PR #340, opened 2019-01-05
[GraphIndex] Detects duplicate edges on simple graphs

### LINKED ISSUE #112: Forbid multiple edges when multigraph is False in DGLGraph
If I understand correctly, when a `dgl.DGLGraph` object is created, by default it is not a multigraph. However when multiple edges are added with the same end nodes and direction, no warnings or errors are raised and it gets multiple edges created as if the graph object is a multigraph.

BarclayII:
## Description
Detects and fails on duplicate edges when adding to simple graphs (fixes #112).  The solution comes from igraph where the successor vertex IDs are always sorted.

The performance for the original and proposed implementation is in https://hackmd.io/lVM6zI6VRIOq4-fzE8vuNw?both

If the performance is not satisfactory then I can try using hashmaps instead.

EDIT: also fixes compiler warnings


jermainewang:
How about let's hold this for a while, wait for the immutable graph PR?

BarclayII:
Sure.  Closing this for now.

BarclayII:
Reopening this since immutable graph is merged.

CMakeLists.txt
```diff
@@ -41,8 +41,8 @@ if(MSVC)
 else(MSVC)
   include(CheckCXXCompilerFlag)
   check_cxx_compiler_flag("-std=c++11"    SUPPORT_CXX11)
-  set(CMAKE_C_FLAGS "-O2 -Wall -fPIC ${CMAKE_C_FLAGS}")
-  set(CMAKE_CXX_FLAGS "-O2 -Wall -fPIC -std=c++11 ${CMAKE_CXX_FLAGS}")
+  set(CMAKE_C_FLAGS "-g -O0 -Wall -fPIC ${CMAKE_C_FLAGS}")
+  set(CMAKE_CXX_FLAGS "-g -O0 -Wall -fPIC -std=c++11 ${CMAKE_CXX_FLAGS}")
```
zheng-da: why turn off the optimization?

BarclayII: That wasn't intentional

zheng-da:
so there isn't performance impact?
could you put the performance results side by side so we can easily see the performance impact. I think we only need to compare the performance for modifying 1,000,000 nodes/edges.

BarclayII:
Modifying an existing graph of 1M nodes and 5M edges is significantly slower:

|# nodes|# edges|Old implementation|New implementation|
|-|-|-|-|
|+100|+500|0.0004|0.0090|
|+1K|+5K|0.0028|0.0141|
|+10K|+50K|0.0306|0.0501|
|+100K|+500K|0.3362|0.5594|
|+1M|+5M|2.5600|4.0326|

jermainewang:
What do you think of the solution in #419 ? namely having `dgl.to_simple` to remove multi-edges. I feel this could avoid the performance/usability dilemma by having the call controlled by users.

BarclayII:
Sure.  What if someone wants to add edges batch by batch while keeping it a simple graph (in the case of streaming graphs)?  Just a theoretical question though - I don't know which situation is more common

jermainewang:
> Sure. What if someone wants to add edges batch by batch while keeping it a simple graph (in the case of streaming graphs)? Just a theoretical question though - I don't know which situation is more common

That's good consideration. For streaming graph, my guess is some new solution is needed. We may need to read some papers about streaming graph for some solutions. For the time being, let's use a naive solution.

BarclayII:
Agreed.

Issue #334, opened 2019-01-03
Creating DGLGraph from non-consecutive int nodes


sufeidechabei:
## To Reproduce

```
    train_remove_length = len(np.where(data.train_mask== 1)[0]) * 0.1
    valid_remove_length = len(np.where(data.val_mask==1)[0]) * 0.1
    test_remove_length = len(np.where(data.test_mask== 1)[0]) * 0.1
    remove_train = np.random.choice(range(140), int(train_remove_length), replace=False)
    remove_valid = np.random.choice(range(200, 500), int(valid_remove_length), replace=False)
    remove_test = np.random.choice(range(500, 1500), int(test_remove_length), replace=False)
    total_delete_list = list(remove_train) + list(remove_valid) + list(remove_test)
    data.graph.remove_nodes_from(total_delete_list)
    graph = DGLGraph(data.graph)
```
```
dgl._ffi.base.DGLError: Error while creating graph from input of type "<class 'networkx.classes.digraph.DiGraph'>".
```

## Expected behavior

networkx graph to dgl graph

## Environment

 - DGL Version (e.g., 1.0): 0.1.3
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): Pytorch 0.4.1
 - OS (e.g., Linux): Linux
 - How you installed DGL (`conda`, `pip`, source): pip
 - Python version:3.6.7
 - CUDA/cuDNN version (if applicable): 9.0
 - GPU models and configuration (e.g. V100): V100
 - Any other relevant information:networkx(2.2)

mufeili:
When you perform `data.graph.remove_nodes_from(total_delete_list)`, the nodes left will not be re-indexed. In `dgl` we are always indexing nodes with `0, ..., num_nodes-1`. Therefore you can end up asking to add edges between nodes dgl consider not to exist. A temporary workaround can be done by re-indexing the nodes and maintaining a mapping between the original node index and the node index after re-indexing.

jermainewang:
We could provide a utility function similar to networkx's [`convert_node_labels_to_integers`](https://networkx.github.io/documentation/stable/reference/generated/networkx.relabel.convert_node_labels_to_integers.html#networkx.relabel.convert_node_labels_to_integers), and give a better error message for this.

jermainewang:
@mufeili would you help on this?

mufeili:
Sure.

mufeili:
Fixed in #341

Issue #306, opened 2018-12-14
How can I implement the function dgl.max_nodes?


zztant:
A key operation in PointNet for point cloud processing.

VoVAllen:
Do you mean in batched graph or in only one graph?

jermainewang:
Hi @zztant , please ask questions in our discussion forum https://discuss.dgl.ai. We try to keep our issue list only about bugs and features. Thank you.

PR #341, opened 2019-01-06
[Feature] Max readout and consecutive labeling for networkx

mufeili:
## Checklist
- [x] Code is well-documented

## Changes
- As requested in #306 , a naive implementation for max readout on nodes and edges is provided with `max_nodes` and `max_edges` in `batched_graph.py`. A more efficient implementation will require `segment_max`. The test cases are provided in `tests/pytorch/test_readout.py`
- As discussed in #334 , if we want to create a `DGLGraph` object from a networkx graph with non-consecutive node labels, we relabel the nodes using consecutive integers first before converting it into a `DGLGraph`. The changes are `from_networkx` in `graph.py`, `graph_index.py`, `immutable_graph_index.py`. The test cases are provided in `tests/compute/test_basics.py`.
- In `from_networkx` of `graph.py`, I also convert the networkx graph to be directed first, otherwise only half of the `DGLGraph` edges will have features.


yzh119:
@mufeili does the `segment_max` you mentioned similar to `segment_sum` in tensorflow?

mufeili:
@yzh119 Yeah, see [doc](https://www.tensorflow.org/api_docs/python/tf/math/segment_max) and [code](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/segment_reduction_ops.cc).

BarclayII:
I think later on we will need to replace the for loop in `max_nodes` etc with a custom op, e.g. `scatter_max`.

mufeili:
@BarclayII I agree, this is just a very naive workaround.

python/dgl/graph.py
```diff
@@ -1165,6 +1170,10 @@ def from_networkx(self, nx_graph, node_attrs=None, edge_attrs=None):
                 [2., 2., 2., 2.],
                 [1., 1., 1., 1.]])
         """
+        nx_graph = nx_graph.to_directed()
+        # Relabel nodes using consecutive integers
+        nx_graph = nx.convert_node_labels_to_integers(nx_graph)
```
jermainewang: Not needed. The graph index already calls it (see below).

python/dgl/graph.py
```diff
@@ -1137,7 +1141,8 @@ def from_networkx(self, nx_graph, node_attrs=None, edge_attrs=None):
         Parameters
```
jermainewang: Describe if the reordering happens, what ordering will be used. From nx: "inherit node ordering from G.nodes() “sorted” : inherit node ordering from sorted(G.nodes())"

python/dgl/graph.py
```diff
@@ -1165,6 +1170,10 @@ def from_networkx(self, nx_graph, node_attrs=None, edge_attrs=None):
                 [2., 2., 2., 2.],
                 [1., 1., 1., 1.]])
         """
+        nx_graph = nx_graph.to_directed()
```
jermainewang: What will happen for the node/edge attributes after `to_directed()`?

python/dgl/graph.py
mufeili: When we convert networkx graph into dgl graph with node features, it assumes that nodes are labeled with consecutive integers due to `for nid in range(self.number_of_nodes()):` in line 1193. Initially I thought relabeling the graph will help circumvent not only this issue for node feature but also for edge feature. It turns out that we cannot circumvent line 1208 anyway and I guess you are right that we should remove line 1173 and change line 1193 to be `for nid in nx_graph.nodes()` instead.

python/dgl/graph.py
mufeili: `to_directed` returns a deep copy of the original undirected graph where all `(node, node_attrs)` and `(edge, edge_attrs)` pairs are preserved. Besides, originally the undirected graph only has `u, v` and its associated attributes. `to_directed` creates `v, u` edges whose attributes are the same as `u, v`. See [`to_directed` source code](https://github.com/networkx/networkx/blob/master/networkx/classes/graph.py#L1602). This is useful as otherwise we will get `None` attributes for the new `v, u` edges and an error will occur in [L1178 of graph.py](https://github.com/dmlc/dgl/blob/master/python/dgl/graph.py#L1178) as you are concatenating tensors with None type objects.

mufeili: I will add a condition check as the function seems still returns a deepcopy even if the graph is already directed.

jermainewang: Got that. Maybe add some comments about this behavior?

python/dgl/graph.py
jermainewang: > It turns out that we cannot circumvent line 1208 anyway

Why is that? Also, technically `for nid in nx_graph.nodes()` does not assume any order so `for nid in range(self.number_of_nodes())` is clearer.

python/dgl/graph.py
mufeili: Comments added. I also added the condition of checking whether a networkx graph is directed in `graph_index.py` and `immutable_graph_index.py`.

python/dgl/graph.py
mufeili: 1. @ylfdq1118 Could you please take a look at whether my understanding is correct for the rest of item 1. From [L1196-1199 of graph.py](https://github.com/dmlc/dgl/blob/master/python/dgl/graph.py#L1196), we assume that there exists a prior specification of edge ids and the edge attributes are stored in the rows specified by these edge ids. @jermainewang Then if the edge ids are broken after we remove some nodes there will still be a problem. I guess this is not something that we need to take care.
2. If you do `for nid in range(self.number_of_nodes())`, since we did not relabel the nodes with consecutive integers in the `from_networkx` in `graph.py`, a KeyError can occur. I can do `for nid in sorted(nx_graph.nodes())` if this works for you.

jermainewang: I see what happens here. Then, how about this?
(1) Relabel the networkx in `graph.py` because the order is required to set node features.
(2) `graph_index.py` will not relabel the nx graph and assume it is correct.

The benefit is we don't need to sort the nodes twice. As long as the `graph_index.from_networkx` is only called by `graph.from_networkx`, it would be fine. What do you think?

mufeili: Sounds good to me. Done.

python/dgl/graph.py
```diff
@@ -1181,7 +1194,7 @@ def _batcher(lst):
         if node_attrs is not None:
             # mapping from feature name to a list of tensors to be concatenated
             attr_dict = defaultdict(list)
-            for nid in range(self.number_of_nodes()):
+            for nid in nx_graph.nodes():
```
jermainewang: Then revert this since the graph has been relabeled.

python/dgl/graph.py
jermainewang: For the edge id, maybe we could loop over the list and check whether they is a `None`? and if so, we raise error about the consecutive edge id blah blah.

python/dgl/graph.py
mufeili: Done.

python/dgl/graph.py
mufeili: Yeah I can do that. `attrs['id']` can also lead to an index out of range error, shall we also handle that?

jermainewang: Yeah, that's a good idea.

mufeili: Done.

jermainewang:
Great! Thanks Murphy.

PR #342, opened 2019-01-07
[Feature] Reimplement Immutable graph index in DGL

zheng-da:
## Description
This is to reimplement immutable graph index in DGL. This is a framework-agnostic implementation.

## Checklist
- [x] Code is well-documented


include/dgl/immutable_graph.h
```diff
@@ -0,0 +1,480 @@
...
+bool binary_search(ForwardIt first, ForwardIt last, const T& value) {
+  first = std::lower_bound(first, last, value);
+  return (!(first == last) && !(value < *first));
+}
+
+/*!
+ * \brief Base dgl immutable graph index class.
+ *
+ */
+class ImmutableGraph {
```
jermainewang: I think we could define an interface for graph interface and ImmutableGraph simply inherits that. As a result, we don't need to:
(1) Expose all the related CAPI again.l
(2) The GraphIndex class in python could be resued for immutable graph.
Thoughts?

zheng-da: I was struggling with it. There are a few things that these two indices are incompatible.
GraphIndex supports:
* graph modification,
* FindEdge (this operation is very slow in CSR, so I didn't implement it).
* something like this `const std::vector<dgl_id_t>& SuccVec` is not possible for ImmutableGraphIndex.

ImmutableGraphIndex supports:
* sampling,
* Get CSR arrays,

We could merge the CAPIs for the basic operations.

jermainewang: For graph modification, we could throw errors.

For `FindEdge`, it's OK to have very naive implementation.

For `SuccVec`, we could change it. It is used to implement graph traversal (not exposed to CAPI at all). What would you suggest? Please read how it is used in graph traversal.

For `sampling` and Get CSR arrays, we could put it in the new graph class not in the base class. And when the graph pointer is passed down from python, we just cast it to the new class pointer.

Strongly suggest we define a common abstract right now. It would be very painful to do that later.

Also, change `ImmutableGraphIndex` to `CSRGraphIndex`? We might have more formats later.

include/dgl/immutable_graph.h
```diff
@@ -0,0 +1,480 @@
+/*!
+ *  Copyright (c) 2018 by Contributors
+ * \file dgl/graph.h
```
szha: need to correct the filename

include/dgl/immutable_graph.h
BarclayII: I'm in line with @jermainewang .

Also is there a particular reason why (Mutable)GraphIndex *does not* support samling?  In fact I think it makes sense to support sampling there as well, since that will be helpful for learning on large dynamic graphs in the (far) future.

zheng-da: Sampling requires very efficient implementation. Mutable graph index doesn't allow efficient implementation.

zheng-da: I'm fine with implementing a common interface.

Actually ImmutableGraphIndex contains both CSR and CSC for the graph.

zheng-da: The main problem with `SuccVec` is that it returns a reference. ImmutableGraphIndex can't return a reference. One option is to return `std::vector` directly. however, this may cause unnecessary overhead for GraphIndex.

jermainewang: How about an iterator?

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,944 @@
...
+ * \brief DGL immutable graph index implementation
+ */
+
+#include <stdlib.h>
+#include <dgl/immutable_graph.h>
+
+#ifdef _MSC_VER
+// TODO(zhengda) it seems MS compiler doesn't have rand_r. Let's
+// use rand for now. This may not have good performance because rand has
+// a global variable shared by all threads.
```
BarclayII: Not the case for Windows; [VC Runtime is always multithreaded](https://social.msdn.microsoft.com/Forums/vstudio/en-US/3263ab0b-da3f-4737-9d80-0788a5b426db/randr-identifier-not-found?forum=vcgeneral).

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,944 @@
...
+ private:
+  int vec_size_;  // sample size
+  int bit_len_;   // bit size
+  int limit_;
+  std::vector<float> heap_;
+};
+
+static void RandomSample(size_t set_size,
+                         size_t num,
+                         std::vector<size_t>* out,
```
BarclayII: Any particular reason why a pointer instead of a reference is used here (and below)?

include/dgl/immutable_graph.h
BarclayII: @zheng-da I assume that "efficient implementation" refers to storing the graph in CSR.  Could you point out which part of the sampling depends on CSR in particular?  I went through the code and didn't find it..

zheng-da: The efficiency comes from the format. CSR stores data in a single piece of contiguous memory, which makes memory management much easier. The mutable graph index stores the adjacnecy lists in individual vectors, which results in many small memory allocation. As far as I remember, subgraph construction is significantly faster on CSR than on the mutable graph index.

BarclayII: In this case, should we switch to the single contiguous memory solution for the Mutable graph index as well?  igraph goes this way, and therefore adding edges takes linear time of the number of edges of the new graph.  I guess this is okay since we don't have a lot of separate edge additions in general.

zheng-da: Our original graph index was designed for general purpose. I think the current impl is quite reasonable. I think it's better to add a function that converts a general graph index into a read-only graph index if someone really wants to sample from the general graph index.

zheng-da:
do you guys have any other comments?

python/dgl/graph_index.py
```diff
@@ -127,6 +140,7 @@ def number_of_nodes(self):
         int
             The number of nodes
         """
+        assert self._handle is not None
```
jermainewang: The assertion is not needed. Or you could think that all APIs require such assertion which means we should assume the _handle is never None after initialization.

python/dgl/graph_index.py
```diff
@@ -703,12 +745,18 @@ def from_scipy_sparse_matrix(self, adj):
...
         adj : scipy sparse matrix
         """
-        self.clear()
-        self.add_nodes(adj.shape[0])
+        assert isinstance(adj, (scipy.sparse.csr_matrix, scipy.sparse.coo_matrix)), \
+                "The input matrix has to be a SciPy sparse matrix."
+        if not self.is_readonly():
+            self.clear()
+        # what if the adj matrix isn't symmetric.
+        num_nodes = max(adj.shape[0], adj.shape[1])
```
jermainewang: I think we should raise error when the given matrix is not square.

python/dgl/graph_index.py
```diff
@@ -900,12 +952,14 @@ def create_graph_index(graph_data=None, multigraph=False, readonly=False):
         return graph_data
 
     if readonly:
-        return create_immutable_graph_index(graph_data)
-
-    handle = _CAPI_DGLGraphCreate(multigraph)
-    gidx = GraphIndex(handle)
+        gidx = GraphIndex(None, multigraph, readonly)
```
jermainewang: To avoid this awkward `None`, the solution is to first process all the graph data. There are functions to convert different types of graph data to `src, dst, edge_id, num_nodes` first, so the handler could be created and then the python shim `GraphIndex`. Leave this for now. But would you put a `FIXME` here?

python/dgl/graph_index.py
```diff
@@ -933,4 +987,29 @@ def create_graph_index(graph_data=None, multigraph=False, readonly=False):
 
     return gidx
 
+
 _init_api("dgl.graph_index")
+
+_NEIGHBOR_SAMPLING_APIS = {
```
jermainewang: Put a "FIXME" or "TODO" about supporting variable-length returns.

tests/mxnet/test_graph_index.py
```diff
@@ -1,12 +1,27 @@
 import os
```
jermainewang: I didn't see unittest with pytorch backend.

include/dgl/graph_interface.h
```diff
@@ -0,0 +1,383 @@
...
+ */
+class GraphInterface {
+ public:
+  /* \brief structure used to represent a list of edges */
+  typedef struct {
+    /* \brief the two endpoints and the id of the edge */
+    IdArray src, dst, id;
+  } EdgeArray;
+
+  typedef std::shared_ptr<GraphInterface> ptr;
```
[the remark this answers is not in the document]
zheng-da: I feel uncomfortable to use raw pointer. I create a Reset to move data to an object pointed by a raw pointer when it's necessary.

include/dgl/graph_interface.h
```diff
@@ -0,0 +1,383 @@
...
+  std::string msg;
+ public:
+  explicit NotImplemented(const std::string &name) {
+    this->msg = name + " isn't implemented";
+  }
+
+  virtual const char* what() const noexcept {
+    return msg.c_str();
+  }
+};
```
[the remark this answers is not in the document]
zheng-da: which `base.h`?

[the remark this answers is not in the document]
zheng-da: `./third_party/dmlc-core/include/dmlc/base.h` is in another repo. Adding code there makes it difficult to merge the PR.

include/dgl/graph_interface.h
```diff
@@ -0,0 +1,383 @@
...
+   * \param vid The vertex id.
+   * \return the in edge id vector iterator pair.
+   */
+  virtual DGLIdIters InEdgeVec(dgl_id_t vid) const = 0;
+
+  /*!
+   * \brief Reset the data in the graph and move its data to the returned graph object.
+   * \return a raw pointer to the graph object.
+   */
+  virtual GraphInterface *Reset() = 0;
```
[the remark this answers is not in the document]
zheng-da: https://github.com/dmlc/dgl/pull/342/files#diff-edd00dbcb0c7e5f13164c417fede7156R56
we can't use move constructor. We don't know the actual type.

include/dgl/graph_interface.h
```diff
@@ -0,0 +1,383 @@
...
+
+  /*!
+   * \brief Get the adjacency matrix of the graph.
+   *
+   * By default, a row of returned adjacency matrix represents the destination
+   * of an edge and the column represents the source.
+   * \param transpose A flag to transpose the returned adjacency matrix.
+   * \param fmt the format of the returned adjacency matrix.
+   * \return a vector of three IdArray.
```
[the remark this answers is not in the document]
zheng-da: std::vector doesn't have large overhead, considering how much computation is required by this method. 
the main reason of using std::vector is that the number of returned arrays changes for different formats.

include/dgl/graph_interface.h
```diff
@@ -0,0 +1,383 @@
...
+struct SampledSubgraph: public Subgraph {
+  /*!
+   * \brief the layer of a sampled vertex in the subgraph.
+   */
+  IdArray layer_ids;
+  /*!
+   * \brief the probability that a vertex is sampled.
+   */
+  runtime::NDArray sample_prob;
+};
```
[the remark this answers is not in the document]
zheng-da: not sure what you mean. this data structure is used to store the sampled results. it will be used in graph_apis.cc to return the results to python.

include/dgl/immutable_graph.h
```diff
@@ -0,0 +1,531 @@
...
+class ImmutableGraph: public GraphInterface {
+ public:
+  typedef struct {
+    IdArray indptr, indices, id;
+  } CSRArray;
+
+  struct edge {
+    dgl_id_t end_points[2];
+    dgl_id_t edge_id;
+  };
```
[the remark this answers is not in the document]
zheng-da: we can't use `src` and `dst` here. I want to have a single piece of code that works for src or dst.
https://github.com/dmlc/dgl/pull/342/files#diff-dc95a0d3acb4ca34468e75c512dd42dfR185

include/dgl/immutable_graph.h
```diff
@@ -0,0 +1,531 @@
...
+
+  struct csr {
```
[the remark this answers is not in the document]
zheng-da: If we move it to .cc file, a lot of small methods in ImmutableGraphIndex can't be implemented in .h.

include/dgl/immutable_graph.h
```diff
@@ -0,0 +1,531 @@
...
+    EdgeArray GetEdges(dgl_id_t vid) const;
+    EdgeArray GetEdges(IdArray vids) const;
+    std::pair<const dgl_id_t *, const dgl_id_t *> GetIndexRef(dgl_id_t v) const {
+      int64_t start = indptr[v];
+      int64_t end = indptr[v + 1];
+      return std::pair<const dgl_id_t *, const dgl_id_t *>(&indices[start], &indices[end]);
+    }
+    csr::ptr Transpose() const;
+    std::pair<csr::ptr, IdArray> VertexSubgraph(IdArray vids) const;
+    static csr::ptr from_edges(std::vector<edge> *edges, int sort_on, int64_t num_nodes);
```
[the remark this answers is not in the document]
zheng-da: this method needs to sort the input edges. we don't want to make a copy of the edges.

include/dgl/immutable_graph.h
```diff
@@ -0,0 +1,531 @@
...
+  /*!
+   * \return whether the graph is read-only
+   */
+  virtual bool IsReadonly() const {
+    return true;
+  }
+
+  /*! \return the number of vertices in the graph.*/
+  uint64_t NumVertices() const {
+    if (in_csr_)
```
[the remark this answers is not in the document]
zheng-da: this is right. we don't want to call `GetInCSR` all the time. `GetInCSR` generates in_csr if it doesn't exist. When an immutable graph index is generated, there must exist either in_csr_ or out_csr_. if in_csr_ doesn't exist, we can get the info from out_csr_

include/dgl/immutable_graph.h
```diff
@@ -0,0 +1,531 @@
...
+      return in_csr_->NumVertices();
+    else
+      return out_csr_->NumVertices();
+  }
+
+  /*! \return the number of edges in the graph.*/
+  uint64_t NumEdges() const {
+    if (in_csr_)
```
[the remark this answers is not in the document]
zheng-da: the same comment above.

include/dgl/immutable_graph.h
```diff
@@ -0,0 +1,531 @@
...
+    if (!HasVertex(src) || !HasVertex(dst)) return false;
+    if (this->in_csr_) {
+      auto pred = this->in_csr_->GetIndexRef(dst);
+      return binary_search(pred.first, pred.second, src);
+    } else {
+      assert(this->out_csr_);
+      auto succ = this->out_csr_->GetIndexRef(src);
+      return binary_search(succ.first, succ.second, dst);
+    }
+  }
```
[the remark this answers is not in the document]
zheng-da: why? what is the standard of having code in .h or .cc?

include/dgl/immutable_graph.h
```diff
@@ -0,0 +1,531 @@
...
+  /*
+   * When we get in csr or out csr, we try to get the one cached in the structure.
+   * If not, we transpose the other one to get the one we need.
+   */
+  csr::ptr GetInCSR() const {
+    if (in_csr_) {
+      return in_csr_;
+    } else {
+      assert(out_csr_ != nullptr);
+      const_cast<ImmutableGraph *>(this)->in_csr_ = out_csr_->Transpose();
```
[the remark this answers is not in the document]
zheng-da: when a subgraph is generated, i only construct in_csr or out_csr. in a lot of cases, we only need one of them. subgraph is generated frequently and its lifetime is short. it's really unnecessary to construct both.

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,942 @@
...
+  int64_t *indptr_data = static_cast<int64_t*>(indptr->data);
+  dgl_id_t* indices_data = static_cast<dgl_id_t*>(indices->data);
+  dgl_id_t* eid_data = static_cast<dgl_id_t*>(eids->data);
+  std::copy(out_csr->indptr.begin(), out_csr->indptr.end(), indptr_data);
+  std::copy(out_csr->indices.begin(), out_csr->indices.end(), indices_data);
+  std::copy(out_csr->edge_ids.begin(), out_csr->edge_ids.end(), eid_data);
+  return CSRArray{indptr, indices, eids};
+}
+
+////////////////////////////// Graph Sampling ///////////////////////////////
```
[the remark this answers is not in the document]
zheng-da: let's do it in another PR.

zheng-da:
finally addressed all comments.

include/dgl/graph_interface.h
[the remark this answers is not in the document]
BarclayII: I think @jermainewang meant to create a new one.  Should that be in `c_api_common.h`?

include/dgl/immutable_graph.h
```diff
@@ -0,0 +1,531 @@
...
+#include "runtime/ndarray.h"
+#include "graph_interface.h"
+
+namespace dgl {
+
+template<class ForwardIt, class T>
+bool binary_search(ForwardIt first, ForwardIt last, const T& value) {
+  first = std::lower_bound(first, last, value);
+  return (!(first == last) && !(value < *first));
+}
```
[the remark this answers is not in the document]
BarclayII: That's surprising.  Isn't it a C++98 standard?  Which complier doesn't support that?

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,942 @@
...
+      this->other_end = other_end;
+    }
+    bool operator()(const edge &e1, const edge &e2) {
+      if (e1.end_points[sort_on] == e2.end_points[sort_on])
+        return e1.end_points[other_end] < e2.end_points[other_end];
+      else
+        return e1.end_points[sort_on] < e2.end_points[sort_on];
+    }
+  };
+  std::sort(edges->begin(), edges->end(), compare(sort_on, other_end));
```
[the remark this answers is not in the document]
BarclayII: You can put all the buckets in a single vector.

src/graph/immutable_graph.cc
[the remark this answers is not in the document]
BarclayII: I would prefer do it now since layer sampler & random walk sampler would probably stay there as well.  @jermainewang What do you think?

include/dgl/graph_interface.h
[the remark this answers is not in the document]
jermainewang: Could forward declare it in the header and define it in `graph_apis.cc`.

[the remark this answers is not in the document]
zheng-da: i don't think it should be defined in graph_api.cc. the structure is also used by immutable_graph.cc

include/dgl/immutable_graph.h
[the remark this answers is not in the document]
zheng-da: i don't remember exactly. one of the compilers fails. most likely the MS compiler.

python/dgl/contrib/sampling/sampler.py
```diff
@@ -200,7 +195,7 @@ def __iter__(self):
 
 def NeighborSampler(g, batch_size, expand_factor, num_hops=1,
                     neighbor_type='in', node_prob=None, seed_nodes=None,
-                    shuffle=False, num_workers=1, max_subgraph_size=None,
+                    shuffle=False, num_workers=1,
                     return_seed_id=False, prefetch=False):
```
BarclayII: Also I assume the note "This method currently only supports MXNet backend" outdated?

src/graph/immutable_graph.cc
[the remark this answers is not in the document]
zheng-da: i feel radix sort is difficult here. each element is a pair. since radix sort doesn't exist in STL and the general sort usually works well, I'll leave the code as it is.

include/dgl/immutable_graph.h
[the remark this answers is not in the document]
jermainewang: Still, the const cast should be avoided until the last moment. So what's the way out. One solution we could borrow is from [boost graph](https://www.boost.org/doc/libs/1_64_0/libs/graph/doc/adjacency_list.html), where they have **three** types of directions: directed, undirected and bidirected. In your case, you want the subgraph generated to be simply "directed". In the general case, the graph should be "bidirected".

For now, I think we should document it clearly. How the subgraph is constructed? I remembered the constructor requires both in_csr and our_csr. Then at least we should have a constructor overload that accepts only in_csr (the docstring can also go there). Also, can user chooses which csr to use at all?

include/dgl/immutable_graph.h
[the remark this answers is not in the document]
BarclayII: Let's leave it here and I'll inspect it in my multigraph PR.

src/graph/immutable_graph.cc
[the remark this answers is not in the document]
BarclayII: Fair enough.

include/dgl/immutable_graph.h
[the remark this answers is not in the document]
jermainewang: It should be standard. https://en.wikipedia.org/wiki/Algorithm_(C%2B%2B).

include/dgl/graph_interface.h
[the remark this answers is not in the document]
zheng-da: I agree. we should move the sampling code to sampler.h and sampler.cc. i can refactor the code when i work on neighbor sampling (this is my next task).

include/dgl/immutable_graph.h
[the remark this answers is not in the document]
zheng-da: make sense. i'll move the code tomorrow.

include/dgl/immutable_graph.h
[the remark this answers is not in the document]
zheng-da: in_csr and out_csr stores the same edge. it's just stored them in different orders, which is required for different operations.

how a subgraph is constructed should be documented.

python/dgl/graph_index.py
zheng-da: edge list input handles a similar case.
https://github.com/dmlc/dgl/pull/342/files/3abbf81f267a61a3fcc1a69f55aaaa88c29a6f94#diff-57dab646fc76271b216df6298141ebbeR776

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+    for (size_t j = 0; j < row_len; ++j) {
+      const dgl_id_t oldsucc = neigh_idx[j];
+      const dgl_id_t eid = eids[j];
+      Collect(oldsucc, eid, new_neigh_idx, orig_eids);
+    }
+  }
+};
+
+std::pair<ImmutableGraph::CSR::Ptr, IdArray> ImmutableGraph::CSR::VertexSubgraph(
+    IdArray vids) const {
```
jermainewang: Check valid id array

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+  const int64_t len = vids->shape[0];
+
+  HashTableChecker def_check(vid_data, len);
+  // check if varr is sorted.
+  CHECK(std::is_sorted(vid_data, vid_data + len)) << "The input vertex list has to be sorted";
+
+  // Collect the non-zero entries in from the original graph.
+  std::vector<dgl_id_t> orig_edge_ids;
+  orig_edge_ids.reserve(len * 50);
+  auto sub_csr = std::make_shared<CSR>(len, len * 50);
```
jermainewang: What is 50 here?

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+/*
+ * This uses a hashtable to check if a node is in the given node list.
+ */
+class HashTableChecker {
+  std::unordered_map<dgl_id_t, dgl_id_t> oldv2newv;
+  // This bitmap is used as a bloom filter to remove some lookups.
+  // Hashtable is very slow. Using bloom filter can significantly speed up lookups.
+  Bitmap map;
+
+  void Collect(const dgl_id_t old_id, const dgl_id_t old_eid,
```
jermainewang: Could you give some brief docstring about what is this Collect and also the CollectOnRow is doing? Explain what's the meaning of the args. The docstring does not need to be as complete as the one in the interface, but it will help developers a lot later.

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+      this->sort_on = sort_on;
+      this->other_end = other_end;
+    }
+    bool operator()(const Edge &e1, const Edge &e2) {
+      if (e1.end_points[sort_on] == e2.end_points[sort_on])
+        return e1.end_points[other_end] < e2.end_points[other_end];
+      else
+        return e1.end_points[sort_on] < e2.end_points[sort_on];
+    }
+  };
```
jermainewang: You are using local structure here. Is there any particular reason? If not, put it in the anonymous namespace in this cc file.

src/graph/immutable_graph.cc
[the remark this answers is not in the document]
jermainewang: I actually agree with @BarclayII that radix sort is better since all these values are integers and it is easy to compute hash function. I suggest put a TODO here. I'm thinking about doing another PR that tries to enable graph ops on immutable graph too. I could incorporate this change in that PR.

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+
+  IdArray rst_eids = IdArray::Empty({static_cast<int64_t>(orig_edge_ids.size())},
+                                    DLDataType{kDLInt, 64, 1}, DLContext{kDLCPU, 0});
+  dgl_id_t* eid_data = static_cast<dgl_id_t*>(rst_eids->data);
+  std::copy(orig_edge_ids.begin(), orig_edge_ids.end(), eid_data);
+
+  return std::pair<ImmutableGraph::CSR::Ptr, IdArray>(sub_csr, rst_eids);
+}
+
+ImmutableGraph::CSR::Ptr ImmutableGraph::CSR::FromEdges(std::vector<Edge> *edges,
```
jermainewang: I don't know whether you've put docstring in the CSR structure (seems that you haven't). For me, it is not a straight-forward convert-from-edge that takes a list of edges. Just briefly explain what `sort_on` means would help a lot.

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+  t->indices.resize(edges->size());
+  t->edge_ids.resize(edges->size());
+  for (size_t i = 0; i < edges->size(); i++) {
+    t->indices[i] = edges->at(i).end_points[other_end];
+    CHECK(t->indices[i] < num_nodes);
+    t->edge_ids[i] = edges->at(i).edge_id;
+    dgl_id_t vid = edges->at(i).end_points[sort_on];
+    CHECK(vid < num_nodes);
+    while (vid > 0 && t->indptr.size() <= static_cast<size_t>(vid))
+      t->indptr.push_back(i);
```
jermainewang: brace

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+    CHECK(t->indptr.size() == vid + 1);
+  }
+  while (t->indptr.size() < num_nodes + 1)
+    t->indptr.push_back(edges->size());
```
jermainewang: brace

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+    const dgl_id_t *eid_begin = &edge_ids[indptr[i]];
+    for (size_t j = 0; j < GetDegree(i); j++) {
+      Edge e;
+      e.end_points[0] = i;
+      e.end_points[1] = indices_begin[j];
+      e.edge_id = eid_begin[j];
+      edges[indptr[i] + j] = e;
+    }
+  }
+  return FromEdges(&edges, 1, NumVertices());
```
jermainewang: Magic number `1`? Use a bool?

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+}
+
+ImmutableGraph::ImmutableGraph(IdArray src_ids, IdArray dst_ids, IdArray edge_ids, size_t num_nodes,
+                               bool multigraph) : is_multigraph_(multigraph) {
```
jermainewang: Check valid id array

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+  int64_t len = src_ids->shape[0];
```
jermainewang: const

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+  const int64_t len = succ.second - succ.first;
+  IdArray rst = IdArray::Empty({len}, DLDataType{kDLInt, 64, 1}, DLContext{kDLCPU, 0});
+  dgl_id_t* rst_data = static_cast<dgl_id_t*>(rst->data);
+
+  std::copy(succ.first, succ.second, rst_data);
+  return rst;
+}
+
+std::pair<const dgl_id_t *, const dgl_id_t *> ImmutableGraph::GetInEdgeIdRef(dgl_id_t src,
+                                                                         dgl_id_t dst) const {
```
jermainewang: I wonder can we return the DGLIdIter you defined before?

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+}
+
+IdArray ImmutableGraph::EdgeId(dgl_id_t src, dgl_id_t dst) const {
+  CHECK(HasVertex(src) && HasVertex(dst)) << "invalid edge: " << src << " -> " << dst;
+
+  std::pair<const dgl_id_t *, const dgl_id_t *> edge_ids;
+  if (in_csr_)
+    edge_ids = GetInEdgeIdRef(src, dst);
+  else
+    edge_ids = GetOutEdgeIdRef(src, dst);
```
jermainewang: brace

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+
+  return rst;
+}
+
+ImmutableGraph::EdgeArray ImmutableGraph::EdgeIds(IdArray src_ids, IdArray dst_ids) const {
+  CHECK(IsValidIdArray(src_ids)) << "Invalid src id array.";
+  CHECK(IsValidIdArray(dst_ids)) << "Invalid dst id array.";
+  const auto srclen = src_ids->shape[0];
+  const auto dstlen = dst_ids->shape[0];
+  int64_t i, j;
```
jermainewang: Are you using i and j else where? why not put them in loop variables?

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+        "invalid edge: " << src_id << " -> " << dst_id;
+
+    std::pair<const dgl_id_t *, const dgl_id_t *> edges;
+    if (this->in_csr_)
+      edges = this->GetInEdgeIdRef(src_id, dst_id);
+    else
+      edges = this->GetOutEdgeIdRef(src_id, dst_id);
+
+    size_t len = edges.second - edges.first;
+    for (size_t i = 0; i < len; i++) {
```
jermainewang: shadow variable `i`

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+    CHECK(HasVertex(src_id) && HasVertex(dst_id)) <<
+        "invalid edge: " << src_id << " -> " << dst_id;
+
+    std::pair<const dgl_id_t *, const dgl_id_t *> edges;
+    if (this->in_csr_)
+      edges = this->GetInEdgeIdRef(src_id, dst_id);
+    else
+      edges = this->GetOutEdgeIdRef(src_id, dst_id);
+
+    size_t len = edges.second - edges.first;
```
jermainewang: put `len` in the loop below

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+    for (size_t i = 0; i < len; i++) {
+        src.push_back(src_id);
+        dst.push_back(dst_id);
+        eid.push_back(edges.first[i]);
+    }
+  }
+
+  int64_t rstlen = src.size();
```
jermainewang: const

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+  auto out_csr = GetOutCSR();
+  // If sorted, the returned edges are sorted by the source Id and dest Id.
+  for (size_t i = 0; i < out_csr->indptr.size() - 1; i++) {
+    std::fill(rst_src_data + out_csr->indptr[i], rst_src_data + out_csr->indptr[i + 1],
+              static_cast<dgl_id_t>(i));
+  }
+  std::copy(out_csr->indices.begin(), out_csr->indices.end(), rst_dst_data);
+  std::copy(out_csr->edge_ids.begin(), out_csr->edge_ids.end(), rst_eid_data);
+
+  // TODO(zhengda) do I need to sort the edges if sorted = false?
```
jermainewang: It should be in the edge id order if sorted=False

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
...
+  std::copy(out_csr->indices.begin(), out_csr->indices.end(), indices_data);
+  std::copy(out_csr->edge_ids.begin(), out_csr->edge_ids.end(), eid_data);
+  return CSRArray{indptr, indices, eids};
+}
+
+std::vector<IdArray> ImmutableGraph::GetAdj(bool transpose, const std::string &fmt) const {
+  if (fmt == "csr") {
+    CSRArray arrs = transpose ? this->GetOutCSRArray() : this->GetInCSRArray();
+    return std::vector<IdArray>{arrs.indptr, arrs.indices, arrs.id};
+  } else if (fmt == "coo") {
```
jermainewang: Isn't this equal to GetEdges?

python/dgl/graph_index.py
zheng-da: I tried to fix it and realized that we may not need to fix it. Think about edge list input. It needs to handle a similar case (max source id might be smaller or larger than the destination id).
https://github.com/dmlc/dgl/pull/342/files/3abbf81f267a61a3fcc1a69f55aaaa88c29a6f94#diff-57dab646fc76271b216df6298141ebbeR776

src/graph/graph_apis.cc
```diff
@@ -33,13 +35,25 @@ PackedFunc ConvertEdgeArrayToPackedFunc(const Graph::EdgeArray& ea) {
...
+}
+
 // Convert Subgraph structure to PackedFunc.
 PackedFunc ConvertSubgraphToPackedFunc(const Subgraph& sg) {
   auto body = [sg] (DGLArgs args, DGLRetValue* rv) {
       const int which = args[0];
       if (which == 0) {
-        Graph* gptr = new Graph();
-        *gptr = std::move(sg.graph);
+        GraphInterface* gptr = sg.graph->Reset();
```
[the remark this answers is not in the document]
zheng-da: I agree that the current solution is ideal. But I don't know what is a better solution.

```C++
GraphInterface* MoveGraph(GraphPtr gptr);
```
How to implement this? we need to explicitly cast the pointer to one of the graph type (Graph or ImmutableGraph). I don't feel comfortable with this solution either.

src/graph/graph_apis.cc
```diff
@@ -53,34 +67,75 @@ PackedFunc ConvertSubgraphToPackedFunc(const Subgraph& sg) {
...
+    };
+  return PackedFunc(body);
+}
+
 }  // namespace
 
-DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLGraphCreate")
+///////////////////////////// Graph API ///////////////////////////////////
+
+DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLGraphCreateMutable")
```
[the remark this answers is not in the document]
zheng-da: We do. `_CAPI_DGLGraphCreate` takes the whole graph in the COO format as input. `_CAPI_DGLGraphCreateMutable` creates an empty mutable graph.

src/graph/immutable_graph.cc
zheng-da: this is the rough estimation the ration of #edges/#nodes. it's not very accurate.

src/graph/graph_apis.cc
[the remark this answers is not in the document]
jermainewang: Wait. Isn't GraphPtr alias of `std::shared_ptr<GraphInterface>`? Then move it to GraphInterface* is simply what's your code is doing right?

src/graph/immutable_graph.cc
jermainewang: What if we have much fewer edges than nodes? Will this cause too much memory request for the allocator? The magic number is hard to maintain.

src/graph/immutable_graph.cc
jermainewang: If it is only 1 or 0, change the type of the argument to bool and use names like `sort_on_src_or_dst` and doc it.

GaiYu0:
I am fine as long as there is no change in major interface.

src/graph/immutable_graph.cc
zheng-da: i see the C++ documentation says that. However, sort all edges in their eid order can be very expensive in immutable graph. It seems this is an common operations. Is this sementics used anywhere?

src/graph/graph_apis.cc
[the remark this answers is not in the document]
zheng-da: gptr still owns the object. if gptr disappears, the memory is gone. We don't have any place to store gptr.

src/graph/immutable_graph.cc
zheng-da: we could assume #edges is the same as #vertices  and let it grow when necessary.

src/graph/immutable_graph.cc
```diff
@@ -0,0 +1,975 @@
+/*!
+ *  Copyright (c) 2018 by Contributors
+ * \file graph/immutable_graph.cc
+ * \brief DGL immutable graph index implementation
+ */
+
+#include <stdlib.h>
```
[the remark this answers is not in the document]
zheng-da: is there any difference?

src/graph/graph_apis.cc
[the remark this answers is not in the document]
jermainewang: See your code:
```c++
virtual GraphInterface *Reset() {
  Graph* gptr = new Graph();
  *gptr = std::move(*this);
  return gptr;
}
# in c_api.cc
GraphInterface* gptr = sg.graph->Reset();  # this sg.graph is a GraphPtr type
```
After Reset, the memory pointed by the shared pointer has already gone. What I proposed is exactly the same as this logic, but just refactor it to a global function.

src/graph/immutable_graph.cc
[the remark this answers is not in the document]
jermainewang: Not much. But using c++ header is more consistent. https://stackoverflow.com/questions/2900785/whats-the-difference-between-cstdlib-and-stdlib-h

src/graph/graph_apis.cc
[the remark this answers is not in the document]
jermainewang: A proposed implementation of this function:
```c++
GraphInterface* MoveGraph(GraphPtr gptr) {
  Graph* ret = new Graph();
  *ret = std::move(*gptr);
  return ret;
}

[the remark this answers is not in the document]
Comment (jermainewang) [reply src/graph/graph_apis.cc]: I see the problem of inheritance here and why you do it in a member function. As I've mentioned in another comment, since python only accepts raw pointers, so in order to fix this, is to return python a **pointer to the shared pointer**:
```
GraphPtr* ret = new GraphPtr;
*ret = sg.graph;
```
**Good thing:** Now if any C++ side code holds the shared pointer, it could be preserved without being reset.
**Bad thing:** Then all other graph apis need to be changed accordingly. Also the pointer to the pointer is kind of confusing.

That's why I directly use `Graph` object previously.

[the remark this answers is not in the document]
zheng-da: The problem is that we can't use `Graph` object here.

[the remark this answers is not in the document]
zheng-da: After discussing different options, I still don't see why my current solution is worse than others. There isn't a good solution here. `Reset` may not be a good name. Something like MoveGraph can be an option.

jermainewang:
Approved. Thanks Da!

Issue #335, opened 2019-01-03
Custom OP


yzh119:
## 🚀 Feature
Custom op.
- [x] Kernel for `masked_matrix_multiplication` (both forward and backward)
- [x] Kernel for `sparse_softmax` (both forward and backward)
- [x] Kernel for `vector-shape spmm` (both forward and backward)
- [x] PyTorch wrapper.
- [x] Multi-Head
- [ ] CPU
- [ ] MXNet
- [ ] ...

## Motivation

Current self-attention implementation in DGL is not efficient and uses too much GPU memory.
Custom Op support is required to accelerate some graph operations like `masked_mm` and `sparse_softmax` used in the self-attention module.

## Alternatives

In the future, there might be elegant solutions but currently, we write custom op for these operations on our own.

## Additional context

You may find my primitive custom op implementations [here(private repo)](https://github.com/yzh119/custom_op_benchmark), note that I've not covered MXNet yet and I hope team members familiar with MXNet would help.

PR #348, opened 2019-01-09
[Model] Improve GAT models

jermainewang:
## Description

A more efficient GAT implementation:
* Use SPMV to compute softmax.
* Batch multiple attention head together for more efficiency.

Results
-------

 | Dataset | Test Accuracy | Time(s) | Baseline#1 times(s) | Baseline#2 times(s) |
| ------- | ------------- | ------- | ------------------- | ------------------- |
| Cora | 84.0% | 0.0127 | 0.0982 (**7.7x**) | 0.0424 (**3.3x**) |
| Citeseer | 70.7% | 0.0123 | n/a | n/a |
| Pubmed | 78.1% | 0.0302 | n/a | n/a |

 * All the accuracy numbers are obtained after 200 epochs.
* The time measures how long it takes to train one epoch.
* All time is measured on EC2 p3.2xlarge instance w/ V100 GPU.
* Baseline#1: [https://github.com/PetarV-/GAT](https://github.com/PetarV-/GAT).
* Baseline#2: [https://github.com/Diego999/pyGAT](https://github.com/Diego999/pyGAT).

## Checklist
- [x] Code is well-documented

@zheng-da @eric-haibin-lin  Cannot implement it in MXNet because MXNet does not support autograd on the sparse matrix for matmul. Based on the experience from GCN, if MXNet is able to support this, MXNet should be faster.


jermainewang:
@sufeidechabei

examples/pytorch/gat/train.py
```diff
@@ -0,0 +1,271 @@
...
+            if self.residual_fc:
+                ret = self.residual_fc(h) + ret
+            else:
+                ret = h + ret
+        return ret
+
+    def edge_attention(self, edges):
+        # an edge UDF to compute unnormalized attention values from src and dst
+        a = self.leaky_relu(edges.src['a1'] + edges.dst['a2'])
+        a = torch.exp(a).clamp(-10, 10)  # use clamp to avoid overflow
```
BarclayII: A caveat is that the softmax is technically still not the "correct and numerically-stable" softmax.  For example, if the pre-softmax scores are `[5, 10, 5, 5]` then the results would become, since here it just clips the whole thing to have a maximum absolute value of 10.

We should leave a comment there.

jermainewang: I see. Is this a reasonable approximation? I found in the torch implementation
https://github.com/Diego999/pyGAT/blob/3301fd68cbf349925d7458d9d47766e48021cf78/layers.py#L111
They even use `torch.exp(-leaky_relu(...))`. Not sure whether this is equivalent.

BarclayII: `-leaky_relu` is actually better (although not perfect due to underflow) since the exponential of that would be always no greater than 1.  The opposite is not since the result will easily overflow.

jermainewang: But is it still correct? It seems to me that they change the activation function to `-leaky_relu`.

BarclayII: The value range of the pre-softmax value doesn't matter.

jermainewang: Hmm, I think `-leaky_relu` is just a little better than `leaky_relu` because the negative slope (the author used 0.2) is less steep than the positive slope (is 1.0). It can still go beyond 10 right?

BarclayII: Oops I was thinking of `relu`, my bad.  Still, I think it will be hard to have the value explode in the negative case.

yzh119:
@ylfdq1118 There are two available methods to avoid `clamp`:
1. Calling an extra `update_all` to calculate the max of pre-softmax attention values.
2. Calculate the max of pre-softmax attention values in reduce function.

If my understanding is right, I think the in the first method we compute the max value in a degree-bucketing way, then we could use SPMV to accelerate the following steps(`src_mul_edge`). But in the second method we execute `src_mul_edge` in a degree-bucketing way, which might be a heavy cost. 

Is that the point why you mentioned adding an extra `update_all` round?

VoVAllen:
According to pytorch's implementations, they minus the max before doing the softmax.
Max operation is expected to be a kernal in the future. Similar to Max Readout operation, they are both segment max(segment reduction) operation. I think it's worth time to investigate how to put this into dgl.

yzh119:
@VoVAllen It's unnecessary to call `segment_max` for softmax, the backward function might be slower compared to a single op.
Custom op is in progress(#335), you may find the sparse softmax implementation there.

jermainewang:
One thing I noticed from the tensorflow baseline. Their loss value is much larger than ours (all the hyperparameters are the same) and the training accuracy is lower than validation accuracy. Do you guys know why this will happen?

jermainewang:
Figured out what happens. There is a weird dropout. Adding that and multi output attention heads gives better results. Updated.

PR #352, opened 2019-01-11
[Docs] Move file_pattern option from Makefile to conf.py

VoVAllen:
## Description

Move file_pattern option from Makefile to conf.py and add doc build config file. This is for further doc autobuild and versioning settings.

## Checklist


docs/source/conf.py
```diff
@@ -213,3 +213,9 @@
     'filename_pattern' : '.py',
     'download_all_examples' : False,
 }
+
+import os
```
ylfdq1118: `os` is already imported at line 15

docs/source/conf.py
```diff
@@ -213,3 +213,9 @@
...
+if os.environ['DGLBACKEND'] == 'mxnet':
+    sphinx_gallery_conf['filename_pattern'] = "/*(?<=mx)\.py"
+if os.environ['DGLBACKEND'] == 'pytorch':
+    sphinx_gallery_conf['filename_pattern'] = "/*(?<!mx)\.py"
```
szha: mx?

VoVAllen:
I'd like to use a self-host readthedocs system to build the docs, which is currently hosted at doc-build.dgl.ai. Main reason is that it's the only system supports versioning (sphinxcontrib-versioning is too old to be compatiable to current sphinx). And at the same time, it also supports doc auto building.
This also enables search capability for docs, which is available [here](http://doc-build.dgl.ai/docs/jj-dgl/en/latest/). It's currently built from my own fork of dgl.
I'll make the doc build system private and link doc.dgl.ai with the lateset doc later.

VoVAllen:
And it's the same for 0.1.x branch. Since the build system directly checkout the commit and halt when errors, we also need to modify the 0.1.x branch to build the docs.

docs/source/conf.py
VoVAllen: Removed

PR #361, opened 2019-01-21
[Feature] add NodeFlow API

zheng-da:
## Description
The PR adds the NodeFlow API. This API is defined for mini-batch training on graphs. Many sampling methods should also use this API to expose the sampled results to users. I adjust the neighbor sampling algorithm for this API. We can use neighbor sampling to construct full graph NodeFlow, mini-batch NodeFlow and neighbor sampling NodeFlow.

## Checklist
- [x] Code is well-documented


src/graph/immutable_graph.cc
```diff
@@ -820,85 +818,73 @@ SampledSubgraph ImmutableGraph::SampleSubgraph(IdArray seed_arr,
...
+      }
+      num_edges += tmp_sampled_src_list.size();
+      for (size_t i = 0; i < tmp_sampled_src_list.size(); ++i) {
+        // We need to add the neighbor in the hashtable here. This ensures that
+        // the vertex in the queue is unique. If we see a vertex before, we don't
+        // need to add it to the queue again.
+        auto ret = sub_ver_map.insert(tmp_sampled_src_list[i]);
+        // If the sampled neighbor is inserted to the map successfully.
+        if (ret.second) {
+          sub_vers.emplace_back(tmp_sampled_src_list[i], cur_node_level + 1);
```
BarclayII: I don't know how to describe it in words, but here's my understanding of sampling duplicate nodes based on the current implementation:

Consider the following bidirectional graph:
```
C     E
|\   /
| A-B
|/   \
D     F
```
and we are sampling with 3 hops, one neighbor per hop, and the seed node being A.

If in the first hop C is enqueued, then for the second hop, *A can still be sampled, even if A is already in the subgraph*.
If indeed A is sampled, then since A already exists in the nodes subset, *no new neighbors will be sampled from A if A is being sampled again*.  So in the example above, the sampling algorithm will stop there.

Please correct me if I'm wrong.  Otherwise I think this particular property is very subtle, and we need to document it somewhere (perhaps in the Python API).

zheng-da: sample
What you described isn't correct. For a graph in (a), we can sample a subgraph where the same node is sampled in different layers (b). When a node is resampled, we can continue sampling from this node.
We have an optimization here, as shown in (c). If a node is sampled multiple times in the same layer, we only sample it once. We could sample the node multiple times and duplicate the node in the layer. However, we feel this is unnecessary. The main goal of having a node in a layer is to estimate its node embedding. Duplicating the node doesn't really help us estimate the node embedding.
The same node can appear in different layer because it has different functions in the GNN model. A node in different layer estimates node embeddings in different layers.

The implementation of the sampling algorithm needs to make sure the sampling probability is strictly followed.

You are right, we should document this behavior clearly.

zheng-da:
@ZiyueHuang @GaiYu0 please help review this PR.

jermainewang:
Is this code ready? Please check the corresponding boxes.

jermainewang:
Again, similar to #362 , please put all the sampler-related codes to `sampler.cc` and `sampler.h`.

src/graph/immutable_graph.cc
BarclayII: But in line 877-879, if a node already exist in `sub_ver_map`, then `ret.second` will be `false`.  So the item `(A, cur_node_level + 1)` won't be enqueued, and the neighbors of A won't be sampled again?

zheng-da: this is to avoid resample the same node in the same layer. `sub_ver_map` is cleared at the beginning when we get to a new layer.

src/graph/immutable_graph.cc
```diff
@@ -820,85 +818,73 @@ SampledSubgraph ImmutableGraph::SampleSubgraph(IdArray seed_arr,
...
 
-  // Copy sub_ver_map to output[0]
-  // Copy layer
-  int64_t num_vertices = sub_ver_map.size();
-  std::sort(sub_vers.begin(), sub_vers.end(),
-            [](const std::pair<dgl_id_t, dgl_id_t> &a1, const std::pair<dgl_id_t, dgl_id_t> &a2) {
-    return a1.first < a2.first;
-  });
-
+  int64_t num_vertices = sub_vers.size();
```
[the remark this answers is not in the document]
BarclayII: Not urgent for me, since for random walks I can just return ID arrays.  Maybe layer-wise sampling can reuse the code a bit?

zheng-da:
i can keep the sampling code in the original file and you move it to sampler.cc?

BarclayII:
Or we can merge this PR first, then my PR and #362 follows this up?

zheng-da:
that also works.

python/dgl/node_flow.py
```diff
@@ -0,0 +1,408 @@
...
+    def copy_from_parent(self):
+        """Copy node/edge features from the parent graph.
+
+        All old features will be removed.
+        """
+        # TODO we need to avoid copying the same node embedding to all layers.
+        if self._parent._node_frame.num_rows != 0:
+            for i in range(self.num_layers):
+                nid = utils.toindex(self.layer_parent_nid(i))
+                self._node_frames[i] = FrameRef(Frame(self._parent._node_frame[nid]))
```
ZiyueHuang: Actually I think there is no additional memory allocation in the current implementation (FrameRef of a Frame creating from an existing frame, since `select_rows` is lazily). When user copy the NodeFrame to gpu externally (like https://github.com/dmlc/dgl/blob/master/examples/mxnet/sse/sse_batch.py#L155), then some keys are wasted.


Could we implement it as

```
def copy_from_parent(self, layer_id_to_keys):
    ...
    self._node_frames[i] = FrameRef(Frame(self._parent._node_frame[nid], layer_id_to_keys[i]))
```

Also slightly modify the the creation of a new frame from an existing frame (adding `keys=None`)

```
class Frame(MutableMapping):
    def __init__(self, data=None, num_rows=0, keys=None):
        ...
        self._columns = {k : Column.create(v) for k, v in data.items() if k in keys} # line 203 in frame.py
```

python/dgl/node_flow.py
```diff
@@ -0,0 +1,408 @@
...
+                self._edge_frames[i] = FrameRef(Frame(self._parent._edge_frame[eid]))
+
+    def copy_to_parent(self):
+        """Copy node/edge embeddings to the parent graph.
+        """
+        #TODO We need to take care of the following things:
+        #    * copy right node embeddings back. For instance, we should copy temporary
+        #      node embeddings back; we don't need to copy read-only node embeddings back.
+        #    * When nodes in different layers have the same node embedding, we need
+        #      to avoid conflicts.
```
ZiyueHuang: Do we really need to handle these complicated cases? For example, in which method will there be conflicts (when nodes in different layers have the same node embedding)?  In Control Variate method, each node has `num_layers` different history which need to be copy to parent.

I think it is better and more flexible to let users specify which keys to copy back to the parent graph.


Then, could we implement as

```
def copy_to_parent(self, layer_id_to_keys):
    for i in range(self.num_layers):
        nid = self.map_to_parent_nid(self.layer_parent_nid(i))
        self._parent._node_frame.update_rows(nid,
            {k :  self._node_frames[i][k] for k in layer_id_to_keys[i]}, inplace=True)
        
```

python/dgl/node_flow.py
```diff
@@ -0,0 +1,408 @@
...
+                                                    apply_func=func,
+                                                    inplace=inplace)
+            Runtime.run(prog)
+
+    def _conv_local_nid(self, nid, layer_id):
+        layer_id = self._get_layer_id(layer_id)
+        return nid - self._layer_offsets[layer_id]
+
+    def flow_compute(self, message_func="default", reduce_func="default",
+                     apply_node_func="default", range=ALL, inplace=False):
```
ZiyueHuang: Could we change `range` since it is a keyword in python

src/graph/sampler.cc
```diff
@@ -0,0 +1,462 @@
...
+  neigh_pos.reserve(num_seeds);
+  std::vector<dgl_id_t> neighbor_list;
+  std::vector<size_t> layer_offsets(num_hops + 1);
+  int64_t num_edges = 0;
+
+  layer_offsets[0] = 0;
+  layer_offsets[1] = sub_vers.size();
+  size_t idx = 0;
+  for (size_t layer_id = 1; layer_id < num_hops; layer_id++) {
+    // We need to avoid resampling the same node in a layer, but we allow a node
```
ZiyueHuang: This is wrong. The same node can be sampled multiple times in a layer.

Neighbor sampling works as, independently for each node `v`, `min(n, m(v))` neighbors are sampled from `m(v)` total neighbors of `v`.

ZiyueHuang:
Could we get the number of nodes in previous layer connected to a given node? This might be needed in sampling.

zheng-da:
@jermainewang @BarclayII @liangfu @ZiyueHuang @GaiYu0 Could you please review this PR? I think the code is ready for review.

tutorials/models/1_gnn/8_sse_mx.py
```diff
@@ -417,32 +418,62 @@ def train(g, label_nodes, steady_state_operator, predictor, trainer):
...
+            x = edges.src['x']
+            h = edges.src['h']
+            return {'m' : mx.nd.concat(x, h, dim=1)}
+
+        def reduce_func(nodes):
+            m = mx.nd.sum(nodes.mailbox['m'], axis=1)
+            z = mx.nd.concat(nodes.data['x'], m, dim=1)
+            return {'h' : self.dense2(self.dense1(z))}
+
+        subg.block_compute(0, message_func, reduce_func)
```
BarclayII: I realized that we are using NodeFlow interfaces here.  Should we move the tutorial after the NodeFlow one?  Especially since it's a brand new concept and we are not talking (and probably don't want to talk) about NodeFlow here?

Also it seems necessary to adapt the text a little bit, since we are technically not "computing on the subgraph".

tutorials/models/1_gnn/8_sse_mx.py
```diff
@@ -417,32 +418,62 @@ def train(g, label_nodes, steady_state_operator, predictor, trainer):
...
+    def forward(self, subg):
+        def message_func(edges):
+            x = edges.src['x']
+            h = edges.src['h']
+            return {'m' : mx.nd.concat(x, h, dim=1)}
+
+        def reduce_func(nodes):
+            m = mx.nd.sum(nodes.mailbox['m'], axis=1)
+            z = mx.nd.concat(nodes.data['x'], m, dim=1)
+            return {'h' : self.dense2(self.dense1(z))}
```
BarclayII: Better to move `message_func` and `reduce_func` outside `forward` as methods.  And use
```python
subg.block_compute(0, self.message_func, self.reduce_func)
```

tests/compute/test_node_flow.py
```diff
@@ -0,0 +1,255 @@
...
+
+    deg = nf.layer_in_degree(0)
+    assert F.array_equal(deg, F.zeros((nf.layer_size(0)), F.int64))
+    deg = nf.layer_out_degree(-1)
+    assert F.array_equal(deg, F.zeros((nf.layer_size(-1)), F.int64))
+    for i in range(1, nf.num_layers):
+        in_deg = nf.layer_in_degree(i)
+        out_deg = nf.layer_out_degree(i - 1)
+        assert F.asnumpy(F.sum(in_deg, 0) == F.sum(out_deg, 0))
+
```
BarclayII: Move the section**s** below to `test_copy`?  I feel that some of the logics are duplicated.

src/graph/sampler.cc
```diff
@@ -0,0 +1,462 @@
...
+                                   DLDataType{kDLInt, 64, 1}, DLContext{kDLCPU, 0});
+  nf.layer_offsets = IdArray::Empty({static_cast<int64_t>(num_hops + 1)},
+                                    DLDataType{kDLInt, 64, 1}, DLContext{kDLCPU, 0});
+  nf.flow_offsets = IdArray::Empty({static_cast<int64_t>(num_hops)},
+                                    DLDataType{kDLInt, 64, 1}, DLContext{kDLCPU, 0});
+
+  dgl_id_t *out = static_cast<dgl_id_t *>(nf.node_mapping->data);
+  dgl_id_t *out_layer = static_cast<dgl_id_t *>(nf.layer_offsets->data);
+  dgl_id_t *out_flow = static_cast<dgl_id_t *>(nf.flow_offsets->data);
+  dgl_id_t* val_list_out = static_cast<dgl_id_t *>(nf.edge_mapping->data);
```
BarclayII: These variable names don't look good...  Maybe something like `node_map_data`,`layer_off_data`, `flow_off_data`, `edge_map_data`.

src/graph/sampler.cc
```diff
@@ -0,0 +1,462 @@
...
+      out[out_node_idx++] = sub_vers[i].first;
+      layer_ver_maps[layer_id].insert(std::pair<dgl_id_t, dgl_id_t>(sub_vers[i].first, ver_id++));
+      assert(sub_vers[i].second == layer_id);
+    }
+  }
+  CHECK(out_node_idx == num_vertices);
+
+  // Copy sampled results to a CSR.
+  // When we copy data, we need to move the last layer to the first layer
+  // because we consider the layer with the leaf nodes as layer 0.
```
BarclayII: Any particular reason for reversing the direction?

tutorials/models/1_gnn/8_sse_mx.py
zheng-da: you are right. i need to fix the tutorial to pass the CI test.

tests/compute/test_node_flow.py
zheng-da: what logic is duplicated?

src/graph/sampler.cc
zheng-da: sampling algorithms have to start from the seed nodes, so the seed nodes are in the first layer and the input nodes are in the last layer. When we expose the sampled graph to a Python user, we say the input nodes are in the first layer and the seed nodes are in the last layer.

src/graph/sampler.cc
zheng-da: good point. i have updated the names accordingly

tests/compute/test_node_flow.py
BarclayII: L49-L66 duplicates L188-L198?

src/graph/sampler.cc
BarclayII: I see.  Maybe replace the comment with your reply?

src/graph/graph.cc
```diff
@@ -20,8 +21,10 @@ Graph::Graph(IdArray src_ids, IdArray dst_ids, IdArray edge_ids, size_t num_node
   CHECK(IsValidIdArray(edge_ids));
   this->AddVertices(num_nodes);
   num_edges_ = src_ids->shape[0];
-  CHECK(static_cast<int64_t>(num_edges_) == dst_ids->shape[0]) << "vectors in COO must have the same length";
-  CHECK(static_cast<int64_t>(num_edges_) == edge_ids->shape[0]) << "vectors in COO must have the same length";
+  CHECK(static_cast<int64_t>(num_edges_) == dst_ids->shape[0])
```
BarclayII: Later on I shall change `NDArray` to use `uint64_t` shapes.  These `static_cast`'s are hurting my eyes.

src/c_api_common.cc
```diff
@@ -25,7 +25,7 @@ DLManagedTensor* CreateTmpDLManagedTensor(const DGLArgValue& arg) {
 PackedFunc ConvertNDArrayVectorToPackedFunc(const std::vector<NDArray>& vec) {
     auto body = [vec](DGLArgs args, DGLRetValue* rv) {
         const int which = args[0];
-        if (which >= vec.size()) {
+        if (which >= static_cast<int>(vec.size())) {
```
BarclayII: Use `const uint64_t which` instead.

python/dgl/runtime/scheduler.py
```diff
@@ -513,6 +663,9 @@ def _gen_reduce(graph, reduce_func, edge_tuples, recv_nodes):
 
 def _gen_send_reduce(
         graph,
+        src_node_frame,
+        dst_node_frame,
+        edge_frame,
```
BarclayII: Update the docstrings accordingly.

tests/compute/test_node_flow.py
zheng-da: you are right.

python/dgl/node_flow.py
```diff
@@ -0,0 +1,845 @@
...
+
+        Parameters
+        ----------
+        block_id : int
+            the specified block to return the number of edges.
+        """
+        block_id = self._get_block_id(block_id)
+        return int(self._block_offsets[block_id + 1]) - int(self._block_offsets[block_id])
+
+    def copy_from_parent(self, node_embed_names=ALL, edge_embed_names=ALL):
```
jermainewang: Is this the same logic from subgraph? Can we reuse it?

python/dgl/node_flow.py
jermainewang: I agree. I think it's easier just to prefix the feature name with `l0/`, `l1/`, .... The prefix can be configured by the user.

python/dgl/node_flow.py
```diff
@@ -0,0 +1,845 @@
...
+                             for i in range(self.num_layers)]
+        self._edge_frames = [FrameRef(Frame(num_rows=self.block_size(i))) \
+                             for i in range(self.num_blocks)]
+        # registered functions
+        self._message_funcs = [None] * self.num_blocks
+        self._reduce_funcs = [None] * self.num_blocks
+        self._apply_node_funcs = [None] * self.num_blocks
+        self._apply_edge_funcs = [None] * self.num_blocks
+
+    def _get_layer_id(self, layer_id):
```
jermainewang: Docstring. "Handle layer_id < 0 case, blahblahblah ..."

python/dgl/node_flow.py
```diff
@@ -0,0 +1,845 @@
...
+        if layer_id >= 0:
+            return layer_id
+        else:
+            return self.num_layers + layer_id
+
+    def _get_block_id(self, block_id):
```
jermainewang: Same as above

python/dgl/node_flow.py
```diff
@@ -0,0 +1,845 @@
...
+        Tensor
+            The node id array.
+        """
+        layer_id = self._get_layer_id(layer_id)
+        assert layer_id + 1 < len(self._layer_offsets)
+        start = self._layer_offsets[layer_id]
+        end = self._layer_offsets[layer_id + 1]
+        return F.arange(int(start), int(end))
+
+    def layer_parent_nid(self, layer_id):
```
jermainewang: Do we need this API? To me it seems to be equal to:
```python
nflow = ...
nid = nflow.layer_nid(i)
parent_nid = nflow.map_to_parent_nid(nid)
```

python/dgl/node_flow.py
```diff
@@ -0,0 +1,845 @@
...
+            an :mod:`Edge UDF <dgl.udf>`.
+        block_id : int or ALL
+            the block to register the message function.
+        """
+        if is_all(block_id):
+            self._message_funcs = [func] * self.num_blocks
+        else:
+            self._message_funcs[block_id] = func
+
+    def register_reduce_func(self, func, block_id=ALL):
```
jermainewang: Does this means no reduce function can be registered for the first layer?

python/dgl/node_flow.py
```diff
@@ -0,0 +1,845 @@
...
+            v = utils.toindex(v)
+        with ir.prog() as prog:
+            scheduler.schedule_nodeflow_apply_nodes(graph=self,
+                                                    layer_id=layer_id,
+                                                    v=v,
+                                                    apply_func=func,
+                                                    inplace=inplace)
+            Runtime.run(prog)
+
+    def _layer_local_nid(self, layer_id):
```
jermainewang: I found there are only two places using this function. Maybe just inline it?

python/dgl/node_flow.py
```diff
@@ -0,0 +1,845 @@
...
+        func : callable or None, optional
+            Apply function on the nodes. The function should be
+            a :mod:`Node UDF <dgl.udf>`.
+        v : a list of vertex Ids or ALL.
+            The vertices to run the node update function.
+        inplace : bool, optional
+            If True, update will be done in place, but autograd will break.
+        """
+        if func == "default":
+            func = self._apply_node_funcs[block_id]
```
jermainewang: `layer_id` instead of `block_id`? Has this been tested?

python/dgl/node_flow.py
```diff
@@ -0,0 +1,845 @@
...
+
+        Parameters
+        ----------
+        layer_id : int
+            The specified layer to update node embeddings.
+        func : callable or None, optional
+            Apply function on the nodes. The function should be
+            a :mod:`Node UDF <dgl.udf>`.
+        v : a list of vertex Ids or ALL.
+            The vertices to run the node update function.
```
jermainewang: Is `v` local or global?

python/dgl/node_flow.py
```diff
@@ -0,0 +1,845 @@
...
+
+        Parameters
+        ----------
+        block_id : int
+            The specified block to update edge embeddings.
+        func : callable or None, optional
+            Apply function on the nodes. The function should be
+            a :mod:`Node UDF <dgl.udf>`.
+        e : a list of edge Ids or ALL.
+            The edges to run the edge update function.
```
jermainewang: `edges` instead of `e`; Is the id local or global? Does this support (u, v) tuple? If (u, v) tuple, local or global?

python/dgl/node_flow.py
```diff
@@ -0,0 +1,845 @@
...
+            scheduler.schedule_nodeflow_apply_edges(graph=self,
+                                                    block_id=block_id,
+                                                    u=u,
+                                                    v=v,
+                                                    eid=eid,
+                                                    apply_func=func,
+                                                    inplace=inplace)
+            Runtime.run(prog)
+
+    def _conv_local_nid(self, nid, layer_id):
```
jermainewang: `_glb2lcl_nid`

python/dgl/node_flow.py
```diff
@@ -0,0 +1,845 @@
...
+        layer_id = self._get_layer_id(layer_id)
+        return nid - int(self._layer_offsets[layer_id])
+
+    def _conv_local_eid(self, eid, block_id):
```
jermainewang: `_glb2lcl_eid`

python/dgl/node_flow.py
```diff
@@ -0,0 +1,845 @@
...
+        message_funcs : a callable, a list of callable, optional
+            Message functions on the edges. The function should be
+            an :mod:`Edge UDF <dgl.udf>`.
+        reduce_funcs : a callable, a list of callable, optional
+            Reduce functions on the node. The function should be
+            a :mod:`Node UDF <dgl.udf>`.
+        apply_node_funcs : a callable, a list of callable, optional
+            Apply functions on the nodes. The function should be
+            a :mod:`Node UDF <dgl.udf>`.
+        range : int or a slice or ALL.
```
jermainewang: `range` <- `flow_range`

src/graph/graph.cc
jermainewang: Sparse Attention?

jermainewang: I guess a better way is to always use `int64_t`. We only use `uint64_t` when we are very sure about it.

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+  for (size_t layer_id = 1; layer_id < num_hops; layer_id++) {
+    // We need to avoid resampling the same node in a layer, but we allow a node
+    // to be resampled in multiple layers. We use `sub_ver_map` to keep track of
+    // sampled nodes in a layer, and clear it when entering a new layer.
+    sub_ver_map.clear();
+    // sub_vers is used both as a node collection and a queue.
+    // In the while loop, we iterate over sub_vers and new nodes are added to the vector.
+    // A vertex in the vector only needs to be accessed once. If there is a vertex behind idx
+    // isn't in the last level, we will sample its neighbors. If not, the while loop terminates.
+    while (idx < sub_vers.size() && layer_id - 1 == sub_vers[idx].second) {
```
jermainewang: This while loop condition is complicated. How about:
```
for (size_t i = layer_offsets[layer_id-1]; i < layer_offsets[layer_id]; ++i) {
  ...
}
```

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+      dgl_id_t dst_id = sub_vers[idx].first;
+      int cur_node_level = sub_vers[idx].second;
```
jermainewang: const above

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+                             const size_t max_num_neighbor,
+                             std::vector<dgl_id_t>* out_ver,
+                             std::vector<dgl_id_t>* out_edge,
+                             unsigned int* seed) {
+  // Copy ver_list to output
+  if (ver_len <= max_num_neighbor) {
+    for (size_t i = 0; i < ver_len; ++i) {
+      out_ver->push_back(ver_list[i]);
+      out_edge->push_back(val_list[i]);
+    }
```
jermainewang: use `insert` to simplify the code

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+                                const size_t max_num_neighbor,
+                                std::vector<dgl_id_t>* out_ver,
+                                std::vector<dgl_id_t>* out_edge,
+                                unsigned int* seed) {
+  // Copy ver_list to output
+  if (ver_len <= max_num_neighbor) {
+    for (size_t i = 0; i < ver_len; ++i) {
+      out_ver->push_back(ver_list[i]);
+      out_edge->push_back(val_list[i]);
+    }
```
jermainewang: use `insert` to simplify the code

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+  }
+  // verify the result.
+  CHECK_EQ(sorted_idxs.size(), max_num_neighbor);
+  for (size_t i = 1; i < sorted_idxs.size(); i++) {
+    CHECK_GT(sorted_idxs[i], sorted_idxs[i - 1]);
+  }
+  for (auto idx : sorted_idxs) {
+    out_ver->push_back(ver_list[idx]);
+    out_edge->push_back(val_list[idx]);
+  }
```
jermainewang: use `insert` to simplify the code

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+/*
+ * Uniform sample vertices from a list of vertices.
+ */
+static void GetUniformSample(const dgl_id_t* val_list,
+                             const dgl_id_t* ver_list,
+                             const size_t ver_len,
+                             const size_t max_num_neighbor,
+                             std::vector<dgl_id_t>* out_ver,
+                             std::vector<dgl_id_t>* out_edge,
+                             unsigned int* seed) {
```
jermainewang: The name `val_list` and `ver_list` is confusing. Is it equal to `edge_id_list` `vert_id_list`?

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+                            col_list + *(indptr + dst_id),
+                            ver_len,
+                            num_neighbor,
+                            &tmp_sampled_src_list,
+                            &tmp_sampled_edge_list,
+                            &time_seed);
+      }
+      CHECK_EQ(tmp_sampled_src_list.size(), tmp_sampled_edge_list.size());
+      size_t pos = neighbor_list.size();
+      neigh_pos.emplace_back(dst_id, pos);
```
jermainewang: Merge the above two lines. `pos` is never used again.

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+      for (size_t i = 0; i < tmp_sampled_src_list.size(); ++i) {
+        // We need to add the neighbor in the hashtable here. This ensures that
+        // the vertex in the queue is unique. If we see a vertex before, we don't
+        // need to add it to the queue again.
+        auto ret = sub_ver_map.insert(tmp_sampled_src_list[i]);
+        // If the sampled neighbor is inserted to the map successfully.
+        if (ret.second) {
+          sub_vers.emplace_back(tmp_sampled_src_list[i], cur_node_level + 1);
+        }
+      }
```
jermainewang: Give `num_neighbor=10` and a node have 100 neighbors, does this means that it is possible no neighbor is sampled for that nodes if they *all* happen to be picked by other nodes before?

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+    }
+    layer_offsets[layer_id + 1] = layer_offsets[layer_id] + sub_ver_map.size();
+  }
+
+  uint64_t num_vertices = sub_vers.size();
```
jermainewang: const

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+    }
+  }
+  std::vector<dgl_id_t> tmp_sampled_src_list;
+  std::vector<dgl_id_t> tmp_sampled_edge_list;
+  // ver_id, position
+  std::vector<std::pair<dgl_id_t, size_t> > neigh_pos;
+  neigh_pos.reserve(num_seeds);
+  std::vector<dgl_id_t> neighbor_list;
+  std::vector<size_t> layer_offsets(num_hops + 1);
+  int64_t num_edges = 0;
```
jermainewang: `num_edges` can be directly calculated from `neighbor_list.size()`?

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+      sub_vers.emplace_back(seed[i], 0);
+    }
+  }
+  std::vector<dgl_id_t> tmp_sampled_src_list;
+  std::vector<dgl_id_t> tmp_sampled_edge_list;
+  // ver_id, position
+  std::vector<std::pair<dgl_id_t, size_t> > neigh_pos;
+  neigh_pos.reserve(num_seeds);
+  std::vector<dgl_id_t> neighbor_list;
+  std::vector<size_t> layer_offsets(num_hops + 1);
```
jermainewang: `std::vector<size_t> layer_offsets(num_hops + 1, 0);`

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+  CHECK(out_node_idx == num_vertices);
+
+  // sampling algorithms have to start from the seed nodes, so the seed nodes are
+  // in the first layer and the input nodes are in the last layer.
+  // When we expose the sampled graph to a Python user, we say the input nodes
+  // are in the first layer and the seed nodes are in the last layer.
+  // Thus, when we copy sampled results to a CSR, we need to reverse the order of layers.
+  size_t row_idx = 0;
+  for (size_t i = layer_offsets[num_hops - 1]; i < layer_offsets[num_hops]; i++)
+    indptr_out[row_idx++] = 0;
```
jermainewang: brace

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+      // First we push the size of neighbor vector
+      neighbor_list.push_back(tmp_sampled_edge_list.size());
+      // Then push the vertices
+      for (size_t i = 0; i < tmp_sampled_src_list.size(); ++i) {
+        neighbor_list.push_back(tmp_sampled_src_list[i]);
+      }
+      // Finally we push the edge list
+      for (size_t i = 0; i < tmp_sampled_edge_list.size(); ++i) {
+        neighbor_list.push_back(tmp_sampled_edge_list[i]);
+      }
```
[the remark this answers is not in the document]
jermainewang: Also, I wonder how costly it is to use a `vector<vector<` data structure for the node/edge ids in different layers? The number of layers should be manageable and I feel using `vector<vector<` is much cleaner (no need to maintain stuffs like `neigh_pos`.

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+  for (int layer_id = num_hops - 2; layer_id >= 0; layer_id--) {
+    std::sort(neigh_pos.begin() + layer_offsets[layer_id],
+              neigh_pos.begin() + layer_offsets[layer_id + 1],
+              [](const std::pair<dgl_id_t, size_t> &a1, const std::pair<dgl_id_t, size_t> &a2) {
+                return a1.first < a2.first;
+              });
+
+    for (size_t i = layer_offsets[layer_id]; i < layer_offsets[layer_id + 1]; i++) {
+      dgl_id_t dst_id = sub_vers[i].first;
+      assert(dst_id == neigh_pos[i].first);
```
jermainewang: Use CHECK. Also replace other assert with CHECK.

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+  size_t out_layer_idx = 1;
+  for (int layer_id = num_hops - 2; layer_id >= 0; layer_id--) {
+    std::sort(neigh_pos.begin() + layer_offsets[layer_id],
+              neigh_pos.begin() + layer_offsets[layer_id + 1],
+              [](const std::pair<dgl_id_t, size_t> &a1, const std::pair<dgl_id_t, size_t> &a2) {
+                return a1.first < a2.first;
+              });
+
+    for (size_t i = layer_offsets[layer_id]; i < layer_offsets[layer_id + 1]; i++) {
+      dgl_id_t dst_id = sub_vers[i].first;
```
jermainewang: const

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+      assert(dst_id == neigh_pos[i].first);
+      size_t pos = neigh_pos[i].second;
+      CHECK_LT(pos, neighbor_list.size());
+      size_t num_edges = neighbor_list[pos];
```
jermainewang: This `num_edges` actually shadows the variable in L264. That's why I suggest put nodeflow construction in another function.

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+  for (int i = 0; i < layer_offsets.size() - 2; i++) {
+    size_t num_edges = subg_csr->GetDegree(layer_off_data[i + 1], layer_off_data[i + 2]);
+    flow_off_data[out_flow_idx + 1] = flow_off_data[out_flow_idx] + num_edges;
+    out_flow_idx++;
+  }
+  CHECK(out_flow_idx == num_hops - 1);
+  CHECK(flow_off_data[num_hops - 1] == num_edges);
+
+  for (size_t i = 0; i < subg_csr->edge_ids.size(); i++)
+    subg_csr->edge_ids[i] = i;
```
jermainewang: brace

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+
+  if (neigh_type == "in")
+    nf.graph = GraphPtr(new ImmutableGraph(subg_csr, nullptr, IsMultigraph()));
+  else
+    nf.graph = GraphPtr(new ImmutableGraph(nullptr, subg_csr, IsMultigraph()));
```
jermainewang: brace

src/graph/sampler.cc
```diff
@@ -0,0 +1,463 @@
...
+#include <dgl/sampler.h>
+#include <dgl/immutable_graph.h>
+#include <algorithm>
+
+#ifdef _MSC_VER
+// rand in MS compiler works well in multi-threading.
+int rand_r(unsigned *seed) {
+  return rand();
+}
+#endif
```
jermainewang: Can we use std::random from c11 to avoid this platform check?

include/dgl/immutable_graph.h
```diff
@@ -56,6 +56,11 @@ class ImmutableGraph: public GraphInterface {
       return indices.size();
     }
 
+    /* This gets the sum of vertex degrees in the range. */
+    uint64_t GetDegree(dgl_id_t start, dgl_id_t end) const {
```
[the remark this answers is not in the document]
zheng-da: no, this is a method in CSR. You can search for it in immutable_graph.cc

python/dgl/node_flow.py
zheng-da: Similar, but it's not the same, mainly because of the concept of layers.

python/dgl/runtime/scheduler.py
```diff
@@ -232,6 +234,44 @@ def _afunc_wrapper(node_data):
...
+        If True, the update will be done in place
+
+    Returns
+    -------
+    A list of executors for DGL Runtime
+    """
+    var_nf = var.FEAT_DICT(graph._get_node_frame(layer_id), name='nf')
+    var_v = var.IDX(v)
+    v_nf = ir.READ_ROW(var_nf, var_v)
+    # TODO what is node_data?
```
ylfdq1118: remove this commend?

python/dgl/runtime/scheduler.py
```diff
@@ -403,6 +491,68 @@ def schedule_group_apply_edge(graph,
...
 
+def schedule_nodeflow_compute(graph,
+                              block_id,
+                              u, v, eid,
+                              dest_nodes,
+                              message_func,
+                              reduce_func,
+                              apply_func,
+                              inplace):
+    """get pull schedule
```
ylfdq1118: docstring

python/dgl/node_flow.py
zheng-da: do we need to write docstring for hidden methods?

python/dgl/node_flow.py
zheng-da: This is a more efficient way of implementing it. Getting the parent nid of a layer is a common operation.

python/dgl/node_flow.py
zheng-da: all registered UDFs are for blocks. not sure why reduce should be special.

python/dgl/node_flow.py
zheng-da: what is the definition of local and global here?

src/graph/sampler.cc
zheng-da: we need to use rand_r for the sake of performance. rand or std::random has a global variable for seed. It causes very bad performance in multithreads.

src/graph/graph.cc
BarclayII: > I guess a better way is to always use `int64_t`. We only use `uint64_t` when we are very sure about it.

We are very sure that *pretty much everywhere* should have used `uint64_t` (like `num_edges_`, `NumVertices()`, and a bunch of `vector.size()`'s).  The only issue is that `DLTensor`'s have *signed* integer shapes, which requires us to do `static_cast`'s or punches your face with compiler warnings otherwise.

I don't know why they made this decision; what does a negative shape stand for?

src/graph/sampler.cc
zheng-da: i don't think it's possible. all sampled neighbors will be stored.

zheng-da: sub_vers stores the nodes whose neighbors will be sampled in the next layer.

src/graph/sampler.cc
zheng-da: what's the difference?

src/graph/sampler.cc
zheng-da: i don't think this can be done with insert. this is to permute data and push data to the vectors.

python/dgl/node_flow.py
jermainewang: local is the node id of each layer; global is the id of the whole graph

src/graph/sampler.cc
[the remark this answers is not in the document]
zheng-da: i prefer to use two vectors. vector of vectors isn't very efficient in terms of memory allocation.

python/dgl/node_flow.py
zheng-da: i didn't allow node id of each layer. it's kind of confusing. there would be parent vertex id space, nodeflow vertex id space and layer id space.

src/graph/graph.cc
jermainewang: Yeah, I think so. We could do that later.

src/graph/sampler.cc
jermainewang: There seems to be solution from [here](https://stackoverflow.com/questions/24007263/is-random-number-generation-in-multiple-threads-using-c11-random-library-slow). I'm actually curious about the exact difference. The similar situation happens in #392 . We could improve this there.

PR #359, opened 2019-01-18
[GraphIndex][Fix] signed and unsigned integer conversion

GaiYu0:
## Checklist
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [$CATEGORY] (such as [Model], [Doc], [Feature]])
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [ ] To the my best knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR

PR #363, opened 2019-01-23
[Bugfix] 'const size_t' (aka 'const unsigned long') is ambiguous on MacOS

aksnzhy:
## Description
This small pr fix the compilation error in Mac OS X.


jermainewang:
Could you post the error message here?

zheng-da:
I got the same compilation error in Mac.

jermainewang:
I see what happened. This bug is introduced in #359 (@GaiYu0 be aware of this subtlety) . Before that, we were using `int`.  I feel we need a macos ci to avoid such thing to happened again.

PR #374, opened 2019-02-03
Fix  bug when dropout = 0 in GAT

sufeidechabei:
GAT will throw an error when dropout =0, I have fixed it.

### DIFF
--- a/examples/mxnet/gat/README.md
+++ b/examples/mxnet/gat/README.md
@@ -7,8 +7,17 @@ Graph Attention Networks (GAT)
 
 Note that the original code is implemented with Tensorflow for the paper.
 
+### Dependencies
+* MXNet nightly build
+* requests
 
-## Usage (make sure that DGLBACKEND is changed into mxnet)
+```bash
+pip install mxnet --pre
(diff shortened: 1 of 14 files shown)


examples/pytorch/gat/train.py
```diff
@@ -93,7 +93,9 @@ def edge_attention(self, edges):
         a = torch.exp(a).clamp(-10, 10)  # use clamp to avoid overflow
         if self.attn_drop:
             a_drop = self.attn_drop(a)
-        return {'a' : a, 'a_drop' : a_drop}
+            return {'a' : a, 'a_drop' : a_drop}
+        return {'a': a, 'a_drop': a }
```
jermainewang: Change L95 to `a = self.attn_drop(a)` and use one return statement.

jermainewang: Sorry my bad. Use
```python
if self.attn_drop:
    a_drop = self.attn_drop(a)
else:
    a_drop = a
return {'a' : a, 'a_drop' : a_drop}
```
So there is only one return statement.

Issue #387, opened 2019-02-13
A bug in the tutorial of RGCN


yuchenlin:
## 📚 Documentation

For the [Relational Graph Convolutional Network Tutorial](https://docs.dgl.ai/en/latest/tutorials/models/1_gnn/4_rgcn.html#define-full-r-gcn-model), 
there is a line `h2h = self.build_hidden_layer(idx)` that will report a bug since the defination of `self.build_hidden_layer` has no argument for `idx`

yuchenlin:
This `idx` was used to indicate whether it is the final layer, but I think it is not useful in this context.

yzh119:
Thanks for pointing this out. We'll fix this.

yzh119:
Fixed.

PR #388, opened 2019-02-13
[Bugfix] RGCN tutorial

yzh119:
## Description
Bug described in #387 

## Checklist
- [ ] Code is well-documented

## Changes
- [x] `4_rgcn.py`, the problem does not reveal during unit test because we set the number of hidden layers to 0.
- [x] rgcn example, the interface of `build_hidden_layer` in `BaseRGCN` is not compatible with its derived class(However, this does not matter for running the code)


tutorials/models/1_gnn/4_rgcn.py
```diff
@@ -294,7 +294,7 @@ def forward(self, g):
 # configurations
 n_hidden = 16 # number of hidden units
 n_bases = -1 # use number of relations as number of bases
-n_hidden_layers = 0 # use 1 input layer, 1 output layer, no hidden layer
+n_hidden_layers = 1 # use 1 input layer, 1 output layer, 1 hidden layer
```
ylfdq1118: I think in author's rgcn, there are only two layers: one from input to hidden, the other from hidden to output. There is no hidden to hidden transformation.

yzh119: Well, this change is only used for revealing the bug, it's totally ok to follow the author's settings. I've changed that.

ylfdq1118: Or if you think it would be more intuitive, you can also leave line 297 to have value 1, and change the line 235 to
`for _ in range(self.num_hidden_layers - 1):`

Issue #369, opened 2019-01-29
TODO list for ImmutableGraph related features


zheng-da:
## 🚀 Feature

Current `ImmutableGraph` (merged in #342 ) still lacks some functionalities. This might be a problem when user switch from normal graph to it (some APIs will break). For example, ImmutableGraph needs to support FindEdges, EdgeSubgraphs, batching graphs, etc.

We use this issue to book-keep all the missing features.

yzh119:
Could you elaborate more on EdgeSubgraphs?

yzh119:
TODOs:
- FindEdges
1. `FindEdge(dgl_id_t eid)`
2. `FindEdges(IdArray eids)`

yzh119:
@zheng-da I've implemented the two APIs: 
```cpp
std::pair<dgl_id_t, dgl_id_t> ImmutableGraph::FindEdge(dgl_id_t eid) const {
  auto out_csr = GetOutCSR();
  // The edge_ids must have been sorted.
  auto ptr = std::lower_bound(out_csr->edge_ids.begin(), out_csr->edge_ids.end(), eid);
  //auto ptr = std::find(out_csr->edge_ids.begin(), out_csr->edge_ids.end(), eid);
  CHECK(eid == *ptr);
  const int64_t idx = ptr - out_csr->edge_ids.begin();
  dgl_id_t col = out_csr->indices[idx];
  dgl_id_t row = std::lower_bound(out_csr->indptr.begin(), out_csr->indptr.end(), idx) - out_csr->indptr.begin();
  return std::pair<dgl_id_t, dgl_id_t>(row, col);
}

ImmutableGraph::EdgeArray ImmutableGraph::FindEdges(IdArray eids) const {
  dgl_id_t* eid_data = static_cast<dgl_id_t*>(eids->data);
  const auto len = eids->shape[0];
  std::vector<dgl_id_t> src, dst;
  for (size_t k = 0; k < len; k++) {
    auto edge = ImmutableGraph::FindEdge(eid_data[k]);
    src.push_back(edge.first);
    dst.push_back(edge.second);
  }

  IdArray rst_src = IdArray::Empty({len}, eids->dtype, eids->ctx);
  IdArray rst_dst = IdArray::Empty({len}, eids->dtype, eids->ctx);
  dgl_id_t* rst_src_data = static_cast<dgl_id_t*>(rst_src->data);
[...]

PR #389, opened 2019-02-13
[Feature] Add switch readonly state interface

yzh119:
## Description
Switch between immutable and mutable graph. @BarclayII mentioned about it earlier.
This PR is a part of #369 

## Checklist
- [x] Code is well-documented

## Changes
- Add a interface `switch_mode(readonly=True)` in `DGLGraph` class.


python/dgl/graph.py
```diff
@@ -200,6 +200,37 @@ def __init__(self,
         self._apply_node_func = None
         self._apply_edge_func = None
 
+    def switch_mode(self, readonly=True):
```
BarclayII: The name `switch_mode` is too generic and doesn't seem to match the docstring.  Either change the name to `switch_readonly`, or the docstring to `Switch mode of the graph` if future toggles are to be included.

python/dgl/graph_index.py
```diff
@@ -138,6 +138,14 @@ def is_readonly(self):
             self._readonly = bool(_CAPI_DGLGraphIsReadonly(self._handle))
         return self._readonly
 
+    def switch_mode(self, readonly=True):
+        """Switch the readonly state of graph index."""
+        n_nodes, multigraph, _, src, dst = self.__getstate__()
+        self.clear_cache()
+        state = (n_nodes, multigraph, readonly, src, dst)
+        self._readonly = readonly
```
BarclayII: `_readonly` should be set in `__setstate__`; change that instead.

yzh119:
@BarclayII I think `mutable_()`/`readonly_()` is a more concise name, what's your idea?

VoVAllen:
How about API like pytorch's `tensor.requires_grad()` and `tensor.requires_grad(False)`, which means `g.readonly()` and `g.readonly(False)`?

yzh119:
@VoVAllen I think `g.readonly_()` is more appropriate considering it's an in-place operation, PyTorch also has a corresponding [requires_grad_](https://pytorch.org/docs/stable/tensors.html#torch.Tensor.requires_grad_) function.

BarclayII:
PyTorch also has `Module.train()` and `Module.eval()` which changes the mode in-place.

I'm fine with either `readonly(bool)` or `switch_mode()`.  `readonly_()` is too PyTorch and I don't like that too much.

yzh119:
I think it sounds more like creating a new graph by calling `g.readonly(bool)`, this is different from `Module.train()` which does not return anything.
However, I have no preference between `readonly(bool)`/`readonly_(bool)`, it's up to you guys.

BarclayII:
I think `readonly` is better.  We don't have the convention that `<func>` creates a new tensor and `<func>_` modifies in-place.

yzh119:
@BarclayII updated.

zheng-da:
i just fixed a bug in the sampler, which fails the CI frequently: https://github.com/dmlc/dgl/pull/401
please rebase to the master branch.

PR #392, opened 2019-02-14
[Feature] Random walk traces generation

BarclayII:
## Description
`dgl.contrib.sampling.random_walk(g, seeds, num_traces, num_hops)`: Batch-generates same-length random walk traces.

EDIT: also moved C++ sampler code from `immutable_graph.cc` to `sampler.cc`

## Checklist
- [x] Code is well-documented

I wonder where is the best place to put `GraphIndex` sampler methods (e.g. `GraphIndex.random_walk` or `GraphIndex.neighbor_sampling`).  I would expect to have more samplers in the future, and maybe `GraphIndex` would look too heavy if we put everything there?


python/dgl/graph_index.py
```diff
@@ -677,6 +677,20 @@ def neighbor_sampling(self, seed_ids, expand_factor, num_hops, neighbor_type, no
...
+        (num_seeds, num_traces, num_hops + 1)
+        """
+        if len(seeds) == 0:
+            return utils.toindex([])
+
+        seeds = seeds.todgltensor()
+        traces = _CAPI_DGLGraphRandomWalk(self._handle, seeds, num_traces, num_hops)
+
+        return F.zerocopy_from_dlpack(traces.to_dlpack())
+
```
jermainewang: Change it to global function.

src/graph/sampler.cc
```diff
@@ -0,0 +1,483 @@
...
+                                   int num_traces,
+                                   int num_hops) const {
+  const int num_nodes = seeds->shape[0];
+  const dgl_id_t *seed_ids = static_cast<dgl_id_t *>(seeds->data);
+  IdArray traces = IdArray::Empty(
+      {num_nodes, num_traces, num_hops + 1},
+      DLDataType{kDLInt, 64, 1},
+      DLContext{kDLCPU, 0});
+  dgl_id_t *trace_data = static_cast<dgl_id_t *>(traces->data);
+  unsigned int random_seed = time(nullptr);
```
jermainewang: I guess we should allow user set random seed. To avoid having `seed` argument everywhere, we could export an API similar to `numpy.random.seed`.

include/dgl/immutable_graph.h
```diff
@@ -475,6 +476,15 @@ class ImmutableGraph: public GraphInterface {
...
   virtual std::vector<IdArray> GetAdj(bool transpose, const std::string &fmt) const;
 
+  /*!
+   * \brief Batch-generate random walk traces
+   * \param seeds The array of starting vertex IDs
+   * \param num_traces The number of traces to generate for each seed
+   * \param num_hops The number of hops for each trace
+   * \return a flat ID array with shape (num_seeds, num_traces, num_hops + 1)
+   */
+  IdArray RandomWalk(IdArray seeds, int num_traces, int num_hops) const;
```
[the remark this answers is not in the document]
BarclayII: Sounds like a more reasonable solution would involve a `Sampler` class that takes in a graph and an RNG, and produce different sampling outputs:
```C++
class GraphSampler {
  public:
    GraphSampler(GraphInterface *gptr, uint64_t seed);
    void SetRandomSeed(uint64_t seed);
    NodeFlow NeighborUniformSample(IdArray seeds, std::string &neigh_type, int num_hops, int expand_factor);
    IdArray RandomWalk(IdArray seeds, int num_traces, int num_hops);
};
```

I would prefer refactoring these as a separate PR after merging this and the NodeFlow PR.

python/dgl/graph_index.py
BarclayII: Similarly, we probably need a `Sampler` class that wraps the C++ `Sampler` class?

src/graph/sampler.cc
BarclayII: The `Sampler` class should solve the problem of setting RNG states.

include/dgl/immutable_graph.h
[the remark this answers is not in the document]
jermainewang: I don't know whether we need a `GraphSampler` class. All the sampler APIs are stateless, meaning the class is likely to be stateless as well. The only member seems to be `seed`, which could be a global(thread local) variable as it should be.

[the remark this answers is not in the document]
BarclayII: So make all these sampler functions friends?

include/dgl/immutable_graph.h
```diff
@@ -538,6 +548,13 @@ class ImmutableGraph: public GraphInterface {
    * When a multiedge is added, this flag switches to true.
    */
   bool is_multigraph_ = false;
+ private:
+  /*! \brief Get a random successor given the current node.  Changes the seed */
+  inline dgl_id_t GetRandomSuccessor(dgl_id_t vid, unsigned int *seed) const {
+    const DGLIdIters &succ = this->GetOutCSR()->GetIndexRef(vid);
+    const size_t size = succ.size();
+    return succ[rand_r(seed) % size];
+  }
```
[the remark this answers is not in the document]
BarclayII: I feel this function to be small enough to be inlined.  But I'm not sure where to put it.  Another friend function in `sampler.cc`?

include/dgl/immutable_graph.h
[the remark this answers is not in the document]
jermainewang: Ideally, we should rely on the public methods in `GraphInterface` for implementation. Making it friend functions would probably fail in the future when the inheritance hierarchy becomes more complicated. Do you need any internal structure to implement random walk right now?

include/dgl/immutable_graph.h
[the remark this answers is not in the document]
jermainewang: Can we use the `SuccVec` interface instead?
https://github.com/dmlc/dgl/blob/220a1e68fbdb1026c3c3efa71eeb2b9b0a23378d/include/dgl/graph_interface.h#L296

This makes it only relying on `GraphInterface` and thus could be implemented as an inline global function in the cc file.

python/dgl/graph_index.py
jermainewang: Agree.

include/dgl/immutable_graph.h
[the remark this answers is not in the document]
BarclayII: `GetOutCSR()` is a protected method I'm using.

include/dgl/immutable_graph.h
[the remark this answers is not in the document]
BarclayII: Good point.  That should allow me to have a global function.

include/dgl/runtime/util.h
```diff
@@ -8,6 +8,14 @@
...
 #include "c_runtime_api.h"
 
+#ifdef _MSC_VER
+// rand in MS compiler works well in multi-threading.
+static inline int rand_r(unsigned *seed) {
+  return rand();
+}
+#define _CRT_RAND_S
+#endif
+
```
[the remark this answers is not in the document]
BarclayII: I'm not sure how to use `std::uniform_int_distribution` properly for this purpose since the range of random number is always changing.  It seems that they want us to initialize a different RNG every time the lower/upper bound changes?  Here's the reference code:
```C++
    std::random_device rd;  //Will be used to obtain a seed for the random number engine
    std::mt19937 gen(rd()); //Standard mersenne_twister_engine seeded with rd()
    std::uniform_int_distribution<> dis(1, 6);
 
    for (int n=0; n<10; ++n)
        //Use dis to transform the random unsigned int generated by gen into an int in [1, 6]
        std::cout << dis(gen) << ' ';
```

src/graph/sampler.cc
zheng-da: agree

src/graph/sampler.cc
```diff
@@ -0,0 +1,487 @@
...
+
+  // TODO: openmp support
```
zheng-da: adding openmp should be easy.

zheng-da:
it seems the random walk sampling isn't the one used by PinSage. what is the main use of this random walk sampler?

BarclayII:
It *is* used by PinSage.  It generates the traces of random walks, used for computing the weights of the nodes.

python/dgl/contrib/sampling/randomwalk.py
```diff
@@ -0,0 +1,31 @@
...
+    Returns
+    -------
+    traces : Tensor
+        A 3-dimensional node ID tensor with shape
+
+            (num_seeds, num_traces, num_hops + 1)
+
+        traces[i, j, 0] are always starting nodes (i.e. seed[i]).
+    """
+    return g._graph.random_walk(utils.toindex(seeds), num_traces, num_hops)
```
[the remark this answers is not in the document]
BarclayII: Right now I think this interface is the bare minimum to provide necessary functionality, and users can do whatever minibatch sampling they want.  I could add a DataLoader that iterates over minibatches of seed nodes, but that would be another PR.

zheng-da:
i thought PinSage does random walk and picks the top k visited nodes as neighbors. it doesn't really care about the trace itself.

BarclayII:
You need to actually generate the traces to figure out which nodes are to be picked anyway.  Plus I think the random walk function is more generic.

src/graph/sampler.cc
BarclayII: I'm not sure whether each thread should receive its own `random_seed`.

include/dgl/runtime/util.h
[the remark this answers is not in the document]
jermainewang: According to [cppreference](http://www.cplusplus.com/reference/random/), generator and distribution are two things.
* Generator for generating random numbers from the seed.
* Distribution for converting random numbers to certain distribution.
In our case, I think each call will create a new distribution object that uses the same generator object.

For multi-thread case, according to [this stackoverflow](https://stackoverflow.com/questions/24007263/is-random-number-generation-in-multiple-threads-using-c11-random-library-slow), we should create one generator for each thread so there is no interlock between threads. There will be only one `random_device` and locking is required to initialize the generator but the overhead is one-time.

We need to write benchmark code for this.

zheng-da:
i agree. so this PR is to build a generic random walk sampler?

The next step is to use the trace to find topK visited vertices and construct a NodeFlow?

src/graph/sampler.cc
```diff
@@ -305,7 +307,7 @@ NodeFlow ConstructNodeFlow(std::vector<dgl_id_t> neighbor_list,
   }
   layer_off_data[0] = 0;
   layer_off_data[1] = layer_offsets[num_hops] - layer_offsets[num_hops - 1];
-  size_t out_layer_idx = 1;
+  int out_layer_idx = 1;
```
zheng-da: why changing the types?

src/graph/sampler.cc
zheng-da: every thread should have their own seeds to avoid write conflicts.

src/graph/sampler.cc
BarclayII: It throws type comparison warnings otherwise.

BarclayII:
For now I'm doing that in Python (https://github.com/BarclayII/dgl/blob/recommendation/examples/pytorch/recommendation/rec/model/pinsage.py#L97-L102), and I need to see whether that could be a bottleneck, and whether moving top-K and nodeflow into C is a priority.

src/graph/sampler.cc
```diff
@@ -345,14 +348,14 @@ NodeFlow ConstructNodeFlow(std::vector<dgl_id_t> neighbor_list,
 
   // Copy flow offsets.
   flow_off_data[0] = 0;
-  size_t out_flow_idx = 0;
-  for (int i = 0; i < layer_offsets.size() - 2; i++) {
+  int out_flow_idx = 0;
+  for (int i = 0; i < (int)layer_offsets.size() - 2; i++) {
```
zheng-da: Please define i as size_t

src/graph/sampler.cc
```diff
@@ -345,14 +348,14 @@ NodeFlow ConstructNodeFlow(std::vector<dgl_id_t> neighbor_list,
...
     size_t num_edges = subg_csr->GetDegree(layer_off_data[i + 1], layer_off_data[i + 2]);
     flow_off_data[out_flow_idx + 1] = flow_off_data[out_flow_idx] + num_edges;
     out_flow_idx++;
   }
   CHECK(out_flow_idx == num_hops - 1);
-  CHECK(flow_off_data[num_hops - 1] == num_edges);
+  CHECK(flow_off_data[num_hops - 1] == (uint64_t)num_edges);
```
zheng-da: Let’s not use c style type cast

zheng-da:
Feel free to merge the pr after you fix the type cast.

PR #395, opened 2019-02-15
[Model]PPI dataloader and inductive learning script.

sufeidechabei:
## Changes
1. PPI dataloader.
2. Training script.

### DIFF
--- a/examples/pytorch/gat/train_ppi.py
+++ b/examples/pytorch/gat/train_ppi.py
@@ -17,7 +17,7 @@ import torch.nn.functional as F
 import argparse
 from sklearn.metrics import f1_score
 from gat import GAT
-from dgl.data.ppi import PPIDataset, load_ppi
+from dgl.data.ppi import PPIDataset
 from torch.utils.data import DataLoader
 
 def collate(sample):
@@ -54,10 +54,9 @@ def main(args):
     # define loss function
     loss_fcn = torch.nn.BCEWithLogitsLoss()
     # create the dataset
-    data = load_ppi()
-    train_dataset = PPIDataset(mode='train', data=data)
-    valid_dataset = PPIDataset(mode='valid', data=data)
-    test_dataset = PPIDataset(mode='test', data=data)
+    train_dataset = PPIDataset(mode='train')
+    valid_dataset = PPIDataset(mode='valid')
+    test_dataset = PPIDataset(mode='test')
     train_dataloader = DataLoader(train_dataset, batch_size=batch_size, collate_fn=collate)
     valid_dataloader = DataLoader(valid_dataset, batch_size=batch_size, collate_fn=collate)
     test_dataloader = DataLoader(test_dataset, batch_size=batch_size, collate_fn=collate)
--- a/examples/pytorch/gat/train_ppi.py
+++ b/examples/pytorch/gat/train_ppi.py
@@ -10,7 +10,6 @@ Author's code: https://github.com/PetarV-/GAT
 Pytorch implementation: https://github.com/Diego999/pyGAT
 """
 
-
 import numpy as np
 import torch
 import dgl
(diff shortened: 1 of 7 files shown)


examples/pytorch/gat/train_ppi.py
```diff
@@ -0,0 +1,169 @@
...
+"""
+
+
+import numpy as np
+import torch
+import dgl
+import torch.nn.functional as F
+import argparse
+from sklearn.metrics import f1_score
+from train import GAT
```
jermainewang: Put `GAT` in `gat.py`. Let `train.py` and `train_ppi.py` import from it.

python/dgl/data/ppi.py
```diff
@@ -0,0 +1,121 @@
...
+
+    with open('{}/ppi/ppi-class_map.json'.format(dir)) as jsonfile:
+        class_map = json.load(jsonfile)
+    features = np.load('{}/ppi/ppi-feats.npy'.format(dir))
+    label_list = []
+    for i in range(len(class_map)):
+        label_list.append(np.expand_dims(np.array(class_map[str(i)]), axis=0))
+    labels = np.concatenate(label_list)
+    graph = DGLGraph(nx.DiGraph(json_graph.node_link_graph(g_data)))
+    graph_id = np.load('{}/ppi/graph_id.npy'.format(dir))
```
jermainewang: Put the above codes in `__init__`.

python/dgl/data/ppi.py
```diff
@@ -0,0 +1,121 @@
...
+
+        Paramters
+        ---------
+        mode : str
+            ('train', 'valid', 'test').
+        """
+        self.mode = mode
+        self._preprocess()
+        self._normalize()
+    def _preprocess(self):
```
jermainewang: Add a newline above.

python/dgl/data/ppi.py
```diff
@@ -0,0 +1,121 @@
...
+        if self.mode == 'test':
+            self.test_mask_list = []
+            self.test_graphs = []
+            self.test_labels = []
+            for test_graph_id in range(23, 25):
+                test_graph_mask = np.where(self.graph_id == test_graph_id)[0]
+                self.test_mask_list.append(test_graph_mask)
+                self.test_graphs.append(self.graph.subgraph(test_graph_mask))
+                self.test_labels.append(self.labels[test_graph_mask])
+    def _normalize(self):
```
jermainewang: Add a newline above.

python/dgl/data/ppi.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        Normalize the features
+        """
+
+        train_feats = self.features[np.concatenate(self.train_mask_list)]
+        scaler = StandardScaler()
+        scaler.fit(train_feats)
+        self.features = scaler.transform(self.features)
+    def __len__(self):
```
jermainewang: Add a newline above.

python/dgl/data/ppi.py
sufeidechabei: If I put this part into `__init__`, it will load the data three times when create the train, test, validation and spend more time.

python/dgl/data/ppi.py
sufeidechabei: I have done them.

python/dgl/data/ppi.py
jermainewang: My reply is accidentally cancelled. Have you received it?

jermainewang: You are using class attributes in a wrong way. Class attributes will be evaluated with the class definition.
```python
class C:
  x = 1
  print('xxx')
# 'xxx' will be printed immediately after the above definition.
print(C.x)  # x become an attribute of the class not an instance
```

This means if user import `dgl.data.ppi`, the dataset will be loaded no matter what. User cannot control it. Even worse, if `from ppi import PPIDataset` is in `dgl.data.__init__`, then a user who wants to load other dataset will accidentally load PPIDataset.

There are many ways to fix it:
* Load the whole dataset and split the train/val/test sets outside the dataloader. I believer there should be clean ways to achieve that.
* Split the dataset beforehand.
* Add an `offset` argument in the `init` function to skip graphs before.

sufeidechabei: I have add the `load_ppi` function to load the data, `PPIDataset` is used for preprocessing the dataset.

jermainewang: That's not a good practice. Please have a look at [Pytorch's MNIST dataset](https://pytorch.org/docs/0.4.0/_modules/torchvision/datasets/mnist.html). They also have train and test set. See how they use one Dataset API to solve the problem.

jermainewang:
Thanks @sufeidechabei !

PR #403, opened 2019-02-21
[Model][Pytorch] GraphSAGE

hbsun2113:
## Description
simple implemention of GraphSAGE. 

## Checklist
- [x] Code is well-documented

### DIFF
--- a/examples/pytorch/graphsage/graphsage.py
+++ b/examples/pytorch/graphsage/graphsage.py
@@ -4,53 +4,43 @@ Paper: http://papers.nips.cc/paper/6703-inductive-representation-learning-on-lar
 Code: https://github.com/williamleif/graphsage-simple
 Simple reference implementation of GraphSAGE.
 """
-import argparse, time, math
+import argparse
+import time
+import abc
 import numpy as np
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
 from dgl import DGLGraph
 from dgl.data import register_data_args, load_data
-
-
-def graphsage_msg(edge):
-    msg = edge.src['h']
-    return {'m': msg}
+import dgl.function as fn
 
 
 class Aggregator(nn.Module):
     def __init__(self, g, in_feats, out_feats, activation=None, bias=True):
         super(Aggregator, self).__init__()
         self.g = g
-        self.weight = nn.Parameter(torch.Tensor(in_feats, out_feats))  # (F,EF)
-        if bias:
-            self.bias = nn.Parameter(torch.Tensor(out_feats))  # (EF,1)
-        else:
-            self.bias = None
+        self.linear = nn.Linear(in_feats, out_feats, bias=bias)  # (F, EF) or (2F, EF)
         self.activation = activation
-        self.reset_parameters()
-
-    def reset_parameters(self):
-        weight_stdv = 1. / math.sqrt(self.weight.size(1))
-        self.weight.data.uniform_(-weight_stdv, weight_stdv)
-        if self.bias is not None:
-            bias_stdv = 1. / math.sqrt(self.bias.size(0))
-            self.bias.data.uniform_(-bias_stdv, bias_stdv)
+        nn.init.xavier_uniform_(self.linear.weight, gain=nn.init.calculate_gain('relu'))
 
     def forward(self, node):
         nei = node.mailbox['m']  # (B, N, F)
         h = node.data['h']  # (B, F)
-        h = self.concat(h, nei, node)  # (B, F)
-        h = torch.mm(h, self.weight)  # (B, EF)
-        if self.bias is not None:
-            h = h + self.bias
+        h = self.concat(h, nei, node)  # (B, F) or (B, 2F)
+        h = self.linear(h)   # (B, EF)
         if self.activation:
             h = self.activation(h)
+        norm = torch.pow(h, 2)
+        norm = torch.sum(norm, 1, keepdim=True)
+        norm = torch.pow(norm, -0.5)
+        norm[torch.isinf(norm)] = 0
+        # h = h * norm
         return {'h': h}
 
+    @abc.abstractmethod
     def concat(self, h, nei, nodes):
-        print('no implementation !')
-        pass
+        raise NotImplementedError
 
 
 class MeanAggregator(Aggregator):
@@ -63,7 +53,7 @@ class MeanAggregator(Aggregator):
             degs = degs.cuda(h.device)
         concatenate = torch.cat((nei, h.unsqueeze(1)), 1)
         concatenate = torch.sum(concatenate, 1) / degs.unsqueeze(1)
-        return concatenate
+        return concatenate  # (B, F)
 
 
 class PoolingAggregator(Aggregator):
@@ -72,41 +62,24 @@ class PoolingAggregator(Aggregator):
         self.mlp = PoolingAggregator.MLP(in_feats, in_feats, F.relu, False, True)
 
     def concat(self, h, nei, nodes):
-        nei = self.mlp(nei)
-        concatenate = torch.cat((nei, h), 1)
+        nei = self.mlp(nei)  # (B, F)
+        concatenate = torch.cat((nei, h), 1)  # (B, 2F)
         return concatenate
 
     class MLP(nn.Module):
         def __init__(self, in_feats, out_feats, activation, dropout, bias):  # (F, F)
             super(PoolingAggregator.MLP, self).__init__()
-            self.weight = nn.Parameter(torch.Tensor(in_feats, out_feats))  # (F,EF)
-            if dropout:
-                self.dropout = nn.Dropout(p=dropout)
-            else:
-                self.dropout = 0.
-            if bias:
-                self.bias = nn.Parameter(torch.Tensor(out_feats))  # (EF)
-            else:
-                self.bias = None
+            self.linear = nn.Linear(in_feats, out_feats, bias=bias)  # (F, F)
+            self.dropout = nn.Dropout(p=dropout)
             self.activation = activation
-            self.reset_parameters()
+            nn.init.xavier_uniform_(self.linear.weight, gain=nn.init.calculate_gain('relu'))
 
-        def reset_parameters(self):
-            weight_stdv = 1. / math.sqrt(self.weight.size(1))
-            self.weight.data.uniform_(-weight_stdv, weight_stdv)
-            if self.bias is not None:
-                bias_stdv = 1. / math.sqrt(self.bias.size(0))
-                self.bias.data.uniform_(-bias_stdv, bias_stdv)
-
-        def forward(self, nei):  # (B, N, F)
-            if self.dropout:
-                nei = self.dropout(nei)
-            nei = torch.matmul(nei, self.weight)  # (B, N, EF)
-            if self.bias is not None:
-                nei = nei + self.bias
+        def forward(self, nei):
+            nei = self.dropout(nei)  # (B, N, F)
+            nei = self.linear(nei)
             if self.activation:
                 nei = self.activation(nei)
-            max_value = torch.max(nei, dim=1)[0]  # (B, EF)
+            max_value = torch.max(nei, dim=1)[0]  # (B, F)
             return max_value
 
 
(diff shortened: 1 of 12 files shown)


examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+        if aggregator_type == "pooling":
+            self.aggregator = PoolingAggregator(g, in_feats, out_feats, activation, bias)
+        else:
+            self.aggregator = MeanAggregator(g, in_feats, out_feats, activation, bias)
+
+    def forward(self, h):
+        if self.dropout:
+            h = self.dropout(h)
+        self.g.ndata['h'] = h
+        self.g.update_all(graphsage_msg, self.aggregator)
```
ZiyueHuang: You can use the built-in message function `fn.copy_src(src='h', out='m')`

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+        h = self.concat(h, nei, node)  # (B, F)
+        h = torch.mm(h, self.weight)  # (B, EF)
+        if self.bias is not None:
+            h = h + self.bias
+        if self.activation:
+            h = self.activation(h)
+        return {'h': h}
+
+    def concat(self, h, nei, nodes):
+        print('no implementation !')
```
ZiyueHuang: raise error here?

ZiyueHuang:
Do you have any plan to implement other aggregators such as LSTM aggregator?

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+
+class Aggregator(nn.Module):
+    def __init__(self, g, in_feats, out_feats, activation=None, bias=True):
+        super(Aggregator, self).__init__()
+        self.g = g
+        self.weight = nn.Parameter(torch.Tensor(in_feats, out_feats))  # (F,EF)
+        if bias:
+            self.bias = nn.Parameter(torch.Tensor(out_feats))  # (EF,1)
+        else:
+            self.bias = None
```
BarclayII: L25-L29: You can simply use `nn.Linear` instead, like this:
```python
    self.linear = nn.Linear(in_feats, out_feats, bias=bias)
```

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+        self.activation = activation
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        weight_stdv = 1. / math.sqrt(self.weight.size(1))
+        self.weight.data.uniform_(-weight_stdv, weight_stdv)
+        if self.bias is not None:
+            bias_stdv = 1. / math.sqrt(self.bias.size(0))
+            self.bias.data.uniform_(-bias_stdv, bias_stdv)
```
BarclayII: This procedure can be replaced with `nn.init`.  For example, if you are using ReLU activations, the following runs Xavier uniform initialization with a correction:
```python
        nn.init.xavier_uniform_(self.linear.weight, gain=nn.init.calculate_gain('relu'))
        # if you stick with your code with individual parameters that will be self.weight.
```

Only a hint, since your initialization is different from Xavier uniform.  I saw the original implementation using Xavier uniform though.

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+
+    def forward(self, node):
+        nei = node.mailbox['m']  # (B, N, F)
+        h = node.data['h']  # (B, F)
+        h = self.concat(h, nei, node)  # (B, F)
+        h = torch.mm(h, self.weight)  # (B, EF)
+        if self.bias is not None:
+            h = h + self.bias
```
BarclayII: L44-46 can then be replaced with
```python
        h = self.linear(h)
```

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+        if self.activation:
+            h = self.activation(h)
+        return {'h': h}
+
+    def concat(self, h, nei, nodes):
+        print('no implementation !')
+        pass
```
BarclayII:
```python
        raise NotImplementedError
```
Also remove the `pass` statement.

examples/pytorch/graphsage/graphsage.py
hbsun2113: Thanks for your reivew! I got it!

examples/pytorch/graphsage/graphsage.py
hbsun2113: No, it is just an abstract method.

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+    def __init__(self, g, in_feats, out_feats, activation, bias):  # (2F, F)
+        super(PoolingAggregator, self).__init__(g, in_feats*2, out_feats, activation, bias)
+        self.mlp = PoolingAggregator.MLP(in_feats, in_feats, F.relu, False, True)
+
+    def concat(self, h, nei, nodes):
+        nei = self.mlp(nei)
+        concatenate = torch.cat((nei, h), 1)
+        return concatenate
+
+    class MLP(nn.Module):
```
BarclayII: Move this outside the `PoolingAggregator` class.

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+        def __init__(self, in_feats, out_feats, activation, dropout, bias):  # (F, F)
+            super(PoolingAggregator.MLP, self).__init__()
+            self.weight = nn.Parameter(torch.Tensor(in_feats, out_feats))  # (F,EF)
+            if dropout:
+                self.dropout = nn.Dropout(p=dropout)
+            else:
+                self.dropout = 0.
```
BarclayII: L83-86: just do
```python
self.dropout = nn.Dropout(p=dropout)
```
`dropout = 0` automatically disables it, so you can always pass the input to the dropout layer regardless.

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+            if bias:
+                self.bias = nn.Parameter(torch.Tensor(out_feats))  # (EF)
+            else:
+                self.bias = None
```
BarclayII: L82, L87-L90: similarly you can just do
```python
self.linear = nn.Linear(in_feats, out_feats, bias=bias)
```

hbsun2113:
> Do you have any plan to implement other aggregators such as LSTM aggregator?

There are three aggregators in the original paper, and I have implemented two of them.

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+            self.activation = activation
+            self.reset_parameters()
+
+        def reset_parameters(self):
+            weight_stdv = 1. / math.sqrt(self.weight.size(1))
+            self.weight.data.uniform_(-weight_stdv, weight_stdv)
+            if self.bias is not None:
+                bias_stdv = 1. / math.sqrt(self.bias.size(0))
+                self.bias.data.uniform_(-bias_stdv, bias_stdv)
```
BarclayII: Again, consider `torch.nn.init.xavier_uniform_` in this function.

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+
+        def forward(self, nei):  # (B, N, F)
+            if self.dropout:
+                nei = self.dropout(nei)
```
BarclayII: Considering the changes in L83-L86, you can simply always do
```python
nei = self.dropout(nei)
```

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+            nei = torch.matmul(nei, self.weight)  # (B, N, EF)
+            if self.bias is not None:
+                nei = nei + self.bias
```
BarclayII: L104-106:
```python
nei = self.linear(nei)
```

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+                 dropout,
+                 aggregator_type,
+                 bias=True,
+                 ):
+        super(GraphSAGELayer, self).__init__()
+        self.g = g
+        if dropout:
+            self.dropout = nn.Dropout(p=dropout)
+        else:
+            self.dropout = 0.
```
BarclayII: Again, just do `self.dropout = nn.Dropout(p=dropout)`

examples/pytorch/graphsage/graphsage.py
BarclayII: Meaning: replace `graphsage_msg` with `fn.copy_src(src='h', out='m')`.

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+        if aggregator_type == "pooling":
+            self.aggregator = PoolingAggregator(g, in_feats, out_feats, activation, bias)
+        else:
+            self.aggregator = MeanAggregator(g, in_feats, out_feats, activation, bias)
+
+    def forward(self, h):
+        if self.dropout:
+            h = self.dropout(h)
```
BarclayII: Again, always do `h = self.dropout(h)`.

examples/pytorch/graphsage/graphsage.py
hbsun2113: What is the difference between them?

examples/pytorch/graphsage/graphsage.py
hbsun2113: I fix it by:
```
import abc
@abc.abstractmethod
    def concat(self, h, nei, nodes):
        raise NotImplementedError
```

examples/pytorch/graphsage/graphsage.py
BarclayII: There is no difference semantically.  However,
* You write one less Python function,
* It allows possibility for DGL to further optimize on speed, since you are using a DGL-provided builtin function.

examples/pytorch/graphsage/graphsage.py
hbsun2113: Thanks for your review!! I learned a lot!
 I wonder why it is better?

hbsun2113:
@BarclayII @ZiyueHuang 
Thanks for your review! I have fixed the code as you suggest. Can you tell me how to update the pull request?

examples/pytorch/graphsage/graphsage.py
```diff
@@ -0,0 +1,295 @@
...
+import argparse, time, math
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from dgl import DGLGraph
+from dgl.data import register_data_args, load_data
+
+
+def graphsage_msg(edge):
```
hbsun2113: test

jermainewang:
> Thanks for your review! I have fixed the code as you suggest. Can you tell me how to update the pull request?

@hbsun2113 Simply push commits normally to your `graphsage` branch and the PR will be updated accordingly.

examples/pytorch/graphsage/graphsage.py
jermainewang: TIL

examples/pytorch/graphsage/graphsage.py
jermainewang: Nested/Inner class should usually be avoided. To me, there is no benefit. In terms of encapsulation, I could still access the inner class as a class attribute like `PoolingAggregator.MLP`. So why not just `MLP`? Usually when I saw a nested class I will try to think "hmm... why is it needed? is there any magic?", but the answer is usually no.

examples/pytorch/graphsage/graphsage.py
hbsun2113: > TIL

What is the meaning?

hbsun2113:
I have updated the code as @BarclayII and @ZiyueHuang suggest. Thx!

jermainewang:
@BarclayII any further comments?

jermainewang:
Have added you to the contributor list.

BarclayII:
The MLP class is still nested.  No big deal though.

Thanks again for your contribution!

PR #406, opened 2019-02-21
[WIP] [NN] Refactor NN package

jermainewang:
## Description

This PR refactor the out-of-date NN package. Some guidelines:
* We won't try to unify different backend for this. User needs to explicitly import, for example `dgl.nn.pytorch`. This is because the interface might be quite different for different backends.
* Try to follow the convention of each backend.
* Don't forget to add docstring.

## Checklist
- [x] Code is well-documented

## Changes

- [x] `dgl.nn.pytorch.conv.GraphConv`
- [x] `dgl.nn.pytorch.softmax.EdgeSoftmax`
- [x] `dgl.nn.mxnet.conv.GraphConv`
- [x] Also change the example to use these predefined modules.

## Additional Note
- I've also done some small fixes on the way to modify the examples
- The incorporation of `nn` modules into the examples seem to have slowed down the speed of execution as there are some redundant operations of setting and popping data.


mufeili:
@jermainewang I assume only the examples will be affected and the tutorials are left unchanged?

jermainewang:
Yes.

mufeili:
@jermainewang I'm done and you can take a look.

mufeili:
@jermainewang I'm done and you may take a look. There are some additional changes that I did not realize until now (the ones in `sampler.h`, `sampler.cc`, `pylintrc` and `task_lint.sh`), I suppose these are your changes?

python/dgl/nn/pytorch/softmax.py
```diff
@@ -0,0 +1,104 @@
...
+        A field name such that g.ndata[max_logits_name] will store
+        :math:`\max_{j\in\mathcal{N}(i)}z_{ij}`.
+    normalizer_name : str
+        A field name such that g.ndata[normalizer_name] will store
+        :math:`\sum_{j\in\mathcal{N}(i)}exp(z_{ij})`
+    """
+    def __init__(self,
+                 logits_name,
+                 max_logits_name,
+                 normalizer_name):
```
jermainewang: Give default names for them?

python/dgl/nn/pytorch/softmax.py
```diff
@@ -0,0 +1,104 @@
...
+            The input edge feature
+        graph : DGLGraph
+            The graph.
+
+        Returns
+        -------
+        Unnormalized scores : torch.Tensor
+            This part gives :math:`\exp(z_{ij})`'s
+        Normalizer : torch.Tensor
+            This part gives :math:`\sum_{j\in\mathcal{N}(i)}\exp(z_{ij})`
```
jermainewang: I guess eventually we want to just return the normalized scores as normal softmax does. But I'm fine with this right now. We could change this when we have our kernel to do that more easily.

python/dgl/nn/pytorch/softmax.py
mufeili: Though the chance is not very high, I want to avoid any potential risk of undesired overwriting.

jermainewang: We could also detect whether the name has been taken and use another one. Then those temporary variables will not be seen by the user.

jermainewang:
> There are some additional changes that I did not realize until now (the ones in `sampler.h`, `sampler.cc`, `pylintrc` and `task_lint.sh`), I suppose these are your changes?

Yes

python/dgl/nn/pytorch/softmax.py
mufeili: Done.

python/dgl/nn/mxnet/conv.py
```diff
@@ -0,0 +1,158 @@
...
+        The learnable bias tensor.
+    """
+    def __init__(self,
+                 in_feats,
+                 out_feats,
+                 norm=True,
+                 dropout=0.,
+                 bias=False,
+                 activation=mx.nd.relu,
+                 feat_name="_gconv_feat"):
```
jermainewang: Let's not put dropout and activation here. My idea is to make it similar to torch.nn.conv2d. Activation and dropout can be handled outside.

python/dgl/nn/pytorch/softmax.py
jermainewang: Then, it's totally safe to remove those "_name" args from the arglist. To think it as a normal pytorch module, the internal local variables are not exposed at all. So are these names. We could also change `GraphConv` in the same way. What do you think?

mufeili: Yeah, I agree. I just forgot to make those changes. Thanks.

python/dgl/nn/mxnet/conv.py
mufeili: My motivation is to make this more like Kipf's initial implementation of [GraphConvolutionLayer](https://github.com/tkipf/gcn/blob/master/gcn/layers.py#L132) where both dropout and activation are part of it. This also make potential model code more compact and expressive like `nn.Linear(bias=True)` is more compact than performing matrix multiplication, check whether bias is to be added and then perform a bias addition if needed.

jermainewang: Ok, to me this is a more [Tensorflow/Keras way](https://keras.io/layers/core/), but I'm think it's not a big deal. One thing is that you might want to set `activation=None` by default, which is more commonly seen.

jermainewang: Also, dropout is actually a tricky stuff. Where is the right place to apply it? On the input? On the aggregated value? Or after activation? I feel there is so many decisions so it's good to leave it to users. (TF/Keras also does not include dropout).

mufeili: Done.

jermainewang:
Did some minor fix on MXNet side. Thanks Murphy!

PR #408, opened 2019-02-22
[RPC] add C++ RPC infrastructure and distributed sampler

aksnzhy:
## Description
This PR add the C++ RPC infrastructure and distributed sampler API.

Now we need to add Windows wrapper for the c socket API for tcp_socket.cc. Anyone can help me do this job? I don't have the windows development environment.

Also, this code has been tested on multiple machines outside the unittest. If we need to add unittest test for the network code?

## Checklist
- [x] Code is well-documented

## Changes
- [x]  python/dgl/contrib/sampling/dis_sampler.py
- [x]  python/dgl/network.py
- [x]  src/graph/network/communicator.h
- [x]  src/graph/network/msg_queue.cc
- [x]  src/graph/network/msg_queue.h
- [x]  src/graph/network/serialize.cc
- [x]  src/graph/network/serialize.h
- [x]  src/graph/network/socket_communicator.cc
- [x]  src/graph/network/socket_communicator.h
- [x]  src/graph/network/tcp_socket.cc
- [x]  src/graph/network/tcp_socket.h
- [x]  src/graph/network.cc
- [x]  src/graph/network.h


BarclayII:
I assume that I only need to port `TCPSocket` to Windows?  The implementation would be a little bit different between WinSock and UNIX sockets.

Re unit test: I think it's better to have some sort of tests, otherwise I wouldn't know whether Windows can work.

BarclayII:
Also, what about PR #325 ?

aksnzhy:
> Also, what about PR #325 ?

I closed this PR. It is out of date.

aksnzhy:
> I assume that I only need to port `TCPSocket` to Windows? The implementation would be a little bit different between WinSock and UNIX sockets.
> 
> Re unit test: I think it's better to have some sort of tests, otherwise I wouldn't know whether Windows can work. IMO they don't _have to_ be integrated into CI (like unit tests), but I need some scripts to set up a simulated, or connect to an existing, distributed environment.

Yes, only the TCPSocket class need to be changed. The other components are all platform-agnostic.

I will add some unittests for distributed sampler.

src/graph/graph_apis.cc
```diff
@@ -494,4 +499,72 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLGraphRandomWalk")
...
+    if (size <= 0) {
+      LOG(ERROR) << "Receive error: (size: " << size << ")";
+    }
+    NodeFlow nf;
+    network::DeserializeSampledSubgraph(&nf, data_buffer);
+    std::vector<NodeFlow> subgs(1);
+    subgs[0] = nf;
+    *rv = ConvertSubgraphToPackedFunc(subgs);
+  });
+
```
zheng-da: It seems the CAPIs can be in another namespace.

src/graph/network/serialize.cc
```diff
@@ -0,0 +1,157 @@
...
+#include <dgl/immutable_graph.h>
+
+#include "serialize.h"
+
+namespace dgl {
+namespace network {
+
+const int kNumTensor = 7;  // We need to serialize 7 conponents (tensor) here
+
+int64_t SerializeSampledSubgraph(char* &data,
```
zheng-da: a reference to a pointer is weird.

src/graph/graph_apis.cc
```diff
@@ -494,4 +499,72 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLGraphRandomWalk")
...
+    network::Communicator* comm = new network::SocketCommunicator();
+    comm->Initialize(false, ip.c_str(), port, num_sender, queue_size);
+    CommunicatorHandle chandle = static_cast<CommunicatorHandle>(comm);
+    *rv = chandle;
+  });
+
+DGL_REGISTER_GLOBAL("graph_index._CAPI_ReceiverRecvSubgraph")
+.set_body([] (DGLArgs args, DGLRetValue* rv) {
+    // The data buffer will be allocated just once 
+    static char* data_buffer = new char[kMaxBufferSize];
```
zheng-da: The memory is allocated when the process starts regardless of how people use DGL. Maybe allocate memory more explicitly. e.g., only when people starts to use the distributed module.

src/graph/network/socket_communicator.cc
```diff
@@ -0,0 +1,191 @@
...
+  } else {
+    is_sender_ = false;
+    return InitReceiver(ip, port, num_sender, queue_size);
+  }
+}
+
+bool SocketCommunicator::InitSender(const char* ip, int port) {
+  // Sender only has a client socket
+  socket_.resize(1);
+  socket_[0] = new TCPSocket();
```
zheng-da: why not use shared_ptr?

src/graph/network/msg_queue.cc
```diff
@@ -0,0 +1,189 @@
...
+
+MessageQueue::~MessageQueue() {
+  std::lock_guard<std::mutex> lock(mutex_);
+  if (nullptr != queue_) {
+    delete [] queue_;
+    queue_ = nullptr;
+  }
+}
+
+int MessageQueue::Add(const char* src, int size, bool is_blocking) {
```
zheng-da: this is a large memory copy.

src/graph/network/msg_queue.cc
```diff
@@ -0,0 +1,189 @@
...
+    LOG(ERROR) << "Message size (" << size << ") is negative or zero.";
+    return -1;
+  }
+  std::unique_lock<std::mutex> lock(mutex_);
+  if (finished_producers_.size() >= num_producers_) {
+    LOG(ERROR) << "Can't add to buffer when flag_no_more is set.";
+    return -1;
+  }
+  if (size > free_size_ && !is_blocking) {
+    return 0;
```
zheng-da: a message is dropped. no warning?

python/dgl/graph_index.py
```diff
@@ -1148,3 +1151,69 @@ def _uniform_sampling(gidx, seed_ids, neigh_type, num_hops, expand_factor):
     assert len(seed_ids) in _NEIGHBOR_SAMPLING_APIS.keys()
     return _NEIGHBOR_SAMPLING_APIS[len(seed_ids)](gidx._handle, *seed_ids, neigh_type,
                                                   num_hops, expand_factor, num_seeds)
+
+##### Distributed sampler infrastructure #####
```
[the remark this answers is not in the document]
aksnzhy: I moved the code into network.py

src/graph/graph_apis.cc
aksnzhy: I moved the code into network.cc

src/graph/network/serialize.cc
aksnzhy: Changed it to `char** data `

src/graph/graph_apis.cc
aksnzhy: Changed it by using a global pointer, which will only be allocated by `CreateReceiver` API

src/graph/network/msg_queue.cc
aksnzhy: It's not very easy to implement zero-copy here

src/graph/network/msg_queue.cc
aksnzhy: Add warning here

src/graph/network/socket_communicator.cc
aksnzhy: I think it is better to control the pointer by handle in Finalize() method.

jermainewang:
First of all, @aksnzhy @zheng-da do we agree that this will not merged into master because this feature is not part of the release plan?

python/dgl/contrib/sampling/dis_sampler.py
```diff
@@ -0,0 +1,88 @@
...
+class Sender(object):
+    """The Sender class for distributed sampler.
+
+    Users use the Sender class to send sampled subgraph 
+    to remote trainer machine.
+
+    Parameters
+    ----------
+    ip : ip address of remote machine
+    port : listen port of remote machine
```
jermainewang: Please follow [numpy comment style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html).

python/dgl/contrib/sampling/dis_sampler.py
```diff
@@ -0,0 +1,88 @@
+# This file contains distributed samplers.
+from ...node_flow import NodeFlow
+from ...network import _create_sender, _create_receiver
+from ...network import _send_subgraph, _recv_subgraph
+from ...network import _finalize_sender, _finalize_receiver
+
+class Sender(object):
```
jermainewang: Rename it to `SamplerSender` caz it is not a general-purpose sender in RPC.

python/dgl/contrib/sampling/dis_sampler.py
```diff
@@ -0,0 +1,88 @@
...
+        nodeflow_list : a list of nodeflow object
+        """
+        _batch_send_subgraph(self._sender, nodeflow_list)
+
+    def Finalize(self):
+        """Finalize Sender
+        """
+        _finalize_sender(self._sender)
+
+class Receiver(object):
```
jermainewang: Receiver <- SamplerReceiver

python/dgl/contrib/sampling/dis_sampler.py
```diff
@@ -0,0 +1,88 @@
...
+        Parameter
+        ---------
+        nodeflow_list : a list of nodeflow object
+        """
+        _batch_send_subgraph(self._sender, nodeflow_list)
+
+    def Finalize(self):
+        """Finalize Sender
+        """
+        _finalize_sender(self._sender)
```
jermainewang: When the `Finalize` is called? In other word, why not put in `__del__`? Unless you have following scenario:
```python
sender = Sender(...)
# do some sending stuff
sender.Finalize()
sender.Restart(...)  # however, I didn't see any restart functions here.
```

aksnzhy:
> First of all, @aksnzhy @zheng-da do we agree that this will not merged into master at the moment because this feature is not part of the release plan?

I agree that we will not merge it into 0.2 version. I suggest that we can merge the distributed training feature with a complete application demo that users can play with.

python/dgl/contrib/sampling/dis_sampler.py
```diff
@@ -0,0 +1,88 @@
...
+        self._queue_size = queue_size
+        self._receiver = _create_receiver(ip, port, num_sender, queue_size)
+
+    def Receive(self):
+        """ Receive data from sender and construct sampled subgraph.
+        Note that, in distributed sampler, the parent graph of NodeFlow 
+        in will be set to None object.
+        """
+        sgi = _recv_subgraph(self._receiver)
+        return NodeFlow(None, sgi)
```
jermainewang: Could you add a line of comment about the `None` (e.g. the received nodeflow has no parent graph blahblah...)?

python/dgl/contrib/sampling/dis_sampler.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def BatchReceive(self):
```
jermainewang: I wonder what is the batch size?

python/dgl/contrib/sampling/dis_sampler.py
```diff
@@ -0,0 +1,88 @@
...
+        sgi_list = _batch_recv_subgraph(self._receiver)
+        nodeflow_list = []
+        for sgi in sgi_list:
+            nodeflow_list.append(NodeFlow(None, sgi))
+        return nodeflow_list
+
+    def Finalize(self):
+        """Finalize Receiver
+        """
+        _finalize_receiver(self._receiver)
```
jermainewang: Same as the Finalize function in Sender.

python/dgl/graph_index.py
```diff
@@ -915,7 +915,13 @@ class NodeFlowIndex(GraphIndex):
...
     """
     def __init__(self, handle, parent, node_mapping, edge_mapping, layers, flows):
-        super(NodeFlowIndex, self).__init__(handle, parent.is_multigraph(), parent.is_readonly())
+        if parent is None:
+            super(NodeFlowIndex, self).__init__(handle, False, True)
+        else:
+            super(NodeFlowIndex, self).__init__(
+                handle,
+                parent.is_multigraph(),
+                parent.is_readonly())
```
jermainewang: @zheng-da does NodeFlow support non-readonly mode right now?

jermainewang:
> > First of all, @aksnzhy @zheng-da do we agree that this will not merged into master at the moment because this feature is not part of the release plan?
> 
> I agree that we will not merge it into 0.2 version. I suggest that we can merge the distributed training feature with a complete application demo that users can play with.

Don't worry. We could merge one component at a time until we got the complete application.

python/dgl/contrib/sampling/dis_sampler.py
aksnzhy: Solved.

python/dgl/contrib/sampling/dis_sampler.py
aksnzhy: Solved.

python/dgl/contrib/sampling/dis_sampler.py
aksnzhy: Solved.

python/dgl/contrib/sampling/dis_sampler.py
aksnzhy: Finalize() will send a special msg to Receiver node, and Receive node can clean up the corresponding backend threads when a Sender finishes its job. If user don't invoke this method, nothing will happen in sender. But it is a waste of thread for Receive node.

aksnzhy: I think it's more flexible to add a Finalize() API to users to control the Sender.

aksnzhy: I don't know if we have the scenario, which we should invoke the Finalize() API and then start a new Sender, and these two Sender have the same life cycle.

python/dgl/contrib/sampling/dis_sampler.py
aksnzhy: Solved. I think we need a better support for the `none-parent` graph.

python/dgl/contrib/sampling/dis_sampler.py
aksnzhy: BatchReceive will return a list of NodeFlow. But now I'm considering if we need the `BatchSend` and `BatchReceive` API. These two API have not beed implemented yet.

python/dgl/contrib/sampling/dis_sampler.py
aksnzhy: The Finalize() of Receiver will clean up some socket memory and stop listening. I don't know if we have the scenario we need to stop and start Receiver many times.

zheng-da:
let's merge this component to this branch first: https://github.com/dmlc/dgl/tree/giant
We do need to put all the components together as soon as possible to demonstrate/evaluate our solution on the giant graph.

python/dgl/graph_index.py
zheng-da: NodeFlow is readonly right now.

python/dgl/contrib/sampling/dis_sampler.py
jermainewang: If `Finalize` is the right function to be called, then my point is to invoke it automatically once the sender object is out-of-scope. This makes it easier to handle exceptions and resources can be correctly released. For example:
```python
with SamplerSender(...) as sender:
  # do a  lot of stuff
# sender goes out-of-scope, __del__ function will call Finalize.
```

python/dgl/contrib/sampling/dis_sampler.py
jermainewang: > and the size of list is controlled by remote sampler

What does it mean? Is there any batch size argument I can set? Or the batch size is non-deterministic?

python/dgl/graph_index.py
jermainewang: Another question is if the parent is None, what happens when `copy_to/from_parent` is called?

jermainewang: Will it be changed in the future?

python/dgl/contrib/sampling/dis_sampler.py
aksnzhy: Ok, I know your concern. If this is the only scenario we may meet in DGL, that's no problem to put `finalize` into `__del__`.

python/dgl/contrib/sampling/dis_sampler.py
aksnzhy: The batch size is non-derterministic.

src/graph/network/msg_queue.cc
zheng-da: how much does the memory copy affect the performance?

python/dgl/contrib/sampling/dis_sampler.py
zheng-da: This batch size isn't observed by users. The remote sampler sends sampled results in a batch size that works the best for the sampler. It should be fine.

python/dgl/graph_index.py
zheng-da: A user should get `NodeFlow` whose `parent` is None. This is a tempoary solution because we don't have a distributed DGLGraph. Once we have it, we can set `parent` correctly.

zheng-da:
@aksnzhy could you remove the windows test from the CI for now? This PR is going to be merged to a dev branch. I think it's fine to not include the windows test for now.

src/graph/network/msg_queue.cc
aksnzhy: Let me evaluate it by removing the message queue for a test.

BarclayII:
I have updated `tcp_socket.cc` with Windows socket implementation.

aksnzhy:
It seems that the NodeFlow and Sampler API has been changed. I need to update the distributed sampler code.

aksnzhy:
@jermainewang Could we change the branch from `dgl:gaint` to `dgl:master`. The windows support has been completed and I have updated my code with the new NodeFlow code in master branch.

src/graph/network.cc
```diff
@@ -105,18 +82,18 @@ DGL_REGISTER_GLOBAL("network._CAPI_ReceiverRecvSubgraph")
     if (size <= 0) {
       LOG(ERROR) << "Receive error: (size: " << size << ")";
     }
-    NodeFlow nf;
+    NodeFlow* nf = new NodeFlow();
```
zheng-da: how is the memory free'd?

aksnzhy: @zheng-da 
I do the same allocation like this:

https://github.com/dmlc/dgl/blob/eb1acecd0d9057d116fa8c8687123600a5065557/src/graph/sampler.cc#L836

aksnzhy: How does the sampler free the memory?

BarclayII: They are manually freed from Python code after instantiation of `NodeFlow` objects (`python/dgl/nodeflow.py:59`)

src/graph/network/serialize.cc
```diff
@@ -0,0 +1,165 @@
...
+namespace network {
+
+const int kNumTensor = 7;  // We need to serialize 7 conponents (tensor) here
+
+int64_t SerializeSampledSubgraph(char* data,
+                                 const ImmutableGraph::CSR::Ptr csr,
+                                 const IdArray& node_mapping,
+                                 const IdArray& edge_mapping,
+                                 const IdArray& layer_offsets,
+                                 const IdArray& flow_offsets) {
```
BarclayII: This interface makes changing the nodeflow structure subtle; one needs to also change the interface here (and the deserialization, and all the calling statements).

Is it possible to change this to something like `size_t SerializeNodeFlow(char *buffer, const NodeFlow &nf)`?

aksnzhy: No problem. I'll do this change after new NodeFlow stricture merged.

BarclayII: If "new NodeFlow structure" refers to the one with optional tensor data (which is the one I'm using) then it won't be merged until after PR #453.  It's your call but I would prefer changing the interface now so later on NodeFlow could be extended further more easily.

PR #413, opened 2019-02-25
[Bugfix] Removing non-determinism in data loading

### LINKED ISSUE #412: Fixing random seeds does not yield deterministic results
- [Related discussion thread](https://discuss.dgl.ai/t/how-to-fix-random-seeds-in-dgl-in-training/65/3)

It has been observed by Erutan-pku, @hbsun2113 , me and @yzh119 that after fixing NumPy random seed, PyTorch random seed and `cudnn.deterministic=True` we still get different results across runs. There can be two possibilities:
1. This is a bug.
2. Some system optimization involves randomness, in which case we should provide APIs for users to fix results.

BarclayII:
## Description
Fixes #412.
@mufeili @yzh119 @hbsun2113 Please check if this fixes non-determinism with custom seed.


yzh119:
LGTM

mufeili:
LGTM

PR #416, opened 2019-02-26
[Feature]Uniform layer-wise sampler

GaiYu0:
## Description
Uniform layer-wise sampler with test cases.

## Checklist
- [x] Code is well-documented

## Changes
- [x] Uniform layer-wise sampler with test cases.


src/graph/sampler.cc
```diff
@@ -524,4 +525,198 @@ IdArray SamplerOp::RandomWalk(
...
+  const dgl_id_t* seed_data = static_cast<dgl_id_t*>(seed_array->data);
+  size_t seed_len = seed_array->shape[0];
+  std::copy(seed_data, seed_data + seed_len, std::back_inserter(*node_mapping));
+  actl_layer_sizes->push_back(node_mapping->size());
+  probabilities->insert(probabilities->end(), node_mapping->size(), 1);
+
+  size_t curr = 0;
+  size_t next = node_mapping->size();
+  unsigned int rand_seed = time(nullptr);
+  for (auto i = layer_sizes.rbegin(); i != layer_sizes.rend(); ++i) {
```
zheng-da: maybe rename `i` with `layer_size`?

src/graph/sampler.cc
```diff
@@ -524,4 +525,198 @@ IdArray SamplerOp::RandomWalk(
...
+    std::copy(candidate_set.begin(), candidate_set.end(),
+              std::back_inserter(candidate_vector));
+
+    std::unordered_map<dgl_id_t, size_t> n_occurrences;
+    auto n_candidates = candidate_vector.size();
+    for (size_t j = 0; j != *i; ++j) {
+      auto dst = candidate_vector[rand_r(&rand_seed) % n_candidates];
+      if (!n_occurrences.insert(std::make_pair(dst, 1)).second) {
+        ++n_occurrences[dst];
+      }
```
zheng-da: so it doesn't sample the specified number of vertices in a layer?

src/graph/sampler.cc
```diff
@@ -524,4 +525,198 @@ IdArray SamplerOp::RandomWalk(
...
+    }
+
+    for (auto const &pair : n_occurrences) {
+      node_mapping->push_back(pair.first);
+      float p = pair.second * n_candidates / static_cast<float>(*i);
+      probabilities->push_back(p);
```
zheng-da: I wonder if this should be considered as the sampling probability.

src/graph/sampler.cc
```diff
@@ -524,4 +525,198 @@ IdArray SamplerOp::RandomWalk(
...
+void ConstructFlows(const int64_t *indptr,
+                    const dgl_id_t *indices,
+                    const dgl_id_t *eids,
+                    const std::vector<dgl_id_t> &node_mapping,
+                    const std::vector<int64_t> &actl_layer_sizes,
+                    std::vector<int64_t> *sub_indptr,
+                    std::vector<dgl_id_t> *sub_indices,
+                    std::vector<dgl_id_t> *sub_eids,
+                    std::vector<dgl_id_t> *flow_offsets,
+                    std::vector<dgl_id_t> *edge_mapping) {
```
zheng-da: please add comments to describe this function.

src/graph/sampler.cc
```diff
@@ -524,4 +525,198 @@ IdArray SamplerOp::RandomWalk(
...
+  std::cout << "layer_offsets:" << layer_offsets.size() << std::endl
+            << "node_mapping:" << node_mapping.size() << std::endl
+            << "actl_layer_sizes:" << actl_layer_sizes.size() << std::endl
+            << "probabilities:" << probabilities.size() << std::endl
+            << "sub_indptr:" << sub_indptr.size() << std::endl
+            << "sub_indices:" << sub_indices.size() << std::endl
+            << "sub_eids:" << sub_eids.size() << std::endl
+            << "flow_offsets:" << flow_offsets.size() << std::endl
+            << "edge_mapping:" << edge_mapping.size() << std::endl;
+  */
```
zheng-da: please remove them.

src/graph/sampler.cc
```diff
@@ -524,4 +525,198 @@ IdArray SamplerOp::RandomWalk(
   return traces;
 }
 
+void ConstructLayers(const int64_t *indptr,
```
zheng-da: put this function and the other internal functions inside the anonymous namespace.

src/graph/sampler.cc
```diff
@@ -524,4 +525,198 @@ IdArray SamplerOp::RandomWalk(
...
+                     const dgl_id_t *indices,
+                     const IdArray seed_array,
+                     const std::vector<size_t> &layer_sizes,
+                     std::vector<dgl_id_t> *layer_offsets,
+                     std::vector<dgl_id_t> *node_mapping,
```
zheng-da: `node_mapping` is confusing. maybe rename it with `node_set` or `layer_node_set`?

src/graph/sampler.cc
```diff
@@ -524,4 +525,198 @@ IdArray SamplerOp::RandomWalk(
...
+    for (int64_t j = 0; j < src_size; ++j) {
+      source_map.insert(std::make_pair(node_mapping[first + j], first + j));
+    }
+    auto dst_size = actl_layer_sizes[i + 1];
+    for (int64_t j = 0; j < dst_size; ++j) {
+      auto dst = node_mapping[first + src_size + j];
+      typedef std::pair<dgl_id_t, dgl_id_t> id_pair;
+      std::vector<id_pair> neighbor_indices;
+      for (int64_t k = indptr[dst]; k < indptr[dst + 1]; ++k) {
+        auto ret = source_map.find(indices[k]);
```
zheng-da: There are a lot of hashtable lookups here. We might need to accelerate it in the future. Please add a comment here.

src/graph/sampler.cc
```diff
@@ -524,4 +525,198 @@ IdArray SamplerOp::RandomWalk(
...
+  */
+
+  NodeFlow nf;
+
+  int64_t n_nodes = node_mapping.size();
+  int64_t n_edges = edge_mapping.size();
+  auto sub_csr = std::make_shared<ImmutableGraph::CSR>(n_nodes, n_edges);
+  std::copy(sub_indptr.begin(), sub_indptr.end(), sub_csr->indptr.begin());
+  std::copy(sub_indices.begin(), sub_indices.end(), std::back_inserter(sub_csr->indices));
+  std::copy(sub_eids.begin(), sub_eids.end(), std::back_inserter(sub_csr->edge_ids));
```
zheng-da: potentially, you can give `sub_csr->indptr`, `sub_csr->indices` and `sub_csr->edge_ids` to `ConstructFlows` to avoid data copy.

tests/compute/test_node_flow.py
```diff
@@ -233,9 +237,15 @@ def reduce_func(node, ind):
...
     test_basic()
+    print(1)
     test_copy()
+    print(2)
     test_apply_nodes()
+    print(3)
     test_apply_edges()
+    print(4)
     test_flow_compute()
+    print(5)
```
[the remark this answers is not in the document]
GaiYu0: Sorry for these...

src/graph/sampler.cc
GaiYu0: `node_mapping` here corresponds to `NodeFlow::node_mapping`.

python/dgl/contrib/sampling/sampler.py
```diff
@@ -248,8 +259,59 @@ def NeighborSampler(g, batch_size, expand_factor, num_hops=1,
...
+                 shuffle=False, num_workers=1,
+                 return_seed_id=False, prefetch=False):
+    '''Create a sampler that samples neighborhood.
+
+    This creates a NodeFlow loader that samples subgraphs from the input graph
+    with layer-wise sampling. This sampling method is implemented in C and can perform
+    sampling very efficiently.
+
+    The NodeFlow loader returns a list of NodeFlows and a dictionary of additional
+    information about the NodeFlows. The size of the NodeFlow list is the number of workers.
```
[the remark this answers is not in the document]
GaiYu0: I have modified the documentation of `LayerSampler`. Maybe I should leave other modifications for you?

src/graph/sampler.cc
GaiYu0: The sampling is with replacement, which is required by the Monte Carlo method. So the number of sampled nodes may be less than `layer_size`. The number of times that a node is sampled is reflected in its corresponding sampling probability.

src/graph/sampler.cc
GaiYu0: Well, this is not really a probability. Please see my comment above.

GaiYu0:
@zheng-da Please take another look.

python/dgl/contrib/sampling/sampler.py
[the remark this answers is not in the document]
zheng-da: yes, my PR fixes for neighbor sampling.

python/dgl/contrib/sampling/sampler.py
```diff
@@ -248,8 +259,54 @@ def NeighborSampler(g, batch_size, expand_factor, num_hops=1,
...
+    batch_size: The number of NodeFlows in a batch.
+    layer_size: A list of layer sizes.
+    node_prob: the probability that a neighbor node is sampled.
+        Not implemented.
+    seed_nodes: a list of nodes where we sample NodeFlows from.
+        If it's None, the seed vertices are all vertices in the graph.
+    shuffle: indicates the sampled NodeFlows are shuffled.
+    num_workers: the number of worker threads that sample NodeFlows in parallel.
+    return_seed_id: indicates whether to return seed ids along with the NodeFlows.
+        The seed Ids are in the parent graph.
```
zheng-da: please remove this argument.

python/dgl/contrib/sampling/sampler.py
```diff
@@ -248,8 +259,54 @@ def NeighborSampler(g, batch_size, expand_factor, num_hops=1,
...
+
+    Returns
+    -------
+    A NodeFlow iterator
+        The iterator returns a list of batched NodeFlows.
+    '''
+    loader = SampledSubgraphLoader(g, batch_size, 'layer', layer_sizes=layer_sizes,
+                                   neighbor_type=neighbor_type, node_prob=node_prob,
+                                   seed_nodes=seed_nodes, shuffle=shuffle,
+                                   num_workers=num_workers, return_seed_id=return_seed_id)
```
zheng-da: please remove `return_seed_id`.

src/graph/sampler.cc
zheng-da: it should work for both with or without replacement. @ZiyueHuang could you confirm?

zheng-da: neighbor sampling samples neighbors without replacement.

GaiYu0:
The reason for sampling with replacement is that the Monte Carlo method requires I.I.D. samples. Sampling without replacement is non-I.I.D.

zheng-da:
so do you mean we shouldn't do sampling without replacement? does the same apply to neighbor sampling?

GaiYu0:
I think the same applies to neighbor sampling.

ZiyueHuang:
I.I.D. samples are not necessary, neither in practice or analysis, for Monte Carlo methods. Sampling with replacement can get I.I.D. samples which is easy for analysis thus many ML papers prefer this setting...

In neighbor sampling, there are many nodes whose neighborhood size is less than `num_neighbors`, in this case sampling with replacement is too bad which requires more samples than sampling without replacement and the result might not be accurate. But in layerwise sampling, the population size is the number of nodes in the graph, then I think sampling with replacement is also OK.

Sampling with replacement is easy to implement for weighted sampling, but this is not a must even for optimal variance and this might be bad in some cases, for example, when you want to draw samples from a very skew distribution, you may get few distinct elements even your sample size is large. This is too bad in training GCN since many parent nodes might not have sampled neighbors. [This well-known paper from SODA](https://arxiv.org/abs/0803.0473) propose a fantastic algorithm for weighted sampling without replacement, it can even get the optimal result for estimating arbitrary subset sum, not only for estimating the sum of the whole set, for a fixed sample size. I think this algorithm is more appropriate for GCN training, since actually we care about the subset sum (each parent node aggregates their neighbors).

Actually I think the difference between with replacement or without replacement is subtle (I have only read a few famous papers in the sampling area, the experts can correct me if wrong), good algorithms for sampling without replacement is difficult to design since there exists covariance. In practice, more distinct elements in the sampled subset might be better, although we cannot directly see this from the formulas of expectation or variance...

We can merge this PR first, I can reproduce some experiments for layerwise sampling and implement the weighted version when I have time.

PR #424, opened 2019-03-01
[Build][CAPI] Enable cython & add support for vector<void*> return

jermainewang:
## Description

This PR enables cython build to speedup the python->CAPI call. On my machine, I observed a 10x speedup. It also implemented helper functions for returning `vector<void*>` type from C to python. It is useful for returning multiple graph objects (e.g. unbatch API returns multiple graph samples, sampling API returns multiple NodeFlows). An example usage is as follows:

On C++ side,
```c++
// in a.cc file

// A demo API that returns a list of graph handlers
DGL_REGISTER_GLOBAL("a._CAPICreateGraphList")
    const int n = args[0];
    std::vector<GraphInterface*> graphs(n);
    for (int i = 0; i < n; ++i) {
      graphs[i] = new Graph();
    }
    *rv = WrapVectorReturn(std::move(graphs));  // WrapVectorReturn defined in c_api_common.h
  });
```

On python side:
```python
# in a.py
import dgl.utils as utils

def create_graph_list(n):
  wrap = _CAPICreateGraphList(n)
  rst = utils.unwrap_to_ptr_list(wrap)  # rst is now a python list of ctypes.c_void_p
  # create graph index using the graph handlers in the list.
  return [GraphIndex(x) for x in rst]
  
_init_api('dgl.a')
```

## Benchmark
I use following script to compare the time to construct a list of integers directly in python and a list of empty graphs from C:
```python
import time
import dgl.a as a

N = 100000

t0 = time.time()
l = list(range(N))
print(time.time() - t0)

t0 = time.time()
l = a.create_graph_list(n)
print(time.time() - t0)
```

On my machine, it takes ~2ms to create a 100K integer array in python and ~20ms to create 100K graphs from C (with Cython enabled). I think the overhead is pretty OK.

@zheng-da Please see whether the above usage works for you. My next step is to apply it to refactor [here](https://github.com/dmlc/dgl/blob/master/python/dgl/graph_index.py#L1129).

## Checklist
- [x] Code is well-documented

## Details

Currently, the PR used a wrapper structure defined as:
```c++
struct CAPIVectorWrapper {
  std::vector<void*> pointers;
};
```
The returning process from C to python is as follows:
On C side:
  1. C/C++ function generates a `vector<Graph*>` result (Graph can be any class).
  2. New a `CAPIVectorWrapper` and copy data to the `pointers`.
  3. Cast `CAPIVectorWrapper*` as `void*` and pass it to python.

On python side:
  1. Get a `ctypes.c_void_p` that represents the `CAPIVectorWrapper*` pointer.
  2. Get the size and the data pointer using other two C APIs.
  3. Cast the data pointer to `ctypes.POINTER(ctypes.c_void_p * size)`
  4. Copy it to a python list.
  5. Free the `CAPIVectorWrapper` and return.

@tqchen I don't know whether this is the most optimal way of using TVM runtime. Could you help?


jermainewang:
To build cython for developers:
```bash
cd python
python setup.py build_ext --inplace
```

tqchen:
If you would like to adopt TVM's Node system, then you can simply return Array<Graph>, where Graph is backed by the Node system

jermainewang:
@tqchen I think the `Graph` in TVM is more for IR. DGL has it's own `Graph` that is for data graph. You could also view this problem as how to return a vector of arbitrary C-defined data structure to python (thus the type `vector<void*>`).

tqchen:
what I mean is that the node system can be subclassed to put in arbitrary field, so you can have dgl.graph as a node. It allows you to enjoy all the benefit of the node. You can pass things back and forth

BarclayII:
Are we completely migrating to Cython from ctypes then?

jermainewang:
No. ctypes is still available for non-cython users. But for binary distribution, we will use cython.

PR #432, opened 2019-03-05
[BUG] Random walk fpe

BarclayII:
## Description
Currently random walk will crash with floating point exception if the current visited vertex has no successors (hence unable to continue random walking).  This PR fails random walk gracefully in this case.

It seems that OpenMP does not work well with existing C++ exception handling (i.e. `LOG(FATAL)`).  So I removed OpenMP in that routine as well.


BarclayII:
Oops I based off the wrong branch.  Opening a new one.

PR #434, opened 2019-03-06
[Perf] Accelerate block_compute when all nodes are invoked.

zheng-da:
## Description
When all nodes are invoked in block_compute, we should implement it as update_all in DGLGraph.

## Checklist
- [x] Code is well-documented


include/dgl/graph_interface.h
```diff
@@ -333,6 +333,10 @@ class GraphInterface {
    * \return a vector of IdArrays.
    */
   virtual std::vector<IdArray> GetAdj(bool transpose, const std::string &fmt) const = 0;
+
+  virtual std::vector<IdArray> GetNodeFlowSlice(bool transpose, const std::string &fmt,
+                                                size_t layer0_size, size_t layer1_start,
+                                                size_t layer1_end, bool remap) const = 0;
```
jermainewang: Let's try not add more APIs to `GraphInterface` header files, esp. APIs for nodeflow.

python/dgl/nodeflow.py
```diff
@@ -17,6 +17,96 @@
 
 NodeFlowHandle = ctypes.c_void_p
 
+class NodeFlowIndex(GraphIndex):
```
jermainewang: Do we need a `NodeFlowIndex` object? During implementation, I found it unnecessary for most cases. The `block_edges` and `block_adjacency_matrix` could be put in `NodeFlow` directly. If they are internal functions that should not be seen by users. We could make them global functions and pass the necessary arguments to it.

zheng-da: we need. it's mainly used by schedulers and the runtime. These components rely on graph index.

jermainewang: Graph index should be accessible using `nodeflow._graph` right? My point is that all the members of this `NodeFlowIndex` is passed by the `NodeFlow` class, so why bother maintaining another class? They can be put in `NodeFlow` class and accessible from there.

zheng-da: but these components don't have a reference to NodeFlow.

jermainewang: Oh, you want to have `schedule_nodeflow_compute` to reuse schedule logics from `DGLGraph`? but I saw the `self` object has been passed to the function so scheduler and runtime could hold a reference to NodeFlow right?

jermainewang: Another concern is duplicated logic. For example what's the difference of `layer_size` in L25 and L224?

zheng-da: oh, i see. you always pass DGLGraph around.

zheng-da: i see a case that NodeFlowIndex is important. In the distributed sampler, it's much easier to have NodeFlowIndex around.

jermainewang: Ok? Could you explain a little bit?

zheng-da: I just discussed with @aksnzhy, the distributed sampler works fine without NodeFlowIndex.

include/dgl/nodeflow.h
```diff
@@ -0,0 +1,68 @@
...
+ * \param layer0_size the size of the first layer in the block.
+ * \param layer1_start the location where the second layer starts.
+ * \param layer1_end the location where the secnd layer ends.
+ * \param remap Indicates to remap all vertex ids and edge Ids to local Id
+ * space.
+ * \return a vector of IdArrays.
+ */
+std::vector<IdArray> GetNodeFlowSlice(const ImmutableGraph &graph, const std::string &fmt,
+                                      size_t layer0_size, size_t layer1_start,
+                                      size_t layer1_end, bool remap);
```
jermainewang: Must the given `ImmutableGraph` be the graph of `NodeFlow`? If so, maybe use your newly defined `NodeFlow` object as the input?

zheng-da: After NodeFlow is returned to Python, it doesn't exist in C any more.

jermainewang: Oh, I see. The Nodeflow object is destroyed after construction. Is this the problem?

jermainewang: Yeah, just realized.

zheng-da:
BTW, I did some experiments on Reddit. The optimization has about 50% speedup when using MXNet as the backend.

src/graph/nodeflow.cc
```diff
@@ -0,0 +1,75 @@
...
+    IdArray idx = IdArray::Empty({2 * nnz}, DLDataType{kDLInt, 64, 1}, DLContext{kDLCPU, 0});
+    IdArray eid = IdArray::Empty({nnz}, DLDataType{kDLInt, 64, 1}, DLContext{kDLCPU, 0});
+    int64_t *idx_data = static_cast<int64_t*>(idx->data);
+    dgl_id_t *eid_data = static_cast<dgl_id_t*>(eid->data);
+    size_t num_edges = 0;
+    for (size_t i = layer1_start; i < layer1_end; i++) {
+      for (int64_t j = csr->indptr[i]; j < csr->indptr[i + 1]; j++) {
+        // These nodes are all in a layer. We need to remap them to the node id
+        // local to the layer.
+        idx_data[num_edges] = remap ? i - layer1_start : i;
```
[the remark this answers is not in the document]
zheng-da: the current implementation is much more efficient.

PR #452, opened 2019-03-16
[BugFix] Fix performance bug in EdgeSoftmax

ylfdq1118:
## Description
`EdgeSoftmax` module did not pop the temporary max feature. In GAT model, this bug leads to huge slow down in `get_ndata_name` function because GAT uses one single graph, and after `n` epochs, there are `n` max feature in `g.ndata`, and `get_ndata_name` has to retry `n` times to find a usable feature name. Also, the bug leads to extra memory consumption because the reference to tensor is still kept by DGLGraph.

This explains why when running GAT models, the training time of one epoch keeps increasing.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/nn/pytorch/softmax.py
+++ b/python/dgl/nn/pytorch/softmax.py
@@ -84,7 +84,7 @@ class EdgeSoftmax(nn.Module):
         graph.apply_edges(
             lambda edges: {self._logits_name : th.exp(edges.data[self._logits_name] -
                                                       edges.dst[self._max_logits_name])})
-        # pop out temporary feature _max_logits, otherwise get_ndata_name will have huge overhead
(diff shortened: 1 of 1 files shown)


python/dgl/nn/pytorch/softmax.py
```diff
@@ -84,6 +84,8 @@ def forward(self, logits, graph):
         graph.apply_edges(
             lambda edges: {self._logits_name : th.exp(edges.data[self._logits_name] -
                                                       edges.dst[self._max_logits_name])})
+        # pop out temporary feature _max_logits, otherwise get_ndata_name could have huge overhead
+        graph.ndata.pop(self._max_logits_name)
```
jermainewang: How huge? Isn't it a dictionary lookup?

Issue #438, opened 2019-03-08
Picking/unpickling DGLGraph breaks feature set/get


jackroos:
Hi, I am trying to emit BatchedDGLGraph in torch.utils.data.DataLoader, but it failed when num_workers > 0. Could you help me solve this problem?
The error message is as following:
> Traceback (most recent call last):
> File "/xxx/anaconda3/envs/xxx/lib/python3.6/multiprocessing/queues.py", line 234, in _feed
> obj = _ForkingPickler.dumps(obj)
> File "/xxx/anaconda3/envs/xxx/lib/python3.6/multiprocessing/reduction.py", line 51, in dumps
> cls(buf, protocol).dump(obj)
> TypeError: can't pickle module objects

jermainewang:
Hi, could you paste the code here? My guess is you use "register_reduce_func" somewhere and the function/module object is not pickleable. A walkaround is directly pass the reduce func to the message passing APIs you call (e.g. update_all, send_and_recv, etc.).

jackroos:
> Hi, could you paste the code here? My guess is you use "register_reduce_func" somewhere and the function/module object is not pickleable. A walkaround is directly pass the reduce func to the message passing APIs you call (e.g. update_all, send_and_recv, etc.).

I do not use register_reduce_func . I just construct a DGLGraph with nodes, edges, ndata, edata in __getitem__() of my dataset, and use dgl.batch() in collate_fn of dataloader. My original code is a little complicated. I am trying to reproduce this error using toy example. But I find I can't reproduce the error now. By the way, if I remove the BatchedDGLGraph in the output of collate_fn. This error will disappear. So it must be related to the graph.

jackroos:
Actually, I can't even pickle graph.ndata.

jermainewang:
I'll look into this.

jackroos:
It is very weird. When I debug line by line in my code, I find the graph can't be pickled just after it be created and call add_nodes. But when I try to reproduce this in following simple code:
````python
import dgl
import torch
import _pickle

graph = dgl.DGLGraph()
graph.add_nodes(n_boxes, {'type': torch.zeros(n_boxes, dtype=torch.int64)})
 _pickle.dumps(graph)
````
the graph is pickled successfully. Do you have any ideas?

jackroos:
In addition, in my original code, after graph created by:
````
graph = dgl.DGLGraph()
````
[...]

PR #456, opened 2019-03-19
[Bugfix] fix bug in pickling

jermainewang:
## Description

Two bugs w.r.t. pickling (Issue #438 ):
* Pickling `torch.dtype` sometimes fails. So always use our own pickling method.
* `utils.index` is not properly pickled when it represents a slice data.

## Checklist
- [x] Code is well-documented


BarclayII:
Looks good.  Does it also fix the cuda problem in #438 ?

jermainewang:
It works on my machine. Unfortunately, we don't have cuda utest yet. That's sth we should add later.

PR #465, opened 2019-03-27
[RPC] add C++ RPC infrastructure and distributed sampler 

aksnzhy:
@jermainewang @zheng-da @BarclayII  

I re-open a new PR for easier code review.  After the polish of AMLC, I think the C++ RPC and distributed sampler component is reliable and ready to merge. Our evaluation shows that developer can gain 20%-40% speed improvement when using distributed sampler.

Now the API is very easy to use:

sampler:
```python
g = generate_rand_graph(100)
sender = dgl.contrib.sampling.SamplerSender(ip="127.0.0.1", port=50051)
for i, subg in enumerate(dgl.contrib.sampling.NeighborSampler(
            g, 1, 100, neighbor_type='in', num_workers=4)):
    sender.Send(subg)
```

trainer:
```python
g = generate_rand_graph(100)
receiver = dgl.contrib.sampling.SamplerReceiver(ip="127.0.0.1", port=50051)
while True:
    nodeflow = receiver.Receive(g)
    ...
```

We can also start many sampler processes in different machines, and the trainer can recv messages from them concurrently.

 After this merging, I will implement a `distributed kvstore` for giant graph training, by re-using the C++ RPC infrastructure.


python/dgl/contrib/sampling/dis_sampler.py
```diff
@@ -0,0 +1,114 @@
...
+        self._sender = _create_sampler_sender(ip, port)
+
+    def __del__(self):
+        """Finalize Sender
+        """
+        # _finalize_sampler_sender will send a special message
+        # to tell the remote trainer machine that it has finished its job.
+        _finalize_sampler_sender(self._sender)
+
+    def Send(self, nodeflow):
```
jermainewang: Python functions should be lower-case (different from C++).

I might not have time on this. I will leave @BarclayII and @zheng-da on this PR. Thanks for the effort!

zheng-da:
can we use the receiver just like we do for NeighborSampler?

```python
g = generate_rand_graph(100)
for subg in dgl.contrib.sampling.SamplerReceiver(ip="127.0.0.1", port=50051):
   ....
```

zheng-da:
The other thing is that when we create multiple sampler processes to generate subgraphs, a user only needs to load the graph structure once.

python/dgl/contrib/sampling/dis_sampler.py
```diff
@@ -0,0 +1,114 @@
...
+    Parameters
+    ----------
+    ip : str
+        ip address of current trainer machine
+    port : int
+        port of current trainer machine
+    num_sender : int
+        total number of sampler nodes, use 1 by default
+    """
+    def __init__(self, ip, port, num_sender=1):
```
zheng-da: do users have to provide `num_sender` if more than one sender is used?

python/dgl/network.py
```diff
@@ -0,0 +1,132 @@
...
+    ----------
+    sender : ctypes.c_void_p
+        C sender handle
+    nodeflow : NodeFlow
+        NodeFlow object
+    """
+    graph_handle = nodeflow._graph._handle
+    node_mapping = nodeflow._node_mapping.todgltensor()
+    edge_mapping = nodeflow._edge_mapping.todgltensor()
+    # Can we convert NDArray to tensor directly, instead of using toindex()?
```
zheng-da: I think this is a right way.

python/dgl/network.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Parameters
+    ----------
+    sender : ctypes.c_void_p
+        C sender handle
+    nodeflow_list : list
+        a list of NodeFlow objects
+    """
+    print(sender)
+    print(nodeflow_list)
```
zheng-da: remove print

python/dgl/contrib/sampling/dis_sampler.py
```diff
@@ -0,0 +1,114 @@
...
+        """Send sampled subgraph (NodeFlow) to remote trainer.
+
+        Parameters
+        ----------
+        nodeflow : NodeFlow
+            sampled NodeFlow object
+        """
+        _send_subgraph(self._sender, nodeflow)
+
+    def BatchSend(self, nodeflow_list):
```
zheng-da: why this interface is added but not implemented? Do we expect the batched version runs faster?

src/graph/network.cc
```diff
@@ -0,0 +1,124 @@
...
+.set_body([] (DGLArgs args, DGLRetValue* rv) {
+    CommunicatorHandle chandle = args[0];
+    GraphHandle ghandle = args[1];
+    const IdArray node_mapping = IdArray::FromDLPack(CreateTmpDLManagedTensor(args[2]));
+    const IdArray edge_mapping = IdArray::FromDLPack(CreateTmpDLManagedTensor(args[3]));
+    const IdArray layer_offsets = IdArray::FromDLPack(CreateTmpDLManagedTensor(args[4]));
+    const IdArray flow_offsets = IdArray::FromDLPack(CreateTmpDLManagedTensor(args[5]));
+    ImmutableGraph *ptr = static_cast<ImmutableGraph*>(ghandle);
+    network::Communicator* comm = static_cast<network::Communicator*>(chandle);
+    auto csr = ptr->GetInCSR();
```
zheng-da: I think it's better to check out csr as well. Although NodeFlow probably has only in csr, we should add a check for now.

src/graph/network/msg_queue.cc
```diff
@@ -0,0 +1,190 @@
...
+
+MessageQueue::~MessageQueue() {
+  std::lock_guard<std::mutex> lock(mutex_);
+  if (nullptr != queue_) {
+    delete [] queue_;
+    queue_ = nullptr;
+  }
+}
+
+int64_t MessageQueue::Add(const char* src, int64_t size, bool is_blocking) {
```
zheng-da: I actually don't understand why this message queue implementation is preferred. Why not always allocate a new piece of memory to store received data and add the piece of memory into the message queue? In this way, there isn't two extra memory copy (one memory copy when adding to the queue, another memory copy when removing from the queue). Since NodeFlow can be quite large, the extra memory copy can cause significant overhead.

src/graph/network/msg_queue.cc
```diff
@@ -0,0 +1,190 @@
...
+    write_pointer_ = size - size_partial;
+  }
+
+  // not empty signal
+  cond_not_empty_.notify_one();
+
+  return size;
+}
+
+int64_t MessageQueue::Add(const string &src, bool is_blocking) {
```
zheng-da: what is the use case of this method?

src/graph/network/serialize.h
```diff
@@ -0,0 +1,53 @@
...
+ * \param layer_offsets layer offsets in NodeFlowIndex
+ * \param flow_offsets flow offsets in NodeFlowIndex
+ * \return the total size of the serialized binary data
+ */
+int64_t SerializeSampledSubgraph(char* data,
+                                 const ImmutableGraph::CSR::Ptr csr,
+                                 const IdArray& node_mapping,
+                                 const IdArray& edge_mapping,
+                                 const IdArray& layer_offsets,
+                                 const IdArray& flow_offsets);
```
zheng-da: this interface is defined specifically for the current NodeFlow. It'll be extended after NodeFlow from random walker sampler is added. Can we make this interface a little more flexible?

src/graph/network/socket_communicator.cc
```diff
@@ -0,0 +1,213 @@
...
+  } else {
+    is_sender_ = false;
+    return InitReceiver(ip, port, num_sender, queue_size);
+  }
+}
+
+bool SocketCommunicator::InitSender(const char* ip, int port) {
+  // Sender only has a client socket
+  socket_.resize(1);
+  socket_[0] = new TCPSocket();
```
zheng-da: why not use shared_pointer?

src/graph/network/socket_communicator.cc
```diff
@@ -0,0 +1,213 @@
...
+bool SocketCommunicator::InitReceiver(const char* ip,
+                                      int port,
+                                      int num_sender,
+                                      int64_t queue_size) {
+  CHECK_GE(num_sender, 1);
+  CHECK_GT(queue_size, 0);
+  // Init message queue
+  num_sender_ = num_sender;
+  queue_size_ = queue_size;
+  queue_ = new MessageQueue(queue_size_, num_sender_);
```
zheng-da: why not use shared_pointer?

src/graph/network/socket_communicator.cc
```diff
@@ -0,0 +1,213 @@
...
+  // Init socket, and socket_[0] is the server socket
+  socket_.resize(num_sender+1);
+  thread_.resize(num_sender);
+  socket_[0] = new TCPSocket();
```
zheng-da: why not use shared_pointer?

src/graph/network/socket_communicator.cc
```diff
@@ -0,0 +1,213 @@
...
+  std::string accept_ip;
+  int accept_port;
+  for (int i = 1; i <= num_sender_; ++i) {
+    socket_[i] = new TCPSocket();
+    if (server->Accept(socket_[i], &accept_ip, &accept_port) == false) {
+      LOG(ERROR) << "Error on accept socket.";
+      return false;
+    }
+    // new thread for the socket
+    thread_[i-1] = new std::thread(MsgHandler, socket_[i], queue_);
```
zheng-da: where is thread object deleted?

src/graph/network/socket_communicator.cc
```diff
@@ -0,0 +1,213 @@
...
+    // First recv the size
+    int64_t received_bytes = 0;
+    int64_t data_size = 0;
+    while (received_bytes < sizeof(int64_t)) {
+      int64_t max_len = sizeof(int64_t) - received_bytes;
+      int64_t tmp = socket->Receive(
+        reinterpret_cast<char*>(&data_size)+received_bytes,
+        max_len);
+      received_bytes += tmp;
+    }
```
zheng-da: it seems you use a while loop to send and receive data in a buffer everywhere. why not move the while loop in the Send and Receive method?

tests/compute/test_dis_sampler.py
```diff
@@ -0,0 +1,49 @@
...
+def start_trainer():
+    g = generate_rand_graph(100)
+    recv = dgl.contrib.sampling.SamplerReceiver(ip="127.0.0.1", port=50051)
+    nodeflow = recv.Receive(g)
+    print("----- trainer ------")
+    print(nodeflow._node_mapping.todgltensor())
+    print(nodeflow._edge_mapping.todgltensor())
+    print(nodeflow._layer_offsets)
+    print(nodeflow._block_offsets)
+    print("----- trainer ------")
```
zheng-da: I think it's better to add checks in the unit test instead of just printing them.

aksnzhy:
> can we use the receiver just like we do for NeighborSampler?
> 
> ```python
> g = generate_rand_graph(100)
> for subg in dgl.contrib.sampling.SamplerReceiver(ip="127.0.0.1", port=50051):
>    ....
> ```

aksnzhy:
> can we use the receiver just like we do for NeighborSampler?
> 
> ```python
> g = generate_rand_graph(100)
> for subg in dgl.contrib.sampling.SamplerReceiver(ip="127.0.0.1", port=50051):
>    ....
> ```

python/dgl/contrib/sampling/dis_sampler.py
aksnzhy: Yes

src/graph/network.cc
aksnzhy: You mean that we can just serialize an empty out-csr for now?

src/graph/network/msg_queue.cc
aksnzhy: Allocate new memory has the following disadvantages:

1. Allocate a new piece of memory is much more expensive than copy a piece of memory.
2. We should control the buffer size to avoid allocating too much memory when we have lot of remote samplers.

src/graph/network/msg_queue.cc
aksnzhy: It is just a wrapper because in many cases we use std::string to hold tmp data.

src/graph/network/serialize.h
aksnzhy: I discussed this with @BarclayII  and find that it is not very easy to change this interface.

src/graph/network/socket_communicator.cc
aksnzhy: We don't know the message size inside the Send() and Recv(). So we should design the protocol outside the Send() and Recv().

src/graph/network.cc
zheng-da: i guess just double check if GetOutCSR() == nll.

aksnzhy:
@zheng-da I updated this PR, and it is ready to merge. The main update is that I add a SamplerPool to help users start multiple processes to avoid copy Graph in memory many time. I also change the unittest for distributed sampler. I remain the original implementation in somewhere and it may be changed in the future.

PR #468, opened 2019-03-29
[Feature] Create shared memory graph store.

zheng-da:
## Description
This is to create a shared memory graph store. In this way, the graph store can be created in a separate process and loads a graph to memory. Worker processes can link to the graph store and create node embeddings and run graph computation in the graph store directly.

Potentially, there is race condtion when we update the node/edge embeddings in the graph store. We'll test if this can cause troubles later (e.g., how does it affact the training in GCN with control variate).

## Checklist
- [x] Code is well-documented


python/dgl/contrib/graph_store.py
```diff
@@ -0,0 +1,355 @@
...
+            idx = create_graph_index(graph_data=graph_data, multigraph=multigraph, readonly=True)
+            transpose = False if edge_dir == 'in' else True
+            csr = idx.adjacency_matrix_scipy(transpose, 'csr')
+            return csr.indptr, csr.indices
+
+class SharedMemoryStoreServer:
+    """The graph store server.
+
+    The server loads graph structure and node embeddings and edge embeddings
+    and store them in shared memory.
```
BarclayII: Describe how the graph name would be used in greater detail (e.g. `The shared memory objects are named after the name of the graph supplied during initialization and the name of the node/edge embeddings`).

python/dgl/contrib/graph_store.py
```diff
@@ -0,0 +1,355 @@
...
+        graph_idx = GraphIndex(multigraph=multigraph, readonly=True)
+        indptr, indices = _to_csr(graph_data, edge_dir, multigraph)
+        graph_idx.from_csr_matrix(indptr, indices, edge_dir, _get_graph_path(graph_name))
+
+        self._graph = DGLGraph(graph_idx, multigraph=multigraph, readonly=False)
+        self._num_workers = num_workers
+        self._graph_name = graph_name
+        self._edge_dir = edge_dir
+
+        def get_graph_info():
```
BarclayII: Docstrings, even if they are RPC methods (same below).

python/dgl/contrib/graph_store.py
```diff
@@ -0,0 +1,355 @@
...
+        dgl.DGLGraph.edges
+        """
+        return EdgeDataView(self._graph, ALL, self._graph_name)
+
+    def run(self):
+        while self._num_workers > 0:
+            self.server.handle_request()
+        self._graph = None
+
+class SharedMemoryGraphStore(DGLGraph):
```
BarclayII: `SharedMemoryGraphStore` -> `SharedMemoryGraph` (it's a `DGLGraph`, right?)

python/dgl/contrib/graph_store.py
```diff
@@ -0,0 +1,355 @@
...
+    """The shared-memory graph store.
```
BarclayII: `DGLGraph using shared memory.`

python/dgl/contrib/graph_store.py
```diff
@@ -0,0 +1,355 @@
...
+        The number of workers that will connect to the server.
+    port : int
+        The port that the server listens to.
+    """
+    def __init__(self, graph_data, edge_dir, graph_name, multigraph, num_workers, port):
+        graph_idx = GraphIndex(multigraph=multigraph, readonly=True)
+        indptr, indices = _to_csr(graph_data, edge_dir, multigraph)
+        graph_idx.from_csr_matrix(indptr, indices, edge_dir, _get_graph_path(graph_name))
+
+        self._graph = DGLGraph(graph_idx, multigraph=multigraph, readonly=False)
```
BarclayII: So we first convert the graph data to CSR, create an *immutable* graph index and initialize from CSR, and then initialize a *mutable* `DGLGraph` atop that?

python/dgl/graph_index.py
```diff
@@ -812,6 +812,56 @@ def from_scipy_sparse_matrix(self, adj):
...
+        indptr = utils.toindex(indptr)
+        indices = utils.toindex(indices)
+        edge_ids = utils.toindex(F.arange(0, len(indices)))
+        self._handle = _CAPI_DGLGraphCSRCreate(
+            indptr.todgltensor(),
+            indices.todgltensor(),
+            edge_ids.todgltensor(),
+            shared_mem_name,
+            self._multigraph,
+            edge_dir)
```
BarclayII: So `from_csr_matrix` actually *creates* a shared memory object if needed.

I feel it's better to have a utility function `create_shared_mem_from_csr_matrix()`, so we can replace this whole method with `from_shared_mem_csr_matrix`.

src/graph/data_apis.cc
```diff
@@ -0,0 +1,61 @@
...
+
+namespace dgl {
+
+std::unordered_map<std::string, DLDataType> dtype_map = {
+  {"float32", DLDataType{kDLFloat, 32, 1}},
+  {"int", DLDataType{kDLInt, 32, 1}},
+  {"uint", DLDataType{kDLUInt, 32, 1}},
+};
+
+DGL_REGISTER_GLOBAL("contrib.graph_store._CAPI_DGLCreateSharedMem")
```
BarclayII: `DGLCreateSharedMem` -> `DGLOpenSharedMem`

src/graph/data_apis.cc
```diff
@@ -0,0 +1,61 @@
...
+    for (size_t i = 0; i < shape.size(); i++)
+      shape[i] = shape_data[i];
+    NDArray arr = NDArray::EmptyShared(mem_name, shape, dtype,
+                                       DLContext{kDLCPU, 0}, is_create);
+    *rv = arr;
+    if (fill == "zero" && is_create)
+      memset(arr->data, 0, arr.GetSize());
+  });
+
+DGL_REGISTER_GLOBAL("contrib.graph_store._CAPI_DGLCreateSharedMemWithData")
```
BarclayII: ditto

python/dgl/contrib/graph_store.py
```diff
@@ -0,0 +1,355 @@
...
+
+    Returns
+    -------
+    SharedMemoryStoreServer
+        The graph store server
+    """
+    return SharedMemoryStoreServer(graph_data, edge_dir, graph_name, multigraph,
+                                   num_workers, port)
+
+def create_graph_store_client(graph_name, store_type, port=8000):
```
BarclayII: `create_graph_from_store` since it is returning a `DGLGraph` and users could run stuff like `in_degrees()` on it.

python/dgl/contrib/graph_store.py
```diff
@@ -0,0 +1,355 @@
...
+        It first creates the node embedding in the server and maps it to the current process
+        with shared memory.
+
+        Parameters
+        ----------
+        ndata_name : string
+            The name of node embedding
+        shape : tuple
+            The shape of the node embedding
+        dtype : string
```
BarclayII: Put valid choices here.

python/dgl/contrib/graph_store.py
```diff
@@ -0,0 +1,355 @@
...
+        Data to initialize graph. Same as networkx's semantics.
+    graph_name : string
+        Define the name of the graph.
+    store_type : string
+        The type of the graph store.
+    num_workers : int
+        The number of workers that will connect to the server.
+    multigraph : bool, optional
+        Whether the graph would be a multigraph (default: False)
+    edge_dir : string
```
BarclayII: Put valid choices here.

python/dgl/contrib/graph_store.py
```diff
@@ -0,0 +1,355 @@
...
+    After the server runs, the graph store clients can access the graph data
+    with the specified graph name.
+
+    Parameters
+    ----------
+    graph_data : graph data
+        Data to initialize graph. Same as networkx's semantics.
+    graph_name : string
+        Define the name of the graph.
+    store_type : string
```
BarclayII: Put valid choices here.

python/dgl/contrib/graph_store.py
```diff
@@ -0,0 +1,355 @@
...
+    """Create the graph store client.
+
+    The client constructs the graph structure and node embeddings and edge embeddings
+    in shared memory that has been loaded by the graph store server.
+
+    Parameters
+    ----------
+    graph_name : string
+        Define the name of the graph.
+    store_type : string
```
BarclayII: Put valid choices here.

python/dgl/graph_index.py
```diff
@@ -812,6 +812,56 @@ def from_scipy_sparse_matrix(self, adj):
...
+    def from_csr_matrix(self, indptr, indices, edge_dir, shared_mem_name=""):
+        """Load a graph from the CSR matrix.
+
+        Parameters
+        ----------
+        indptr : a 1D tensor
+            index pointer in the CSR format
+        indices : a 1D tensor
+            column index array in the CSR format
+        edge_dir : string
```
BarclayII: Put valid choices here.

python/dgl/graph_index.py
```diff
@@ -812,6 +812,56 @@ def from_scipy_sparse_matrix(self, adj):
...
+
+        Parameters
+        ----------
+        shared_mem_name : string
+            the name of shared memory
+        num_nodes : int
+            the number of nodes
+        num_edges : int
+            the number of edges
+        edge_dir : string
```
BarclayII: Put valid choices here.

src/graph/graph_apis.cc
```diff
@@ -96,6 +96,45 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLGraphCreate")
     *rv = ghandle;
   });
 
+DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLGraphCSRCreate")
```
BarclayII: Similar to the rationale of `from_csr_matrix`, I feel that a utility function that creates a shared memory object from CSR + `DGLGraphCSRCreateMMap` would be cleaner.

src/graph/immutable_graph.cc
```diff
@@ -20,6 +22,80 @@ bool binary_search(ForwardIt first, ForwardIt last, const T& value) {
...
+  const dgl_id_t *indices_data = static_cast<dgl_id_t*>(index_arr->data);
+  const dgl_id_t *edge_id_data = static_cast<dgl_id_t*>(edge_id_arr->data);
+  CHECK_EQ(indptr_data[0], 0);
+  CHECK_EQ(indptr_data[num_vertices], num_edges);
+  indptr.insert_back(indptr_data, num_vertices + 1);
+  indices.insert_back(indices_data, num_edges);
+  edge_ids.insert_back(edge_id_data, num_edges);
+  this->mem = mem;
+#else
+  LOG(FATAL) << "Windows doesn't support ImmutableGraph with shared memory";
```
BarclayII: `Shared memory support not yet implemented`.

Windows does support shared memory.

src/graph/immutable_graph.cc
```diff
@@ -20,6 +22,80 @@ bool binary_search(ForwardIt first, ForwardIt last, const T& value) {
...
+  indptr.init(addr1, num_vertices + 1, num_vertices + 1);
+  void *addr = addr1 + num_vertices + 1;
+  dgl_id_t *addr2 = static_cast<dgl_id_t *>(addr);
+  indices.init(addr2, num_edges, num_edges);
+  addr = addr2 + num_edges;
+  dgl_id_t *addr3 = static_cast<dgl_id_t *>(addr);
+  edge_ids.init(addr3, num_edges, num_edges);
+  this->mem = mem;
+#else
+  LOG(FATAL) << "Windows doesn't support ImmutableGraph with shared memory";
```
BarclayII: ditto

include/dgl/runtime/ndarray.h
```diff
@@ -186,6 +205,46 @@ class NDArray {
  */
 inline bool SaveDLTensor(dmlc::Stream* strm, const DLTensor* tensor);
 
+#ifndef _WIN32
```
BarclayII: Can we have this in another header file (like `dgl/runtime/shared_memory.h`)?

src/runtime/ndarray.cc
```diff
@@ -15,6 +21,50 @@ extern "C" void NDArrayDLPackDeleter(DLManagedTensor* tensor);
 namespace dgl {
 namespace runtime {
 
+#ifndef _WIN32
```
BarclayII: Similarly can we have these in another source file (like `src/runtime/shared_memory.cpp`)?

python/dgl/contrib/graph_store.py
zheng-da: yes

python/dgl/graph_index.py
zheng-da: not necessary. if `shared_mem_name` is given, csr is loaded to shared memory. otherwise, it's loaded to regular memory.

src/graph/data_apis.cc
zheng-da: why use `open`? it is creating/allocating shared memory

src/graph/graph_apis.cc
zheng-da: this is to create a CSR graph. If `shared_mem_name` is provided, it creates shared memory. otherwise, allocates regular memory.

BarclayII: Got it

python/dgl/contrib/graph_store.py
BarclayII: Why do we want a mutable DGLGraph atop an immutable index then?

src/graph/data_apis.cc
BarclayII: Because if `is_create` is False then you are just opening an existing one?

python/dgl/contrib/graph_store.py
zheng-da: it's a bug. it has been fixed.

src/graph/data_apis.cc
zheng-da: i rename it to _CAPI_DGLAllocateSharedMem.

zheng-da:
@BarclayII do you have further comments?

src/graph/data_apis.cc
```diff
@@ -0,0 +1,60 @@
...
+};
+
+DGL_REGISTER_GLOBAL("contrib.graph_store._CAPI_DGLAllocateSharedMem")
+.set_body([] (DGLArgs args, DGLRetValue* rv) {
+    std::string mem_name = args[0];
+    dgl::runtime::NDArray shape_arr = args[1];
+    std::string dtype_str = args[2];
+    std::string fill = args[3];
+    bool is_create = args[4];
+    auto it = dtype_map.find(dtype_str);
```
jermainewang: For dtype, we shall use the type defined [here](https://github.com/dmlc/dgl/blob/master/python/dgl/_ffi/runtime_ctypes.py#L36). To pass it to CAPI, checkout [`DGLArrayAlloc`](https://github.com/dmlc/dgl/blob/cd4f0a2bfce4b58c4f5006d9760eb446df05289b/include/dgl/runtime/c_runtime_api.h#L373).

src/graph/data_apis.cc
```diff
@@ -0,0 +1,60 @@
...
+.set_body([] (DGLArgs args, DGLRetValue* rv) {
+    std::string mem_name = args[0];
+    NDArray data = args[1];
+    std::vector<int64_t> shape(data->shape, data->shape + data->ndim);
+    size_t mem_size = 1;
+    for (auto s : shape)
+      mem_size *= s;
+    mem_size *= data->dtype.bits / 8;
+    NDArray arr = NDArray::EmptyShared(mem_name, shape, data->dtype,
+                                       DLContext{kDLCPU, 0}, true);
```
jermainewang: Why `is_create` is always true here?

src/graph/immutable_graph.cc
```diff
@@ -58,6 +134,8 @@ ImmutableGraph::EdgeArray ImmutableGraph::CSR::GetEdges(IdArray vids) const {
     dgl_id_t vid = vid_data[i];
     int64_t off = this->indptr[vid];
     const int64_t len = this->GetDegree(vid);
+    if (len == 0)
+      continue;
```
jermainewang: Will this happen?

include/dgl/vector.h
```diff
@@ -0,0 +1,208 @@
...
+    }
+    for (size_t i = this->curr; i < new_size; i++)
+    this->arr[i] = 0;
+    this->curr = new_size;
+  }
+
+  /* Similar to std::vector::clear. */
+  void clear() {
+    this->curr = 0;
+  }
```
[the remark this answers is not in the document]
zheng-da: this semantic is the same as std::vector.

include/dgl/vector.h
```diff
@@ -0,0 +1,208 @@
...
+    if (!this->own) {
+      CHECK_LE(curr + len, capacity);
+    } else if (curr + len > capacity) {
+      this->capacity = curr + len;
+      this->arr = static_cast<T *>(realloc(this->arr, this->capacity * sizeof(T)));
+      CHECK(this->arr) << "can't allocate memory for a larger vector.";
+    }
+    std::copy(val, val + len, this->arr + curr);
+    curr += len;
+  }
```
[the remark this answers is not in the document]
zheng-da: push_back is the same as the one in std::vector. insert_back doesn't exist in std::vector.

python/dgl/contrib/graph_store.py
```diff
@@ -0,0 +1,373 @@
+import os
+import time
+import scipy
+from xmlrpc.server import SimpleXMLRPCServer
+import xmlrpc.client
```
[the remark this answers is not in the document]
zheng-da: i searched for rpc in python. it seems this one is the most convinient one. it is distributed with python by default. so we don't need to install any python package.

src/graph/sampler.cc
```diff
@@ -601,17 +601,18 @@ namespace {
                       const dgl_id_t *eids,
                       const std::vector<dgl_id_t> &node_mapping,
                       const std::vector<int64_t> &actl_layer_sizes,
-                      std::vector<int64_t> *sub_indptr,
-                      std::vector<dgl_id_t> *sub_indices,
-                      std::vector<dgl_id_t> *sub_eids,
+                      vector<int64_t> *sub_indptr,
+                      vector<dgl_id_t> *sub_indices,
+                      vector<dgl_id_t> *sub_eids,
```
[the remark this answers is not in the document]
zheng-da: i can change its name, but I don't think there is any problem of using the same set of API.

src/graph/data_apis.cc
zheng-da: actually, the data type string should be consistent with https://github.com/dmlc/dgl/blob/master/python/dgl/backend/pytorch/tensor.py#L10
The data type is extract from the backend tensor.

src/graph/immutable_graph.cc
zheng-da: yes, I ran into this corner case before.

python/dgl/contrib/graph_store.py
```diff
@@ -0,0 +1,373 @@
...
+    This is a client to access data in the shared-memory graph store that has loads
+    the graph structure and node embeddings and edge embeddings to shared memory.
+    It provides the DGLGraph interface.
+
+    Parameters
+    ----------
+    graph_name : string
+        Define the name of the graph.
+    port : int
+        The port that the server listens to.
```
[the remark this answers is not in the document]
zheng-da: I don't know you are referring to the implementation of the server side and the client side.

GraphIndex is easy. Just allocate shared memory for CSR.

It seems that you are talking about the server side. More specifically, how do we handle 
`g.ndata['feat'] = arr`?
The behavior I like to have is to create shared memory tensor and copy data in `arr` to the tensor. Potentially, we can modify `Frame` to do so, but I don't think it's cleaner or easier.

include/dgl/vector.h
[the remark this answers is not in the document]
jermainewang: No, `std::vector::clear` will call the destructor of each element stored.

There are also couple of other details. For example, the `push_back` will fail if the size is larger than the capacity but it is not the case for `std::vector`. To avoid confusion, it's better to simply expose APIs for direct memory access such as  `operator []` or `data`. Checkout [`array_view`](https://github.com/dmlc/dmlc-core/blob/master/include/dmlc/array_view.h) from `dmlc-core`. I feel the case is quite similar. The difference is that we need non-const `operator []`, `begin` and `end`.

Are you planning to use the vector together with some other libraries that require `std::vector` interfaces? This would be a strong motivation to have a fully-compatible vector class.

python/dgl/contrib/graph_store.py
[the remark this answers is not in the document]
jermainewang: Client side.

Implement it in `Frame` will be cleaner. `g.ndata['feat'] = arr` is `set_column` or `set_row` in frame level and you could still implement the shared memory logic there. Other benefits of doing it in frames:
* Directly support syntax like `g.nodes[[0, 1]].data['h']`.
* The code is reused for node frame, edge frame and potentially message frame.
* No need for a new graph class. We could modify DGLGraph to create shared memory frames and graph index and the rest of the logic are the same.

src/graph/immutable_graph.cc
jermainewang: I mean if `len==0`, the code should not enter look in L133. Why this will be triggered?

zheng-da: ah, `len` is reused. i should rename it.

include/dgl/vector.h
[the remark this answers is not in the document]
zheng-da: I think a better option is to put `vector` inside `CSR`. after all, `vector` is specifically designed for `CSR`. It doesn't mean to be general-purpose.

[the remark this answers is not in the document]
zheng-da: It was implemented so until @BarclayII suggested moving it out. Inside `CSR`, it's just a helper class to implement CSR.

python/dgl/contrib/graph_store.py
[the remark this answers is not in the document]
zheng-da: `SharedMemDGLGraph` is necessary. It's not just mapping data in Frame to shared memory. It also send commands to the server to run operations in the server.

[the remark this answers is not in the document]
zheng-da: On the client side, `g.ndata['feat'] = arr` shouldn't happen. Logically, the client should only access part of node embeddings. You should think `SharedMemDGLGraph` has the same behavior of `DistributedDGLGraph`.

include/dgl/vector.h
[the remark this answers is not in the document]
BarclayII: I see.  I got confused thinking that this `vector` is meant to be general-purpose.

python/dgl/contrib/graph_store.py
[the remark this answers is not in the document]
futurely: xmrpc is slow with large data.

https://bugs.python.org/issue26049

Other choices include apache thrift and google grpc.

https://blog.thriftly.io/know-your-api-protocols

Apache thrift has high throughput while grpc has low latency.

https://colobu.com/2016/09/05/benchmarks-of-popular-rpc-frameworks/

Apache thrift is the fastest in this benchmark.

http://szelei.me/rpc-benchmark-part1/

golong benchmark.

https://github.com/chrislee87/rpc_benchmark

Serde performance comparison.

https://github.com/eishay/jvm-serializers/wiki

https://www.eprosima.com/index.php/resources-all/performance/apache-thrift-vs-protocol-buffers-vs-fast-buffers

https://labs.criteo.com/2017/05/serialization/

Overrall, grpc is probably the best choice.

https://dzone.com/articles/moving-from-apache-thrift-to-grpc-a-perspective-fr

PR #474, opened 2019-04-01
Add demo for distributed sampler

aksnzhy:
## Description
This PR add a demo for distributed sampler, using neighbor sampler on GCN.

## Checklist
- [x] Code is well-documented

## Changes
- [x] Add gcn_ns_sampler.py
- [x] Add  gcn_trainer.py
- [x] Add README.md
- [x] Add run_sampler.sh
- [x] Add run_trainer.sh


examples/mxnet/sampling/dis_sampling/gcn_ns_sampler.py
```diff
@@ -0,0 +1,83 @@
...
+
+def main(args):
+    pool = MySamplerPool()
+    pool.start(args.num_sender, args)
+ 
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='GCN')
+    register_data_args(parser)
+    parser.add_argument("--gpu", type=int, default=-1,
+            help="gpu")
```
zheng-da: why do we need GPU?

examples/mxnet/sampling/dis_sampling/gcn_trainer.py
```diff
@@ -0,0 +1,269 @@
...
+
+    # use optimizer
+    print(model.collect_params())
+    trainer = gluon.Trainer(model.collect_params(), 'adam',
+                            {'learning_rate': args.lr, 'wd': args.weight_decay},
+                            kvstore=mx.kv.create('local'))
+
+    # initialize graph
+    dur = []
+    total_count = 153
```
zheng-da: it is very difficult for users to figure what is the right number of batches in an epoch.

examples/mxnet/sampling/dis_sampling/gcn_ns_sampler.py
aksnzhy: Removed this argument.

examples/mxnet/sampling/dis_sampling/gcn_trainer.py
aksnzhy: Yes it is. Actually I don't know how to solve this problem easily for now.

aksnzhy: The potential solution could be:

1. Send a number to trainer from sender to tell trainer what the this number is.
2. Batch send all the sub-graph in an epoch to trainer. But it may cause the memory and network issue because this batched-nodeflow could be extremely large.

aksnzhy: I think we can just leave this problem here, and solve it after all of the distributed training pipeline completed.

BarclayII: Why do we want the number of batches here?  I don't see why this is a big deal...

aksnzhy:
@BarclayII Do you agree to merge?

examples/mxnet/sampling/dis_sampling/gcn_trainer.py
zheng-da: We need to deliver runnable code to users before 0.3 release.

To really fix this problem, I think we should provide users the notion of iterator and epoch. Just like normal samplers.

zheng-da: what do you think of this API?
```python
receiver = dgl.contrib.sampling.SamplerReceiver(ip=args.ip, port=args.port, num_sender=args.num_sender)
for epoch in range(n_epochs):
    for subg in receiver.get_sampler():
        ...
```

zheng-da: Maybe we propose a set of interface for distributed sampler and have everyone involved in discussion, just like what we did for all other interface before?

zheng-da:
I'm OK with merging this PR. I just think we should provide a way for users to use the distributed sampler easily.

I'm not sure if it's really necessary to have the notion of epoch during the training. But I guess people get used to this concept.

Let's merge this PR and have more discussions about what is the right API for users and system implementation.

examples/mxnet/sampling/dis_sampling/gcn_trainer.py
aksnzhy: This API is good. I think we should figure out a way to define the number of nodeflow in an epoch when all the sampler processes's out are mixed.

aksnzhy: A navie way to solve this problem is that we let one sampler pre-sample an epoch, and send the size of sampled nodeflow to trainer node, and we can use this number to define the boundary of epoch. But it is not very friendly for users who write code in sampler.

PR #481, opened 2019-04-08
[Giant] Distributed KVStore

aksnzhy:
## Description
This PR is under developing. I push it here for reviewing the API of distributed GraphStore.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/contrib/kvstore/kvstore.py
+++ b/python/dgl/contrib/kvstore/kvstore.py
@@ -1,9 +1,8 @@
-# This file contains DGL distributed graph store APIs.
+# This file contains DGL distributed kvstore APIs.
 from ...network import _create_sender, _create_receiver
 from ...network import _finalize_sender, _finalize_receiver
 from ...network import _recv_ip_port, _send_local_ip_port
 from ...network import _recv_graph_store_msg, _send_graph_store_msg
-from ...network import _pull, _push
 
 import time
 import signal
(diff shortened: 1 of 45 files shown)


python/dgl/contrib/kvstore/kvstore.py
```diff
@@ -0,0 +1,186 @@
...
+
+    KVStore also provides three APIs for KVStore-client:
+
+      * init_client(sever_ip, server_port) initialize KVStore-client.
+
+      * pull(feat_name, node_ids) requests the feature values associated with corresponding 
+        feature name and node IDs.
+
+      * push(feat_name, node_ids, node_feats) sends feature name, node IDs, and node features 
+        to the KVStore-server, which uses these data to updating graph embeddings.
```
zheng-da: I think it's better to split the server and the client. It's quite confusing why these two components are in the same class.

python/dgl/network.py
```diff
@@ -95,3 +118,63 @@ def _finalize_receiver(receiver):
         C receiver handle
     """
     _CAPI_DGLFinalizeCommunicator(receiver)
+
+def _send_local_ip_port(sender):
```
zheng-da: why do you need to send ip and port?

python/dgl/contrib/kvstore/kvstore.py
aksnzhy: Ok, I will do this change.

PR #482, opened 2019-04-09
fix self-edge bug in GCN and GAT.

sufeidechabei:
## Description
I find that  some nodes in citation dataset have self-edge orginally, if we use `g.add_edges(g.nodes(), g.nodes())` directly, some nodes will have more than one self-edge.  This PR fixes that.

### DIFF
--- a/examples/mxnet/gat/gat_batch.py
+++ b/examples/mxnet/gat/gat_batch.py
@@ -178,9 +178,12 @@ def main(args):
         test_mask = test_mask.as_in_context(ctx)
         val_mask = val_mask.as_in_context(ctx)
     # create graph
-    g = DGLGraph(data.graph)
+    g = data.graph
     # add self-loop
+    g.remove_edges_from(g.selfloop_edges())
+    g = DGLGraph(g)
     g.add_edges(g.nodes(), g.nodes())
+    
 
     # create model
     model = GAT(g,
--- a/examples/mxnet/gat/gat_batch.py
+++ b/examples/mxnet/gat/gat_batch.py
@@ -183,8 +183,6 @@ def main(args):
     g.remove_edges_from(g.selfloop_edges())
     g = DGLGraph(g)
     g.add_edges(g.nodes(), g.nodes())
-    
-
     # create model
     model = GAT(g,
                 args.num_layers,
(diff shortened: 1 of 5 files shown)


mufeili:
I just took a look and it seems that `data.graph` does not have self edges. Have you verified the issue yourself?

examples/mxnet/gat/gat_batch.py
```diff
@@ -178,9 +178,12 @@ def main(args):
         test_mask = test_mask.as_in_context(ctx)
         val_mask = val_mask.as_in_context(ctx)
     # create graph
-    g = DGLGraph(data.graph)
+    g = data.graph
     # add self-loop
+    g.remove_edges_from(g.selfloop_edges())
+    g = DGLGraph(g)
     g.add_edges(g.nodes(), g.nodes())
+
```
mufeili: One empty line only to be consistent with the rest of the file.

examples/pytorch/gcn/gcn_mp.py
```diff
@@ -147,10 +147,13 @@ def main(args):
...
     # graph preprocess and calculate normalization factor
-    g = DGLGraph(data.graph)
-    n_edges = g.number_of_edges()
+    g = data.graph
+    g.remove_edges_from(g.selfloop_edges())
+    g = DGLGraph(g)
     # add self loop
     g.add_edges(g.nodes(), g.nodes())
+    n_edges = g.number_of_edges()
+    # create model
```
mufeili: Remove L156, which seems to be wrong.

sufeidechabei:
> Looks good overall. Some minor comments.

Done!

yzh119:
Why more than one self-edge matters? AFAIN self edge does not affect Laplacian matrix, does DGL have problem handling multi-edge?

mufeili:
@yzh119 , see below for your reference.

```
import torch
from dgl import DGLGraph

g = DGLGraph()
g.add_nodes(1)
g.add_edges([0, 0], [0, 0])
g.ndata['h'] = torch.tensor([[1.]])
g.update_all(fn.copy_src('h', 'm'), fn.sum('m', 'h'))
```

shengwenLeong:
In the tutorial of GCN in the document, the code didn't contain adding the self-edge operation. so should we fix it?

mufeili:
@shengwenLeong Nice catch! Thank you for the notice.

@HQ01 Could you please help fixing it?

HQ01:
@shengwenLeong tutorial has been fixed

PR #484, opened 2019-04-09
[Model] add multiprocessing training with sampling.

zheng-da:
## Description
This PR enables multi-processing training with shared-memory graph store. It now supports multi-GPU training and NUMA training.

## Checklist
- [ ] Code is well-documented


examples/mxnet/sampling/README.md
```diff
@@ -64,3 +64,4 @@ reddit: test accuracy 96.1% with `--num-neighbors 1` and `--batch-size 1000`, ~9
 DGLBACKEND=mxnet python examples/mxnet/sampling/train.py --model graphsage_cv --batch-size 1000 --test-batch-size 5000 --n-epochs 50 --dataset reddit --num-neighbors 1 --n-hidden 128 --dropout 0.2 --weight-decay 0
 ```
 
+python3 ../incubator-mxnet/tools/launch.py -n 1 -s 1 --launcher local python3 examples/mxnet/sampling/multi_process_train.py --graph-name pubmed --model gcn_ns
```
BarclayII: What is this?

examples/mxnet/sampling/gcn_cv_sc.py
```diff
@@ -139,25 +139,38 @@ def forward(self, nf):
...
+
+    if distributed:
+        g.dist_update_all(fn.copy_src(src='features', out='m'),
+                          fn.sum(msg='m', out='preprocess'),
+                          lambda node : {'preprocess': node.data['preprocess'] * node.data['norm']})
+        for i in range(n_layers):
+            g.init_ndata('h_{}'.format(i), (features.shape[0], args.n_hidden), 'float32')
+            g.init_ndata('agg_h_{}'.format(i), (features.shape[0], args.n_hidden), 'float32')
+        g.init_ndata('h_{}'.format(n_layers-1), (features.shape[0], 2*args.n_hidden), 'float32')
+        g.init_ndata('agg_h_{}'.format(n_layers-1), (features.shape[0], 2*args.n_hidden), 'float32')
```
BarclayII: I think having a separate `dist_update_all()` API is quite redundant.  In fact, we can already see that the two branches of the `if distributed` block is essentially doing the same thing except with different paradigm (distributed vs local).

Could we at least have another class that overrides `update_all()`, like `DGLDistributedGraph.update_all()` and `DGLDistributedGraph.ndata[]` etc. instead?

examples/mxnet/sampling/graphsage_cv.py
```diff
@@ -181,25 +178,33 @@ def forward(self, nf):
...
-    for i in range(n_layers):
-        g.ndata['h_{}'.format(i)] = mx.nd.zeros((features.shape[0], args.n_hidden), ctx=ctx)
+
+    if distributed:
+        g.dist_update_all(fn.copy_src(src='features', out='m'),
+                          fn.sum(msg='m', out='preprocess'),
+                          lambda node : {'preprocess': node.data['preprocess'] * node.data['norm']})
+        for i in range(n_layers):
+            g.init_ndata('h_{}'.format(i), (features.shape[0], args.n_hidden), 'float32')
+            g.init_ndata('agg_h_{}'.format(i), (features.shape[0], args.n_hidden), 'float32')
```
BarclayII: ditto

examples/mxnet/sampling/gcn_cv_sc.py
zheng-da: we could reuse `update_all` here. However, my main concern is that if we do so, the semantics of `update_all` is changed. The main difference between `dist_update_all` and `update_all` is that `dist_update_all` requires that all trainers invoke this method and it will return only when all trainers finish their own portion of computation (there is a barrier at the end of this method). If one trainer doesn’t invoke the method, the method won’t be able to return because of the barrier. That’s why I think we should make this distributed computation explicit.

For example, the following code may not work:
```python
if ...:
    g.dist_update_all(...)
```
If one of the trainers doesn't invoke `dist_update_all`, all other trainers will be blocked. I want to distinguish this semantic from the original semantic of `update_all`.

examples/mxnet/sampling/graphsage_cv.py
zheng-da: I actually try to explain these two new APIs in the tutorial: https://hackmd.io/szg0gaArQg-DVbo1_JjTcA?both#

zheng-da:
I don't see why it's not supported in Pytorch. All the components are built in DGL. It only requires the underlying frameworks to support multi-processing training. What it lack is an example of Pytorch version. actually, the kernel fushion release doesn't have examples of mxnet version either...

examples/mxnet/sampling/gcn_cv_sc.py
BarclayII: I see.  So instead of a master process issuing an `update_all()` command to make the worker processes compute outputs of "their" nodes, we're making the workers connect to the master and call `dist_update_all()` individually.

I'm personally fine with the latter paradigm (since PyTorch is somehow using this paradigm as well - each worker computes gradients and calls `all_reduce` individually, with a sync barrier inside `all_reduce`).  The former still makes more sense to me in terms of graph-level computation though.

BarclayII:
Yeah I meant a PyTorch example.  Sorry.

examples/mxnet/sampling/gcn_cv_sc.py
zheng-da: It's interesting that you also mention issuing `update_all` in the master process. @jermainewang had the same comment. We could do that, but I don't know how it works with mini-batch training. Typically, deep learning frameworks use multiple workers and each of them executes python code to process mini-batches. If we use a mater process to issue commands, we would have to ditch this programming paradigm and adopt the one used by Spark (the master process issues commands, and worker processes split the data and execute the commands on parts of the data individually).

Both paradigms work. However, I'm not sure if it's reasonable to ditch the programming paradigm that the DL users are familiar with? Secondly, the master-worker approach doesn't work well with mini-batch training. The master process has to drive the computation in a mini-batch on worker processes. In this case, the master process is better to use task parallelism instead of data parallelism to have better performance. Even if it uses task parallelism, the performance is still an issue because the communication overhead can still be very large.

zheng-da:
I agree that we should have a Pytorch example. Let's merge this PR first and have a Pytorch example before the tutorial is released.

examples/mxnet/sampling/gcn_cv_sc.py
BarclayII: > We could do that, but I don't know how it works with mini-batch training.
Why does `update_all()` have anything to do with mini-batch training?  Doesn't `update_all()` always update the features of *all* nodes regardless?

The only issue I can think of for the master-worker approach is how the master process could send the message/reduce/apply *functions* to the workers.  For instance:
```python
distributed_graph.update_all(message_func=msg_nn, reduce_func=sum, apply_func=apply_nn)
```
where `msg_nn` and `apply_nn` are, say, PyTorch/MXNet neural networks.

I could go with the current API for now since we need this functionality.  However, if we do find a solution to the problem above in the future, I would very much prefer the `distributed_graph.update_all()` style eventually.

zheng-da: All of user code runs in trainers. This is where mini-batch is processed. Below is some pseudocode to show you how it works with the current implementation of control-variate sampling.

```python
g = create_from_graph_store(...)
# This preprocess the input features on nodes so we can reduce the number of hops to sample by 1.
g.dist_update_all(msg_func=..., reduce_func=..., update_func=...)

for nf in NeighborSampler():
    # This accumulates the embeddings of all neighbors of some nodes in the NodeFlow.
    g.pull(...)
    # We perform computation on the nodeflow to train the model.
    nf.block_compute(....)
```

If you want to use the master-worker approach, what is the master process? What is its role in the multi-processing training?

BarclayII: The master-worker approach would work like Spark.  I would imagine the master process work as follows:
```python
# DGLGraph going distributed; you can partition the graph & features over multiple storage
# servers in the argument list if you want to.
# It also specifies the number of workers needed for computation.
g = dgl.DGLDistributedGraph(storage_server='local', num_workers=20)
# Assign features to a graph.  This could replace `init_ndata` if the tensor can be held on a single
# machine.
g.ndata['h'] = ...
# Preprocess the input features; replaces dist_update_all().
# The nodes to be processed are then distributed across the workers, and each worker would
# compute the representation of its own portion.
g.update_all(msg_func=..., reduce_func=..., apply_node_func=...)

# Your minibatch trainer function
def minibatch_trainer(g: dgl.DGLDistributedGraph):
    for nf in NeighborSampler(g):
        g.pull(...)
        nf.block_compute(...)

# This is a new API that tells all the workers to run the same function.
g.run_parallel(target=minibatch_trainer, args=(g,))
```

zheng-da: What is the semantic of `pull` in `DGLDistributedGraph`? If its `update_all` means distributed update_all and there is a synchronization at the end, what about `pull`? Should it be distributed pull? If it is, what is the definition of distributed pull? Different pull has different computation (the vertices to pull are different). In addition, there will be serious performance issue because of synchronization. If it's not, I feel this is quite confusing to users.

There is also an implementation difficulty for `run_parallel`. In the case of multi-processing training, if we fork a new process, we can't run computation in the child process with openmp (this is a limitation of openmp). I'm not sure how to implement `run_parallel` given this limitation.

PR #499, opened 2019-04-17
[Sampler] Change Distributed Sampler API

aksnzhy:
## Description
This PR change distributed sampler API. Now users can use distributed sampler on trainer like this:

```python
sampler = dgl.contrib.sampling.SamplerReceiver(graph=g, addr='127.0.0.1:50051', num_sender=1)
for subg in sampler:
    # do anything here
```

## Checklist
- [x] Code is well-documented

### DIFF
--- a/examples/mxnet/sampling/dis_sampling/gcn_trainer.py
+++ b/examples/mxnet/sampling/dis_sampling/gcn_trainer.py
@@ -194,7 +194,7 @@ def main(args):
     for epoch in range(args.n_epochs):
         idx = 0
         for nf in sampler:
-            print("epoch: %d, subgraph: %d", %(epoch, idx))
+            print("epoch: %d, subgraph: %d" %(epoch, idx))
             idx += 1
             nf.copy_from_parent()
             # forward
--- a/examples/mxnet/sampling/dis_sampling/gcn_trainer.py
+++ b/examples/mxnet/sampling/dis_sampling/gcn_trainer.py
@@ -160,7 +160,7 @@ def main(args):
     norm = mx.nd.expand_dims(1./degs, 1)
     g.ndata['norm'] = norm
 
-        # Create sampler receiver
+    # Create sampler receiver
     sampler = dgl.contrib.sampling.SamplerReceiver(graph=g, addr=args.ip, num_sender=args.num_sender)
 
     model = GCNSampling(in_feats,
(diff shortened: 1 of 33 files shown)


examples/mxnet/sampling/dis_sampling/gcn_trainer.py
```diff
@@ -191,11 +191,12 @@ def main(args):
...
     # initialize graph
     dur = []
-    total_count = 153
     for epoch in range(args.n_epochs):
-        for subg_count in range(total_count):
-            print(subg_count)
-            nf = receiver.recv(g)
+        print('epoch: %d' % epoch)
+        idx = 0
+        for nf in sampler:
```
zheng-da: my main concern here is that an epoch might have a different number of mini-bathes when a distributed sampler is used. this might be confusing to people.

zheng-da: one possible solution is to maintain a receiving queue for each epoch.

src/graph/network.cc
```diff
@@ -23,6 +23,25 @@ namespace network {
 static char* SEND_BUFFER = nullptr;
 static char* RECV_BUFFER = nullptr;
 
+static void SendData(network::Sender* sender,
```
zheng-da: please add comments on what these two functions do.

src/graph/network.cc
```diff
@@ -74,20 +93,30 @@ DGL_REGISTER_GLOBAL("network._CAPI_SenderSendSubgraph")
     ImmutableGraph *ptr = static_cast<ImmutableGraph*>(ghandle);
     network::Sender* sender = static_cast<network::Sender*>(chandle);
     auto csr = ptr->GetInCSR();
+    // Write control message
+    *SEND_BUFFER = CNTROL_NODEFLOW;
```
zheng-da: do you mean `CONTROL_NODEFLOW`?

examples/mxnet/sampling/dis_sampling/gcn_trainer.py
aksnzhy: @zheng-da The main problem is that we don't know the how many subgraph will be generated in each epoch. This issue comes from the design of sampler API. If we can fix the number of subgraph in the sampler API, this problem can be solved easily. Otherwise, I don't think the receiving queue can address this problem very well, and it will also bring some other system overhead.

aksnzhy:
@zheng-da I think we use this API for now and discuss with mingjie and zz in tutorial.

PR #509, opened 2019-04-24
[BUGFIX] fix the error message in the data loader.

zheng-da:
## Description
The error message is confusing when users don't specify the dataset.

## Checklist
- [ ] Code is well-documented


VoVAllen:
I changed the required argument to True to ensure users pass this argument into program. And add some help notes.

zheng-da:
The reason that it was changed to be not required is that we should allow users to specify a graph file to load a graph. SSE is one example that users can load a graph with `--graph-file` argument.

VoVAllen:
I changed it back to required=False

PR #439, opened 2019-03-08
[Model]Implementation of diffpool and a dataset loader for TU dataset.

HQ01:
## Description

This PR implement the batched version of diffpool model, together with a re-factored simple graphsage(without sampling) with LSTM/Maxpool/Mean aggregator, and a simple dataset loader for graph dataset available in https://ls11-www.cs.tu-dortmund.de/staff/morris/graphkerneldatasets.
## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [Model], [Doc], [Feature]])
- [x] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage
- [x] Code is well-documented
- [x] To the my best knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [x] Related issue is referred in this PR

## Changes

Essentially, this PR uses DGL's batchedGraph API to implement diffpool. The graph convolution layer is handled by DGL, but computation of the pooled adjacency matrix is still done by Pytorch sparse mm. Batched Normalization is not implemented, as we batch graphs into a large graph. This implementation match the paper's accuracy on ENZYMES.

PR #517, opened 2019-05-01
Add mean pooling aggregator

sufeidechabei:
## Description
There are two pooling methods in official code (max pooling and mean pooling), I found dgl's example only has max pooling.


HQ01:
In the not-yet-finished Diffpool PR #439, LSTM pooling and max pooling are both implemented. Perhaps that could help

yzh119:
The graphsage example would be rewritten with dgl.nn.*.GraphSAGE.

lihuiknight:
In [Benchmarking Graph Neural Networks](https://arxiv.org/pdf/2003.00982.pdf), the author said the max pooling should be more powerful than the average pooling (Appendix E.2). Why this statement is right? Any ideas?

PR #521, opened 2019-05-07
[WIP][Feature] Add locking for concurrent data access to the embeddings in the shared-memory graph store

zheng-da:
## Description
The concurrent data access to the embeddings in the shared-memory graph store should be protected by locks. In this design, the embedding of each vertex has a lock to prevent from concurrent access.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/include/dgl/runtime/c_runtime_api.h
+++ b/include/dgl/runtime/c_runtime_api.h
@@ -457,6 +457,16 @@ DGL_DLL int DGLArrayCopyFromTo(DGLArrayHandle from,
                                DGLArrayHandle to,
                                DGLStreamHandle stream);
 
+DGL_DLL int DGLArraySharedMemGatherRows(DGLArrayHandle from,
+                                        DGLArrayHandle lock,
+                                        DGLArrayHandle idx,
+                                        DGLArrayHandle to);
+
+DGL_DLL int DGLArraySharedMemScatterRows(DGLArrayHandle from,
+                                         DGLArrayHandle lock,
+                                         DGLArrayHandle idx,
+                                         DGLArrayHandle to);
+
 /*!
  * \brief Produce an array from the DLManagedTensor that shares data memory
  * with the DLManagedTensor.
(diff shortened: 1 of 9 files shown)


jermainewang:
@zheng-da I see you have two PR. Which one should I review first?

futurely:
Is multiprocess/multi-gpu training affected by this?

https://github.com/dmlc/dgl/search?q=multiprocessing&unscoped_q=multiprocessing

https://github.com/dmlc/dgl/issues/838

futurely:
It is more suitable to save embeddings in the [distributed KVStore](https://github.com/dmlc/dgl/blob/6731ea3af775ee95404d99dc5a60d8d4f514f94d/python/dgl/contrib/dis_kvstore.py) than in the single machine shared-memory graph store. First, the memory requirement of the embeddings can be larger than the available memory on a single machine. Second, each class should have a single responsibility for easier development and more flexible deployment. The graph manages the input data and the KVStore manages runtime output data. 

Each KVServer only uses [a single-thread while loop](https://github.com/dmlc/dgl/blob/6731ea3af775ee95404d99dc5a60d8d4f514f94d/python/dgl/contrib/dis_kvstore.py#L338) to receive and process incoming messages. There is no race condition can happen. Using multi-processing or asynchronous programming model can improve the performance of each KVServer. Starting multiple KVServer processes and partitioning the embeddings among them can also increase the overall throughput.

zheng-da:
@futurely we'll unify shared-memory store and distributed kvstore. Right now the distributed kvstore can communicate with workers through both network and shared memory. It uses shared memory if the kvstore server and the worker run on the same machine. Shared memory has much lower communication overhead. When we partition a graph properly, we expect most of data access is from the kvstore server on the local machine.

The reason that I reopen this is because shared memory access without locking have data access conflicts and may result in data corruption during the training. I want to double check the locking mechanism in the PR. I don't intend to merge this PR.

PR #526, opened 2019-05-11
[BUGFix] Improve multi-processing training

zheng-da:
## Description
The previous multi-processing training couldn't converge at the same rate as the single-processing training. There are two reasons:
* each process generates very similar mini-batches because all processes use the same random seed to shuffle the seed vertices.
* there was a bug in dist_update_all. Previously, we construct shared-memory tensors in two steps: 1) create shared-memory tensor, 2) initialize them. It turns out there is race condition in dist_update_all. Some update_all results computed by one process are overwritten by the initialization in another process. Therefore, when we create and initialize a new column in the shared-memory graph store, we need to complete the initialization in the graph store server in one RPC call.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/contrib/graph_store.py
+++ b/python/dgl/contrib/graph_store.py
@@ -16,6 +16,7 @@ from ..graph_index import GraphIndex, create_graph_index
 from .._ffi.ndarray import empty_shared_mem
 from .._ffi.function import _init_api
 from .. import ndarray as nd
+from ..init import zero_initializer
 
 def _get_ndata_path(graph_name, ndata_name):
     return "/" + graph_name + "_node_" + ndata_name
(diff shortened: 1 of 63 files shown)


python/dgl/contrib/graph_store.py
```diff
@@ -490,7 +560,12 @@ def init_ndata(self, ndata_name, shape, dtype):
...
             are "float32" and "int32".
         """
-        self.proxy.init_ndata(ndata_name, shape, dtype)
+        init = self._node_frame.get_initializer(ndata_name)
+        if init is None:
+            self._node_frame._frame._warn_and_set_initializer()
+        init = self._node_frame.get_initializer(ndata_name)
+        init = self._init_manager.serialize(init)
+        self.proxy.init_ndata(init, ndata_name, shape, dtype)
         self._init_ndata(ndata_name, shape, dtype)
```
jermainewang: It looks like the data is initialized twice? (once remotely, once locally?)

tests/mxnet/test_shared_mem_store.py
```diff
@@ -1,61 +1,93 @@
...
+            assert np.all(arr[0].asnumpy() == i)
+
+def check_init_func(worker_id, graph_name):
     time.sleep(3)
     print("worker starts")
     np.random.seed(0)
     csr = (spsp.random(num_nodes, num_nodes, density=0.1, format='csr') != 0).astype(np.int64)
 
-    g = dgl.contrib.graph_store.create_graph_from_store("test_graph5", "shared_mem")
+    g = dgl.contrib.graph_store.create_graph_from_store(graph_name, "shared_mem", port=rand_port)
```
jermainewang: What's the type of `g` here? `GraphStore`?

python/dgl/contrib/graph_store.py
```diff
@@ -404,29 +475,28 @@ def __init__(self, graph_name, port):
...
-            arr1[:] = arr
-            return arr1
-        self._node_frame.set_remote_initializer(node_initializer)
-        self._edge_frame.set_remote_initializer(edge_initializer)
-        self._msg_frame.set_remote_initializer(edge_initializer)
+            return F.zerocopy_from_dlpack(dlpack)
+
+        self._node_frame.set_remote_init_builder(lambda init: partial(node_initializer, init))
+        self._edge_frame.set_remote_init_builder(lambda init: partial(edge_initializer, init))
+        self._msg_frame.set_remote_init_builder(lambda init: partial(edge_initializer, init))
```
jermainewang: Do we need another `set_remote_init_builder`? I feel the existing `set_initializer` is enough. The codes below are from my understanding, may not directly match your implementation.
```python
proxy = ...
node_frame = Frame(...)
def remote_init_fn(shape, dtype, ctx, name):
  proxy.init_data('zero', shape, dtype, name)
  data = empty_shared_mem(...)  # btw, the name empty here is actually confusing, the content is actually *not* empty at all.
  return data
node_frame.set_initializer(remote_init_fn)
```

python/dgl/contrib/graph_store.py
zheng-da: the first init is to initialize the data in the remote server. the second one is actually to map data in the server to the local process. there is no initialization.

tests/mxnet/test_shared_mem_store.py
zheng-da: it's a graph store.

python/dgl/contrib/graph_store.py
zheng-da: We can't only use `set_initializer`. When creating a column, we may want to create it from the shared memory. In this case, we need to get a remote initializer. However, when creating initialized data for zero-degree vertices, we only need one created in the local memory.
In addition, the signatures for the two initializers are different. the remote version needs an argument "name".

Issue #502, opened 2019-04-20
Construct BatchedDGLGraph from readonly DGLGraph


mwcvitkovic:
## 🚀 Feature

Currently readonly DGLGraphs cannot be combined into a BatchedDGLGraph.  But since a BatchedDGLGraph is itself readonly, this seems like an unnecessary blockage.

jermainewang:
You are right. That should be fixed. @VoVAllen @mufeili  do you guys have time to look into this?

jermainewang:
The feature should be supported via PR #539.

PR #539, opened 2019-05-16
[Feature] Add batch and unbatch for immutable graph

VoVAllen:
## Description

https://github.com/dmlc/dgl/issues/502
Add batch and unbatch for immutable graph

## Checklist
- [x] Code is well-documented


jermainewang:
@VoVAllen , nice work. Let's merge your change before PR #543 since I changed immutable graph quite a lot. Some general questions:
* We have two types of graph (mutable and immutable), but when converting to python object, the type is lost (both are represented by a GraphInterface*). It looks like we are likely to have separate implementations (the DisjointUnion in your change). Currently, you are using dynamic_cast to check the type. Do you think we need an explicit type enum? For example:
  ```c++
  enum GraphType {
    kMutable = 0,
    kImmutable,
  };
  class GraphInterface {
    virtual GraphType Type() const = 0;
  };
  ```
* Since we have two types of graphs, the operand type can be complicated. For example, DisjointUnion could accept a list of mixed graph objects. I saw you currently didn't handle it. But I want to discuss about the right behavior here.
  - (Type inference question) If the input list of graphs is a mix of mutable and immutable graph, what is the return type?
  - (Implementation question) What should be the right implementation for DisjointUnion of mixed graph list?

VoVAllen:
I think adding GraphType to GraphInterface is a good idea, since in many cases we need to know the type of graph. But it seems many codes would be changed for this? Might be better to implement after immutable graph refactor pr merged?

Currently I only supported batching for one type of graph, and mixed type will raise an error. 
My consideration: 
- it's rare to batch over mutable and immutable graph. User can change to consistent type in the preprocessing stage. 
- If we would like to support this, the batched graph should be immutable since it won't change. Therefore we would calculate `csr` for mutable graph in the backend. The total computation cost would be the same(or more if `csr` is not cached and conversion happened multiple times) comparing to converting at user's side, which is not ideal. 
Therefore I think it's better to let user to make the graph type consistent.

src/graph/graph_apis.cc
```diff
@@ -383,58 +434,45 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLDisjointUnion")
...
-      CHECK(gr) << "_CAPI_DGLDisjointUnion isn't implemented in immutable graph";
-      graphs.push_back(gr);
+    const GraphInterface *ptr = static_cast<const GraphInterface *>(inhandles[0]);
+    const ImmutableGraph *im_gr = dynamic_cast<const ImmutableGraph *>(ptr);
+    const Graph *gr = dynamic_cast<const Graph *>(ptr);
+    if (gr) {
+      DGLDisjointUnion<Graph>(inhandles, list_size, rv);
+    }
+    if (im_gr) {
+      DGLDisjointUnion<ImmutableGraph>(inhandles, list_size, rv);
```
jermainewang:
```c++
else {
  CHECK(im_gr) << "...";
  DGLDisjointUnion<ImmutableGraph>(inhandles, list_size, rv);
}
```

src/graph/graph_apis.cc
```diff
@@ -383,58 +434,45 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLDisjointUnion")
...
-    for (size_t i = 0; i < rst.size(); ++i) {
-      Graph* ptr = new Graph();
-      *ptr = std::move(rst[i]);
-      ptr_array_data[i] = reinterpret_cast<std::intptr_t>(ptr);
+    const ImmutableGraph* im_gptr = dynamic_cast<const ImmutableGraph*>(ptr);
+    if (gptr) {
+      DGLDisjointPartitionByNum(gptr, args, rv);
+    }
+    if (im_gptr){
+      DGLDisjointPartitionByNum(im_gptr, args, rv);
```
jermainewang: Same above

src/graph/graph_apis.cc
```diff
@@ -383,58 +434,45 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLDisjointUnion")
...
-      ptr_array_data[i] = reinterpret_cast<std::intptr_t>(ptr);
+    const ImmutableGraph* im_gptr = dynamic_cast<const ImmutableGraph*>(ptr);
+    if (gptr) {
+      DGLDisjointPartitionBySizes(gptr, sizes, rv);
     }
-    *rv = ptr_array;
-  });
+    if (im_gptr){
+      DGLDisjointPartitionBySizes(im_gptr, sizes, rv);
+    }
```
jermainewang: same above

src/graph/graph_op.cc
```diff
@@ -104,9 +106,103 @@ std::vector<Graph> GraphOp::DisjointPartitionBySizes(const Graph* graph, IdArray
...
+  dgl_id_t cum_num_edges = 0;
+  dgl_id_t indptr_idx = 1;
+  for (const ImmutableGraph *gr : graphs) {
+    const ImmutableGraph::CSR::Ptr &g_csrptr = gr->GetInCSR();
+    dgl_id_t g_num_nodes = g_csrptr->NumVertices();
+    dgl_id_t g_num_edges = g_csrptr->NumEdges();
+    ImmutableGraph::CSR::vector<dgl_id_t> &g_indices = g_csrptr->indices;
+    ImmutableGraph::CSR::vector<int64_t> &g_indptr = g_csrptr->indptr;
+    ImmutableGraph::CSR::vector<dgl_id_t> &g_edge_ids = g_csrptr->edge_ids;
+    batched_csr_ptr->indptr[0] = 0;
```
jermainewang: LN129 outside of the loop?

src/graph/graph_apis.cc
```diff
@@ -71,6 +71,57 @@ PackedFunc ConvertSubgraphToPackedFunc(const Subgraph& sg) {
 
 }  // namespace
 
+namespace {
+// This namespace contains template functions for batching
+// and unbatching over graph and immutable graph
+template<typename T>
+void DGLDisjointPartitionByNum(const T *gptr, DGLArgs args, DGLRetValue *rv) {
+  int64_t num = args[1];
+  std::vector<T> &&rst = GraphOp::DisjointPartitionByNum(gptr, num);
```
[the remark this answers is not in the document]
VoVAllen: Fixed

src/graph/graph_apis.cc
VoVAllen: Fixed. (Don't know why it isn't showed here)

src/graph/graph_apis.cc
VoVAllen: Fixed. (Don't know why it isn't showed here)

src/graph/graph_apis.cc
VoVAllen: Fixed. (Don't know why it isn't showed here)

src/graph/graph_op.cc
VoVAllen: Fixed.

PR #543, opened 2019-05-17
[Refactor] Immutable graph index

jermainewang:
## Description

Refactor the immutable graph class. Now it contains three cached graph formats (in_csr, out_csr, coo). All graph formats use runtime::NDArray as storage.

@zheng-da @ylfdq1118 @BarclayII please review.

## Checklist
- [x] Code is well-documented


include/dgl/immutable_graph.h
```diff
@@ -489,39 +603,51 @@ class ImmutableGraph: public GraphInterface {
...
   /*!
    * \brief Get the in edges of the vertex.
    * \note The returned dst id array is filled with vid.
    * \param vid The vertex id.
    * \return the edges
    */
-  EdgeArray InEdges(dgl_id_t vid) const {
-    return this->GetInCSR()->GetEdges(vid);
+  EdgeArray InEdges(dgl_id_t vid) const override {
+    const EdgeArray& ret = GetInCSR()->OutEdges(vid);
```
zheng-da: a reference to a return value. This is very dangerous.

include/dgl/immutable_graph.h
```diff
@@ -489,39 +603,51 @@ class ImmutableGraph: public GraphInterface {
...
 
   /*!
    * \brief Get the in edges of the vertices.
    * \param vids The vertex id array.
    * \return the id arrays of the two endpoints of the edges.
    */
-  EdgeArray InEdges(IdArray vids) const {
-    return this->GetInCSR()->GetEdges(vids);
+  EdgeArray InEdges(IdArray vids) const override {
+    const EdgeArray& ret = GetInCSR()->OutEdges(vids);
```
zheng-da: the same here.

include/dgl/immutable_graph.h
```diff
@@ -699,95 +820,78 @@ class ImmutableGraph: public GraphInterface {
...
-      return in_csr_;
-    } else {
-      CHECK(out_csr_ != nullptr) << "one of the CSRs must exist";
-      const_cast<ImmutableGraph *>(this)->in_csr_ = out_csr_->Transpose();
-      return in_csr_;
+  /* !\brief Return in csr. If not exist, transpose the other one.*/
+  CSRPtr GetInCSR() const {
+    if (!in_csr_) {
+      if (out_csr_) {
+        in_csr_ = out_csr_->Transpose();
```
zheng-da: is transpose from out csr more efficient than conversion from coo directly?

include/dgl/immutable_graph.h
```diff
@@ -699,95 +820,78 @@ class ImmutableGraph: public GraphInterface {
...
-    } else {
-      CHECK(in_csr_ != nullptr) << "one of the CSRs must exist";
-      const_cast<ImmutableGraph *>(this)->out_csr_ = in_csr_->Transpose();
-      return out_csr_;
+
+  /* !\brief Return out csr. If not exist, transpose the other one.*/
+  CSRPtr GetOutCSR() const {
+    if (!out_csr_) {
+      if (in_csr_) {
+        out_csr_ = in_csr_->Transpose();
```
zheng-da: the same comment here.

python/dgl/graph_index.py
```diff
@@ -51,21 +51,19 @@ def __setstate__(self, state):
...
-            self._init(src, dst, utils.toindex(F.arange(0, len(src))), n_nodes)
+            self._init(src, dst, n_nodes)
         else:
             self._handle = _CAPI_DGLGraphCreateMutable(multigraph)
             self.clear()
             self.add_nodes(n_nodes)
             self.add_edges(src, dst)
 
-    def _init(self, src_ids, dst_ids, edge_ids, num_nodes):
+    def _init(self, src_ids, dst_ids, num_nodes):
```
zheng-da: is it possible that users may provide their own edge ids?

src/array.cc
```diff
@@ -0,0 +1,160 @@
...
+    ret_data[i + L] = arr2_data[i];
+  }
+  return ret;
+}
+
+CSRMatrix SliceRows(const CSRMatrix& csr, int64_t start, int64_t end) {
+  // TODO(minjie): TBD
+  LOG(FATAL) << "Not implemented.";
+  return {};
+}
```
zheng-da: will this be implemented?

include/dgl/immutable_graph.h
```diff
@@ -17,317 +17,444 @@
...
 
-  // Edge list indexed by edge id;
-  struct EdgeList {
-    typedef std::shared_ptr<EdgeList> Ptr;
-    std::vector<dgl_id_t> src_points;
-    std::vector<dgl_id_t> dst_points;
+class COO : public GraphInterface {
+ public:
+  // Create a coo graph that shares the given src and dst
+  COO(int64_t num_vertices, IdArray src, IdArray dst);
```
zheng-da: this should also support shared memory. please add a TODO here for me.

src/graph/immutable_graph.cc
```diff
@@ -4,150 +4,179 @@
...
-  return (!(first == last) && !(value < *first));
-}
+  // Return true if the given id is contained in this hashmap.
+  bool Contains(dgl_id_t id) const {
+    return filter_.test(id & kFilterMask) && oldv2newv_.count(id);
+  }
+
+  // Return the new id of the given id.
+  dgl_id_t Map(dgl_id_t id) const {
+    return oldv2newv_.at(id);
```
zheng-da: this implementation seems to require much more hashtable lookups. it can potentially significantly slow down sampling.

src/graph/immutable_graph.cc
```diff
@@ -4,150 +4,179 @@
...
+
+  IdArray sm_array = IdArray::EmptyShared(
+      shared_mem_name, {file_size}, DLDataType{kDLInt, 8, 1}, DLContext{kDLCPU, 0}, true);
+  // Create views from the shared memory array. Note that we don't need to save
+  //   the sm_array because the refcount is maintained by the view arrays.
+  indptr_ = sm_array.CreateView({num_verts + 1}, DLDataType{kDLInt, 64, 1});
+  indices_ = sm_array.CreateView({num_edges}, DLDataType{kDLInt, 64, 1},
+      (num_verts + 1) * sizeof(dgl_id_t));
+  edge_ids_ = sm_array.CreateView({num_edges}, DLDataType{kDLInt, 64, 1},
+      (num_verts + 1 + num_edges) * sizeof(dgl_id_t));
```
zheng-da: now all arrays have a reference to the shared-memory object?

src/graph/immutable_graph.cc
```diff
@@ -156,392 +185,55 @@ DegreeArray ImmutableGraph::CSR::GetDegrees(IdArray vids) const {
...
+      ids.push_back(eid_data[i]);
+    }
   }
-
-  return rst;
+  return VecToIdArray(ids);
 }
 
-ImmutableGraph::EdgeArray ImmutableGraph::EdgeIds(IdArray src_ids, IdArray dst_ids) const {
+CSR::EdgeArray CSR::EdgeIds(IdArray src_ids, IdArray dst_ids) const {
```
zheng-da: it seems this operation is heavily used by the computation API, for example `send_and_recv`. I think it's important to optimize its performance. because the edges in the neighborhood of a vertex aren't sorted, we need linearly search in the neighborhood. it can potentially be much slower than the previous implementation.

src/graph/immutable_graph.cc
```diff
@@ -555,187 +247,332 @@ ImmutableGraph::EdgeArray ImmutableGraph::EdgeIds(IdArray src_ids, IdArray dst_i
...
-  std::copy(eid.begin(), eid.end(), rst_eid_data);
+  // If sorted, the returned edges are sorted by the source Id and dest Id.
+  for (dgl_id_t src = 0; src < NumVertices(); ++src) {
+    std::fill(rst_src_data + indptr_data[src],
+              rst_src_data + indptr_data[src + 1],
+              src);
+  }
+
+  return CSR::EdgeArray{rst_src, indices_, edge_ids_};
+}
```
zheng-da: bookmark. i'll continue review later.

include/dgl/immutable_graph.h
jermainewang: This is totally fine for handling readonly temporary object.

include/dgl/immutable_graph.h
jermainewang: I think they are similar in terms of complexity. But for cache locality, I'm not sure.

python/dgl/graph_index.py
jermainewang: I don't see any usage in DGL. For example, when converting nx graph to DGL graph, we will first figure out the right edge id order and pack the edge features into one tensor by that order so they are consistent.

src/array.cc
jermainewang: I'm on it. It is used in nodeflow.cc but somehow not covered in unittest.

src/graph/immutable_graph.cc
jermainewang: https://github.com/dmlc/dgl/blob/a7d0b8a0a416c0a253215e714ce274bb9dd88518/src/graph/immutable_graph.cc#L308-L309

So you think it is better to change it to sth like?
```c++
newv = hashMap.ContainsAndMap(...);
if (IsValid(newv)) {
  // do something.
}
```

src/graph/immutable_graph.cc
jermainewang: In `NDArray::CreateView`, it increases the reference count of the base array (the `sm_array` in this context):
https://github.com/dmlc/dgl/blob/a7d0b8a0a416c0a253215e714ce274bb9dd88518/src/runtime/ndarray.cc#L136-L154

As a result, `sm_array` will not be freed untill all indices_, indptr_ and edge_ids_ are freed. Since `sm_array` holds a shared pointer to the shared memory object. That means we don't need to have that explicitly anymore.

src/graph/immutable_graph.cc
```diff
@@ -4,150 +4,179 @@
...
     CHECK(HasVertex(vid)) << "Invalid vertex: " << vid;
-    rstlen += this->GetDegree(vid);
+    rstlen += OutDegree(vid);
   }
-  IdArray src = IdArray::Empty({rstlen}, vids->dtype, vids->ctx);
-  IdArray dst = IdArray::Empty({rstlen}, vids->dtype, vids->ctx);
-  IdArray eid = IdArray::Empty({rstlen}, vids->dtype, vids->ctx);
+  IdArray src = NewIdArray(rstlen);
+  IdArray dst = NewIdArray(rstlen);
+  IdArray eid = NewIdArray(rstlen);
```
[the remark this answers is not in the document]
jermainewang: Yup.

src/graph/immutable_graph.cc
```diff
@@ -156,392 +185,55 @@ DegreeArray ImmutableGraph::CSR::GetDegrees(IdArray vids) const {
...
 
-IdArray ImmutableGraph::Predecessors(dgl_id_t vid, uint64_t radius) const {
+IdArray CSR::Successors(dgl_id_t vid, uint64_t radius) const {
   CHECK(HasVertex(vid)) << "invalid vertex: " << vid;
-  CHECK(radius >= 1) << "invalid radius: " << radius;
-
-  auto pred = this->GetInCSR()->GetIndexRef(vid);
-  const int64_t len = pred.size();
-  IdArray rst = IdArray::Empty({len}, DLDataType{kDLInt, 64, 1}, DLContext{kDLCPU, 0});
+  CHECK(radius == 1) << "invalid radius: " << radius;
```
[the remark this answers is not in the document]
jermainewang: I think we don't have implementation for radius >=2 right now?

include/dgl/immutable_graph.h
zheng-da: I see. I didn't know that.
https://stackoverflow.com/questions/2822243/store-return-value-of-function-in-reference-c

src/graph/immutable_graph.cc
zheng-da: yes, something like this.

src/graph/immutable_graph.cc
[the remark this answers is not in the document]
zheng-da: oh, i misunderstood the check.

include/dgl/immutable_graph.h
zheng-da: is this allowed from the very beginning of C++ specification? This semantic looks really weird to me.

src/graph/immutable_graph.cc
zheng-da: it seems it's not used only for node_subgraph and edge_subgraph instead of neighbor sampling. so far, these two operations aren't commonly used. i guess it's fine for now if you think it's not necessary to change it.

include/dgl/immutable_graph.h
zheng-da: i see how csr transpose and coo=>csr works now. in the worst case, both should have the same cache locality. However, coo=>csr might work better in practice because coo is already sorted. i'm not sure either if this is a common case.

tests/mxnet/test_specialization.py
```diff
@@ -227,7 +227,7 @@ def message_func_edge(edges):
     def reduce_func(nodes):
         return {'v2': mx.nd.sum(nodes.mailbox['m2'], axis=1)}
 
-    g = generate_graph(100)
+    g = generate_graph(10)
```
zheng-da: original size is too large?

jermainewang: Oh, I change it to debug. I could change it back if you want.

src/graph/immutable_graph.cc
jermainewang: Yeah. Do you have any benchmark I could refer to? As we've discussed before, I could implement the `sorted_indices` if that's really needed.

src/graph/immutable_graph.cc
jermainewang: I could change that. It's easy.

include/dgl/immutable_graph.h
jermainewang: coo is not sorted in src order but in eid order, so one could expect the locality is pretty bad

src/graph/immutable_graph.cc
zheng-da: I don't unfortunately. But we could create a relatively large graph and run `send_and_recv` to benchmark it.

tests/mxnet/test_specialization.py
zheng-da: yes, let's change it back.

BarclayII:
I noticed that we are creating NDArrays with `NewXXXArray` helper functions.  Should we do the same for `graph.cc` and others in this PR?

jermainewang:
> I noticed that we are creating NDArrays with `NewXXXArray` helper functions. Should we do the same for `graph.cc` and others in this PR?

Let's do it later and keep this PR slim.

src/graph/graph_apis.cc
```diff
@@ -82,33 +82,42 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLGraphCreateMutable")
...
     GraphHandle ghandle;
-    if (readonly)
-      ghandle = new ImmutableGraph(src_ids, dst_ids, edge_ids, num_nodes, multigraph);
-    else
-      ghandle = new Graph(src_ids, dst_ids, edge_ids, num_nodes, multigraph);
+    if (readonly) {
+      // TODO(minjie): The array copy here is unnecessary and adds extra overhead.
+      //   However, with MXNet backend, the memory would be corrupted if we directly
+      //   save the passed-in ndarrays into DGL's graph object. We hope MXNet team
+      //   could help look into this.
```
jermainewang: @zheng-da here is an issue we found for MX backend. The issue occurs when we directly borrow the passed-in ndarrays into CSR graph. This is very efficient because there is no memory copy. However, with MX backend, the program will crash due to memory double-free. We ends up with a sad workaround using memory copy. We cannot have a clean reproducible example. I hope you could connect @ylfdq1118 with who is in charge of DLPack in MX team and he could show you step-by-step on how to reproduce it in DGL.

zheng-da: The DLPack in MXNet is implemented by https://github.com/wkcn. He isn't in the MX team.

src/array.cc
jermainewang: done

include/dgl/immutable_graph.h
jermainewang: done

tests/mxnet/test_specialization.py
jermainewang: done

src/graph/immutable_graph.cc
jermainewang: Done. Please see whether it looks good to you.

jermainewang:
@BarclayII , the WIN CI fails with a weird `Stackoverflow error` for the `tests/compute/test_graph_index.test_node_subgraph` test case. I currently disable the test and everything else works fine. Could you have a look? Is it related to the WIN VM configuration?

include/dgl/immutable_graph.h
```diff
@@ -17,317 +17,444 @@
...
-      void init(T *arr, size_t size) {
-        CHECK(this->arr == nullptr);
-        this->arr = arr;
-        this->capacity = size;
-        this->curr = 0;
-        this->own = false;
-      }
+  bool IsMultigraph() const override {
+    return false;
+  }
```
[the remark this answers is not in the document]
jermainewang: How about this. I will change the member to `Lazy<bool> is_multigraph_`. User has an option to not specify whether the graph is a multigraph or not during the construction. When `IsMultigraph()` is called, it will try `Get` the value by a scan of the full graph. Since the graph is immutable, the scan should only be done once. (in fact, for mutable graph, we probably should also do this as a cache).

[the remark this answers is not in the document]
zheng-da: It seems IsMultigraph() isn't used. People only call ImmutableGraph::IsMultigraph(). I guess it doesn't really matter that much what value is returned from IsMultigraph().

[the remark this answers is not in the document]
jermainewang: Immutable graph should return `AnyGraph()->IsMultigraph()` technically. My current implementation is not correct.

BarclayII:
I think it's a Windows-specific issue since it's also occurring on my laptop.

BarclayII:
Seems that IdHashMap is taking too much stack space (2M).  Can we make the filter size smaller or move the bitset to the heap?

include/dgl/immutable_graph.h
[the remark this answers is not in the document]
jermainewang: Done. See if works for you.

jermainewang:
> Seems that IdHashMap is taking too much stack space (2M). Can we make the filter size smaller or move the bitset to the heap?

Done.

src/graph/immutable_graph.cc
jermainewang: I put down a TODO note here. Let's do that later?

zheng-da: sure

src/graph/immutable_graph.cc
```diff
@@ -80,51 +110,70 @@ CSR::CSR(IdArray indptr, IdArray indices, IdArray edge_ids)
...
+          const dgl_id_t dst = indices_data[eid];
+          if (hashmap.count(dst)) {
+            return true;
+          } else {
+            hashmap.insert(dst);
+          }
+        }
+      }
+      return false;
+    });
```
zheng-da: please add comments here. `is_multigraph_` is initialized with a value. Is the lambda always called regardless of the initial value of `is_multigraph_`?

include/dgl/immutable_graph.h
```diff
@@ -454,9 +459,7 @@ class COO : public GraphInterface {
 class ImmutableGraph: public GraphInterface {
  public:
   /*! \brief Construct an immutable graph from the COO format. */
-  explicit ImmutableGraph(COOPtr coo, bool multigraph = false)
-    : coo_(coo), is_multigraph_(multigraph) {
-  }
+  explicit ImmutableGraph(COOPtr coo, bool multigraph = false): coo_(coo) { }
```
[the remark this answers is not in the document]
jermainewang: done

src/graph/graph_op.cc
```diff
@@ -142,10 +143,10 @@ ImmutableGraph GraphOp::DisjointUnion(std::vector<const ImmutableGraph *> graphs
     }
     cum_num_nodes += g_num_nodes;
     cum_num_edges += g_num_edges;
-    is_multigraph |= gr->IsMultigraph();
   }
 
-  return ImmutableGraph(batched_csr_ptr, nullptr, is_multigraph);
+  CSRPtr batched_csr_ptr = CSRPtr(new CSR(indptr_arr, indices_arr, edge_ids_arr));
+  return ImmutableGraph(batched_csr_ptr, nullptr);
```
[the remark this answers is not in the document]
jermainewang: If any of the graph in the given list is a multigraph, the result batched graph is a multi-graph. However, I don't want to explicitly call `IsMultigraph` for each input graph (could be costly). So I set the result graph flag to be unknown.

src/graph/immutable_graph.cc
jermainewang: done

PR #545, opened 2019-05-19
[Feature][RFC] Heterogeneous graph interface

BarclayII:
## Description
The Python interface proposal for heterogeneous graphs (heterograph for short - although [the word heterograph means something unrelated](https://study.com/academy/lesson/heterographs-definition-examples-quiz.html)).

* A heterograph in DGL additionally requires that the edges with the same type MUST have the same source node type and destination node type.
* Users can only assign features and perform computation on one node/edge type at a time.
* Users can induce subgraphs from heterograph with node/edge types.
* Different node/edge types may have different node/edge feature schemes.

Please feel free to add other features you wish to have or make comments.


python/dgl/heterograph.py
```diff
@@ -0,0 +1,1214 @@
...
+    One can construct the graph as follows:
+
+    >>> import networkx as nx
+    >>> metagraph = nx.MultiGraph([
+    ...     ('user', 'user', 'follows'),
+    ...     ('user', 'game', 'plays'),
+    ...     ('developer', 'game', 'develops')])
+    >>> g = DGLBaseHeteroGraph(
+    ...     metagraph=metagraph,
+    ...     node_types=['user'] * 4 + ['game'] * 2 + ['developer'] * 2,
```
VoVAllen: What if I have 1M users? Would it be `['user']*1,000,000`?

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1214 @@
...
+            The edge type name.  Must exist in the metagraph.
+        data : dict, optional
+            Feature data of the added edge.
+
+        Examples
+        --------
+        The variable ``g`` is constructed from the example in
+        DGLBaseHeteroGraph.
+
+        >>> g['user', 'game', 'plays'].number_of_edges()
```
VoVAllen: Is it possible to directly call `g['plays']`(assume no semantic ambiguity)? `g['user', 'game', 'plays']` seems too long?

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1214 @@
...
+        4
+        >>> g.add_edge(2, 0, 'user', 'game', 'plays')
```
VoVAllen: How about `g['user', 'game', 'plays'].add_edges(2, 0)`?

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1214 @@
...
+        Developer name  Nintendo    Mojang
+        Developer ID    0           1
+
+    One can construct the graph as follows:
+
+    >>> import networkx as nx
+    >>> metagraph = nx.MultiGraph([
+    ...     ('user', 'user', 'follows'),
+    ...     ('user', 'game', 'plays'),
+    ...     ('developer', 'game', 'develops')])
```
zheng-da: Is this networkx syntax? the last element in the tuple represents edge type?

python/dgl/heterograph.py
zheng-da: after avoiding global node id, i thought we can avoid defining node types for each node?

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1214 @@
...
+    # pylint: disable=unused-argument
+    def __init__(
+            self,
+            metagraph,
+            node_types,
+            edge_connections_by_type):
+        super(DGLBaseHeteroGraph, self).__init__()
+
+    # TODO: should this return DGLBaseHeteroGraph or something like
+    # DGLBaseHeteroSubgraph?
```
zheng-da: i think it should be a subgraph type. otherwise, it's weird that the updates on the node data of this graph also reflect on the nodes of the parent graph.

zheng-da: BTW, if the heterogeneous graph is mutable graph, getting a subgraph is going to be expensive. nodes of the same type are no longer assigned to a single Id range. slicing requires id remapping.

python/dgl/heterograph.py
zheng-da: multiple edge types can have the same edge type name

python/dgl/heterograph.py
zheng-da: we probably don't allow mutating a subgraph.

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1214 @@
...
+        single node type.
+
+        Notes
+        -----
+        An error is raised if the graph contains multiple node types.  Use
+
+            g[ntype]
+
+        to select nodes with type ``ntype``.
+        """
```
zheng-da: can we add a node type argument here? if `g.nodes` doesn't work for heterogeneous graphs,  `g['user', 'item'].nodes` doesn't work either. the only way that works is `g['user'].nodes`?

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1214 @@
...
+        to select nodes with type ``ntype``.
+
+        Parameters
+        ----------
+        func : callable or None
+            Apply function on the nodes. The function should be
+            a :mod:`Node UDF <dgl.udf>`.
+        v : int, iterable of int, tensor, optional
+            The (type-specific) node (ids) on which to apply ``func``. The
+            default value is all the nodes (with the same type).
```
zheng-da: v can't be string?

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1214 @@
...
+        Consecutive sends on different edge types will also raise an error.
+
+        Parameters
+        ----------
+        edges : valid edges type, optional
+            Edges on which to apply ``message_func``. Default is sending along all
+            the edges.
+        message_func : callable
+            Message function on the edges. The function should be
+            an :mod:`Edge UDF <dgl.udf>`.
```
zheng-da: should we support list of message functions?

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1214 @@
...
+        pass
+
+    def edge_subgraph(self, edges):
+        """Return the subgraph induced on given edges.
+
+        Parameters
+        ----------
+        edges : dict[etype, list or iterable]
+            A dictionary of edge types to edge ID array to construct
+            subgraph.
```
zheng-da: edge types are strings or tuples of strings?

python/dgl/heterograph.py
BarclayII: Yes

python/dgl/heterograph.py
BarclayII: Instead of `g.apply_nodes('nodetype', func)`, one does `g['nodetype'].apply_nodes(func)`

python/dgl/heterograph.py
BarclayII: More accurately, edge types could share the same names *while* connecting to different node types.  For instance:
```
Piano <---(Plays)--- User ---(Plays)---> Game
```
The problem is: do we really want to allow different edge types having the same name?  Or should we force the users to have distinct names for each edge type (like, **force** the users to rename `Plays` to `Plays-piano` and `Plays-game`)?

python/dgl/heterograph.py
BarclayII: Re subgraph: adopted
Re mutable: In the case of mutable heterogeneous graph, I guess we have to make a `node_type x number_of_nodes x number_of_successors` adjacency list.  For now we can focus on implementing this interface for immutable graphs only since that's more important.

python/dgl/heterograph.py
BarclayII: Only `g['user'].nodes` is allowed; `g['user', 'item'].nodes` would throw an error.

python/dgl/heterograph.py
BarclayII: You mean a list of message functions for different edge types?

python/dgl/heterograph.py
BarclayII: Tuple of strings (see L946)

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1213 @@
...
+    ...         # Nintendo develops Tetris and Mojang develops Minecraft
+    ...         ('developer', 'game', 'develops'): ([0, 1], [0, 1])})
+    """
+
+    # pylint: disable=unused-argument
+    def __init__(
+            self,
+            metagraph,
+            number_of_nodes_by_type,
+            edge_connections_by_type):
```
jermainewang: Could we support conversion from other graph formats? For example, each relation subgraphs could be stored in scipy sparse matrix.

python/dgl/heterograph.py
jermainewang: Another problem is `g['plays']` returns edge subgraph while `g['user']` returns node induced subgraph. This makes the type system inconsistent.

python/dgl/heterograph.py
jermainewang: Wait, `g['user', 'item'].nodes['user']` is allowed? No?

python/dgl/heterograph.py
zheng-da: yes

python/dgl/heterograph.py
zheng-da: i wonder if it's too expensive to always call `g['nodetype'].apply_nodes`. after all, it constructs a subgraph for just updating nodes.

python/dgl/heterograph.py
zheng-da: what is `node_type x number_of_nodes x number_of_successors` adjacency list?

python/dgl/heterograph.py
zheng-da: given yesterday's discussion, i think we agree to adopt the following syntax:
```python
g[‘user’, ‘item’].send_and_recv(‘purchase’, fn...)
g[‘user’, ‘item’, ‘purchase’].pull(‘item’, fn...)
g.apply_nodes(‘item’, fn...)
```

python/dgl/heterograph.py
BarclayII: No.  `nodes[]` takes in IDs.

python/dgl/heterograph.py
BarclayII: `std::vector_by_type< std::vector_by_id< std::pair<type, id>>>`

python/dgl/heterograph.py
BarclayII: Can we make `g['nodetype']` lazy?  Like, not actually constructing the subgraph unless the user wants to query the graph topology itself?

Besides, what if the users want to do something like `g['item'].pull([1, 3, 5])`?

python/dgl/heterograph.py
BarclayII: Did we agree on having a dictionary of message/reduce functions last time?  I thought we said that the dictionary would look ugly or something.

python/dgl/heterograph.py
zheng-da: I think people can still do `pull([1, 3, 5])`. we can use the input type to determine the behavior.

python/dgl/heterograph.py
VoVAllen: > More accurately, edge types could share the same names _while_ connecting to different node types. For instance:
> 
> ```
> Piano <---(Plays)--- User ---(Plays)---> Game
> ```
> 
> The problem is: do we really want to allow different edge types having the same name? Or should we force the users to have distinct names for each edge type (like, **force** the users to rename `Plays` to `Plays-piano` and `Plays-game`)?

In this case, forcing unique name can make the code more concise and I'm for it. My concern is something like:
```
Piano <---(have)--- User ---(have)---> Game
```
Will user have this kind of relation and want to do the computation on all `have` edges?

python/dgl/heterograph.py
zheng-da: I thought we agreed that we'll allow both. Having a dict of message functions allows us to perform all computation in one API, which has two advantages:
* potentially have better optimizations
* avoid storing many temporary node data in frames.

python/dgl/heterograph.py
VoVAllen: We can teach user to write `user_g=g['user']; user_g.apply_nodes(func)`. Another advantage of this is if `user_g` is written wrong(such as `user_gg`), it can be detected at compilation stage (or by smart IDE).

python/dgl/heterograph.py
zheng-da: i see. I guess having the data structure won't reduce the overhead of constructing a subgraph. This is my concern. Extensively using the subgraph API for computation might make the computation very slow.

python/dgl/heterograph.py
BarclayII: > I think people can still do `pull([1, 3, 5])`. we can use the input type to determine the behavior.

No.  Users have to write
```python
g.pull('item', [1, 3, 5])
```
instead if we put node type inside the argument list.  The point is that the node list argument cannot be omitted in any case.

python/dgl/heterograph.py
BarclayII: Then what if I do something like this:
```python
g.update_all(
    message_func={
        ('A', 'B', 'AB'): lambda edges: {'x': funcAB(edges)},
        ('C', 'B', 'CB'): lambda edges: {'y': funcCB(edges)}},
    reduce_func={'B': lambda nodes: {'m': funcB(nodes)}})
```
Note that the messages to B nodes do not align.

Typed mailboxes can solve the issue though (i.e. `nodes.typed_mailbox[('A', 'B', 'AB')]['x']` and `nodes.typed_mailbox[('C', 'B', 'CB')]['y']`), so that messages are batched by edge type.  What do you think?

python/dgl/heterograph.py
zheng-da: what does a user do if he wants to pull on one type of nodes in a bipartite graph of 'user' and 'item'? following your syntax, i guess a user needs to do
```
g.pull('item', ALL, ...)
```

python/dgl/heterograph.py
BarclayII: For computation like `g['user'].update_all(...)` I think we can translate the call to `g.update_all('user', ...)` *without* materializing the subgraph (i.e. no remapping stuff).  I updated the return type to `DGLBaseHeteroGraphView` to demonstrate my idea.

python/dgl/heterograph.py
zheng-da: should the received messages be stored in `mailbox['x']` and `mailbox['y']`?

python/dgl/heterograph.py
zheng-da: does it mean DGLHeteroGraph also supports `update_all('user', ...)`?

BarclayII: No.  For `update_all()` the graph must have one node type and one edge type (see docstring).

zheng-da: do we allow `update_all` on a bipartite graph?

python/dgl/heterograph.py
BarclayII: > what does a user do if he wants to pull on one type of nodes in a bipartite graph of 'user' and 'item'? following your syntax, i guess a user needs to do
> 
> ```
> g.pull('item', ALL, ...)
> ```
> 
> if so, we still need to extend the API to accept a node type in pull.

In this case they need to write `g['user', 'item', 'purchase'].send_and_recv(...)` instead.

python/dgl/heterograph.py
BarclayII: The problem is that AB and CB could technically return different messages with the same name `x` - even with different shapes.

python/dgl/heterograph.py
BarclayII: > My concern is something like:
> 
> ```
> Piano <---(have)--- User ---(have)---> Game
> ```
> 
> Will user have this kind of relation and want to do the computation on all have edges?

We assume not; they need to compute one edge type at a time (the idea itself is debatable; see discussion in L659)

EDIT: I accidentally replied by editing the previous comment...

python/dgl/heterograph.py
BarclayII: I believe `g['user', 'item', 'purchase'].send_and_recv(ALL)` should suffice for at least GCMC-like computations: this computes the message from each user to each of his/her purchased item, and the items aggregate the messages from the users and update themselves.

python/dgl/heterograph.py
zheng-da: so `pull` can't be used in a bipartite graph? there will be a performance issue. in general, `update_all` is much more efficient than `pull`, `pull` is much more efficient than `send_and_recv`. I think we should allow all of these API at a bipartite graph for the sake of efficiency.

also, what confuses me is that sometimes you say the following is supported.
```python
g.pull('item', [1, 3, 5])
```
sometimes, you say the API doesn't accept node type as input. i'm not sure what is supported and what is not supported.

python/dgl/heterograph.py
zheng-da: i see. this is problematic. we could simply report an error.

in general, i think we should define large APIs. This principle is used from the very beginning. That's why we have multiple levels of API. We believe large APIs give us more opportunities for optimization in the future.

python/dgl/heterograph.py
BarclayII: Makes sense.  I'll add that.

python/dgl/heterograph.py
BarclayII: > i see. this is problematic. we could simply report an error.

Why is `typed_mailbox` bad?  Is it simply because we have another API?

BarclayII:
As per @zheng-da 's request I allowed specifying a set of type-specific message/reduce/apply functions as dicts in message passing APIs.  The docstrings become quite complicated as a result.  Please take a look and see if examples are needed.

python/dgl/heterograph.py
zheng-da: I'm OK with `typed_mailbox`. I'm a little worried that it may make the programming in reduce very complex. For example, in RGCN, neighbor embeddings of different node types are projected into the same space, so we can do simple reduction in the reduce function. If we use `typed_mailbox`, i'm not sure how to write a simple reduce function for this case. If a model really wants to send messages with different dimensions to the mailbox, it has the option of using a different name for each node type if they want.
We might want to investigate what is the common case. Do most of the models send messages of the same dimension to the destination nodes?

zheng-da:
could you please rebase to the master branch? Somehow github messed up the diff.

BarclayII:
I'll reopen one - don't know how to fix the diff.

PR #553, opened 2019-05-23
[Feature][RFC] Heterogeneous graph interface

BarclayII:
## Description
The Python interface proposal for heterogeneous graphs (heterograph for short - although [the word heterograph means something unrelated](https://study.com/academy/lesson/heterographs-definition-examples-quiz.html)).

* A heterograph in DGL additionally requires that the edges with the same type MUST have the same source node type and destination node type.
* Users can only assign features on one node/edge type at a time.
* Users can induce subgraphs from heterograph with node/edge types.
* Different node/edge types may have different node/edge feature schemes.
* Message functions and edge apply functions can be dictionary of edge types and edge UDFs.
* Reduce functions can be dictionary of edge types and node UDFs.
* Node apply functions can be dictionary of node types and node UDFs.

Please feel free to add other features you wish to have or make comments.

I accidentally merged from master and messed up the diff so I'm reopening a new PR (and closed the old one)

### DIFF
--- a/python/dgl/heterograph.py
+++ b/python/dgl/heterograph.py
@@ -372,7 +372,7 @@ class DGLBaseHeteroGraph(object):
         Parameters
         ----------
         v : int
-            The node.
+            The node of destination type.
 
         Returns
         -------
@@ -412,7 +412,7 @@ class DGLBaseHeteroGraph(object):
         Parameters
         ----------
         v : int
-            The node.
+            The node of source type.
 
         Returns
         -------
@@ -572,7 +572,7 @@ class DGLBaseHeteroGraph(object):
         Parameters
         ----------
         v : int, list, tensor
-            The node(s).
+            The node(s) of destination type.
         form : str, optional
             The return form. Currently support:
 
@@ -614,7 +614,7 @@ class DGLBaseHeteroGraph(object):
         Parameters
         ----------
         v : int, list, tensor
-            The node(s).
+            The node(s) of source type.
         form : str, optional
             The return form. Currently support:
 
@@ -703,7 +703,7 @@ class DGLBaseHeteroGraph(object):
         Parameters
         ----------
         v : int
-            The node ID.
+            The node ID of destination type.
 
         Returns
         -------
@@ -737,7 +737,8 @@ class DGLBaseHeteroGraph(object):
         Parameters
         ----------
         v : list, tensor, optional.
-            The node ID array. Default is to return the degrees of all the nodes.
+            The node ID array of destination type. Default is to return the
+            degrees of all the nodes.
 
         Returns
         -------
@@ -771,7 +772,7 @@ class DGLBaseHeteroGraph(object):
         Parameters
         ----------
         v : int
-            The node ID.
+            The node ID of source type.
 
         Returns
         -------
@@ -805,7 +806,8 @@ class DGLBaseHeteroGraph(object):
         Parameters
         ----------
         v : list, tensor
-            The node ID array. Default is to return the degrees of all the nodes.
+            The node ID array of source type. Default is to return the degrees
+            of all the nodes.
 
         Returns
         -------
--- a/python/dgl/heterograph.py
+++ b/python/dgl/heterograph.py
@@ -18,8 +18,8 @@ class DGLBaseHeteroGraph(object):
         metagraph.
         The edge types are specified as edge keys on the NetworkX MultiGraph.
         The node types and edge types must be strings.
-    node_types : list
-        The node types.
+    number_of_nodes_by_type : dict[str, int]
+        Number of nodes for each node type.
     edge_connections_by_type : dict
         Specifies how edges would connect nodes of the source type to nodes of
         the destination type in the following form:
@@ -86,7 +86,7 @@ class DGLBaseHeteroGraph(object):
     ...     ('developer', 'game', 'develops')])
     >>> g = DGLBaseHeteroGraph(
     ...     metagraph=metagraph,
-    ...     node_types=['user'] * 4 + ['game'] * 2 + ['developer'] * 2,
+    ...     number_of_nodes_by_type={'user': 4, 'game': 2, 'developer': 2},
     ...     edge_connections_by_type={
     ...         # Alice follows Bob and Bob follows Carol
     ...         ('user', 'user', 'follows'): ([0, 1], [1, 2]),
@@ -100,7 +100,7 @@ class DGLBaseHeteroGraph(object):
     def __init__(
             self,
             metagraph,
-            node_types,
+            number_of_nodes_by_type,
             edge_connections_by_type):
         super(DGLBaseHeteroGraph, self).__init__()
 
@@ -143,7 +143,7 @@ class DGLHeteroGraph(DGLBaseHeteroGraph):
 
     Parameters
     ----------
-    metagraph, node_types, edge_connections_by_type :
+    metagraph, number_of_nodes_by_type, edge_connections_by_type :
         See DGLBaseHeteroGraph
     node_frame : dict[str, FrameRef], optional
         Node feature storage per type
@@ -156,12 +156,13 @@ class DGLHeteroGraph(DGLBaseHeteroGraph):
     def __init__(
             self,
             metagraph,
-            node_types,
+            number_of_nodes_by_type,
             edge_connections_by_type,
             node_frame=None,
             edge_frame=None,
             readonly=False):
-        super(DGLHeteroGraph, self).__init__(metagraph, node_types, edge_connections_by_type)
+        super(DGLHeteroGraph, self).__init__(
+                metagraph, number_of_nodes_by_type, edge_connections_by_type)
 
     # TODO: REVIEW
     def add_nodes(self, num, node_type, data=None):
--- a/python/dgl/heterograph.py
+++ b/python/dgl/heterograph.py
@@ -125,6 +125,9 @@ class DGLBaseHeteroGraph(object):
           source type name ``src_type_name``, destination type name
           ``dst_type_name``, and edge type name ``edge_type_name``.
 
+        The view would share the frames with the parent graph; any
+        modifications on one's frames would reflect on the other.
+
         Note that the subgraph itself is not materialized until someone
         queries the subgraph structure.  This implies that calling computation
         methods such as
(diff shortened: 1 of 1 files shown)


python/dgl/heterograph.py
```diff
@@ -0,0 +1,1387 @@
...
+    """View on a heterogeneous graph, constructed from
+    DGLBaseHeteroGraph.__getitem__().
+
+    It is semantically the same as a subgraph, except that
+
+    * The subgraph itself is not materialized until the user explicitly
+    queries the subgraph structure (e.g. calling ``in_edges``, but not
+    ``update_all``).
+    """
+    pass
```
zheng-da: what are the methods in this class?

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1387 @@
...
+
+        Notes
+        -----
+        An error is raised if the graph contains multiple node types.  Use
+
+            g[ntype]
+
+        to select nodes with type ``ntype``.
+        """
+        pass
```
zheng-da: it's a bit weird here. Suppose we have a bipartite graph. If we don't allow `nodes` on a bipartite graph and force people to use `g[ntype]`, `g[ntype]` from a bipartite graph actually has no edges at all. It should be OK for accessing data on nodes and performing computation on nodes. It's just a little weird that this is the only way of accessing node data on a bipartite graph.

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1387 @@
...
+            edge type.
+            The edge type is characterized by a triplet of source type name,
+            destination type name, and edge type name.
+            If the graph has more than one edge type and ``message_func`` is
+            not a dict, it will throw an error.
+
+        Notes
+        -----
+        On multigraphs, if :math:`u` and :math:`v` are specified, then the messages will be sent
+        along all edges between :math:`u` and :math:`v`.
```
zheng-da: How do we specify `u` and `v` if they have different node types? Do we have to use global ID?

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1387 @@
...
+        """
+        pass
+
+    def recv(self,
+             v=ALL,
+             reduce_func=None,
+             apply_node_func=None,
+             inplace=False):
+        """Receive and reduce incoming messages and update the features of node(s) :math:`v`.
```
zheng-da: the same comments here. `v` has to be global IDs?

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1387 @@
...
+    def send(self, edges=ALL, message_func=None):
+        """Send messages along the given edges with the same edge type.
+
+        ``edges`` can be any of the following types:
+
+        * ``int`` : Specify one edge using its edge id.
+        * ``pair of int`` : Specify one edge using its endpoints.
+        * ``int iterable`` / ``tensor`` : Specify multiple edges using their edge ids.
+        * ``pair of int iterable`` / ``pair of tensors`` :
+          Specify multiple edges using their endpoints.
```
zheng-da: The description here misses 'a dict of int, iterable of int, or tensors'

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1387 @@
...
+            number_of_nodes_by_type,
+            edge_connections_by_type,
+            node_frame=None,
+            edge_frame=None,
+            readonly=False):
+        super(DGLHeteroGraph, self).__init__(
+            metagraph, number_of_nodes_by_type, edge_connections_by_type)
+
+    # TODO: REVIEW
+    def add_nodes(self, num, node_type, data=None):
```
zheng-da: Should these interfaces be declared in `DGLBaseHeteroGraph `?

python/dgl/heterograph.py
```diff
@@ -0,0 +1,1387 @@
...
+        Parameters
+        ----------
+        data : tensor or dict of tensor
+            Edge representation.
+        etype : tuple[str, str, str]
+            The edge type, characterized by a triplet of source type name,
+            destination type name, and edge type name.
+        edges : edges
+            Edges can be a pair of endpoint nodes (u, v), or a
+            tensor of edge ids. The default value is all the edges.
```
zheng-da: if it's `(u, v)`, `u` and `v` should be global IDs, or should they be local IDs?

python/dgl/heterograph.py
BarclayII: The methods in `DGLBaseGraph` *would* materialize the subgraph since they explicitly ask for the structure of the subgraph.
For message passing APIs we clearly don't have to materialize it.

python/dgl/heterograph.py
BarclayII: I don't know.  I followed `DGLBaseGraph` which doesn't have `add_nodes` method.

zheng-da:
BTW, the interface doesn't contain the ones in `DGLBaseGraph`.

python/dgl/heterograph.py
BarclayII: I think this should be sufficient.  For instance, if you want to set item data on a `user -- item` bipartite graph, you would do this:
```
g['item'].ndata['logprice'] = ...
g['item'].nodes[[1, 3, 5]].data['category'] = ...
```
which looks natural.

That being said, the docstring indeed looks confusing.  I added an example for them.

python/dgl/heterograph.py
BarclayII: They are local IDs of source and destination types (added).

python/dgl/heterograph.py
zheng-da: I feel it's more appropriate to have all of these methods in `DGLBaseHeteroGraph`. Otherwise, I don't which class `DGLBipartiteGraph` should inherit from.

zheng-da:
I think the class also needs the following functions.
```python
def register_message_func(self, func):
def register_reduce_func(self, func):
def register_apply_node_func(self, func):
def register_apply_edge_func(self, func):
```

python/dgl/heterograph.py
```diff
@@ -0,0 +1,2092 @@
...
+        >>> g['user', 'game', 'plays'].has_edges_between([0, 0], [1, 2])
+        tensor([1, 0])
+
+        See Also
+        --------
+        has_edge_between
+        """
+        pass
+
+    def predecessors(self, v):
```
zheng-da: this method needs node type. otherwise, it doesn't work with bipartite graph.

python/dgl/heterograph.py
```diff
@@ -0,0 +1,2092 @@
...
+        --------
+        The following example uses PyTorch backend.
+
+        Find the user and game of gameplay #0 and #2:
+        >>> g['user', 'game', 'plays'].find_edges([0, 2])
+        (tensor([0, 1]), tensor([0, 1]))
+        """
+        pass
+
+    def in_edges(self, v, form='uv'):
```
zheng-da: this also needs node type.

python/dgl/heterograph.py
```diff
@@ -0,0 +1,2092 @@
...
+        --------
+        The following example uses PyTorch backend.
+
+        Find the gameplay IDs of game #0 (Tetris)
+        >>> g['user', 'game', 'plays'].in_edges(0, 'eid')
+        tensor([0, 1])
+        """
+        pass
+
+    def out_edges(self, v, form='uv'):
```
zheng-da: the same here.

python/dgl/heterograph.py
```diff
@@ -0,0 +1,2092 @@
...
+        --------
+        The following example uses PyTorch backend.
+
+        Find the user-game pairs for all gameplays:
+        >>> g['user', 'game', 'plays'].all_edges('uv')
+        (tensor([0, 1, 1, 2]), tensor([0, 0, 1, 1]))
+        """
+        pass
+
+    def in_degree(self, v):
```
zheng-da: this also require node type.

zheng-da: how to get degree in the original heterogeneous graph.

python/dgl/heterograph.py
BarclayII: If we require the subgraph to have only one edge type then it does.  `v` would always be of the node type of destination.

python/dgl/heterograph.py
BarclayII: Ditto.

python/dgl/heterograph.py
BarclayII: Ditto but of source type.

python/dgl/heterograph.py
BarclayII: > how to get degree in the original heterogeneous graph.

If you want to you can loop over the in-edges of the metagraph, get all the edge types connecting to the node type you're considering, and call this method for each edge type.

And I don't know if the degree of all types is meaningful enough; usually different entities like "User" and "Developer" should not be counted together anyway.

python/dgl/heterograph.py
```diff
@@ -0,0 +1,2094 @@
...
+    ...         # Nintendo develops Tetris and Mojang develops Minecraft
+    ...         ('developer', 'game', 'develops'): ([0, 1], [0, 1])})
+    """
+
+    # pylint: disable=unused-argument
+    def __init__(
+            self,
+            metagraph,
+            number_of_nodes_by_type,
+            edge_connections_by_type):
```
jermainewang: How to support converting from other format like networkx, scipy.

BarclayII: Added `from_networkx`.  I'm not sure how to handle scipy matrices since it's likely that we'll specify a list of scipy matrices for each edge type (which is handled by `__init__`)

python/dgl/heterograph.py
zheng-da: so you assume that all edges in a graph go from one type of nodes to another type? Is it always true?

BarclayII: We assume that all edges belonging to the same edge type always have the same source node type and destination node type (which is the assumption of the metagraph)

python/dgl/heterograph.py
zheng-da: I don't know what people want to do. Since this is query API, I feel it should be reasonably flexible. If a hetero graph has 10 node types in `ntypes`. Your solution will be something like this:
```python
in_deg = zeros(...)
for ntype in ntypes:
    in_deg = in_deg + g[ntype].in_degree(v)
for ntype1 in ntypes:
    for ntype2 in ntypes:
        for etype in etypes:
            in_deg = in_deg + g[ntype1, ntype2, etype].in_degree(v)
```
something like this. I think my code is probably incorrect. My point is that users' code can becomes unnecessarily complex and also very slow.

For a bipartite graph, what are `v`? Do they have to be left nodes? right nodes? It has to be one type of nodes, right? It's pretty confusing which side it should be.

BarclayII: > I don't know what people want to do. Since this is query API, I feel it should be reasonably flexible. If a hetero graph has 10 node types in `ntypes`. Your solution will be something like this:
> 
> ```python
> in_deg = zeros(...)
> for ntype in ntypes:
>     in_deg = in_deg + g[ntype].in_degree(v)
> for ntype1 in ntypes:
>     for ntype2 in ntypes:
>         for etype in etypes:
>             in_deg = in_deg + g[ntype1, ntype2, etype].in_degree(v)
> ```
> 
> something like this. I think my code is probably incorrect. My point is that users' code can becomes unnecessarily complex and also very slow.

Let's keep the interface simple first, as the current query interface is consistent with `DGLBaseGraph`.  I doubt if people would only care about the total number of predecessors without caring about the types.  If users do want more flexible interfaces we can later add new ones such as `typed_successors`.

> For a bipartite graph, what are `v`? Do they have to be left nodes? right nodes? It has to be one type of nodes, right? It's pretty confusing which side it should be.

They will always be "destination" nodes, since the bipartite graph would be *directed*.  A undirected bipartite graph would be "directionized" and the result would have a relationship (edge type) pointing left to right, and another pointing right to left.  So one can still do
```
g['left', 'right', 'ltor'].in_degrees(...)  # in degrees of right nodes
g['right', 'left', 'rtol'].in_degrees(...)  # in degrees of left nodes
```

zheng-da: i see. i feel this is a little counter-intuitive, but it works. we could have users to give us feedback later.

python/dgl/heterograph.py
zheng-da: If the class is inherited from `DGLBaseHeteroGraph`, there won't be computation APIs. It should be inherited from `DGLHeteroGraph`?

zheng-da: I don't think we should materialize the subgraph for query APIs. For example, `g['user'].number_of_nodes()`. If we need to materialize the subgraph before getting the number of nodes. It'll be too expensive.

Issue #558, opened 2019-05-23
RGCNEntityDataset not working if bfs_level = 0


askliar:
## 🐛 Bug

When trying to use `RGCNEntityDataset` from `dgl.contrib.data.knowledge_graph` with `bfs_level = 0` (so, without removing any nodes), it breaks due to wrong naming of the variables.

## To Reproduce

1. Pull code from [https://github.com/dmlc/dgl/blob/master/examples/pytorch/rgcn](https://github.com/dmlc/dgl/blob/master/examples/pytorch/rgcn)
2. In file entity_classify.py change `args.bfs_level = args.n_layers + 1 # pruning used nodes for memory` to `args.bfs_level = 0`.
3. Run `python entity_classify.py`
4. Get error that `self.edge_dst` and `self.edge_src` are not defined.

## Expected behavior

There should be no error in case of any bfs_level as long as it is not larger than the maximum distance in the graph. In the case of zero, simply no nodes should be removed. 

## Environment

 - DGL Version : 2.0
 - Backend Library & Version: Pytorch

PR #559, opened 2019-05-23
[BugFix] Fix error when bfs_level = 0 in Entity Classification with RGCN

askliar:
## Description
This pull request serves as a fix for #558. It renames edge_src and edge_dst to src and dst in all files related to entity classification task with RGCN.


askliar:
@ylfdq1118 I included your suggestions. Indeed, that was a simpler fix.

PR #573, opened 2019-05-27
[Feature] add bipartite graph.

zheng-da:
## Description
This PR implements DGLBipartiteGraph with the hetero graph API for bipartite graph. The edges in DGLBipartiteGraph have to go from one side of nodes to the other side. If we want to support a bipartite graph with bidirectional edges, we need to use two DGLBipartiteGraph.

In addition, this implementation reuses the original graph structure for bipartite graphs. To support bipartite graph, we put nodes of the two sides into the same space. To be more specific, the source nodes have Ids between [0, #source nodes - 1] and the destination nodes have Ids between [#source nodes, #source nodes + #destination nodes - 1]. We'll improve this graph structure in the future and make it more specific for bipartite graphs.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/include/dgl/immutable_graph.h
+++ b/include/dgl/immutable_graph.h
@@ -920,6 +920,34 @@ CSR::CSR(int64_t num_vertices, int64_t num_edges,
   }
 }
 
+class ImmutableBiGraph: public ImmutableGraph {
+ public:
+  /*! \brief Construct an immutable graph from the COO format. */
+  explicit ImmutableBiGraph(COOPtr coo, size_t num_src, size_t num_dst): ImmutableGraph(coo) {
+    this->num_src = num_src;
+    this->num_dst = num_dst;
+  }
+  /*! \brief Construct an immutable graph from the CSR format. */
+  ImmutableBiGraph(CSRPtr in_csr, CSRPtr out_csr, size_t num_src, size_t num_dst)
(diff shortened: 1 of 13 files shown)


yzh119:
@zheng-da , I'm writing the pooling layers for DGL, and I found BipartiteGraph might be useful.
For example, a diffpool layer might look like this:
```python
class DiffPoolLayer(nn.Module):
    def __init__(self, *args, **kwargs):
        ...

    def forward(self, h, g):
        ...
        return h_new, g_new, bipartite_graph(g, g_new)
```
And apply `send_and_recv` on the edges of the bipartite_graph.
For U-Net like model, we might need `bipartite_graph(g_new, g)`.

Proposal of `bipartite_graph`:

```python
def bipartite_graph(g_u, g_v):
    r"""Construct a bipartite graph between the nodes in g_u and nodes in g_v.

    Parameters
    ----------
    g_u, g_v: DGLGraph
        
    Returns
    -------
    g_bi: DGLBipartiteGraph

    The node data frame of the returned graph would be FrameRef of the concatenation of the node data frame of g_u and node data frame of g_v.
    """
    ...
```

BarclayII:
Should we rebase it to master since `heterograph` is merged?

zheng-da:
yes, i'll rebase it to the master branch.

PR #583, opened 2019-05-29
[Docker]Roll back CI to use MXNet nightly build 20190523

ylfdq1118:
## Description
Latest mxnet has a bug in its NDArray::Save function, which breaks DGL code. Rollback CI to use MXNet nightly build on 20190523


jermainewang:
Shall we close this PR?

PR #584, opened 2019-05-29
[BugFix] Some negative layer_id/block_id are not converted to the actual one.

zengxy:
## Description
Some negative layer_id/block_id are not converted to the actual ones, thus wrong result is got when refer to `block_id + k`.

For example,  `block_id, block_id + 1, block_id + 2` are used  in function `NodeFlow.block_edges` as below shows.
https://github.com/dmlc/dgl/blob/ada3786c5761136415915066e3de9c7207b58159/python/dgl/nodeflow.py#L417-L421
 And when `block_id` is negative it will throw an exception or cause ipython crash in my laptop.

Codes can be used to reproduce.
```python
import dgl
g = dgl.DGLGraph()
g.add_nodes(5)
g.add_edges([0,1,1,2,3,3,4], [3,4,1,3,4,2,0])
g.readonly();nf = next(iter(dgl.contrib.sampling.NeighborSampler(g, 2, 2, 4)))

# throw an exception but expected to be the same with nf.block_edges(3)
nf.block_edges(-1)

# cause my ipython crashed but expected to be the same with nf.block_edges(2)
nf.block_edges(-2)

```

## Checklist
- [ ] Code is well-documented


zheng-da:
Thanks for fixing the bugs. @zengxy Could you add some tests for the fix?

zengxy:
I've add some tests for changed functions. @zheng-da 
I add a test for `block_incidence_matrix` with param `typestr =  "in" and "out"` but without `typestr =  "both"` for shape value `(n, m)` is not defined for the matrix when `typestr = both`
https://github.com/dmlc/dgl/blob/e0ce190dfe6ead2fb2dfd4b2755af95177519fee/python/dgl/nodeflow.py#L536-L552

I don't know how to fix it  because I thought that row_index of the matrix may refer to the local nid in a certain sample layer. But add both src and dst, which are in different sample layers, to a single matrix is ambiguous for me.

zheng-da:
@zengxy sorry for the late reply. You are right. I also realized that a bipartite graph doesn't have an incidence matrix for both in-edges and out-edges. I think we should remove the support of "both" and throw an exception saying that this is an invalid type.

zengxy:
@zheng-da Got it. Thank you!

PR #587, opened 2019-05-30
[Graph] Add API to convert graph to simple graph

jermainewang:
## Description

Add `dgl.to_simple_graph` to convert a graph to a simple graph. Also refactor the graph_index a little bit.

## Checklist
- [x] Code is well-documented

## Changes
- [x] `dgl.to_simple_graph` API.
- [x] Refactor the `from_XXX` functions in `graph_index.py`. They are not class methods anymore. They take in the corresponding graph data and return a graph index object.
- [x] Double check all the places where `multigraph` flag is used, so that:
  * `multigraph` flag is default to be None (previously False). None means the flag will be determined by the data.
  * If `multigraph` flag is given by the user, the behavior is the same as before.
  * For mutable graph, the flag is default to be False (same behavior as before).


python/dgl/graph.py
```diff
@@ -765,7 +765,8 @@ class DGLGraph(DGLBaseGraph):
     edge_frame : FrameRef, optional
         Edge feature storage.
     multigraph : bool, optional
-        Whether the graph would be a multigraph (default: False)
+        Whether the graph would be a multigraph. If none, the flag will be determined
+        by scanning the whole graph. (default: None)
```
zheng-da: this sounds very expensive. then we should force users to provide `multigraph` for graph store.

zheng-da: please remove the default value here. https://github.com/dmlc/dgl/blob/master/python/dgl/contrib/graph_store.py#L1020

jermainewang: For immutable graph, the flag will be only computed once and only when `is_multigraph` is called. That'll be fine?

jermainewang:
I found some issues with our CI. Will fix it in another PR.

zheng-da:
I see there are many modifications in Jenkins. What are the modifications for?

python/dgl/graph.py
zheng-da: For giant graphs, I think it's better to ask users to provide this information? Is it fast to test it for a very large graph?

python/dgl/graph_index.py
```diff
@@ -45,28 +46,29 @@ def __setstate__(self, state):
...
-            dst_ids.todgltensor(),
-            self._multigraph,
-            int(num_nodes),
-            self._readonly)
+            self._handle = _CAPI_DGLGraphCreate(
+                src.todgltensor(),
+                dst.todgltensor(),
+                multigraph,
+                int(num_nodes),
+                readonly)
```
zheng-da: will it be better to use one API and the values of multigraph can be -1, 0, 1? too hacky?

src/graph/graph_op.cc
```diff
@@ -273,4 +297,22 @@ IdArray GraphOp::ExpandIds(IdArray ids, IdArray offset) {
...
+      if (!hashmap.count(dst)) {
+        indices.push_back(dst);
+        hashmap.insert(dst);
+      }
+    }
+    indptr[src+1] = indices.size();
+  }
+  CSRPtr csr(new CSR(graph->NumVertices(), indices.size(),
+        indptr.begin(), indices.begin(), RangeIter(0), false));
+  return ImmutableGraph(csr);
```
zheng-da: if the original graph index has only in_csr. the code first generates out_csr and return out_csr. Later on, the user's code needs in_csr. Maybe we don't need to spend time optimizing this operation. I just raise a potential issue. It's up to you to decide.

tests/graph_index/test_graph_index.py
```diff
@@ -1,175 +0,0 @@
-from dgl import DGLError
```
zheng-da: why removing this file?

jermainewang: It looks a duplicate of `tests/compute/test_graph_index.py`.

jermainewang:
> I see there are many modifications in Jenkins. What are the modifications for?

I'm testing the CI and found some bugs. I'll fix that in another PR and revert the changes here.

PLEASE DO NOT MERGE THIS PR

python/dgl/graph_index.py
jermainewang: That's fine as long as it is not exposed to user side. Will do that to eliminate some copy-pasting.

src/graph/graph_op.cc
jermainewang: Actually `SuccVec` will create `out_csr`, but you are right, it is always a decision to create which csr. I don't have good solution. It looks like a layout optimization problem that requires more information than simply one operator. For example, if we have a computation graph, then we know when and where to use which format so the format conversion cost + computation cost is minimized. Pretty far goal I will say.

python/dgl/graph.py
jermainewang: Yeah, we still allow user to directly provide the flag. I guess in the giant graph example, it's better to specify that explicitly. What do you think?

zheng-da: yes, this is what I meaned.

python/dgl/graph_index.py
jermainewang: Done

PR #596, opened 2019-06-02
[Feature][Kernel] DGL kernel support

ylfdq1118:
## Description

This PR adds fused message passing kernels to DGL.

Code review plan:
Please read the PR breakdown below first.
- [x] C++ side changes (@ylfdq1118 @zheng-da  )
- [x] Python side changes (@jermainewang @BarclayII )
   - [x] Scheduler changes (@zheng-da )
- [x] Others (@jermainewang )

## Checklist
- [x] Code is well-documented

## Changes
**C++ side changes**

_Graph_
Add `override` qualifier;  Two new APIs: `Context` and `NumBits`;  `CopyTo` API for immutable graph.
* `include/dgl/graph.h`
* `include/dgl/graph_interface.h`
* `include/dgl/immutable_graph.h`
* `src/graph/graph_apis.cc`
* `src/graph/immutable_graph.cc`
* `src/graph/nodeflow.cc`

_Runtime_
CUDA runtime codes copied from TVM project.
* `src/runtime/cuda/cuda_common.h`
* `src/runtime/cuda/cuda_device_api.cc`

_Kernel_
The main kernel is `BinaryReduce`. It implements `C=Sum(A op B)` where A, B could be either src/dst/edge features, op could be add/mul/sub/div and Sum could be sum/max/min/prod. To support all these combinations, template is widely used. When exposing them to python, we use narrow API design -- only `binary_reduce` is exposed. Thus, in C++, we need to dispatch the given runtime arguments to different template specialization. That's why you will see many macro unrolling in the code. Also, to speedup the compilation (as we have so many combinations), we separate the definition in multiple translation units so they can be compiled in parallel. For example, `src/kernel/cpu/binary_reduce_sum.cc` defines all the kernels whose reduce function is sum. Since there are many common parts in CPU and GPU implementation, we extract the common parts in `src/kernel/binary_reduce_impl.h` and put the device-specific implementation in `src/kernel/xpu`.

All the kernel source files are in `src/kernel/` folder. It is organized as follows:
* `src/kernel/common.h`: Common headers for all kernels.
* `src/kernel/binary_reduce.h`: Non-template kernel functions. The functions there are exposed to python in `kernel.py` and can support CPU/GPU.
* `src/kernel/binary_reduce.cc`: CAPI definition.
* `src/kernel/binary_reduce_common.h`: Common headers for binary reduce kernels.
* `src/kernel/binary_reduce_impl.h`: Common implementation that are shared by CPU and GPU.
* `src/kernel/binary_reduce_impl_decl.h`
* `src/kernel/cpu/`: All CPU kernels.
* `src/kernel/cuda/`: All GPU kernels.

_ Other C++ refactor and bugfix_
* `include/dgl/runtime/pack_func.h`: Support passing DLContext object between Python and C.
* `include/dgl/array.h`, `src/array.cc`: Add `AsNumBits`.

**Python side changes**
We first expose fused `BinaryReduce` kernels to python (`kernel.py`). We then convert it to framework-dependent operators (`dgl/backend` package). We then invoke them in the scheduler (`dgl/runtime`). Finally, we expose them as new built-in functions (`dgl/function`). Compare with previous invoke stack, one big difference is that an immutable graph object is passed down from python to C++ instead of framework-specific sparse matrix.

_Framework-specific adapters of new kernels_
* `python/dgl/backend/backend.py`
* `python/dgl/backend/mxnet/tensor.py`
* `python/dgl/backend/pytorch/tensor.py`

_Message and reduce functions_
Replace old builtin functions (e.g. `SrcMulEdge`) with `BinaryMessageFunction` and so on. Change its constructor functions with proper arguments. Implement its schedule logic in the `__call__` operator overload.
* `python/dgl/function/base.py`
* `python/dgl/function/message.py`
* `python/dgl/function/reducer.py`

_Graph & kernels_
* `python/dgl/graph_index.py`: Cache an `ImmutableGraph` object instead of the previous adj arrays. Add `to_immutable`, `ctx`, 
  `copy_to`, `nbits`, etc APIs so we could cache immutable graph of various device contexts and nbits 
  (int32 or int64). Currently GPU only supports int32 graph while int64 graph is needed for giant graph 
  training in CPU.
* `python/dgl/kernel.py`: The python interface for the fused message passing kernels. The main one is `binary_op_reduce`.
* `python/dgl/nn/pytorch/softmax.py`: The new softmax module using new builtins.
* `python/dgl/nodeflow.py`

_Scheduler change_
* `python/dgl/runtime/ir/executor.py`
* `python/dgl/runtime/ir/var.py`
* `python/dgl/runtime/scheduler.py`
* `python/dgl/runtime/spmv.py`

_Other python refactor/bugfix changes_
* `python/dgl/_ffi/_ctypes/types.py`: Support passing DLContext object between Python and C.

**Build system changes:**
* `.gitmodules`: Add `minigun` dependency.
* `CMakeLists.txt`: CUDA build
* Files under `cmake` folders: cmake utilities (copied from MXNet and TVM projects)
  - `cmake/config.make`: cmake option file
  - `cmake/modules/CUDA.cmake`
  - `cmake/util/FindCUDA.cmake`, `cmake/util/MshadowUtil.cmake`, `cmake/util/Util.cmake`.

**CI changes:**
* `Jenkinsfile`: Add GPU unittest.
* Files under `tests/scripts`.
* Unittests (files under `tests/compute`)

## TODO
- mxnet test_share_mem (@zheng-da )
- (next PR) Nodeflow create graph index from csr
- (next PR) support bipartite graph index
- (next PR) support mean reduction.
- (next PR) enable layer sampling.


include/dgl/immutable_graph.h
```diff
@@ -68,6 +68,14 @@ class CSR : public GraphInterface {
     LOG(FATAL) << "CSR graph does not allow mutation.";
   }
 
+  DLContext Context() const override {
+    return indptr_->ctx;
+  }
+
+  uint8_t NumBits() const override {
+    return indptr_->dtype.bits;
```
zheng-da: can we use `indices_->dtype.bits`? in theory, `indptr` and `indices` can have different number of bits. we can scale to a graph with more than 4B edges and still use 32-bit indices.

python/dgl/backend/mxnet/tensor.py
```diff
@@ -246,3 +257,124 @@ def zerocopy_to_numpy(arr):
...
+            self.in_map[1], self.out_map[1])
+        return grad_in
+
+
+def copy_reduce(reducer, graph, target, in_data, out_size, in_map, out_map):
+    func = CopyReduce(reducer, graph, target, out_size, in_map, out_map)
+    return func(in_data)
+
+
+def _reduce_grad(grad, shape):
```
zheng-da: what is this function for?

python/dgl/backend/pytorch/tensor.py
```diff
@@ -31,24 +34,12 @@ def get_preferred_sparse_format():
...
-            raise TypeError('Pytorch backend only supports COO format. But got %s.' % fmt)
-        spmat = th.sparse_coo_tensor(index[1], data, shape)
-        # No conversion is required.
-        return spmat, None
+def sparse_matrix(data, index, shape, force_format=False):
+    fmt = index[0]
+    if fmt != 'coo':
+        raise TypeError('Pytorch backend only supports COO format. But got %s.' % fmt)
+    spmat = th.sparse_coo_tensor(index[1], data, shape)
+    return spmat, None
```
zheng-da: we will not support pytorch 0.x?

python/dgl/graph_index.py
```diff
@@ -598,8 +599,38 @@ def adjacency_matrix_scipy(self, transpose, fmt):
...
+            The first element of the tuple is the shuffle order for outward graph
+            The second element of the tuple is the shuffle order for inward graph
+        """
+        csr = _CAPI_DGLGraphGetAdj(self._handle, True, "csr")
+        order = csr(2)
+        rev_csr = _CAPI_DGLGraphGetAdj(self._handle, False, "csr")
+        rev_order = rev_csr(2)
+        return utils.toindex(order), utils.toindex(rev_order)
 
-    @utils.cached_member(cache='_cache', prefix='adj')
```
zheng-da: why removing the caching?

python/dgl/graph_index.py
```diff
@@ -650,7 +681,6 @@ def adjacency_matrix(self, transpose, ctx):
         else:
             raise Exception("unknown format")
 
-    @utils.cached_member(cache='_cache', prefix='inc')
```
zheng-da: the same here.

python/dgl/graph_index.py
```diff
@@ -761,6 +791,88 @@ def line_graph(self, backtracking=True):
...
+
+        Returns
+        -------
+        int
+            The number of bits needed
+        """
+        if self.number_of_edges() >= 0x80000000:
+            return 64
+        elif self.number_of_nodes() >= 0x80000000:
+            return 64
```
zheng-da: shouldn't be "or"? #edges can be smaller than 2^31 but #nodes is > 2^31.

python/dgl/backend/mxnet/tensor.py
ylfdq1118: If the forward pass does broadcasting on certain feature dimension, then in the backward pass the gradient should be reduced on that dimension. Currently for simplicity, DGL C++ kernel does not do the gradient reduction, and therefore, it's done in python side.

python/dgl/nodeflow.py
```diff
@@ -244,7 +246,8 @@ def map_to_parent_nid(self, nid):
         Tensor
             The parent node id array.
         """
-        return self._node_mapping.tousertensor()[nid]
+        nid = utils.toindex(nid)
```
zheng-da: the input nid should be a tenor. why do we need to convert it dgl tensor first?

python/dgl/nodeflow.py
```diff
@@ -259,7 +262,8 @@ def map_to_parent_eid(self, eid):
         Tensor
             The parent edge id array.
         """
-        return self._edge_mapping.tousertensor()[eid]
+        eid = utils.toindex(eid)
```
zheng-da: the same here.

python/dgl/runtime/scheduler.py
```diff
@@ -42,19 +44,22 @@ def schedule_send(graph, u, v, eid, message_func):
     message_func: callable or list of callable
         The message function
     """
-    message_func = _standardize_func_usage(message_func, 'message')
-    mfunc_is_list = utils.is_iterable(message_func)
-    if mfunc_is_list:
-        message_func = BundledFunction(message_func)
```
zheng-da: why are these lines removed? do we still support list of callables?

python/dgl/runtime/scheduler.py
```diff
@@ -146,14 +151,23 @@ def schedule_snr(graph,
     var_recv_nodes = var.IDX(recv_nodes, name='recv_nodes')
     # generate send and reduce schedule
     uv_getter = lambda: (var_u, var_v)
-    adj_creator = lambda: spmv.build_adj_matrix_uv((u, v), recv_nodes, graph.number_of_nodes())
-    inc_creator = lambda: spmv.build_inc_matrix_dst(v, recv_nodes)
```
zheng-da: why remove incidence matrix? do we still support the inputs that message function is UDF and reduce function is a builtin?

python/dgl/runtime/scheduler.py
```diff
@@ -193,14 +206,22 @@ def schedule_update_all(graph,
         def uv_getter():
             src, dst, _ = graph._graph.edges('eid')
             return var.IDX(src), var.IDX(dst)
-        adj_creator = lambda: spmv.build_adj_matrix_graph(graph)
-        inc_creator = lambda: spmv.build_inc_matrix_graph(graph)
```
zheng-da: the same comment here.

python/dgl/runtime/scheduler.py
```diff
@@ -301,24 +322,17 @@ def schedule_apply_edges(graph,
     A list of executors for DGL Runtime
     """
     # vars
-    var_nf = var.FEAT_DICT(graph._node_frame, name='nf')
+    var_nf = var.FEAT_DICT(graph._node_frame)
+    var_ef = var.FEAT_DICT(graph._edge_frame)
+    var_out = _gen_send(graph=graph, u=u, v=v, eid=eid, mfunc=apply_func,
+                        var_src_nf=var_nf, var_dst_nf=var_nf, var_ef=var_ef)
```
zheng-da: use the send scheduler for apply_edges?

python/dgl/runtime/spmv.py
```diff
@@ -165,236 +130,86 @@ def build_adj_matrix_graph(graph):
...
+    if edge_tuples is None:
+        u, v, eid = graph.block_edges(block_id, remap=True)
+        u = utils.toindex(u)
+        v = utils.toindex(v)
+        eid = utils.toindex(eid)
+    else:
+        u, v, eid = edge_tuples
+    num_nodes = max(graph.layer_size(block_id), graph.layer_size(block_id + 1))
+    gidx, edge_map, nbits = build_gidx_and_mapping_uv((u, v, eid), num_nodes)
+    return gidx, edge_map, nbits
```
zheng-da: please slice block csr directly.

src/kernel/binary_reduce.cc
```diff
@@ -0,0 +1,506 @@
...
+    CHECK_EQ(arrays[i]->ndim, 1);
+    CHECK_EQ(bits, arrays[i]->dtype.bits)
+      << "Expected " << bits << " integer array. But got "
+      << arrays[i]->dtype.bits << " for " << names[i] << ".";
+  }
+}
+
+// Return true if the operator is commutative and lhs and rhs need
+// to be switched. For example, Add(kDst, kSrc) needs to be changed
+// to Add(kSrc, kDst)
```
zheng-da: i don't understand. why we need to switch "Add(kDst, kSrc)" to "Add(kSrc, kDst)".

src/kernel/binary_reduce.cc
```diff
@@ -0,0 +1,506 @@
...
+  CheckIdArray(graph->NumBits(),
+      {lhs_mapping, rhs_mapping, out_mapping},
+      {"lhs_mapping", "rhs_mapping", "out_mapping"});
+  // Switch order for commutative operation
+  if (NeedSwitchOrder(op, lhs, rhs)) {
+    BinaryOpReduce(reducer, op, graph,
+        rhs, lhs, rhs_data, lhs_data, out_data,
+        rhs_mapping, lhs_mapping, out_mapping);
+  } else {
+    if (HasBcast(lhs_data, rhs_data)) {
```
zheng-da: is it necessary to support broadcasting? what is the use case?

src/kernel/binary_reduce.cc
```diff
@@ -0,0 +1,506 @@
...
+          reducer, op, graph,
+          lhs, rhs,
+          lhs_mapping, rhs_mapping, out_mapping,
+          lhs_data, rhs_data, out_data, grad_out_data,
+          grad_lhs_data, utils::NoneArray());
+    }
+  }
+}
+
+DGL_REGISTER_GLOBAL("kernel._CAPI_DGLKernelBackwardLhsBinaryOpReduce")
```
zheng-da: why do you need to break it into two CAPIs?

src/kernel/binary_reduce_common.h
```diff
@@ -0,0 +1,450 @@
...
+static const char kMul[] = "mul";
+static const char kDiv[] = "div";
+static const char kUseLhs[] = "use_lhs";
+
+enum Target {
+  kSrc = 0,
+  kDst,
+  kEdge,
+  kNone,
+};
```
zheng-da: what does this mean? please write some comments.

src/kernel/binary_reduce_impl.h
```diff
@@ -0,0 +1,414 @@
...
+  }
+  if (!utils::IsNoneArray(rhs_mapping)) {
+    gdata.rhs_mapping = static_cast<Idx*>(rhs_mapping->data);
+  }
+  if (!utils::IsNoneArray(out_mapping)) {
+    gdata.out_mapping = static_cast<Idx*>(out_mapping->data);
+  }
+  // fill out data with zero values
+  utils::Fill<XPU>(ctx, gdata.out_data, utils::NElements(out_data), Zero<Reducer>::value);
+  return gdata;
```
zheng-da: what is GData?

python/dgl/function/message.py
```diff
@@ -25,233 +27,193 @@ def name(self):
...
+        in_frames = (src_frame, dst_frame, edge_frame)
+        in_maps = (src_map, dst_map, edge_map)
+        lhs_data = ir.READ_COL(in_frames[self.lhs], var.STR(self.lhs_field))
+        rhs_data = ir.READ_COL(in_frames[self.rhs], var.STR(self.rhs_field))
+        lhs_map = var.MAP(in_maps[self.lhs])
+        rhs_map = var.MAP(in_maps[self.rhs])
+        out_map = var.MAP(out_map)
+        return ir.BINARY_REDUCE(reducer, self.binary_op, graph, self.lhs,
+                                self.rhs, lhs_data, rhs_data, out_size,
+                                lhs_map, rhs_map, out_map)
```
BarclayII: Add a TODO here: ideally `dgl.function` shouldn't depend on IR.

python/dgl/function/reducer.py
```diff
@@ -23,34 +24,37 @@ def name(self):
...
+    def __call__(self, graph, edge_frame, out_size, edge_map=None,
+                 out_map=None):
+        """Symbolic execution of this builtin function"""
+        reducer = self._name
+        graph = var.GRAPH(graph)
+        edge_map = var.MAP(edge_map)
+        out_map = var.MAP(out_map)
+        edge_data = ir.READ_COL(edge_frame, var.STR(self.msg_field))
+        return ir.COPY_REDUCE(reducer, graph, TargetCode.EDGE, edge_data,
+                              out_size, edge_map, out_map)
```
BarclayII: Same TODO

src/kernel/binary_reduce.cc
jermainewang: Multi-head graph attention network. The node feature shape is (H, D) and the edge feature shape is (H, 1).

src/kernel/binary_reduce.cc
jermainewang: This is to speedup the compilation. We don't want to generate kernels for both Add(kDst, kSrc) and Add(kSrc, kDst). They are the same.

python/dgl/backend/pytorch/tensor.py
ylfdq1118: We support 0.x. But sparse_matrix is only for users. DGL scheduler will not need it any more. So we will just use the standard API to create the coo tensor

python/dgl/graph_index.py
ylfdq1118: Because adj_matrix and inc_matrix are only for users. DGL scheduler won't need them any more. If user need to use them multiple times, they should cache by themselves

python/dgl/graph_index.py
ylfdq1118: Same as adjmat, we ask users to cache them.

python/dgl/graph_index.py
ylfdq1118: What do you suggest? I think if either # edges or # nodes is larger than 2^31 -1, we need to use 64 bits.

python/dgl/nodeflow.py
ylfdq1118: The reason is for expression `a[b]`, MX requires `a` and `b` to be on the same device. However, sometimes, the `nid` is a GPU tensor while `self._node_mapping` is always on CPU. Converting to index not only handles the case that `nid` is not a tensor, but also does a GPU to CPU copying when the next line calls `nid.tousertensor()`

python/dgl/nodeflow.py
ylfdq1118: Same as above

python/dgl/runtime/spmv.py
ylfdq1118: Will do in next PR. There are some subtlety I need to carefully think about.

python/dgl/function/message.py
ylfdq1118: Hmm... Unfortunately, this is a "feature"... Because we want to unify all DGL kernel interfaces to support more kernels, the message function interface has to be unified. I designed this so that scheduler does not need knowledge about the what builtin message and reduce function are. If you think this is problematic, we probably need to discuss if other choices are available.

If you just don't like my using __call__ because user might use it as well, I can instead use some other private function name for the symbolic execution

python/dgl/function/reducer.py
ylfdq1118: Same as above

python/dgl/backend/pytorch/tensor.py
BarclayII: I think not supporting 0.x is OK now.

src/kernel/common.h
```diff
@@ -0,0 +1,108 @@
...
+    typedef float DType;                                    \
+    {__VA_ARGS__}                                           \
+  } else if (val.code == kDLFloat && val.bits == 64) {      \
+    typedef double DType;                                   \
+    {__VA_ARGS__}                                           \
+  } else {                                                  \
+    LOG(FATAL) << "Unsupported dtype: " << val.code << "_"  \
+               << val.bits;                                 \
+  }
+#else
```
BarclayII: So we only support float32?

python/dgl/kernel.py
```diff
@@ -0,0 +1,260 @@
...
+    rhs_data : NDArray
+        The rhs data.
+    out_data : NDArray
+        The out data.
+    lhs_mapping : NDArray
+        The lhs id mapping array.
+    rhs_mapping : NDArray
+        The rhs id mapping array.
+    out_mapping : NDArray
+        The out id mapping array.
```
BarclayII: If `xxx_mapping` are given, I assume that `lhs`, `rhs` and `out` must stay consistent?  Like
* If `reducer, lhs, rhs = none, src, edge`, then `rhs_mapping == out_mapping && lhs_mapping == src_nodes_of(rhs_mapping)` must hold.
* If `reducer, lhs, rhs = sum, dst, edge`, then `out_mapping == dst_nodes_of(rhs_mapping) == lhs_mapping` must hold.

python/dgl/function/message.py
BarclayII: > If you just don't like my using **call** because user might use it as well, I can instead use some other private function name for the symbolic execution

Yeah let's do that.

python/dgl/backend/backend.py
```diff
@@ -842,6 +833,105 @@ def zerocopy_from_numpy(np_array):
...
+    rhs_data : Tensor
+        The rhs data
+    out_size : int
+        Size of first dimension of output data
+    lhs_map : tuple of two dgl.ndarray.NDArray
+        The lhs id mapping array
+    rhs_map : tuple of two dgl.ndarray.NDArray
+        The rhs id mapping array
+    out_map : tuple of two dgl.ndarray.NDArray
+        The out id mapping array
```
BarclayII: Why are they tuples?

python/dgl/nn/pytorch/softmax.py
```diff
@@ -50,46 +47,86 @@ def forward(self, logits, graph):
...
+        gidx = graph._graph.get_immutable_gidx(ctx)
+        _, dst, _ = graph._graph.edges()
+        dst = dst.tousertensor(F.context(logits))
+        empty_map = (None, None)
+        max_logits_ = F.copy_reduce("max", gidx, fn.TargetCode.EDGE, logits,
+                                    num_nodes, empty_map, empty_map)
+        logits = (logits - max_logits_.index_select(0, dst)).exp()
+        norm = F.copy_reduce("sum", gidx, fn.TargetCode.EDGE, logits,
+                             num_nodes, empty_map, empty_map)
+        return logits / norm.index_select(0, dst)
```
BarclayII: I assume that this class should be removed and replaced with the class below.

python/dgl/graph_index.py
jermainewang: `if self.number_of_edges() >= ... or self.number_of_nodes() >= ...`

python/dgl/runtime/ir/executor.py
```diff
@@ -1169,3 +995,254 @@ def CLEAR_FRAME_(fd):
...
+        graph = self.graph.data(ctx)
+        in_map = self.in_map.data(ctx) if self.in_map.data else None
+        out_map = self.out_map.data(ctx) if self.out_map.data else None
+        if not isinstance(in_map, tuple):
+            in_map = (in_map, in_map)
+        if not isinstance(out_map, tuple):
+            out_map = (out_map, out_map)
+        self.ret.data = F.copy_reduce(
+            self.reducer, graph, self.target, in_data, self.out_size, in_map,
+            out_map)
```
BarclayII: Did we ever benchmark our binary/copy reducer against PyTorch SPMM?  Just curious.  We don't have to do it.

python/dgl/backend/mxnet/tensor.py
jermainewang: Please add comment

include/dgl/immutable_graph.h
jermainewang: Done.

python/dgl/runtime/scheduler.py
```diff
@@ -746,26 +774,36 @@ def _gen_send_reduce(
...
+        Function that returns a pair of var.IDX (u, v) for the triggered edges.
     adj_creator : callable
-        A function that returns the adjmat and the shuffle index.
-    inc_creator : callable
-        A function that returns the incmat and the shuffle index.
+        Function that returns the adjmat, edge order of csr matrix, and edge
+        order of transposed csr matrix
+    out_map_creator: callable
+        A function that returns a mapping from reduce_nodes to relabeled
+        consecutive ids
```
BarclayII: Why do we need this as an argument?

python/dgl/runtime/scheduler.py
```diff
@@ -774,61 +812,84 @@ def _gen_send_reduce(
...
-        mfunc = BundledFunction(mfunc)
-
-    # generate UDF send schedule
-    var_u, var_v = uv_getter()
-    var_mf = _gen_send(graph, var_src_nf, var_dst_nf, var_ef, var_u, var_v, var_eid, mfunc)
+        # messages are builtin but reduce is UDF
+        # Create a tmp frame to hold the message.
+        n_message = len(var_eid.data)
+        tmp_msg_frame = FrameRef(frame_like(edge_frame._frame, n_message))
+        var_mf = var.FEAT_DICT(data=tmp_msg_frame)
```
BarclayII: Does the TODO in L815 apply here and in L919?

src/kernel/cpu/binary_reduce_impl.h
```diff
@@ -0,0 +1,228 @@
...
+    gdata->rhs_mapping = static_cast<Idx*>(outcsr->edge_ids()->data);
+  }
+  if (OutSelector<Reducer>::Type::target == binary_op::kEdge
+      && gdata->out_mapping == nullptr) {
+    gdata->out_mapping = static_cast<Idx*>(outcsr->edge_ids()->data);
+  }
+  // TODO(minjie): allocator
+  minigun::advance::Advance<XPU, Idx, cpu::AdvanceConfig,
+    BcastGData<NDim, Idx, DType>, UDF>(
+        rtcfg, csr, gdata, minigun::IntArray1D<Idx>());
```
BarclayII: Question about minigun: is output frontier being used in DGL in any way?  I noticed that a chunk of code in minigun fills in the output frontier buffer which seems to be discarded afterwards.

jermainewang: No, it is not. Currently, for subgraph update, we generate explicit subgraph and perform full advance on it. The partial frontier could be useful to speedup `pull` so no explicit subgraph is needed.

src/kernel/binary_reduce.cc
```diff
@@ -0,0 +1,506 @@
...
+  }
+  for (int i = 1; i < lhs->ndim; ++i) {
+    if (lhs->shape[i] != rhs->shape[i]) {
+      return false;
+    }
+  }
+  return true;
+}
+
+bool HasBcast(NDArray lhs, NDArray rhs) {
```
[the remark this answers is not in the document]
jermainewang: added more comments. pls checkout.

src/kernel/binary_reduce.cc
jermainewang: More comments. Pls checkout.

src/kernel/binary_reduce.cc
jermainewang: Because it is possible that only one gradient is needed. For example, in `u_mul_e` case, if e is a fixed edge feature that does not require gradient, then only BackwardLhsBinaryOpReduce will be called.

src/kernel/binary_reduce_common.h
jermainewang: More comments. Pls checkout.

src/kernel/binary_reduce_impl.h
jermainewang: More comments (in binary_reduce_impl_decl.h). Pls checkout.

python/dgl/backend/mxnet/tensor.py
ylfdq1118: Done

python/dgl/graph_index.py
ylfdq1118: Done

src/kernel/common.h
jermainewang: Yes. I plan to enable more data types later, but the compilation will become even slower ...

python/dgl/function/message.py
ylfdq1118: Done

src/kernel/common.h
ylfdq1118: Yes

src/kernel/cpu/functor.h
```diff
@@ -0,0 +1,76 @@
+/*!
+ *  Copyright (c) 2019 by Contributors
+ * \file kernel/cpu/functor.h
```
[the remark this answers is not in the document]
jermainewang: There might be other functors, though currently only reducer here. The idea is to put device-dependent functor in this header (for example, the LDGReader functor in cuda folder).

python/dgl/kernel.py
ylfdq1118: I don't quite get this one...

lhs_mapping is for lhs, rhs_mapping is for rhs, out_mapping is for out. Whenever we want to access XXX, we need to go through XXX_mapping first. It's just one layer of indirection.

python/dgl/runtime/ir/executor.py
jermainewang: We did. Our copy_reduce and torch.spmm both use cusparse, but our backward is more efficient since we cache the reverse csr. Compared with scatter, the result in our blogpost applies. PyG essentially used torch's scatter_add.

python/dgl/backend/backend.py
ylfdq1118: Because the mapping for forward pass and backward pass can possibly be different. It's the same for nodes, but different for edges because when converting coo to csr and csc, the edge order is different.

python/dgl/nn/pytorch/softmax.py
ylfdq1118: I suppose so. Will remove and we benchmark the speed.

python/dgl/runtime/ir/executor.py
ylfdq1118: Not yet. Will do for sure.

python/dgl/runtime/scheduler.py
ylfdq1118: Think about performing recv on only a few nodes. If we use UDF reducer, even if these recv nodes are not consecutive (let's say node 5, 10, 20, 100), the output will just be a compact tensor of 4 rows. You can think of this as an implicit mapping from global id 5, 10, 20, 100 to local id 0, 1, 2, 3. 

Builtin function should result in the same behavior as UDF. But kernels think the entire graph. Therefore, we need a way to tell kernel that it needs to write the output to some local id space 0, 1, 2, 3 instead of the global id space. That's why we need the output map.

The output map is different for different APIs. Update_all needs all nodes, so local space == global space. But send_and_recv needs this out mapping. To unify, we have this out_map_creator and use it to create out_map lazily.

python/dgl/runtime/scheduler.py
ylfdq1118: Yes. Done

python/dgl/nodeflow.py
zheng-da: could you add comments to explain the context conversion? and explain to users that the output tensor is always in the CPU context.

ylfdq1118: Yeah. I wonder where I should add those comments. Actually, I discussed with @jermainewang about this yesterday and we agreed that we should let user know that graph itself has context and currently it's always on CPU. So user need to be aware that all graph query will should take CPU tensor as input and return CPU tensor as output.

jermainewang: I'd prefer not to be crystal clear about this since we are likely to improve this part in the near release cycle. I think it's ok for now as the behavior is the same as before (both the argument and return tensor for index are CPU only).

python/dgl/kernel.py
jermainewang: Added some description here. @BarclayII see if looks better for you.

python/dgl/backend/backend.py
```diff
@@ -842,6 +833,105 @@ def zerocopy_from_numpy(np_array):
...
+
+    Parameters
+    ----------
+    input : dgl.ndarray.NDArray
+
+    Returns
+    -------
+    Tensor
+    """
+    pass
```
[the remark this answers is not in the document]
ylfdq1118: zerocopy_from_dgl_ndarray and zerocopy_to_dgl_ndarray are both utility functions. The first copies a dgl ndarray to a framework-specific tensor, the second copies a framework-specific tensor to dgl ndarray

python/dgl/runtime/spmv.py
```diff
@@ -4,159 +4,126 @@
...
 
 from . import ir
 from .ir import var
 
-def analyze_v2v_spmv(graph, mfunc, rfunc):
-    """Analyze if SPMV from node space to node space can be applied.
+
+def gen_v2v_spmv_schedule(graph, mfunc, rfunc, src_frame, dst_frame,
+                          edge_frame, out, out_size, src_map=None,
+                          dst_map=None, edge_map=None, out_map=None):
```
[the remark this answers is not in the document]
ylfdq1118: That's very correct. The reason I have them here is 1) to make code general in case we need it; 2) improve the readability of the logic in message function where lhs and rhs is chosen (https://github.com/dmlc/dgl/blob/6720483d7eae701b51fb733578a0f1e51a75a58b/python/dgl/function/message.py#L53)

python/dgl/runtime/spmv.py
```diff
@@ -165,236 +132,86 @@ def build_adj_matrix_graph(graph):
...
-    return utils.CtxCachedObject(lambda ctx: F.copy_to(mat, ctx)), None
+    nbits = gidx.bits_needed()
+    forward_map = utils.to_nbits_int(eid[forward.tousertensor()], nbits)
+    backward_map = utils.to_nbits_int(eid[backward.tousertensor()], nbits)
+    forward_map = F.zerocopy_to_dgl_ndarray(forward_map)
+    backward_map = F.zerocopy_to_dgl_ndarray(backward_map)
+    edge_map = utils.CtxCachedObject(
+        lambda ctx: (nd.array(forward_map, ctx=ctx),
+                     nd.array(backward_map, ctx=ctx)))
+    return utils.CtxCachedObject(gidx.get_immutable_gidx), edge_map, nbits
```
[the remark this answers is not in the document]
ylfdq1118: Done

python/dgl/function/base.py
```diff
@@ -1,38 +1,25 @@
...
-class BuiltinFunction(object):
-    """Base builtin function class."""
-    @property
-    def name(self):
-        """Return the name of this builtin function."""
-        raise NotImplementedError
 
-class BundledFunction(object):
-    """A utility class that bundles multiple functions.
+class TargetCode(object):
```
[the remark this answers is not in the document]
ylfdq1118: Done

python/dgl/backend/mxnet/tensor.py
```diff
@@ -6,6 +6,14 @@
...
 import mxnet.ndarray as nd
 import numbers
+from ... import ndarray as dglnd
+from ... import kernel as K
+
+MX_VERSION = LooseVersion(mx.__version__)
+# After MXNet 1.5, empty tensors aren't supprted by default.
+# after we turn on the numpy compatible flag, MXNet supports empty NDArray.
+if MX_VERSION.version[0] == 1 and MX_VERSION.version[1] >= 5:
+    mx.set_np_compat(True)
```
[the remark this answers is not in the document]
ylfdq1118: Done

tests/compute/test_kernel.py
```diff
@@ -0,0 +1,235 @@
...
+        assert F.allclose(r1, r2)
+        assert(F.allclose(n_grad1, n_grad2))
+        assert(F.allclose(e_grad1, e_grad2))
+
+    _test('sum')
+    _test('sum', 'src')
+    _test('sum', 'edge')
+    _test('max')
+    _test('max', 'src')
+    _test('max', 'edge')
```
zheng-da: why do we need this test? `test_all_binary_builtins` doesn't cover this case?

tests/compute/test_specialization.py
```diff
@@ -418,18 +413,6 @@ def _afunc(nodes):
...
-    g.update_all([fn.src_mul_edge(src='h', edge='w1', out='m1'),
-                  fn.src_mul_edge(src='h', edge='w2', out='m2'),
-                  fn.src_mul_edge(src='h', edge='w1', out='m3')],
-                 [fn.sum(msg='m1', out='o1'),
-                  fn.sum(msg='m2', out='o2'),
-                  fn.max(msg='m3', out='o3')],
-                 _afunc)
-    assert F.allclose(o1, g.ndata.pop('o1'))
-    assert F.allclose(o2, g.ndata.pop('o2'))
-    assert F.allclose(o3, g.ndata.pop('o3'))
```
zheng-da: why is this test removed?

tests/compute/test_specialization.py
```diff
@@ -47,18 +47,18 @@ def apply_func(nodes):
...
         # update all with edge weights
         v1 = g.ndata[fld]
-        g.update_all(fn.src_mul_edge(src=fld, edge='e1', out='m'),
-                fn.sum(msg='m', out=fld), apply_func)
-        v2 = g.ndata[fld]
-        g.set_n_repr({fld : v1})
-        g.update_all(fn.src_mul_edge(src=fld, edge='e2', out='m'),
-                fn.sum(msg='m', out=fld), apply_func)
-        v3 = g.ndata[fld]
+        if len(g.ndata[fld].shape) == 1:
```
zheng-da: it seems the new builtin doesn't support some of the cases?

src/kernel/binary_reduce.cc
```diff
@@ -0,0 +1,541 @@
...
+// to be switched. For example, Add(kDst, kSrc) needs to be changed
+// to Add(kSrc, kDst).
+// This is because we only generate kernels for
+//  Add(kSrc, kDst), Add(kDst, kEdge), Add(kSrc, kDst)
+// to save compilation time.
+inline bool NeedSwitchOrder(const std::string& op,
+    binary_op::Target lhs, binary_op::Target rhs) {
+  CHECK_NE(lhs, rhs);
+  return (op == binary_op::kAdd || op == binary_op::kMul)
+    && lhs > rhs;
```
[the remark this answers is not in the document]
jermainewang: Add,Mul,Div,Sub  (see binary_reduce_common.h)

python/dgl/runtime/scheduler.py
```diff
@@ -746,26 +790,36 @@ def _gen_send_reduce(
...
+        unique(v) + 0deg nodes.
     uv_getter : callable
-        A function that returns a pair of var.IDX (u, v) for the triggered edges.
+        Function that returns a pair of var.IDX (u, v) for the triggered edges.
     adj_creator : callable
-        A function that returns the adjmat and the shuffle index.
-    inc_creator : callable
-        A function that returns the incmat and the shuffle index.
+        Function that returns the adjmat, edge order of csr matrix, and edge
+        order of transposed csr matrix
```
BarclayII: The third return of adj_creator() seems to be bit width instead of transposed CSR.

tests/compute/test_specialization.py
jermainewang: DGL should be able to handle all f1*e1, f1*e2, f2*e1, f2*e2.

tests/compute/test_specialization.py
jermainewang: Because there is no fallback to degree-bucketing anymore, the test is the same as testing multiple builtin functions (e.g. L404-L409)

python/dgl/runtime/scheduler.py
```diff
@@ -837,4 +916,75 @@ def _mfunc_wrapper(src_data, edge_data, dst_data):
...
+    Returns
+    -------
+    old_to_new : CtxCachedObject
+        The mapping from old id to new id. It is a vector of length MAX(x).
+        One can use advanced indexing to convert an old id tensor to a
+        new id tensor: new_id = old_to_new[old_id]
+    """
+    x = idx.tousertensor()
+    map_len = int(F.asnumpy(F.max(x, dim=0))) + 1
+    old_to_new = F.zeros((map_len,), dtype=F.int64, ctx=F.cpu())
```
BarclayII: Can we fill the n/a's with -1 since 0 is a valid value?  It's hard to debug otherwise.

python/dgl/kernel.py
BarclayII: I see.  So essentially `xxx_mapping` are actually slicing the lhs/rhs/output data to match the nodes/edges of `graph`.  And `graph` could be either the entire original graph or a part of it as in `send_and_recv`.

python/dgl/nn/pytorch/softmax.py
BarclayII: Do we still need to write the backward pass for edge softmax?

jermainewang: Yes, so we could avoid backward the max operation for example.

jermainewang:
I'm done with the review and approved.

tests/compute/test_kernel.py
ylfdq1118: This one is an old test. The 'edge', 'src' is specifying that broadcasting should be tested on edge or src feature. We can have all comprehensive tests (including broadcasting dimension) for all builtin combinations in test_all_binary_builtins. It's just going to take longer for testing to finish. But yeah it's worth it. Will do.

BarclayII:
Let me refine the description in `kernel.py` and I will be done.

ylfdq1118:
I've finished reviewing C++ code. Looks good.

jermainewang:
Cool! Thanks everyone for this huge change!

Issue #419, opened 2019-02-28
[Usability] to_undirected_graph method when constructing DGL graph using edgelist


HQ01:
## 🚀 Feature

Add a simple method to allow users to construct an "undirected" (which is essentially bi-directed) DGL graph from edge list when constructing the graph.
## Motivation
Some datasets (like ENZYMES) store graphs as edges list, and default the graph type to undirected. It would be convenient for users to initialize a DGL graph from an edgelist of undirected graph.
 

## Alternatives

Currently one could first construct a nx graph, then cast it to DGL graph. Or one could first double the edge list and make it bi-directional first. 
## Pitch

a = dgl.DGLGraph()
a.add_nodes(3)
edgeList = [(0,1),(1,0),(2,1)]
a.add_edges(edgeList)
a.to_undirected()
print(a.edges)
result: [(0,1),(1,0),(2,1),(1,2)]
## Additional context
Circumstances exists when users may not have prior knowledge of networkx, or the edge list is not purely undirected. If let user to handle it, it could be the case that 1) the constructed DGL graph is not "undirected", or it's a multigraph.

jermainewang:
My question is if the graph already contains features. What will happen if `to_undirected` is called?

Also, suggest `to_bidirected`.

HQ01:
If edges already have features, then one thing we can do is to simply copy the edge feature to the reverse direction, with the option of letting users specify them (reverse sign, different edge feature, etc.)

Yeah I agree, to_bidirected sounds better!

jermainewang:
Sounds good. Two more questions:

* Would you prefer returning a new graph? or perform mutation inplace?
* As you've noticed, the `bi_directed` graph could become a multi-graph if edges are added accidentally. This is related to #340 by @BarclayII . The issue with that PR is it slows down the graph construction time a lot for extra duplication check. I've been thinking about this question a lot. My current intention is to (a) no duplication check during mutation; (b) provide a utility function called `to_simple_graph` that will prune out duplication. As a result, the extra time/space is confined in this API and is called by user explicitly. What do you guys think?

HQ01:
1) I would prefer mutation in place since the most common use case of this will be in data-preprocessing and we don't need the original graph.

2) If I understand it correctly -- in this design, `to_bidirected` will simply double the number of edges (even if the original graph is already a multi-graph), and `to_simple_graph` will prune any graph with multiple edges between pairs of nodes to simple graph?

jermainewang:
[...]

PR #598, opened 2019-06-02
[Feature] Add API to convert graph to bidirected graph

mufeili:
## Description
See issue #419 . For simplicity, we only handle the graph topology and leave feature part to the users.

## Checklist
- [x] Code is well-documented

## Changes
- [x] `dgl.to_bidirected` API


VoVAllen:
1. Shall we support this operation on immutable graph?
2. What will happen if graph is a multi graph?

mufeili:
> 1. Shall we support this operation on immutable graph?

This will return a new graph. For immutable graphs, maybe we can check first if it is readonly and then return a readonly one if so?

> 2. What will happen if graph is a multi graph?

If a graph originally has m edges for i -> j and n edges for j -> i, the new graph will have max(m, n) edges for both i -> j and j -> i. So you will still get a multigraph.

yzh119:
I think such operations should only be applicable to mutable and single graphs.

mufeili:
@VoVAllen @yzh119 In fact currently this does not support immutable graphs . Probably I should also check whether this is a batched graph.

jermainewang:
Since we maintained many different types of graphs in python (mutable, immutable, subgraph, batched_graph, nodeflow), it'll become quite complicated if we want to cover all the cases. Here is a proposal to keep the APIs in `dgl.transform` consistent.
1) The default behavior only transforms graph structure. Additional arguments could be specified to transform features as well.
2) The API is functional, aka a new graph structure is returned.
3) The returned graph is always an immutable graph. Thus, if the transformation happens to be identical and the input graph is also an immutable graph, the input graph could be returned.

Let me know your thoughts.

mufeili:
@jermainewang I think the current implementation already matches the first two points. I'm a bit uncertain about the third point. Is it possible to convert an immutable graph to a mutable graph now? What can be pros and cons with the current system to return a normal graph vs an immutable graph?

VoVAllen:
@jermainewang I'm a bit confused about the third point. The returned graph is identical or not cannot be known beforehand. Therefore user have to do extra check whether this is the input graph after transform.  Maybe always return a new graph and provide some API like `g.is_identical(g2)` can help here?

jermainewang:
> @jermainewang I think the current implementation already matches the first two points. I'm a bit uncertain about the third point. Is it possible to convert an immutable graph to a mutable graph now? What can be pros and cons with the current system to return a normal graph vs an immutable graph?

Immutable graph should be more efficient for most of the operations. The problem is no mutation will be allowed afterwards.

> The returned graph is identical or not cannot be known beforehand. Therefore after transform user have to do extra check whether this is the input graph. It's weird.

The third point tries to ease the implementation. For example, when implementing `to_simple` transformation, if I know that the input graph is already a simple graph (by looking at the flag), I'm safe to return the input graph as the output graph directly without a deep copy.

mufeili:
@jermainewang What if we first check whether the input graph is immutable and then handle the two cases separately?

jermainewang:
> @jermainewang What if we first check whether the input graph is immutable and then handle the two cases separately?

That's fine. You could handle it that way.

mufeili:
@jermainewang The API now returns an immutable graph by default and allows returning mutable graphs if needed.

src/graph/graph_apis.cc
```diff
@@ -572,4 +572,24 @@ DGL_REGISTER_GLOBAL("transform._CAPI_DGLToSimpleGraph")
...
+    *bgptr = GraphOp::ToBidirectedMutableGraph(gptr);
+    GraphHandle bghandle = bgptr;
+    *rv = bghandle;
+  });
+
+DGL_REGISTER_GLOBAL("transform._CAPI_DGLToBidirectedImmutableGraph")
+.set_body([] (DGLArgs args, DGLRetValue* rv) {
+    GraphHandle ghandle = args[0];
+    const GraphInterface *ptr = static_cast<const GraphInterface *>(ghandle);
+    const Graph* gptr = static_cast<const Graph*>(ptr);
```
VoVAllen: gptr can be null if graph is immutable graph

mufeili: Should be fixed now.

src/graph/graph_op.cc
```diff
@@ -315,4 +315,74 @@ ImmutableGraph GraphOp::ToSimpleGraph(const GraphInterface* graph) {
...
+    }
+  }
+
+  Graph bg;
+  bg.AddVertices(g->NumVertices());
+  for (dgl_id_t u = 0; u < g->NumVertices(); ++u) {
+    for (dgl_id_t v = u; v < g->NumVertices(); ++v) {
+      const auto new_n_e = std::max(n_e[u][v], n_e[v][u]);
+      if (new_n_e > 0) {
+        IdArray us = IdArray::Empty({new_n_e}, DLDataType{kDLInt, 64, 1}, DLContext{kDLCPU, 0});
```
jermainewang: Try use the new routines in `include/dgl/array.h`.

src/graph/graph_op.cc
```diff
@@ -315,4 +315,74 @@ ImmutableGraph GraphOp::ToSimpleGraph(const GraphInterface* graph) {
...
+      const auto new_n_e = std::max(n_e[u][v], n_e[v][u]);
+      for (size_t i = 0; i < new_n_e; ++i) {
+        indices.push_back(v);
+      }
+    }
+    indptr[u+1] = indices.size();
+  }
+  CSRPtr csr(new CSR(g->NumVertices(), indices.size(),
+        indptr.begin(), indices.begin(), RangeIter(0), g->IsMultigraph()));
+  return ImmutableGraph(csr);
```
jermainewang: The immutable graph supports construction from a COO structure. This should be easier and more efficient.

tests/compute/test_transform.py
```diff
@@ -95,9 +95,27 @@ def test_simple_graph():
     eset = set(zip(list(F.asnumpy(src)), list(F.asnumpy(dst))))
     assert eset == set(elist)
 
+def test_bidirected_graph():
+    def _test(in_readonly, out_readonly):
+        elist = [(0, 0), (0, 1), (0, 1), (1, 0), (1, 1), (2, 1), (2, 2), (2, 2)]
+        g = dgl.DGLGraph(elist, readonly=in_readonly)
+        elist.append((1, 2))
+        elist = set(elist)
+        big = dgl.to_bidirected(g, out_readonly)
```
jermainewang: It looks like your `to_bidirected` implicitly convert a multigraph to a simple graph, which might not be the expected behavior. It's generally better to let each operator serve only one purpose, and allow the power of combination to cover rich scenarios.

src/graph/graph_op.cc
```diff
@@ -315,4 +315,74 @@ ImmutableGraph GraphOp::ToSimpleGraph(const GraphInterface* graph) {
   return ImmutableGraph(csr);
 }
 
+Graph GraphOp::ToBidirectedMutableGraph(const GraphInterface* g) {
+  std::unordered_map<int, std::unordered_map<int, int>> n_e;
+  for (dgl_id_t u = 0; u < g->NumVertices(); ++u) {
+    for (const dgl_id_t v : g->SuccVec(u)) {
+      n_e[u][v]++;
+    }
+  }
```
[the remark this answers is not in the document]
mufeili: If we have `m` edges for `i->j` and `n` edges for `j->i`, then the current implementation will give `max(m, n)` edges for both direction in the returned graph. The hashmap is just used for a first pass edge counting.

tests/compute/test_transform.py
mufeili: No, the returned graph is still a multi-graph.

src/graph/graph_op.cc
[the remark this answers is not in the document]
mufeili: If you simply make `m` edges for `j->i` and `n` edges for `i->j`, I think this is what `reverse` intends to do.

src/graph/graph_op.cc
mufeili: Done.

src/graph/graph_op.cc
mufeili: Done.

jermainewang:
Chatted with @mufeili offline. @mufeili is right about the semantics.

python/dgl/transform.py
```diff
@@ -124,4 +124,28 @@ def to_simple_graph(g):
     newgidx = GraphIndex(_CAPI_DGLToSimpleGraph(g._graph.handle))
     return DGLGraph(newgidx, readonly=True)
 
+def to_bidirected(g, readonly=True):
+    """Convert the graph to a bidirected graph.
+
+    The function generates a new graph with no node/edge feature.
+    If g has m edges for i->j and n edges for j->i, then the
+    returned graph will have max(m, n) edges for both i->j and j->i.
```
jermainewang: Give some examples in the comment?

jermainewang:
Approved! One last little thing, could you add some examples in the comment about two cases (one simple graph, one multigraph). Otherwise, LGTM and you could go ahead merge it afterwards.

mufeili:
@jermainewang Examples have been added.

Issue #467, opened 2019-03-28
[Discussion] Graph backend and Boost graph library


Wimmerer:
## Feature 
Adding support for a graph backend like Boost in place of or in addition to the current custom.

## Motivations
The current graph backend doesn't support a number of important operations, importantly removing nodes and edges.

Integration with third party libraries may also be easier, in particular mutating the graph from external programs might be easier with a standard library.

Easier to support more graph types in the future and any of the already implemented algoritms in Boost which could be useful. 

Potentially better performance, but the lightweight implementation already here may be faster. 

Could support calling Boost based graph algorithms from DGL to integrate our own C++ code.

## Alternatives
Keeping and iterating on the current backend.

Choosing another library, maybe Graph-Tools. 

This is definitely a major undertaking, and I could lend my limited skills, but it would be quite beneficial for a number of my use cases.

jermainewang:
Hi @Pangurbon , thank you very much for your interest and proposal. Really appreciate it.

I think the major question is what kinds of graph operations/algorithms are urgently or likely needed. The final technical solution will likely depend on the requirements. Also, user can always fallback to `networkx` if some operations are missing. However, such fallback is not cheap, thus we want to have efficient implementation for important/common operations.

From my side, I see following factors to be considered:
* What kinds of models/application require such operations?
  - For example, we currently don't find any models/applications that remove nodes/edges dynamically.
  - Traversal algorithms are common so we implement them in DGL (namely topological traversal, DFS, BFS). As a result, supporting more traversal algorithms might be useful for future models.
  - Graph generators such as many random graph models, stochastic block model, tree generator, bipartite generator, etc are useful for applications that can utilize synthetic graphs.
  - Graph transformations such as line graph, reverse graph, power graph are also seen in some models.
  - More complicated graph analytic algorithms like connected component, triangle counting haven't been explored by graph deep learning. We don't know whether people are interested in combining them with deep learning.
* What kind of hardware?
  - Currently, we only have CPU graph, but GPU graph looks very important and there are already some efforts in this line of work (e.g. cugraph).
* Special graph type?
  - In terms of data structure, is CSR and adjlist enough?
  - In terms of upper-level semantics, do we need special treatment for trees, cliques, bipartite graphs?

I'd like to hear more about your use cases. I think we could use this discussion to plot a clearer roadmap for this support.

Pangurbon:
>I think the major question is what kinds of graph operations/algorithms are urgently or likely needed

[...]

PR #599, opened 2019-06-02
[Feature] Delete nodes/edges.

yzh119:
## Description
This is an in-efficient implementation of removing nodes/edges in `DGLGraph`. As in issue #467 .

## Checklist
- [x] Code is well-documented

## Changes
In this PR, I use node/edge subgraph to implement nodes/edges removal.
However, the behavior of edge subgraph is different from edge removal. Because in edge subgraph, nodes without in/out edges would be removed, but it's node the expected behavior of edge removal.

One solution is to add an argument `preserve_nodes` in `EdgeSubgraph` function to indicate whether to preserve nodes in the graph.
Do you have any suggestions?


python/dgl/graph.py
```diff
@@ -1101,6 +1101,64 @@ def add_edges(self, u, v, data=None):
...
+            self._node_frame = FrameRef(Frame(self._node_frame[sgi.induced_nodes]))
+        else:
+            self._node_frame = FrameRef(self._node_frame, sgi.induced_nodes)
+
+        if isinstance(self._edge_frame, FrameRef):
+            self._edge_frame = FrameRef(Frame(self._edge_frame[sgi.induced_edges]))
+        else:
+            self._edge_frame = FrameRef(self._edge_frame, sgi.induced_edges)
+
+        # FIXME(zihao): the following impl is not efficient, but directly reusing the handle of subgraph index would
```
yzh119: @jermainewang and @zheng-da , could you please have a look at this? After discussion with @BarclayII we think the problem resides in "GraphInterface.Reset".

yzh119: @jermainewang and @zheng-da could you please have a look at this, after discussion with @BarclayII we think the problem resides in "GraphInterface.Reset".

src/graph/graph_apis.cc
```diff
@@ -452,7 +452,8 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLGraphEdgeSubgraph")
     GraphHandle ghandle = args[0];
     const GraphInterface *gptr = static_cast<GraphInterface*>(ghandle);
     const IdArray eids = args[1];
-    *rv = ConvertSubgraphToPackedFunc(gptr->EdgeSubgraph(eids));
+    bool preserve_nodes = args.size() > 2 ? args[2] : false;
```
zheng-da: the number of arguments can change? from python, you always pass 3 arguments.

src/graph/immutable_graph.cc
```diff
@@ -640,7 +641,8 @@ Subgraph ImmutableGraph::VertexSubgraph(IdArray vids) const {
                   sg.induced_vertices, sg.induced_edges};
 }
 
-Subgraph ImmutableGraph::EdgeSubgraph(IdArray eids) const {
+Subgraph ImmutableGraph::EdgeSubgraph(IdArray eids, bool preserve_nodes) const {
+  // TODO(zihao): to handle the case of preserve_nodes=true.
```
zheng-da: maybe you can add an assertion here and above.

tests/compute/test_removal.py
```diff
@@ -0,0 +1,149 @@
...
+    g.del_edges(range(10, 20))
+    assert g.number_of_nodes() == 10
+    assert g.number_of_edges() == 48
+
+if __name__ == '__main__':
+    test_node_removal()
+    test_edge_removal()
+    test_multigraph_node_removal()
+    test_multigraph_edge_removal()
+    test_node_and_edge_removal()
```
zheng-da: please check the data in the node frame and edge frame as well.

python/dgl/graph.py
jermainewang: You cannot let two GraphIndex object share the same internal handle. You will get double free problem. The reason is that there is no refcount for the handle (which is essentially a pointer to a C object). We rely on python's native object refcount to delete the handle properly. Checkout L37-40 in `graph_index.py`. The handle is deleted in the `__del__` function, which will be called **only** when the GraphIndex object is destroyed. Python makes sure that if a GraphIndex object is shallow-copied to another variable, only one `__del__` function call will be made.

If you create a new GraphIndex and reuse the internal handle of an existing GraphIndex object, two `__del__` function will delete the same handle.

jermainewang: My solution:
* Change `graph_index.SubgraphIndex` to **not** inherit `GraphIndex`. Actually, our C++ data structure is like this. The subgraph index only contains the graph structure, the parent graph, induced nodes and edges. As a result, the object returned by `node_subgraph` is not necessarily readonly graph or not.
* In `del_nodes/edges` functions, directly set the returned graph structure as the `_graph` field in the `DGLGraph` object.

python/dgl/graph.py
```diff
@@ -1101,6 +1101,57 @@ def add_edges(self, u, v, data=None):
             self._msg_index = self._msg_index.append_zeros(num)
         self._msg_frame.add_rows(num)
 
+    def del_nodes(self, vids):
```
jermainewang: Use `remove_nodes` and `remove_edges` to be consistent with networkx.

python/dgl/graph.py
```diff
@@ -1101,6 +1101,57 @@ def add_edges(self, u, v, data=None):
...
+            self._node_frame = FrameRef(self._node_frame, sgi.induced_nodes)
+
+        if isinstance(self._edge_frame, FrameRef):
+            self._edge_frame = FrameRef(Frame(self._edge_frame[sgi.induced_edges]))
+        else:
+            self._edge_frame = FrameRef(self._edge_frame, sgi.induced_edges)
+
+        self._graph = sgi.graph
+
+    def del_edges(self, eids):
```
jermainewang: same

python/dgl/graph_index.py
```diff
@@ -536,23 +537,26 @@ def node_subgraphs(self, vs_arr):
...
-    def edge_subgraph(self, e):
+    def edge_subgraph(self, e, preserve_nodes=False):
         """Return the induced edge subgraph.
 
         Parameters
         ----------
         e : utils.Index
             The edges.
+        preserve_nodes: bool
+            Indicates whether to preserve all nodes or not. (when set to false, all nodes without edge connected to it would be removed.)
```
jermainewang: If true, keep the nodes which have no edge connected in the subgraph.
It is equivalent to a sparser graph that contains only the given edges.

python/dgl/subgraph.py
```diff
@@ -32,35 +32,31 @@ class DGLSubGraph(DGLGraph):
...
+        Internal subgraph data structure.
     shared : bool, optional
         Whether the subgraph shares node/edge features with the parent graph.
     """
-    def __init__(self, parent, parent_nid, parent_eid, graph_idx, shared=False):
-        super(DGLSubGraph, self).__init__(graph_data=graph_idx,
-                                          readonly=graph_idx.is_readonly())
+    def __init__(self, parent, sgi, shared=False):
+        super(DGLSubGraph, self).__init__(graph_data=sgi.graph,
+                                          readonly=sgi.graph.is_readonly())
```
jermainewang: `readonly=True`

src/graph/graph.cc
```diff
@@ -467,37 +467,56 @@ Subgraph Graph::VertexSubgraph(IdArray vids) const {
...
+        {static_cast<int64_t>(nodes.size())}, eids->dtype, eids->ctx);
+    std::copy(nodes.begin(), nodes.end(), static_cast<int64_t*>(rst.induced_vertices->data));
+  } else {
+    rst.graph = std::make_shared<Graph>(IsMultigraph());
+    rst.induced_edges = eids;
+    rst.graph->AddVertices(NumVertices());
+
+    for (int64_t i = 0; i < len; ++i) {
+      dgl_id_t src_id = all_edges_src_[eid_data[i]];
+      dgl_id_t dst_id = all_edges_dst_[eid_data[i]];
```
jermainewang: const

src/graph/graph.cc
```diff
@@ -467,37 +467,56 @@ Subgraph Graph::VertexSubgraph(IdArray vids) const {
...
-    dgl_id_t dst_id = all_edges_dst_[eid_data[i]];
-    rst.graph->AddEdge(oldv2newv[src_id], oldv2newv[dst_id]);
-  }
+    rst.graph = std::make_shared<Graph>(IsMultigraph());
+    rst.induced_edges = eids;
+    rst.graph->AddVertices(nodes.size());
+
+    for (int64_t i = 0; i < len; ++i) {
+      dgl_id_t src_id = all_edges_src_[eid_data[i]];
+      dgl_id_t dst_id = all_edges_dst_[eid_data[i]];
```
jermainewang: const

src/graph/graph.cc
```diff
@@ -467,37 +467,56 @@ Subgraph Graph::VertexSubgraph(IdArray vids) const {
...
 
-  rst.induced_vertices = IdArray::Empty(
-      {static_cast<int64_t>(nodes.size())}, eids->dtype, eids->ctx);
-  std::copy(nodes.begin(), nodes.end(), static_cast<int64_t*>(rst.induced_vertices->data));
+    for (int64_t i = 0; i < NumVertices(); ++i)
+      nodes.push_back(i);
+
+    rst.induced_vertices = IdArray::Empty(
+        {static_cast<int64_t>(nodes.size())}, eids->dtype, eids->ctx);
+    std::copy(nodes.begin(), nodes.end(), static_cast<int64_t*>(rst.induced_vertices->data));
```
jermainewang: Could you help refactor this function by using the new IdArray routines defined [here](https://github.com/dmlc/dgl/blob/master/include/dgl/array.h)? such as NewIdArray, VecToIdArray, etc. Thx!

src/graph/immutable_graph.cc
```diff
@@ -6,6 +6,7 @@
 
 #include <dgl/immutable_graph.h>
 #include <string.h>
+#include <cassert>
```
jermainewang: remove

src/graph/immutable_graph.cc
```diff
@@ -549,7 +550,8 @@ COO::EdgeArray COO::Edges(const std::string &order) const {
   return EdgeArray{src_, dst_, rst_eid};
 }
 
-Subgraph COO::EdgeSubgraph(IdArray eids) const {
+Subgraph COO::EdgeSubgraph(IdArray eids, bool preserve_nodes) const {
+  assert(!preserve_nodes); // TODO(zihao): handle the case of preserve_nodes == true
```
jermainewang: Could you help implement this? It shouldn't be too hard. Just create a COO with the given src,dst id and the same number of nodes.

src/graph/immutable_graph.cc
```diff
@@ -696,7 +698,8 @@ Subgraph ImmutableGraph::VertexSubgraph(IdArray vids) const {
                   sg.induced_vertices, sg.induced_edges};
 }
 
-Subgraph ImmutableGraph::EdgeSubgraph(IdArray eids) const {
+Subgraph ImmutableGraph::EdgeSubgraph(IdArray eids, bool preserve_nodes) const {
+  assert(!preserve_nodes); // TODO(zihao): handle the case of preserve_nodes == true
```
jermainewang: After implemented COO::EdgeSubgraph, pls remove this.

yzh119:
@jermainewang , I've updated the code according to your review.
As for the refactor of IdArray routines, I found that many parts in the code follow the old style. How about leaving it to the next PR?

jermainewang:
Cool, LGTM!

PR #616, opened 2019-06-06
[BUGFIX] fix sampler.

zheng-da:
## Description
When our ImmutableGraph is constructed, it may not have one of the CSRs. For example, when we construct ImmutableGraph from a csr input, we only have out-csr. When in-csr is required, it is constructed when required. However, we can't reconstruct the in-csr inside OpenMP. If we do so, multiple threads try to construct the in-csr in parallel and override the one created by another thread. That's why we experience segfault in the sampling test. The right way is to construct the in-csr in advance before we enter OpenMP.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/src/graph/sampler.cc
+++ b/src/graph/sampler.cc
@@ -700,6 +700,18 @@ NodeFlow SamplerOp::LayerUniformSample(const ImmutableGraph *graph,
   return nf;
 }
 
+void build_csr(const ImmutableGraph &g, const std::string neigh_type) {
+  if (neigh_type == "in") {
+    auto csr = g.GetInCSR();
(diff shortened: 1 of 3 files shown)


src/graph/sampler.cc
```diff
@@ -700,6 +700,18 @@ NodeFlow SamplerOp::LayerUniformSample(const ImmutableGraph *graph,
   return nf;
 }
 
+void build_csr(const ImmutableGraph &g, const std::string neigh_type) {
```
jermainewang: `BuildCsr` <- `build_csr`

PR #619, opened 2019-06-07
[Build] Support older CMake & OpenMP toggle

BarclayII:
* Lowered the required CMake version from 3.9 to 3.5 since Ubuntu 16.04 ships with CMake 3.5 only.
* `USE_OPENMP` now toggles OpenMP compilation.
* Fixes Windows builds where OpenMP is never turned on.
* Now if users wish to select their own CUDA installation path, they have to do it via `CUDA_TOOLKIT_ROOT_DIR` as per `FindCUDA` in CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.9)
+cmake_minimum_required(VERSION 3.5)
 ########################################
 # Borrowed and adapted from TVM project
 ########################################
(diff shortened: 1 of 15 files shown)


jermainewang:
> * Lowered the required CMake version from 3.9 to 3.5 since Ubuntu 16.04 ships with CMake 3.5 only.

@szha do you remember what's the problem of using old cmake for building cuda programs? I saw a lot of `FIRST_CUDA`, `OLD_CUDA` macro in mxnet project.

PR #634, opened 2019-06-10
[BUGFIX] copy graph index to shared memory.

zheng-da:
## Description
If users provide a constructed graph index to SharedMemoryStoreServer, the graph index may not be in shared memory. We have to copy it to shared memory if it's not.

In addition, this PR gives shared memory for in-CSR and out-CSR different names. This is useful if a shared-memory graph index requires both in-CSR and out-CSR.

## Checklist
- [ ] Code is well-documented


include/dgl/immutable_graph.h
```diff
@@ -889,6 +914,9 @@ class ImmutableGraph: public GraphInterface {
     if (!in_csr_) {
       if (out_csr_) {
         const_cast<ImmutableGraph*>(this)->in_csr_ = out_csr_->Transpose();
+        if (out_csr_->IsSharedMem())
+          LOG(WARNING) << "We just construct an in-CSR from a shared-memory out CSR. "
+                       << "It may dramatically increase memory consumption.";
```
jermainewang: What does this mean? Is the `in_csr` also in shared memory? Is there any solution to this warning?

python/dgl/contrib/graph_store.py
```diff
@@ -318,7 +318,12 @@ class SharedMemoryStoreServer(object):
     """
     def __init__(self, graph_data, edge_dir, graph_name, multigraph, num_workers, port):
         self.server = None
-        if isinstance(graph_data, (GraphIndex, DGLGraph)):
+        if isinstance(graph_data, GraphIndex):
+            graph_data = graph_data.copyto_shared_mem(edge_dir, _get_graph_path(graph_name))
```
jermainewang: We probably should make shared memory as a special device type (similar to kDLCPUPinned in current DLPack). I created a issue in DLPack (https://github.com/dmlc/dlpack/issues/41) for this. This is definitely sth after this release though.

python/dgl/contrib/graph_store.py
```diff
@@ -318,7 +318,12 @@ class SharedMemoryStoreServer(object):
...
+            self._graph = DGLGraph(graph_data, multigraph=multigraph, readonly=True)
+        elif isinstance(graph_data, DGLGraph):
+            print(graph_data._graph.number_of_nodes())
```
jermainewang: remove

jermainewang:
Could you briefly update the PR description about what is the bug?

include/dgl/immutable_graph.h
zheng-da: I add this warning in case this happens. We could make sure the newly constructed csr is also in shared memory.

python/dgl/contrib/graph_store.py
zheng-da: yes, I agree.

Issue #637, opened 2019-06-10
The GPU usage keep growing as the training is going, if graph is passed into the model every time


yangysc:
## 🐛 Bug

Hello, I'm using the GAT model from the dgl library. But I made a small modification, since my input graph is changing all the time, so I do not store a graph in the model (I commented the code `self.g=g`), instead that I **treat the graph and feature tensor as the input to the model** every time.  Then the problem occured: **the GPU usage is keeping growing as the training is going.**

```
Processes:                                                           GPU Memory   
 GPU  PID         Type   Process name                                Usage   

   0     19583      C   ...noone/anaconda3/envs/tf_test/bin/python   697MiB
   0     19731      C   ...noone/anaconda3/envs/tf_test/bin/python  1159MiB 
   0     19583      C   ...noone/anaconda3/envs/tf_test/bin/python  1459MiB 
   0     19731      C   ...noone/anaconda3/envs/tf_test/bin/python  1929MiB 
   0     19731      C   ...noone/anaconda3/envs/tf_test/bin/python  2297MiB
   0     19731      C   ...noone/anaconda3/envs/tf_test/bin/python  6023MiB
```

**Is this because some data still exist in the GPU memory, and begin to accumulate whenever I call the model** like `probs = net(graph, feature)`?  If this is true, what information should I delete when the model is called?
Thanks in advance!
  
## To Reproduce

1. Install the latest dgl.
1. Download the files I uploaded  in [here](https://github.com/yangysc/Test_dgl_GAT/tree/master).
1. Run the following command to train
 `python3 train.py --dataset=cora --gpu=0 --epochs=20000`

1. Watch the GPU usage every 3 seconds.
`watch -n 3 nvidia-smi`

``` python
# gat model
"""
Graph Attention Networks in DGL using SPMV optimization.
References
----------
Paper: https://arxiv.org/abs/1710.10903
Author's code: https://github.com/PetarV-/GAT
Pytorch implementation: https://github.com/Diego999/pyGAT
"""

import torch
import torch.nn as nn
import dgl.function as fn
from dgl.nn.pytorch import edge_softmax

class GraphAttention(nn.Module):
    def __init__(self,
                 in_dim,
                 out_dim,
                 num_heads,
                 feat_drop,
                 attn_drop,
                 alpha,
                 residual=False):
        super(GraphAttention, self).__init__()
        # self.g = g
        self.num_heads = num_heads
[...]

PR #643, opened 2019-06-10
[BugFix] Fix memory leak in DGL edge softmax module

ylfdq1118:
## Description
Current DGL operators leak memory, and this happens for both PyTorch (#637 ) and MXNet backend.

For PyTorch, the cause is we used a separate cache instead of PyTorch's `ctx.save_for_backward` API to save tensors from forward pass but forgot to explicitly clear the cache. And the reason for not using `ctx.save_for_backward` API is we need to save non-tensor type data. Similar problem was mentioned in [this thread](https://discuss.pytorch.org/t/when-should-you-save-for-backward-vs-storing-in-ctx/6522).

For MXNet, the interesting thing is, even if we use its standard `save_for_backward` API, there is still memory leak. 

The Solution is simply explicitly clearing the cache.

## Checklist
- [x] Code is well-documented


SunderlandAJ-1130:
Hello, this problem still exists in DGL 0.6.1 with mxnet as backend. Could you please tell me how to fix it. Thanks!

PR #658, opened 2019-06-14
[Example] Add reddit example for SGC. 

Tiiiger:
provide an example of training SGC on the Reddit dataset.


yzh119:
LGTM, thanks! @Tiiiger

PR #679, opened 2019-06-24
[Network] Refactoring Communicator

aksnzhy:
## Description

1. implemented a zero-copy message queue to avoid large memory copy during send/recv.
2. Implemented non-blocking send that returns immediately.
3. Implemented multi-thread send/recv that maximal the throughput network communication.
4. Implemented multi-part send/recv to avoid copying ndarray data in serialization by meta-data pre-sending.
5. More general API for MPI, which will be used in GPU-direct communication in the future.

## Checklist
- [x] Code is well-documented


src/graph/network/common.h
```diff
@@ -0,0 +1,132 @@
...
+//
+//   substrings.size() == 3
+//   substrings[0] == "apple"
+//   substrings[1] == "orange"
+//   substrings[2] == "banana"
+//------------------------------------------------------------------------------
+
+void SplitStringUsing(const std::string& full,
+                      const char* delim,
+                      std::vector<std::string>* result);
```
jermainewang: Check out functions in [dmlc-core/strtonum.h](https://github.com/dmlc/dmlc-core/blob/master/include/dmlc/strtonum.h) to see whether there are sth useful.

src/graph/network/socket_communicator.cc
```diff
@@ -18,14 +19,29 @@
...
+/////////////////////////////////////// SocketSender ///////////////////////////////////////////
+
+void SocketSender::AddReceiver(const char* addr, int recv_id) {
+  std::vector<std::string> substring;
+  SplitStringUsing(addr, "//", &substring);
+  // Check address format
+  if (strcmp(substring[0].c_str(), "socket:") != 0 || substring.size() != 2) {
+    LOG(FATAL) << "Incorrect address format:" << addr 
+               << " Please provide right address format, "
+               << "e.g, 'ip://127.0.0.1:50051'.";
```
jermainewang: Are you using "socket://..." ?

src/graph/network/socket_communicator.cc
```diff
@@ -93,23 +109,28 @@ void SocketSender::Finalize() {
...
                           int queue_size) {
   CHECK_GE(num_sender, 1);
   CHECK_GT(queue_size, 0);
+  std::vector<std::string> substring;
+  SplitStringUsing(addr, "//", &substring);
+  // Check address format
+  if (strcmp(substring[0].c_str(), "socket:") != 0 || substring.size() != 2) {
+    LOG(FATAL) << "Incorrect address format:" << addr 
+               << " Please provide right address format, "
+               << "e.g, 'ip://127.0.0.1:50051'.";
```
jermainewang: same

src/graph/network.cc
```diff
@@ -8,19 +8,29 @@
...
 using dgl::runtime::DGLRetValue;
 using dgl::runtime::PackedFunc;
 using dgl::runtime::NDArray;
+using dgl::network::StringPrintf;
 
 namespace dgl {
 namespace network {
 
-// Wrapper for Send api
+static std::unordered_map<void*, char*> MemoryPool;
```
jermainewang: Check out memory pool implementation in [dmlc-core/memory.h](https://github.com/dmlc/dmlc-core/blob/master/include/dmlc/memory.h). Can we reuse that?

zheng-da: can we have a thread-safe memory pool?

src/graph/network.cc
```diff
@@ -44,13 +54,26 @@ static void RecvData(network::Receiver* receiver,
...
+DGL_REGISTER_GLOBAL("network._CAPI_DGLReceiverCreate")
+.set_body([] (DGLArgs args, DGLRetValue* rv) {
+    network::Receiver* receiver = new network::SocketReceiver();
+    CommunicatorHandle chandle = static_cast<CommunicatorHandle>(receiver);
+    try {
+      char* buffer = new char[kMaxBufferSize];
+      MemoryPool[chandle] = buffer;
+    } catch (const std::bad_alloc&) {
+      LOG(FATAL) << "Not enough memory for receiver buffer: " << kMaxBufferSize;
+    }
```
zheng-da: did you just move the location of the code?

src/graph/network/communicator.h
```diff
@@ -16,7 +16,7 @@ namespace network {
  *
  * Sender is an abstract class that defines a set of APIs for sending 
  * binary data over network. It can be implemented by different underlying 
- * networking libraries such TCP socket and ZMQ. One Sender can connect to 
+ * networking libraries such TCP socket and MPI. One Sender can connect to
```
zheng-da: Why do we prefer MPI over ZMQ?

zheng-da:
Is it right that the main modification is how to pass address?

zheng-da:
I think more importantly, what we need to address is how to send and recv data? Should we send data asynchronously? Should we send and receive data with multiple threads?
For support 100Gbps network, multi-threading is necessary, but I'm not sure about asynchronicity.

src/graph/network.cc
aksnzhy: > did you just move the location of the code?

I change the code: `MemoryPool[chandle] = buffer;`

src/graph/network/communicator.h
aksnzhy: > Why do we prefer MPI over ZMQ?

MPI can support GPU-direct communication easily.

src/graph/network/common.h
aksnzhy: It seems that nothing we can borrow from this file.

aksnzhy:
> I think more importantly, what we need to address is how to send and recv data? Should we send data asynchronously? Should we send and receive data with multiple threads?
> For support 100Gbps network, multi-threading is necessary, but I'm not sure about asynchronicity.
> For GPU communication, I don't know what access is required to achieve max throughput. Any ideas?

Multi-thread send/recv is not very easy to implement and actually I don't know which library can support multi-thread send/recv. This is because the Linux C socket is not thread-safe. (See reference:https://stackoverflow.com/questions/2354417/c-socket-api-is-thread-safe). To support multi-thread send/recv, we have to maintain multiple sockets for one connect, and we also need to divide and re-construct message by hand. 

For now, DGL communicator is sync mode, and it is very easy to be extended to async model by just adding a message queue in sender.

For multi-process GPU-communication, the best way is to use MPI, which did all the things for us.

src/graph/network.cc
aksnzhy: Actually, we don't need a memory pool here. We just need to a container that stores the pointer of the data  buffer.

aksnzhy:
> Is it right that the main modification is how to pass address?

Yes, it is.

zheng-da:
We can discuss the pros and cons of multi-threading and multi-processing.

First, when using multi-threading, we'll have a socket/MPI handle/etc for each thread. So there is no thread-safty issue. You can think of a socket as a channel. A message is sent through a channel without being broken up, so you don't need to assemble them. As an abstraction, the sender and receiver maintains a message queue. Each message is sent through a channel in a thread and is received in a similar fashion. All the user needs to do is to add multiple messages to the queue.

The advantages of multi-threading data transmission includes:
* This abstraction decouples the parallelization strategy in the user's computation from data transmission. To use a machine with 32 CPU cores, we can have users to specify any number of threads for parallel data transmission and any number of threads for computation. It's very easy for users to set it up.
* We don't need to maintain so many data partitions. When we want to push computation to data, having a smaller number of partitions has significant advantage. One simple example is that when we push aggregation computation to data partitions, small number of partitions means much smaller data transmission.

Multi-processing may make the server-side implementation much easier. What do you do for the client? It seems we still need to maintain multiple threads in the client side to receive messages. Otherwise, do we need to run many trainers to access the kvstore servers?

zheng-da:
How do you send messages from a client to multiple servers? do you think them async or with multi-threading?

aksnzhy:
> We can discuss the pros and cons of multi-threading and multi-processing.
> 
> First, when using multi-threading, we'll have a socket/MPI handle/etc for each thread. So there is no thread-safty issue. You can think of a socket as a channel. A message is sent through a channel without being broken up, so you don't need to assemble them. As an abstraction, the sender and receiver maintains a message queue. Each message is sent through a channel in a thread and is received in a similar fashion. All the user needs to do is to add multiple messages to the queue.
> 
> The advantages of multi-threading data transmission includes:
> 
> * This abstraction decouples the parallelization strategy in the user's computation from data transmission. To use a machine with 32 CPU cores, we can have users to specify any number of threads for parallel data transmission and any number of threads for computation. It's very easy for users to set it up.
> * We don't need to maintain so many data partitions. When we want to push computation to data, having a smaller number of partitions has significant advantage. One simple example is that when we push aggregation computation to data partitions, small number of partitions means much smaller data transmission.
> 
> Multi-processing may make the server-side implementation much easier. What do you do for the client? It seems we still need to maintain multiple threads in the client side to receive messages. Otherwise, do we need to run many trainers to access the kvstore servers?

But for users,

aksnzhy:
> How do you send messages from a client to multiple servers? do you think them async or with multi-threading?

For now, the Sender uses sync/single-thread and the receiver uses sync/multi-thread.

aksnzhy:
> We can discuss the pros and cons of multi-threading and multi-processing.
> 
> First, when using multi-threading, we'll have a socket/MPI handle/etc for each thread. So there is no thread-safty issue. You can think of a socket as a channel. A message is sent through a channel without being broken up, so you don't need to assemble them. As an abstraction, the sender and receiver maintains a message queue. Each message is sent through a channel in a thread and is received in a similar fashion. All the user needs to do is to add multiple messages to the queue.
> 
> The advantages of multi-threading data transmission includes:
> 
> * This abstraction decouples the parallelization strategy in the user's computation from data transmission. To use a machine with 32 CPU cores, we can have users to specify any number of threads for parallel data transmission and any number of threads for computation. It's very easy for users to set it up.
> * We don't need to maintain so many data partitions. When we want to push computation to data, having a smaller number of partitions has significant advantage. One simple example is that when we push aggregation computation to data partitions, small number of partitions means much smaller data transmission.
> 
> Multi-processing may make the server-side implementation much easier. What do you do for the client? It seems we still need to maintain multiple threads in the client side to receive messages. Otherwise, do we need to run many trainers to access the kvstore servers?

So you mean that multi-thread sending should be handled by application?

aksnzhy:
> > We can discuss the pros and cons of multi-threading and multi-processing.
> > First, when using multi-threading, we'll have a socket/MPI handle/etc for each thread. So there is no thread-safty issue. You can think of a socket as a channel. A message is sent through a channel without being broken up, so you don't need to assemble them. As an abstraction, the sender and receiver maintains a message queue. Each message is sent through a channel in a thread and is received in a similar fashion. All the user needs to do is to add multiple messages to the queue.
> > The advantages of multi-threading data transmission includes:
> > 
> > * This abstraction decouples the parallelization strategy in the user's computation from data transmission. To use a machine with 32 CPU cores, we can have users to specify any number of threads for parallel data transmission and any number of threads for computation. It's very easy for users to set it up.
> > * We don't need to maintain so many data partitions. When we want to push computation to data, having a smaller number of partitions has significant advantage. One simple example is that when we push aggregation computation to data partitions, small number of partitions means much smaller data transmission.
> > 
> > Multi-processing may make the server-side implementation much easier. What do you do for the client? It seems we still need to maintain multiple threads in the client side to receive messages. Otherwise, do we need to run many trainers to access the kvstore servers?
> 
> So you mean that multi-thread sending should be handled by application?

zheng-da:
No, what I mean is that we can implement a message queue for message passing. Application code simply adds messages to the message queue and the transmission code can send messages with multiple threads.

aksnzhy:
> No, what I mean is that we can implement a message queue for message passing. Application code simply adds messages to the message queue and the transmission code can send messages with multiple threads.

Ok, I get your point. I think it's not very hard to implement this on current code. Actually, the Receiver has already used this model, and I can just add a message_queue to sender.

jermainewang:
Regarding the memory buffer, it's quite hacky to implement it in current way. At least two issues:
(1) Why always kMaxSize? The max size is hard to tune.
(2) Why one buffer per client? What if multiple messages are sent to the same client? Will you block?

Essentially, there are two modes of send/recv, blocking and non-blocking. Let's take send as an example
* In blocking mode, user (who has an in-memory object to send such as a Nodeflow) serializes it to a buffer (created by user-side memory pool or whatever), pass it to `Send`. Blocking mode will essentially copy the buffer to its internal buffer so when `Send` returns, the user-provided buffer can be safely freed.
* In non-blocking mode, user still creates a buffer and serializes an object to it and pass the buffer to `Send`. However, the `Send` will simply queue the buffer into its internal task queue and return. User should not free the buffer because **the communicator has assumed responsibility for it**. User could optionally provide a deallocator function to it.

In either way, we don't need that `MemoryBuffer` object. Your CAPI part logic is equal to the user role in the above, so it should use a memory pool for acquiring serialization buffers. The communicator could be blocking or non-blocking. We state it crystal clear. If it's non-blocking, you need to have an extra argument for deallocator and an internal message queue to queue the buffer (just as @zheng-da suggested). The message will be destroyed after it is internally sent.

Please have a look at this zeromq doc: http://zeromq.org/blog:zero-copy

zheng-da:
We need to define the semantics of the communicator clearly:
* Is it blocking or non-blocking? 
* What is the thread context of application? Should they be the same for transmission threads?

My previous proposal will require that the communicator be non-blocking and that application threads are always separate from IO threads.

zheng-da:
I agree that we should use a memory pool to manage memory and recycle them, unless we want to use the blocking semantic of sending and receiving.

For memory management, we should avoid two things as much as possible: memory copy and memory allocation. Currently, the received data is copied to the memory buffer before being used by application code, which will be very expensive for 100Gbps network. Memory allocation of large piece of memory is also very expensive, which requires many pagefault interrupts to allocate physical memory pages for the allocated memory space.

We should get this right either in this PR or the next PR.

aksnzhy:
The PR679 is completed, main changes:

1. Implemented a zero-copy message queue to avoid large memory copy during send/recv.
2. Implemented non-block send that returns immediately.
3. Implemented multi-thread send/recv that maximal the throughput network communication.
4. Implemented multi-part send/recv to avoid copy ndarray data in serialization by meta-data pre-sending.
5. More general API for MPI, which will be used in GPU-direct communication in the future.

src/graph/network.cc
```diff
@@ -78,104 +91,148 @@ DGL_REGISTER_GLOBAL("network._CAPI_DGLSenderConnect")
...
+      csr->indptr().GetSize(),
+      recv_id);
+    sender->Send(
+      static_cast<char*>(csr->indices()->data),
+      csr->indices().GetSize(),
+      recv_id);
+    sender->Send(
+      static_cast<char*>(csr->edge_ids()->data),
+      csr->edge_ids().GetSize(),
+      recv_id);
```
jermainewang: Since the `sender` supports multi-threads, how to assure that the receiver receives all these messages in one bundle?

src/graph/network.cc
```diff
@@ -78,104 +91,148 @@ DGL_REGISTER_GLOBAL("network._CAPI_DGLSenderConnect")
...
-                                          &(nf->node_mapping),
-                                          &(nf->edge_mapping),
-                                          &(nf->layer_offsets),
-                                          &(nf->flow_offsets));
+      // node_mapping
+      std::unique_ptr<char> array_0(receiver->RecvFrom(&data_size, send_id));
+      CHECK_EQ(msg.data_shape_[0], 1);
+      nf->node_mapping = NDArray::Empty(
+        {msg.data_shape_[1]}, DLDataType{kDLInt, 64, 1}, DLContext{kDLCPU, 0});
+      // TODO(chao): Can we avoid memory copy here? ndarray->data is const pointer.
```
jermainewang: Use [`NDArray::FromDLPack`](https://github.com/dmlc/dgl/blob/master/include/dgl/runtime/ndarray.h#L184).

src/graph/network/socket_communicator.cc
```diff
@@ -57,101 +86,201 @@ bool SocketSender::Connect() {
...
-void SocketReceiver::MsgHandler(TCPSocket* socket, MessageQueue* queue, int id) {
-  char* buffer = new char[kMaxBufferSize];
+char* SocketReceiver::Recv(int64_t* size, int* send_id) {
+  // Get the top message
+  for (;;) {  // loop until get a message
+    for (auto& mq : msg_queue_) {
+      if (mq.second->Empty() == false) {
+        *send_id = mq.first;
+        return msg_queue_[*send_id]->Remove(size);
+      }
```
zheng-da: I'm confused that who owns memory.

src/graph/network/socket_communicator.cc
```diff
@@ -160,54 +289,31 @@ void SocketReceiver::MsgHandler(TCPSocket* socket, MessageQueue* queue, int id)
...
-    if (socket_[i] != nullptr) {
-      socket_[i]->Close();
-      delete socket_[i];
-      socket_[i] = nullptr;
+    if (data_size < 0) {
+      LOG(FATAL) << "Recv data error (data_size: " << data_size << ")";
+    } else if (data_size == 0) {  // This is a end-signal sent by client
+      return;
+    } else {
+      char* buffer = new char[data_size];
```
zheng-da: the memory for receiving data is allocated here. but when is it deallocated?

tests/cpp/socket_communicator_test.cc
```diff
@@ -73,21 +73,24 @@ TEST(SocketCommunicatorTest, SendAndRecv) {
...
+  SocketReceiver receiver(500*1024);
+  receiver.Wait("socket://127.0.0.1:50091", 1);
+  int64_t size = 0;
+  char* data = nullptr;
+  for (int i = 0; i < 100; ++i) {
+    data = receiver.RecvFrom(&size, 0);
+  }
   receiver.Finalize();
-  return string("123456789") == string(serbuff);
+  return string("123456789") == string(data, size);
```
zheng-da: please test the code with a more normal scenario. For example, construct NodeFlows, send and receive them and see how it works. It seems the current implementation has memory leak and potential memory error if the sender destroys memory before the message is sent.

src/graph/network.h
```diff
@@ -8,24 +8,155 @@
...
 
-#define IS_SENDER    true
-#define IS_RECEIVER  false
+// Max size of message queue for communicator is 200 MB
+// TODO(chao): Make this number configurable
+const int64_t kQueueSize = 200 * 1024 * 1024;
+
+// Message type
+const int NF_MSG = 0;
+const int END_MSG = 1;
```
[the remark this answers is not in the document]
jermainewang:
```c++
enum MessageType {
  // comment ...
  kNodeFlowMsg = 0,
  // comment ...
  kEndMsg,
};
```

src/graph/network.h
```diff
@@ -8,24 +8,155 @@
...
+  int Type() const {
+    return msg_type_;
+  }
+
+  /*!
+   * \return count of ndarray
+   */
+  int NDArrayCount() const {
+    return ndarray_count_;
+  }
```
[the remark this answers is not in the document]
jermainewang: That's a relatively arguable convention. You could choose not to use it.

src/graph/network/common.h
```diff
@@ -0,0 +1,132 @@
...
+//------------------------------------------------------------------------------
+
+std::string StringPrintf(const char* format, ...);
+void SStringPrintf(std::string* dst, const char* format, ...);
+void StringAppendF(std::string* dst, const char* format, ...);
+
+}  // namespace network
+}  // namespace dgl
+
+#endif  // DGL_GRAPH_NETWORK_COMMON_H_
```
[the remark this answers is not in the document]
jermainewang: https://github.com/dmlc/dmlc-core/blob/master/include/dmlc/io.h#L525 . DMLC handles `hdfs://...`, `s3://...`. Don't know how the function is applicable here.

src/graph/network/socket_communicator.cc
```diff
@@ -57,101 +86,201 @@ bool SocketSender::Connect() {
...
-    sent_bytes += tmp;
-  }
-
-  return size + sizeof(int64_t);
+  CHECK_NOTNULL(data);
+  if (recv_id < 0) {
+    LOG(FATAL) << "recv_id cannot be a negative number.";
+  }
+  // Add data to message queue
+  int64_t send_size = msg_queue_[recv_id]->Add(data, size);
```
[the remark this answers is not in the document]
jermainewang: We cannot. In this case, user should provide a deallocator and be aware that the communicator will handle the deallocation. Zeromq also adopts this practice (http://api.zeromq.org/4-0:zmq-msg-send). 

>A successful invocation of zmq_msg_send() does not indicate that the message has been transmitted to the network, only that it has been queued on the socket and ØMQ has assumed responsibility for the message. You do not need to call zmq_msg_close() after a successful zmq_msg_send().

src/graph/network/socket_communicator.cc
jermainewang: For receiver side, the buffer should be allocated by the communicator. Then the ownership should be transferred to the user side. Take NDArray as an example:
1. The buffer allocated in communicator and stored in the message queue.
2. User calls `Recv` and get the data buffer (communicator no longer owns the memory).
3. User creates a DLTensor using the data buffer and correctly set the deallocator.
4. User creates an NDArray from that DLTensor. NDArray is equipped with reference counting, so the data buffer will be destroyed once the NDArray has no reference.

src/graph/network/socket_communicator.cc
```diff
@@ -160,54 +289,31 @@ void SocketReceiver::MsgHandler(TCPSocket* socket, MessageQueue* queue, int id)
...
-        sizeof(int64_t));
-    }
-    if (socket_[i] != nullptr) {
-      socket_[i]->Close();
-      delete socket_[i];
-      socket_[i] = nullptr;
+    if (data_size < 0) {
+      LOG(FATAL) << "Recv data error (data_size: " << data_size << ")";
+    } else if (data_size == 0) {  // This is a end-signal sent by client
+      return;
```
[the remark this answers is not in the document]
aksnzhy: These are two exit conditions that controlled by different components. `queue->EmptyAndNoMoreAdd()` is to check if user call `Signal()`, and `else if (data_size == 0)` is prevent listening remote sender in a dead loop when sender is shut down.

src/graph/network.cc
aksnzhy: > Since the `sender` supports multi-threads, how to assure that the receiver receives all these messages in one bundle?

For sender, multi-thread will be used for sending messages to different remote nodes, for receiver, each thread has an independent message queue for the guarantee of the message order.

aksnzhy: > I remember we discussed some API like sending a vector of arrays.
> Also, is `Send` non-blocking? how do you manage the memory? Both meta message memory and data in NodeFlow?

I implemented multi-part send into application level. And send is non-blocking, it will push data into a message queue and return immediately. All the message will be simple new/delete in this case.

src/graph/network/communicator.h
```diff
@@ -54,64 +67,80 @@ class Sender {
...
-   * \param max_size maximul size of data buffer
-   * \return bytes we received
-   *   > 0 : bytes we received
-   *   - 1 : error
+   * \brief Recv data from Sender
+   * \param size data size we received
+   * \param send_id which sender current msg comes from
+   * \return data buffer we received
+   */
+  virtual char* Recv(int64_t* size, int* send_id) = 0;
```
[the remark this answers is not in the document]
aksnzhy: Yes, actually, we just need `deallocator` in `Message` because we use zero-copy here, and the data has already been allocated by users. e.g., the `NodeFlow` will be created before sending.

src/graph/network/msg_queue.cc
```diff
@@ -14,28 +14,18 @@ namespace network {
...
 
-MessageQueue::~MessageQueue() {
-  std::lock_guard<std::mutex> lock(mutex_);
-  if (nullptr != queue_) {
-    delete [] queue_;
-    queue_ = nullptr;
-  }
-}
-
 int64_t MessageQueue::Add(const char* src, int64_t size, bool is_blocking) {
```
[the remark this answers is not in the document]
aksnzhy: `src` is created by user, e.g `NodeFlow` object

src/graph/network.cc
jermainewang: > > I remember we discussed some API like sending a vector of arrays.
> > Also, is `Send` non-blocking? how do you manage the memory? Both meta message memory and data in NodeFlow?
> 
> I implemented multi-part send into application level. And send is non-blocking, it will push data into a message queue and return immediately. All the message will be simple new/delete in this case.

Could you share a pseudo-code of how you achieve it?

zheng-da: Here is the place that concerns me: https://github.com/dmlc/dgl/pull/679/files/a8ec9fd31538357ae6726fd73a685bc9d7f0e021#diff-3c750fcd5279ed512c14a3cfa930446cR140
You send data in NodeFlow with multiple sends, which will be sent by multiple threads if the sender has multiple threads. How do you assemble the data in the receiver and construct NodeFlow?

src/graph/network/msg_queue.cc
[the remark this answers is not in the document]
zheng-da: Does it work? The user doesn't know when the data is sent by the sender and doesn't know what is the right time to deallocate the memory. I think it's better to pass the ownership to the sender or both the user and the sender share the ownership of the memory (i.e., using `shared_ptr`).

src/graph/network/communicator.h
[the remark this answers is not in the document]
aksnzhy: Yes, actually we just need `deallocator` in the `message` because we use zero-copy here, and the data will be allocated by users before.

src/graph/network/socket_communicator.h
```diff
@@ -32,118 +33,147 @@ struct Addr {
...
+   * \return bytes of data
    *   > 0 : bytes we sent
    *   - 1 : error
+   *
+   * Note that, DO NOT delete the data pointer in Send() because we use zero-copy here. 
+   * The Send() API is a non-blocking API, which will return from function immediantely 
+   * if the message queue is not full, and it will be blocked when message queue is full.
+   *
+   * The Send() API is thread-safe, but we DO NOT guarantee the order of message 
+   * in multi-threading sending.
```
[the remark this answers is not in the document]
aksnzhy: For my implementation,  if we set deallocator to nullptr, the deallocator will do nothing on the data, because sometimes users don't want to delete their message in communicator.

src/graph/network/socket_communicator.cc
```diff
@@ -57,101 +86,201 @@ bool SocketSender::Connect() {
...
 
-void SocketReceiver::MsgHandler(TCPSocket* socket, MessageQueue* queue, int id) {
-  char* buffer = new char[kMaxBufferSize];
+char* SocketReceiver::Recv(int64_t* size, int* send_id) {
+  // Get the top message
+  for (;;) {  // loop until get a message
+    for (auto& mq : msg_queue_) {
+      if (mq.second->Empty() == false) {
+        *send_id = mq.first;
+        return msg_queue_[*send_id]->Remove(size);
```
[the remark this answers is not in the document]
aksnzhy: That is a question. In your case, the Remove() will be blocked until its msg_queue is not empty. For our case, we only have one thread to invoke the recv() API, so that is not a problem. For multi-threading, do you have any suggestion to make it efficient?  It is too expensive if we use a global mutex here,

src/graph/network.cc
aksnzhy: > Here is the place that concerns me: https://github.com/dmlc/dgl/pull/679/files/a8ec9fd31538357ae6726fd73a685bc9d7f0e021#diff-3c750fcd5279ed512c14a3cfa930446cR140
> You send data in NodeFlow with multiple sends, which will be sent by multiple threads if the sender has multiple threads. How do you assemble the data in the receiver and construct NodeFlow?

Here is a misunderstanding of my implementation. We send `nodeflow` in multiple send() here, but all the send() will push data into an independent message queue and send it in one thread. Multi-thread send will be only performed for sending data to different nodes with different `recv_id`.  One thread for one queue, and one queue for one node.

aksnzhy: > > > I remember we discussed some API like sending a vector of arrays.
> > > Also, is `Send` non-blocking? how do you manage the memory? Both meta message memory and data in NodeFlow?
> > 
> > 
> > I implemented multi-part send into application level. And send is non-blocking, it will push data into a message queue and return immediately. All the message will be simple new/delete in this case.
> 
> Could you share a pseudo-code of how you achieve it?

For example, if you want to send a nodeflow to one remote node, you can just do that:

```
send(meta_data)
send(nf.node_mapping)
send(nf.edge_mapping)
send(nf.layer_offset)
send(nf.flow_offset)
...
```

and in receiver:

```
recv(meta_data)
recv(node_mapping)
recv(edge_mapping)
recv(layer_offset)
recv(flow_offset)
```

As we give each socket connection an independent message queue, so you don't need to worry about the order of the message.

src/graph/network/msg_queue.cc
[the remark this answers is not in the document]
aksnzhy: The deallocator will delete the message when finish sending. If the user doesn't want to delete the message in communicator, they can just set deallocator to nullptr.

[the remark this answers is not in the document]
zheng-da: so the user passes the ownership of the memory to the sender? very much like std::uniq_ptr? It's fine. did you change the code in the local repo? When I review your code, I didn't see your code do that.

src/graph/network.cc
zheng-da: I think I can see why you want to parallelize communication between destination machines, instead of between messages. If we want to parallelize communication between a pair of machines, we need to create multiple sockets between the two nodes.

However, I'm wondering if it can achieve the maximal throughput of the network. Many graphs have a power-law distribution. it means that communication to different nodes can be very unbalanced, especially in a short period of time. Later on, we need to observe the network throughput. It might be a problem.

Could you please document this in the comment? Both how is parallelization implemented and the potential problem with the current implementation?

src/graph/network/communicator.h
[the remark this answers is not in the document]
zheng-da: Currently, we use `new`. We should have better memory management. Allocating a large chunk of memory can be very expensive.

src/graph/network.cc
aksnzhy: > I think I can see why you want to parallelize communication between destination machines, instead of between messages. If we want to parallelize communication between a pair of machines, we need to create multiple sockets between the two nodes.
> 
> However, I'm wondering if it can achieve the maximal throughput of the network. Many graphs have a power-law distribution. it means that communication to different nodes can be very unbalanced, especially in a short period of time. Later on, we need to observe the network throughput. It might be a problem.
> 
> Could you please document this in the comment? Both how is parallelization implemented and the potential problem with the current implementation?

Sure. Actually, if we don't need multi-part sending and pack all the data into one message, we can implement multi-thread sending between messages easily. We can leave this potential improvement for the future.

src/graph/network/msg_queue.cc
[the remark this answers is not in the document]
aksnzhy: > so the user passes the ownership of the memory to the sender? very much like std::uniq_ptr? It's fine. did you change the code in the local repo? When I review your code, I didn't see your code do that.

I just updated my code for the deallocator.

tests/cpp/socket_communicator_test.cc
aksnzhy: I will add more test in socket_communicator_test.cc, and I test sending nodeflow in python.

src/graph/network/socket_communicator.cc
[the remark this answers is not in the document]
jermainewang: Well, technically we are using global mutex already when entering `MessageQueue::Remove`. The problem here could be solved by letting `Remove` returns size 0 if the queue is empty:
```c++
for (;;) {
  for (auto& mq : msg_queue_) {
    *send_id = mq.first;
    const int64_t size = mq.second->Remove(msg);
    if (size < 0) {
      LOG(..) << ..;
      return -1;
    }
    if (size > 0) {
      return size;
    }
  }
}
```
Also, I feel quite redundant to return the data size when the `Message` structure already contains a size. Consider return some status code?

src/graph/network/msg_queue.h
```diff
@@ -13,30 +13,54 @@
...
+ * \brief Message used by network communicator and message queue.
+ */
+struct Message {
+  /*!
+   * \brief message data
+   */
+  char* data;
+  /*!
+   * \brief message size
+   */
```
jermainewang: Message size in bytes.

src/graph/network/msg_queue.h
```diff
@@ -13,30 +13,54 @@
...
+  /*!
+   * \brief user-defined deallocator, which can be nullptr
+   */
+  std::function<void(Message*)> deallocator = nullptr;
+};
+
+/*!
+ * \brief Free memory buffer for message
+ */
+inline void msg_clear(Message* msg) { delete [] msg->data; }
```
jermainewang: `DefaultMessageDeleter`

src/graph/network/communicator.h
```diff
@@ -6,112 +6,144 @@
...
+   * \brief Send data to specified Receiver.
+   * \param msg data message
    * \param recv_id receiver's ID
-   * \return bytes we sent
+   * \return bytes of data
    *   > 0 : bytes we sent
    *   - 1 : error
    */
-  virtual int64_t Send(const char* data, int64_t size, int recv_id) = 0;
+  virtual int64_t Send(Message msg, int recv_id) = 0;
```
jermainewang: Please update the comment:
(1) The send is non-blocking. There is no guarantee that the message has been physically sent out when the function returns.
(2) The communicator will assume the responsibility of the given message.
(3) The API is multi-thread safe.
(4) Messages sent to the same receiver are guaranteed to be received in the same order. There is no guarantee for messages sent to different receivers.

src/graph/network/communicator.h
```diff
@@ -6,112 +6,144 @@
...
   /*!
-   * \brief Add receiver address and it's ID to the namebook
-   * \param ip receviver's IP address
-   * \param port receiver's port
+   * \brief Add receiver's address and ID to the sender's namebook
+   * \param addr Networking address, e.g., 'socket://127.0.0.1:50091', 'mpi://0'
    * \param id receiver's ID
    */
-  virtual void AddReceiver(const char* ip, int port, int id) = 0;
+  virtual void AddReceiver(const char* addr, int id) = 0;
```
jermainewang: Please update the docstring about whether this API is multi-thread safe or not.

src/graph/network/communicator.h
```diff
@@ -6,112 +6,144 @@
...
-   * \return bytes we received
-   *   > 0 : bytes we received
+   * \brief Recv data from Sender
+   * \param msg pointer of data message
+   * \param send_id which sender current msg comes from
+   * \return bytes of data
+   *   > 0 : bytes we sent
+   *   - 1 : error
+   */
+  virtual int64_t Recv(Message* msg, int* send_id) = 0;
```
jermainewang: Similar to `Send`, update the docstring.
(1) Multi-thread safe
(2) Blocking
(3) Memory allocated by communicator but will not own it after the function returns.

src/graph/network/communicator.h
```diff
@@ -6,112 +6,144 @@
...
+  /*!
+   * \brief Recv data from a specified Sender
+   * \param msg pointer of data message
+   * \param send_id sender's ID
+   * \return bytes of data
+   *   > 0 : bytes we sent
    *   - 1 : error
    */
-  virtual int64_t Recv(char* dest, int64_t max_size) = 0;
+  virtual int64_t RecvFrom(Message* msg, int send_id) = 0;
```
jermainewang: Same

src/graph/network/communicator.h
```diff
@@ -6,112 +6,144 @@
...
-   * \param port Receiver's port
+   * \brief Wait for all the Senders to connect
+   * \param addr Networking address, e.g., 'socket://127.0.0.1:50051', 'mpi://0'
    * \param num_sender total number of Senders
-   * \param queue_size size of message queue
-   * \return True for sucess and False for fail
+   * \return True for success and False for fail
    */
-  virtual bool Wait(const char* ip, int port, int num_sender, int queue_size) = 0;
+  virtual bool Wait(const char* addr, int num_sender) = 0;
```
jermainewang: Docstring

src/graph/network.cc
```diff
@@ -78,104 +155,176 @@ DGL_REGISTER_GLOBAL("network._CAPI_DGLSenderConnect")
...
-                                          &(nf->edge_mapping),
-                                          &(nf->layer_offsets),
-                                          &(nf->flow_offsets));
+      // node_mapping
+      Message array_0;
+      CHECK_NE(receiver->RecvFrom(&array_0, send_id), -1);
+      CHECK_EQ(msg.data_shape_[0], 1);
+      nf->node_mapping = NDArray::Empty(
+        {msg.data_shape_[1]}, DLDataType{kDLInt, 64, 1}, DLContext{kDLCPU, 0});
+      memcpy(static_cast<char*>(nf->node_mapping->data), array_0.data, array_0.size);
```
jermainewang: Is the memory copy problem resolved? Look like you are still using memory copy.

jermainewang:
One more thing. I still suggest change the return value of `Send` and `Recv` to be some status code. It is redundant to return message size given the message structure already has that.

src/graph/network/communicator.h
aksnzhy: I will update it in socket_communicator because the behavior of different communicator implementation could be different.

src/graph/network/socket_communicator.cc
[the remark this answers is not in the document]
aksnzhy: For example, SocketComminactor is thread-safe but it's not in MPI implementation.

src/graph/network/communicator.h
aksnzhy: > I will update it in socket_communicator because the behavior of different communicator implementation could be different.

For example, SocketComminactor is thread-safe but it's not in MPI implementation.

jermainewang: Why MPI implementation cannot be thread-safe?

src/graph/network/communicator.h
aksnzhy: same

src/graph/network/communicator.h
aksnzhy: same

src/graph/network/communicator.h
aksnzhy: > Why MPI implementation cannot be thread-safe?

If let me implement mpi_communicator, the send() API will call mpi send directly. But I remember that all the mpi send/recv must be invoked in a single thread.

src/graph/network/communicator.h
aksnzhy: updated in socket_communicator

src/graph/network/communicator.h
aksnzhy: updated in socket_communicator

aksnzhy:
@jermainewang @zheng-da Hi, thanks so much for the comments. I have already solved most of the important issues and just leave a few small issues that I'm not sure if it should be changed or I think it should be left for the furture. Please let me know where I should improve for this PR and if we can merge it. After merging this PR, I will post a new kvstore PR based on this PR for review. Thanks again!

src/graph/network/communicator.h
jermainewang: You could still implement a thread-safe version by the message queue design. The design of the `Communicator` interface, as you've written in the docstring, is to provide a unified abstraction no matter how it is implemented (mpi or socket). If socket and mpi implementations have different behaviors and the usage will depend on that, does that mean when switching to MPI implementation the codes will be broken? then why need an interface? That's why these behavior definition should be in the interface class.

jermainewang:
One more comment and everything else looks fine for me. Great job! Thanks. @zheng-da do you have other comments?

src/graph/network.cc
```diff
@@ -78,104 +156,233 @@ DGL_REGISTER_GLOBAL("network._CAPI_DGLSenderConnect")
...
+      node_mapping_tensor.data = array_0.data;
+      node_mapping_tensor.ctx = DLContext{kDLCPU, 0};
+      node_mapping_tensor.ndim = 1;
+      node_mapping_tensor.dtype = DLDataType{kDLInt, 64, 1};
+      node_mapping_tensor.shape = new int64_t[1];
+      node_mapping_tensor.shape[0] = msg.data_shape_[1];
+      node_mapping_tensor.byte_offset = 0;
+      DLManagedTensor *node_mapping_managed_tensor = new DLManagedTensor();
+      node_mapping_managed_tensor->dl_tensor = node_mapping_tensor;
+      nf->node_mapping = NDArray::FromDLPack(node_mapping_managed_tensor);
```
zheng-da: This looks very lengthy. Can we put them into a function?

src/graph/network.cc
```diff
@@ -78,104 +156,233 @@ DGL_REGISTER_GLOBAL("network._CAPI_DGLSenderConnect")
...
+    Message send_msg;
+    send_msg.data = data;
+    send_msg.size = size;
+    send_msg.deallocator = DefaultMessageDeleter;
+    CHECK_NE(sender->Send(send_msg, recv_id), -1);
+    // send node_mapping
+    Message node_mapping_msg;
+    node_mapping_msg.data = static_cast<char*>(node_mapping->data);
+    node_mapping_msg.size = node_mapping.GetSize();
+    CHECK_NE(sender->Send(node_mapping_msg, recv_id), -1);
```
zheng-da: pass the ownership of the memory in NDArray to sender. The simplest thing to do is to add the NDArray to the deallocator. @jermainewang is it the right way of avoiding the memory in NDArray from being free'd?

tests/cpp/socket_communicator_test.cc
```diff
@@ -1,32 +1,103 @@
...
+  SocketSender sender(kQueueSize);
+  for (int i = 0; i < kNumReceiver; ++i) {
+    sender.AddReceiver(ip_addr[i], i);
+  }
+  sender.Connect();
+  std::string str_data("123456789");
+  for (int i = 0; i < kNumMessage; ++i) {
+    for (int n = 0; n < kNumReceiver; ++n) {
+      Message msg = {const_cast<char*>(str_data.data()), 9};
+      EXPECT_EQ(sender.Send(msg, n), ADD_SUCCESS);
```
zheng-da: Test memory management in the unit test.

src/graph/network.cc
aksnzhy: @jermainewang When sending ndarray, we assume that user will use this NDArray in the future, so we don't set deallocator to the message here. If we want to clear reference here, how can we do it safely (avoid free the memory that user will use later)?

jermainewang: Da is right. The simplest way to achieve this is to let the deallocator hold a reference to the NDArray. The deallocator body does nothing because when the message structure is destroyed, the deallocator function pointer is destroyed and so does the NDArray and its data.

src/graph/network.cc
```diff
@@ -78,104 +156,218 @@ DGL_REGISTER_GLOBAL("network._CAPI_DGLSenderConnect")
...
+    node_mapping_msg.size = node_mapping.GetSize();
+    node_mapping_msg.aux_handler = &node_mapping;
+    node_mapping_msg.deallocator = NDArrayDeleter;
+    CHECK_NE(sender->Send(node_mapping_msg, recv_id), -1);
+    // send edege_mapping
+    Message edge_mapping_msg;
+    edge_mapping_msg.data = static_cast<char*>(edge_mapping->data);
+    edge_mapping_msg.size = edge_mapping.GetSize();
+    edge_mapping_msg.aux_handler = &edge_mapping;
+    edge_mapping_msg.deallocator = NDArrayDeleter;
```
zheng-da: Is this  really a right way of deallocating the NDArray? `edge_mapping` is an object in the stack. But `NDArrayDeleter` call `delete` to free the pointer to the object in the stack. It seems to cause serious memory error. Do I miss something here?

aksnzhy: It is an error and will be fixed in a new small PR.

PR #693, opened 2019-06-28
[FFI] FFI container support, custom structure extension via Object

jermainewang:
## Description

This PR adds common container types to CFFI. It also enables quick prototype using Object system. The codes are mainly borrowed from TVM's [Node system](https://docs.tvm.ai/dev/runtime.html#tvm-node-and-compiler-stack). The change is that TVM's Node system is quite deeply coupled with DSL concept, while DGL needs a runtime mechanism to support a collection of NDArrays for example. Thus, there are following changes:
* Decouple TVM's original Node system from DSL components (e.g. HalideIR, Expr, etc.). Rename it to Object.
* An Object can contain any DGLValue. In fact, a special `ValueObject` container handles that.
* It does not support serialization to json string.

Besides the above, the new DGL Object system inherits many benefits from TVM's Node system. Namely,
* Easy support of (nested) collection type for CAPI arguments and return values. For example, calling a CAPI with an argument of python list/tuple/dict type works like magic.
* Easy extension (prototyping) of python-exposed C data structure. Simply interit `ObjectBase` and use `register_object` decorate does the trick.
* Reflection mechanism to visit the internal fields of a python-exposed C data structure.

See the "Usage Guide" section for more information. The section will also be included in the document for developers. Note that this change will NOT affect any of the existing APIs, because they are not using this new mechanism. However, I do think we should try to use it for following components:
* Graph
* Subgraph
* Nodeflow
* BatchedGraph

Note that the change cannot have unit tests at the moment, because it cannot be fully tested in either C or python side alone. When a CAPI uses it, it would be tested naturally.

Thanks @tqchen for many advices.

TBD:
* Some benchmarks.

## Checklist
- [x] Code is well-documented

### Defining new C API
We all like Python because it is easy to manipulate. We all like C because it is fast, reliable and typed. To have the merits of both ends, DGL is mostly in python, for quick prototyping, while lowers the performance-critical part to C. Thus, DGL developers frequently face the scenario to write a C routine and has it exposed to python, via a mechanism called Foreign Function Interface (FFI).

There are many FFI solutions out there. In DGL, we want to keep it simple, intuitive and efficient for critical use cases. That's why we come across the FFI solution in the TVM project, and we love it. It exploits the idea of functional programming so that it exposes only a dozens of CAPIs and new APIs can be built upon it.

We (shamelessly) borrowed the idea. For example, to define a C function that is exposed to python is only a few lines of codes:
```c++
// Create a file calculator.cc anywhere in dgl/src
#include <dgl/runtime/packed_func.h>
#include <dgl/runtime/registry.h>

DGL_REGISTER_GLOBAL("calculator.MyAdd")
    int a = args[0];
    int b = args[1];
    *rv = a * b;
  });
```
Compile and build the library. And on the python side, create a `calculator.py` file under `dgl/python/dgl/`
```python
# file calculator.py
from ._ffi.function import _init_api

def add(a, b):
  # MyAdd has been registered via `_ini_api` call below
  return MyAdd(a, b)

_init_api("dgl.calculator")
```

The trick is that the FFI system first masks the type information of the function arguments, so all the C function calls can go through one C API (`DGLFuncCall`). The type information is retrieved in the function body by static conversion, and we will do runtime type check to make sure that the type conversion is correct. The overhead of such back-and-forth is negligible as long as the function call is not too light (the above example is actually a bad one). TVM's [PackedFunc document](https://docs.tvm.ai/dev/runtime.html#packedfunc) has more details.

### Defining new types
`DGLArgs` and `DGLRetValue` only support a limited number of types:
* Numerical values: int, float, double, ...
* string
* Function (in the form of PackedFunc)
* NDArray

Though limited, the above type system is very powerful because it supports function as a first-class citizen. For example, if you want to return multiple values, you can return a PackedFunc which returns each value given an integer index. However, in many cases, new types are still desired to ease the development process:
* The argument/return value is a composition of collections (e.g. dictionary of dictionary of list).
* Sometimes we just want to have a notion of "structure" (e.g. given an apple, get its color string by `apple.color`).

To achieve this, we introduce the Object type system. For example, to define a new type `Calculator`:
```c++
// in file: calculator.cc
#include <dgl/packed_func_ext.h>
using namespace runtime;
class CalculatorObject : public Object {
 public:
  std::string brand;
  int price;
  
  void VisitAttrs(AttrVisitor *v) final {
    v->Visit("brand", &brand);
    v->Visit("price", &price);
  }

  static constexpr const char* _type_key = "Calculator";
  DGL_DECLARE_OBJECT_TYPE_INFO(CalculatorObject, Object);
};

// This is to define a reference class (the wrapper of an object shared pointer).
// A minimal implementation is as follows, but you could define extra methods.
class Calculator : public ObjectRef {
 public:
  const CalculatorObject* operator->() const {
    return static_cast<const CalculatorObject*>(obj_.get());
  }
  using ContainerType = CalculatorObject;
};

DGL_REGISTER_GLOBAL("calculator.CreateCaculator")
  std::string brand = args[0];
  int price = args[1];
  auto o = std::make_shared<CalculatorObject>();
  o->brand = brand;
  o->price = price;
  *rv = o;
}
```
On the python side:
```python
# calculator.py
from dgl._ffi.object import register_object, ObjectBase
from ._ffi.function import _init_api

@register_object
class Calculator(ObjectBase):
  @staticmethod
  def create(brand, price):
    # invoke a C API, the return value is of `Calculator` type
    return CreateCalculator(brand, price)

_init_api("dgl.calculator")
```
We can then simply create `Calculator` object by:
```python
calc = Calculator.create("casio", 100)
```

What is nice about this object is that, it defines a visitor pattern that is essentially a reflection mechanism to get its internal attributes. For example, you can print the calculator's brand and by simply accessing its attributes.
```
print(calc.brand)
print(cal.price)
```

The reflection is indeed a little slow due to the string key lookup. To speedup this, you could define an attribute access API:
```c++
// in file: calculator.cc
DGL_REGISTER_GLOBAL("calculator.CaculatorGetBrand")
  Calculator calc = args[0];
  *rv = calc->brand;
}
```

### Containers
Containers are also objects. For example, the C API below accepts a list of integers and return their sum:
```c++
// in file: calculator.cc
#include <dgl/runtime/container.h>
using namespace runtime;
DGL_REGISTER_GLOBAL("calculator.Sum")
  // All the DGL supported values are represented as a ValueObject, which
  //   contains a data field.
  List<Value> values = args[0];
  int sum = 0;
  for (int i = 0; i < values.size(); ++i) {
    sum += static_cast<int>(values[i]->data);
  }
}
```

Invoking this API is simple. Just pass a python list of integers. DGL FFI will automatically convert python list/tuple/dictionary to the corresponding object type.
```python
# in file: calculator.py
from ._ffi.function import _init_api

Sum([0, 1, 2, 3, 4, 5])

_init_api("dgl.calculator")
```

The elements in the containers can be any objects, which allows the containers to be composed. Below is an API that accepts a list of calculators and print out their price:
```c++
# in file: calculator.cc
// in file: calculator.cc
#include <dgl/runtime/container.h>
using namespace runtime;
DGL_REGISTER_GLOBAL("calculator.PrintCalculators")
  List<Calculator> calcs = args[0];
  for (int i = 0; i < calcs.size(); ++i) {
    std::cout << calcs[i]->price << std::endl;
  }
}
```

Please note that containers are NOT meant for passing a large collection of items from/to C APIs. They will be quite slow in these cases. It is recommended to benchmark first. As an alternative, use NDArray for a large collection of numerical values and use BatchedDGLGraph for a lot of graphs.


jermainewang:
@VoVAllen please read the usage guide to see whether you have any questions.

jermainewang:
@BarclayII @ylfdq1118 please try your best to review the codes. Many codes are borrowed from TVM, so they should be bug-proof to some extent. I also did a careful line-by-line pass when copying the codes. I think it's more important to read through the usage guide, since you will likely to use it in the future.

VoVAllen:
Works like magic :satisfied:

jermainewang:
@BarclayII the windows CI failed with a strange error. Looks like the workspace is messed up somehow. Could you have a look?

PR #699, opened 2019-07-08
[Refactor] Separating graph and sparse matrix operations

jermainewang:
## Description
When I try to implement bipartite graph, I find that there are tons of codes that can be reused in ImmutableGraph. Thus, this PR tries to extract common logic into a sparse matrix operation module (`aten`) so it can be shared by other parts. This change is also the first effort towards what @zheng-da required -- having a fully functional int32 graph.

## Changes
* We now has a new namespace `aten` to put all the NDArray and sparse matrix related operators. The operators in the `aten` namespace should 1) can be run on both cpu/gpu; 2) can support different data type (int32, int64, etc.).
* Currently, only CPU operators are implemented.
* Sparse operators are modified so that they can support general sparse matrix (not necessarily square shape). This will be used in bipartite graph implementation.
* Immutable graph class is refactored. It is now implemented using `aten` operators. Thus, in the future, when `aten` operators can support multiple devices/types, the graph class can naturally benefit from that.

## Checklist
- [x] Code is well-documented


jermainewang:
Some review guideline: @zheng-da @BarclayII @mufeili @yzh119 
* `include/array/array.h` declares the `aten` operators.
* `src/array/array.cc` dispatches operators to different implementation according to the input array's device context and data type. The macro is defined in `src/array/common.h`.
* `src/array/array_op.h` declares the template implementation.
* `src/array/cpu/array_op_impl.cc` defines the array op implementation and also explicitly specialize the template arguments.
* `src/array/cpu/spmat_op_impl.cc` defines the sparse op implementation. Most of them are moved from previous `src/graph/immutable_graph.cc`. You could compare the two during the code review.

src/array/array.cc
```diff
@@ -0,0 +1,372 @@
...
+  ATEN_XPU_SWITCH(lhs->ctx.device_type, XPU, {
+    ATEN_ID_TYPE_SWITCH(lhs->dtype, IdType, {
+      ret = impl::BinaryElewise<XPU, IdType, arith::Div>(lhs, rhs);
+    });
+  });
+  return ret;
+}
+
+IdArray Add(dgl_id_t lhs, IdArray rhs) {
+  return Add(rhs, lhs);
```
mufeili: Why don't you do `impl::BinaryElewise<XPU, IdType, arith::Add>(lhs, rhs);` here?

src/array/array.cc
```diff
@@ -0,0 +1,372 @@
...
+  ATEN_XPU_SWITCH(rhs->ctx.device_type, XPU, {
+    ATEN_ID_TYPE_SWITCH(rhs->dtype, IdType, {
+      ret = impl::BinaryElewise<XPU, IdType, arith::Sub>(lhs, rhs);
+    });
+  });
+  return ret;
+}
+
+IdArray Mul(dgl_id_t lhs, IdArray rhs) {
+  return Mul(rhs, lhs);
```
mufeili: Same.

src/array/cpu/spmat_op_impl.cc
```diff
@@ -0,0 +1,619 @@
...
+  CHECK(col >= 0 && col < csr.num_cols) << "Invalid col index: " << col;
+  const IdType* indptr_data = static_cast<IdType*>(csr.indptr->data);
+  const IdType* indices_data = static_cast<IdType*>(csr.indices->data);
+  for (IdType i = indptr_data[row]; i < indptr_data[row + 1]; ++i) {
+    if (indices_data[i] == col) {
+      return true;
+    }
+  }
+  return false;
+}
```
mufeili: Seems we are missing the version for `runtime::NDArray CSRIsNonZero(CSRMatrix, runtime::NDArray row, runtime::NDArray col);`?

src/array/cpu/spmat_op_impl.cc
```diff
@@ -0,0 +1,619 @@
...
+
+template NDArray CSRGetRowData<kDLCPU, int32_t, int32_t>(CSRMatrix, int64_t);
+template NDArray CSRGetRowData<kDLCPU, int64_t, int64_t>(CSRMatrix, int64_t);
+
+///////////////////////////// CSRGetData /////////////////////////////
+
+template <DLDeviceType XPU, typename IdType, typename DType>
+NDArray CSRGetData(CSRMatrix csr, int64_t row, int64_t col) {
+  CHECK(CSRHasData(csr)) << "missing data array";
+  // TODO(minjie): use more efficient binary search when the column indices
```
mufeili: This is probably redundant.

src/array/array.cc
jermainewang: Because it's commutative, so I reuse the function to save several lines of codes.

src/array/cpu/spmat_op_impl.cc
jermainewang: You are right. That is `HasEdgesBetween(runtime::NDArray src, runtime::NDArray dst)`. I might forget to move that here.

src/array/cpu/spmat_op_impl.cc
jermainewang: The comment sentence is broken. Will fix that. Ideally, if the column indices are sorted for each neighborhood, a binary search could be used instead of the current linear search.

src/array/cpu/spmat_op_impl.cc
```diff
@@ -0,0 +1,619 @@
...
+  std::vector<DType> ret_vec;
+  const IdType* indptr_data = static_cast<IdType*>(csr.indptr->data);
+  const IdType* indices_data = static_cast<IdType*>(csr.indices->data);
+  const DType* data = static_cast<DType*>(csr.data->data);
+  for (IdType i = indptr_data[row]; i < indptr_data[row+1]; ++i) {
+    if (indices_data[i] == col) {
+      ret_vec.push_back(data[i]);
+    }
+  }
+  const int64_t len = ret_vec.size();
```
mufeili: Replace L207-L211 with `return VecToNDArray(ret_vec, csr.data->dtype, csr.data->ctx);`

include/dgl/array.h
```diff
@@ -57,30 +105,169 @@ IdArray Sub(dgl_id_t lhs, IdArray rhs);
...
+
+/* \brief Get the data and the row,col indices for each returned entries. */
+std::vector<runtime::NDArray> CSRGetDataAndIndices(
+    CSRMatrix , runtime::NDArray rows, runtime::NDArray cols);
+
+/*! \brief Return a transposed CSR matrix */
+CSRMatrix CSRTranspose(CSRMatrix csr);
+
+/*!
+ * \brief Convert COO matrix to CSR matrix.
```
mufeili: Should be "CSR matrix to COO matrix" instead

include/dgl/array.h
```diff
@@ -57,30 +105,169 @@ IdArray Sub(dgl_id_t lhs, IdArray rhs);
...
+ * \param csr Input csr matrix
+ * \param data_as_order If true, the data array in the input csr matrix contains the order
+ *                      by which the resulting COO tuples are stored. In this case, the
+ *                      data array of the resulting COO matrix will be empty because it
+ *                      is essentially a consecutive range.
+ * \return a coo matrix
```
mufeili: The `data_as_order` seems to be a very special case. When do we have such scenario?

src/graph/immutable_graph.cc
```diff
@@ -144,328 +97,112 @@ CSR::CSR(IdArray indptr, IdArray indices, IdArray edge_ids, bool is_multigraph,
...
 CSR::CSR(const std::string &shared_mem_name,
          int64_t num_verts, int64_t num_edges, bool is_multigraph)
   : is_multigraph_(is_multigraph), shared_mem_name_(shared_mem_name) {
-  std::tie(indptr_, indices_, edge_ids_) = MapFromSharedMemory(
+  CHECK(!(num_verts == 0 && num_edges != 0));
+  adj_.num_rows = num_verts;
+  adj_.num_cols = num_verts;
+  std::tie(adj_.indptr, adj_.indices, adj_.data) = MapFromSharedMemory(
       shared_mem_name, num_verts, num_edges, false);
 }
```
mufeili: Seems we are missing the implementation of 

```c++
  CSR(int64_t num_vertices, int64_t num_edges,
      IndptrIter indptr_begin, IndicesIter indices_begin, EdgeIdIter edge_ids_begin,
      bool is_multigraph);
```

declared in L44 of immutable_graph.h.

src/graph/immutable_graph.cc
```diff
@@ -144,328 +97,112 @@ CSR::CSR(IdArray indptr, IdArray indices, IdArray edge_ids, bool is_multigraph,
...
-  const dgl_id_t* vid_data = static_cast<dgl_id_t*>(vids->data);
-  DegreeArray rst = DegreeArray::Empty({len}, vids->dtype, vids->ctx);
-  dgl_id_t* rst_data = static_cast<dgl_id_t*>(rst->data);
-  for (int64_t i = 0; i < len; ++i) {
-    const auto vid = vid_data[i];
-    CHECK(HasVertex(vid)) << "Invalid vertex: " << vid;
-    rst_data[i] = OutDegree(vid);
-  }
-  return rst;
+  return aten::CSRGetRowNNZ(adj_, vids);
```
mufeili: Are we missing the implementation for `uint64_t OutDegree(dgl_id_t vid)`?

jermainewang: Here: https://github.com/dmlc/dgl/blob/a1a355f7e7c6437abff39d8be1aa011ab4c4723f/include/dgl/immutable_graph.h#L149

include/dgl/runtime/ndarray.h
```diff
@@ -292,16 +292,18 @@ struct NDArray::Container {
 // the usages of functions are documented in place.
 inline NDArray::NDArray(Container* data)
   : data_(data) {
-  data_->IncRef();
+  if (data_)
+    data_->IncRef();
```
zheng-da: do you see a case that `data_` is nullptr?

src/array/array.cc
```diff
@@ -0,0 +1,372 @@
...
+
+namespace aten {
+
+IdArray NewIdArray(int64_t length, DLContext ctx, uint8_t nbits) {
+  return IdArray::Empty({length}, DLDataType{kDLInt, nbits, 1}, ctx);
+}
+
+BoolArray NewBoolArray(int64_t length, DLContext ctx) {
+  return BoolArray::Empty({length}, DLDataType{kDLInt, 64, 1}, ctx);
+}
```
zheng-da: why not using a small number of bits here?

src/array/array.cc
```diff
@@ -0,0 +1,372 @@
...
+  return ret;
+}
+
+CSRMatrix COOToCSR(COOMatrix coo) {
+  CSRMatrix ret;
+  ATEN_COO_SWITCH(coo, XPU, IdType, DType, {
+    ret = impl::COOToCSR<XPU, IdType, DType>(coo);
+  });
+  return ret;
+}
```
zheng-da: can we have c++ unit tests for these functions?

include/dgl/runtime/ndarray.h
jermainewang: Yes, see following codes:
```c++
NDArray arr;  // This NDArray contains null data_
NDArray arr2 = arr;  // This will call the copy constructor above and segfault if not correctly protected.
```

src/array/cpu/array_op_impl.cc
```diff
@@ -0,0 +1,211 @@
...
+      static_cast<IdType*>(arr->data)[i] = oldv2newv[id];
+    }
+  }
+  // map array
+  IdArray maparr = NewIdArray(newid);
+  IdType* maparr_data = static_cast<IdType*>(maparr->data);
+  for (const auto& kv : oldv2newv) {
+    maparr_data[kv.second] = kv.first;
+  }
+  return maparr;
```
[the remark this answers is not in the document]
jermainewang: Similar but slightly different. (1) The function needs to relabel a union of multiple input id arrays while the IdHashMap only handles one. (2) The function needs to iterates all relabeled ids one and only one time and no bloom filter is needed.

include/dgl/array.h
```diff
@@ -57,30 +105,169 @@ IdArray Sub(dgl_id_t lhs, IdArray rhs);
 IdArray Mul(dgl_id_t lhs, IdArray rhs);
 IdArray Div(dgl_id_t lhs, IdArray rhs);
 
-
 /*! \brief Stack two arrays (of len L) into a 2*L length array */
 IdArray HStack(IdArray arr1, IdArray arr2);
 
+/*! \brief Return the data under the index. In numpy notation, A[I] */
+int64_t IndexSelect(IdArray array, int64_t index);
+IdArray IndexSelect(IdArray array, IdArray index);
```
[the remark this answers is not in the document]
jermainewang: agree

src/array/array.cc
jermainewang: It's a workaround. First, pytorch does not have bool array. Second, the return will be catched by `utils.Index` which assumes the input is an int64 id array. Since this method is used only in `HasEdgesBetween` and thus in `CSRIsNonZero`, I changed the implementation:
(1) This function is removed.
(2) Instead, when creating the return array, it will has the same bit width as the input CSR matrix.

include/dgl/array.h
jermainewang: Done.

src/array/cpu/spmat_op_impl.cc
jermainewang: Done.

src/array/cpu/spmat_op_impl.cc
jermainewang: Done.

src/array/cpu/spmat_op_impl.cc
jermainewang: Done.

src/array/cpu/spmat_op_impl.cc
```diff
@@ -0,0 +1,635 @@
...
+template NDArray CSRIsNonZero<kDLCPU, int32_t>(CSRMatrix, NDArray, NDArray);
+template NDArray CSRIsNonZero<kDLCPU, int64_t>(CSRMatrix, NDArray, NDArray);
+
+///////////////////////////// CSRHasDuplicate /////////////////////////////
+
+template <DLDeviceType XPU, typename IdType>
+bool CSRHasDuplicate(CSRMatrix csr) {
+  const IdType* indptr_data = static_cast<IdType*>(csr.indptr->data);
+  const IdType* indices_data = static_cast<IdType*>(csr.indices->data);
+  for (IdType src = 0; src < csr.num_rows; ++src) {
```
zheng-da: theoretically, the number of rows can still go above 2B nodes and the 32-bit CSR matrix still works fine.

src/graph/immutable_graph.cc
```diff
@@ -144,328 +97,118 @@ CSR::CSR(IdArray indptr, IdArray indices, IdArray edge_ids, bool is_multigraph,
...
-      *(eid_ptr++) = eids[j];
-    }
-  }
-  return CSR::EdgeArray{src, dst, eid};
+  auto csrsubmat = aten::CSRSliceRows(adj_, vids);
+  auto coosubmat = aten::CSRToCOO(csrsubmat, false);
+  // Note that the row id in the csr submat is relabled, so
+  // we need to recover it using an index select.
+  auto row = aten::IndexSelect(vids, coosubmat.row);
+  return CSR::EdgeArray{row, coosubmat.col, coosubmat.data};
```
zheng-da: is this implementation much slower than before? now we have to reconstruct csr and then coo and remap the row ids to the original graph.

tests/compute/test_graph.py
```diff
@@ -65,9 +65,8 @@ def _test_one(g):
             assert g.has_node(i)
             assert i in g
         assert not g.has_node(11)
-        assert not g.has_node(-1)
-        assert not -1 in g
```
zheng-da: why remove these two lines?

src/array/cpu/spmat_op_impl.cc
```diff
@@ -0,0 +1,635 @@
...
+
+///////////////////////////// COOHasDuplicate /////////////////////////////
+
+template <DLDeviceType XPU, typename IdType>
+bool COOHasDuplicate(COOMatrix coo) {
+  std::unordered_set<std::pair<IdType, IdType>, PairHash> hashmap;
+  const IdType* src_data = static_cast<IdType*>(coo.row->data);
+  const IdType* dst_data = static_cast<IdType*>(coo.col->data);
+  const auto nnz = coo.row->shape[0];
+  for (IdType eid = 0; eid < nnz; ++eid) {
```
zheng-da: this is definitely not right. 2B x 2B matrix can have more than 2B edges. we should use int64 here. a general suggest: try using int64 as much as possible for index in for loops.

src/array/array.cc
jermainewang: Done.

src/graph/immutable_graph.cc
jermainewang: No. They are the same. The new implementation just breakdown the original logic into three functions. The CSRToCOO (without shuffling) is very efficient. The column indices and data arrays are just shallow copy.

tests/compute/test_graph.py
jermainewang: -1 is not being checked anymore in the new implementation because we use `int64_t` and less than operator. In fact, the negative id was not tested correctly even in the previous implementation. This is because `dgl_id_t` is `uin64_t` so -1 simply becomes very large positive integer number, making it redundant.

zheng-da:
We might need to deal with indexing type consistently. otherwise, it looks good.

jermainewang:
> We might need to deal with indexing type consistently. otherwise, it looks good.

If we do restrict the size of int32 graph, then the rule is quite simple: both IdType or int64_t are good and safe.

PR #711, opened 2019-07-15
[NodeFlow] Non-uniform neighbor sampling

BarclayII:
## Description
This PR adds in non-uniform neighbor sampler with unnormalized transition probabilities on edges.

Essentially, it replaces the reserved `node_prob` argument with `transition_prob` which can either be a static vector specified during initialization, or a column name in which case the probabilities are loaded from the frame for each iteration.

## Checklist
- [x] Code is well-documented

## Changes
Most changes take place in
- [ ] `python/dgl/contrib/sampling/sampler.py`
- [ ] `src/graph/sampler.cc`


include/dgl/sampler.h
```diff
@@ -56,13 +56,16 @@ class SamplerOp {
...
-                                        const std::string &edge_type,
-                                        int num_hops, int expand_factor,
-                                        const bool add_self_loop);
+  template<typename ValueType>
+  static NodeFlow NeighborSample(const ImmutableGraph *graph,
+                                 const std::vector<dgl_id_t>& seeds,
+                                 const std::string &edge_type,
+                                 int num_hops, int expand_factor,
+                                 const bool add_self_loop,
+                                 const ValueType *probability);
```
zheng-da: does it matter to accept double floating-points for probability?

python/dgl/contrib/sampling/sampler.py
```diff
@@ -259,10 +259,30 @@ class NeighborSampler(NodeFlowSampler):
...
+           p(u, v) = \frac{w_{u, v}}{\sum_{u', (u', v) \in E} w_{u', v}}
+
+        If neighbor type is "out", then the probability is instead normalized by the sum
+        grouping by source node:
+
+        .. math::
+
+           p(v, u) = \frac{w_{v, u}}{\sum_{u', (v, u') \in E} w_{v, u'}}
+
+        If a str is given, the edge weight will be loaded from the feature column with
```
zheng-da: is it "edge feature column"?

tests/compute/test_sampler.py
```diff
@@ -155,10 +155,53 @@ def test_layer_sampler():
...
+
+    # Test the reverse direction
+    sampler = dgl.contrib.sampling.NeighborSampler(
+        g, 1, 1, 99, 'out', transition_prob='w', seed_nodes=[0])
+    nf = next(iter(sampler))
+
+    assert nf.num_layers == 100
+    for i in range(nf.num_layers):
+        assert nf.layer_size(i) == 1
+        assert nf.layer_parent_nid(i)[0] == 99 - i
```
zheng-da: is it possible to check the sample distribution?

src/graph/sampler.cc
```diff
@@ -766,9 +771,75 @@ DGL_REGISTER_GLOBAL("sampling._CAPI_UniformSampling")
...
+    const IdArray seed_nodes = args[1];
+    const int64_t batch_start_id = args[2];
+    const int64_t batch_size = args[3];
+    const int64_t max_num_workers = args[4];
+    const int64_t expand_factor = args[5];
+    const int64_t num_hops = args[6];
+    const std::string neigh_type = args[7];
+    const bool add_self_loop = args[8];
+
+    const GraphInterface *ptr = static_cast<const GraphInterface *>(ghdl);
```
jermainewang: Consider merge this after #716 ?

src/graph/sampler.cc
```diff
@@ -766,9 +771,75 @@ DGL_REGISTER_GLOBAL("sampling._CAPI_UniformSampling")
...
+    if (probability->dtype.bits == 32) {
+      const float *prob32 = static_cast<const float *>(probability->data);
+      nflows = _CAPI_NeighborSampling(
+          ptr, seed_nodes, batch_start_id, batch_size, max_num_workers,
+          expand_factor, num_hops, neigh_type, add_self_loop, prob32);
+    } else if (probability->dtype.bits == 64) {
+      const double *prob64 = static_cast<const double *>(probability->data);
+      nflows = _CAPI_NeighborSampling(
+          ptr, seed_nodes, batch_start_id, batch_size, max_num_workers,
+          expand_factor, num_hops, neigh_type, add_self_loop, prob64);
```
jermainewang: Looks like you are dispatching implementation based on data types. Check out my implementation in `aten` package (https://github.com/dmlc/dgl/blob/master/src/array/array.cc). You could use a macro. Also, suggest directly use `NDArray` as argument instead of raw `ValueType*` pointer.

BarclayII: Which function are you referring to?  I saw in your implementation involving bit-width (like `Full`) you are still dispatching with if-statements.

src/graph/sampler.cc
BarclayII: Of course.

tests/compute/test_sampler.py
BarclayII: Could you give an example of how to do so?

include/dgl/sampler.h
BarclayII: I think it's OK to support both if the code does not become significantly complicated and the performance does not hurt.

python/dgl/contrib/sampling/sampler.py
BarclayII: Yes

src/graph/sampler.cc
jermainewang: Functions like `Add`, `Sub`.

BarclayII: Done

python/dgl/contrib/sampling/sampler.py
```diff
@@ -300,26 +320,48 @@ def __init__(
...
-            self.batch_size,        # batch size
-            self._num_workers,      # num batches
-            self._expand_factor,
-            self._num_hops,
-            self._neighbor_type,
-            self._add_self_loop))
+        if self._transition_prob is None:
+            prob = None
+        elif isinstance(self._transition_prob, str):
+            prob = self.g.edata[self._transition_prob]
```
jermainewang: Add a check to make sure the edge feature is scalar.

python/dgl/contrib/sampling/sampler.py
```diff
@@ -300,26 +320,48 @@ def __init__(
...
+                self.g.c_handle,
+                self.seed_nodes.todgltensor(),
+                current_nodeflow_index, # start batch id
+                self.batch_size,        # batch size
+                self._num_workers,      # num batches
+                self._expand_factor,
+                self._num_hops,
+                self._neighbor_type,
+                self._add_self_loop,
+                F.zerocopy_to_dgl_ndarray(prob))
```
jermainewang: Can we merge the two? How about if the prob array is an empty array then it is uniform sampling?

src/graph/sampler.cc
```diff
@@ -182,8 +184,11 @@ void GetUniformSample(const dgl_id_t* edge_id_list,
 
 /*
  * Non-uniform sample via ArrayHeap
+ *
+ * \param probability Transition probability on the entire graph, indexed by edge ID
  */
-void GetNonUniformSample(const float* probability,
+template<typename ValueType>
+void GetNonUniformSample(const ValueType* probability,
```
jermainewang: Still suggest use `NDArray` for some of these arguments (IdArray for vid_list etc.). This could save argument like `ver_len` and make it potentially cleaner. Also, could you change dgl_id_t to template argument as well? We might have int32 graph in the future. You could use ATEN_ID_TYPE_SWITCH macro to dispatch the corresponding type.

src/graph/sampler.cc
```diff
@@ -532,14 +538,16 @@ DGL_REGISTER_GLOBAL("nodeflow._CAPI_NodeFlowFree")
     delete nflow;
   });
```
jermainewang: This _CAPI_NodeFlowFree is actually not needed anymore. If I haven't deleted it before, you could delete it here.

src/graph/sampler.cc
```diff
@@ -733,20 +741,18 @@ void BuildCsr(const ImmutableGraph &g, const std::string neigh_type) {
...
-    const int64_t batch_size = args[3];
-    const int64_t max_num_workers = args[4];
-    const int64_t expand_factor = args[5];
-    const int64_t num_hops = args[6];
-    const std::string neigh_type = args[7];
-    const bool add_self_loop = args[8];
-    // process args
-    const GraphInterface *ptr = static_cast<const GraphInterface *>(ghdl);
+template<typename ValueType>
+std::vector<NodeFlow*> _CAPI_NeighborSampling(
```
jermainewang: NeighborSamplingImpl <- _CAPI_NeighborSampling

src/graph/sampler.cc
BarclayII: Actually those are *views* of the `vid_list`.  You can see the usage in L433-L450: what the routines actually took in is something like `val_list[indptr[dst_id]:indptr[dst_id] + ver_len]`.

python/dgl/contrib/sampling/sampler.py
BarclayII: It's done in C.

python/dgl/contrib/sampling/sampler.py
BarclayII: I merged the two in Python and made `_CAPI_NeighborSampling` distinguish the two cases.

BarclayII: Done that.

src/graph/sampler.cc
jermainewang: Replied in the other issue. Basically, for scalar type argument, use int64 is safe. For ndarray, it is type agnostic and the pointer will only be casted in the template implementation.

BarclayII: Then should we have two IdType's `VIDType` and `EIDType`?  Technically a graph with int32 node IDs can still have more than 2B edges (i.e. int64 edge IDs).  The same question is raised in #699 but somehow it is not addressed.

jermainewang: We decided to simplify the situation. If a graph is 32bit, then both its number of nodes and edges will not exceed 2^32. Otherwise, a 64 bit graph is used. This means that all the ndarrays in a graph must have the same bit width, making it easier to maintain. So using one id type is fine.

BarclayII: Oh OK I see what you meant.  Can we do it in another PR?  We need to template-ize layer sampler and NodeFlow construction routines as well.  I feel this is beyond the scope of this PR.

jermainewang: Sure

include/dgl/sampler.h
zheng-da: template in general increases the binary code and compliation time. Also, a large portion of code has to be in the header file. I would suggest to not use template if necessary.

MXNet is an example of overusing templates. I'm hoping that DGL doesn't do the same again.

tests/compute/test_sampler.py
zheng-da: If we sample many times, the frequency of the sampled result will be close to the probability distribution, although it's really tricky to set the epsilon. Have you observed the sampled results manually?

include/dgl/sampler.h
BarclayII: If we were to support multiple data types with the same logic (for instance, int32/int64 graphs, float32/float64 probabilities), then templates seem to be one of the better solutions we have for now.

Other choices include
* [Using `#define`'s all over the place](https://stackoverflow.com/questions/10950828/simulation-of-templates-in-c-for-a-queue-data-type)
* Write your own template (e.g. in Jinja) and render them with a helper script into C code with different data types ([`numpy`](https://github.com/numpy/numpy/blob/master/numpy/core/src/umath/scalarmath.c.src)'s solution).

tests/compute/test_sampler.py
BarclayII: Yes and it looked fine, although I need to put a small delay between samples because we initialize the seed with `time(nullptr)` every time `SampleSubgraph` is called.  I plan to put a better RNG in C to address this issue as well as providing reproducibility in sampling.

include/dgl/sampler.h
zheng-da: Template is definitely better than macro. If you want to support multiple data types, template is definitely a way to go. The question is: do we really need to support both single-precision and double-precision for transition matrix? To support both, we have to put a large amount of code in template. I'm wondering how much supporting both affects the sampling precision?

tests/compute/test_sampler.py
zheng-da: I think no one is working on it. Thanks for picking this task.

include/dgl/sampler.h
BarclayII: There shouldn't be a lot of code in template.  @jermainewang mentioned that the arrays should be in NDArrays instead of raw pointers, and I agree with that.  With the dispatchers in `aten` I think the only place where we would have templates is the `ArrayHeap`.  I'm leaving the refactor to another PR.

tests/compute/test_sampler.py
jermainewang: Great! Take a look at TVM's random engine codes [here](https://github.com/dmlc/tvm/blob/cffb4fba03ea582417e2630bd163bca773756af6/src/contrib/random/mt_random_engine.cc) and [here](https://github.com/dmlc/tvm/blob/cffb4fba03ea582417e2630bd163bca773756af6/src/contrib/random/random.cc). Two things to pay attention to: (1) consider using std <random> which is multi-platform; (2) use thread local entry to handle multi-thread case. Thread local entry acts like a singleton but is thread-safe. Then we can expose API to set seed to python.

include/dgl/sampler.h
jermainewang: @zheng-da, we could solve the binary size problem by using declaration in header file and explicit full specialization in other .cc units. This practice makes it not necessary to inline everything for template functions. For example, suppose you have following:

```c++
template <typename T>
void Foo() {
  // very large function body that you don't want to inline
}

template <typename T>
void Bar() {
  Foo<T>();  // Unfortunately Foo is being used in many other function such as Bar here.
}
```
We could separate the declaration and definition by:
```c++
// foo.h
template <typename T> void Foo();  // only declaration
```
```c++
// foo.cc
template <typename T>
void Foo() {
  // very large function body
}

// explicit specialization for the type that is used
template Foo<int32_t>();
template Foo<int64_t>();
...
```
```c++
// bar.cc
#include "foo.h"
template <typename T>
void Bar() {
  Foo<T>();  // use Foo here will not inline the function.
}
```
This also makes compilation faster and you could even split the template definitions in multiple translation units to parallelize it (the trick used in BinaryOp kernels).

zheng-da: The templated code includes `SampleSubgraph`, `GetNonUniformSample` and `ArrayHeap`, so it's still significant.

I'm not trying to block the PR. I'm just saying that we should use template with caution and don't repeat what MXNet did. If you think supporting both sing precision and double precision are definitely necessary, I'm fine with the current implementation.

@jermainewang I know template specialization helps us to move implementation to .cc file and split the template definitions in multiple files. Are you sure it accelerates compilation? I'm not familiar with how compilers deal with template specialization.

jermainewang: Let push through this. I think we will gradually move sampling algorithm into `aten` namespace in the future. You could view sampling as a type of sparse operators. As a result, there will be some refactoring going on so if it is currently not too much an issue, let's merge this PR so the functionality is there. What do you think?

zheng-da: sounds good to me.

PR #716, opened 2019-07-17
[Refactor] Use object system for all CAPIs

jermainewang:
## Description
The PR refactors our CAPIs to use the new object FFI solution introduced in https://github.com/dmlc/dgl/pull/693 . The principle is following:

* Use shared pointer everywhere in C++. For example, `GraphPtr` is of type `shared_ptr<GraphInterface>`. All the graph operators should accept `GraphPtr` type arguments and return `GraphPtr` type results.
* When returning a graph to python. You could either directly return `GraphPtr` or use `GraphRef` to wrap it and return. They are the same.
* When returning a collection of graphs to python. use `dgl::runtime::List<GraphRef>`.
* On python side, the `dgl.graph_index.GraphIndex` object has been registered as the wrapper class for graph type values. Therefore, you could directly use `dgl.graph_index.GraphIndex` object as the CAPI argument and return type. No internal C handle is exposed.

The new FFI greatly simplifies the CAPI logics (**no more hack of vector type return**). Note that `Nodeflow` now also uses object FFI, and I suggest we change BatchedGraph and Subgraph as well in the future.

Other subtlety:
* Move `GraphInterface::EdgeArray` out of class scope so it can be used by other subclasses.
* Move `GraphInterface::Reverse` out of graph interface as a functional API in GraphOp.
* `LazyObject` -> `Lazy` to avoid naming confusion.
* Merge the two implementation of `GraphOp::DisjointUnion` for mutable and immutable graph into one. It now accepts `vector<GraphPtr>` and returns `GraphPtr`. I think this is cleaner. @yzh119
* Suggest @aksnzhy to take a look at this new practice. I think your `CommunicatorHandler` can also be refactored into `Object`. I didn't touch that part though.
* @VoVAllen Please aware of this change as you might be working on similar stuff.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/include/dgl/graph.h
+++ b/include/dgl/graph.h
@@ -20,20 +20,7 @@ namespace dgl {
 class Graph;
 class GraphOp;
 
-/*!
- * \brief Base dgl graph index class.
- *
- * DGL's graph is directed. Vertices are integers enumerated from zero.
- *
- * Removal of vertices/edges is not allowed. Instead, the graph can only be "cleared"
- * by removing all the vertices and edges.
- *
- * When calling functions supporing multiple edges (e.g. AddEdges, HasEdges),
- * the input edges are represented by two id arrays for source and destination
- * vertex ids. In the general case, the two arrays should have the same length.
- * If the length of src id array is one, it represents one-many connections.
- * If the length of dst id array is one, it represents many-one connections.
- */
+/*! \brief Mutable graph based on adjacency list. */
 class Graph: public GraphInterface {
  public:
   /*! \brief default constructor */
@@ -359,16 +346,6 @@ class Graph: public GraphInterface {
   }
 
   /*!
-   * \brief Reset the data in the graph and move its data to the returned graph object.
-   * \return a raw pointer to the graph object.
-   */
-  GraphInterface *Reset() override {
-    Graph* gptr = new Graph();
-    *gptr = std::move(*this);
-    return gptr;
-  }
-
-  /*!
    * \brief Get the adjacency matrix of the graph.
    *
    * By default, a row of returned adjacency matrix represents the destination
@@ -378,6 +355,9 @@ class Graph: public GraphInterface {
    * \return a vector of three IdArray.
    */
   std::vector<IdArray> GetAdj(bool transpose, const std::string &fmt) const override;
+
+  static constexpr const char* _type_key = "graph.Graph";
+  DGL_DECLARE_OBJECT_TYPE_INFO(Graph, GraphInterface);
 
  protected:
   friend class GraphOp;
@@ -410,6 +390,32 @@ class Graph: public GraphInterface {
   bool is_multigraph_ = false;
   /*! \brief number of edges */
   uint64_t num_edges_ = 0;
+};
+
+/*! \brief Graph reference class */
+class GraphRef : public runtime::ObjectRef {
+ public:
+  /*! \brief empty reference */
+  GraphRef() {}
+  explicit GraphRef(std::shared_ptr<runtime::Object> obj): runtime::ObjectRef(obj) {}
+  const Graph* operator->() const {
+    return static_cast<const Graph*>(obj_.get());
+  }
+  Graph* operator->() {
+    return static_cast<Graph*>(obj_.get());
+  }
+  using ContainerType = Graph;
+
+  /*! \brief Create an empty graph */
+  static GraphRef Create(bool multigraph = false) {
+    return GraphRef(std::make_shared<Graph>(multigraph));
+  }
+
+  /*! \brief Create from coo */
+  static GraphRef CreateFromCOO(
+      int64_t num_nodes, IdArray src_ids, IdArray dst_ids, bool multigraph = false) {
+    return GraphRef(std::make_shared<Graph>(src_ids, dst_ids, num_nodes, multigraph));
+  }
 };
 
 }  // namespace dgl
--- a/include/dgl/graph.h
+++ b/include/dgl/graph.h
@@ -20,7 +20,7 @@ namespace dgl {
 
 class Graph;
 class GraphOp;
-typedef std::shared_ptr<Graph> MGraphPtr;
+typedef std::shared_ptr<Graph> MutableGraphPtr;
 
 /*! \brief Mutable graph based on adjacency list. */
 class Graph: public GraphInterface {
@@ -350,12 +350,12 @@ class Graph: public GraphInterface {
   std::vector<IdArray> GetAdj(bool transpose, const std::string &fmt) const override;
 
   /*! \brief Create an empty graph */
-  static MGraphPtr Create(bool multigraph = false) {
+  static MutableGraphPtr Create(bool multigraph = false) {
     return std::make_shared<Graph>(multigraph);
   }
 
   /*! \brief Create from coo */
-  static MGraphPtr CreateFromCOO(
+  static MutableGraphPtr CreateFromCOO(
       int64_t num_nodes, IdArray src_ids, IdArray dst_ids, bool multigraph = false) {
     return std::make_shared<Graph>(src_ids, dst_ids, num_nodes, multigraph);
   }
--- a/include/dgl/graph.h
+++ b/include/dgl/graph.h
@@ -12,6 +12,7 @@
 #include <cstdint>
 #include <utility>
 #include <tuple>
+#include <memory>
 
 #include "graph_interface.h"
 
@@ -19,6 +20,7 @@ namespace dgl {
 
 class Graph;
 class GraphOp;
+typedef std::shared_ptr<Graph> MGraphPtr;
 
 /*! \brief Mutable graph based on adjacency list. */
 class Graph: public GraphInterface {
@@ -293,15 +295,6 @@ class Graph: public GraphInterface {
   Subgraph EdgeSubgraph(IdArray eids, bool preserve_nodes = false) const override;
 
   /*!
-   * \brief Return a new graph with all the edges reversed.
-   *
-   * The returned graph preserves the vertex and edge index in the original graph.
-   *
-   * \return the reversed graph
-   */
-  GraphPtr Reverse() const override;
-
-  /*!
    * \brief Return the successor vector
    * \param vid The vertex id.
    * \return the successor vector
@@ -356,8 +349,16 @@ class Graph: public GraphInterface {
    */
   std::vector<IdArray> GetAdj(bool transpose, const std::string &fmt) const override;
 
-  static constexpr const char* _type_key = "graph.Graph";
-  DGL_DECLARE_OBJECT_TYPE_INFO(Graph, GraphInterface);
+  /*! \brief Create an empty graph */
+  static MGraphPtr Create(bool multigraph = false) {
+    return std::make_shared<Graph>(multigraph);
+  }
+
+  /*! \brief Create from coo */
+  static MGraphPtr CreateFromCOO(
+      int64_t num_nodes, IdArray src_ids, IdArray dst_ids, bool multigraph = false) {
+    return std::make_shared<Graph>(src_ids, dst_ids, num_nodes, multigraph);
+  }
 
  protected:
   friend class GraphOp;
@@ -390,32 +391,6 @@ class Graph: public GraphInterface {
   bool is_multigraph_ = false;
   /*! \brief number of edges */
   uint64_t num_edges_ = 0;
-};
-
-/*! \brief Graph reference class */
-class GraphRef : public runtime::ObjectRef {
- public:
-  /*! \brief empty reference */
-  GraphRef() {}
-  explicit GraphRef(std::shared_ptr<runtime::Object> obj): runtime::ObjectRef(obj) {}
-  const Graph* operator->() const {
-    return static_cast<const Graph*>(obj_.get());
-  }
-  Graph* operator->() {
-    return static_cast<Graph*>(obj_.get());
-  }
-  using ContainerType = Graph;
-
-  /*! \brief Create an empty graph */
-  static GraphRef Create(bool multigraph = false) {
-    return GraphRef(std::make_shared<Graph>(multigraph));
-  }
-
-  /*! \brief Create from coo */
-  static GraphRef CreateFromCOO(
-      int64_t num_nodes, IdArray src_ids, IdArray dst_ids, bool multigraph = false) {
-    return GraphRef(std::make_shared<Graph>(src_ids, dst_ids, num_nodes, multigraph));
-  }
 };
 
 }  // namespace dgl
--- a/include/dgl/graph_interface.h
+++ b/include/dgl/graph_interface.h
@@ -57,7 +57,7 @@ typedef struct {
 
 // forward declaration
 struct Subgraph;
-class BaseGraphRef;
+class GraphRef;
 class GraphInterface;
 typedef std::shared_ptr<GraphInterface> GraphPtr;
 
@@ -301,15 +301,6 @@ class GraphInterface : public runtime::Object {
   virtual Subgraph EdgeSubgraph(IdArray eids, bool preserve_nodes = false) const = 0;
 
   /*!
-   * \brief Return a new graph with all the edges reversed.
-   *
-   * The returned graph preserves the vertex and edge index in the original graph.
-   *
-   * \return the reversed graph
-   */
-  virtual GraphPtr Reverse() const = 0;
-
-  /*!
    * \brief Return the successor vector
    * \param vid The vertex id.
    * \return the successor vector iterator pair.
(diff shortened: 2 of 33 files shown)


python/dgl/graph_index.py
```diff
@@ -1,43 +1,44 @@
...
-    def __del__(self):
-        """Free this graph index object."""
-        if hasattr(self, '_handle'):
-            _CAPI_DGLGraphFree(self._handle)
+    def __new__(cls):
+        obj = ObjectBase.__new__(cls)
+        obj._multigraph = None  # python-side cache of the flag
+        obj._readonly = None  # python-side cache of the flag
+        obj._cache = {}
+        return obj
```
zheng-da: this method is called in the graph index CAPI?

python/dgl/nodeflow.py
```diff
@@ -293,6 +335,7 @@ def map_to_parent_nid(self, nid):
             The parent node id array.
         """
         nid = utils.toindex(nid)
+        # TODO(minjie): should not directly use []
```
zheng-da: why?

include/dgl/graph_interface.h
```diff
@@ -353,6 +345,32 @@ class GraphInterface {
    * \return a vector of IdArrays.
    */
   virtual std::vector<IdArray> GetAdj(bool transpose, const std::string &fmt) const = 0;
+
+  static constexpr const char* _type_key = "graph.Graph";
+  DGL_DECLARE_OBJECT_TYPE_INFO(GraphInterface, runtime::Object);
+};
+
+/*! \brief Base graph reference */
+class GraphRef : public runtime::ObjectRef {
```
VoVAllen: What's the difference between GraphRef/ GraphPtr/ GraphInterface?

include/dgl/immutable_graph.h
```diff
@@ -517,7 +500,7 @@ class COO : public GraphInterface {
   aten::COOMatrix adj_;
 
   /*! \brief whether the graph is a multi-graph */
-  LazyObject<bool> is_multigraph_;
+  Lazy<bool> is_multigraph_;
```
VoVAllen: Why we need to make it lazy here?

python/dgl/graph_index.py
VoVAllen: What's these flag for?

src/graph/graph_op.cc
```diff
@@ -38,199 +42,208 @@ class RangeIter : public std::iterator<std::input_iterator_tag, dgl_id_t> {
...
 
-    for (dgl_id_t i = 0; i < g_num_edges; ++i) {
-      edge_ids[cum_num_edges + i] = g_edge_ids[i] + cum_num_edges;
+    indptr[0] = 0;
+    dgl_id_t cum_num_nodes = 0;
+    dgl_id_t cum_num_edges = 0;
+    for (auto g : graphs) {
+      ImGraphPtr gr = std::dynamic_pointer_cast<ImmutableGraph>(g);
+      CHECK(gr) << "All the input graphs should be immutable graphs.";
+      // TODO(minjie): why in csr?
```
VoVAllen: Just for simplicity, because the construction of immutable graph uses incsr.

src/graph/immutable_graph.cc
```diff
@@ -467,53 +474,177 @@ std::vector<IdArray> ImmutableGraph::GetAdj(bool transpose, const std::string &f
   }
 }
 
-ImmutableGraph ImmutableGraph::ToImmutable(const GraphInterface* graph) {
-  const ImmutableGraph* ig = dynamic_cast<const ImmutableGraph*>(graph);
+ImGraphPtr ImmutableGraph::CreateFromCSR(
+    IdArray indptr, IdArray indices, IdArray edge_ids, const std::string &edge_dir) {
+    CSRPtr csr(new CSR(indptr, indices, edge_ids));
```
VoVAllen: Better to use `make_shared` instead of `new`?

src/graph/graph_op.cc
```diff
@@ -38,199 +42,208 @@ class RangeIter : public std::iterator<std::input_iterator_tag, dgl_id_t> {
...
-    g_indptr[0] = 0;
-    for (int l = start_pos + 1; l < end_pos + 1; ++l) {
-      g_indptr[l - start_pos] = indptr[l] - indptr[start_pos];
-    }
 
-    for (int j = indptr[start_pos]; j < indptr[end_pos]; ++j) {
-      g_indices[j - idoff] = indices[j] - cumsum[i];
+  std::vector<GraphPtr> rst;
+  if (IsMutable(batched_graph)) {
+    // Input is a mutable graph. Partition it into several multiple graphs.
```
VoVAllen: Mutable instead of multiple?

src/graph/graph_op.cc
```diff
@@ -38,199 +42,208 @@ class RangeIter : public std::iterator<std::input_iterator_tag, dgl_id_t> {
...
+      // push to rst
+      rst.push_back(mg);
+      // update offset
+      CHECK_EQ(rst[i]->NumVertices(), sizes_data[i]);
+      CHECK_EQ(rst[i]->NumEdges(), num_edges);
+      node_offset += sizes_data[i];
+      edge_offset += num_edges;
     }
+  } else {
+    // Input is an mutable graph. Partition it into several multiple graphs.
```
VoVAllen: Input is an immutable graph

python/dgl/nodeflow.py
BarclayII: Because this syntax doesn't work with Chainer variables.  I fixed these in my Chainer PR.

python/dgl/graph_index.py
jermainewang: Nop. This is called on python side when a CAPI returns an Object. See here: https://github.com/dmlc/dgl/blob/master/python/dgl/_ffi/_ctypes/object.py#L20

The flags indicate whether the graph is a multigraph or a readonly graph. We could use CAPI to get the flag. Here it also saves a copy (a bool value) in python.

include/dgl/graph_interface.h
jermainewang: `GraphInterface` is an interface class. GraphPtr is `std::shared_ptr<GraphInterface>`. GraphRef is equivalent to GraphPtr except that it can be used in container. The general practice I found useful is that:

* Use shared pointer (i.e, GraphPtr) in C++ implementation.
* When returning GraphPtr to python, wrap it as GraphRef. Similarly, when getting the argument from python, also use GraphRef.
```c++
DGL_REGISTER_GLOBAL("SomeCAPI")
.set_body([] (DGLArgs args, DGLRetValue *rv) {
  GraphRef g = args[0];      // use graph ref here
  GraphPtr sptr = g.sptr();  // get the shared pointer
  GraphPtr retptr = sptr->SomeCall();   // C++ functions all work on GraphPtr
  *rv = GraphRef(retptr);    // return graph ref
});
```

Side notes:
* `GraphRef` also implements the pointer operator `->`, so you could use it just like a `GraphPtr`.
* Returning `GraphPtr` to python is also allowed. It is the same as returning GraphRef. However, I suggest we keep it consistent.

include/dgl/immutable_graph.h
jermainewang: Yes. Checking whether the graph is a multigraph is costly, and since the graph is immutable, we could postpone the check till the first time the flag is requested.

src/graph/immutable_graph.cc
jermainewang: Haha, I also tried that. I found that it does not work for private constructor like this (https://github.com/dmlc/dgl/blob/master/include/dgl/immutable_graph.h#L1033).

src/graph/graph_op.cc
jermainewang: I left a note here because the immutable graph might have only out_csr. In this case, the function will first convert it to in_csr, which is somewhat redundant because we could directly batch out_csr and construct the result graph using out_csr. However, I don't know what is a better solution here because this means the function needs to know the internal state of an immutable graph object, breaking the encapsulation.

jermainewang:
Let me explain a little more about what's inside of the Object FFI. We know that when passing a complex data structure from/to python, we need to use `void*`. This is essentially what we have done in our previous code. This practice has at least three issues:

* When defining a C API, we need to continuously write static conversion:
  ```c++
  void* ptr = args[0];
  GraphInterface* gptr = static_cast<GraphInterface*>(ptr);
  // do some stuff
  ```
* No runtime type checking. The static cast is not safe.
* Cannot use shared pointer. This means when a structure (e.g. a Graph) is returned to python, it cannot be used any more in C++. That's why we have a strange `Reset` function in `GraphInterface` before.

To solve these issues, the Object FFI does following:
* When defining a structure, you also need to provide a literal name (i.e, a string) that should be unique and is used as the type name. That's why you need to use the `DGL_DECLARE_OBJECT_TYPE_INFO` macro and define `_type_key` static member.
* On the python side, you could define a python class that **maps** to the c++ structure by using the same name in `register_object(name)`.
* Object FFI lets you use shared pointer everywhere and allows you to return a shared pointer to python. Suppose we have a shared pointer `std::shared_ptr<SomeObject> ptr`.
  - When running `*rv = ptr;`:
    * it calls the [conversion overload](https://github.com/dmlc/dgl/blob/master/include/dgl/runtime/packed_func.h#L738), which is defined [here](https://github.com/dmlc/dgl/blob/master/include/dgl/packed_func_ext.h#L139)
    * It basically creates a pointer to the shared pointer (i.e, `shared_ptr<SomeObject>*`), converts it to `void*` and returns it to python. The type handler type is `kObjectHandle`.
    * On python side, the function finds that the return is an object handle and calls the corresponding object return value maker [here](https://github.com/dmlc/dgl/blob/master/python/dgl/_ffi/_ctypes/function.py#L193).
    * The [return value maker function](https://github.com/dmlc/dgl/blob/master/python/dgl/_ffi/_ctypes/object.py#L20) first looks up the type string and then looks up the python class constructor you registered under the same name. That's why you need to make sure that you use the same name for `register_object` and `_type_key`. It then calls the `__new__` function of that class to create the object and return.
    * The end-to-end result is that you see a `GraphIndex` object returned directly from a C API call.
  - When running `_SOME_CAPI_(obj)` on python:
    * The FFI first checks the type code of the argument and understands that it is an object handler. [here](https://github.com/dmlc/dgl/blob/master/python/dgl/_ffi/_ctypes/function.py#L97)
    * It then passes the `void*` handler to C++.
    * On C++ side, when `std::shared_ptr<SomeObject> ptr = args[0]` is called, it internally perform a runtime check to see whether the type string matches or not, and then does a static_cast that converts the `void*` to `std::shared_ptr<SomeObject>*` and returns the value. The actual code is more complicated because the type checking needs to handle type hierarchy but the general flow is the same. [code](https://github.com/dmlc/dgl/blob/master/include/dgl/packed_func_ext.h#L109).

To summarize, the FFI
* implements a reflection mechanism when an object is returned from C to python.
* implements runtime type checking when an object is passed from python to C.
* uses pointer to shared pointer as `void*` handler, which is safe on both sides.

@BarclayII @zheng-da @VoVAllen Hope this helps.

include/dgl/immutable_graph.h
VoVAllen: I see. I didn't understand the implementation before. It seems like a cache more than a lazy initialized object to me.

include/dgl/graph_interface.h
VoVAllen: Thanks for super clear explanation. We could add it to our developer guides also.

VoVAllen:
I found NDArray/DLTensor doesn't have corresponding ObjectRef, which prevents directly use List with ffi. Do we need to add this also?

VoVAllen:
Also I found List<std::string> is not supported, shall we add this also?

src/graph/network.cc
```diff
@@ -169,9 +169,9 @@ DGL_REGISTER_GLOBAL("network._CAPI_ReceiverRecvSubgraph")
                                           &(nf->layer_offsets),
                                           &(nf->flow_offsets));
       nf->graph = GraphPtr(new ImmutableGraph(csr, nullptr));
-      std::vector<NodeFlow*> subgs(1);
-      subgs[0] = nf;
-      *rv = WrapVectorReturn(subgs);
+      List<NodeFlow> subgs;
+      subgs.push_back(nf);
+      *rv = subgs;
```
aksnzhy: If we have only one nodeflow here, can we still use `*rv = &nf` here?

include/dgl/nodeflow.h
```diff
@@ -42,6 +43,31 @@ struct NodeFlow {
    * \brief The edge mapping from the NodeFlow graph to the parent graph.
    */
   IdArray edge_mapping;
+
+  static constexpr const char* _type_key = "graph.NodeFlow";
+  DGL_DECLARE_OBJECT_TYPE_INFO(NodeFlowObject, runtime::Object);
+};
+
+class NodeFlow : public runtime::ObjectRef {
```
VoVAllen: Seems we can use macro to simplify this with macro ([code from tvm](https://github.com/dmlc/tvm/blob/415a270def327c4e1beeea75ce7f296e4683a5c0/include/tvm/base.h#L48)

include/dgl/immutable_graph.h
```diff
@@ -930,100 +917,39 @@ class ImmutableGraph: public GraphInterface {
...
-    }
-  }
+  /*!
+   * \brief Return a new graph with all the edges reversed.
+   *
+   * The returned graph preserves the vertex and edge index in the original graph.
+   *
+   * \return the reversed graph
+   */
+  ImGraphPtr Reverse() const;
```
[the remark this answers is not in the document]
BarclayII: `func() const` are functions that don't change internal state of the object.
`static` are just for class methods.  I'm also not sure what is the general guideline of member v.s. static v.s. global functions.  [Google has a C++ guideline](https://google.github.io/styleguide/cppguide.html#Nonmember,_Static_Member,_and_Global_Functions) but we don't seem to follow it anyway.

src/graph/network.cc
```diff
@@ -84,12 +82,14 @@ DGL_REGISTER_GLOBAL("network._CAPI_SenderSendSubgraph")
...
-    GraphHandle ghandle = args[2];
+    // TODO(minjie): could simply use NodeFlow nf = args[2];
+    GraphRef g = args[2];
     const IdArray node_mapping = args[3];
     const IdArray edge_mapping = args[4];
     const IdArray layer_offsets = args[5];
     const IdArray flow_offsets = args[6];
-    ImmutableGraph *ptr = static_cast<ImmutableGraph*>(ghandle);
+    auto ptr = std::dynamic_pointer_cast<ImmutableGraph>(g.sptr());
+    CHECK(ptr) << "only immutable graph is allowed in send/recv";
```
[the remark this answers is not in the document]
BarclayII: Because we haven't implemented send/recv on mutable graph yet?

src/graph/network.cc
jermainewang: Just use `*rv=nf`. Nodeflow has inherited `ObjectRef` so it is now a reference object (equivalent to a shared pointer).

include/dgl/immutable_graph.h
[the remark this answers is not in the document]
jermainewang: We should try follow the Google coding style. I agree that in general we should avoid grouping a bunch of static functions such as in `GraphOp`. The reason it happened (and perhaps we should fix it) is that some operations are more efficient if they can directly access the internal state of a graph object. Again, this is bad practice as it breaks the encapsulation. We should get rid of the `GraphOp` bundle class. `Reverse` here is similar because it needs to know whether CSR or COO has been materialized to choose a more efficient implementation. I think a better way to handle this is to add methods to indicate whether CSR or COO has been materialized so we could move the logic outside and make `Reverse` a global function. Let's do that gradually in the following PRs.

src/graph/immutable_graph.cc
```diff
@@ -467,53 +474,177 @@ std::vector<IdArray> ImmutableGraph::GetAdj(bool transpose, const std::string &f
...
+    return ImGraphPtr(new ImmutableGraph(csr, nullptr));
+  } else if (edge_dir == "out") {
+    return ImGraphPtr(new ImmutableGraph(nullptr, csr));
+  } else {
+    LOG(FATAL) << "Unknown edge direction: " << edge_dir;
+    return ImGraphPtr();
+  }
+}
+
+ImGraphPtr ImmutableGraph::CreateFromCSR(
```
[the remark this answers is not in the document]
jermainewang: `CSRPtr` is a shared pointer, so yes. What's your concern for distributed training?

src/graph/network.cc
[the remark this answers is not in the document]
jermainewang: That's correct.

include/dgl/nodeflow.h
jermainewang: Great find.

include/dgl/immutable_graph.h
jermainewang: yes

include/dgl/nodeflow.h
jermainewang: Done with the macro.

include/dgl/immutable_graph.h
```diff
@@ -23,6 +23,9 @@ class COO;
 typedef std::shared_ptr<CSR> CSRPtr;
 typedef std::shared_ptr<COO> COOPtr;
 
+class ImmutableGraph;
+typedef std::shared_ptr<ImmutableGraph> ImGraphPtr;
```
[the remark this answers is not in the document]
jermainewang: Done

include/dgl/graph.h
```diff
@@ -12,28 +12,17 @@
...
 #include <tuple>
+#include <memory>
 
 #include "graph_interface.h"
 
 namespace dgl {
 
 class Graph;
 class GraphOp;
+typedef std::shared_ptr<Graph> MGraphPtr;
```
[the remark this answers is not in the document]
jermainewang: Done. Still leave `Graph` as it is.

src/graph/graph_op.cc
jermainewang: Done.

src/graph/graph_op.cc
jermainewang: Done.

jermainewang: Done

src/graph/graph_op.cc
jermainewang: Done

Issue #668, opened 2019-06-17
Support for chainer?


cmpute:
## 🚀 Feature

Support chainer as backend

## Motivation

I'm a chainer user and it would be great if chainer is supported as backend. Usage of Chainer is quite similar to Pytorch, so I think it would not be that hard to port it.

## Additional context

[Chainer Repo](https://github.com/chainer/chainer)
[Cupy Repo](https://github.com/cupy/cupy)

jermainewang:
Hi, thank you for the suggestion. We do love to have more backends so more users can use DGL. One question, does Chainer support [DLPack](https://github.com/dmlc/dlpack) protocol? It allows framework-specific tensors to expose their raw memory so DGL can perform computation on it in C.

cmpute:
Sorry, I'm not sure about this. Chainer is built based on numpy and cupy. I think this could be answered by chainer developers.

@kmaehashi Sorry to quoting you. I hope this won't makes you too many troubles. Could you kindly help with this issue? Thanks!

BarclayII:
Seems that Chainer tensors are just numpy and cupy arrays, both of which support DLPack (for cupy, see https://github.com/cupy/cupy/blob/master/docs/source/reference/interoperability.rst#dlpack)

BarclayII:
Seems that Chainer tensors are just numpy and cupy arrays, both of which support DLPack (for cupy, see https://github.com/cupy/cupy/blob/master/docs/source/reference/interoperability.rst#dlpack).

@cmpute Could you try implementing the Chainer backend?  Since DGL is platform-agnostic, all you need to do to support a new backend is to create a directory `python/dgl/backend/chainer` and fill in the implementations.  You could refer to `python/dgl/backend/pytorch/tensor.py` to see what are the needed functions and provide the Chainer equivalents accordingly.

cmpute:
I can have a try if time permitted. However I'm not working on graph learning recently, the progress may not be much

BarclayII:
> I can have a try if time permitted. However I'm not working on graph learning recently, the progress may not be much

No problem.  I am currently working on Chainer support now.  It should take a few days.

yzh119:
@BarclayII for pooling layer we need more backend functions, please check https://github.com/dmlc/dgl/pull/669/files, thanks

PR #720, opened 2019-07-19
[Backend] Chainer support

BarclayII:
## Description
Chainer support as per request in #668.  Turns out to be a lot more effort that I expected.

In particular, zerocopy of Chainer variables to DGL tensors is not very good.  Maybe I missed the correct way to implement this, but here are my issues:

* Chainer CPU variables directly wrap around numpy NDArrays, and I could not make zerocopying between numpy and DGL tensors work.
* As a result I have to resort to old-style Chainer custom layers, which is not a big deal since the primary benefit of new-style layers is second-order gradients, which is OK for now since we fuse the gradient operators anyway.
* Another consequence is that I could not get shared memory to work: DGL frees the node features and edge features in shared memory right after `zerocopy_from_dgl_ndarray`.

The issues are commented in `python/dgl/backend/chainer/tensor.py`.

Other changes include (lots of) rewrites in unit tests, replacing tensor operators with backend functions and numpy functions.

@cmpute I may include one or two examples for Chainer in the (potentially far) future, but it would be greatly appreciated if you could try it out after this PR is merged and maybe contribute an example.

@yzh119 I suspect that this PR would be merged earlier than your pooling PR, so please ping me if you need any help to implement the Chainer equivalents.

### Minor change

Added a `return_edge_id` optional argument in `DGLGraph.adjacency_matrix_scipy()` method which defaults to True.  If False, then 1's are returned.

The reason to do this is that if edge IDs are returned, then one of the nonzero entries will be actually 0, which may be confusing.

## Checklist
- [x] Code is well-documented
- [x] New CI image is pushed


yzh119:
I understand, btw you are crazy...

cmpute:
Thanks for you wonderful work!.. I will see whether I can test it

python/dgl/backend/backend.py
```diff
@@ -866,8 +866,8 @@ def zerocopy_from_dgl_ndarray(input):
 # kernels (see kernel.py), and plug into tensor framework using custom op
 # extensions.
 
-def binary_reduce(reducer, binary_op, graph, lhs, rhs, lhs_data, rhs_data,
-                  out_size, lhs_map, rhs_map, out_map):
+def binary_reduce(reducer, op, G, A_target, B_target, A, B,
```
yzh119: I think the argument names are the same as c++ backend, should we refactor c++ impls too?

python/dgl/backend/chainer/tensor.py
```diff
@@ -0,0 +1,536 @@
...
+    num_to_squeeze = len(grad_shape) - len(in_shape)
+    # pad in_shape
+    in_shape = (1,) * num_to_squeeze + in_shape
+    reduce_idx = np.nonzero(np.array(grad_shape) - np.array(in_shape))[0]
+    reduce_idx += 1 # skip batch dim
+    grad = grad.sum(axis=tuple(reduce_idx), keepdims=True)
+    return grad.reshape(shape)
+
+
+def sync():
```
yzh119: It looks like a unique operation just for chainer, should we incorporate this?

python/dgl/backend/chainer/utils.py
```diff
@@ -0,0 +1,24 @@
+import numpy as np
+from chainer.backend import get_array_module as _get_array_module
+try:
+    import cupy
```
yzh119: Would this affect the requirements of DGL?

python/dgl/graph.py
```diff
@@ -3045,7 +3045,7 @@ def edge_subgraph(self, edges, preserve_nodes=False):
         sgi = self._graph.edge_subgraph(induced_edges, preserve_nodes=preserve_nodes)
         return subgraph.DGLSubGraph(self, sgi)
 
-    def adjacency_matrix_scipy(self, transpose=False, fmt='csr'):
+    def adjacency_matrix_scipy(self, transpose=False, fmt='csr', return_edge_ids=True):
```
yzh119: I agree, and I think it should defaults to False.

python/dgl/graph_index.py
```diff
@@ -562,7 +562,7 @@ def edge_subgraph(self, e, preserve_nodes=False):
         return SubgraphIndex(gidx, self, induced_nodes, e)
 
     @utils.cached_member(cache='_cache', prefix='scipy_adj')
-    def adjacency_matrix_scipy(self, transpose, fmt):
+    def adjacency_matrix_scipy(self, transpose, fmt, return_edge_ids=True):
```
yzh119: The same as in `graph.py`.

tests/backend/chainer/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+def array_equal(a, b):
+    return (a.data == b.data).all().item()
+
+def allclose(a, b, rtol=1e-4, atol=1e-4):
+    return get_array_module(a, b).allclose(a.data, b.data, rtol, atol)
+
+def randn(shape):
+    return chainer.as_variable(np.random.randn(*shape).astype('float32'))
+
+def attach_grad(x):
```
yzh119: What's this test used for?

tests/compute/test_basics.py
```diff
@@ -124,6 +127,8 @@ def _pfc(x):
     v = F.tensor([3, 4, 5])
     assert _pfc(g.edges[u, v].data['l']) == [1., 1., 1.]
 
+@unittest.skipIf(dgl.env.get_backend() == 'chainer',
+                 'Chainer does not support head gradients')
```
yzh119: What do you mean by head gradients?

tests/compute/test_kernel.py
```diff
@@ -69,7 +74,7 @@ def _test(red):
             g.update_all(fn.copy_src(src='u', out='m'),
                          builtin[red](msg='m', out='r1'))
             r1 = g.ndata['r1']
-            F.backward(r1.sum())
+            F.backward(F.reduce_sum(r1))
```
yzh119: I suppose `reduce_sum` means taking the sum of all elements?

tests/compute/test_nodeflow.py
```diff
@@ -98,7 +98,7 @@ def check_basic(g, nf):
     for i in range(1, nf.num_layers):
         in_deg = nf.layer_in_degree(i)
         out_deg = nf.layer_out_degree(i - 1)
-        assert F.asnumpy(F.sum(in_deg, 0) == F.sum(out_deg, 0))
+        assert F.asnumpy(in_deg).sum(0) == F.asnumpy(out_deg).sum(0)
```
yzh119: It seesm these two impls are equivalent, why converting to numpy first?

python/dgl/graph.py
BarclayII: Do we have existing code that depend on this behavior?

tests/compute/test_basics.py
BarclayII: `loss.backward(blah)`

tests/compute/test_nodeflow.py
BarclayII: Chainer variables don't support things like `a == b`.

tests/compute/test_kernel.py
BarclayII: Yes

python/dgl/backend/chainer/utils.py
BarclayII: Yes; Chainer could install with or without CuPy.

python/dgl/backend/chainer/tensor.py
BarclayII: You mean MXNet?

python/dgl/backend/backend.py
BarclayII: It could be another separate PR I think...  Lots of stuff to change there.

python/dgl/graph.py
yzh119: My transformer depend on this behavior and I would refactor the code.
Some unittests may depend on this function but I guess they expect this function to return all 1s.

tests/compute/test_nodeflow.py
yzh119: I see...

tests/compute/test_kernel.py
yzh119: OK, then I would refactor my code for pooling layers following the same naming style.

python/dgl/backend/chainer/tensor.py
```diff
@@ -0,0 +1,536 @@
...
+
+def zerocopy_from_dgl_ndarray(input):
+    return chainer.as_variable(_zerocopy_from_dgl_ndarray(input))
+
+
+# TODO(quan):
+# Currently I could not unify the NumPy-DLPack and CuPy-DLPack interaction
+# logic, so I resorted to old-style Chainer functions.
+# The new array backend ChainerX currently does not support DLPack.
+class BinaryReduce(chainer.Function):
```
jermainewang: @ylfdq1118 I highly suggest we could further separate the logic into frame-specific and frame-agnostic part. For example, I feel what we need is an adaptor class `backend.Function` which unifies `torch.autograd.Function`, `mxnet.autograd.Function`, `chainer.Function`. By doing so, we could put this `BinaryReduce` class outside of backend, otherwise it's just too much of copy-pasting.

jermainewang:
> * Chainer CPU variables directly wrap around numpy NDArrays, and I could not make zerocopying between numpy and DGL tensors work.

Investigated some solutions. Although numpy has a FFI to support from/to ctype array, it does not handle memory ownership like DLPack does. Thus, in order to support zerocopy between numpy and DLPack, we need to use [numpy's C library](https://docs.scipy.org/doc/numpy-1.13.0/reference/c-api.array.html) (see [pytorch codes](https://github.com/pytorch/pytorch/blob/master/torch/csrc/utils/tensor_numpy.cpp#L74) for usage). It adds an extra dependency during compilation but it looks manageable. What do you think? Shall we take that route?

> * Another consequence is that I could not get shared memory to work: DGL frees the node features and edge features in shared memory right after zerocopy_from_dgl_ndarray.

Could you be more specific about this issue? One side note. @zheng-da and I had a discussion before and we both think that shared memory should be treated as one of the device types. We also raised an issue (https://github.com/dmlc/dlpack/issues/41) in DLPack to collect some feedback.

BarclayII:
> Investigated some solutions. Although numpy has a FFI to support from/to ctype array, it does not handle memory ownership like DLPack does. Thus, in order to support zerocopy between numpy and DLPack, we need to use [numpy's C library](https://docs.scipy.org/doc/numpy-1.13.0/reference/c-api.array.html) (see [pytorch codes](https://github.com/pytorch/pytorch/blob/master/torch/csrc/utils/tensor_numpy.cpp#L74) for usage). It adds an extra dependency during compilation but it looks manageable. What do you think? Shall we take that route?

By asking "Shall we take that route" you mean you have other solutions available?

> Could you be more specific about this issue? One side note. @zheng-da and I had a discussion before and we both think that shared memory should be treated as one of the device types. We also raised an issue ([dmlc/dlpack#41](https://github.com/dmlc/dlpack/issues/41)) in DLPack to collect some feedback.

The issue is in `_move_data_to_shared_mem_array` in `python/dgl/contrib/graph_store.py`.
The function is supposed to return a framework-specific *view* of the DGL array in shared memory, meaning that the object it returns is a, say, PyTorch tensor, with its contents residing in shared memory.  However, since zerocopying from DGL arrays to numpy does not work yet, the resulting Chainer variable would have its memory independent of the DGL array in shared memory.  So after returning, the DGL array would be recycled.

As per device type, I think a better generalization would be having **file** as a device type.  Because technically we can store tensors on any storage as long as it support random access (e.g. shared memory, file on a disk, or even sockets).

BarclayII:
Closing this since the Chainer team is migrating to PyTorch.

Issue #714, opened 2019-07-17
[Feature Request] Frame for Graph Level Data


MilkshakeForReal:
## Motivation
If I have some global information about the graph, id for example, how do I add it to the DGLGraph object?

## Pitch

```
g = dgl.DGLGraph()
g.gdata['xx'] = xxx
```

mufeili:
Simply try `g.xx=xxx`?

MilkshakeForReal:
> Simply try `g.xx=xxx`?

Yes it works, but the problem is that if I run `print(g)` the attributions would not show up

MilkshakeForReal:
One can do `dir(g)` to see the additional data but there are a bunch of attributions, it's very hard to check which one is what I call `gdata`

mufeili:
@MilkshakeForReal That's a good point. Yeah, we should probably support this.

jermainewang:
Hi @MilkshakeForReal @mufeili @VoVAllen, I created an RFC for this feature. Let's move to that thread for discussion.

PR #728, opened 2019-07-25
[Feature] Data format

VoVAllen:
## Description
Data format

## Checklist
- [ ] Code is well-documented

## Changes

```
The storage structure is
{
  // MetaData Section
  uint64_t kDGLSerializeMagic
  uint64_t kVersion
  uint64_t GraphType
  ** Reserved Area till 4kB **
  dgl_id_t num_graphs
  vector<dgl_id_t> graph_indices (start address of each graph)
  vector<int64_t> nodes_num_list (list of number of nodes for each graph)
  vector<int64_t> edges_num_list (list of number of edges for each graph)
  vector<GraphData> graph_datas;
}
Storage of GraphData is
{
  // Everything uses in csr
  NDArray indptr
  NDArray indices
  NDArray edge_ids
  vector<pair<string, NDArray>> node_tensors;
  vector<pair<string, NDArray>> edge_tensors;
}
```


VoVAllen:
@BarclayII Could you take a look on why my PR failed to build on windows? Thanks a lot.

python/dgl/graph_serialize.py
```diff
@@ -0,0 +1,63 @@
...
+        return _CAPI_MakeGraphData(ghandle, node_tensors, edge_tensors)
+
+    def get_graph(self):
+        """Get DGLGraph from GraphData"""
+        ghandle = _CAPI_GDataGraphHandle(self)
+        g = DGLGraph(graph_data=ghandle, readonly=True)
+        node_tensors_items = _CAPI_GDataNodeTensors(self).items()
+        edge_tensors_items = _CAPI_GDataEdgeTensors(self).items()
+        for k, v in node_tensors_items:
+            g.ndata[k] = F.zerocopy_from_dgl_ndarray(_api_internal._ValueGet(v))
```
jermainewang: I forgot to expose the `Value` class in python. It is fixed in PR #725 , [here](https://github.com/dmlc/dgl/blob/f3c417c1b000d930196322dd14531efa08f7564a/python/dgl/container.py#L73), which means you could directly use `v.data`.

python/dgl/graph_serialize.py
```diff
@@ -0,0 +1,63 @@
...
+
+def save_graphs(filename, g_list):
+    """
+    Save DGLGraphs to file
+    :param filename: file to store DGLGraphs
+    :param g_list: list of DGLGraph
+    :return:
+    """
+    gdata_list = [GraphData.create(g) for g in g_list]
+    _CAPI_DGLSaveGraphs(filename, gdata_list)
```
jermainewang: One question. The Object FFI is not very efficient for large list. Do you expect the given list to be large? We might need some benchmarks.

VoVAllen: Agree benchmark is needed.

VoVAllen: I tried saving 30000 graphs, with 30 nodes and ~90 edges, and feature size of 5 for edata and ndata. It just took about 3 seconds, which is totally acceptable.

VoVAllen:
@jermainewang Another problem is storing graph-level labels on DGLGraphs. It seems `g.gdata` could be a good solution for this as mentioned [here](https://github.com/dmlc/dgl/issues/714). Do you this this is a good solution? Shall we implement that before this PR?

VoVAllen:
dmlc-core has own windows compilation settings.
https://github.com/dmlc/dmlc-core/tree/master/windows

BarclayII:
> dmlc-core has own windows compilation settings.
> https://github.com/dmlc/dmlc-core/tree/master/windows

It seems that `dmlc-core` with CMake on Windows always builds with static CRT (i.e. built with `/MT` flag) while DGL requires a dynamic CRT (i.e. built with `/MD` flag).

I could successfully build the library by manually changing `third_party/dmlc_core/CMakeLists.txt` to switch the `/MD` and `/MT` flag:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index ccf0356..167f53c 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,9 +63,9 @@ if(MSVC)
   foreach(flag_var
         CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
         CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
-    if(${flag_var} MATCHES "/MD")
-      string(REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
-    endif(${flag_var} MATCHES "/MD")
+    if(${flag_var} MATCHES "/MT")
+      string(REGEX REPLACE "/MT" "/MD" ${flag_var} "${${flag_var}}")
+    endif(${flag_var} MATCHES "/MT")
   endforeach(flag_var)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc")
 else(MSVC)
```

We should probably make a PR to `dmlc-core` to include a "build-with-`/MD`" CMake flag.

VoVAllen:
Thanks a lot. I'm not familiar with this part and  am afraid could not help a lot here. Why dmlc-core is added here is `SeekStream::CreateForRead` is not implemented in the header file but at [here](https://github.com/dmlc/dmlc-core/blob/692216e8633eacae001a023ebbdcafbad0985854/src/io.cc#L140). 
I'm not sure whether the current way is the correct one to add this.

VoVAllen:
I found tvm has similar setting as @BarclayII mentioned ([link](https://github.com/dmlc/tvm/blob/master/CMakeLists.txt#L82)).

BarclayII:
> I found tvm has similar setting as @BarclayII mentioned ([link](https://github.com/dmlc/tvm/blob/master/CMakeLists.txt#L82)).

Precisely what I was referring to.  I made a PR in https://github.com/dmlc/dmlc-core/pull/547.
After that we can update the Windows build script.  If you want to test the CI you can temporarily disable Windows test and add it back later.

jermainewang:
> Another problem is storing graph-level labels on DGLGraphs. It seems `g.gdata` could be a good solution for this as mentioned [here](https://github.com/dmlc/dgl/issues/714). Do you this this is a good solution? Shall we implement that before this PR?

Let's do that in another PR. However, if `g.gdata` is available, do you still wish to store graphs in the format like a list of graph? By contrast, storing graphs in the format of BatchedGraph will be more efficient.

python/dgl/graph_serialize.py
jermainewang: Ok, let's push this PR through first and see if it needs to be improved when we have more realistic benchmarks.

VoVAllen:
I still prefer storing list of graphs. My current thought is storing label at metadata part. The scenario here would be that user may want stratified split of dataset, so they might want label data without loading all the graphs.

VoVAllen:
Ready for review. 
Current version doesn't store graph-level label/data. Need discussion on how to store them.

futurely:
Euler is the only gnn framework that support heterogeneous and distributed graph as far as I know. https://github.com/dmlc/dgl/pull/738#issuecomment-519826751

Its implementation could be a source of inspiration.

https://github.com/alibaba/euler/tree/master/euler/core

yzh119:
@futurely thanks for your suggestion, however your comment is not related to this PR.

VoVAllen:
@jermainewang I changed API a little bit.
Now `load_graphs(filename)` will return list of graphs and labels from files.
And `load_labels(filename)` will only return the labels.
`load_metadata` would be confused for user of what's metadata, therefore I didn't added it.

Everything is ready for review.

mufeili:
@yzh119 Any update on the review?

src/graph/graph_serialize.cc
```diff
@@ -0,0 +1,346 @@
...
+}
+
+ImmutableGraphPtr ToImmutableGraph(GraphPtr g) {
+  ImmutableGraphPtr imgr = std::dynamic_pointer_cast<ImmutableGraph>(g);
+  if (imgr) {
+    return imgr;
+  } else {
+    MutableGraphPtr mgr = std::dynamic_pointer_cast<Graph>(g);
+    CHECK(mgr) << "Invalid Graph Pointer";
+    IdArray srcs_array = mgr->Edges("srcdst").src;
```
yzh119: May lost edge order information.

src/graph/graph_serialize.cc
```diff
@@ -0,0 +1,346 @@
...
+    IdArray dsts_array = mgr->Edges("srcdst").dst;
+    ImmutableGraphPtr imgptr = ImmutableGraph::CreateFromCOO(mgr->NumVertices(), srcs_array,
+                                                             dsts_array);
+    return imgptr;
+  }
+}
+
+void StorageMetaDataObject::setMetaData(dgl_id_t num_graph,
```
yzh119: Uppercase `set`

src/graph/graph_serialize.cc
```diff
@@ -0,0 +1,346 @@
...
+                                        std::vector<NamedTensor> labels_list) {
+  this->num_graph = num_graph;
+  this->nodes_num_list = Value(MakeValue(aten::VecToIdArray(nodes_num_list)));
+  this->edges_num_list = Value(MakeValue(aten::VecToIdArray(edges_num_list)));
+  for (auto kv : labels_list) {
+    this->labels_list.Set(kv.first, Value(MakeValue(kv.second)));
+  }
+}
+
+void StorageMetaDataObject::setGraphData(std::vector<GraphData> gdata) {
```
yzh119: ditto.

yzh119:
Aslo, I was wondering how do you serialize a batched graph.

src/graph/graph_serialize.cc
VoVAllen: Fixed to `eid`

VoVAllen:
BatchedGraph will not be supported and will raise error.

PR #729, opened 2019-07-26
[Refactor] C random number generator

BarclayII:
## Description
This PR replaces `rand_r` with C++ random number generators, dedicated for OpenMP multithreading.  This should provide reproducibility in sampling as well.

Each OpenMP thread would get its own random number generator associated with the thread ID.

New C++ interfaces:
* ~~`Random::SetSeed(seed)`~~ `RandomEngine::ThreadLocal()->SetSeed(seed)`: set the seed of C random number generator.
* ~~`Random::RandInt(a, b)`~~ `RandomEngine::ThreadLocal()->RandInt(a, b)`: generate a random integer in `[a, b)` uniformly.
* ~~`Random::Uniform(a, b)`~~ `RandomEngine::ThreadLocal()->Uniform(a, b)`: generate a random float in `[a, b)` uniformly.

New Python interface:
* ~~`dgl.contrib.sampling.seed(val)`~~ `dgl.random.seed(val)`: set the seed of C random number generator from Python.

EDIT: moved to the thread-local style of TVM, hence changing the C++ interface.

## Checklist
- [x] Code is well-documented


tests/compute/test_sampler.py
```diff
@@ -155,10 +155,29 @@ def test_layer_sampler():
     _test_layer_sampler()
     _test_layer_sampler(prefetch=True)
 
+def test_setseed():
```
BarclayII: @zheng-da @jermainewang @mufeili (since you're involved in a discussion thread)
I require that the NodeFlow sequence emitted by the sampler should be the same given the same seed, even with multiple workers.  I wonder if the requirement is appropriate or too strict, especially since we are specifying multiple workers.

jermainewang:
> dgl.contrib.sampling.seed(val): set the seed of C random number generator from Python.

How about `dgl.random.seed(val)`? The random seed might be used by other components as well.

include/dgl/runtime/random.h
```diff
@@ -0,0 +1,87 @@
...
+  /*! \brief Resize the RNG array */
+  void Resize(int size);
+
+  /*! \brief Meta RNG that seeds the thread-specific RNGs */
+  std::mt19937 meta_rng_;
+  /*! \brief Thread-specific RNGs that have omp_get_num_threads() elements */
+  std::vector<std::mt19937_64> rngs_;
+};
+
+};  // namespace runtime
```
jermainewang: Put it in `dgl` namespace. It shouldn't belong to `runtime`.

include/dgl/runtime/random.h
```diff
@@ -0,0 +1,87 @@
...
+#include <chrono>
+
+namespace dgl {
+namespace runtime {
+
+/*!
+ * \brief OpenMP thread-safe Random Number Generator.
+ *
+ * This is only intended for OpenMP!  It does so by creating omp_get_num_threads()
+ * random number generators.
```
jermainewang: What do you mean this is only intended for OpenMP? I must use it with OpenMP?

tests/compute/test_sampler.py
jermainewang: I think it's too strict. Basically, for multi-thread, use separate thread-local rng for each thread. It makes sure that each thread generates the same sequence but does not care about the concurrency.

jermainewang:
I think concurrency and rng are two things. If a user want to get a deterministic result, s/he should set the number of workers to be one **and** fix the seed. Otherwise, the behavior may or may not be non-deterministic (undefined), making implementation easier. I don't like the idea of coupling RNG with omp, they are two things. My suggestion:
* Use a global singleton seed that could be set/get by the end user. We could assume that only one thread is setting the seed so we don't need to protect it.
* Use thread-local rng for each thread. It will use the global seed if available, otherwise, use local time seed.
* Do nothing about concurrency.

BarclayII:
This PR has adopted the solution from TVM, and should be ready for review.

python/dgl/random.py
```diff
@@ -0,0 +1,16 @@
+"""Pyhton interfaces to DGL random number generators."""
+from ._ffi.function import _init_api
+
+def seed(val):
```
jermainewang: Add it to API doc?

src/random.cc
```diff
@@ -0,0 +1,96 @@
...
+template int32_t RandomEngine::RandInt<int32_t>(int32_t, int32_t);
+template uint32_t RandomEngine::RandInt<uint32_t>(uint32_t, uint32_t);
+template int64_t RandomEngine::RandInt<int64_t>(int64_t, int64_t);
+template uint64_t RandomEngine::RandInt<uint64_t>(uint64_t, uint64_t);
+
+template float RandomEngine::Uniform<float>();
+template double RandomEngine::Uniform<double>();
+
+template float RandomEngine::Uniform<float>(float, float);
+template double RandomEngine::Uniform<double>(double, double);
```
jermainewang: Since the functions are pretty light and are called very frequently (mostly inside a loop), it might worth just inline the implementation to the header file. I will leave it to you.

PR #730, opened 2019-07-29
[Misc] Provide a "return 1s instead of edge IDs" option in scipy adjacency matrix

BarclayII:
## Description
Added a return_edge_id optional argument in DGLGraph.adjacency_matrix_scipy() method which defaults to True. If False, then 1's are returned.

The reason to do this is that if edge IDs are returned, then one of the nonzero entries will be actually 0, which may be confusing.

In the next major release (0.5) it will default to False to keep consistent with `adjacency_matrix` interface.  Since existing code may depend on the current behavior that returns edge IDs, I added a `FutureWarning`.

This PR is separated from #720 

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/graph_index.py
+++ b/python/dgl/graph_index.py
@@ -577,7 +577,7 @@ class GraphIndex(ObjectBase):
         return SubgraphIndex(gidx, self, induced_nodes, e)
 
     @utils.cached_member(cache='_cache', prefix='scipy_adj')
-    def adjacency_matrix_scipy(self, transpose, fmt):
+    def adjacency_matrix_scipy(self, transpose, fmt, return_edge_ids=None):
         """Return the scipy adjacency matrix representation of this graph.
 
         By default, a row of returned adjacency matrix represents the destination
@@ -586,14 +586,14 @@ class GraphIndex(ObjectBase):
         When transpose is True, a row represents the source and a column represents
         a destination.
 
-        The elements in the adajency matrix are edge ids.
-
         Parameters
         ----------
         transpose : bool
             A flag to transpose the returned adjacency matrix.
         fmt : str
             Indicates the format of returned adjacency matrix.
+        return_edge_ids : bool
+            Indicates whether to return edge IDs or 1 as elements.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


python/dgl/graph_index.py
```diff
@@ -603,6 +603,15 @@ def adjacency_matrix_scipy(self, transpose, fmt):
         if not isinstance(transpose, bool):
             raise DGLError('Expect bool value for "transpose" arg,'
                            ' but got %s.' % (type(transpose)))
+
+        if return_edge_ids is None:
+            import warnings
+            warnings.warn(
```
yzh119: How about switching to `dgl.warning`?

python/dgl/graph_index.py
```diff
@@ -603,19 +603,29 @@ def adjacency_matrix_scipy(self, transpose, fmt):
...
+                " and 0 will be eliminated otherwise.",
+                FutureWarning)
+            return_edge_ids = True
+
         rst = _CAPI_DGLGraphGetAdj(self, transpose, fmt)
         if fmt == "csr":
             indptr = utils.toindex(rst(0)).tonumpy()
             indices = utils.toindex(rst(1)).tonumpy()
-            shuffle = utils.toindex(rst(2)).tonumpy()
+            shuffle = utils.toindex(rst(2)).tonumpy() if return_edge_ids else np.ones_like(indices)
```
yzh119: I recommend to rename this variable as data.

python/dgl/graph_index.py
```diff
@@ -603,19 +603,29 @@ def adjacency_matrix_scipy(self, transpose, fmt):
...
             n = self.number_of_nodes()
             return scipy.sparse.csr_matrix((shuffle, indices, indptr), shape=(n, n))
         elif fmt == 'coo':
             idx = utils.toindex(rst(0)).tonumpy()
             n = self.number_of_nodes()
             m = self.number_of_edges()
             row, col = np.reshape(idx, (2, m))
-            shuffle = np.arange(0, m)
+            shuffle = np.arange(0, m) if return_edge_ids else np.ones_like(row)
```
yzh119: The same.

Issue #742, opened 2019-08-06
Adding classes in RGCN example to the main body of library


sharifza:
## 🚀 Feature

I suggest to add the classes in RGCN example to the main body of the library. 
## Motivation

I'm using RGCN in another library and I think it would be much more convenient for everyone like me to directly import e.g. the layers from the library rather than adding them from the examples.

jermainewang:
I think that's a good idea. Will do.

jermainewang:
@sharifza created a PR on this. Currently implemented the pytorch module (basically copy-paste the codes from the example). Please take a look. Will merge after the documentation and mxnet modules are finished.

sharifza:
Looks good, thanks. Should we include the base class that uses the layers, also somewhere in the library?
https://github.com/dmlc/dgl/blob/565f0c88fc5cfbcfd5476da8146e92c793af4834/examples/pytorch/rgcn/model.py#L3

jermainewang:
@sharifza Sorry didn't notice your comment here. The module has been merged and it now has only one class `RelGraphConv`. The regularization ("basis" or "bdd") is an option. If you have other questions, please reopen this issue. Thanks!

PR #744, opened 2019-08-06
[NN] RGCN modules

jermainewang:
## Description

Requested by #742 . Added `RelGraphConv` module to NN.

TBD:
- [x] Update MXNet RGCN example

## Checklist
- [x] Code is well-documented


python/dgl/nn/pytorch/conv.py
```diff
@@ -148,3 +148,219 @@ def extra_repr(self):
...
+            node_repr = node_repr + self.h_bias
+        if self.self_loop:
+            node_repr = node_repr + loop_message
+        if self.activation:
+            node_repr = self.activation(node_repr)
+
+        return node_repr
+
+class RelGraphConvBasis(_BaseRelGraphConv):
+    """Relational Graph Convolution with basis regularization.
```
sharifza: """Apply relational graph convolution with basis regularization graph convolution over an input signal. Relational graph convolution is introduced in `RGCN <https://arxiv.org/abs/1703.06103>`__
    and can be described as below:

jermainewang:
@ylfdq1118 @yzh119 Please review.

Correctness and speed:
The new module is able to achieve the reported numbers in the README for aifb, mutag and BGS. However, I observed quite large variance in these results. For link prediction task, I could achieve 0.144 MRR after 500 iterations.

The code maintains the same training speed.

jermainewang:
@ylfdq1118 I don't know whether the normalizer is correctly implemented or not. From the paper, it should be c_{r,i}, which means each node should have r number of normalizers. However, the code only has one.

ylfdq1118:
> However, the code only has one.

Do you mean the link prediction code?

Please check out section 5.2 of the [paper](https://arxiv.org/pdf/1703.06103.pdf):
> We found a normalization constant defined as $c_{i,r} = c_i =\sum_r|N_i^r|$ - in other words, applied across relation types – to work best

examples/pytorch/rgcn/entity_classify.py
```diff
@@ -151,6 +156,8 @@ def main(args):
             help="l2 norm coef")
     parser.add_argument("--relabel", default=False, action='store_true',
             help="remove untouched nodes and relabel")
+    parser.add_argument("--use-self-loop", default=False, action='store_true',
+            help="include self feature as a special relation")
```
ylfdq1118: Actually, I don't think self loop is a special relation. It works more like skip connection especially because self loop is not done in message passing and is not normalized.

Even if it is a special relation, the preprocessing code of entity classification already adds a self-relation to the graph. I agree this argument makes the usage layer more consistent. But if user sets this argument to be true, then self loop will then be added twice. So maybe we should remove the code in data preprocessing that adds the self relation.

jermainewang: It's more like a special relation because of the W_0 weight matrix. By contrast, GraphSAGE models it as $W * (mean({h_j}) + h_i)$ which is more like a skip connection. Does the author's code add self loops?

jermainewang:
@ylfdq1118 Found the accuracy numbers dropped a little bit (still better than paper's numbers) for MXNet. Cannot find the reason at the moment.

examples/pytorch/rgcn/entity_classify.py
ylfdq1118: Skip connection usually involves a transformation matrix $W$ to make sure dimension matches. 

Author's code adds an identity matrix here:
https://github.com/tkipf/relational-gcn/blob/master/rgcn/prepare_dataset.py#L39

Issue #734, opened 2019-07-31
EdgeBatch.edges() call attempts tuple item assignment


lukovnikov:
## 🐛 Bug

TypeError: 'tuple' object does not support item assignment

## To Reproduce

```python
import dgl
import torch
g = dgl.DGLGraph()
g.add_nodes(11)
h = torch.rand(11, 7)
g.nodes[:].data["h"] = h
g.add_edges(list(range(10)), 10, {"etype": torch.ones(10)})
g.add_edges(list(range(10)), list(range(10)), {"etype": torch.zeros(10)})
qnodes = torch.tensor([10])
def edgefilter(eb: dgl.EdgeBatch):
    ret = (eb.edges()[1].unsqueeze(0) == qnodes.unsqueeze(1)).any(0)
    ret &= eb.data["etype"] == 1
    return ret

g.filter_edges(edgefilter)
```

This throws the above mentioned exception, from the call to EdgeView.edges()

## Expected behavior

Not to throw errors.

## Environment

 - DGL Version (e.g., 1.0): **0.3**
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): **PyTorch 1.1.0**
 - OS (e.g., Linux): **Windows**
 - How you installed DGL (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version: **3.6**
 - CUDA/cuDNN version (if applicable): N/A
 - GPU models and configuration (e.g. V100): None
 - Any other relevant information:

yzh119:
@lukovnikov solved in #747 .

PR #747, opened 2019-08-07
[Bugfix] EdgeBatch.edges call attempts tuple item assignment

yzh119:
## Description
Related issue #734 .

## Checklist
- [x] Code is well-documented

## Changes
- python/dgl/udf.py
- python/dgl/graph.py
- tests/compute/test_udf.py

### DIFF
--- a/python/dgl/graph.py
+++ b/python/dgl/graph.py
@@ -3258,7 +3258,7 @@ class DGLGraph(DGLBaseGraph):
         """
         if is_all(edges):
             eid = ALL
-            u, v, _ = self._graph.edges()
+            u, v, _ = self._graph.edges('eid')
         elif isinstance(edges, tuple):
             u, v = edges
             u = utils.toindex(u)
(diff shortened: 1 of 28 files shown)


python/dgl/graph.py
```diff
@@ -3258,7 +3258,7 @@ def filter_edges(self, predicate, edges=ALL):
         """
         if is_all(edges):
             eid = ALL
-            u, v, _ = self._graph.edges()
```
jermainewang: Need it to be 'eid' order? By default, the order is arbitrary.

yzh119: I think it has to be eid order, if you check the logic in dgl/udf.py you will find we convert `ALL` placeholder to `F.arange(0, self._g.number_of_edges())`.
On C++ end, by default `edges()` would use `eid` order if order is not specified. But I think we should specify it here if we later changed the default behavior of `edges()`.

jermainewang:
Just to leave a note about the edge order. Since the graph is a set, there
is no intrinsic order. On the C++ side, it's better to keep the flexibility
that if the order argument is not given, then arbitrary order would be
returned. This is because the graph could be either csr or copying, which
is suitable for different types of orders. On the Python side, we should
choose which one to use smartly. For this PR, choosing Eid order seems to
be the right one.

PR #749, opened 2019-08-09
[Refactor] Decoupling ImmutableGraph from Kernels

BarclayII:
## Description
This PR attempts to decouple `ImmutableGraph` from Kernels as much as possible.  This is necessary for running fused message passing on heterogeneous graphs.

Actual kernel routines in C take in `CSRWrapper` references, which can be constructed from either `ImmutableGraph` or `Bipartite` in the future.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/src/kernel/csr_interface.h
+++ b/src/kernel/csr_interface.h
@@ -8,6 +8,7 @@
 #define DGL_KERNEL_CSR_INTERFACE_H_
 
 #include <dgl/array.h>
+#include <dgl/runtime/c_runtime_api.h>
 
 namespace dgl {
 namespace kernel {
@@ -16,6 +17,8 @@ class CSRWrapper {
  public:
   virtual aten::CSRMatrix GetInCSRMatrix() const = 0;
   virtual aten::CSRMatrix GetOutCSRMatrix() const = 0;
+  virtual DGLContext Context() const = 0;
+  virtual int NumBits() const = 0;
 };
 
 };  // namespace kernel
--- a/src/kernel/csr_interface.h
+++ b/src/kernel/csr_interface.h
@@ -1,0 +1,24 @@
+/*!
+ *  Copyright (c) 2019 by Contributors
+ * \file kernel/csr_interface.h
+ * \brief Kernel common utilities
+ */
+
+#ifndef DGL_KERNEL_CSR_INTERFACE_H_
+#define DGL_KERNEL_CSR_INTERFACE_H_
+
+#include <dgl/array.h>
+
+namespace dgl {
+namespace kernel {
+
+class CSRWrapper {
+ public:
+  virtual aten::CSRMatrix GetInCSRMatrix() const = 0;
+  virtual aten::CSRMatrix GetOutCSRMatrix() const = 0;
+};
+
+};  // namespace kernel
+};  // namespace dgl
+
+#endif  // DGL_KERNEL_CSR_INTERFACE_H_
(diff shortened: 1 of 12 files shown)


src/kernel/csr_interface.h
```diff
@@ -0,0 +1,27 @@
...
+#ifndef DGL_KERNEL_CSR_INTERFACE_H_
+#define DGL_KERNEL_CSR_INTERFACE_H_
+
+#include <dgl/array.h>
+#include <dgl/runtime/c_runtime_api.h>
+
+namespace dgl {
+namespace kernel {
+
+class CSRWrapper {
```
jermainewang: Some brief docstring is appreciated.

ylfdq1118:
Looks good! @BarclayII 

I just have a few questions (maybe because I don't have enough context):
1. You added a new class called `CSRWrapper`. Is that assuming DGL kernel will only use CSR format? Currently, `ImmutableGraph` itself does not implying any storage format.
2. From https://github.com/dmlc/dgl/blob/4a0a7eba1d33aac52518a913b3dc90a0c088b9f6/src/kernel/binary_reduce.cc#L228
The `CSRWrapper` here is not really holding a CSR. Instead, it's holding a pointer to `ImmutableGraph` object, which is more general than a CSR. So the implementation here is a little bit not that straight-forward. I guess the reason here is because DGL kernels actually need two CSRs.
3. This wrapper is added at the level of kernel's CAPI
https://github.com/dmlc/dgl/blob/4a0a7eba1d33aac52518a913b3dc90a0c088b9f6/src/kernel/binary_reduce.cc#L385-L387
So it's assuming that the python code is still going to pass down an `ImmutableGraph` object. Then my question is how you are going to support HeteroGraph. Are you going to add another set of CAPIs to support HeteroGraph? If that's the case, python side also needs to duplicate code to invoke kernels for HeteroGraph?

BarclayII:
> 1. You added a new class called `CSRWrapper`. Is that assuming DGL kernel will only use CSR format? Currently, `ImmutableGraph` itself does not implying any storage format.

The object is merely to unify `ImmutableGraph` and `Bipartite` objects, as they don't have the same base class.  `ImmutableGraph` is wrapped with an `ImmutableGraphCSRWrapper`, and `Bipartite` would be wrapped with `BipartiteCSRWrapper`.  Of course, such wrapper can have another interface called `GetCOOMatrix`.

I discussed with @jermainewang and agreed that the ideal case would be refactor the code to unify `ImmutableGraph` and `Bipartite` to one single class, since they share most of the code.

> 2. From https://github.com/dmlc/dgl/blob/4a0a7eba1d33aac52518a913b3dc90a0c088b9f6/src/kernel/binary_reduce.cc#L228
>    
>    The `CSRWrapper` here is not really holding a CSR. Instead, it's holding a pointer to `ImmutableGraph` object, which is more general than a CSR. So the implementation here is a little bit not that straight-forward. I guess the reason here is because DGL kernels actually need two CSRs.

Yes.

> 3. This wrapper is added at the level of kernel's CAPI
>    https://github.com/dmlc/dgl/blob/4a0a7eba1d33aac52518a913b3dc90a0c088b9f6/src/kernel/binary_reduce.cc#L385-L387
>    
>    So it's assuming that the python code is still going to pass down an `ImmutableGraph` object. Then my question is how you are going to support HeteroGraph. Are you going to add another set of CAPIs to support HeteroGraph? If that's the case, python side also needs to duplicate code to invoke kernels for HeteroGraph?

You don't have to write two separate CAPIs.  The idea is to check the type of the passed argument during runtime:

https://github.com/dmlc/dgl/blob/ad60f4a70289f6165fcc067e931ee4b7f3f13264/src/kernel/binary_reduce.cc#L322-L342

PR #751, opened 2019-08-10
[Fix] Synchronize frame initializers for local_var and local_scope & Remove warning

mufeili:
## Description
Currently `local_var` and `local_scope` do not use the frame initializers of the original `_node_frame` and `_edge_frame`. This causes warnings about initializers even when we have initializers set, which can be confusing for example when using `nn modules`.

Also as discussed before, warnings about frame initializers are removed.

## Checklist
- [x] Code is well-documented

## Changes
- Synchronize frame initializers after the copy of frames
- Test cases are added

### DIFF
--- a/python/dgl/graph.py
+++ b/python/dgl/graph.py
@@ -9,7 +9,7 @@ import dgl
 from .base import ALL, is_all, DGLError
 from . import backend as F
 from . import init
-from .frame import FrameRef, Frame, Scheme
+from .frame import FrameRef, Frame, Scheme, sync_frame_initializer
 from . import graph_index
 from .runtime import ir, scheduler, Runtime
 from . import utils
@@ -3460,19 +3460,3 @@ class DGLGraph(DGLBaseGraph):
         yield
         self._node_frame = old_nframe
         self._edge_frame = old_eframe
-
-def sync_frame_initializer(new_frame, reference_frame):
-    """Set the initializers of the new_frame to be the same as the reference_frame.
-
-    Parameters
-    ----------
-    new_frame : Frame
-        The frame to set initializers
-    reference_frame : Frame
-        The frame to copy initializers
-    """
-    new_frame._default_initializer = reference_frame._default_initializer
-    # set per-col initializer
-    # TODO(minjie): hack; cannot rely on keys as the _initializers
-    #   now supports non-exist columns.
-    new_frame._initializers = reference_frame._initializers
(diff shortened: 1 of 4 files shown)


python/dgl/graph.py
```diff
@@ -3451,6 +3455,24 @@ def local_scope(self):
         old_eframe = self._edge_frame
         self._node_frame = FrameRef(Frame(self._node_frame._frame))
         self._edge_frame = FrameRef(Frame(self._edge_frame._frame))
+        sync_frame_initializer(self._node_frame._frame, old_nframe._frame)
+        sync_frame_initializer(self._edge_frame._frame, old_eframe._frame)
         yield
         self._node_frame = old_nframe
         self._edge_frame = old_eframe
+
+def sync_frame_initializer(new_frame, reference_frame):
```
yzh119: I recommend moving this function to frame.py

mufeili: Done.

python/dgl/graph.py
```diff
@@ -3401,9 +3401,13 @@ def local_var(self):
         DGLGraph
             The graph object that can be used as a local variable.
         """
+        local_node_frame = FrameRef(Frame(self._node_frame._frame))
```
yzh119: Better explain the behavior of initializer in the docstring.

python/dgl/graph.py
```diff
@@ -3451,6 +3455,8 @@ def local_scope(self):
         old_eframe = self._edge_frame
         self._node_frame = FrameRef(Frame(self._node_frame._frame))
         self._edge_frame = FrameRef(Frame(self._edge_frame._frame))
+        sync_frame_initializer(self._node_frame._frame, old_nframe._frame)
```
yzh119: The same.

jermainewang:
Hey @mufeili, if we just get rid of the warning, could it be easier? I suppose we only need to change `warn_and_set_initializer`?

python/dgl/graph.py
mufeili: Better?

python/dgl/graph.py
mufeili: Better?

mufeili:
@jermainewang I've removed the warnings. However, I think this synchronization is still necessary as long as we do not completely deprecate initializers.

python/dgl/graph.py
yzh119: I mean users should know the `local_var` use the same node/edge initializer as the original `DGLGraph`, maybe you could mention it in the Notes section.

mufeili: Done.

mufeili:
@yzh119 Any further concerns?

yzh119:
Please go ahead and merge it.

PR #760, opened 2019-08-13
[Dataset] Tox21

VoVAllen:
## Description
@mufeili If you are fine with the code, I'll add the documentation soon.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/data/chem/csv_dataset.py
+++ b/python/dgl/data/chem/csv_dataset.py
@@ -11,18 +11,51 @@ from .utils import smile2graph
 from ..utils import download, get_download_dir, _get_dgl_url, Subset
 
 class CSVDataset(object):
+    """CSVDataset
+
+    This is a general class for loading data from csv or pd.DataFrame.
+
+    In data pre-processing, we set non-existing labels to be 0, and returning mask with 1 where label exists.
+
+    All molecules are converted into DGLGraphs. After the first-time construction, the
+    DGLGraphs will be saved for reloading so that we do not need to reconstruct them every time.
+
+    Parameters
+    ----------
+    df: pandas.DataFrame
+    Dataframe including smiles and labels. Can be loaded by pandas.read_csv(file_path).
+    One column includes smiles and other columns for binary labels.
+    Column names other than smiles column would be considered as task names.
+
+    smile2graph: callable, str -> DGLGraph
+    A function turns smiles into a DGLGraph. Default one can be found 
+    at python/dgl/data/chem/utils.py named with smile2graph.
+
+    smile_column: str
+    Column name that including smiles
+
+    cache_file_path: str
+    Path to store the preprocessed data
+    """
     
-    def __init__(self, df, smile2graph=smile2graph, smile_name='smiles', cache_file_path="csvdata_dglgraph.pkl"):
+    def __init__(self, df, smile2graph=smile2graph, smile_column='smiles', cache_file_path="csvdata_dglgraph.pkl"):
         if 'rdkit' not in sys.modules:
             from ...base import dgl_warning
             dgl_warning("Please install RDKit (Recommended Version is 2018.09.3)")
         self.df = df
-        self.smiles = self.df[smile_name].tolist()
-        self.task_names = self.df.columns.drop([smile_name]).tolist()
+        self.smiles = self.df[smile_column].tolist()
+        self.task_names = self.df.columns.drop([smile_column]).tolist()
         self.cache_file_path = cache_file_path
         self._pre_process(smile2graph)
 
     def _pre_process(self, smile2graph):
+        """Pre-process the dataset
+
+        * Convert molecules from smiles format into DGLGraphs
+          and featurize their atoms
+        * Set missing labels to be 0 and use a binary masking
+          matrix to mask them
+        """
         if os.path.exists(self.cache_file_path):
             # DGLGraphs have been constructed before, reload them
             print('Loading previously saved dgl graphs...')
@@ -37,11 +70,32 @@ class CSVDataset(object):
                 pickle.dump(self.graphs, f)
 
         _label_values = self.df[self.task_names].values
-        self.labels = np.nan_to_num(_label_values)
-        self.mask = ~np.isnan(_label_values)
+        # np.nan_to_num will also turn inf into a very large number
+        self.labels =  F.zerocopy_from_numpy(np.nan_to_num(_label_values))
+        self.mask =  F.zerocopy_from_numpy(~np.isnan(_label_values).astype(np.float32))
 
     def __getitem__(self, item):
+        """Get the ith datapoint
+
+        Returns
+        -------
+        str
+            SMILES for the ith datapoint
+        DGLGraph
+            DGLGraph for the ith datapoint
+        Tensor of dtype float32
+            Labels of the datapoint for all tasks
+        Tensor of dtype float32
+            Weights of the datapoint for all tasks
+        """
         return self.smiles[item], self.graphs[item], self.labels[item], self.mask[item]
 
     def __len__(self):
+        """Length of Dataset
+        
+        Return
+        ------
+        int
+            Length of Dataset
+        """
         return len(self.smiles)
--- a/python/dgl/data/chem/csv_dataset.py
+++ b/python/dgl/data/chem/csv_dataset.py
@@ -4,6 +4,7 @@ import dgl.backend as F
 import numpy as np
 import os
 import pickle
+import sys
 
 from dgl import DGLGraph
 from .utils import smile2graph
@@ -12,6 +13,9 @@ from ..utils import download, get_download_dir, _get_dgl_url, Subset
 class CSVDataset(object):
     
     def __init__(self, df, smile2graph=smile2graph, smile_name='smiles', cache_file_path="csvdata_dglgraph.pkl"):
+        if 'rdkit' not in sys.modules:
+            from ...base import dgl_warning
+            dgl_warning("Please install RDKit (Recommended Version is 2018.09.3)")
         self.df = df
         self.smiles = self.df[smile_name].tolist()
         self.task_names = self.df.columns.drop([smile_name]).tolist()
--- a/python/dgl/data/chem/csv_dataset.py
+++ b/python/dgl/data/chem/csv_dataset.py
@@ -24,7 +24,7 @@ class CSVDataset(object):
     ----------
     df: pandas.DataFrame
     Dataframe including smiles and labels. Can be loaded by pandas.read_csv(file_path).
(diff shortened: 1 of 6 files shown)


python/dgl/data/chem/utils.py
```diff
@@ -0,0 +1,175 @@
...
+import os
+import pickle
+
+from dgl import DGLGraph
+
+try:
+    from rdkit import Chem
+    from rdkit.Chem import rdmolfiles, rdmolops
+except ImportError:
+    raise ImportError("Please install rdkit")
```
mufeili: Add rdkit version.

python/dgl/data/chem/utils.py
```diff
@@ -0,0 +1,175 @@
...
+    ----------
+    x : str, int or Chem.rdchem.HybridizationType
+    allowable_set : list
+        The elements of the allowable_set should be of the
+        same type as x.
+
+    Returns
+    -------
+    list
+        List of boolean values where only one value is True.
```
mufeili: It's possible that the whole vector is 0. I'd replace "only one" by "at most one".

python/dgl/data/chem/utils.py
```diff
@@ -0,0 +1,175 @@
...
+        x == allowable_set[i].
+    """
+    return list(map(lambda s: x == s, allowable_set))
+
+
+class BaseAtomFeaturizer(object):
+    """An abstract class for atom featurizers
+
+    All atom featurizers that map a molecule to atom features should subclass it.
+    All subclasses should overrite ``_featurize_atom``, which featurizes a single
```
mufeili: "overrite" should be "overwrite".

python/dgl/data/chem/utils.py
```diff
@@ -0,0 +1,175 @@
...
+    atom and ``__call__``, which featurize all atoms in a molecule.
```
mufeili: "featurize" should be "featurizes".

python/dgl/data/chem/utils.py
```diff
@@ -0,0 +1,175 @@
...
+            atom = mol.GetAtomWithIdx(i)
+            atom_features.append(self._featurize_atom(atom))
+        atom_features = np.stack(atom_features)
+        atom_features = F.zerocopy_from_numpy(atom_features.astype(np.float32))
+
+        return {self.atom_data_field: atom_features}
+
+
+def smile2graph(smile, add_self_loop=False, atom_featurizer=None, bond_featurizer=None):
+    # Graph construction
```
mufeili: Add documentation here, particularly the order of adding edges.

python/dgl/data/chem/csv_dataset.py
```diff
@@ -0,0 +1,43 @@
...
+import dgl.backend as F
+import numpy as np
+import os
+import pickle
+
+from dgl import DGLGraph
+from .utils import smile2graph
+from ..utils import download, get_download_dir, _get_dgl_url, Subset
+
+class CSVDataset(object):
```
mufeili: Add documentation for the class and methods.

python/dgl/data/chem/csv_dataset.py
```diff
@@ -0,0 +1,43 @@
...
+        else:
+            self.graphs = []
+            for id, s in enumerate(self.smiles):
+                self.graphs.append(smile2graph(s))
+
+            with open(self.cache_file_path, 'wb') as f:
+                pickle.dump(self.graphs, f)
+
+        _label_values = self.df[self.task_names].values
+        self.labels = np.nan_to_num(_label_values)
```
mufeili: One caveat is that `np.nan_to_num` will also replace infinity with large finite numbers. Add a notice here.

python/dgl/data/chem/tox21.py
```diff
@@ -0,0 +1,31 @@
...
+from .utils import smile2graph
+from ..utils import get_download_dir, download, _get_dgl_url, Subset
+
+try:
+    import pandas as pd
+except ImportError:
+    raise ImportError("Please install pandas")
+
+
+class Tox21(CSVDataset):
```
mufeili: Add documentation.

python/dgl/data/chem/tox21.py
```diff
@@ -0,0 +1,31 @@
...
+    _urls = 'dataset/tox21.csv.gz'
```
mufeili: `_urls` should rather be `_url`?

python/dgl/data/chem/tox21.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def __init__(self, smile2graph=smile2graph):
+        data_path = get_download_dir() + '/tox21.csv.gz'
+        download(_get_dgl_url(self._urls), path=data_path)
```
mufeili: Same

python/dgl/data/chem/tox21.py
```diff
@@ -0,0 +1,31 @@
...
+        df = pd.read_csv(data_path)
+        df = df.drop(columns=['mol_id'])
```
mufeili: I don't think we should drop them.

python/dgl/data/chem/tox21.py
```diff
@@ -0,0 +1,31 @@
...
+    raise ImportError("Please install pandas")
+
+
+class Tox21(CSVDataset):
+    _urls = 'dataset/tox21.csv.gz'
+
+    def __init__(self, smile2graph=smile2graph):
+        data_path = get_download_dir() + '/tox21.csv.gz'
+        download(_get_dgl_url(self._urls), path=data_path)
+        df = pd.read_csv(data_path)
```
mufeili: Add an option for shuffling the dataset.

python/dgl/data/chem/tox21.py
```diff
@@ -0,0 +1,31 @@
...
+        df = df.drop(columns=['mol_id'])
+        super().__init__(df, smile2graph, cache_file_path="tox21_dglgraph.pkl")
+        self._weight_balancing()
+
+    def _weight_balancing(self):
+        num_pos = np.sum(self.labels, axis=0)
```
mufeili: This should rather be considered as an optional processing step.

python/dgl/data/utils.py
```diff
@@ -24,6 +25,22 @@ def _get_dgl_url(file_url):
     return repo_url + file_url
 
 
+def split_dataset(dataset, frac_list=[0.8, 0.1, 0.1], shuffle=False, random_state=42):
```
mufeili: Many people are against using mutable default value as an argument, e.g. [Using a mutable default value as an argument
](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html). I tend to agree them and maybe it's more secure to set None as default argument and do

```python
if frac_list is None:
    frac_list = [0.8, 0.1, 0.1]
```

mufeili: Also maybe you use `42` because this is a magic number, but this can be confusing for people who do not know the story.

mufeili:
Overall well done, some minor comments:

1. Have you tested the implementation by re-training the model? 
2. Try to make more detailed documentation.

python/dgl/data/utils.py
VoVAllen: Changed to None.

VoVAllen: Done

python/dgl/data/chem/tox21.py
VoVAllen: It's just an additional field, seems no side effect if we execute this.

python/dgl/data/chem/tox21.py
VoVAllen: I prefer user using `split_dataset` function in the main file.

python/dgl/data/chem/tox21.py
VoVAllen: Added `self.id=df['mol_id']`. Because some csv dataset may not have id column, therefore I assume the dataframe passed to CSVDataset is already preprcessed with only smiles and labels.

python/dgl/data/chem/tox21.py
VoVAllen: Done

python/dgl/data/chem/tox21.py
mufeili: I understand that, but that part might still be useful for some users.

python/dgl/data/chem/tox21.py
mufeili: Allright, but make sure to add documentation about that.

python/dgl/data/chem/tox21.py
VoVAllen: You can maintain the id in the inherited class. It's still there.

mufeili: I'm fine with additional `self.id`, just not to completely remove them.

python/dgl/data/chem/csv_dataset.py
```diff
@@ -11,18 +11,51 @@
...
+    In data pre-processing, we set non-existing labels to be 0, and returning mask with 1 where label exists.
+
+    All molecules are converted into DGLGraphs. After the first-time construction, the
+    DGLGraphs will be saved for reloading so that we do not need to reconstruct them every time.
+
+    Parameters
+    ----------
+    df: pandas.DataFrame
+    Dataframe including smiles and labels. Can be loaded by pandas.read_csv(file_path).
+    One column includes smiles and other columns for binary labels.
```
mufeili: I suppose this dataset does not necessarily need to be for binary classification?

mufeili: The current implementation does not necessarily need to be for binary classification.

python/dgl/data/chem/csv_dataset.py
```diff
@@ -11,18 +11,51 @@
...
+    Column names other than smiles column would be considered as task names.
+
+    smile2graph: callable, str -> DGLGraph
+    A function turns smiles into a DGLGraph. Default one can be found 
+    at python/dgl/data/chem/utils.py named with smile2graph.
```
mufeili: I think it will be better to update `docs/source/api/python/data.rst` and directly link to `smile2graph` with `:func:`. Make sure you build the doc and take a look at the visual results.

mufeili: Similarly, update `data.rst` for other data related APIs that users should have access to.

python/dgl/data/chem/csv_dataset.py
VoVAllen: Removed

python/dgl/data/chem/csv_dataset.py
VoVAllen: Would add this later. Let's merge this first.

PR #776, opened 2019-08-19
[Dataset] Fix CSVDataset Interface

VoVAllen:
## Description
Change reading dataframe to file path

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/data/chem/csv_dataset.py
+++ b/python/dgl/data/chem/csv_dataset.py
@@ -10,6 +10,12 @@ from dgl import DGLGraph
 from .utils import smile2graph
 
 
+try:
+    import pandas as pd
+    import rdkit
+except ImportError:
+    pass
+
 class CSVDataset(object):
     """CSVDataset
 
@@ -39,13 +45,19 @@ class CSVDataset(object):
     Path to store the preprocessed data
     """
 
-    def __init__(self, df, smile2graph=smile2graph, smile_column='smiles',
+    def __init__(self, csvfile_path, smile2graph=smile2graph, smile_column='smiles', id_column=None,
                  cache_file_path="csvdata_dglgraph.pkl"):
         if 'rdkit' not in sys.modules:
             from ...base import dgl_warning
             dgl_warning(
                 "Please install RDKit (Recommended Version is 2018.09.3)")
-        self.df = df
+        if 'pandas' not in sys.modules:
+            from ...base import dgl_warning
+            dgl_warning("Please install pandas")
+        self.df = pd.read_csv(csvfile_path)
+        if id_column is not None:
+            self.id = self.df[id_column]
+            self.df = self.df.drop(columns=[id_column])
         self.smiles = self.df[smile_column].tolist()
         self.task_names = self.df.columns.drop([smile_column]).tolist()
         self.n_tasks = len(self.task_names)
@@ -90,8 +102,8 @@ class CSVDataset(object):
             Weights of the datapoint for all tasks
         """
         return self.smiles[item], self.graphs[item], \
-               F.zerocopy_from_numpy(self.labels[item]),  \
-               F.zerocopy_from_numpy(self.mask[item])
+            F.zerocopy_from_numpy(self.labels[item]),  \
+            F.zerocopy_from_numpy(self.mask[item])
 
     def __len__(self):
         """Length of Dataset
--- a/python/dgl/data/chem/tox21.py
+++ b/python/dgl/data/chem/tox21.py
@@ -43,12 +43,7 @@ class Tox21(CSVDataset):
 
         data_path = get_download_dir() + '/tox21.csv.gz'
         download(_get_dgl_url(self._url), path=data_path)
-        df = pd.read_csv(data_path)
-        self.id = df['mol_id']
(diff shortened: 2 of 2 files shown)


python/dgl/data/chem/tox21.py
```diff
@@ -43,12 +43,7 @@ def __init__(self, smile2graph=smile2graph):
 
         data_path = get_download_dir() + '/tox21.csv.gz'
         download(_get_dgl_url(self._url), path=data_path)
-        df = pd.read_csv(data_path)
-        self.id = df['mol_id']
-
-        df = df.drop(columns=['mol_id'])
-
-        super().__init__(df, smile2graph, cache_file_path="tox21_dglgraph.pkl")
+        super().__init__(data_path, smile2graph, cache_file_path="tox21_dglgraph.pkl")
```
mufeili: Seems you forgot to pass `id_column` for Tox21.

python/dgl/data/chem/csv_dataset.py
```diff
@@ -10,6 +10,12 @@
 from .utils import smile2graph
 
 
+try:
+    import pandas as pd
+    import rdkit
+except ImportError:
+    pass
+
 class CSVDataset(object):
```
jermainewang: CSVDataset -> MoleculeCSVDataset. In the docstring, would you mind describing a little bit about the schema of the input CSV?

python/dgl/data/chem/csv_dataset.py
```diff
@@ -39,13 +45,19 @@ class CSVDataset(object):
...
+    def __init__(self, csvfile_path, smile2graph=smile2graph, smile_column='smiles', id_column=None,
                  cache_file_path="csvdata_dglgraph.pkl"):
         if 'rdkit' not in sys.modules:
             from ...base import dgl_warning
             dgl_warning(
                 "Please install RDKit (Recommended Version is 2018.09.3)")
-        self.df = df
+        if 'pandas' not in sys.modules:
+            from ...base import dgl_warning
+            dgl_warning("Please install pandas")
```
jermainewang: Is pandas required or not? If required, you could just throw error in L17

python/dgl/data/chem/csv_dataset.py
```diff
@@ -90,8 +102,8 @@ def __getitem__(self, item):
             Weights of the datapoint for all tasks
         """
         return self.smiles[item], self.graphs[item], \
-               F.zerocopy_from_numpy(self.labels[item]),  \
-               F.zerocopy_from_numpy(self.mask[item])
+            F.zerocopy_from_numpy(self.labels[item]),  \
+            F.zerocopy_from_numpy(self.mask[item])
```
jermainewang: Let not use backend API in data package. Also returning dgl ndarray to user is not friendly. Directly return numpy arrays?

python/dgl/data/chem/csv_dataset.py
VoVAllen: Sure.

python/dgl/data/chem/csv_dataset.py
VoVAllen: Currently in dgl/data/__init__.py having `from .chem import Tox21`, which triggers import rdkit and pandas when importing other dataset. I'll change this.

mufeili:
@jermainewang Why do you think it's a bad idea to return framework tensors than numpy arrays?

jermainewang:
@mufeili My bad. I thought `F.zerocopy_from_numpy` returns `dgl.ndarray`. Resolved the comment and  LGTM.

jermainewang:
One side note. I think the question has been around for a while. Should `dgl.data` depends on `dgl.backend`? If you were to write a blog about defining custom dataset, do you suggest the user to learn backend APIs?

mufeili:
@jermainewang I think it's fine to have different standards depending on the scenario:
1. It's fine to let `dgl.data` depend on `dgl.backend` since we want to make `dgl` framework agnostic.
2. If I were to write a blog about dataset customization, I will create an example with one framework since most users use one only. This is also the way we develop examples in documentation.
3. If users want to contribute their code for new datasets and their code is framework agnostic. I'd rather merge their code first and then let our folks handle the rest work for bringing framework agnostic.

VoVAllen:
It's better to set the feature in DGLGraph as backend tensor instead of numpy array. Since if we set ndata as numpy array, user would have to do something like `g.ndata['h']=torch.tensor(g.ndata['h'])`, which is redundant. Therefore we should not block using `dgl.backend` in `dgl.data`.

jermainewang:
Merged the PR. But feel free to discuss more in this thread or we could create another thread for this.

It looks like using backend APIs cannot be avoided. For example, it is very clean to save the features in the graph and this requires the features to be stored in as framework tensors. However, to lower the learning curve, I suggest we process the data in numpy as much as possible and convert it to framework tensors only right before return. As such, I guess only `F.zerocopy_from_numpy` is mostly desired? We could also make it prettier by some syntax tricks.

mufeili:
@VoVAllen This PR is merged into `dmlc:model_zoo` rather than `dmlc:master`, is this intended?

PR #783, opened 2019-08-22
[Model Zoo] DGMG for molecule generation

mufeili:
## Description
1. DGMG for molecule generation, with both training scripts and pre-trained model support.
2. Remove `examples/pytorch/model_zoo/chem/property_prediction/inference_pretrained.ipynb` as dicussed before. @VoVAllen If you want to keep it, you can make a PR [here](https://github.com/dglai/DGL-GTC2019) and add links in a README.

Todo (@mufeili): Final test should be performed before merge.

## Checklist
- [x] Code is well-documented


examples/pytorch/model_zoo/chem/generative_models/dgmg/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Using new datasets
+
+If you want to train with new datasets, you can proceed as follows:
+1. Get the list of atom types and bond types from a list of molecule SMILES. 
+You can use `get_atom_and_bond_types` at `L370, utils.py`.
+2. Since DGMG has some limitation in generating molecules (E.g. it ignores things like protonation and chirality), we
+need to standardize our dataset to avoid false novel molecules. You can preprocess your dataset with 
+`preprocess_dataset` at `L638, utils.py`. 
+3. Modify the dataset class `MoleculeDataset._setup` at `L753, utils.py`.
```
VoVAllen: It's not practical for user to use their own dataset under current interface. At least I still don't have a clear idea on how to customize after 15-min reading. Also it's not a good practice to ask user to rewrite a function which modifies `self`.

python/dgl/model_zoo/chem/pretrain.py
```diff
@@ -14,17 +48,21 @@ def load_pretrained(model_name):
...
         model = GCNClassifier(in_feats=74,
                               gcn_hidden_feats=[64, 64],
                               n_tasks=12,
                               classifier_hidden_feats=64)
-        checkpoint = torch.load(local_pretrained_path)
-        model.load_state_dict(checkpoint['model_state_dict'])
-        return model
-    else:
-        raise RuntimeError("Cannot find a pretrained model with name {}".format(model_name))
+    elif model_name[:4] == 'DGMG':
```
VoVAllen: using `model.startswith`

examples/pytorch/model_zoo/chem/generative_models/dgmg/sascorer.py
```diff
@@ -0,0 +1,178 @@
...
+        sascore = 10.0
+    elif sascore < 1.:
+        sascore = 1.0
+
+    return sascore
+
+def processMols(mols):
+    print('smiles\tName\tsa_score')
+    for i, m in enumerate(mols):
+        if m is None:
```
VoVAllen: What's this?

mufeili: This is a descriptor computation function from rdkit. It fails to handle a corner case so some modifications are performed.

mufeili:
@VoVAllen I do not prefer moving files to `example/dgmg`. We should only import from modules like `dgl`. Examples themselves should stand alone and to import across examples should not be encouraged. In addition, many of the utilities are not related to the previous DGMG example and this can lead to additional confusion. I think it is fine as long as we make the entry points clear enough.

python/dgl/model_zoo/chem/pretrain.py
mufeili: Done.

examples/pytorch/model_zoo/chem/generative_models/dgmg/README.md
mufeili: Agree. I've made an update and see if you feel more comfortable now.

examples/pytorch/model_zoo/chem/generative_models/dgmg/eval.py
```diff
@@ -0,0 +1,161 @@
...
+    # Setup dataset and data loader
+    dataset = MoleculeDataset(args['dataset'], subset_id=rank, n_subsets=args['num_processes'])
+    env = dataset.env
+
+    # Initialize model
+    if rank == args['num_processes'] - 1:
+        print('Loading the trained model...')
+
+    if not args['pretrained']:
+        model = model_zoo.chem.DGMG(env=env, node_hidden_size=args['node_hidden_size'],
```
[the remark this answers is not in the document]
mufeili: Users will always run `train.py` first, in which case pre-processing for new datasets will be performed. `env` objects are created with dataset initialization and users should not directly instantiate them.

mufeili:
> Minor comments:
> 
> * Should `MoleculeEnv` be part of model or dataset? Looks can be part of model to me instead of passing from dataset to model
> * What's the goal of DGMG?
> * I assume user would only call `model_zoo.chem.DGMG` in their own files, therefore I wouldn't request changes on other parts since they are encapsulated in dgl.

1. `MoleculeEnv` cannot be instantiated without the particular dataset as it depends on the type of atoms and bonds appearing in the dataset. Also in preprocessing datasets we need to directly initialize them for verification.
2. To learn the distribution of molecules so that we can get new molecules with similar properties.

VoVAllen:
This would make dgmg purely an example instead of a model zoo api. If you don't expect it to be called, why we would like to put it in the model zoo instead of an example? The purpose of model zoo API is for user to easily reuse the model code.

mufeili:
@VoVAllen This is the nature of many complex models like the generative models. You can develop your own code if you really want. But in most cases it will be easiest to use the provided code. I imagine the same story for JTNN.

VoVAllen:
It's just a design problem. For JTNN I just put everything including vocab into model_zoo/chem/gen_model/jtnn (not exposed to user), and provide one default option for this. If some utils/dataset is necessary for the model to use, you can also put them in the folders. This is not a problem with model complexity, just a design for the interface. The key is to make the `env` optional, which may not prevent user from using it at the first look.

examples/pytorch/model_zoo/chem/generative_models/dgmg/README.md
```diff
@@ -0,0 +1,107 @@
+# Learning Deep Generative Models of Graphs (DGMG)
+
+Yujia Li, Oriol Vinyals, Chris Dyer, Razvan Pascanu, and Peter Battaglia. 
+Learning Deep Generative Models of Graphs. *arXiv preprint arXiv:1803.03324*, 2018.
+
+DGMG generates graphs by progressively adding nodes and edges as below:
+![](https://user-images.githubusercontent.com/19576924/48605003-7f11e900-e9b6-11e8-8880-87362348e154.png)
```
VoVAllen: > To learn the distribution of molecules so that we can get new molecules with similar properties. 

Add tasks description here

examples/pytorch/model_zoo/chem/generative_models/dgmg/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+![](https://s3.us-east-2.amazonaws.com/dgl.ai/model_zoo/drug_discovery/dgmg/tensorboard.png)
+
+To use tensorboard, you need to install [tensorboardX](https://github.com/lanpa/tensorboardX) and 
+[TensorFlow](https://www.tensorflow.org/). You can lunch tensorboard with `tensorboard --logdir=.`
+
+### Evaluation
+
+To evaluate your own model, do `python eval.py -d X -o Y -p Z -np A -mn B -gt C`, where:
+- `X` is the dataset to use, such as `ChEMBL` and `ZINC`
```
VoVAllen:
```
optional arguments:
  -h, --help            show this help message and exit
  -s SEED, --seed SEED  random seed
  -w WARMUP_EPOCHS, --warmup-epochs WARMUP_EPOCHS
                        Number of epochs where no lr decay is performed.
  -d DATASET, --dataset DATASET
                        dataset to use
  -o {random,canonical}, --order {random,canonical}
                        order to generate graphs
  -tf TRAIN_FILE, --train-file TRAIN_FILE
                        Path to a file with one SMILES a line for training
                        data. This is only necessary if you want to use a new
                        dataset.
  -vf VAL_FILE, --val-file VAL_FILE
                        Path to a file with one SMILES a line for validation
                        data. This is only necessary if you want to use a new
                        dataset.
  -l LOG_DIR, --log-dir LOG_DIR
                        folder to save info like experiment configuration
  -np NUM_PROCESSES, --num-processes NUM_PROCESSES
                        number of processes to use
```

XYZABC is hard to track when reading for me. `python train.py --help` generate acceptable result shows above. You can merge your description to python file. And I found [this](https://stackoverflow.com/questions/12151306/argparse-way-to-include-default-values-in-help) could add default value to `--help` also

examples/pytorch/model_zoo/chem/generative_models/dgmg/README.md
```diff
@@ -0,0 +1,107 @@
...
+By default we use `25`.
+- `C` is the max time (in seconds) allowed for molecule generation. By default we use `600` (10 minutes).
+
+To use a pretrained model, replace `-p Z` with `-pr`.
+
+After the evaluation, 100000 molecules will be generated and stored in `generated_smiles.txt` under `eval_results`
+directory, with three statistics logged in `generation_stats.txt` under `eval_results`:
+1. `Validity among all` gives the percentage of molecules that are valid
+2. `Uniqueness among valid ones` gives the percentage of valid molecules that are unique
+3. `Novelty among unique ones` gives the percentage of unique valid molecules that are novel (not seen in training data)
```
VoVAllen: Mention a little bit about evaluation metric when describing task above.

examples/pytorch/model_zoo/chem/generative_models/dgmg/README.md
mufeili: Down.

Issue #785, opened 2019-08-23
[RFC] Builtin mean reduce function.


yzh119:
## 🚀 Feature
Many GNN models use mean reducer:
![image](https://user-images.githubusercontent.com/11773619/63561906-2f3c0000-c58e-11e9-9cb0-1846a00690ad.png)
Currently we only support sum/max/min/prod as builtin reducers, there should be a builtin mean reducer in `dgl.function.reducer`.

## Alternatives

Sum reduce + normalization(by in degrees) has the same the effect as mean reducer:
```python
import dgl.function as fn
g.update_all(fn.copy_u('h', 'm'), fn.sum('m', 'h'))
degs = graph.in_degrees().float().unsqueeze(-1)
g.ndata['h'] = g.ndata['h'] / degs
```
However, it's more convenient to call `fn.mean` directly.

## Pitch
Currently we do not have kernel support for mean reducers, I was wondering is it possible to decompose `fn.mean`to a reduce function `fn.sum` and an apply_node function that normalize the result.

Another concern is how do we deal with zero degree nodes? I prefer return an all-zero vector for these nodes.

mufeili:
I agree with both points.

jermainewang:
Since we don't have a fully functional int32 graph or CUDA graph, the question is where to generate the correct degree tensor. Apparently, we could not compute the degree tensor inside `binary_reduce` kernel function (https://github.com/dmlc/dgl/blob/master/python/dgl/kernel.py#L26) due to the aforementioned reason. I think as a workaround we could put the code right before the graph is converted to the required index type and device context. @ylfdq1118 could you provide some advises on this?

ylfdq1118:
Personally, I think adding the norm at kernel function level is the cleanest way. But as Minjie mentioned, unfortunately, we only support graph structure query for 64-bit CPU graph.

In current scheduler design, builtin function interfaces are unified, and scheduler does not need to distinguish which reduce function is used (in other words, the code path for max/min/sum/prod is exactly the same). So it's actually not easy to add node degree for mean.

After stepping into this line:
https://github.com/dmlc/dgl/blob/708765f0a11d29b3d3f27857aeebfd10498f1937/python/dgl/runtime/scheduler.py#L869
you will lose access to the original 64-bit CPU DGLGraph object. So you need to check the reduce function and generate the node degree tensor before this line and pass it all the way down to kernel. And I believe this would require non-trivial and tedious code change.

[...]

PR #787, opened 2019-08-23
[Feature] Add builtin mean reducer

yzh119:
## Description
Implement the builtin mean reducer mentioned in #785. 

## Example
PyTorch
```python
>>> import dgl
>>> import dgl.function as fn
>>> import torch as th
>>> g = dgl.DGLGraph()
>>> g.add_nodes(4)
>>> g.add_edges([0, 0, 0, 1, 1, 2], [0, 1, 2, 1, 2, 2])
>>> feat = th.arange(1, 5).float().unsqueeze(-1)
>>> feat.requires_grad = True
>>> g.ndata['x'] = feat
>>> g.update_all(fn.copy_u('x', 'm'), fn.mean('m', 'o'))
>>> print(g.ndata['o'])
tensor([[1.0000],
        [1.5000],
        [2.0000],
        [0.0000]], grad_fn=<CopyReduceBackward>)
>>> g.ndata['o'].sum().backward()
>>> print(g.ndata['x'].grad)
tensor([[1.8333],
        [0.8333],
        [0.3333],
        [0.0000]], device='cuda:0')
```

MXNet:
```python
>>> import dgl
>>> import dgl.function as fn
>>> import mxnet as mx
>>> from mxnet import nd
>>> g = dgl.DGLGraph()
>>> g.add_nodes(4)
>>> g.add_edges([0, 0, 0, 1, 1, 2], [0, 1, 2, 1, 2, 2])
>>> feat = nd.arange(1, 5).astype('float32').expand_dims(1) #th.arange(1, 5).float().unsqueeze(-1)
>>> feat.attach_grad()
>>> g.ndata['x'] = feat
>>> with mx.autograd.record():
>>>    g.update_all(fn.copy_u('x', 'm'), fn.mean('m', 'o'))
>>>    print(g.ndata['o'])

[[1. ]
 [1.5]
 [2. ]
 [0. ]]
<NDArray 4x1 @cpu(0)>
>>>    g.ndata['o'].sum().backward()
>>>    print(g.ndata['x'].grad)

[[1.8333334 ]
 [0.8333334 ]
 [0.33333334]
 [0.        ]]
<NDArray 4x1 @cpu(0)>
```

The combination of `*_op_* + mean` is also supported. 

## Checklist
- [x] Code is well-documented

## Changes
- The changes are only on python end: in `dgl.backend.*.tensor.py`, if the reducer is of type `mean`, we convert it to `sum`, and pre/post process the output/gradients accordingly.


python/dgl/backend/mxnet/tensor.py
```diff
@@ -368,28 +368,57 @@ def forward(self, lhs_data, rhs_data):
                             ctx=lhs_data.context, dtype=lhs_data.dtype)
         out_data_nd = zerocopy_to_dgl_ndarray_for_write(out_data)
         K.binary_op_reduce(
-            self.reducer, self.binary_op, self.graph, self.lhs, self.rhs,
+            self.reducer if self.reducer != 'mean' else 'sum',
+            self.binary_op, self.graph, self.lhs, self.rhs,
             lhs_data_nd, rhs_data_nd, out_data_nd, self.lhs_map[0],
             self.rhs_map[0], self.out_map[0])
+        # normalize if mean reducer
```
BarclayII: Could you leave a note that this is a temporary hack and we should have a better solution (either in kernel or to support int32 queries) in the future?

yzh119: DONE

python/dgl/backend/mxnet/tensor.py
```diff
@@ -368,28 +368,58 @@ def forward(self, lhs_data, rhs_data):
...
+        # note(zihao): this is a temporal hack and we should have better solution in the future.
```
jermainewang: note -> NOTE
temporal -> temporary

python/dgl/backend/mxnet/tensor.py
```diff
@@ -368,28 +368,58 @@ def forward(self, lhs_data, rhs_data):
...
+        if self.reducer == 'mean':
+            degs = nd.empty((out_data.shape[0],),
+                            ctx=out_data.context, dtype=out_data.dtype)
+            degs_nd = zerocopy_to_dgl_ndarray(degs)
+            if self.lhs != 1: # 1: target code of dst node
```
jermainewang: Use the TargetCode in dgl.function.base

python/dgl/backend/mxnet/tensor.py
```diff
@@ -368,28 +368,58 @@ def forward(self, lhs_data, rhs_data):
...
+                in_ones = nd.ones((lhs_data.shape[0],),
+                                  ctx=lhs_data.context, dtype=lhs_data.dtype)
+                in_ones_nd = zerocopy_to_dgl_ndarray(in_ones)
+                K.copy_reduce(
+                    'sum', self.graph, self.lhs, in_ones_nd, degs_nd, 
+                    self.lhs_map[0], self.lhs_map[0])
+            else: # self.rhs != 1
```
jermainewang: I don't quite understand this if here. Shouldn't we always compute in-coming degrees?

python/dgl/backend/mxnet/tensor.py
```diff
@@ -423,18 +453,39 @@ def forward(self, in_data):
         in_data_nd = zerocopy_to_dgl_ndarray(in_data)
         out_data_nd = zerocopy_to_dgl_ndarray_for_write(out_data)
         K.copy_reduce(
-            self.reducer, self.graph, self.target, in_data_nd, out_data_nd,
+            self.reducer if self.reducer != 'mean' else 'sum',
+            self.graph, self.target, in_data_nd, out_data_nd,
             self.in_map[0], self.out_map[0])
-        self.save_for_backward(in_data_nd, out_data_nd)
+        # normalize if mean reducer
+        # note(zihao): this is a temporal hack and we should have better solution in the future.
```
jermainewang: note -> NOTE
temporal -> temporary

python/dgl/backend/pytorch/tensor.py
```diff
@@ -289,34 +289,59 @@ def forward(ctx, reducer, binary_op, graph, lhs, rhs, lhs_data, rhs_data,
         out_data = lhs_data.new_empty((out_size,) + feat_shape)
         out_data_nd = zerocopy_to_dgl_ndarray(out_data)
         K.binary_op_reduce(
-            reducer, binary_op, graph, lhs, rhs, lhs_data_nd, rhs_data_nd,
+            reducer if reducer != 'mean' else 'sum', 
+            binary_op, graph, lhs, rhs, lhs_data_nd, rhs_data_nd,
             out_data_nd, lhs_map[0], rhs_map[0], out_map[0])
+        # normalize if mean reducer
+        # note(zihao): this is a temporal hack and we should have better solution in the future.
```
jermainewang: note -> NOTE
temporal -> temporary

python/dgl/backend/pytorch/tensor.py
```diff
@@ -289,34 +289,59 @@ def forward(ctx, reducer, binary_op, graph, lhs, rhs, lhs_data, rhs_data,
...
+        if reducer == 'mean':
+            degs = lhs_data.new_empty((out_data.shape[0],))
+            degs_nd = zerocopy_to_dgl_ndarray(degs)
+            if lhs != 1: # 1: target code of dst node
+                in_ones = lhs_data.new_ones((lhs_data.shape[0],))
+                in_ones_nd = zerocopy_to_dgl_ndarray(in_ones)
+                K.copy_reduce(
+                    'sum', graph, lhs, in_ones_nd, degs_nd, lhs_map[0], lhs_map[0]) 
+            else: # rhs != 1
```
jermainewang: same here

python/dgl/backend/mxnet/tensor.py
yzh119: Yes we should always compute in-coming degrees.
When the input target is `src` or `edge`,  the output of this `copy_reduce` call is in-degrees of each node. But when the input target is `dst`, what's the expected behavior?
The `if` statement is to discriminate the case where lhs is on dst target (e.g. `v_op_e`)

yzh119: The real problem is `copy_reduce` does not accept input target as dst nodes.
```
dgl._ffi.base.DGLError: [03:39:04] /home/ec2-user/dgl/src/kernel/cpu/../binary_reduce_impl.h:104: Unsupported operation: op=use_lhs lhs=1 rhs=3
```
So I need to avoid this case.

python/dgl/backend/mxnet/tensor.py
yzh119: DONE

python/dgl/backend/mxnet/tensor.py
yzh119: DONE

python/dgl/backend/mxnet/tensor.py
```diff
@@ -368,28 +369,58 @@ def forward(self, lhs_data, rhs_data):
...
+                K.copy_reduce(
+                    'sum', self.graph, self.lhs, in_ones_nd, degs_nd, 
+                    self.lhs_map[0], self.lhs_map[0])
+            else: # self.rhs != TargetCode.DST
+                in_ones = nd.ones((rhs_data.shape[0],),
+                                  ctx=rhs_data.context, dtype=rhs_data.dtype)
+                in_ones_nd = zerocopy_to_dgl_ndarray(in_ones)
+                K.copy_reduce(
+                    'sum', self.graph, self.rhs, in_ones_nd, degs_nd, 
+                    self.rhs_map[0], self.rhs_map[0])
```
jermainewang: Does the following work?
```python
degs = nd.empty((out_data.shape[0],), ctx=out_data.context, dtype=out_data.dtype)
degs_nd = zerocopy_to_dgl_ndarray(degs)
in_ones = nd.ones((graph.number_of_nodes(),), ctx=lhs_data.context, dtype=lhs_data.dtype)
in_ones_nd = zerocopy_to_dgl_ndarray(in_ones)
K.copy_reduce('sum', self.graph, TargetCode.SRC, in_ones_nd, degs_nd)
```

The mapping arrays can be omitted in your case.

yzh119: Yes, it works. But I was wondering why could we not use mapping information? What if we update a subgraph instead of the whole graph?

yzh119: btw, by calling `g.send_and_recv(eids, msg_func, reduce_func)` I met another bug:
```
  File "/home/ec2-user/dgl/python/dgl/graph.py", line 2523, in send_and_recv
    inplace=inplace)
  File "/home/ec2-user/dgl/python/dgl/runtime/scheduler.py", line 183, in schedule_snr
    out_map_creator=out_map_creator)
  File "/home/ec2-user/dgl/python/dgl/runtime/scheduler.py", line 859, in _gen_send_reduce
    adj, edge_map, nbits = adj_creator()
  File "/home/ec2-user/dgl/python/dgl/runtime/scheduler.py", line 171, in <lambda>
    edge_tuples, graph._number_of_src_nodes(), graph._number_of_dst_nodes())
  File "/home/ec2-user/dgl/python/dgl/runtime/spmv.py", line 181, in build_gidx_and_mapping_uv
    gidx = create_bipartite_from_coo(num_src, num_dst, u, v)
  File "/home/ec2-user/dgl/python/dgl/heterograph_index.py", line 799, in create_bipartite_from_coo
    return _CAPI_DGLHeteroCreateBipartiteFromCOO(
NameError: name '_CAPI_DGLHeteroCreateBipartiteFromCOO' is not defined
```

yzh119: I'm afraid not using mapping is problematic, for example:
```python
import dgl
import dgl.function as fn
import torch as th
g = dgl.DGLGraph()
g.add_nodes(4)
g.add_edges([2, 0, 0, 1, 1, 0], [2, 1, 2, 1, 2, 0])
feat = th.arange(1, 5).float().unsqueeze(-1).cuda()
feat.requires_grad = True
g.ndata['x'] = feat
g.edata['x'] = th.ones(6, 1).cuda()
g.send_and_recv(th.LongTensor([0, 2, 4, 5]), fn.u_mul_e('x', 'x', 'm'), fn.mean('m', 'o'))
print(g.ndata['o'])
g.ndata['o'].sum().backward()
print(g.ndata['x'].grad)
```
With mapping information, the output is:
```
tensor([[1.],
        [0.],
        [2.],
        [0.]], device='cuda:0', grad_fn=<IndexCopyBackward>)
tensor([[1.3333],
        [0.3333],
        [0.3333],
        [0.0000]], device='cuda:0')
```
Without mapping information, the output is:
```python
tensor([[1.],
        [0.],
        [6.],
        [0.]], device='cuda:0', grad_fn=<IndexCopyBackward>)
tensor([[2.],
        [1.],
        [1.],
        [0.]], device='cuda:0')
```

yzh119: `K.copy_reduce('sum', self.graph, TargetCode.SRC, in_ones_nd, degs_nd, None, self.out_map[0])` would solve the problem.

jermainewang:
LGTM. Approved!

python/dgl/backend/mxnet/tensor.py
ylfdq1118: @jermainewang When doing `send_and_recv`, `self.graph` here is a bipartite graph index, and using `graph.number_of_nodes()` can be problematic. Here if we use `copy_reduce(target=src)`, we actually need `number_of_nodes` to return number of source nodes.

I think an alternative is to use `copy_edge` for computing node degree. And then the one tensor can be created with `nd.ones((graph.number_of_edges(),))`.

However, the problem is the bipartite graph interface is slightly different. I checked the C++ implementation of bipartite graph, and found that it subclasses heterograph and therefore its API needs to take in an argument `etype`. For example:
https://github.com/dmlc/dgl/blob/73b2668fe260017bb09e74553b65dc4a6094afa6/src/graph/bipartite.cc#L109-L111
The `etype` argument was not really used, but it makes the bipartite graph index interface different from ordinary graph index. @jermainewang @BarclayII I wonder why bipartite graph is designed this way. Is there a reason bipartite graph needs to handle more than one edge type?

ylfdq1118:
Current solution from @yzh119 is correct. However, it's not very straight-forward. I wonder if we can use `copy_reduce(target=code)` for node degree. We need to fix bipartite graph's interface... (see inline comment for detail)

python/dgl/backend/mxnet/tensor.py
jermainewang: The bipartite graph interface is not stable enough for use. Let's not add extra complexity here. For @ylfdq1118 's question, etype here is simply 0. A check could be enforced too. We will gradually move all the kernels to general sparse matrix so we will definitely come back to this later.

Issue #719, opened 2019-07-18
tutorials and models for point clouds dataset 


yangj257:
point clouds processing is also an important application area for GCN，But I didn't find anything related in the tutorial.Could you please add tutorials and add examples about processing point cloud datasets(the small dataset like ModelNet10, S3DIS, and the big dataset like Semantic 3D).
If tutorials include examples about converting the point cloud dataset into a graph dataset,and implementing the models in some papers like“Graph Attention Convolution for Point Cloud Segmentation” (paper link: https://engineering.purdue.edu/~jshan/publications/2018/Lei%20Wang%20Graph%20Attention%20Convolution%20for%20Point%20Cloud%20Segmentation%20CVPR2019.pdf), that will be much better.

EtoDemerzel0427:
Yes, I noticed that compared with Pytorch Geometric, DGL gives us fewer examples and tutorials.  I think to make an open-source library popular, you have to pay more attention to the documentation so as to smooth the learning curve for new users.

 It will be more attractive if you could keep track of the recent advances of more GCN models and implement them in DGL, because researchers will feel like to reproduce results in a uniform platform instead of setting up a new enviroment for a new model again and again. To this point, pytorch geometric is better, but their documentation is not that good either.

jermainewang:
Hey there, paste my reply to the other similar question on the discussion forum (https://discuss.dgl.ai/t/dgl-vs-pytorch-geometric/346/6).

Overall, I think both frameworks have their merits. PyG is very light-weighted and has lots of off-the-shelf examples. In DGL, we put a lot of efforts to cover a wider range of scenarios. Many of them are not necessarily GNNs but share the principles of structural/relational learning. Examples are [CapsuleNet](https://docs.dgl.ai/tutorials/models/4_old_wines/2_capsule.html), [Transformer](https://docs.dgl.ai/tutorials/models/4_old_wines/7_transformer.html) and [TreeLSTM](https://docs.dgl.ai/tutorials/models/2_small_graph/3_tree-lstm.html). We also notice that graph generative models are important and could be quite flexible (i.e, adding/removing one node/edge at a time), so we spend extra efforts in the design for them (examples are [DGMG](https://docs.dgl.ai/tutorials/models/3_generative_model/5_dgmg.html) and [JTNN](https://github.com/dmlc/dgl/tree/master/examples/pytorch/jtnn)). Many of them have not been covered by other frameworks.

[...]

PR #789, opened 2019-08-26
[Example] Dynamic Graph CNN on Point Cloud

BarclayII:
## Description
This PR implements the [Dynamic Graph CNN paper](https://arxiv.org/pdf/1801.07829.pdf), as well as the `EdgeConv` module and `DynamicEdgeConv` module in #748 .  Namely, it provides
* `EdgeConv` in the said paper.
* `NearestNeighborGraph`: A module that constructs a graph from a feature matrix based on nearest neighbors.
* (Alternating between `NearestNeighborGraph` and `EdgeConv` yields `DynamicEdgeConv`.)

Hopefully this could be a starting point for issue #719.

This runs on ModelNet40 dataset and got a best validation accuracy of 93.5% and test accuracy of 91.8%.

## Checklist
- [x] Code is well-documented


yzh119:
@BarclayII , could you please move the EdgeConv module to `dgl.nn.*.conv`? Plus, I recommend putting NearestNeighborGraph to `dgl.transforms`, thanks!

BarclayII:
> @BarclayII , could you please move the EdgeConv module to `dgl.nn.*.conv`? Plus, I recommend putting NearestNeighborGraph to `dgl.transforms`, thanks!

I'm actually thinking what the form of `EdgeConv` and `NearestNeighborGraph` should be.  In particular, whether we want to support point cloud with different number of points.

BarclayII:
Temporarily closing to save CI as I'm migrating code to `dgl.nn`.

python/dgl/transform.py
```diff
@@ -4,15 +4,113 @@
...
 from .graph import DGLGraph
+from . import backend as F
 from .graph_index import from_coo
 from .batched_graph import BatchedDGLGraph, unbatch
-from .backend import asnumpy, tensor
 
 
 __all__ = ['line_graph', 'khop_adj', 'khop_graph', 'reverse', 'to_simple_graph', 'to_bidirected',
-           'laplacian_lambda_max']
+           'laplacian_lambda_max', 'nearest_neighbor_graph', 'segmented_nearest_neighbor_graph']
```
yzh119: Please also mention these two functions in https://github.com/dmlc/dgl/blob/master/docs/source/api/python/transform.rst

python/dgl/nn/pytorch/factory.py
```diff
@@ -0,0 +1,103 @@
+"""Modules that transforms between graphs and between graph and tensors."""
+import torch.nn as nn
+from ...transform import nearest_neighbor_graph, segmented_nearest_neighbor_graph
+
+def pairwise_squared_distance(x):
```
yzh119: May I ask why these functions has to be `nn.Modules`, it seems they have no parameters and are just wrappers of `nearest_neighbor_graph` and `segmented_nearest_neighbor_graph`.

BarclayII: Anything can be a neural network module.  We also have both `torch.nn.functional.relu` and `torch.nn.ReLU` which is a wrapper of the former.

PR #797, opened 2019-08-27
[Test] set seed for nn test graph generation

BarclayII:


BarclayII:
Closed since this is intended for debugging CI and ChebConv

Issue #784, opened 2019-08-22
[Kernel][Perf] Message builtin with broadcasting is slow compared to UDF


jermainewang:
## 🐛 Performance Issue

One example is when using `u_mul_e` with broadcasting, the speed is quite slow compared to using UDF. Note that UDF should ideally be slower than built-in because node data needs to be duplicated to edges. This issue is found in NNConv module #779 .

## To Reproduce

Use the following testing script:
```python
import dgl
import time
import numpy as np
import scipy as sp
import torch

adj = sp.sparse.random(550, 550, 0.05)
G = dgl.DGLGraph(adj)
N = G.number_of_nodes()
M = G.number_of_edges()
print(N)
print(M)

def udf(edges):
    return {'m' : edges.src['h'] * edges.data['w']}

def time_udf(G):
    dur = []
    for i in range(50):
        if i >= 5:
            t0 = time.time()
        G.update_all(udf, dgl.function.sum('m', 'hh'))
        if i >= 5:
            dur.append(time.time() - t0)
    return np.average(dur)

def time_builtin(G):
    dur = []
    for i in range(50):
        if i >= 5:
            t0 = time.time()
        G.update_all(dgl.function.u_mul_e('h', 'w', 'm'), dgl.function.sum('m', 'hh'))
        if i >= 5:
            dur.append(time.time() - t0)
    return np.average(dur)

G.ndata['h'] = torch.randn((N, 64, 1))
G.edata['w'] = torch.randn((M, 64, 64))
G.to(torch.device('cuda:0'))
print('udf time:', time_udf(G))
print('builtin time:', time_builtin(G))

G.ndata['h'] = torch.randn((N, 64, 64))
G.edata['w'] = torch.randn((M, 64, 64))
G.to(torch.device('cuda:0'))
print('udf time:', time_udf(G))
print('builtin time:', time_builtin(G))
```

Run it with `CUDA_LAUNCH_BLOCKING=1 python t.py`. On a P3 machine, I got following result:

```
550
15125
udf time: 0.0021312501695421007
builtin time: 0.003366051779852973
udf time: 0.003186459011501736
builtin time: 0.001470041275024414
```

As you can see, when node feature shape is (64, 1) and edge feature shape is (64, 64), builtin kernel is ~1.5 times slower than UDF. As a comparison, if the node feature shape is also (64, 64) so no broadcasting is required, DGL builtin is much faster as expected.

## Additional context

[...]

PR #815, opened 2019-08-31
[Kernel][Perf] Message builtin with broadcasting performance optimization

classicsong:
## Description
Here is the performance optimization for builtin broadcasting opts (e.g., u_add_v, u_mul_v, etc.). 
This optimization is based on edge partition (current minigun version), this may need further optimization if we change the underling minigum graph execution policy.

The optimization includes:
1. Combined Unravel and Ravel to improve data loading pipline.
2. Figuring out which one side(lhs, rhs) is to broadcast to eliminate unnecessary ops.

The test code follows #784, The overall performance is (The test was done on a P3 machine):

| A mul B (ms) | UDF | Original | URRevel | Original/UDF | URRevel/UDF |
| --- | --- | --- | --- | --- | --- |
| 64x1,64x16 | 0.957447 | 1.238871 | 1.086717 | 129% | 114% |
| 64x16,64x1 | 1.143344 | 1.201614 | 1.047124 | 105% | 92% |
| 64x1,64x32 | 1.334397 | 1.917760 | 1.579565 | 144% | 118% |
| 64x32,64x1 | 1.709864 | 1.720285 | 1.431730 | 101% | 84% |
| 64x1,64x64 | 2.072053 | 2.924156 | 2.454148 | 141% | 118% |
| 64x64,64x1 | 2.900102 | 2.776416 | 2.311436 | 96% | 80% |
| 256x1,256x32 | 3.796005 | 5.262550 | 4.375209 | 139% | 115% |
| 256x32,256x1 | 5.252356 | 4.976400 | 4.107655 | 95% | 78% |
| 256x1,256x64 | 6.907299 | 9.877385 | 8.164803 | 143% | 118% |
| 256x64,256x1 | 9.879049 | 9.287098 | 7.624669 | 94% | 77% |
| 256x1,256x128 | 13.496282 | 19.118198 | 15.871180 | 142% | 118% |
| 256x128,256x1 | 19.603629 | 17.201047 | 13.960610 | 88% | 71% |
| Overall |   |   |   | 118% | 99% |

#784 

## Checklist
- [x] Code is well-documented


src/kernel/cuda/binary_reduce_impl.cuh
```diff
@@ -72,6 +72,53 @@ __device__ __forceinline__ int64_t Ravel(
   return out;
 }
 
+/*
+ * Merge convert flattened index to multi-dimension index according to output shape
+ * and Convert multi-dimension index to flattened index for lhs or rhs (assume row-major).
+ */
+__device__ __forceinline__ void URRavel(
```
yzh119: This function basically fuses unravel with lhs_ravel and rhs_ravel. Besides, it uses loop unroll to reduce the overhead of for loop.
I think the function name is not intuitive, could you provide an explaination in comments?

src/kernel/cuda/binary_reduce_impl.cuh
```diff
@@ -72,6 +72,53 @@ __device__ __forceinline__ int64_t Ravel(
...
+    int64_t idx, int ndim, const int64_t* shape, const int64_t* stride,
```
yzh119:
```suggestion
    int64_t idx, int ndim, const int64_t* out_shape, const int64_t* out_stride,
```

src/kernel/cuda/binary_reduce_impl.cuh
```diff
@@ -72,6 +72,53 @@ __device__ __forceinline__ int64_t Ravel(
...
+    for (int d = 0; d < ndim; ++d) {
+      int64_t o_sh = shape[d];
+      int64_t o_st = stride[d];
+      int64_t rhs_sh = rhs_shape[d];
+      int64_t rhs_st = rhs_stride[d];
+      
+      int64_t i = (idx / o_st) % o_sh;
+      /*
+       * Simplfied for rhs_out += min(i, rhs_sh - 1) * rhs_st;
+       * rhs_sh be N or 1
```
yzh119: explain what is `N`, I suspect it is `o_sh`?

src/kernel/cuda/binary_reduce_impl.cuh
```diff
@@ -72,6 +72,53 @@ __device__ __forceinline__ int64_t Ravel(
...
+       */
+      if (rhs_sh > i) {
```
yzh119: I'm not sure which one is more intuitive, if you use `if` style, I suggest not to mention `min(i, rhs_sh - 1) * rhs_st` in the doc, which might be confusing.

src/kernel/cuda/binary_reduce_impl.cuh
```diff
@@ -72,6 +72,53 @@ __device__ __forceinline__ int64_t Ravel(
...
+      int64_t o_st = stride[d];
+      int64_t lhs_sh = lhs_shape[d];
+      int64_t lhs_st = lhs_stride[d];
+  
+      int64_t i = (idx / o_st) % o_sh;
+      /*
+       * Simplfied for lhs_out += min(i, lhs_sh - 1) * lhs_st;
+       * lhs_sh be N or 1
+       */
+      if (lhs_sh > i) {
```
yzh119: ditto

yzh119:
If we use this style, we should remove `ravel` function and `unravel` function.

src/kernel/cuda/binary_reduce_impl.cuh
jermainewang: +1 with the function name. We generally recommend using full name for C++ functions and variables.

src/kernel/cuda/binary_reduce_impl.cuh
```diff
@@ -72,6 +72,53 @@ __device__ __forceinline__ int64_t Ravel(
...
 }
 
+/*
+ * Merge convert flattened index to multi-dimension index according to output shape
+ * and Convert multi-dimension index to flattened index for lhs or rhs (assume row-major).
+ */
+__device__ __forceinline__ void URRavel(
+    int64_t idx, int ndim, const int64_t* shape, const int64_t* stride,
+    const int64_t* lhs_shape, const int64_t* lhs_stride,
+    const int64_t* rhs_shape, const int64_t* rhs_stride, int64_t &lhs_out, int64_t &rhs_out) {
```
jermainewang: For output argument, I suggest use `int64_t* lhs_out` so it's more obvious.

src/kernel/cuda/binary_reduce_impl.cuh
```diff
@@ -98,13 +145,14 @@ struct BinaryReduceBcast {
     DType* lhsoff = gdata->lhs_data + lid * gdata->lhs_len;
     DType* rhsoff = gdata->rhs_data + rid * gdata->rhs_len;
     DType* outoff = gdata->out_data + oid * gdata->out_len;
-    int64_t tmp[NDim];  // store unraveled idx.
```
jermainewang: Does this mean `NDim` template argument is not needed any more?

classicsong: > Does this mean `NDim` template argument is not needed any more?

Yes, it is not needed in this func. Maybe further opt may need this again.

src/kernel/cuda/binary_reduce_impl.cuh
classicsong: I think rhs_out += min(i, rhs_sh - 1) * rhs_st; is more intuitive to understand.
So I left it in the comment.

src/kernel/cuda/binary_reduce_impl.cuh
```diff
@@ -54,24 +54,56 @@ struct BinaryReduce {
...
-  for (int d = 0; d < ndim; ++d) {
-    out[d] = (idx / stride[d]) % shape[d];
+/*
+ * This func do the followings:
+ *   1. Convert flattened index to multi-dimension index
+ *      according to output shape (assume row-major).
+ *   2. Convert multi-dimension index to flattened index for lhs.
+ *   3. Convert multi-dimension index to flattened index for rhs.
+ */
+__device__ __forceinline__ void UnravelRravel(
```
yzh119: Why not `UnravelRavel`?

classicsong: > Why not `UnravelRavel`?

oops

src/kernel/cuda/binary_reduce_impl.cuh
jermainewang: Could you remove the argument? It will make the compilation faster.

classicsong: > Could you remove the argument? It will make the compilation faster.

Oops, NDim is not need by the code, but it is needed by one of the arguments of ApplyEdge func: static __device__ __forceinline__ void ApplyEdge(Idx src, Idx dst, Idx eid, **BcastGData<NDim, Idx, DType>* gdata**) {

Issue #737, opened 2019-08-03
[RFC] Graph level data


jermainewang:
Moved from #714 . Currently, we have node-level and edge-level data (e.g. `ndata` and `edata`). The RFC proposes to add a dictionary for graph-level data (using torch syntax as an example):

```python
g.gdata['embed'] = torch.randn((100,)) # could be multi-dimension array
g.gdata['label'] = torch.array([0]) 
```

I could foresee following considerations
1. **Restriction on the graph data.** Our node-level data and edge-level data have to be ndarrays. Shall we restrict the graph data to be the same? One subtlety is that when setting scalar type graph data, the user needs to use a length-one array (the 'label' data in the above example).
2. **Batching multiple graphs.** `dgl.batch` needs to handle graph-level data too. The behavior should refer to how node/edge-level data are batched. For example, we need to throw errors if graph-level data under the same key cannot be stacked (or the key is missing for some graphs). Similar change needs to be made to `dgl.unbatch`.
3. **Graph storage format.** A new section to store graph-level data needs to be added. This means if the RFC is accepted, it should wait for PR #728 to be finished.
4. **Message passing UDFs.** We might need to add graph-level data frame to the arguments of node/edge UDFs.
  ```python
  def my_edge_udf(edges, gdata):
      pass
  def my_node_udf(nodes, gdata):
      pass
  ```
  Compared with our previous suggestion to use function closure:
  ```python
  some_graph_data = ...
  def my_edge_udf(edges):
      foo(some_graph_data)  # could use graph data here
      pass
  ```
  Also, the change will break backward compatibility. Any good idea?

## Alternatives
Note that graph-level data could be maintained outside of DGLGraph because they are dense tensors and the most common operations are stacking/unstacking them. UDFs can also access them using function closure which is quite neat TBH. On the other hand, adding graph-level data needs change to the UDF signatures and potentially breaks user codes.

VoVAllen:
Can we use edges.g.data/nodes.g.data in UDF instead of another argument graph_data?

mufeili:
1. What can be some other possibilities if we do not restrict graph data to be ndarrays?
2. We can also allow users to specify initializers if a key does not exist for some graphs.
3. If there will not exist really serious performance issues, I think I'd rather prefer the suggested alternative and make the handle light.

VoVAllen:
[...]

PR #824, opened 2019-09-02
[Hetero] New syntax

jermainewang:
## Description

Finished APIs for metagraph, graph query, and views. Currently implement a very crappy cross type reducer.

Tasks
- [x] Support both triplet relation syntax and shortcut syntax.
- [x] Change `Bipartite` class to `UnitGraph` and supports the case where there is only one node type.
- [x] Test message passing with cross type reducer
- [x] Support wildcard syntax (e.g, `g['user', :, 'game']`)
- [x] `convert.py`
  - [x] convert from/to networkx
  - [x] convert from/to homograph
- [x] Make it compatible with `DGLGraph`
  - [x] `prop_nodes`
  - [x] `prop_edges`
  - [x] `subgraph`
  - [x] `edge_subgraph`
  - [x] `adjacency_matrix`
  - [x] `incidence_matrix`
  - [x] `filter_nodes`
  - [x] `filter_edges`
  - [x] `__repr__`
- [x] Docstring

## Checklist
- [ ] Code is well-documented

## Things are not sure
Feel that the `G.nodes[id].data['h']` and `G.edges[id].data['h']` syntaxes are quite awkward when to introduce types, especially for `G.edges`:
```python
u = [0, 1]
v = [2, 3]
eid = [5, 10]
G.edges[u, v, 'plays']  # is this allowed?
G.edges[u, v, 'user', 'plays', 'game']  # is this allowed?
G.edges[(u, v), ('user', 'plays', 'game')]  # ?
G.edges[eid, 'user', 'plays', 'game']?
```
It looks strange and makes the implementation quite complicated. My current idea is to deprecate `G.nodes[id].data['h']` because it is the same as `G.ndata['h'][id]`. Then, we could use `G.nodes[ntype].data['h']` and `G.edges[stype, etype, dtype].data['h']`.


jermainewang:
Currently, new message passing APIs are implemented in `multi_recv`, `multi_send_and_recv`, `multi_pull` and `multi_update_all`. Since the arguments are drastically different from original `recv`, `send_and_recv` etc, I found separating them makes it cleaner and easier to develop. Note that there is no `multi_send` and `multi_push` because there is no cross-type reducer in them.

jermainewang:
@BarclayII I implemented two APIs to construct basic graphs.
* `dgl.graph(data, ntype, etype)` creates graph with only one type of nodes and edges.
* `dgl.bipartite(data, utype, etype, vtype)` creates a uni-directional bipartite graph.
The `data` can be edge list, pair of node lists and scipy matrix. Networkx support hasn't been done. In general, I want to have only few (short) APIs that can be easily remembered and quite flexible. Think about `torch.tensor`.

BarclayII:
> ## Things are not sure
> Feel that the `G.nodes[id].data['h']` and `G.edges[id].data['h']` syntaxes are quite awkward when to introduce types, especially for `G.edges`:
> 
> ```python
> u = [0, 1]
> v = [2, 3]
> eid = [5, 10]
> G.edges[u, v, 'plays']  # is this allowed?
> G.edges[u, v, 'user', 'plays', 'game']  # is this allowed?
> G.edges[(u, v), ('user', 'plays', 'game')]  # ?
> G.edges[eid, 'user', 'plays', 'game']?
> ```

Since we allow `G.edges[eid, etype]`, maybe we could first support replacing `eid` with `(u, v)` and `etype` with `(srctype, etype, dsttype)`.  Therefore,

```python
G.edges[u, v, 'plays']  # no -> G.edges[(u, v), 'plays']
G.edges[u, v, 'user', 'plays', 'game']  # no -> G.edges[(u, v), ('user', 'plays', 'game')]
G.edges[(u, v), ('user', 'plays', 'game')]  # yes
G.edges[eid, 'user', 'plays', 'game'] # no -> G.edges[eid, ('user', 'plays', 'game')]
```

> 
> It looks strange and makes the implementation quite complicated. My current idea is to deprecate `G.nodes[id].data['h']` because it is the same as `G.ndata['h'][id]`. Then, we could use `G.nodes[ntype].data['h']` and `G.edges[stype, etype, dtype].data['h']`.

What was the point of having `G.nodes[id].data` in the first place?  I thought the reason was to allow picking some rows of the data (e.g. distributed across multiple machines) without loading the entire data into a single piece of memory (with `G.ndata`).

jermainewang:
> What was the point of having G.nodes[id].data in the first place? I thought the reason was to allow picking some rows of the data (e.g. distributed across multiple machines) without loading the entire data into a single piece of memory (with G.ndata).

I don't think it's for distributed computation since it is likely to be handled by a distributed KV store. Thus, I feel it is redundant and could be recycled for better purposes.

BarclayII:
> > What was the point of having G.nodes[id].data in the first place? I thought the reason was to allow picking some rows of the data (e.g. distributed across multiple machines) without loading the entire data into a single piece of memory (with G.ndata).
> 
> I don't think it's for distributed computation since it is likely to be handled by a distributed KV store. Thus, I feel it is redundant and could be recycled for better purposes.

But the APIs of distributed KVStore are equivalent to `G.nodes[ids].data`, right?  And I feel the latter is more intuitive and can also directly interface with the message passing APIs.

include/dgl/base_heterograph.h
```diff
@@ -381,6 +398,62 @@ struct HeteroSubgraph : public runtime::Object {
   DGL_DECLARE_OBJECT_TYPE_INFO(HeteroSubgraph, runtime::Object);
 };
 
+/*! \brief The flattened heterograph */
+struct FlattenedHeteroGraph : public runtime::Object {
```
jermainewang: My general concern is can we use existing python data structure to realize this? For example, induced_eid and induced_nid can be stored as node/edge features. Induced srctype and dsttype can just be the `ntypes` and `etypes` argument to the `DGLHeterograph` initializer and can be stored together natively with the returned graph.

BarclayII: I thought when we do wildcard slicing, we need to combine the different source/destination/edge types into one source/destination/edge type and return a unitgraph.  So to keep track of the original node/edge type/ID in the heterograph, we cannot reuse the `ntypes` and `etypes` argument since they actually only contain one type.
Regarding to storing `induced_foo` as features, I'm not sure how to decide the column names to store the node/edge type/IDs.  Should we reserve some names?

jermainewang:
@BarclayII Since you've merged FlattendHeteroGraph to just hetero graph, how about HeteroSubgraph? Can we do something similar there?

tests/compute/test_heterograph.py
```diff
@@ -10,151 +9,388 @@
...
+
+        for i, (etype, eid) in enumerate(zip(etypes, eids)):
+            src_g, dst_g = g.find_edges([eid], g.canonical_etypes[etype])
+            src_fg, dst_fg = fg.find_edges([i])
+            # TODO(gq): I feel this code is quite redundant; can we just add new members (like
+            # "induced_srcid") to returned heterograph object and not store them as features?
+            assert src_g == fg.nodes[SRC].data[dgl.NID][src_fg]
+            assert g.canonical_etypes[etype][0] == g.ntypes[fg.nodes[SRC].data[dgl.NTYPE][src_fg]]
+            assert dst_g == fg.nodes[DST].data[dgl.NID][dst_fg]
+            assert g.canonical_etypes[etype][2] == g.ntypes[fg.nodes[DST].data[dgl.NTYPE][dst_fg]]
```
jermainewang: I feel it's pretty ok if you use a variable here:
```python
mapping = fg.nodes[SRC].data[dgl.NID]
assert src_g = mapping[src_fg]
```
To make it better, I think we should try to shorten the `fg.nodes[type].data[fld]`.

BarclayII: No, I mean, I feel the code would look cleaner if the mapping can be accessed with an attribute like `fg.induced_srcid`.

Maybe we could provide a property that acts as a shortcut to the reserved column.

```python
@property
def induced_srcid(self):
    return self.nodes[SRC].data[dgl.NID]
```

BarclayII:
> @BarclayII Since you've merged FlattendHeteroGraph to just hetero graph, how about HeteroSubgraph? Can we do something similar there?

Then we should also do the same for DGLSubgraph?

tests/compute/test_heterograph.py
jermainewang: Comparing the two:
```python
mapping = fg.nodes[SRC].data[dgl.NID]
assert src_g = mapping[src_fg]
```
and
```python
mapping = fg.induced_srcid
assert src_g = mapping[src_fg]
```
I feel the difference is not that huge and the learning effort is similar, if we could beautify the `fg.nodes[SRC].data[dgl.NID]` syntax (for example, `fg.ndata['id', 'src']`). Another option is to have `g.srcdata` and `g.dstdata` and add `mapid` to the data view. This ends up with:
```python
mapping = fg.srcdata.mapid
```
Another question, if the slice returns a homograph, will `induced_srcid` still make sense?

jermainewang:
> > @BarclayII Since you've merged FlattendHeteroGraph to just hetero graph, how about HeteroSubgraph? Can we do something similar there?
> 
> Then we should also do the same for DGLSubgraph?

I think that's the plan but we could first work on heterosubgraph. The idea aligns with the effort to merge XXXSubgraph, BatchedXXXGraph all in one. One idea is to store auxiliary data as features as much as possible. As a result, there will be following changes:
* For XXXSubgraph,
  - `parent_nid` and `parent_eid` APIs will be deprecated. They are equivalent to get the mapping directly from node/edge features.
  - `map_to_subgraph_nid` will be moved to `dgl.map_to_subgraph_nid(parent, subg)`.
  - For `copy_from/to_parent`, I find we usually call these APIs immediately after a subgraph is created, so how about `copy_from_parent` by default? The copy could also be deferred to the first time node/edge features are required, and the write will directly reflect to the parent graph. Is there any use case that such behavior is not desired?
* For BatchedXXXGraph,
  - `batch_size`, `batch_num_nodes` and `batch_num_edges` are all graph-level data. We could (1) return them to the user; (2) Put them in graph-level data frame as in #737 ; or (3) Escalate them as general concepts into `DGLGraph`.

Bring team for the visibility @mufeili @VoVAllen @yzh119 @zheng-da . Also related to this issue #735 .

BarclayII:
> * For XXXSubgraph,
>   
>   * `parent_nid` and `parent_eid` APIs will be deprecated. They are equivalent to get the mapping directly from node/edge features.
>   * `map_to_subgraph_nid` will be moved to `dgl.map_to_subgraph_nid(parent, subg)`.

I'm OK with the two above.

>   * For `copy_from/to_parent`, I find we usually call these APIs immediately after a subgraph is created, so how about `copy_from_parent` by default? The copy could also be deferred to the first time node/edge features are required, and the write will directly reflect to the parent graph. Is there any use case that such behavior is not desired?

Then what if the parent graph modifies the feature?  Would that reflect to the subgraph as well?  If so, then we are essentially supporting `shared` mode.

VoVAllen:
> For BatchedXXXGraph,
batch_size, batch_num_nodes and batch_num_edges are all graph-level data. We could (1) return them to the user; (2) Put them in graph-level data frame as in #737 ; or (3) Escalate them as general concepts into DGLGraph.

I like this idea, and we can reserve name starts with '_' as dgl reserved namespace.

mufeili:
I think we also need `local_var` and `local_scope` for `DGLHeteroGraph`

jermainewang:
> I think we also need `local_var` and `local_scope` for `DGLHeteroGraph`.

There is. https://github.com/jermainewang/dgl/blob/hetero/python/dgl/heterograph.py#L2847

> One question:
> Since most computation can only be done on one etype, what's the difference between `g[etype].update_all()` and `g.update_all(..., etype=...)`. Is there any preference of this?

We don't. They are exactly the same. However, slice is more flexible since it supports merging the results of multiple etypes.

python/dgl/heterograph.py
```diff
@@ -1,373 +1,689 @@
...
+        """
+        if self._nx_metagraph is None:
+            nx_graph = self._graph.metagraph.to_networkx()
+            self._nx_metagraph = nx.MultiDiGraph()
+            for u_v in nx_graph.edges:
+                srctype, etype, dsttype = self.canonical_etypes[nx_graph.edges[u_v]['id']]
+                self._nx_metagraph.add_edge(srctype, dsttype, etype)
+        return self._nx_metagraph
+
+    def to_canonical_etype(self, etype):
```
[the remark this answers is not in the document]
jermainewang: I will add some docstring about this. A canonical etype is a triplet (srctype, etype, dsttype). To simplify the syntax, we support using one etype string whenever it is unambiguous.

python/dgl/heterograph.py
```diff
@@ -1,373 +1,689 @@
...
+    def is_multigraph(self):
+        """True if the graph is a multigraph, False otherwise."""
+        return self._graph.is_multigraph()
+
+    @property
+    def is_readonly(self):
+        """True if the graph is readonly, False otherwise."""
+        return self._graph.is_readonly()
+
+    def has_node(self, vid, ntype=None):
```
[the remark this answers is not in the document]
jermainewang: Currently there is no example using this. Simply inherited from DGLGraph to keep backward compatible.

python/dgl/heterograph_index.py
```diff
@@ -58,10 +61,26 @@ def get_relation_graph(self, etype):
         Returns
         -------
         HeteroGraphIndex
-            The bipartite graph.
+            The unitgraph graph.
         """
         return _CAPI_DGLHeteroGetRelationGraph(self, int(etype))
 
+    def flatten_relations(self, etypes):
+        """Convert the list of requested unitgraph graphs into a single unitgraph
```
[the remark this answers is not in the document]
jermainewang: Unit graph is an _internal_ data structure representing the basic computing block of heterograph. It is either a normal graph or a uni-directional bipartite graph, so it can be represented by one sparse matrix (not necessarily square).

python/dgl/heterograph_index.py
```diff
@@ -839,4 +997,24 @@ def create_heterograph(metagraph, rel_graphs):
...
+        Data to initialize graph structure.
+
+    Returns
+    -------
+    HeteroGraphIndex
+    """
+    if isinstance(graph_data, HeteroGraphIndex):
+        return graph_data
+    elif isinstance(graph_data, DGLGraph):
+        return create_heterograph_from_homo(graph_data)
```
mufeili: Where is this defined/imported?

mufeili: Also for DGLGraph

python/dgl/heterograph.py
```diff
@@ -1,373 +1,689 @@
...
-            return etype_indices
-        elif self.is_edge_type_view:
-            return [self._view_etype_idx]
-        else:
-            return range(len(self._etypes))
+    One can construct the graph as follows:
 
-    def edge_types(self):
-        """Return the list of edge types appearing in the current view.
+    >>> mg = nx.MultiDiGraph([('user', 'user', 'follows'),
```
mufeili: The examples for initializing `DGLHeteroGraph` are now outdated.

jermainewang: I'm working on the docstring. Will update it later.

mufeili:
Got an error with the code below:

```python
import dgl
import numpy as np
from scipy import sparse

A = (np.random.rand(10, 10) > 0.25).astype('int8')
A = sparse.csr_matrix(A)
g = dgl.graph(A, ntype='A', etype='B')
```

python/dgl/convert.py
```diff
@@ -0,0 +1,577 @@
...
+            num_ntypes, num_src, num_dst, row, col)
+    else:
+        spmat = spmat.tocsr()
+        indptr = utils.toindex(spmat.indptr)
+        indices = utils.toindex(spmat.indices)
+        # TODO(minjie): with_edge_id is only reasonable for csr matrix. How to fix?
+        data = utils.toindex(spmat.data if with_edge_id else list(range(len(indices))))
+        hgidx = heterograph_index.create_unitgraph_from_csr(
+            num_ntypes, num_src, num_dst, indptr, indices, data)
+    return DGLHeteroGraph(hgidx, [utype, vtype], [etype])
```
mufeili: Handle the case where `utype == vtype`.

python/dgl/convert.py
```diff
@@ -0,0 +1,577 @@
...
+    """
+    if card is not None:
+        urange, vrange = card, card
+    else:
+        urange, vrange = None, None
+    if isinstance(data, tuple):
+        u, v = data
+        return create_from_edges(u, v, ntype, etype, ntype, urange, vrange)
+    elif isinstance(data, list):
+        return create_from_edge_list(data, ntype, etype, ntype, urange, vrange)
```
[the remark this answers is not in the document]
jermainewang: One solution is to only support (u, v) in `dgl.graph` while have an another function to convert edge pairs to two arrays. However, please notice that checking tuple and list is used in a lot of places in DGL already. For example,  the argument to `g.edges[[0, 1, 2], [2, 3, 4]]` is a tuple while the argument to `g.edges[[0, 1, 2]` is a list, and we treat the previous one as (u, v) while the later one as eid. I didn't see too much complain about this.

python/dgl/heterograph_index.py
jermainewang: I removed this. Currently, a heterograph is created via functions in `convert.py`. The class initializer is not directly exposed so the first argument right now must be HeteroGraphIndex.

python/dgl/convert.py
jermainewang: Done

python/dgl/convert.py
```diff
@@ -0,0 +1,577 @@
...
+    elif isinstance(data, sp.sparse.spmatrix):
+        return create_from_scipy(data, ntype, etype, ntype)
+    elif isinstance(data, nx.Graph):
+        return create_from_networkx(data, ntype, etype, **kwargs)
+    else:
+        raise DGLError('Unsupported graph data type:', type(data))
+
+def bipartite(data, utype='_U', etype='_E', vtype='_V', card=None, **kwargs):
```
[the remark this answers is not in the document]
jermainewang: Examples updated in the docstring.

python/dgl/convert.py
```diff
@@ -0,0 +1,577 @@
...
+__all__ = [
+    'graph',
+    'bipartite',
+    'hetero_from_relations',
+    'hetero_from_homo',
+    'hetero_to_homo',
+    'to_networkx',
+]
+
+def graph(data, ntype='_N', etype='_E', card=None, **kwargs):
```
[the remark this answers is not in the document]
jermainewang: Similar to `torch.tensor` or `numpy.array`, I feel `dgl.graph` is quite intuitive.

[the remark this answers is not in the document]
VoVAllen: I see. `dgl.graph` is better

python/dgl/convert.py
```diff
@@ -0,0 +1,694 @@
...
+        dtid = ntype_dict[dtype]
+        meta_edges.append((stid, dtid))
+        etypes.append(etype)
+    metagraph = graph_index.from_edge_list(meta_edges, True, True)
+    # create graph index
+    hgidx = heterograph_index.create_heterograph_from_relations(
+        metagraph, [rgrh._graph for rgrh in rel_graphs])
+    return DGLHeteroGraph(hgidx, ntypes, etypes)
+
+def hetero_from_homo(G, ntypes, etypes, ntype_field='type', etype_field='type'):
```
mufeili: We should be consistent with `hetero_from_homo` and set the default values of `ntype_field` and `etype_field` to be `NTYPE` and `ETYPE`.

jermainewang: Done

python/dgl/convert.py
```diff
@@ -0,0 +1,722 @@
...
+    node_invmap = np.zeros((G.number_of_nodes(),), dtype=src.dtype) - 1
+    edge_invmap = np.zeros((G.number_of_edges(),), dtype=src.dtype) - 1
+    for grp in node_groups:
+        _make_invmap(grp, node_invmap)
+    for grp in edge_groups:
+        _make_invmap(grp, edge_invmap)
+
+    # infer metagraph
+    canonical_etids = []
+    for etid in range(num_etypes):
```
BarclayII: So are we not allowing the users to have the same edge type ID if the source-destination type pairs differ?  I think we should allow this if we allow such edge types to have the same name.

jermainewang: Same name is allowed. For example, there could be two `"watches"` in the `etypes` argument. However, I feel the IDs should be different. Otherwise, `len(G.etypes)` might not be equal to `len(G.canonical_etypes)`, which is quite strange to me.

BarclayII: No, I mean we need to detect the same edge IDs with different source-destination types, and translate them to different canonical etypes, just like what we did for edge type names.

BarclayII: On the other hand, I don't think we need to maintain the order of edge types in the heterograph to be the same as that in  the `etypes` argument, because after the users called `to_hetero()`, we can safely assume that from then on the users would only access the features and do MP with the edge type names instead of IDs.

jermainewang: For the user provided edge type ID, the question is whether the edge ID is similar to edge type name that allows collision. Current implementation is NO. That is to say the provided edge ID one-one corresponds to the canonical edge type. Your proposal is YES. I think both proposals are OK. Hence the question is how easy (efficient) will the implementation be? You could see that the current implementation avoids looping over all the nodes/edges in python and only loops over node/edge types which is a nice property I want to keep. If edge ID can have collision, is this still achievable? I feel inferring the metagraph will be quite painful (i.e, need to compute some complicated unique operation).

jermainewang: For the edge type order, I think keeping the order makes it easier to jump from IDs to actual names. If we could keep it without too much of an effort, let's try to have that.

BarclayII: > Hence the question is how easy (efficient) will the implementation be?

I don't think sacrificing usability for efficiency is a good rationale in particular for this case.

Essentially, we would be leaving the de-duplication of edge types to functions at higher levels.  For instance, converting from NetworkX/Neptune/Neo4j/... graphs with edge types as strings would lead us into a similar problem.  How would we handle those cases?  If we were to figure out the different canonical etypes for the same etype in those routines, then why not handle them in `to_hetero()`, so that the conversion from NetworkX/Neptune/etc only involves (1) transform the graph into a homograph with features, and (2) call `to_hetero()`?

jermainewang: Ok. I'm convinced that such functionality might be inevitable. However, we should avoid looping over graphs in python as it will be too slow for even moderate size graph. At the moment, I cannot think of a good way to implement this using numpy.

For the discussion of user experience, I want to use an example to check we are on the same page. Suppose a graph has two relations: `user-watches-TV` `user-watches-movie`.
(1) What `G.etypes` should be?
(2) What `G.canonical_etypes` should be?
(3) What is the `etype` argument to `dgl.to_hetero`?
(4) Given the above `etype`, if there are two edges (one for each type), what the type ID should be?

For the current implementation,
(1) `['watches', 'watches']`
(2) `[('user', 'watches', 'TV'), ('user', 'watches', 'movie')]`
(3) `['watches', 'watches']`
(4) `[0, 1]`

BarclayII: Let's say we have the following heterograph:

```
TV #0 <--(watches)-- User #0 --(watches)--> Movie #0 
```

Then I would expect

```python
g = dgl.graph([(1, 0), (1, 2)])
g.ndata[NTYPE] = torch.LongTensor([0, 1, 2])
# Here I would expect 0 to be just an index to the `etypes` argument.
g.edata[ETYPE] = torch.LongTensor([0, 0])
hg = dgl.to_hetero(g, ntypes=['TV', 'user', 'movie'], etypes=['watches'])
assert set(hg.ntypes) == {'user', 'TV', 'movie'}
assert set(hg.canonical_etypes) == {
        ('user', 'watches', 'TV'), ('user', 'watches', 'movie')}
```

BarclayII: For the purpose of efficiency, I think we will eventually need to migrate this into C++.  We can do that in 0.5.

jermainewang: It's better be done this release. This is a very common utility and I'm afraid that the efficiency may make this API not usable.

PR #836, opened 2019-09-06
[Feature] add PBG's negative edge sampling.

zheng-da:
## Description
This implements PBG-like negative sampler. It first groups positive edges into chunks and corrupt all edges in a chunk together. For example, to corrupt the head nodes of edges in a chunk, it uniformly samples `c` nodes to create `c*c` negative edges, where `c` is the chunk size. This is a simplified version of PBG negative sampling. When experimenting on FB15k, it actually works better than the original PBG negative sampler.

## Checklist
- [ ] Code is well-documented


src/graph/sampler.cc
```diff
@@ -997,6 +998,124 @@ Subgraph NegEdgeSubgraph(int64_t num_tot_nodes, const Subgraph &pos_subg,
...
+  const dgl_id_t *induced_eid_data = static_cast<const dgl_id_t *>(pos_subg.induced_edges->data);
+  size_t num_pos_nodes = pos_subg.graph->NumVertices();
+  std::vector<size_t> pos_nodes(induced_vid_data, induced_vid_data + num_pos_nodes);
+
+  dgl_id_t *neg_dst_data = static_cast<dgl_id_t *>(neg_dst->data);
+  dgl_id_t *neg_src_data = static_cast<dgl_id_t *>(neg_src->data);
+  dgl_id_t *neg_eid_data = static_cast<dgl_id_t *>(neg_eid->data);
+  dgl_id_t *induced_neg_eid_data = static_cast<dgl_id_t *>(induced_neg_eid->data);
+
+  bool neg_head = (neg_mode == "head");
```
classicsong: Can it be a global MACRO to handle sample mode?

src/graph/sampler.cc
```diff
@@ -997,6 +998,124 @@ Subgraph NegEdgeSubgraph(int64_t num_tot_nodes, const Subgraph &pos_subg,
...
+  std::vector<size_t> neg_vids;
+  RandomSample(num_tot_nodes,
+               num_chunks * neg_sample_size,
+               &neg_vids);
+
+  dgl_id_t curr_eid = 0;
+  std::unordered_map<dgl_id_t, dgl_id_t> neg_map;
+  for (int64_t i_chunk = 0; i_chunk < num_chunks; i_chunk++) {
+    // for each chunk.
+    size_t neg_idx = neg_sample_size * chunk_size * i_chunk;
```
classicsong: size_t and int64_t may not be compatible

src/graph/sampler.cc
```diff
@@ -1042,10 +1161,16 @@ DGL_REGISTER_GLOBAL("sampling._CAPI_UniformEdgeSampling")
...
       positive_subgs[i] = ConvertRef(subg);
-      if (neg_mode.size() > 0) {
-        Subgraph neg_subg = NegEdgeSubgraph(gptr->NumVertices(), subg,
-                                            neg_mode, neg_sample_size,
-                                            gptr->IsMultigraph(), exclude_positive);
+      // For PBG negative sampling, we accept "PBG-head" for corrupting head
+      // nodes and "PBG-tail" for corrupting tail nodes.
+      if (neg_mode.substr(0, 3) == "PBG") {
+        Subgraph neg_subg = PBGNegEdgeSubgraph(gptr->NumVertices(), subg,
+                                               neg_mode.substr(4), neg_sample_size,
```
classicsong: In the PGB paper, the sampler do both head corrupt and tail corrupt together.
Here it seems the python code should call the PBGhead and PBGtail

zheng-da: There are two modes "PBG-head" and "PBG-tail". A user needs to call this method twice to corrupt head nodes and tail nodes. This is what we do for regular negative sampler.

src/graph/sampler.cc
zheng-da: It should be fine, but I'll make the type consistent.

src/graph/sampler.cc
zheng-da: what do you mean by "global macro"?

classicsong: like:
```
bool neg_head = IS_PBG_HEAD_MODE(neg_mode)

#def IS_PBG_HEAD_MODE(neg_mode)  \
    (neg_mode == "head")
```

This can avoid possible inconsistency if we use bool neg_head = (neg_mode == "head"); in multiple code pieces.

Issue #664, opened 2019-06-17
[Feature Request] Add missing functions in numpy backend.


zengxy:
## 🚀 Feature
There are some missing or  incompatible functions compared with the interface in numpy backend.
The list is following.

#### Missing
+ [ ] device_type
+ [ ] mean
+ [ ] unsorted_1d_segment_sum
+ [ ] unsorted_1d_segment_mean
+ [ ] zerocopy_to_dgl_ndarray
+ [ ] logical_not
+ [ ] zerocopy_to_dlpack
+ [ ] equal
+ [ ] ndim
+ [ ] sync
+ [ ] zerocopy_from_dlpack
+ [ ] copy_reduce
+ [ ] zerocopy_from_dgl_ndarray
+ [ ] boolean_mask
+ [ ] narrow_row
+ [ ] stack
+ [ ] binary_reduce
+ [ ] device_id
+ [ ] zeros_like

#### Not compatible
+ [ ] zeros `(shape, dtype)` v.s. `(shape, dtype, ctx)`
+ [ ] ones `(shape, dtype)`v.s. `(shape, dtype, ctx)`
+ [ ] full_1d `(length, fill_value)` v.s. `(length, fill_value, dtype, ctx)`

## Motivation
In some scenarios, we just want to use message propogation in dgl (e.g. LPA, PageRank) and thus choose numpy as backend.
However due to the missing or incompatible functions in numpy backend, even the simplest example `send(message_func=dgl.function.copy_src("f", "m"))` will raise an exception.

Do you think it's necessary to completing the missing functions? And if so I can help to add some within my ability.

yzh119:
It seems even in the scenarios you mentioned, uses can still use PyTorch as backend, except for the case where some operations in numpy are not supported by PyTorch.
@jermainewang what do you think about it?

zheng-da:
We have plans to add the MXNet Numpy backend, which provides compatible Numpy APIs and support backward computation. Once MXNet Numpy is ready, we'll add it to DGL. Please see the progress of MXNet Numpy: https://github.com/apache/incubator-mxnet/issues/14327

In addition, a few more operators are required by DGL.
- [ ] unique
- [ ] nonzero
- [ ] boolean_mask
- [ ] index_copy
- [ ] segment_sum

zengxy:
Thanks for reply !

[...]

PR #841, opened 2019-09-08
[bugfix] pickle numpy.array in Scheme

seanliu96:
[bugfix] pickle numpy.array in Scheme (issue 840)

### DIFF
--- a/python/dgl/frame.py
+++ b/python/dgl/frame.py
@@ -30,7 +30,7 @@ class Scheme(namedtuple('Scheme', ['shape', 'dtype'])):
         try:
             state = (self.shape, F.reverse_data_type_dict[self.dtype])
             return self._reconstruct_scheme, state
-        except KeyError as e:
+        except KeyError:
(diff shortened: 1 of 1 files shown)


yzh119:
@seanliu96 thanks for your contribution.
I have checked your code and found out the problem is not in `F.reverse_data_type_dict`, if you use  numpy as backend, you must specify environment variable `DGLBACKEND=numpy`(by default it's set to `pytorch`, so dgl is unable to find the value correspond to `np.float32`, you will met the same problem for mxnet if you do not change the variable).

However, numpy backend is not fully supported (we lack the interface from numpy ndarray to dlpack and vice versa.).

seanliu96:
@yzh119 Yes, I find it is not fully supported when I set dgl.load_backend("numpy"). mxnet should not have this problem because keys of its reverse_data_type_dict start with "np."

jermainewang:
Related to #664 . Complete numpy support will not be done in the near future due to the bandwidth limit.

PR #861, opened 2019-09-19
[Sampler] Metapath sampler for metapath2vec

BarclayII:
## Description
Provides the (cyclic) metapath-based random walk sampler required by metapath2vec.

This implementation takes 857 seconds to generate a corpus of the same size as that in Metapath2vec (which takes 1563 seconds) on AMiner.

## Checklist
- [x] Code is well-documented


python/dgl/contrib/sampling/randomwalk.py
```diff
@@ -1,13 +1,20 @@
...
+from ... import ndarray
 
 __all__ = ['random_walk',
            'random_walk_with_restart',
            'bipartite_single_sided_random_walk_with_restart',
+           'metapath_cycle_random_walk',
            ]
 
+@register_object('sampler.RandomWalkTraces')
+class RandomWalkTraces(ObjectBase):
```
jermainewang: Briefly document the available attributes and types.

python/dgl/contrib/sampling/randomwalk.py
```diff
@@ -165,4 +169,52 @@ def bipartite_single_sided_random_walk_with_restart(
             int(max_visit_counts), int(max_frequent_visited_nodes))
     return _split_traces(traces)
 
-_init_api('dgl.randomwalk', __name__)
+
+def metapath_cycle_random_walk(hg, etypes, seeds, num_traces, max_cycles):
```
jermainewang: Could we let user directly give a full metapath sequence so there is no "cycle" behavior in this API? User could directly give `[A,B,A,B,A,B]` and we would follow the metapath but could stop earlier. Also, change `etypes` argument to `metapath`.

python/dgl/contrib/sampling/randomwalk.py
```diff
@@ -165,4 +169,52 @@ def bipartite_single_sided_random_walk_with_restart(
...
+    Returns
+    -------
+    traces : list[list[Tensor]]
+        traces[i][j] is the j-th trace generated for i-th seed.
+        traces[i][j][k] would have node type the same as the destination node type of edge
+        type ``etypes[k % len(etypes)]``
+
+    Notes
+    -----
+    The traces does **not** include the seed nodes themselves.
```
jermainewang: Is this a convention?

python/dgl/convert.py
```diff
@@ -264,7 +264,7 @@ def hetero_from_relations(rel_graphs):
         retg._edge_frames[i].update(rgrh._edge_frames[0])
     return retg
 
-def to_hetero(G, ntypes, etypes, ntype_field=NTYPE, etype_field=ETYPE, metagraph=None):
+def to_hetero(G, ntypes, etypes, ntype_field=NTYPE, etype_field=ETYPE):
```
jermainewang: Why revert these back?

src/graph/sampler/metapath.cc
```diff
@@ -0,0 +1,102 @@
...
+  const auto metagraph = hg->meta_graph();
+  uint64_t num_etypes = etypes->shape[0];
+  uint64_t num_seeds = seeds->shape[0];
+  const dgl_type_t *etype_data = static_cast<dgl_type_t *>(etypes->data);
+  const dgl_id_t *seed_data = static_cast<dgl_id_t *>(seeds->data);
+
+  std::vector<dgl_id_t> vertices;
+  std::vector<size_t> trace_lengths, trace_counts;
+
+  for (uint64_t seed_id = 0; seed_id < num_seeds; ++seed_id) {
```
jermainewang: Consider using omp to embarrassingly parallelize this loop? If so, you might need to trigger `SuccVec` first to materialize the lazy component before entering the critical area. You might also need to split the `vertices` vector into ones for each seed.

python/dgl/contrib/sampling/randomwalk.py
BarclayII: Yes

src/graph/sampler/metapath.cc
BarclayII: Sounds not very trivial to implement...  Can I postpone OpenMP parallelization to 0.5?

python/dgl/contrib/sampling/randomwalk.py
BarclayII: Good point.

src/graph/sampler/metapath.cc
jermainewang: sure. added a TODO note there

PR #862, opened 2019-09-19
[Transform] Add to_self_loop, remove_self_loop, onehot_degree transform

VoVAllen:
## Description
Add transform operation

## Checklist
- [ ] Code is well-documented


python/dgl/backend/backend.py
```diff
@@ -1177,6 +1177,27 @@ def zerocopy_from_dgl_ndarray(input):
     """
     pass
 
+
+def one_hot(t, num_classes):
+    """
+    Conver tensor to one-hot tensor
```
mufeili: "Conver" should be "Convert".

python/dgl/backend/backend.py
```diff
@@ -1177,6 +1177,27 @@ def zerocopy_from_dgl_ndarray(input):
     """
     pass
 
+
+def one_hot(t, num_classes):
```
mufeili: Set the default value of `num_classes` to be -1?

python/dgl/backend/mxnet/tensor.py
```diff
@@ -358,6 +358,10 @@ def zerocopy_to_dgl_ndarray_for_write(arr):
 def zerocopy_from_dgl_ndarray(arr):
     return nd.from_dlpack(arr.to_dlpack())
 
+def one_hot(t, num_classes):
```
mufeili: Same

python/dgl/backend/numpy/tensor.py
```diff
@@ -203,3 +203,11 @@ def zerocopy_to_numpy(input):
 
 def zerocopy_from_numpy(np_array):
     return np_array
+
+def one_hot(t, num_classes):
```
mufeili: Same

python/dgl/backend/pytorch/tensor.py
```diff
@@ -281,6 +281,9 @@ def zerocopy_to_dgl_ndarray(input):
 def zerocopy_from_dgl_ndarray(input):
     return dlpack.from_dlpack(input.to_dlpack())
 
+def one_hot(t, num_classes):
```
mufeili: Same

python/dgl/transform.py
```diff
@@ -403,4 +404,84 @@ def laplacian_lambda_max(g):
...
     return rst
 
+
+def onehot_degree(g, max_degree=-1, out_field='d', direction="in"):
+    """Inplace add one-hot degree vector as node feature
+
+    Parameters
+    -----------
+    g: DGLGraph
+    max_degress: int
```
mufeili: Typo

python/dgl/transform.py
```diff
@@ -403,4 +404,84 @@ def laplacian_lambda_max(g):
...
+        Maximum degree for one-hot encoding. If it's -1,
+        the maximum degree would be infered from the input graph.
```
mufeili: Should be "inferred"

python/dgl/transform.py
```diff
@@ -403,4 +404,84 @@ def laplacian_lambda_max(g):
...
+    if direction == "in":
+        degrees = g.in_degrees()
+    elif direction == "out":
+        degrees = g.out_degrees()
+    else:
+        raise RuntimeError("Invalid Direction")
+    g.ndata[out_field] = F.one_hot(degrees, max_degree)
+
+def to_self_loop(g):
+    """Make graph contains exactly one self loop for each node.
```
mufeili: Should just be "contain"

python/dgl/transform.py
```diff
@@ -403,4 +404,84 @@ def laplacian_lambda_max(g):
...
+    if g.is_readonly:
+        g.readonly(False)
+        g.add_edges(remain_nodes_wo_selfloop, remain_nodes_wo_selfloop)
+        g.readonly()
+    else:
+        g.add_edges(remain_nodes_wo_selfloop, remain_nodes_wo_selfloop)
+
+
+def remove_self_loop(g):
+    """Make graph contains exactly one self loop.
```
mufeili: Change the comment.

python/dgl/transform.py
```diff
@@ -403,4 +404,84 @@ def laplacian_lambda_max(g):
...
+    """
+    if direction == "in":
+        degrees = g.in_degrees()
+    elif direction == "out":
+        degrees = g.out_degrees()
+    else:
+        raise RuntimeError("Invalid Direction")
+    g.ndata[out_field] = F.one_hot(degrees, max_degree)
+
+def to_self_loop(g):
```
mufeili: The operation seems to be in-place. Previously all transforms related to graph topology are not in-place.

python/dgl/transform.py
```diff
@@ -403,4 +404,84 @@ def laplacian_lambda_max(g):
...
+    remain_nodes_wo_selfloop = np.where(has_self_edges == 0)[0]
+    if g.is_readonly:
+        g.readonly(False)
+        g.add_edges(remain_nodes_wo_selfloop, remain_nodes_wo_selfloop)
+        g.readonly()
+    else:
+        g.add_edges(remain_nodes_wo_selfloop, remain_nodes_wo_selfloop)
+
+
+def remove_self_loop(g):
```
mufeili: The operation seems to be in-place. Previously all transforms related to graph topology are not in-place.

VoVAllen:
@mufeili I changed them to non-inplace operation. Could you review again?

python/dgl/transform.py
```diff
@@ -403,4 +404,97 @@ def laplacian_lambda_max(g):
...
+        degrees = g.in_degrees()
+    elif direction == "out":
+        degrees = g.out_degrees()
+    else:
+        raise RuntimeError("Invalid Direction")
+    g.ndata[out_field] = F.one_hot(degrees, max_degree)
+    return g
+
+def to_self_loop(g):
+    """Return a new graph contains exactly one self loop for each node.
```
mufeili: "contains" -> "containing"

jermainewang:
Sorry I haven't got time to write my comment and just saw this PR merged. Several considerations:

* Do we really need `onehot_degree`? One quite popular option is to use identity matrix for featureless graph. Since the graph could be quite large, one might want to use sparse matrix for this. Overall, I feel the API is not general enough (and perhaps hard to be) and is also quite straightforward for users to write. I suggest put it as a utility function in the corresponding dataset preprocessing codes (e.g. TUDataset) so it is not exposed.
* Change `to_self_loop` to `add_self_loop`. Consider the following docstring:
  ```python
  """Return a new graph containing all the edges in the input graph plus self loops of every nodes.

  No duplicate self loop will be added for nodes already having self loops. Edge IDs are not preserved.

  Examples
  -------------
  blahblah
  """
  ```

VoVAllen:
@jermainewang Onehot_degree is intended to be exposed to user. I prefer all datasets to be raw dataset. Therefore adding features to graph without feature would be done by user instead of dataset classes. If you think it's not fine to put it under `transform` namespace, I can move it to something like `data.utils`.

jermainewang:
My thought process about `onehot_degree`. It would be better to return the tensor instead of setting it to graph features to avoid name clash. This makes the whole function quite straightforward: only checking whether the argument is "out" or "in" and then applying a onehot function. This makes me feel the function does not save user any effort. On the other hand, the API is also not general enough. For example, we might want to return a sparse matrix if the degree distribution is power-law. Overall, if it is for preprocessing TUDataset, then it should be there.

PR #868, opened 2019-09-19
[Model][Hetero] HAN

mufeili:
@BarclayII 

## Checklist
- [ ] Code is well-documented


BarclayII:
We provide two versions of HAN:

1. One using the author's preprocessed dataset; no heterogeneous API is involved.
2. One using our own dataset operating on a heterogeneous graph, with `coalesce_metapath()` function that coalesces paths matching the given metapaths into edges.  The resulting dataset (number of papers, relation between the papers, their features and labels, etc.) is different from the one from the authors.

BarclayII:
@jermainewang @zheng-da Please check if doing two experiments (one without heterogeneous API on their dataset, another with heterogeneous API on our own dataset) is acceptable.  I feel that this is a reasonable compromise before the authors release their preprocessing code.
@yzh119 If you would like to have HAN models as `dgl.nn` modules please let us know.

python/dgl/transform.py
```diff
@@ -404,6 +405,56 @@ def laplacian_lambda_max(g):
                                       return_eigenvectors=False)[0].real)
     return rst
 
+def coalesce_metapath(g, metapath):
+    """Transform a heterogeneous graph into either a homogeneous graph or a unidirectional
+    bipartite graph so that metapaths are coalesced into edges.
```
jermainewang: How about `metapath_reacheable_graph`? Also, for the docstring, I won't mention homogeneous or bipartite in the first sentence. How about "Return a graph where the neighbors of a node are the nodes reachable by the given metapath in the input graph"?

python/dgl/transform.py
```diff
@@ -404,6 +405,56 @@ def laplacian_lambda_max(g):
...
+    If the beginning node type ``s`` and ending node type ``t`` are the same, it will return
+    a homogeneous graph with node type ``s = t``.  Otherwise, a unidirectional bipartite graph
+    with source node type ``s`` and destination node type ``t`` is returned.
+
+    In both cases, two nodes ``u`` and ``v`` will be connected with an edge ``(u, v)`` if
+    there exists one path matching the metapath from ``u`` to ``v``.
+
+    Even if a node of type ``s`` does not have any outbound path matching the given metapath, it
+    will still show up in the returned graph.  Similarly, a node of type ``t`` will still
+    show up even if it does not have any inbound path matching the given metapath.
```
jermainewang: The result graph keeps the node set of type ``s`` and ``t`` in the original graph even if they might have no neighbor.

examples/pytorch/han/model_hetero.py
```diff
@@ -0,0 +1,78 @@
...
+            nn.Linear(hidden_size, 1, bias=False)
+        )
+
+    def forward(self, z):
+        w = self.project(z)
+        beta = torch.softmax(w, dim=1)
+
+        return (beta * z).sum(1)
+
+class HANLayer(nn.Module):
```
jermainewang: Some brief explanations of the parameters are appreciated. The same for the `forward` function.

python/dgl/transform.py
BarclayII: Much better...

PR #877, opened 2019-09-23
[Model] Add Pytorch example for Cluster GCN

Zardinality:
## Description

Forum discussion link: [click me](https://discuss.dgl.ai/t/model-add-pytorch-example-for-cluster-gcn/442)

## Checklist
- [ ] Code is well-documented

## Changes

- [x] Graph cluster sampler
- [x] Model itself(including GCN and mean GraphSAGE)
- [ ] Speed of partitioning and adj list extraction
- [ ] Sota deep GCN on PPI dataset
- [ ] Cluster evaluate


examples/pytorch/cluster_gcn/run_ppi.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+
+python cluster_gcn.py --gpu 0 --dataset ppi --lr 1e-2 --weight-decay 0.0 --psize 50 --batch-size 2 --n-epochs 300 --n-hidden 1024 --n-layers 1 --log-every 100 --use-pp --self-loop --note self-loop-ppi-non-sym-ly3-pp-cluster-2-2-wd-5e-4 --dropout 0.0 --model-type graphsage --use-val --normalize
```
futurely: What are the experimental results?

https://github.com/google-research/google-research/tree/master/cluster_gcn#run-experiments

Zardinality: My bad that I forgot to mention. The original repo replicates the results of the paper, which is around 0.994. However, it uses some tricks about enhancing adj diagonal, to retain locality when stacking deep conv net. It might not be worth it to implement it for Reddit since it would not be sota anyway, but make sense to see how it works on PPI. I was not focusing on the performance but rather the speed before, but now I will add a milestone on this. Though I would not have time to work on this before this weekend.

examples/pytorch/cluster_gcn/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+Dependencies
+------------
+- Python 3.7+(for string formatting features)
+- PyTorch 1.1.0+
+- metis
+- sklearn
+
+
+* install clustering toolkit: metis and its Python interface.
```
mufeili: It seems that `metis` has a [pip wheel](https://pypi.org/project/metis/), which shall make the life of people easier for installation.

examples/pytorch/cluster_gcn/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+  METIS - Serial Graph Partitioning and Fill-reducing Matrix Ordering ([official website](http://glaros.dtc.umn.edu/gkhome/metis/metis/overview))
+
+```
+1) Download metis-5.1.0.tar.gz from http://glaros.dtc.umn.edu/gkhome/metis/metis/download and unpack it
+2) cd metis-5.1.0
+3) make config shared=1 prefix=~/.local/
+4) make install
+5) export METIS_DLL=~/.local/lib/libmetis.so
+```
```
mufeili: I think you forgot 6) `pip install metis`.

examples/pytorch/cluster_gcn/cluster_gcn.py
```diff
@@ -0,0 +1,247 @@
...
+            batch_labels = cluster.ndata['labels']
+            batch_train_mask = cluster.ndata['train_mask']
+            loss = loss_f(pred[batch_train_mask],
+                          batch_labels[batch_train_mask])
+
+            optimizer.zero_grad()
+            loss.backward()
+            optimizer.step()
+            if j % args.log_every == 0:
+                print(f"epoch:{epoch}/{args.n_epochs}, Iteration {j}/{len(cluster_iterator)}:training loss", loss.item())
```
mufeili: Seems I'am always getting "Iteration 0/50" when using PPI.

examples/pytorch/cluster_gcn/README.md
Zardinality: Sure. Will add it.

examples/pytorch/cluster_gcn/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    return li
+
+
+def save_log_dir(args):
+    log_dir = './log/{}/{}'.format(args.dataset, args.note)
+    os.makedirs(log_dir, exist_ok=True)
+    return log_dir
+
+
+def get_sqrt_norm(g):
```
mufeili: Not used anywhere.

examples/pytorch/cluster_gcn/cluster_gcn.py
Zardinality: Because I set the `log_every` to be a `100`, which is bigger than ppi's partition size per epoch. If `log_every` is set to be `10`, then every epoch it will print log 5 times. I think it could be useful if one wants to train on a large graph for a few epochs.

mufeili:
Some imports are not used, including:

1) `cluster_gcn.py`

```python
import math

import dgl
import dgl.function as fn

import scipy.sparse as sp
```

2) `modules.py`

```python
import dgl

import numpy as np

from functools import partial
```

3) `partition_utils.py`

```python
import json
import os
import pickle as pkl
import random
import sys

import networkx as nx

import scipy.sparse as sp
from networkx.readwrite import json_graph
```

4) `sampler.py`

```python
import os
import random

import numpy as np

import dgl
```

5) `utils.py`

```python
import json

import time

import scipy.sparse as sp
import sklearn.metrics
import sklearn.preprocessing

import torch.nn as nn

from networkx.readwrite import json_graph
```

Zardinality:
> Some minor comments are made, the rest look good to me.

Will address them ASAP. Thanks for the comments.

examples/pytorch/cluster_gcn/cluster_gcn.py
mufeili: I C. Then maybe add a comment there so people can check the comment and understand what's going on.

Zardinality:
Now the PPI train script should give `0.9924` as F1-mic, which is very close to the reported `0.9940` in the original repo, though I don't impl the `diag-enhance` trick. Would be good if someone can confirm my result replicates.

examples/pytorch/cluster_gcn/run_ppi.sh
Zardinality: New PPI results appended

PR #881, opened 2019-09-25
Improve edge sampler

zheng-da:
## Description
Two small improvements:
* This adds a flag to disable the calculation of false negative edges.
* Improve the sampling when we want to sample negative edges from all nodes.

## Checklist
- [ ] Code is well-documented


python/dgl/contrib/sampling/sampler.py
```diff
@@ -498,7 +498,8 @@ def __init__(
             negative_mode="",
             neg_sample_size=0,
             exclude_positive=False,
-            relations=None):
+            relations=None,
+            check_false_neg=False):
```
jermainewang: docstring?

src/graph/sampler.cc
```diff
@@ -988,40 +1007,59 @@ NegSubgraph NegEdgeSubgraph(GraphPtr gptr, IdArray relations, const Subgraph &po
   dgl_id_t *neg_eid_data = static_cast<dgl_id_t *>(neg_eid->data);
   dgl_id_t *induced_neg_eid_data = static_cast<dgl_id_t *>(induced_neg_eid->data);
 
+  const dgl_id_t *unchanged;
+  dgl_id_t *neg_unchanged;
+  dgl_id_t *neg_changed;
+  if (is_neg_head_mode(neg_mode)) {
```
jermainewang: `is_neg_head_mode` -> `IsNegativeHeadMode`

src/graph/sampler.cc
```diff
@@ -988,40 +1007,59 @@ NegSubgraph NegEdgeSubgraph(GraphPtr gptr, IdArray relations, const Subgraph &po
...
         dgl_id_t global_vid = *it;
         exclude.push_back(global_vid);
       }
+      neg_vids.clear();
       RandomSample(num_tot_nodes, neg_sample_size, exclude, &neg_vids);
-    } else {
+    } else if (neg_sample_size < num_tot_nodes) {
+      neg_vids.clear();
       RandomSample(num_tot_nodes, neg_sample_size, &neg_vids);
+    } else if (exclude_positive) {
```
jermainewang: If it possible that all the above conditions are false?

jermainewang: I see it's impossible. Then just use `else`

zheng-da: it is possible. Sometimes, a user wants to sample negative edges of the same number as the number of nodes in a graph. at the same time, they want to turn on `exclude_positive`.

jermainewang: Then what happens if both `exclude_positive` and `neg_sample_size<num_tot_nodes` are false? Do nothing? If that's the case, I would suggest have an extra else block to make it clearer.
```c++
} else {
  // do nothing (with little explanation)
}
```

jermainewang: Oh, another question. You added `neg_vids.clear()` in some branches. Is it correct? The next statement is `RandomSample(..., &neg_vids)`, which I assume it adds some vids to `neg_vids`. Wouldn't it be cleared in the next iteration?

jermainewang: Oh, never mind. Saw the `neg_vids.clear()` in the original code. It looks like you want to reuse the buffer. Take back my last comment.

zheng-da: OK. I'll add an else block and add comment.

PR #882, opened 2019-09-26
[Hetero] Create heterograph from dictionary of edge lists

BarclayII:
## Description
A more intuitive wrapper for `hetero_from_relations`.

With this function we no longer need to explain how a heterograph can be a union of homographs and bipartites.

## Checklist
- [x] Code is well-documented


python/dgl/convert.py
```diff
@@ -264,6 +265,68 @@ def hetero_from_relations(rel_graphs):
         retg._edge_frames[i].update(rgrh._edge_frames[0])
     return retg
 
+def heterograph(edge_dict, num_nodes_dict=None):
+    """Create a heterogeneous graph from a dictionary between edge types and edge lists.
```
jermainewang: Provide an example in the docstring

python/dgl/convert.py
```diff
@@ -264,6 +265,68 @@ def hetero_from_relations(rel_graphs):
...
+                ndst = max(dst) + 1
+            elif isinstance(edges, sp.sparse.spmatrix):
+                nsrc = edges.shape[0]
+                ndst = edges.shape[1]
+            elif isinstance(edges, nx.Graph):
+                if srctype == dsttype:
+                    nsrc = ndst = edges.number_of_nodes()
+                else:
+                    nsrc = len({n for n, d in edges.nodes(data=True) if d['bipartite'] == 0})
+                    ndst = edges.number_of_nodes() - nsrc
```
jermainewang: Add an `else` branch? Also, I'd like to support DGLHeteroGraph data as well. You could create an DGLHeteroGraph in each branch and call `hetero_from_relations` at the end.

BarclayII: The first loop is to infer number of nodes for each node type before actually creating DGLHeteroGraphs (and handle the last test case in `test_heterograph.test_create()`).

PR #889, opened 2019-09-30
edge sampler return head nodes and tail nodes.

zheng-da:
## Description
When we sample edges for link prediction, we need to know which nodes are head nodes and which are tails nodes. although a user can figure out themselves by call `all_edges` and `unique`. This is expensive for KGE, which samples many negative edges. Here we provide a very efficient implementation.

## Checklist
- [ ] Code is well-documented


python/dgl/contrib/sampling/sampler.py
```diff
@@ -436,6 +436,45 @@ def fetch(self, current_nodeflow_index):
         nflows = [NodeFlow(self.g, obj) for obj in nfobjs]
         return nflows
 
+class EdgeSubgraph(subgraph.DGLSubGraph):
+    ''' The subgraph sampled from an edge sampler.
+
+    A user can access the head nodes and tail nodes of the subgraph directly.
```
jermainewang: Maybe give a concrete subgraph example and what's the output of head_nid and tail_nid?

jermainewang: Actually, such example is better in the docstring of EdgeSampler.

src/graph/sampler.cc
```diff
@@ -979,6 +977,25 @@ IdArray CheckExistence(GraphPtr gptr, IdArray relations,
   return exist;
 }
 
+IdArray CreateArray(const std::vector<dgl_id_t> &ids) {
+  int64_t len = ids.size();
+  IdArray arr = IdArray::Empty({len}, DLDataType{kDLInt, 64, 1}, DLContext{kDLCPU, 0});
+  dgl_id_t *arr_data = static_cast<dgl_id_t *>(arr->data);
+  std::memcpy(arr_data, ids.data(), ids.size() * sizeof(*arr_data));
+  return arr;
+}
```
jermainewang: This one? https://github.com/dmlc/dgl/blob/master/include/dgl/array.h#L45

src/graph/sampler.cc
```diff
@@ -1032,8 +1052,30 @@ NegSubgraph NegEdgeSubgraph(GraphPtr gptr, IdArray relations, const Subgraph &po
...
+
+    // Get all nodes in the positive side.
+    for (int64_t i = 0; i < num_pos_edges; i++) {
+      dgl_id_t vid = induced_vid_data[unchanged[i]];
+      local_pos_vids.push_back(neg_map[vid]);
+    }
+    // There is no guarantee that the nodes in the vector are unique.
+    std::sort(local_pos_vids.begin(), local_pos_vids.end());
+    auto it = std::unique(local_pos_vids.begin(), local_pos_vids.end());
+    local_pos_vids.resize(it - local_pos_vids.begin());
```
jermainewang: My general question is how will this change if a bipartite graph is returned? Will the logic become cleaner?

zheng-da: yes, the logic will be cleaner if it's a bipartite graph.

PR #902, opened 2019-10-05
[KG] reduce memory consumption.

zheng-da:
## Description
The original implementation loads data into a list of triplets, which is stored as a list of tuples. This is slow and memory consuming when the KG is large. We should use numpy arrays to store the edges of a KG.

## Checklist
- [ ] Code is well-documented


apps/kg/dataloader/sampler.py
```diff
@@ -29,35 +29,38 @@ def RelationPartition(edges, n):
...
         parts.append([])
-    for h, r, t in edges:
-        idx = rel_dict[r]
-        parts[idx].append((h, r, t))
+    # let's store the edge index to each partition first.
+    for i, r in enumerate(rels):
+        part_idx = rel_dict[r]
+        parts[part_idx].append(i)
+    for i, part in enumerate(parts):
+        parts[i] = np.array(part, dtype=np.int64)
```
classicsong: Will parts[part_idx] overwritten by parts[i]?
It is doing in-place update.

apps/kg/dataloader/sampler.py
```diff
@@ -221,9 +223,9 @@ def __init__(self, dataset, args):
                     pickle.dump(g, graph_file)
         self.g = g
 
-        self.num_train = len(dataset.train)
-        self.num_valid = len(dataset.valid)
-        self.num_test = len(dataset.test)
+        self.num_train = len(dataset.train[0])
```
classicsong: what is the different between dataset.train and dataset.train[0]. Why it is changed here?

apps/kg/dataloader/sampler.py
zheng-da: I'm not sure what you mean. `parts[part_idx]` isn't in this for loop.

apps/kg/dataloader/sampler.py
zheng-da: originally, `dataset.train` stores a list of tuples. Each of the tuples represents an edge. Now `dataset.train` stores a tuple of numpy ndarray.

PR #920, opened 2019-10-10
[Backend] support Deep Numpy.

zheng-da:
## Description
This is to add Deep Numpy as one of the DGL backends.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/backend/numpy/tensor.py
+++ b/python/dgl/backend/numpy/tensor.py
@@ -67,6 +67,9 @@ def shape(input):
 def dtype(input):
     return input.dtype
 
+def ndim(input):
+    return input.ndim
+
 def context(input):
     return input.context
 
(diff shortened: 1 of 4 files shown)


mufeili:
I think we should not replace the numpy backend with deep numpy, but create a separate backend. NumPy is a default dependency and backend if no DL framework is detected. This modification will have users installed MXNet even they are not MXNet users, which is weird.

python/dgl/backend/numpy/tensor.py
```diff
@@ -1,23 +1,25 @@
 from __future__ import absolute_import
 
-import numpy as np
+import numpy as onp
+from mxnet import numpy as np
+import mxnet as mx
```
futurely: What are differences between mxnet.numpy backend and mxnet backend?

yzh119: `mxnet.numpy` is a numpy-like interface for MXNet 2.0: https://numpy.mxnet.io/, the goal of this project is to make MXNet compatible with numpy and accelerate basic operations with TVM. It's still under development.

VoVAllen:
Need to change `mx.set_np_shape(True)` to `npx.set_np()`

PR #937, opened 2019-10-18
[bugfix]adapt networkx version to run the tutorial

burness:
## new networkx wouldn't support the syntax dgl used
Hi, DGLer

I'm a new dgl developer,  I build a Docker Image according the docker scripts. I found a error when I run the tutorial
image
It was caused by the versions of `networkx` as show in [https://networkx.github.io/documentation/stable/release/migration_guide_from_1.x_to_2.0.html](https://networkx.github.io/documentation/stable/release/migration_guide_from_1.x_to_2.0.html). Unfortunately  network-2.3 wouldn't support the syntax that the dgl used. So we may specify the version of `networkx` or change the all `networkx` api in the dgl code.

### DIFF
--- a/docker/install/ubuntu_install_python_package.sh
+++ b/docker/install/ubuntu_install_python_package.sh
@@ -1,3 +1,3 @@
 # install libraries for python package on ubuntu
 #pip2 install nose numpy cython scipy networkx matplotlib nltk requests[security] tqdm
-pip3 install nose numpy cython scipy networkx matplotlib nltk requests[security] tqdm
+pip3 install nose numpy cython scipy networkx==2.1 matplotlib nltk requests[security] tqdm


classicsong:
The current docker image (built from the dockerfile) is only for CI-test.
I think it is much better to add version notification for corresponding tutorial docs and the example README, but not the dockerfile.

burness:
😃  😃  😃  @classicsong

classicsong:
> The current docker image (built from the dockerfile) is only for CI-test.
> I think it is much better to add version notification for corresponding tutorial docs and the example README, but not the dockerfile.
> 
> BTW @burness It is better to lunch an issue first and then propose your fix.

Same problem also for setup.py. see #948 for more details

burness:
I will review the dgl python code with networkx this weekend. If the time is allowed, I will help to update your code with the new version NetworkX api @classicsong @mufeili

mufeili:
@burness Awesome, thanks for the help.

jermainewang:
@burness #949 has changed all the deprecated `G.selfloop_edges` to functional `nx.selfloop_edges(G)`. This API is compatible with all networkx version after v2.1. Thus, we don't need to upgrade the depending networkx version of DGL anymore. Thanks for the fix and report of this problem.

burness:
👍

PR #945, opened 2019-10-21
[NN]Supporting TransR in app/kg score_func

classicsong:
## Description
Providing TransR score_function support for app/kg with both pytorch and mxnet.
As TransR consumes much more memory, the current GPU implementation can not support large embedding size. 

Currently with pytorch backend we can support embed_dim <= 500:
    Exectime and Result for full gpu training (FB15k): (dim=400)
```
forward: 24.106, backward: 38.071, update: 14.093
Test average MRR at [99999/100000]: 0.5014248703249872
Test average MR at [99999/100000]: 138.66557193885325
Test average HITS@1 at [99999/100000]: 0.2738314909177092
Test average HITS@3 at [99999/100000]: 0.70398334207987
Test average HITS@10 at [99999/100000]: 0.8009513974708402
```

With mxnet backend due to its support for large Tensor, we only support embed_dim <=100.
Exectime and Result for full gpu training (FB15k): (dim=100, The result may not be the best score):
```
641.230327129364
Test average HITS@10 at [99999/100000]: 0.6966701088520594
Test average MRR at [99999/100000]: 0.4867585735211409
Test average HITS@1 at [99999/100000]: 0.3387533645951482
Test average MR at [99999/100000]: 241.1212862487515
Test average HITS@3 at [99999/100000]: 0.6026476612889573
```

## Checklist
- [x] Code is well-documented


apps/kg/models/mxnet/score_fun.py
```diff
@@ -46,6 +57,125 @@ def fn(heads, relations, tails, num_chunks, chunk_size, neg_sample_size):
...
+                    for j in range(num_nnodes):
+                        tail_neg = tail[i][j]
+                        tail_neg = tail_neg.reshape(1, 1, self.entity_dim)
+                        tail_neg = nd.broadcast_axis(tail_neg, axis=0, size=num_rels)
+                        tail_neg = nd.batch_dot(tail_neg, projection[i])
+                        tail_neg = tail_neg.squeeze(axis=1)
+                        tail_negs.append(tail_neg)
+                    tail_negs = nd.stack(*tail_negs, axis=1)
+                    tails.append(tail_negs)
+                tail = nd.stack(*tails)
```
zheng-da: you are iterating over individual nodes. how about projecting negative nodes to edges? in this way, all computation can be batched. It might have better performance.

apps/kg/models/mxnet/score_fun.py
```diff
@@ -46,6 +57,125 @@ def fn(heads, relations, tails, num_chunks, chunk_size, neg_sample_size):
...
+        return {'score': self.gamma - nd.norm(score, ord=1, axis=-1)}
+
+    def prepare(self, g, gpu_id, trace=False):
+        head_ids, tail_ids = g.all_edges(order='eid')
+        projection = self.projection_emb(g.edata['id'], gpu_id, trace)
+        projection = projection.reshape(-1, self.entity_dim, self.relation_dim)
+        head_emb = g.ndata['emb'][head_ids.as_in_context(g.ndata['emb'].context)].expand_dims(axis=-2)
+        tail_emb = g.ndata['emb'][tail_ids.as_in_context(g.ndata['emb'].context)].expand_dims(axis=-2)
+        g.edata['head_emb'] = nd.batch_dot(head_emb, projection).squeeze()
+        g.edata['tail_emb'] = nd.batch_dot(tail_emb, projection).squeeze()
```
zheng-da: i don't feel that we need `prepare`. we can put all computation in `forward`. It shouldn't affect the speed.

apps/kg/models/mxnet/score_fun.py
classicsong: Iterating because I want to save memory with MXNet backend. MXNet do not provide dot operation supporting broadcast. 
If we want to project nodes to edges, we should build the neg-head and neg-tail graph from pos and neg graph. This should break most of our code I think. BTW, in MXNet 1.6, deep numpy will provide einsum API, then we can simplify this code.

apps/kg/models/mxnet/score_fun.py
classicsong: Yes, combining forward and prepare will not affect the speed.
The reason I use prepare here is to avoid breaking the interface of original forward func and make the code more flexible.  If we want to add other score funcs, we may reuse prepare to do some data preprocessing or transformation.

apps/kg/tests/test_score.py
```diff
@@ -99,9 +141,24 @@ def check_score_func(func_name):
...
+def test_score_func_transr():
+    check_score_func('TransR')
 
 if __name__ == '__main__':
-    test_score_func()
+    test_score_func_transe()
+    test_score_func_distmult()
+    test_score_func_complex()
+    test_score_func_rescal()
+    test_score_func_transr()
```
zheng-da: why is this preferred way of running tests?

PR #946, opened 2019-10-23
[BUG] Fix #717

### LINKED ISSUE #717: [Bug] DGLSubGraph.map_to_subgraph_nid error in v0.3
Hello,

The method map_to_subgraph_nid from DGLSubGraph seems to malfunction. 

The following example code:
`import dgl`
`g=dgl.DGLGraph()`
`g.add_nodes(10)`
`h=g.subgraph([0,1,2,5,8])`
`h.map_to_subgraph_nid([0,8,2])`

produces the following error chain: 
`File “.../python3.6/dist-packages/dgl/subgraph.py”, line 139, in map_to_subgraph_nid`
`    return map_to_subgraph_nid(self._graph,utils.toindex(parent_vids)).tousertensor()`
`File “.../python3.6/dist-packages/dgl/graph_index.py”, line 1137, in map_to_subgraph_nid`
`    return utils.toindex(_CAPI_DGLMapSubgraphNID(subgraph.induced_nodes.todgltensor(),`
`AttributeError: ‘GraphIndex’ object has no attribute ‘induced_nodes’
`

This bug does not occur in version 0.2 . The main difference I have found was that in v0.2 (in my code), `type(h._graph)` was 
`<class ‘dgl.graph_index.SubGraphIndex’>`, instead of 
`<class ‘dgl.graph_index.GraphIndex’>` as it is in v0.3 .

Am I using the library incorrectly? Is this behavior intended? If so, is there a replacement for the method?
Or is it a bug?

Thanks!

BarclayII:
## Description
Fixes #717 .

We were only testing on `SubGraphIndex` but not testing on the induced `DGLGraph`, which is why the bug gets away.

### DIFF
--- a/python/dgl/subgraph.py
+++ b/python/dgl/subgraph.py
@@ -46,6 +46,7 @@ class DGLSubGraph(DGLGraph):
         self._parent = parent
         self._parent_nid = sgi.induced_nodes
         self._parent_eid = sgi.induced_edges
+        self._subgraph_index = sgi
 
     # override APIs
     def add_nodes(self, num, data=None):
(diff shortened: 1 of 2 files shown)


VoVAllen:
LGTM

Issue #948, opened 2019-10-23
[BUG] New version of networkx will break some example code


classicsong:
## 🐛 Bug
networkx 2.4 has removed selfloop_edges func which is used by some of DGL examples.
https://networkx.github.io/documentation/stable/release/release_2.4.html?highlight=selfloop_edges

## To Reproduce

```
pip install networkx==2.4
git clone https://github.com/dmlc/dgl.git
cd dgl/examples/mxnet/gcn/
DGLBACKEND=mxnet python3 train.py  --self-loop --dataset cora
```

## Expected behavior
The example should not crash

## Environment

 - DGL Version (e.g., 1.0): 0.4.x
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):MXNet
 - OS (e.g., Linux): Linux 
 - How you installed DGL (`conda`, `pip`, source): pip
 - Python version: 3.6

jermainewang:
Fixed in #949 .

PR #949, opened 2019-10-23
[Bug Fix] Fix package reliability bug of networkx 

classicsong:
## Description
The new version of networkx(>2.4) delete selfloop_edges API which will break several examples.
Update version requirement of networkx in both setup.py and tutorials/requirements.txt

#948 

## Checklist
- [ ] Code is well-documented


python/setup.py
```diff
@@ -120,7 +120,7 @@ def config_cython():
     install_requires=[
         'numpy>=1.14.0',
         'scipy>=1.1.0',
-        'networkx>=2.1',
+        'networkx>=2.1,<2.4',
```
jermainewang: Let's use the newer networkx version. That means we need to drop the usage of stale APIs.

classicsong: It think drop the usage of stale APIs is better. 
But I think we should find a way to avoid annoying error messages in lots of our examples before we fix them. This is not friendly to new DGL users when they run the tutorial codes and have misleading errors like this.
At least we should add some Doc announce maybe.

PR #954, opened 2019-10-25
[Cleanup] Change Byte to Bool for training masks

JakeStevens:
## Description
Using ByteTensor for indexing has been deprecated in PyTorch, leading to many warnings in the example traning scripts. I've updated these to use BoolTensor.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/examples/pytorch/appnp/train.py
+++ b/examples/pytorch/appnp/train.py
@@ -25,9 +25,9 @@ def main(args):
     data = load_data(args)
     features = torch.FloatTensor(data.features)
     labels = torch.LongTensor(data.labels)
-    train_mask = torch.ByteTensor(data.train_mask)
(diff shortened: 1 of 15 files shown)


yzh119:
Thanks for your contribution!
Could you please also mention the minimum version number of PyTorch that supports `BoolTensor` in the README file of these examples? Because users who did not upgrade their PyTorch would receive errors after your changes.

JakeStevens:
I've added a check instead, so that if the version of torch being used has BoolTensors, they are used, thus avoiding the warning. Otherwise, ByteTensor is used as in the original code. This way, users who did not upgrade will not get errors since the code is unchanged for them.

jermainewang:
Thanks @JakeStevens for the contribution.

PR #983, opened 2019-11-09
[Test] Use pytest and fix gather rows

VoVAllen:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/backend/__init__.py
+++ b/python/dgl/backend/__init__.py
@@ -35,6 +35,8 @@ def load_backend(mod_name):
             setattr(thismod,
                     'reverse_data_type_dict',
                     {v: k for k, v in data_type_dict.items()})
+            # log backend name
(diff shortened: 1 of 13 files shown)


VoVAllen:
Need support for `@pytest.mark.skipif` to skip some test with tensorflow. TF doesn't support inplace mutate. And pytest has better error message when assert fails (would print out the assert values)

yzh119:
Hi JJ, could you please explain why do we need pytest, is it for tf integration?
It seems many unittests haven to be re-written, the effort is not trivial...

VoVAllen:
Only changes under `tests/scripts/` is for pytest. Others are for tensorflow. Because tf doesn't support get rows by `tensor[[0,2]]`, but only with `F.gather_row(tensor, (0,2))`

jermainewang:
Shall we close this in favor of #1084?

PR #1004, opened 2019-11-18
[Doc] The drawing will not disappear.

guotong1988:
## Description
The drawing will not disappear.

## Changes
The drawing will not disappear.

PR #1007, opened 2019-11-18
[Doc] the drawing will not disappear

guotong1988:
https://github.com/dmlc/dgl/pull/1004

### DIFF
--- a/tutorials/basics/1_first.py
+++ b/tutorials/basics/1_first.py
@@ -91,6 +91,12 @@ nx_G = G.to_networkx().to_undirected()
 pos = nx.kamada_kawai_layout(nx_G)
 nx.draw(nx_G, pos, with_labels=True, node_color=[[.7, .7, .7]])
 
+# Uncomment the code below, then the drawing will not disappear
+# import matplotlib.pyplot as plt
+# plt.pause(0)
+
+
+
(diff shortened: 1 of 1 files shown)


guotong1988:
On Mac, Pycharm

yzh119:
Thanks for your contribution, however I don't think it's necessary to add such comments:
1. The [tutorial page](https://docs.dgl.ai/tutorials/basics/1_first.html) is generated from the `1_first.py`, `pause` would hang the doc generation. Even if you comment these lines, the text would appear in the tutorial, which is confusing.
2. We assume users running tutorials block by block, in this case the problem you mention will not happen.
3. Even if they run the entire `1_first.py` code, as @VoVAllen mentioned, user could browse previous generated figures by clicking at "back" button.

guotong1988:
> 3. Even if they run the entire `1_first.py` code, as @VoVAllen mentioned, user could browse previous generated figures by clicking at "back" button.

The previous generated figures can not be found.

yzh119:
I see where the problem is, in this tutorial we did not call `plt.show()` so the figure would not appear when you run `1_basics.py` in command line.

One solution is to add a line: `plt.show()` before `plt.close()`.

PR #1027, opened 2019-11-21
[Doc] Old wines new title, edit for grammar and style

john-andrilla:
new descriptive title here and an edit pass @aaronmarkham


tutorials/models/4_old_wines/README.txt
```diff
@@ -1,25 +1,24 @@
 .. _tutorials4-index:
 
 
-Old (new) wines in new bottle
-=============================
+Updates to computer vision and RNNs
```
jermainewang: The new title is not accurate. The idea is that there is a new way to view classical models in computer vision and natural language processing using graphs. It is not an update or improvement over the current models. Also, the NLP example we used here is transformer. What is the problem with the old title?

aaronmarkham: I think the title should be updated too. I get what you're going for, but I had to read it like 3 times, then read the description and then decide that it had nothing to do with wine at all. I was both confused and disappointed.

jermainewang: Then how about "Revisit classics from graph perspective"?

john-andrilla: That's an improvement but what exactly is 'classics'? For search optimization and to help the reader with less ML experience, let's be as straightforward as possible. Consider the other topic titles that are peers to this: Graph neural networks and its variants, Dealing with many small graphs, Generative models, Training on giant graphs. These provide a more simple description of the content therein.

john-andrilla:
```suggestion
Revisit classic models from a graph perspective
```

john-andrilla: Revisit classic models from a graph perspective

john-andrilla:
added a title closer to the author's suggestion

jermainewang:
Thanks John!

PR #1028, opened 2019-11-21
[Bug] fix the data path bug for cora dataset

guotong1988:


yzh119:
Could you please elaborate more? I don't see any problem with the original code.

guotong1988:
I cannot download the zip file in IDE. So I download the zip file by Chrome and put it in ~/.dgl/

BarclayII:
Could you show the error message from your IDE when downloading the ZIP file?

guotong1988:
It is a network error. My Chrome can visit the site. But IDE cannot visit the site.

BarclayII:
> It is a network error. My Chrome can visit the site. But IDE cannot visit the site.

Does running from command line work for you?  If so, it is likely the problem of IDE configuration instead of DGL.

guotong1988:
I will try it tomorrow. Hope you wait about 12 hours.

guotong1988:
By the way, I am in China. I use VPN to visit the site. But somehow it does not work for the IDE.

BarclayII:
> By the way, I am in China. I use VPN to visit the site. But somehow it does not work for the IDE.

I see.  FYI, we are working on hosting a mirror site of DGL data repository within mainland China.

yzh119:
Plus, please complete necessary items in our [pull request template](https://github.com/dmlc/dgl/blob/master/.github/PULL_REQUEST_TEMPLATE.md) rather than leave it blank so that we could know what's the purpose of this PR and what you have done in the PR, thanks!

yzh119:
I think in your case, `cd $HOME/.dgl; mv cora_raw cora` would work, right?
We would fix the network issue recently.

guotong1988:
OK

guotong1988:
Oh, I just find that it is not the network reason. The error is:
```
"Can't connect to HTTPS URL because the SSL " "module is not available."
urllib3.exceptions.SSLError: Can't connect to HTTPS URL because the SSL module is not available.
```
```
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='s3.us-east-2.amazonaws.com', port=443): Max retries exceeded with url: /dgl.ai/dataset/FB15k-237.tgz (Caused by SSLError("Can't connect to HTTPS URL because the SSL module is not available.",)
```
Something wrong with my Python3.

PR #1032, opened 2019-11-22
[Bug] This PR can make model to run on FB15k-237 successfully.

guotong1988:
## Description

I run the link prediction example and encounter this bug.

## Checklist

- [x] Code is well-documented

## Changes

I changed the data path

### DIFF
--- a/python/dgl/contrib/data/knowledge_graph.py
+++ b/python/dgl/contrib/data/knowledge_graph.py
@@ -171,7 +171,7 @@ class RGCNLinkDataset(object):
     def __init__(self, name):
         self.name = name
         self.dir = get_download_dir()
-        tgz_path = os.path.join(self.dir, '{}.tar.gz'.format(self.name))
+        tgz_path = os.path.join(self.dir, '{}.tgz'.format(self.name))
         download(_downlaod_prefix + '{}.tgz'.format(self.name), tgz_path)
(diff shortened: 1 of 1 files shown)


BarclayII:
Hi,

I ran the code without this fix and it worked fine in PyTorch RGCN example.  Could you explain which example you ran and how you ran it (with what command line arguments), and what was the error message?

I remembered that you were mainly running examples in IDE and you could not download the files because of network accessibility of the S3 buckets.  Is this the same reason why your FB15k-237 is failing?

guotong1988:
As my IDE cannot download the tgz file. I download the tgz file and put it in `~/.dgl/`

guotong1988:
Oh, It is not the network reason. The error is:
```
"Can't connect to HTTPS URL because the SSL " "module is not available."
urllib3.exceptions.SSLError: Can't connect to HTTPS URL because the SSL module is not available.
```

BarclayII:
Then this is the same issue as the one in #1028.

The `download` function in line 175 has the following signature:
```python
download(url, path)
```
which saves the file from the given URL to the given path.  If you were to do this manually because of network issues, you actually have to rename your downloaded `.tgz` file to `.tar.gz`.

So I don't think there is any problem with the current code itself.  Once the S3 mirror is online we would fix the network accessibility issue once and for all (hopefully).

PR #1058, opened 2019-11-28
[KVStore] API change of kvstore

aksnzhy:
## Description
This PR provides better user experience of kvstore.

## Checklist
- [x] Code is well-documented


python/dgl/contrib/dis_kvstore.py
```diff
@@ -220,29 +220,31 @@ def _push_handler(self, name, ID, data):
             a vector storing the ID list.
         data : tensor (mx.ndarray or torch.tensor)
             a tensor with the same row size of id
+        target : target tensor (mx.ndarray or torch.tensor)
+            the target tensor
```
zheng-da: what is target tensor?

python/dgl/contrib/dis_kvstore.py
```diff
@@ -275,17 +279,27 @@ class KVClient(object):
           2:'168.12.46.12:50051' }
     client_addr : str
         IP address of current KVClient, e.g., '168.12.23.22:50051'
+    partition_book : list of int
+        A book mapping for global ID and server ID
+    global_to_local : list of int
```
zheng-da: why do we need to map global id to local id?

python/dgl/contrib/dis_kvstore.py
```diff
@@ -305,7 +319,7 @@ def connect(self):
         client_ip, client_port = self._addr.split(':')
         _receiver_wait(self._receiver, client_ip, int(client_port), self._server_count)
 
-    def init_data(self, name, server_id, shape, init_type='zero', low=0.0, high=0.0):
+    def init_remote_data(self, name, server_id, shape, init_type='zero', low=0.0, high=0.0):
```
zheng-da: why it needs server_id?

python/dgl/contrib/dis_kvstore.py
```diff
@@ -338,76 +352,157 @@ def init_data(self, name, server_id, shape, init_type='zero', low=0.0, high=0.0)
             id=tensor_shape,
             data=threshold)
         _send_kv_msg(self._sender, msg, server_id)
+
+    def init_local_data(self, name, data_tensor):
```
zheng-da: `data_tensor` is data for the local partition?

zheng-da:
You didn't add tests to verify the change of the API.

zheng-da:
I'll use this set of API for now. But let's design a more consistent API. Could you summarize the current API in a doc and have a meeting to discuss the API next week? Even though kvstore will be hidden from users most of time, it's still good to have a set of consistent API.

python/dgl/contrib/dis_kvstore.py
aksnzhy: It's self._data_store, not a tensor. The comment line is wrong.

python/dgl/contrib/dis_kvstore.py
aksnzhy: Because when we meet local partition data, we need to convert it to the local id. But the ID passed to pull/push is global ID.

python/dgl/contrib/dis_kvstore.py
aksnzhy: We can remove it.

python/dgl/contrib/dis_kvstore.py
aksnzhy: Yes.

aksnzhy:
> You didn't add tests to verify the change of the API.

I will add test and demo today.

examples/pytorch/dis_kvstore/run.py
```diff
@@ -0,0 +1,161 @@
...
+    server.init_data(name='embed', data_tensor=data_tensor)
+
+    server.start()
+
+def start_client(client_id, data_tensor, global_to_local):
+    time.sleep(2) # wait server start
+
+    client = dgl.contrib.KVClient(
+        client_id=client_id, 
+        local_server_id=client_id,
```
zheng-da: why do we need `local_server_id`?

python/dgl/contrib/dis_kvstore.py
```diff
@@ -220,72 +250,84 @@ def _push_handler(self, name, ID, data):
...
+        (start from 0) and value is the server's IP address and port, e.g.,
 
         { 0:'168.12.23.45:50051', 
           1:'168.12.23.21:50051', 
           2:'168.12.46.12:50051' }
     client_addr : str
-        IP address of current KVClient, e.g., '168.12.23.22:50051'
+        IP address and port of current KVClient, e.g., '168.12.23.22:50051'
+    partition_book : list or tensor (mx.ndarray or torch.tensor)
+        A book that maps global ID to target server ID.
```
zheng-da: if we don't shuffle vertices after metis partitioning, every machine needs to maintain a full map from global vid to their locations. it seems pretty expensive.

python/dgl/contrib/dis_kvstore.py
```diff
@@ -220,195 +363,320 @@ def _push_handler(self, name, ID, data):
...
-            rank=self._client_id,
-            name=name,
-            id=tensor_shape,
-            data=threshold)
-        _send_kv_msg(self._sender, msg, server_id)
-        
-    def push(self, name, server_id, id_tensor, data_tensor):
-        """Push sparse message to target KVServer.
+            type=KVMsgType.IP_ID,
+            rank=0,
```
zheng-da: what does `rank` mean here?

python/dgl/contrib/dis_kvstore.py
```diff
@@ -220,195 +363,320 @@ def _push_handler(self, name, ID, data):
...
+            else:
+                msg = KVStoreMsg(
+                    type=KVMsgType.PUSH, 
+                    rank=self._client_id, 
+                    name=name,
+                    id=partial_id, 
+                    data=partial_data)
+                _send_kv_msg(self._sender, msg, server[idx])
+
+            start += count[idx]
```
zheng-da: ideally, this should be implemented in C. we should benchmark this in the real workloads.

python/dgl/contrib/dis_kvstore.py
```diff
@@ -449,3 +721,116 @@ def get_id(self):
...
+        finally:
+            s.close()
+        
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.bind(("",0))
+        s.listen(1)
+        port = s.getsockname()[1]
+        s.close()
+
+        return IP + ':' + str(port)
```
zheng-da: does this only get one IP? How to handle the machine with multiple IPs.

python/dgl/contrib/dis_kvstore.py
```diff
@@ -449,3 +721,116 @@ def get_id(self):
...
+        name : str
+            data name
+        ID : tensor (mx.ndarray or torch.tensor)
+            a vector storing the ID list.
+        data : tensor (mx.ndarray or torch.tensor)
+            a tensor with the same row size of id
+        target : tensor (mx.ndarray or torch.tensor)
+            the target tensor
+        """
+        target[name][ID] = data
```
zheng-da: we definitely don't want people to have user-defined logic on both clients and servers. we need to fix it later.

src/graph/network.h
```diff
@@ -21,9 +21,9 @@ using dgl::runtime::NDArray;
 namespace dgl {
 namespace network {
 
-// Max size of message queue for communicator is 200 MB
+// Max size of message queue for communicator is 500 MB
 // TODO(chao): Make this number configurable
-const int64_t kQueueSize = 200 * 1024 * 1024;
+const int64_t kQueueSize = 500 * 1024 * 1024;
```
zheng-da: this also needs to be fixed. we shouldn't set a hard-coded maximal size.

tests/compute/test_kvstore.py
```diff
@@ -1,108 +0,0 @@
...
-    server_embed.share_memory_()
-
-    pid = os.fork()
-    if pid == 0:
-        start_server(server_embed)
-    else:
-        time.sleep(2) # wait server start
-        start_client(server_embed)
-
-    assert th.equal(server_embed, th.tensor([ 4., 4., 14., 4., 24.])) == True
```
zheng-da: no test code?

python/dgl/contrib/dis_kvstore.py
aksnzhy: `rank` means the sender's ID. If the ID isn't been assigned, it could be 0 by default.

python/dgl/contrib/dis_kvstore.py
aksnzhy: The getsockname() api will find the default IP address, which should be the IP written by configure file.

python/dgl/contrib/dis_kvstore.py
aksnzhy: Sure. We can benchmark the overhead in real distributed GNN tasks. I test it on KGE tasks, and I found that this code does not take too much time.

tests/compute/test_kvstore.py
aksnzhy: I suggest that we should put the test code into the example folder instead of a CI environment and test it by ourselves. Because the distributed setting could be complicated.

src/graph/network.h
aksnzhy: I fix this by giving user an argument in python.

PR #1059, opened 2019-11-28
[KG][Score Func] Update TransE with L2 distance support.

classicsong:
Now TransE support both l1 distance and l2 distance for calculating scores.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/apps/kg/models/mxnet/score_fun.py
+++ b/apps/kg/models/mxnet/score_fun.py
@@ -4,13 +4,31 @@ from mxnet import gluon
 from mxnet.gluon import nn
 from mxnet import ndarray as nd
 
+def batched_l2_dist(a, b):
+    a_squared = nd.power(nd.norm(a, axis=-1), 2)
+    b_squared = nd.power(nd.norm(b, axis=-1), 2)
+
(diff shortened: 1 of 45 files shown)


apps/kg/models/mxnet/score_fun.py
```diff
@@ -5,16 +5,20 @@
 from mxnet import ndarray as nd
 
 class TransEScore(nn.Block):
-    def __init__(self, gamma):
+    def __init__(self, gamma, dist_func='l2'):
```
zheng-da: can you make the same modification for MXNet?

classicsong: You means supporting l2 distance for transE? The PR already has it.

zheng-da: i mean, can we have the similar implementation as the one in Pytorch version? so mxnet version and pytorch version has similar speed.

classicsong: > i mean, can we have the similar implementation as the one in Pytorch version? so mxnet version and pytorch version has similar speed.

Done.

Issue #1046, opened 2019-11-26
[Bug] Error when backward with retain_graph=True


maximillian91:
## ❓ Questions and Help

Any reason for clearing the ctx.backward_cache explicitly in every backward() method?

This is causing errors when calling `loss.backward(retain_graph=True)` twice: 
`TypeError: 'NoneType' object is not iterable`

Here's an example:
[typeerror-nonetype-object-is-not-iterable-in-loss-backward-in-pytorch](https://stackoverflow.com/questions/58655746/typeerror-nonetype-object-is-not-iterable-in-loss-backward-in-pytorch)

VoVAllen:
Hi,

Why you set retain_graph to true? Do you need second-order derivative?

maximillian91:
Hypothetically yes, but did not encounter the problem for that. For me it was just a quick fix of mistakenly constructing `target` required gradients, so the 2nd epoch `backward()` caused the error same error as in [typeerror-nonetype-object-is-not-iterable-in-loss-backward-in-pytorch](https://stackoverflow.com/questions/58655746/typeerror-nonetype-object-is-not-iterable-in-loss-backward-in-pytorch). In that question I think the DQN basically refer twice to the same instance of tensors and thereby the meeting `None` in the `ctx.backward_cache` on the second call of it. I'm not an expert on the DQN, so might be wrong. 

Anyway I don't understand why we need to explicitly clear the backward cache, `ctx.backward_cache = None`. Maybe you can clarify?

VoVAllen:
We found if we didn't do this, the cache is not properly cleared by pytorch, which results in memory leak.

jermainewang:
@BarclayII is re-confirming the memory leak issue. The latest pytorch may have already resolved this issue. Either way, this is likely a bug on our side and we never test `retain_graph=True`.

BarclayII:
@maximillian91 Could you give a minimal example of `retain_graph=True` failing?  The following seems to work for me on the `master` branch.

```python
import dgl
import numpy as np
import scipy.sparse as ssp
from dgl.nn.pytorch import SAGEConv
import torch
import torch.nn as nn

g = dgl.DGLGraph(ssp.random(20, 20, 0.2))
x = torch.randn(20, 10)
m = nn.Linear(10, 20)
g.ndata['x'] = x
g.ndata['w'] = m(x)

g.update_all(dgl.function.copy_u('w', 'm'), dgl.function.sum('m', 'y'))
[...]

PR #1060, opened 2019-11-29
[Backend] Memory leak fix in PyTorch backend

BarclayII:
## Description

PyTorch documentation recommends saving the tensors required to compute gradients with `ctx.save_for_backward`.  However, DGL currently requires several non-tensor objects, which cannot be saved with `ctx.save_for_backward`, to compute gradients.  We previously worked around this issue by putting everything required in a Python attribute (`ctx.backward_cache`), but without explicitly clearing the cache (i.e. setting `ctx.backward_cache = None`), PyTorch autograd would have trouble deallocating the tensors created in forward() function, hence the memory leakage.  The workaround presumably raises another issue related to backward'ing with `retain_graph=True` according to #1046 .

### Minimal reproducible code for PyTorch autograd memory leakage with custom attributes

The following code is copied and modified from `scatter_mul` function in PyTorch-Scatter, which demonstrates the memory leakage issue.

```python
class ScatterMul(Function):
    @staticmethod
    def forward(ctx, out, src, index, dim):
        func = get_func('scatter_mul', src)
        func(src, index, out, dim)

        ctx.mark_dirty(out)
        # This is the original code in PyTorch-Scatter, does not leak
        #ctx.save_for_backward(out, src, index)
        # Does not leak
        ctx.save_for_backward(out)
        ctx.backward_cache = src, index
        # Leaks
        #ctx.save_for_backward(src, index)
        #ctx.backward_cache = out
        ctx.dim = dim

        return out

    @staticmethod
    def backward(ctx, grad_out):
        # This is the original code in PyTorch-Scatter, does not leak
        #out, src, index = ctx.saved_tensors
        # Does not leak
        src, index = ctx.backward_cache
        out, = ctx.saved_tensors
        # Leaks
        #src, index = ctx.saved_tensors
        #out = ctx.backward_cache

        grad_src = None
        if ctx.needs_input_grad[1]:
            grad_src = (grad_out * out).gather(ctx.dim, index) / src

        return None, grad_src, None, None
```

Presumably, if an output tensor is saved in a custom attribute instead of with `save_for_backward`, then PyTorch Autograd somehow never deallocates the output tensor.  I don't know the reason why this is happening, though.

### Notes

This also coincidentally (?) fixes the issue where DGL `update_all()` leaks memory when not specifying `torch.no_grad()`.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/backend/pytorch/tensor.py
+++ b/python/dgl/backend/pytorch/tensor.py
@@ -326,15 +326,17 @@ class BinaryReduce(th.autograd.Function):
         # save_for_backward can only save variables
         ctx.backward_cache = (reducer, binary_op, graph, lhs, rhs, lhs_map,
                               rhs_map, out_map, lhs_data_nd, rhs_data_nd,
-                              out_data_nd, feat_shape, degs)
+                              feat_shape, degs)
+        ctx.save_for_backward(out_data)
         return out_data
 
     @staticmethod
     def backward(ctx, grad_out):
         reducer, binary_op, graph, lhs, rhs, lhs_map, rhs_map, out_map, \
-            lhs_data_nd, rhs_data_nd, out_data_nd, feat_shape, degs \
+            lhs_data_nd, rhs_data_nd, feat_shape, degs \
             = ctx.backward_cache
-        ctx.backward_cache = None
+        out_data, = ctx.saved_variables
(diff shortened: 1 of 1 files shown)


BarclayII:
This also coincidentally (?) fixes the issue where DGL `update_all()` leaks memory when not specifying `torch.no_grad()`.

PR #1068, opened 2019-12-02
[Example] add graphwriter pytorch example

QipengGuo:
## Description

Add the example of the graphwriter model, the [paper](https://arxiv.org/abs/1904.02342) and the [author's code](https://github.com/rikdz/GraphWriter).

The performance is reported in the README and I also give an output from a well-trained model, it can reproduce the result of the paper.


yzh119:
Good job!
Could you please remove the two temporary files in the output directory?

yzh119:
I've made some changes, @QipengGuo could you please give me a write access to your fork?

PR #1087, opened 2019-12-10
[Feature] No-Uniform Edge Sampler

classicsong:
## Description
We provide a no-uniform edge sampler by allowing users define customized weight tensor for positive nodes and weight tensor for negative nodes during sampling.

The implementation use a binary-tree search algo for random sampling a node. The time complexity for sampling one positive node is O(log(size_of_pos_edges)), the time complexity for sampling one negative node is O(log(num_of_nodes))

## Checklist
- [x] Code is well-documented


zheng-da:
Can we reuse the original EdgeSampler API?

classicsong:
> Can we reuse the original EdgeSampler API?

The use case are different, I suppose providing separate Python APIs is better.
But I plan to refactor the C++ impl of WeightedEdgeSampler and EdgeSampler. Must of code can be reused.

python/dgl/contrib/sampling/sampler.py
```diff
@@ -614,25 +625,233 @@ def fetch(self, current_index):
...
+        Whether the sampler only works on immutable graphs.
+        Subclasses can override this property.
+    '''
+    immutable_only = False
+
+    def __init__(
+            self,
+            g,
+            batch_size,
+            edge_weight,
```
zheng-da: can we make `edge_weight` and `node_weight` be optional so we only have one EdgeSampler?

zheng-da: it seems most of the logic is copied.

src/graph/sampler.cc
```diff
@@ -988,415 +988,769 @@ std::vector<dgl_id_t> Global2Local(const std::vector<size_t> &ids,
...
+});
+
+DGL_REGISTER_GLOBAL("sampling._CAPI_GetEdgeSubgraphTail")
+.set_body([] (DGLArgs args, DGLRetValue* rv) {
+  SubgraphRef g = args[0];
+  auto gptr = std::dynamic_pointer_cast<NegSubgraph>(g.sptr());
+  *rv = gptr->tail_nid;
+});
+
+class EdgeSamplerObject: public Object {
```
zheng-da: what is `Object`?

src/graph/sampler.cc
```diff
@@ -988,415 +988,769 @@ std::vector<dgl_id_t> Global2Local(const std::vector<size_t> &ids,
...
+protected:
+  virtual void randomSample(size_t set_size, size_t num, std::vector<size_t>* out) = 0;
+  virtual void randomSample(size_t set_size, size_t num, const std::vector<size_t> &exclude,
+                    std::vector<size_t>* out) = 0;
+
+  NegSubgraph genNegEdgeSubgraph(const Subgraph &pos_subg,
+                                 const std::string &neg_mode,
+                                 int64_t neg_sample_size,
+                                 bool exclude_positive,
+                                 bool check_false_neg) {
```
zheng-da: can you move the implementation of `genNegEdgeSubgraph` so that diff can work correctly to see what is changed and what isn't?

src/graph/sampler.cc
```diff
@@ -988,415 +988,769 @@ std::vector<dgl_id_t> Global2Local(const std::vector<size_t> &ids,
...
-      LOG(FATAL) << "We can't exclude positive edges when sampling negative edges with all nodes.";
+    }
+    return neg_subg;
+  }
+
+  NegSubgraph genPBGNegEdgeSubgraph(const Subgraph &pos_subg,
+                                    const std::string &neg_mode,
+                                    int64_t neg_sample_size,
+                                    bool exclude_positive,
+                                    bool check_false_neg) {
```
zheng-da: the same here.

src/graph/sampler.cc
```diff
@@ -988,415 +988,769 @@ std::vector<dgl_id_t> Global2Local(const std::vector<size_t> &ids,
...
+  explicit UniformEdgeSamplerObject(const GraphPtr gptr,
+                                    IdArray seed_edges,
+                                    const int64_t batch_size,
+                                    const int64_t num_workers,
+                                    const std::string neg_mode,
+                                    const int64_t neg_sample_size,
+                                    const bool exclude_positive,
+                                    const bool check_false_neg,
+                                    IdArray relations)
+                                    : EdgeSamplerObject{gptr,
```
zheng-da: what is the difference between '{' and '(' here?

src/graph/sampler.cc
```diff
@@ -988,415 +988,769 @@ std::vector<dgl_id_t> Global2Local(const std::vector<size_t> &ids,
...
+    std::vector<SubgraphRef> positive_subgs(num_workers);
+    std::vector<SubgraphRef> negative_subgs(num_workers);
+#pragma omp parallel for
+    for (int64_t i = 0; i < num_workers; i++) {
+      const int64_t start = (batch_curr_id_ + i) * batch_size_;
+      const int64_t end = std::min(start + batch_size_, num_seeds_);
+      const int64_t num_edges = end - start;
+      IdArray worker_seeds = seed_edges_.CreateView({num_edges}, DLDataType{kDLInt, 64, 1},
+                                                   sizeof(dgl_id_t) * start);
+      EdgeArray arr = gptr_->FindEdges(worker_seeds);
```
zheng-da: can you make sure the ImmutableGraph has coo format before entering the openmp for loop? you can simply call GetCOO() before the for loop.

src/graph/sampler.cc
```diff
@@ -988,415 +988,769 @@ std::vector<dgl_id_t> Global2Local(const std::vector<size_t> &ids,
...
+                                        neg_sample_size,
+                                        exclude_positive,
+                                        check_false_neg,
+                                        relations} {
+    const size_t num_edges = edge_weight->shape[0];
+    const ValueType *edge_prob = static_cast<const ValueType*>(edge_weight->data);
+    std::vector<ValueType> eprob(num_edges);
+    for (size_t i = 0; i < num_edges; ++i) {
+      eprob[i] = edge_prob[i];
+    }
```
zheng-da: do you assume all elements in the probability array sum to 1?

src/graph/sampler.cc
```diff
@@ -988,415 +988,769 @@ std::vector<dgl_id_t> Global2Local(const std::vector<size_t> &ids,
...
+                                     IdArray relations)
+                                     : EdgeSamplerObject{gptr,
+                                        seed_edges,
+                                        batch_size,
+                                        num_workers,
+                                        neg_mode,
+                                        neg_sample_size,
+                                        exclude_positive,
+                                        check_false_neg,
+                                        relations} {
```
zheng-da: It seems the weighted sampler and unweighted sampler can also be merged.

src/graph/sampler.cc
classicsong: See include/dgl/runtime/object.h This is used for ffi object explored to python

python/dgl/contrib/sampling/sampler.py
```diff
@@ -614,25 +625,233 @@ def fetch(self, current_index):
         Parameters
         ----------
         current_index : int
-            How many batches the sampler has generated so far.
+            deprecated, not used actually.
```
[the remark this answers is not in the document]
classicsong: We still use fetch now.
The current_index is maintained in C++ now.

src/graph/sampler.cc
classicsong: I think no much difference. Both are OK.

src/graph/sampler.cc
classicsong: Nop.
They can sum to any float number. HeapArray can handle such cases.

python/dgl/contrib/sampling/sampler.py
classicsong: Emm..
We can merge the two EdgeSample in Python side.

src/graph/sampler.cc
classicsong: ImmutableGraph->FindEdges() will automatically do GetCOO()->FindEdges(eids);

src/graph/sampler.cc
classicsong: I don't think so, the fetch() logic are different. 
weighted sampler depends on HeapArray, but unweight sampler don't (it uses a pre-shuffled edge list).

src/graph/sampler.cc
zheng-da: I know. But when you call GetCOO() inside the openmp for loop and if the coo graph doesn't exist, every thread tries to generate the coo graph and assign it to the coo pointer. it may lead to race condition. i got this bug before. When I wrote this sampler, I forgot to do so.

src/graph/sampler.cc
zheng-da: can you use '('? '{' looks weird.

src/graph/sampler.cc
classicsong: I see

zheng-da: I don’t know how you like to refactor it. Currently, we dynamic cast the pointer to ImmutableGraph.

python/dgl/contrib/sampling/sampler.py
```diff
@@ -509,6 +510,13 @@ class EdgeSampler(object):
     The sampler returns EdgeSubgraph, where a user can access the unique head nodes
     and tail nodes directly.
 
+    This sampler also allows us sample positive edges and negative edges (actually, 
+    nodes) according to their weight. When doing sample we can provide two tensors 
+    representing the weight of each edges and weight of each nodes respectively 
+    should be provided. These weights are used to decide the posibility of
```
zheng-da: How about this?

This sampler allows to non-uniformly sample positive edges and negative edges. For non-uniformly sampling positive edges, users need to provide an array of m elements (`m` is the number of edges), each of which represents the sampling probability of an edge.  For non-uniformly sampling negative edges, users need to provide an array of n elements and the sampler samples nodes based on the sampling probability to corrupt a positive edge.

src/graph/sampler.cc
```diff
@@ -97,12 +97,68 @@ class ArrayHeap {
...
+    num_workers_ = num_workers;
+    neg_mode_ = neg_mode;
+    neg_sample_size_ = neg_sample_size;
+    exclude_positive_ = exclude_positive;
+    check_false_neg_ = check_false_neg;
+  }
+
+  ~EdgeSamplerObject() {}
+
+  virtual void fetch(DGLRetValue* rv) = 0;
```
zheng-da: I think DGL C++ convention is to use capital letter for method names.

tests/compute/test_sampler.py
```diff
@@ -283,8 +287,13 @@ def check_negative_sampler(mode, exclude_positive, neg_size):
         else:
             assert F.array_equal(g.has_edges_between(neg_src, neg_dst), exist)
 
+        total_samples += batch_size
+        if (total_samples >= max_samples):
```
zheng-da: Why do we need this check?

tests/compute/test_sampler.py
```diff
@@ -303,11 +312,222 @@ def check_negative_sampler(mode, exclude_positive, neg_size):
...
+    # Test the homogeneous graph.
+    batch_size = 50
+    total_samples = 0
+    max_samples = num_edges
+    for pos_edges, neg_edges in EdgeSampler(g, batch_size,
+                                                    edge_weight=edge_weight,
+                                                    negative_mode=mode,
+                                                    neg_sample_size=neg_size,
+                                                    exclude_positive=exclude_positive,
+                                                    return_false_neg=True):
```
zheng-da: Indent?

tests/compute/test_sampler.py
```diff
@@ -303,11 +312,222 @@ def check_negative_sampler(mode, exclude_positive, neg_size):
...
+    edge_tail_half_cnt = edge_sampled[edge_sampled.shape[0] // 2:-1].sum()
+    edge_rate_tail_half = edge_tail_half_cnt / edge_sampled.sum()
+    assert np.allclose(edge_rate_0, 0.5, atol=0.05)
+    assert np.allclose(edge_rate_tail_half, 0.25, atol=0.05)
+
+    node_rate_0 = node_sampled[0] / node_sampled.sum()
+    node_tail_half_cnt = node_sampled[node_sampled.shape[0] // 2:-1].sum()
+    node_rate_tail_half = node_tail_half_cnt / node_sampled.sum()
+    assert node_rate_0 < 0.02
+    assert np.allclose(node_rate_tail_half, 0.5, atol=0.02)
```
zheng-da: Is it possible that the check can fail?

classicsong: Possible but with very low possibility. The number is changed to have some divergence on the output

tests/compute/test_sampler.py
classicsong: The weighted sampler will not stop automatically. We should do it manually.

PR #1095, opened 2019-12-12
Update README.md

JakeStevens:
Extra open brace in the .bibtex entry results in an error.


yzh119:
Thank you @JakeStevens .

Issue #1112, opened 2019-12-16
edge sampler should support both sampling with or without replacement 


zheng-da:
## 🚀 Feature
Currently, edge sampler behaves differently for uniform sampling and non-uniform sampling. For uniform sampling, it samples without replacement. For non-uniform sampling, it samples with replacement. This results in another difference: uniform sampler returns a fixed number of batches while non-uniform sampler can return any number of batches. We should add support of sampling with and without replacement for both uniform and non-uniform sampling and have a flag to determine whether to use with or without replacement. When sampling with replacement, users also need to provide how many batches to generate.

classicsong:
Resolved

PR #1116, opened 2019-12-18
[Feature Improvement]Edge Sampler w and w/o replacement

classicsong:
## Description
Currently edge sampler behaves differently for uniform sampling and non-uniform sampling.
This Pull request add a replacement argument to the EdgeSampler API so users can decide whether to sampler will samples  a fixed number of batches or samples infinitely.

Related issue #1112 

## Checklist
- [x] Code is well-documented


python/dgl/contrib/sampling/sampler.py
```diff
@@ -530,10 +530,10 @@ class EdgeSampler(object):
...
     matches one of the edges in the graph.
 
-    For uniform sampling, the sampler generates only num_of_edges/batch_size 
-    samples. 
-
-    For uniform sampling, the sampler generates samples infinitly.
+    This sampler samples edges without replacement by default, which means it 
+    returns a fixed number of batches (i.e., num_edges/batch_size). However 
+    it accepts a replacement argument, which can be set to True to let the 
+    sampler generates any number of batches.
```
zheng-da: it seems with or without replacement, the sampler always generate `num_edges/batch_size` batches.

src/graph/sampler.cc
```diff
@@ -153,6 +155,7 @@ class EdgeSamplerObject: public Object {
 
   int64_t batch_size_;
   int64_t num_workers_;
+  bool replacement_;
```
zheng-da: it seems you didn't change the sampling algorithm. if replacement=True, you should use sampling with replacement.

python/dgl/contrib/sampling/sampler.py
classicsong: if replacement == False, it will generate num_edges/batch_sizes.
if replacement == True, It will generate samples infinitely.

src/graph/sampler.cc
```diff
@@ -88,12 +89,19 @@ class ArrayHeap {
...
    */
-  void SampleWithoutReplacement(size_t n, std::vector<size_t>* samples) {
+  size_t SampleWithoutReplacement(size_t n, std::vector<size_t>* samples) {
     // sample n elements
-    for (size_t i = 0; i < n; ++i) {
+    size_t i = 0;
+    for (; i < n; ++i) {
+      // heap is empty
+      if (heap_[1] == 0) {
+        break;
```
zheng-da: what does this mean?

src/graph/sampler.cc
```diff
@@ -1592,17 +1629,31 @@ class WeightedEdgeSamplerObject: public EdgeSamplerObject {
...
-        int64_t edge_id = edge_selector_->Sample();
-        edge_ids[i] = seed_edge_ids[edge_id];
+#pragma omp critical
+      {
+        if (replacement_ == false) {
+          size_t n = batch_size_;
+          size_t num_ids = edge_selector_->SampleWithoutReplacement(n, &edge_ids);
+          while (edge_ids.size() > num_ids) {
+            edge_ids.pop_back();
+          }
```
zheng-da: I'm not sure what it means. Why do you need to pop edge Ids?

src/graph/sampler.cc
classicsong: when use SampleWithoutReplacement, node will be deleted (weight cleaned to 0). So when the heap_[1] == 0, the whole heap is cleaned (no node left)

zheng-da: so there was a bug in the previous implementation?

src/graph/sampler.cc
```diff
@@ -1592,17 +1629,31 @@ class WeightedEdgeSamplerObject: public EdgeSamplerObject {
...
-    for (int i = 0; i < num_workers_; i++) {
+    for (int i = 0; i < num_workers; i++) {
       const dgl_id_t *seed_edge_ids = static_cast<const dgl_id_t *>(seed_edges_->data);
-      std::vector<int64_t> edge_ids(batch_size_);
+      std::vector<size_t> edge_ids(batch_size_);
 
-      for (int i = 0; i < batch_size_; ++i) {
-        int64_t edge_id = edge_selector_->Sample();
-        edge_ids[i] = seed_edge_ids[edge_id];
+#pragma omp critical
```
zheng-da: i think you need to move this line after line 1644.

src/graph/sampler.cc
classicsong: SampleWithoutReplacement may samples items less than required in “n”. （WithoutReplacement will not return sampled items, so at the last minibatch, it may return items < batchsize)

src/graph/sampler.cc
classicsong: I see.

Issue #1114, opened 2019-12-18
GCN sampler: pull instead of push model in copy reduce.


yuk12:
Hi,

Need help.
I am working on GCN sampler example. In current GCN model, which uses 2hop neighbors, during block_compute: in first iteration: layer-0 nodes send (or push) the features to layer-1 and layer-1 then aggregates the neighbors. Similarly, in second iteration: layer-1 sends (or pushes) the features to layer-2 which then aggregates over the neighbors.

What I am trying to implement is a pull model: where each node in layer-1 pulls the features from layer-0 (similarly, layer-2 pulls from layer-1) and then aggregates. Thus, the pull model avoids the collision (and inturn gets rid of the omp atomics in reduce functions).

I am able to change the DGL code and implement the pull model. And the code is working fine for 1 thread. While running the same code on more than 1 thread resulting in NaNs out of optimizer.step().
I have no idea why it is happening. 

I am setting #omp threads as: $ export OMP_NUM_THREADS=2 
I even tried by commenting the omp parallel statement in minigun while setting OMP_NUM_THREADS=2. Still no useful. I am consistently getting NaNs from the optimizer.step() regardless of the number of threads (>1).

Can you help me out here? 
Happy to provide more details if you need.

Thanks.

classicsong:
In the forward step, pull model can help eliminate atomic operations, but in the backward step, the model switch to push. Also you should take care of local variables and global variables in OMP loops.

BTW, we are currently investigating the possibility of applying target-parallelism for builtin message passing kernel which can totally eliminate atomic operations in both forward and backward.

yuk12:
Hi, Thanks for the quick response. Can you tell where the push model is being applied in the backward step? -- the copy reduce operation does not go through the backward step.

Good to know about target parallelism: it will be a significant improvement performance-wise.

Issue #1056, opened 2019-11-28
[Bug] Segfault on CPU BackwardBinaryReduce with -O0 compiler switch


BarclayII:
## 🐛 Bug

If compiler optimization is turned off ('-O0'), DGL would segfault when computing gradients of builtin kernels.

## To Reproduce

1. In `CMakeLists.txt`, replace the two `-O2`'s with `-g -O0`.
2. Build and install normally with CMake and `make`
3. Run the following and observe crash.

```python
import dgl
import numpy as np
import scipy.sparse as ssp
import torch
import torch.nn as nn
import tqdm

device = torch.device('cpu')
g = dgl.DGLGraph(ssp.random(2000, 2000, 0.02))
m = nn.Linear(1000, 2000).to(device)

for _ in tqdm.trange(10000):
    x = torch.randn(2000, 1000).to(device)
    g.ndata['x'] = x
    g.ndata['w'] = m(x)
    g.update_all(dgl.function.copy_u('w', 'm'), dgl.function.sum('m', 'y'))
    loss = g.ndata['y'].sum()
    loss.backward()
```

## Expected behavior

Shouldn't crash

## Environment

 - DGL Version (e.g., 1.0): master
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):
 - OS (e.g., Linux): Ubuntu 16.04 and CentOS 7
 - How you installed DGL (`conda`, `pip`, source): source
 - Build command you used (if compiling from source): see above
 - Python version: 3.6
 - CUDA/cuDNN version (if applicable):
 - GPU models and configuration (e.g. V100):
 - Any other relevant information:

## Additional context

GPU kernels do not crash.

Both OpenMP and non-OpenMP builds would crash.

**Would not crash with -O2 option.**

Relevant stacktrace:
```
#0  0x00007fffa11cb3bb in dgl::kernel::cpu::BackwardFunctorsTempl<int, float, dgl::kernel::SelectDst, dgl::kernel::SelectNone, dgl::ker
nel::BinaryUseLhs<float>, dgl::kernel::ReduceSum<1, float> >::Read (addr=0x7fff29e8e000)
    at /mnt/b/gq/dgl/src/kernel/cpu/./backward_binary_reduce_impl.h:185
#1  0x00007fffa11a0102 in dgl::kernel::cpu::BackwardBinaryReduce<0, int, float, dgl::kernel::cpu::BackwardFunctorsTempl<int, float, dgl
[...]

PR #1137, opened 2019-12-24
[WIP][Kernel] Target parallel optimization

classicsong:
## Description
Currently DGL builtin kernel use edge parallel which causes non-determinism in calculating forward results and backward gradients. A deterministic methodology is required. This PR provides a target parallel strategy: when calculation target point to dst nodes, it parallels execution across different dst nodes; when   calculation target point to edges,  it parallels execution across different edges. This totally eliminate the non-determinism.

#1114 
#1056 Fix this bug here

## Checklist
- [x] Code is well-documented


yzh119:
Windows build bug confirmed: `Fatal error C1002 : compiler is out of heap space in pass 2` ([documentation](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1002?view=vs-2019))

Interestly, caffe/pytorch also had met this problem before, their solution is to split the file:https://caffe2.ai/doxygen-c/html/generic_2_t_h_tensor_math_8cpp_source.html .

Also found in tensorflow windows build: https://github.com/tensorflow/tensorflow/issues/23342, they recommend using bazel instead of cmake.

Also a MXNet windows build problem: https://github.com/apache/incubator-mxnet/issues/13958.

It's said that enlarging precompiled header memory could solve the problem (https://docs.microsoft.com/en-us/cpp/build/reference/zm-specify-precompiled-header-memory-allocation-limit?view=vs-2019), however, this does not work for me. 

Another way is to disable some optimizations by setting `/LTCG:PGOptimize`, however, cmake did not recognize this option (https://stackoverflow.com/questions/47654646/how-to-disable-link-time-code-generation-in-cmake).

yzh119:
Limit the maximum broadcasting dimensionality to 2(https://github.com/dmlc/dgl/blob/master/src/kernel/binary_reduce_common.h#L501-L519) would fix the problem. 4 or 8 doesn't work here.

@jermainewang, @classicsong what's your idea?

BarclayII:
Splitting the CPU reducer source files into 32bit and 64bit reduces compiler memory usage by half and solves the issue.

I guess it's better to have a generator script that generates the `binary_bcast_reduce_${op}_${bits}.cc` files when building, instead of putting all these files in the repo?

yzh119:
This PR is being closed because it's outdated and all functions have been implemented in new PR.

PR #1152, opened 2019-12-30
[Bug Fix]edge sample hotfix

classicsong:
## Description
Fix some bug in edge sampler:
1. none-uniform sampler without replacement has a eid mapping bug.
2. If node-weight is extremely imbalanced, the sampler will take much very long time to sample a mini-batch, andd Note into docs.
3. ArrayHeap::Delete(idx) use minus in heap_ value update when a node is deleted. This will encounter floating point precision vanishing hazard. Change to use add to update heap_.

#1154

## Checklist
- [x] Code is well-documented


src/graph/sampler.cc
```diff
@@ -52,9 +52,11 @@ class ArrayHeap {
...
     size_t i = index + limit_;
     ValueType w = heap_[i];
-    for (int j = bit_len_; j >= 0; --j) {
-      heap_[i] -= w;
-      i = i >> 1;
+    heap_[i] = 0;
+    i /= 2;
+    for (int j = bit_len_-1; j >= 0; --j) {
+      heap_[i] = heap_[i << 1] + heap_[(i << 1) + 1];
+      i /= 2;
```
zheng-da: is there a bug here?

src/graph/sampler.cc
```diff
@@ -1697,9 +1698,12 @@ class WeightedEdgeSamplerObject: public EdgeSamplerObject {
         size_t n = batch_size_;
         size_t num_ids = 0;
 #pragma omp critical
-        num_ids = edge_selector_->SampleWithoutReplacement(n, &edge_ids);
-        while (edge_ids.size() > num_ids) {
-          edge_ids.pop_back();
+        {
+          num_ids = edge_selector_->SampleWithoutReplacement(n, &edge_ids);
+        }
```
zheng-da: does `edge_ids` here store indices?

classicsong: Yes

src/graph/sampler.cc
classicsong: yes

PR #1162, opened 2020-01-02
[KG] Disable filter in evaluation

zheng-da:
## Description
By default, we remove the positive edges from sampled negative edges in evaluation. We should allow disabling the filtering for fair comparison with other frameworks.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/apps/kg/eval.py
+++ b/apps/kg/eval.py
@@ -44,6 +44,8 @@ class ArgParser(argparse.ArgumentParser):
                           help='margin value')
         self.add_argument('--eval_percent', type=float, default=1,
                           help='sample some percentage for evaluation.')
+        self.add_argument('--no_eval_filter', action='store_true',
+                          help='do not filter positive edges among negative edges for evaluation')
 
         self.add_argument('--gpu', type=int, default=-1,
                           help='use GPU')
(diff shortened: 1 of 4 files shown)


apps/kg/eval.py
```diff
@@ -99,17 +101,20 @@ def main(args):
     args.neg_sample_size_test = args.neg_sample_size
     if args.neg_sample_size < 0:
         args.neg_sample_size_test = args.neg_sample_size = eval_dataset.g.number_of_nodes()
+    args.eval_filter = not args.no_eval_filter
```
classicsong: I think we cat directly change self.add_argument('--no_eval_filter' to --eval_filter

apps/kg/train.py
```diff
@@ -59,6 +59,8 @@ def __init__(self):
                           help='margin value')
         self.add_argument('--eval_percent', type=float, default=1,
                           help='sample some percentage for evaluation.')
+        self.add_argument('--no_eval_filter', action='store_true',
```
classicsong: same here no_eval_filter -> eval_filter

apps/kg/eval.py
zheng-da: actually, i want to have filtering by default and disable filtering only when a user explicitly asks for it.

classicsong: We can use: store_false

classicsong: I see. This make sense

PR #1178, opened 2020-01-07
[Feature][KG] Multi-GPU training support for DGL KGE

classicsong:
## Description
Multi-GPU training support (pytorch) for DGL Knowledge Graph Embedding app.

#932 

ComplEx GPU with batchsize=1024, emb size=2048, mix_cpu_gpu with relation partition:
1 GPU speed: 32.530 secs / 1000 steps
4 GPU speed: 41.643 secs / 1000 steps
8 GPU speed: 50.510 secs / 1000 steps

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/apps/kg/models/general_models.py
+++ b/apps/kg/models/general_models.py
@@ -44,7 +44,15 @@ class KEModel(object):
             rel_dim = relation_dim * entity_dim
         else:
             rel_dim = relation_dim
-        self.relation_emb = ExternalEmbedding(args, n_relations, rel_dim, device)
+
+        self.rel_part = args.rel_part
+        if not self.rel_part:
+            self.relation_emb = ExternalEmbedding(args, n_relations, rel_dim, device)
+        else:
+            self.relation_embs = [None] * args.num_proc
+            for i in range(args.num_proc):
+                relation_emb = ExternalEmbedding(args, n_relations, rel_dim, F.cpu())
+                self.relation_embs[i] = relation_emb
 
         if model_name == 'TransE' or model_name == 'TransE_l2':
             self.score_func = TransEScore(gamma, 'l2')
@@ -87,8 +95,10 @@ class KEModel(object):
 
     def reset_parameters(self):
         self.entity_emb.init(self.emb_init)
-        self.relation_emb.init(self.emb_init)
         self.score_func.reset_parameters()
+
+        if not self.rel_part:
+            self.relation_emb.init(self.emb_init)
 
     def predict_score(self, g):
         self.score_func(g)
@@ -248,3 +258,14 @@ class KEModel(object):
         self.entity_emb.update()
         self.relation_emb.update()
         self.score_func.update()
+
+    def prepare_relation(self, gpu_id=-1):
+        device = th.device('cuda:' + str(gpu_id))
+        self.relation_emb = ExternalEmbedding(self.args, self.n_relations, self.rel_dim, F.cpu() if gpu_id == -1 else device)
+        self.relation_emb.init(self.emb_init)
+
+    def writeback_relation(self, rank=0):
+        self.relation_embs[rank][:] = self.relation_emb[:]
+
+    def load_relation(self, rank=0):
+        self.relation_emb = self.relation_embs[rank]
--- a/apps/kg/models/general_models.py
+++ b/apps/kg/models/general_models.py
@@ -51,7 +51,7 @@ class KEModel(object):
         if not self.rel_part:
             self.relation_emb = ExternalEmbedding(args, n_relations, rel_dim, F.cpu() if args.mix_cpu_gpu else device)
         else:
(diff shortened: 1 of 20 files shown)


apps/kg/models/general_models.py
```diff
@@ -44,7 +45,13 @@ def __init__(self, args, model_name, n_entities, n_relations, hidden_dim, gamma,
...
         else:
             rel_dim = relation_dim
-        self.relation_emb = ExternalEmbedding(args, n_relations, rel_dim, device)
+
+        self.rel_dim = rel_dim
+        self.rel_part = args.rel_part
+        if not self.rel_part:
+            self.relation_emb = ExternalEmbedding(args, n_relations, rel_dim, F.cpu() if args.mix_cpu_gpu else device)
+        else:
+            self.global_relation_emb = ExternalEmbedding(args, n_relations, rel_dim, F.cpu())
```
zheng-da: is `n_relations` the total number of relations in the graph?

apps/kg/models/pytorch/tensor_models.py
```diff
@@ -23,7 +23,7 @@
 logsigmoid = functional.logsigmoid
 
 def get_device(args):
-    return th.device('cpu') if args.gpu < 0 else th.device('cuda:' + str(args.gpu))
+    return th.device('cpu') if args.gpu[0] < 0 else th.device('cuda:' + str(args.gpu[0]))
```
zheng-da: this function seems to not work correctly since we have multiple devices now.

apps/kg/train_pytorch.py
```diff
@@ -29,13 +53,22 @@ def load_model_from_checkpoint(logger, args, n_entities, n_relations, ckpt_path)
     model.load_emb(ckpt_path, args.dataset)
     return model
 
-def train(args, model, train_sampler, valid_samplers=None):
+@thread_wrapped_func
+def train(args, model, train_sampler, rank=0, rel_parts=None, valid_samplers=None):
     if args.num_proc > 1:
-        th.set_num_threads(1)
+        th.set_num_threads(8)
```
zheng-da: multithreading works now?

zheng-da:
It seems that each GPU has a local copy of the entire relation embedding and copy the trained embedding back to the global embedding in the end.
It has two problems:
* it will waste a lot of memory for models like TransR.
* in Freebase, there are relations that are so frequent that we have to split the relation into multiple partitions. We have to frequently write popular relation embeddings back to the global embeddings.

apps/kg/train_pytorch.py
```diff
@@ -80,16 +113,29 @@ def train(args, model, train_sampler, valid_samplers=None):
             test(args, model, valid_samplers, mode='Valid')
```
zheng-da: the relation embeddings need to write back here as well.

apps/kg/train_pytorch.py
classicsong: Yes

apps/kg/models/pytorch/tensor_models.py
classicsong: For multiple gpu, mix_cpu_gpu should be used and get_device is not used at that time.

apps/kg/models/mxnet/tensor_models.py
```diff
@@ -71,9 +71,9 @@ def update(self):
                 grad_sum = grad_sum.as_in_context(ctx)
             self.state_sum[grad_indices] += grad_sum
             std = self.state_sum[grad_indices]  # _sparse_mask
-            std_values = nd.expand_dims(nd.sqrt(std) + 1e-10, 1)
-            if self.gpu >= 0:
-                std_values = std_values.as_in_context(mx.gpu(self.args.gpu))
+            if gpu_id >= 0:
+                std = std.as_in_context(mx.gpu(gpu_id))
+                std_values = nd.expand_dims(nd.sqrt(std) + 1e-10, 1)
```
zheng-da: why do you move the code here? shouldn't we call sqrt and expand_dims regardless of the device?

apps/kg/models/pytorch/tensor_models.py
```diff
@@ -85,9 +85,9 @@ def update(self):
                     grad_sum = grad_sum.to(device)
                 self.state_sum.index_add_(0, grad_indices, grad_sum)
                 std = self.state_sum[grad_indices]  # _sparse_mask
-                std_values = std.sqrt_().add_(1e-10).unsqueeze(1)
-                if self.gpu >= 0:
-                    std_values = std_values.cuda(self.args.gpu)
+                if gpu_id >= 0:
+                    std = std.cuda(gpu_id)
+                    std_values = std.sqrt_().add_(1e-10).unsqueeze(1)
```
zheng-da: the same question here.

Issue #1190, opened 2020-01-10
Creating dgl.bipartite graph with the given node_id > card range will cause crashing


classicsong:
## To Reproduce
```python
import dgl
import torch as th
import dgl.function as fn

num_user=10
num_item=10

rrow = [1,2,10]
rcol = [3,5,10]

bg = dgl.bipartite(
  (rrow, rcol),
  'user',
  'like',
  'item',
  card=(num_user, num_item)
)

bg.nodes['user'].data['h'] = th.randn((10,5))
bg.nodes['item'].data['h'] = th.randn((10,5))

funcs={}
funcs['like'] = (fn.u_add_v('h', 'h', 'm'), fn.sum('m', 'hh'))
bg.multi_update_all(funcs, 'stack')
```

## Expected behavior

An error is report during bipartite graph init.

## Environment

 - DGL Version (e.g., 1.0): master
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):
 - OS (e.g., Linux):
 - How you installed DGL (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version (if applicable):
 - GPU models and configuration (e.g. V100):
 - Any other relevant information:

PR #1192, opened 2020-01-10
[Bug Fix] Add boundary check for heterograph build with card as input

classicsong:
## Description
Building graph using dgl.bipartite or dgl.graph with both **data** and **card** set may crash when the node id provied in **data** is larger than the size specified in **card**.

Add boundary check for both dgl.bipartite and dgl.graph creation.

#1190 

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/convert.py
+++ b/python/dgl/convert.py
@@ -1011,7 +1011,8 @@ def create_from_networkx_bipartite(nx_graph,
                 dst.append(bottom_map[e[1]])
     src = utils.toindex(src)
     dst = utils.toindex(dst)
-    g = create_from_edges(src, dst, utype, etype, vtype,
+    g = create_from_edges(
+        src, dst, utype, etype, vtype,
         len(top_nodes), len(bottom_nodes), validate=False)
 
     # TODO attributes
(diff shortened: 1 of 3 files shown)


python/dgl/convert.py
```diff
@@ -746,6 +746,13 @@ def create_from_edges(u, v, utype, etype, vtype, urange=None, vrange=None, valid
     vrange = vrange or (
         0 if len(v) == 0 else (int(F.asnumpy(F.max(v.tousertensor(), dim=0))) + 1))
 
+    if len(u) > 0:
+        assert urange > int(F.asnumpy(F.max(u.tousertensor(), dim=0))), \
+            "The urange from card should be larger than max u node_id"
+    if len(v) > 0:
+        assert vrange > int(F.asnumpy(F.max(v.tousertensor(), dim=0))), \
+            "The vrange from card should be larger than max v node_id"
+
```
jermainewang: Are these checks the same as L738-743? Can we merge them?

classicsong: I c. We can set validate=True to turn on the check by default. But for some internal cases, we can manually turn it off.

Issue #1199, opened 2020-01-13
[RFC] 0.5 Sampler Design for Heterographs


BarclayII:
## 🚀 Feature
We have finally settled down (in principle) the design of the long-awaited sampling on heterogeneous graphs, and we have started working on it.

This issue keeps track of the implementation progress.

**You're welcome to read the proposal publicly available on https://docs.google.com/document/d/1cVvbM_QtaIN0UsqrgB0FhlYti1PjXua2QbSAoMaE2ng/edit**

The document is not fixed and it is subject to changes: you're welcome to provide feedback on either the proposal itself or this issue thread.

## Task tracker

- [ ] C APIs
  - [ ] Neighbor sampling [@jermainewang ]
    - [x] sample_neighbors
    - [x] sample_neighbors_topk
    - [x] in_subgraph
    - [x] out_subgraph
    - [ ] sample_layerwise
  - [ ] Random walks
    - [x] random_walk [@BarclayII ]
    - [x] random_walk_with_restart [@BarclayII ]
    - [ ] node2vec_random_walk
  - [ ] Misc
    - [ ] `dgl::aten` refactor to cover float types as well (only necessary ones)
    - [x] create_graph_from_paths [@BarclayII ]
    - [x] compact_graphs [@BarclayII ]
    - [x] to_simple_graph [@BarclayII ]
    - [x] dgl.choice(prob_array, num_samples) (faster numpy.random.choice) [@yzh119, already done in #1142 , may a python interface is required? ]
    - [x] Graph.remove_edges() [@BarclayII ]
- [ ] Models (rewrite)
  - [ ] GCN [@jermainewang ]
  - [ ] GCN with Control Variate Sampling
  - [x] GraphSAGE (with distributed) [@jermainewang ]
  - [ ] GraphSAGE with Control Variate Sampling
  - [ ] PinSAGE (with distributed) [@BarclayII ]
  - [ ] GCMC [@jermainewang ]
  - [x] RGCN? [@jermainewang ]
  - [ ] GAT?
  - [ ] FastGCN?
  - [ ] DGI?
  - [ ] Adaptive Layerwise Sampling?

Related:
* Unification of DGLGraph, DGLHeteroGraph, SubGraph, BatchedGraph, etc.
* Shared memory storage

BarclayII:
@yzh119 I'm thinking of putting the `dgl.choice` as `RandomEngine::ThreadLocal()->Choice(FloatArray unnormalized_prob)`

PR #1209, opened 2020-01-17
[Feature] Random Walk for 0.5

BarclayII:
## Description
This implements fixed-length random walk operator described in #1199 .

Although the branch name is called `pinsage-refactor`, this is only the very first step towards actually refactoring the PinSAGE implementation.

## Checklist
  - [x] API
  - [x] Migrate examples from existing `random_walk` calls to the new one.
- [x] Code is well-documented

## Misc Changes
* New C++ NDArray operator `Assign(array, index)` and `IsEmpty(array)`.
* New NDArray constructor `VecToNDArray` that works with int32/64 and float32/64.
* `EXPECT` macros to perform array type checks.
* `HeteroGraph::GetEndpointTypes(dgl_type_t etype)` to get node types of both endpoints given the edge type.
* `RandomEngine::Choice(FloatArray array)` that does non-uniform sampling using `TreeSampler`.


BarclayII:
@jermainewang I tried replacing `SuccVec` and `OutEdgeVec` with `OutEdges` and `Successors`, and tested the implementations against metapath2vec sampler in the examples (removing the code for writing the paths to a file).  Unfortunately this led to a slow down of at least 3x:

| Implementation | Iterations per second |
|:-----------:|:---------|
| This one | 48.56 |
| This one + OpenMP | 63.51 |
| `hg->Successors()` | 14.31 |
| `hg->Successors()` + OpenMP | 22.19 |
| `hg->OutEdges()` | 4.50 |

For one iteration I'm generating 1000 paths from a single node with a metapath of 100 'C -> P -> A -> P -> C'.

jermainewang:
> @jermainewang I tried replacing `SuccVec` and `OutEdgeVec` with `OutEdges` and `Successors`, and tested the implementations against metapath2vec sampler in the examples (removing the code for writing the paths to a file). Unfortunately this led to a slow down of at least 3x:

Are you talking about `randomwalks_cpu.cc:80`? If I understand it correctly, `OutEdges` and `Successors` return a new NDArray by memory copy, so it is expected to be much slower. If the goal is to avoid using the `SuccVec` interface, how about directly compute random walk on the CSR arrays from `hg`?

include/dgl/sample_utils.h
```diff
@@ -49,7 +49,8 @@ class AliasSampler: public BaseSampler<Idx, DType, replace> {
   DType accum, taken;             // accumulated likelihood
   std::vector<Idx> K;             // alias table
   std::vector<DType> U;           // probability table
-  std::vector<DType> _prob;       // category distribution
+  const DType *_prob;             // category distribution
```
yzh119: Could you please explain why using a pointer here?

BarclayII:
> If I understand it correctly, `OutEdges` and `Successors` return a new NDArray by memory copy, so it is expected to be much slower.

No.  They all return views on the array.

> If the goal is to avoid using the `SuccVec` interface, how about directly compute random walk on the CSR arrays from `hg`?

K.

include/dgl/array.h
```diff
@@ -285,6 +315,150 @@ IdArray VecToIdArray(const std::vector<T>& vec,
...
+NDArray VecToNDArray(const std::vector<T> &vec, DLDataType dtype, DLContext ctx) {
+  int64_t size = static_cast<int64_t>(vec.size());
+  NDArray ret = NDArray::Empty({size}, dtype, DLContext{kDLCPU, 0});
+  std::copy(vec.begin(), vec.end(), static_cast<T *>(ret->data));
+  return ret.CopyTo(ctx);
+}
+
+///////////////////////// Dispatchers //////////////////////////
+
+#define ATEN_XPU_SWITCH(val, XPU, ...) do {                     \
```
yzh119: These dispatch macros look exactly like [pytorch aten dispatcher](https://github.com/pytorch/pytorch/blob/985fd970aad85799a568b76c2a1b75a1c422fe53/aten/src/ATen/Dispatch.h). Is there any possibility we depend on ATen in the future?

include/dgl/sample_utils.h
BarclayII: I was considering using NDArrays but right now I'm going to directly compute random walk on CSR arrays anyway.  So I'm reverting it.

include/dgl/array.h
BarclayII: Does ATen work with MXNet?

BarclayII:
> > If the goal is to avoid using the `SuccVec` interface, how about directly compute random walk on the CSR arrays from `hg`?
> 
> K.

Turns out that directly computing random walk from `GetAdj()` is still twice as slow than using `SuccVec` for some reason.

include/dgl/array.h
```diff
@@ -121,6 +148,19 @@ inline bool IsValidIdArray(const dgl::runtime::NDArray& arr) {
   return arr->ndim == 1 && arr->dtype.code == kDLInt;
 }
 
+/*!
+ * \brief Pack a padded tensor to a triplet of concatenated tensor, their lengths,
+ * and offsets.
```
jermainewang: This looks like a quite complicated API. Could you describe more about the arguments and returns? Give an example in the docstring will be even better.

include/dgl/array.h
```diff
@@ -121,6 +148,19 @@ inline bool IsValidIdArray(const dgl::runtime::NDArray& arr) {
...
+ */
+template<typename ValueType>
+std::tuple<NDArray, IdArray, IdArray> Pack(NDArray array, ValueType pad_value);
+
+/*!
+ * \brief Concat array[i, 0:lengths[i]] for every i.  Returns the concatenated array
+ * and offsets in the concatenated array for each original slice.
```
jermainewang: Take me a while to understand the docstring. Does it mean that the input `array` must be a 2D matrix? To me, it looks like a batched version of slice. Maybe `BatchSlice` is a more proper name? Also, please write more docstring and perhaps an example.

src/graph/sampling/get_node_types_cpu.cc
```diff
@@ -0,0 +1,67 @@
...
+  dgl_type_t etype = metapath_data[0];
+  dgl_type_t srctype = hg->GetEndpointTypes(etype).first;
+  dgl_type_t curr_type = srctype;
+  result_data[0] = curr_type;
+
+  for (uint64_t i = 0; i < num_etypes; ++i) {
+    etype = metapath_data[i];
+    auto src_dst_type = hg->GetEndpointTypes(etype);
+    dgl_type_t srctype = src_dst_type.first;
+    dgl_type_t dsttype = src_dst_type.second;
```
jermainewang: Use local variable for `etype`. `srctype` shadows another variable outside of the loop. I will shorten L32-L34 as:
```
dgl_type_t curr_type = hg->GetEndpointTypes(metapath_data[0]).first;
```

src/graph/sampling/metapath_randomwalk.h
```diff
@@ -0,0 +1,142 @@
...
+using namespace dgl::runtime;
+using namespace dgl::aten;
+
+namespace sampling {
+
+namespace impl {
+
+namespace {
+
+using TerminatePredicate = std::function<bool(void *, dgl_id_t, int64_t)>;
```
jermainewang: Brief description about the argument and return meanings.

tests/cpp/test_aten.cc
```diff
@@ -144,7 +144,7 @@ TEST(ArrayTest, TestHStack) {
 template <typename IDX>
 void _TestIndexSelect() {
   IdArray a = aten::Range(0, 100, sizeof(IDX)*8, CTX);
-  ASSERT_EQ(aten::IndexSelect(a, 50), 50);
+  ASSERT_EQ(aten::IndexSelect<int>(a, 50), 50);
```
jermainewang: How about the other array operators you added? Have you tested them?

include/dgl/array.h
```diff
@@ -102,8 +115,21 @@ BoolArray LT(IdArray lhs, dgl_id_t rhs);
...
 
 /*! \brief Return the data under the index. In numpy notation, A[I] */
-int64_t IndexSelect(IdArray array, int64_t index);
-IdArray IndexSelect(IdArray array, IdArray index);
+template<typename ValueType>
+ValueType IndexSelect(NDArray array, uint64_t index);
+NDArray IndexSelect(NDArray array, IdArray index);
+
+/*! \brief Check if an array is empty (i.e. have 0 elements) */
+inline bool IsEmpty(NDArray array) {
```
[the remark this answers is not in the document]
BarclayII: I missed `GetSize()`.  Good point.

include/dgl/array.h
```diff
@@ -285,6 +311,150 @@ IdArray VecToIdArray(const std::vector<T>& vec,
   return ret.CopyTo(ctx);
 }
 
+template<typename T>
+NDArray VecToNDArray(const std::vector<T> &vec, DLDataType dtype, DLContext ctx) {
+  int64_t size = static_cast<int64_t>(vec.size());
+  NDArray ret = NDArray::Empty({size}, dtype, DLContext{kDLCPU, 0});
+  std::copy(vec.begin(), vec.end(), static_cast<T *>(ret->data));
+  return ret.CopyTo(ctx);
+}
```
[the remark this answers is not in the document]
BarclayII: I was copying what `VecToIdArray` is doing.  Putting it as static method looks great.  I'll do this.

include/dgl/array.h
```diff
@@ -285,6 +311,150 @@ IdArray VecToIdArray(const std::vector<T>& vec,
...
+  ((a)->dtype.code == kDLInt && (a)->dtype.bits == 32)
+#define IS_INT64(a)  \
+  ((a)->dtype.code == kDLInt && (a)->dtype.bits == 64)
+#define IS_FLOAT32(a)  \
+  ((a)->dtype.code == kDLFloat && (a)->dtype.bits == 32)
+#define IS_FLOAT64(a)  \
+  ((a)->dtype.code == kDLFloat && (a)->dtype.bits == 64)
+
+#define EXPECT(cond, prop, value_name, dtype_name) \
+  CHECK(cond) << "Expecting " << (prop) << " of " << (value_name) << " to be " << (dtype_name)
```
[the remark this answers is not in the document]
BarclayII: Indeed.  I'd rather rename them though.

python/dgl/_ffi/function.py
```diff
@@ -286,14 +286,16 @@ def _init_api_prefix(module_name, prefix):
...
-        target_module = module
 
-        if fname.find(".") != -1:
-            print('Warning: invalid API name "%s".' % fname)
+        if len(name_split) == 1:
+            print('Warning: invalid API name "%s".' % name)
             continue
+        fname = name_split[1]
+        target_module = module
+
```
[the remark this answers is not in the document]
BarclayII: If we have two functions registered as `dgl.sampling.random.foo` and `dgl.sampling.randomwalk.bar`, then when finding functions for module `dgl.sampling.random` it will report an error on `dgl.sampling.randomwalk.bar`.

include/dgl/random.h
```diff
@@ -95,8 +97,14 @@ class RandomEngine {
     return dist(rng_);
   }
 
+  /*!
+   * \brief Pick a random integer between 0 to N-1 according to given probabilities
+   */
+  template<typename IdxType, typename ValueType>
+  IdxType Choice(const std::vector<ValueType> &prob);
```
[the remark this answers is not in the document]
BarclayII: I will switch to NDArray.

python/dgl/sampling/__init__.py
```diff
@@ -0,0 +1,2 @@
+"""Sampler modules."""
+from .randomwalks import *
```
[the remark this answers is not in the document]
BarclayII: I think random walks indeed belong to sampling algorithms.

tests/cpp/test_aten.cc
BarclayII: They are tested in python function `test_pack_traces()`

src/graph/sampling/randomwalks_cpu.h
```diff
@@ -0,0 +1,75 @@
...
+ * \param hg The heterograph.
+ * \param seeds A 1D array of seed nodes, with the type the source type of the first
+ *        edge type in the metapath.
+ * \param max_num_steps The maximum number of steps of a random walk path.
+ * \param step The random walk step function with type \c StepFunc.
+ * \return A 2D array of shape (len(seeds), max_num_steps + 1) with node IDs.
+ */
+template<DLDeviceType XPU, typename IdxType>
+IdArray GenericRandomWalk(
+    const HeteroGraphPtr hg,
```
jermainewang: `hg` is not used in this function. I guess it is bound in the closure of `StepFunc`. Should we remove this arg?

src/graph/sampling/metapath_randomwalk.h
```diff
@@ -0,0 +1,147 @@
...
+  int64_t max_num_steps = metapath->shape[0];
+  const IdxType *metapath_data = static_cast<IdxType *>(metapath->data);
+
+  // Prefetch all edges.
+  // This forces the heterograph to materialize all OutCSR's before the OpenMP loop;
+  // otherwise data races will happen.
+  // TODO(BarclayII): should we later on materialize COO/CSR/CSC anyway unless told otherwise?
+  std::vector<std::vector<IdArray> > edges_by_type;
+  for (dgl_type_t etype = 0; etype < hg->NumEdgeTypes(); ++etype)
+    edges_by_type.push_back(hg->GetAdj(etype, true, "csr"));
```
jermainewang: Will you fix this in PR #1224 ?

src/graph/sampling/metapath_randomwalk.h
```diff
@@ -0,0 +1,147 @@
...
+ * \param edges_by_type Vector of results from \c GetAdj() by edge type.
+ * \param metapath_data Edge types of given metapath.
+ * \param prob Transition probability per edge type.
+ * \param terminate Predicate for terminating the current random walk path.
+ *
+ * \return A pair of ID of next successor (-1 if not exist), as well as whether to terminate.
+ */
+template<DLDeviceType XPU, typename IdxType>
+std::pair<dgl_id_t, bool> MetapathRandomWalkStep(
+    void *data,     // void * because std::function does not accept template typenames.
```
jermainewang: Can't it be `IdxType*`?

src/graph/sampling/metapath_randomwalk.h
```diff
@@ -0,0 +1,147 @@
...
+  // uniform case), we want to reduce the overheads (even from object copies or object
+  // construction) as much as possible.
+  // Using Successors() slows down by 2x.
+  // Using OutEdges() slows down by 10x.
+  const auto &csr_arrays = edges_by_type[etype];
+  const IdxType *offsets = static_cast<IdxType *>(csr_arrays[0]->data);
+  const IdxType *all_succ = static_cast<IdxType *>(csr_arrays[1]->data);
+  const IdxType *succ = all_succ + offsets[curr];
+
+  int64_t size = offsets[curr + 1] - offsets[curr];
```
jermainewang: const

src/graph/sampling/metapath_randomwalk.h
```diff
@@ -0,0 +1,147 @@
...
+    // non-uniform random walk
+    const IdxType *all_eids = static_cast<IdxType *>(csr_arrays[2]->data);
+    const IdxType *eids = all_eids + offsets[curr];
+
+    ATEN_FLOAT_TYPE_SWITCH(prob_etype->dtype, DType, "probability", {
+      FloatArray prob_selected = FloatArray::Empty({size}, prob_etype->dtype, prob_etype->ctx);
+      DType *prob_selected_data = static_cast<DType *>(prob_selected->data);
+      const DType *prob_etype_data = static_cast<DType *>(prob_etype->data);
+      for (int64_t j = 0; j < size; ++j)
+        prob_selected_data[j] = prob_etype_data[eids[j]];
```
jermainewang: Is this an `IndexSelect` operation?

include/dgl/array.h
```diff
@@ -121,6 +129,68 @@ inline bool IsValidIdArray(const dgl::runtime::NDArray& arr) {
...
+ *
+ * If a 1D array is given, then the function is equivalent to
+ *
+ * <code>
+ *     def ConcatSlices(array, lengths):
+ *         slices = [array[:l] for l in lengths]
+ *         packed = np.concatenate(slices)
+ *         offsets = np.cumsum([0] + lengths[:-1])
+ *         return packed, offsets
+ * </code>
```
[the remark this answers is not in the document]
BarclayII: When finding the node type IDs for each random walk path with padding, for instance.

src/graph/sampling/metapath_randomwalk.h
BarclayII: Materializing the CSR arrays beforehand is much faster.

src/runtime/ndarray.cc
```diff
@@ -223,6 +223,46 @@ void NDArray::CopyFromTo(DLTensor* from,
...
+      0,
+      static_cast<T*>(ret->data),
+      0,
+      size * sizeof(T),
+      DLContext{kDLCPU, 0},
+      ctx,
+      dtype,
+      nullptr);
+  return ret;
+}
```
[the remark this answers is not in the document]
BarclayII: The implementation requires `DeviceAPI` from `device_api.h`, which includes `packed_func.h`, which includes `ndarray.h`.  So putting this in `ndarray.h` doesn't work.  Should I put it in another header file?

src/graph/sampling/metapath_randomwalk.h
BarclayII: I'm afraid not, because the first argument of `TerminatePredicate` and `StepFunc` must be `void *`.

src/graph/sampling/metapath_randomwalk.h
BarclayII: Using `CreateView` and `IndexSelect` slows it down by 30%.

BarclayII: Here is what I tried to replace the code above with:
```C++
    // XXX CreateView() mutates the (refcount of the) original array but here we have a const.
    const IdArray eids = const_cast<IdArray &>(csr_arrays[2]).CreateView(
        {offsets[curr + 1] - offsets[curr]},
        csr_arrays[2]->dtype,
        offsets[curr] * sizeof(IdxType));
    const FloatArray prob_selected = std::move(IndexSelect(prob_etype, eids));
    idx = RandomEngine::ThreadLocal()->Choice<IdxType>(prob_selected);
```

src/graph/sampling/randomwalks.cc
```diff
@@ -0,0 +1,185 @@
...
+    List<Value> prob = args[3];
+    FloatArray restart_prob = args[4];
+
+    std::vector<FloatArray> prob_vec;
+    prob_vec.reserve(prob.size());
+    for (Value val : prob)
+      prob_vec.push_back(val->data);
+
+    auto result = sampling::RandomWalkWithRestart(
+        hg.sptr(), seeds, metapath, prob_vec, restart_prob);
```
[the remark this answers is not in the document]
BarclayII: Renamed to `RandomWalkWithStepwiseRestart`.

src/graph/sampling/metapath_randomwalk.h
jermainewang: Would [type alias](https://zh.cppreference.com/w/cpp/language/type_alias) help here?

src/runtime/ndarray.cc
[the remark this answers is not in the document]
jermainewang: I see. Then let's stick with your solution.

src/graph/sampling/metapath_randomwalk.h
jermainewang: I c. Let's use your solution. BTW, the const cast could be avoided by copying out `csr_arrays[2]`, which is internally a shallow copy.

src/graph/sampling/metapath_randomwalk.h
BarclayII: NVM.  The comment above was a mistake.

Issue #1153, opened 2019-12-30
[RFC] Unify the interface of BatchedDGLGraph, DGLSubGraph and DGLGraph


yzh119:
## 🚀 Feature

This is the first step towards a unified graph interface(the next step is to merge HeteroDGLGraph and DGLGraph).

## Motivation

Currently we have several graph classes: HeteroGraph, BatchedHeteroGraph, BatchedDGLGraph, DGLSubGraph and DGLGraph, it's difficult to maintain so many seperated graph structures and make the interface consistent. (e.g. #1146 ).

## Pitch

A much cleaner interface like [torch.tensor](https://pytorch.org/docs/stable/torch.html#torch.tensor) to replace DGLSubGraph and BatchedDGLGraph. The refactor could both ease the developers' workload and make the interface easier for users.

## Additional context

The SubDGLGraph class and BatchedDGLGraph should not be deprecated soon, and we should redirect them to `dgl.graph(is_subgraph=True)` and `dgl.graph(batch_size=xx)`.

Pay attention to the use of `isinstance(g, SubDGLGraph)` and `instance(g, BatchedDGLGraph)` and replace them with DGLGraph class methods.

jermainewang:
Would you mind elaborating more about the new APIs and behaviors of DGLGraph when they are merged? For example, I'd like to see a one-one correspondence between the old APIs and the new ones.

mufeili:
# Data Structure Comparison
## `DGLGraph` vs `BatchedDGLGraph`
- `BatchedDGLGraph` has three additional internal attributes `_batch_size`, `_batch_num_nodes` and `_batch_num_edges` and three associated class functions `batch_size`, `batch_num_nodes` and `batch_num_edges`.
- `BatchedDGLGraph` is readonly. Therefore mutations like `add_nodes`, `add_edge`, `add_edges`, `remove_nodes`, `remove_edges` and `clear` should be forbidden in these cases. Currently, these are not well handled.
- Two class functions to be implemented for `BatchedDGLGraph`: `__getitem__(self, idx)` and `__setitem__(self, idx, val)`.

## `DGLGraph` vs `DGLSubGraph` 
- `DGLSubGraph` has four additional internal attributes `_parent`, `_parent_nid`, `_parent_eid`, `_subgraph_index` and 6 associated class functions `parent_nid`, `_get_parent_eid`, `parent_eid`, `copy_to_parent`, `copy_from_parent` and `map_to_subgraph_nid`.
[...]

Issue #1118, opened 2019-12-19
[RFC] Mutation support for Heterograph. Refactor MutableGraph and ImmutableGraph


jermainewang:
## 🚀 Feature

This is part of the refactoring effort to our core graph data structure.
* Add mutation support using COO format.
  * Implement `add_nodes` by extending the node feature storage (node frame) and padding with zero values.
  * Implement `add_edges` by extending and padding the edge feature storage (edge frame). New edges are appended to the COO storage.
  * Implement `remove_nodes` and `remove_edges` by `subgraph` and `edge_subgraph`.

As such, we don't need MutableGraph based on adjacency list anymore. It also paves the way for storing the graph structure entirely in GPU.

### Implementation

The major effort comes from re-implementing algorithms based on adjacency list with CSR or COO format. Examples are:
* Graph traversal algorithms such as BFS, DFS.
* Batching/unbatching

PR #1216, opened 2020-01-19
[Refactor] Unify DGLGraph, BatchedDGLGraph and DGLSubGraph

yzh119:
## Description
Implement the proposal described in #1153 .

Some behavior changes:
1. readout functions (and together with global pooling modules) now returns a `(1, D)` as single graph is now only a special case of batched graph (according to my experience this would make the program simpler).
2. now we supports subgraph of subgraphs, but there are some missing features:
  * subgraphs of batched graphs.
  * batched graph of subgraphs.
The semantics need to be carefully designed.

@jermainewang , I found out that the subgraph index is by default mutable (because we also use this function in remove nodes/edges, where mutation is allowed as user may want to add nodes/edges after removing some nodes/edges), and we force user not to change the graph structure by overloading add_node(s), add_edge(s) function. If we print `subg.is_readonly` it's still `False`. Could we allow user to specify whether to make the subgraph structure readonly on the C++ side?

## Checklist
- [x] Code is well-documented

## Changes
- dgl.graph.py
- dgl.subgraph.py
- dgl.batched_graph.py
- dgl.transform.py
- dgl.nn.*.glob.py
- dgl.model_zoo.chem.classifier.py
- ...


yzh119:
@mufeili , please notify me about related docstring changes I need to make.

mufeili:
@yzh119 We decide not to include this in v0.4.2 since @jermainewang is refactoring the documentation for BatchedDGLGraph. I will take a look.

jermainewang:
> Some behavior changes:
> 
> 1. readout functions (and together with global pooling modules) now returns a `(1, D)` as single graph is now only a special case of batched graph (according to my experience this would make the program simpler).

Do you mean readout functions will return tensor of shape `(1, D)` when the input batch contains only one graph?

> 2. now we supports subgraph of subgraphs, but there are some missing features:
> 
> * subgraphs of batched graphs.
> * batched graph of subgraphs.
>   The semantics need to be carefully designed.

For subgraphs of batched graphs, a more general question is for any DGLGraph g with batch_size > 1, does our API `gg = dgl.some_api(g)` keep the batch size? My proposal is to keep flexible here. For example, many operators support operating on a batch of graphs (e.g., `dgl.line_graph`). Other operators like `subgraph` could just squeeze the batch size to one. Another question is that some operators such as neighbor sampling can be parallelized along batch dimension so the question is whether we'd like to support batch dimension in their arguments too? I think we should stay simple at the moment and optimize only when needed.

For batching subgraphs, what's the problem here?

> @jermainewang , I found out that the subgraph index is by default mutable (because we also use this function in remove nodes/edges, where mutation is allowed as user may want to add nodes/edges after removing some nodes/edges), and we force user not to change the graph structure by overloading add_node(s), add_edge(s) function. If we print `subg.is_readonly` it's still `False`. Could we allow user to specify whether to make the subgraph structure readonly on the C++ side?

In the future, I suggest remove the `readonly` flag in DGLGraph. DGLGraph should allow mutation at all time. See RFC #1118 .

mufeili:
1. +1 for the suggested behavior change of readout functions. This allows models performing graph-level prediction to work with both single and batched graphs without the need of `g.unsqueeze(0)` for single graphs.
2. Why do we still have a `BatchedDGLGraph` class? I suppose `dgl.batch` should return a `DGLGraph` directly with no wrapping? For deprecation concerns, we only need to leave a pointer `BatchedDGLGraph = DGLGraph`. Same for `DGLSubgraph`.
3. Do we still need ` _subgraph_index` and `parent`? @jermainewang previously suggested removing them, which will make batching subgraphs no longer an issue. Related functions like `copy_to_parent` can be made stand alone. @zheng-da seems to be fine with it. Copy him for visibility.
4. Will it be better to make `__parent_nid`, `_parent_eid` separately members of ndata and edata? Also, probably they should be assigned only in functions creating subgraphs.
5. The example of clustergcn accidentally uses `dgl.BatchedDGLGraph` directly. Could you please also fix [that](https://github.com/dmlc/dgl/blob/master/examples/pytorch/cluster_gcn/utils.py#L70)?

yzh119:
@jermainewang 
Reply to point 1: yes
Reply to point 2: we should not restrict `dgl.some_api(g)` to have the same batch size as `g`, it depends on the functionality of this api and I agree with you on keeping flexibilities. For batching subgraphs, how do we define the parent graph of a batched subgraphs?
Reply to point 3: Thanks for letting me know, I can start working on #1118 after this PR being merged.

@mufeili 
Reply to point 2: The BatchedGraph class should and must be deprecated. Your suggestions make sense and I'll try.
Reply to point 3: What do you mean by "Related functions like copy_to_parent can be made stand alone"?
Reply to point 4: 
> Will it be better to make __parent_nid, _parent_eid separately members of ndata and edata?
Could you please elaborate more?
Reply to point 5: TODO.

mufeili:
> Reply to point 3: What do you mean by "Related functions like copy_to_parent can be made stand alone"?

For example, `g.copy_to_parent` will become `dgl.copy_to_parent` and users need to keep the reference to parent graphs themselves.

> Reply to point 4:
> 
> > Will it be better to make __parent_nid, _parent_eid separately members of ndata and edata?
> 
> Could you please elaborate more?

For example `g.ndata[__parent_nid]` will store the current `g.__parent_nid` just like how we store raw ids/types in `dgl.to_homo`.

python/dgl/batched_graph.py
```diff
@@ -343,15 +87,90 @@ def batch(graph_list, node_attrs=ALL, edge_attrs=ALL):
...
+    for grh in graph_list:
+        # handle the input is again a batched graph.
+        batch_size += grh.batch_size
+        batch_num_nodes += grh.batch_num_nodes
+        batch_num_edges += grh.batch_num_edges
+
+    return DGLGraph(graph_data=batched_index,
+                    node_frame=batched_node_frame,
+                    edge_frame=batched_edge_frame,
+                    batch_size=batch_size,
```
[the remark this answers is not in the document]
jermainewang: This also raises a question: shall we add a `g.flatten()` function?

python/dgl/graph.py
```diff
@@ -902,7 +902,12 @@ def __init__(self,
                  edge_frame=None,
                  multigraph=None,
                  readonly=False,
-                 sort_csr=False):
+                 sort_csr=False,
+                 batch_size=1,
```
jermainewang: Again, this argument could be removed.

python/dgl/graph.py
```diff
@@ -902,7 +902,12 @@ def __init__(self,
...
+                 batch_num_nodes=None,
+                 batch_num_edges=None,
+                 parent=None,
+                 sgi=None):
```
jermainewang: Why `sgi`?

python/dgl/graph.py
```diff
@@ -936,6 +941,20 @@ def __init__(self,
...
+        self._batch_num_edges = batch_num_edges
+
+        # subgraph
+        self._parent = parent
+        self._parent_nid = None
+        self._parent_eid = None
+        self._subgraph_index = sgi
+        if sgi is not None:
+            self._parent_nid = sgi.induced_nodes
+            self._parent_eid = sgi.induced_edges
```
jermainewang: Shouldn't need `self._parent_nid`, `self._parent_eid` and `self._subgraph_index` members. Mappings are saved under `dgl.NID` and `dgl.EID` fields. Please refer to the implementation of `dgl.HeteroGraph.node_subgraph`.

python/dgl/graph.py
```diff
@@ -988,6 +1007,9 @@ def add_nodes(self, num, data=None):
                 [1., 1., 1., 1.],
                 [1., 1., 1., 1.]])
         """
+        if self.is_readonly:
+            raise DGLError("Readonly graph. Mutation is not allowed.")
+
```
jermainewang: Remember to modify `batch_num_nodes` and `batch_num_edges`. The simplest way is to flatten the graph.

python/dgl/graph.py
```diff
@@ -1244,6 +1272,85 @@ def remove_edges(self, eids):
...
+        parent_vids : list, tensor
+            The node ID array in the parent graph.
+
+        Returns
+        -------
+        tensor
+            The node ID array in the subgraph.
+        """
+        v = graph_index.map_to_subgraph_nid(self._subgraph_index, utils.toindex(parent_vids))
+        return v.tousertensor()
```
jermainewang: Should this be a global function?

python/dgl/graph.py
```diff
@@ -1244,6 +1272,85 @@ def remove_edges(self, eids):
...
+
+        The returned tensor can be used as a map from the edge id
+        in this subgraph to the edge id in the parent graph.
+
+        Returns
+        -------
+        Tensor
+            The parent edge id array.
+        """
+        return self._get_parent_eid().tousertensor()
```
jermainewang: I don't see the value of `parent_nid` and `parent_eid`. They will be replaced by:
```python
g.ndata[dgl.NID]
g.edata[dgl.EID]
```

jermainewang: And if a `DGLGraph` will hold a reference to its parent graph, consider add a `parent` property.

python/dgl/graph.py
```diff
@@ -1710,6 +1817,41 @@ def edata(self):
...
+        """Number of nodes of each graph in this batch.
+
+        Returns
+        -------
+        list
+            Number of nodes of each graph in this batch."""
+        if self._batch_num_nodes is None:
+            return [self.number_of_nodes()]
+        else:
+            return self._batch_num_nodes
```
jermainewang: Put `if ... is None` to constructor.

python/dgl/graph.py
```diff
@@ -1710,6 +1817,41 @@ def edata(self):
...
+    @property
+    def batch_num_edges(self):
+        """Number of edges of each graph in this batch.
+
+        Returns
+        -------
+        list
+            Number of edges of each graph in this batch."""
+        if self._batch_num_edges is None:
+            return [self.number_of_edges()]
```
jermainewang: ditto

python/dgl/subgraph.py
```diff
@@ -40,102 +37,8 @@ class DGLSubGraph(DGLGraph):
     """
     def __init__(self, parent, sgi, shared=False):
         super(DGLSubGraph, self).__init__(graph_data=sgi.graph,
-                                          readonly=True)
+                                          readonly=True,
+                                          parent=parent,
+                                          sgi=sgi)
```
jermainewang: Could you help remove `shared`? Please also modify the docstring accordingly. And why do we still have this class?

python/dgl/transform.py
```diff
@@ -291,7 +291,7 @@ def reverse(g, share_ndata=False, share_edata=False):
             [2.],
             [3.]])
     """
-    assert not isinstance(g, BatchedDGLGraph), \
+    assert g.batch_size == 1, \
```
jermainewang: Could support instead? just set the same `batch_num_nodes` and `batch_num_edges`.

python/dgl/batched_graph.py
```diff
@@ -305,7 +50,6 @@ def unbatch(graph):
     --------
     batch
     """
-    assert isinstance(graph, BatchedDGLGraph)
```
jermainewang: Add a shortcut for `batch_size == 1`?

jermainewang: There is an unused `split` function above. Help delete it?

tests/compute/test_readout.py
```diff
@@ -195,7 +195,7 @@ def test_softmax_edges():
 def test_broadcast_nodes():
     # test#1: basic
     g0 = dgl.DGLGraph(nx.path_graph(10))
-    feat0 = F.randn((40,))
+    feat0 = F.randn((1, 40))
```
jermainewang: Could you double-check all examples can still run with this new change?

python/dgl/batched_graph.py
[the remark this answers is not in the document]
yzh119: Support mutation for batched graph is ambiguous, we must specify which graph to add/remove nodes/edges on.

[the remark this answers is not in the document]
yzh119: What do you mean by `g.flatten()`, considering we already have `batch` and `unbatch` api.

python/dgl/batched_graph.py
```diff
@@ -808,21 +612,15 @@ def _broadcast_on(graph, typestr, feat_data):
...
-        num_objs = getattr(graph, num_objs_attr)()
-        if F.ndim(feat_data) == 1:
-            feat_data = F.unsqueeze(feat_data, 0)
-        return F.cat([feat_data] * num_objs, 0)
+    _, batch_num_objs_attr, _ = READOUT_ON_ATTRS[typestr]
+
+    batch_num_objs = getattr(graph, batch_num_objs_attr)
+    index = []
+    for i, num_obj in enumerate(batch_num_objs):
+        index.extend([i] * num_obj)
```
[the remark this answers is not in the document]
yzh119: Thanks, that's a useful api.

python/dgl/batched_graph.py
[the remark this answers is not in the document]
jermainewang: I think we could still support it but mutating a batched graph erases the batch information, making it one single graph. The erasure could be encapsulated in a `g.flatten()` API, which is different from either `batch` and `unbatch`.

python/dgl/subgraph.py
yzh119: removed

python/dgl/graph.py
yzh119: If so, `batch_num_nodes` would return wrong results after we mutate on the graph.

python/dgl/batched_graph.py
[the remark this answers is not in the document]
yzh119: DONE.

python/dgl/graph.py
```diff
@@ -1244,6 +1279,112 @@ def remove_edges(self, eids):
...
+        """Write node/edge features to the parent graph.
+
+        Parameters
+        ----------
+        inplace : bool
+            If true, use inplace write (no gradient but faster)
+        """
+        if self._parent is None:
+            raise DGLError("We only support copy_to_parent for subgraphs.")
+        nids = self.ndata.pop(NID)
```
yzh119: @jermainewang do you have any suggestions on improving this? We don't want the EID/NID field of subgraph to pollute parent graph's frame.

python/dgl/graph.py
```diff
@@ -1244,6 +1279,112 @@ def remove_edges(self, eids):
...
+        self.edata[EID] = eids
+
+    def copy_from_parent(self):
+        """Copy node/edge features from the parent graph.
+
+        All old features will be removed.
+        """
+        if self._parent is None:
+            raise DGLError("We only support copy_from_parent for subgraphs.")
+        nids = self.ndata[NID]
```
yzh119: ditto.

python/dgl/graph.py
yzh119: DONE.

python/dgl/graph.py
yzh119: DONE

python/dgl/graph.py
yzh119: So I suggest to let `batch_num_nodes` and `batch_num_edges` to depend on `number_of_nodes` and `number_of_edges` for single graphs.

python/dgl/graph.py
yzh119: DONE

python/dgl/graph.py
yzh119: DONE

python/dgl/transform.py
yzh119: DONE.

python/dgl/batched_graph.py
```diff
@@ -343,15 +87,90 @@ def batch(graph_list, node_attrs=ALL, edge_attrs=ALL):
...
+            num_items = g.number_of_nodes()
+            attr_types = set(g.node_attr_schemes().keys())
+        elif mode == 'edge':
+            num_items = g.number_of_edges()
+            attr_types = set(g.edge_attr_schemes().keys())
+        return num_items, attr_types
+
+    def _init_attrs(attrs, mode):
+        if attrs is None:
+            return []
```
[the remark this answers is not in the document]
yzh119: Since the `node_attrs` and `edge_attrs` argument in `dgl.batch` could be `None`, I think it necessary to handle `None` here.

python/dgl/batched_graph.py
```diff
@@ -343,15 +87,90 @@ def batch(graph_list, node_attrs=ALL, edge_attrs=ALL):
...
+        elif is_all(attrs):
+            attrs = set()
+            # Check if at least a graph has mode items and associated features.
+            for i, g in enumerate(graph_list):
+                g_num_items, g_attrs = _get_num_item_and_attr_types(g, mode)
+                if g_num_items > 0 and len(g_attrs) > 0:
+                    attrs = g_attrs
+                    ref_g_index = i
+                    break
```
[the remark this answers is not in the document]
yzh119: DONE.

python/dgl/batched_graph.py
```diff
@@ -343,15 +87,90 @@ def batch(graph_list, node_attrs=ALL, edge_attrs=ALL):
...
+    def _get_num_item_and_attr_types(g, mode):
+        if mode == 'node':
+            num_items = g.number_of_nodes()
+            attr_types = set(g.node_attr_schemes().keys())
+        elif mode == 'edge':
+            num_items = g.number_of_edges()
+            attr_types = set(g.edge_attr_schemes().keys())
+        return num_items, attr_types
+
+    def _init_attrs(attrs, mode):
```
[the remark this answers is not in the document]
yzh119: DONE.

python/dgl/graph.py
yzh119: It depends on the subgraph itself so I suppose its a class method.

python/dgl/batched_graph.py
yzh119: DONE.

python/dgl/graph.py
jermainewang: I think your solution is fine. I don't have a better solution at the moment.

yzh119:
@jermainewang , another question is should we keep `DGLSubGraph` and `BatchedDGLGraph` class, I don't think so as they are used in nowhere.

python/dgl/batched_graph.py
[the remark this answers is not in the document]
yzh119: @BarclayII has the api been exposed on python end?

jermainewang:
> @jermainewang , another question is should we keep `DGLSubGraph` and `BatchedDGLGraph` class, I don't think so as they are used in nowhere.

Let's remove them.

Issue #1204, opened 2020-01-14
RGCN entity classification, incorrect loss function?


pbloem:
## 🐛 Bug

In the RGCN [entity classification example](https://github.com/dmlc/dgl/blob/master/examples/pytorch/rgcn/entity_classify.py), the model has a softmax activation, but the loss function, [F.cross_entropy](https://pytorch.org/docs/stable/nn.functional.html#cross-entropy), expects raw logits.

Unless I'm mistaken, either the last layer should have no activation, or the loss should be changed to F.nll_loss.

## To Reproduce

Running the example. Setting the activation to None has no effect on the score for AIFB. I can't easily test larger data like AM atm. If I add a double softmax to my own implementation of RGCN the performance on AM completely collapses (but then, I'm still debugging it).

pbloem:
I've now debugged my own implementation, and having a double softmax doesn't seem to affect performance significantly (which explains why your numbers are still good). Still, it's technically more correct to ditch the activation (or use nll_loss).

jermainewang:
Thanks for the report. You are correct. Will fix it.

-Minjie

PR #1217, opened 2020-01-19
[Bugfix] Correct the loss function in RGCN model

yzh119:
## Description
As mentioned in #1204 , the loss function in our RGCN implementation is problematic. This PR fix the issue.

## Checklist
- [x] Code is well-documented


examples/pytorch/rgcn/entity_classify.py
```diff
@@ -117,8 +117,7 @@ def main(args):
         backward_time.append(t2 - t1)
         print("Epoch {:05d} | Train Forward Time(s) {:.4f} | Backward Time(s) {:.4f}".
               format(epoch, forward_time[-1], backward_time[-1]))
-        train_acc = torch.sum(logits[train_idx].argmax(dim=1) == labels[train_idx]).item() / len(train_idx)
-        val_loss = F.cross_entropy(logits[val_idx], labels[val_idx])
+        cross_entropy(logits[val_idx], labels[val_idx])
```
jermainewang: `F.cross_entropy`? Also, isn't `tran_acc` required in L123?

yzh119: Fixed.

jermainewang:
Would you mind also fixing [the tensorflow example](https://github.com/dmlc/dgl/blob/master/examples/tensorflow/rgcn/entity_classify.py#L40) too?

yzh119:
The accuracy become worse, we need to tune the hyper parameters again...
* AIFB: 94.44
* MUTAG: 72.06
* BGS: 75.86

jermainewang:
@pbloem We are trying to fix the double-softmax problem. However, we found that the accuracy drops quite significantly. Have you observed the same issue?

VoVAllen:
I already fixed it in tensorflow at https://github.com/dmlc/dgl/blob/master/examples/tensorflow/rgcn/entity_classify.py#L109

yzh119:
@VoVAllen have you tuned hyper-parameters accordingly?

VoVAllen:
@yzh119 It's unstable but the highest accuracy could reach the original result.

pbloem:
@jermainewang Nope. I get the following mean accuracies over 10 repeats (standard deviation in brackets):

aifb: 0.961 (0.0194)
bgs: 0.834 (0.0317)
am: 0.892 (0.0167)

BGS is a little below what we had in the original paper, but the variance is high, so I think we just got lucky. All hyperparameters are the same as in the paper, and the weight initialization is taken from your code (ie. uniform xavier with relu gain everywhere).

You can find [my implementation here](https://github.com/pbloem/gated-rgcn/blob/3ec103c11048a8d31e555f5af108d3820dd7f8ee/kgmodels/models.py#L9), and the [training loop here](https://github.com/pbloem/gated-rgcn/blob/3ec103c11048a8d31e555f5af108d3820dd7f8ee/experiments/train.full.py#L100).

I strongly recommend averaging over 10 repeats, since the test sets are so small, but none of my runs on BGS are as low as .75.

yzh119:
@pbloem thanks for your instructions!

PR #1224, opened 2020-01-28
[Bug or feature?] Thread-safe Lazy object and UnitGraph::GetXXX

### LINKED ISSUE #1218: Make heterographs in C++ backend (and Lazy<T>) thread-safe?
## 🚀 Feature

Almost all the interfaces in C++ backend is not thread-safe and may crash when working with multithreading (e.g. OpenMP).  This is due to lazy materialization of the necessary storage format (COO, CSR, CSC), and is especially prominent on very small graphs.

## Motivation

For instance, the following code sometimes crashes:

```C++
HeteroGraphPtr CreateGraphFromCOO() {
  std::vector<dgl_id_t> src = {0,1,2};
  std::vector<dgl_id_t> dst = {1,2,3};
  return UnitGraph::CreateFromCOO(
      2,
      4,
      4,
      aten::VecToIdArray(src),
      aten::VecToIdArray(dst));
}

HeteroGraphPtr CreateGraphFromCSR() {
  std::vector<size_t> indptr = {0,1,2,3,3};
  std::vector<dgl_id_t> data = {0,1,2};
  std::vector<dgl_id_t> indices = {1,2,3};
  return UnitGraph::CreateFromCSR(
      2,
      4,
      4,
      aten::VecToIdArray(indptr),
      aten::VecToIdArray(indices),
      aten::VecToIdArray(data));
}

TEST(ThreadSafetyTest, TestHasVertex) {
  auto _test = [] (HeteroGraphPtr g) {
    std::vector<bool> result(omp_get_max_threads());
#pragma omp parallel for
    for (int i = 0; i < omp_get_max_threads(); ++i)
      result[i] = g->HasVertex(0, 0);
    for (int i = 0; i < omp_get_max_threads(); ++i)
      ASSERT_EQ(result[i], true);
  };
  _test(CreateGraphFromCOO());
  _test(CreateGraphFromCSR());
}

TEST(ThreadSafetyTest, TestHasEdgeBetween) {
  auto _test = [] (HeteroGraphPtr g) {
    std::vector<uint64_t> result(omp_get_max_threads());
#pragma omp parallel for
    for (int i = 0; i < omp_get_max_threads(); ++i)
      result[i] = g->HasEdgeBetween(0, 0, 1);
    for (int i = 0; i < omp_get_max_threads(); ++i)
      ASSERT_EQ(result[i], true);
  };
  _test(CreateGraphFromCOO());
  _test(CreateGraphFromCSR());
}
```

I'm marking this as a "Feature" instead of a "Bug" because this can be avoided by manually materializing COO/CSR/CSC before going into OpenMP, but I'm wondering if there is a better way to address this issue, as debugging when forgetting to do so is painful.

## Additional context

BarclayII:
## Description

Fixes #1218 .  Still finding a simple test case to demonstrate this issue.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/include/dgl/lazy.h
+++ b/include/dgl/lazy.h
@@ -7,14 +7,38 @@
 #define DGL_LAZY_H_
 
 #include <memory>
+#include <mutex>
+#include <atomic>
 
 namespace dgl {
+
+/*!
+ * \brief Lazy initialization
+ */
+template <typename T, typename Fn>
+const std::shared_ptr<T> LazyInit(std::shared_ptr<T>* ptr, std::mutex *mutex, Fn func) {
+  std::shared_ptr<T> result;
+
+  while (!(result = std::atomic_load_explicit(ptr, std::memory_order_acquire))) {
+    std::lock_guard<std::mutex> lock(*mutex);
+    result = *ptr;
+
+    if (!result) {
+      result.reset(new T(func()));
+      std::atomic_store_explicit(ptr, result, std::memory_order_release);
+    }
+  }
+
+  return result;
+}
 
 /*!
  * \brief Lazy object that will be materialized only when being queried.
  *
  * The object should be immutable -- no mutation once materialized.
- * The object is currently not threaad safe.
+ *
+ * \note Implementation taken from
+ * https://www.justsoftwaresolutions.co.uk/threading/multithreading-in-c++0x-part-6-double-checked-locking.html
  */
 template <typename T>
 class Lazy {
@@ -23,7 +47,20 @@ class Lazy {
   Lazy() {}
 
   /*!\brief constructor to construct an object with given value (non-lazy case) */
-  explicit Lazy(const T& val): ptr_(new T(val)) {}
+  explicit Lazy(const T& val): ptr_(new T(val)), preinitialized_(true) {}
+
+  /*!\brief copy constructor which ignores mutex */
+  Lazy(const Lazy& other) {
+    ptr_ = other.ptr_;
+    preinitialized_ = other.preinitialized_;
+  }
+
+  /*!\brief assignment operator which ignores mutex */
+  Lazy& operator=(const Lazy other) {
+    this->ptr_ = other.ptr_;
+    this->preinitialized_ = other.preinitialized_;
+    return *this;
+  }
 
   /*!\brief destructor */
   ~Lazy() = default;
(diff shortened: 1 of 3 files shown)


include/dgl/lazy.h
```diff
@@ -7,14 +7,38 @@
...
+ * \brief Lazy initialization
+ */
+template <typename T, typename Fn>
+const std::shared_ptr<T> LazyInit(std::shared_ptr<T>* ptr, std::mutex *mutex, Fn func) {
+  std::shared_ptr<T> result;
+
+  while (!(result = std::atomic_load_explicit(ptr, std::memory_order_acquire))) {
+    std::lock_guard<std::mutex> lock(*mutex);
+    result = *ptr;
+
```
classicsong: For this kind of check and lock, you should do check, lock and check style.
Otherwise, there is still data race here.

src/graph/unit_graph.cc
```diff
@@ -907,52 +908,68 @@ UnitGraph::UnitGraph(GraphPtr metagraph, CSRPtr in_csr, CSRPtr out_csr, COOPtr c
...
-      CHECK(out_csr_) << "Both CSR are missing.";
-      const auto& newadj = aten::CSRToCOO(out_csr_->adj(), true);
-      const_cast<UnitGraph*>(this)->coo_ = std::make_shared<COO>(meta_graph(), newadj);
-    }
-  }
-  return coo_;
+  return LazyInit(&(const_cast<UnitGraph*>(this)->coo_), &lazy_mutex_, [this] () {
+      if (in_csr_) {
+        const auto& newadj = aten::CSRToCOO(in_csr_->adj(), true);
+        return COO(
```
classicsong: Will this cause extra data copy?

zheng-da:
it seems OK. However, this doesn't fix some fundamental problems: e.g., if `GetInCSR` is called in OMP threads, the In-CSR will be generated many times even if one of the CSRs will be set in the pointer correctly. It's still huge waste. When a graph is large, we should parallelize graph conversion as we discussed, instead of having each thread generates one.

src/graph/unit_graph.cc
BarclayII: Which line are you referring to?

include/dgl/lazy.h
BarclayII: Line 22 does the first check.
Line 23 locks.
Line 26 does the second check.
Correct me if I'm wrong.

BarclayII:
> if `GetInCSR` is called in OMP threads, the In-CSR will be generated many times even if one of the CSRs will be set in the pointer correctly.

The mutex should prevent In-CSR from generating multiple times.

src/graph/unit_graph.cc
classicsong: L964

BarclayII: It shouldn't?  Because in `LazyInit()` this is where the lambda function is called:
```c++
result.reset(new T(func()))
```
So the COO object would be moved to the shared pointer and there shouldn't be any copy.

Correct me if I'm wrong though, as I'm sometimes also confused by when C++ copies and moves an object...

Issue #1120, opened 2019-12-19
[RFC] Explicit device context and index type for DGLGraph


jermainewang:
### Motivation

In DGL, a DGLGraph stores two types of data: graph topology and node/edge features. Currently, it stores graph topology in CPU. During message passing, if the feature data is on GPU, it will copy the graph structure to GPU and *cache* it for later use.

Another design flaw is the index type. Currently, we store graph topology only in int64 type. During message passing, we check and cast the graph to int32 type if the computation occurs on GPU.

From my perspective, the design causes several issues:
* **Storage overhead**: Storing graph topology first on CPU and then copy it to GPU is a waste.
* **Opaque conversion and caching mechanism**: It is hard to keep track of the memory usage when the system caches copies of graph for computation.
* **Index type cannot be controlled by users**: Creating a DGLGraph using int32 is not possible in DGL right now.

### Proposal

Just like that a tensor in MX/PT/TF has explicit device context and data type, the RFC proposes that DGLGraph should have explicit device context and index type as well:

```python
>>> g = DGLGraph(..., device='cuda:0', itype=np.int32)
>>> print(g.device)
cuda:0
>>> print(g.itype)
int32
```

Of course, the `device` and `itype` arguments could have default values. A `DGLGraph` with device type `cuda:0` means that **both graph topology and feature data** are stored on GPU#0.

```python
>>> import torch as th
>>> g = DGLGraph(..., device='cuda:0')
>>> h = th.zeros((g.number_of_nodes(), 5), device='cpu')
>>> g.ndata['h'] = h  # throw error due to mismatched device contexts
```

To copy the graph from one device to another, use the `to` API (already implemented), or use `cpu()`, `cuda(devid)` APIs.

```python
>>> g = DGLGraph(..., device='cuda:0')
>>> g0 = g.to('cuda:1')
>>> print(g0.device)
cuda:1
>>> g1 = g0.cpu()
>>> print(g1.device)
cpu
```

In the future, DGL will forbidden any operations whose input graphs or tensors have different device contexts and index types. Examples are:
* In `dgl.batch(graphs)`, the input graphs must all share the same device and index type.
* In `g.in_edges(nodes)`, if the `nodes` argument is of tensor type, it should have the same device and index type as `g`. Otherwise, the `nodes` argument must be a list or other python iterable and are converted to a tensor automatically.
* In `g.adj()`, the returned adjacency matrix will follow the device and index type of `g`.

These operations are carried out on the same device as their arguments.

[...]

PR #1238, opened 2020-02-01
[Feature] Preference to COO for "hypersparse" unit graphs & graph compaction

BarclayII:
### Preference to COO for hypersparse unit graphs

When the number of edges are significantly less than the number of nodes, saving a unit graph as CSR would waste memory.  This PR aims to switch the storage to COO for those "hypersparse" unit graphs.

The policy of preferring COO over CSR is determined by the internal function `dgl/convert.py::detect_prefer_coo`.  Currently, it is set when the number of edges is less than 20% of number of *source* nodes.

Related to the new sampler design since the sampled graphs per layer would often be hypersparse (e.g. only containing inbound edges of a minibatch of nodes).

### `compact_graphs`

`compact_graphs` API (#1199 ) is also implemented in this PR.

## Checklist
- [x] Code is well-documented


BarclayII:
Should be ready for review.

src/array/cpu/spmat_op_impl_coo.cc
```diff
@@ -0,0 +1,497 @@
...
+  const auto rowlen = row->shape[0];
+  const auto collen = col->shape[0];
+  const auto rstlen = std::max(rowlen, collen);
+  NDArray rst = NDArray::Empty({rstlen}, row->dtype, row->ctx);
+  IdType* rst_data = static_cast<IdType*>(rst->data);
+  const IdType* row_data = static_cast<IdType*>(row->data);
+  const IdType* col_data = static_cast<IdType*>(col->data);
+  const int64_t row_stride = (rowlen == 1 && collen != 1) ? 0 : 1;
+  const int64_t col_stride = (collen == 1 && rowlen != 1) ? 0 : 1;
+  const int64_t kmax = std::max(rowlen, collen);
```
classicsong: Does rowlen always == collen here? row and col are pairs.

src/array/cpu/spmat_op_impl_coo.cc
```diff
@@ -0,0 +1,497 @@
...
+  const IdType* col_data = static_cast<IdType*>(cols->data);
+
+  const IdType* coo_row_data = static_cast<IdType*>(coo.row->data);
+  const IdType* coo_col_data = static_cast<IdType*>(coo.col->data);
+  const DType* data = static_cast<DType*>(coo.data->data);
+
+  std::vector<IdType> ret_rows, ret_cols;
+  std::vector<DType> ret_data;
+
+  for (int64_t i = 0, j = 0; i < rowlen && j < collen; i += row_stride, j += col_stride) {
```
classicsong: OMP can also be used here.
It seems you support broadcast row/col cases as well as normal case here. Suggest to have some code comments here for future maintenance.

src/array/cpu/spmat_op_impl_coo.cc
BarclayII: Not necessarily; we allow broadcasting here (i.e. row and col can contain only one element).

src/array/cpu/spmat_op_impl_coo.cc
BarclayII: I'm pushing elements to a vector and I don't know how many entries to push beforehand.

src/graph/unit_graph.cc
```diff
@@ -851,13 +956,14 @@ HeteroSubgraph UnitGraph::EdgeSubgraph(
 }
 
 HeteroGraphPtr UnitGraph::CreateFromCOO(
-    int64_t num_vtypes, int64_t num_src, int64_t num_dst, IdArray row, IdArray col) {
+    int64_t num_vtypes, int64_t num_src, int64_t num_dst, IdArray row, IdArray col,
+    bool prefer_coo) {
```
jermainewang:
```c++
HeteroGraphPtr UnitGraph::CreateFromCOO(
    int64_t num_vtypes, int64_t num_src, int64_t num_dst, IdArray row, IdArray col,
    SparseFormat store_format)
```
`store_format` default to be `kAny`.

src/graph/unit_graph.cc
```diff
@@ -881,9 +987,17 @@ HeteroGraphPtr UnitGraph::AsNumBits(HeteroGraphPtr g, uint8_t bits) {
...
+    if (!bg->prefer_coo_) {
+      CSRPtr new_incsr = CSRPtr(new CSR(bg->GetInCSR()->AsNumBits(bits)));
+      CSRPtr new_outcsr = CSRPtr(new CSR(bg->GetOutCSR()->AsNumBits(bits)));
+      return HeteroGraphPtr(
+          new UnitGraph(g->meta_graph(), new_incsr, new_outcsr, nullptr, false));
+    } else {
+      COOPtr new_coo = COOPtr(new COO(bg->GetCOO()->AsNumBits(bits)));
+      return HeteroGraphPtr(
+          new UnitGraph(g->meta_graph(), nullptr, nullptr, new_coo, true));
+    }
```
jermainewang: The question is for any function `GG = foo(G)`, if `G` is a hypersparse graph (preferring COO), should `GG` also prefer `COO`? I think maybe a better solution is to lower the `detect_prefer_coo` function in python to C++ so whenever a graph is created, it will try to store it in a more efficient format.

src/graph/unit_graph.cc
```diff
@@ -897,13 +1011,20 @@ HeteroGraphPtr UnitGraph::CopyTo(HeteroGraphPtr g, const DLContext& ctx) {
...
+    COOPtr new_coo = COOPtr(new COO(bg->GetCOO()->CopyTo(ctx)));
+    return HeteroGraphPtr(new UnitGraph(g->meta_graph(), nullptr, nullptr, new_coo, true));
+  }
 }
 
-UnitGraph::UnitGraph(GraphPtr metagraph, CSRPtr in_csr, CSRPtr out_csr, COOPtr coo)
-  : BaseHeteroGraph(metagraph), in_csr_(in_csr), out_csr_(out_csr), coo_(coo) {
+UnitGraph::UnitGraph(GraphPtr metagraph, CSRPtr in_csr, CSRPtr out_csr, COOPtr coo, bool prefer_coo)
+  : BaseHeteroGraph(metagraph), in_csr_(in_csr), out_csr_(out_csr), coo_(coo),
+    prefer_coo_(prefer_coo) {
```
jermainewang: If using the `lock_format` solution I proposed, this constructor should simply accept a `lock_format` argument.

jermainewang:
My major reviews:
* Try make `prefer_coo` flag more flexible since we would like to prefer other formats in the future.
* It might be better to decide storage format in C++ than in Python because many APIs will construct and return graph objects (such as subgraph, batching, etc.), and lowering this functionality to C++ will benefit them. This also makes the format decision more opaque so it is easier to implement new operators. For example, performing a vertex subgraph operation on a hypersparse graph might give a non-hypersparse graph, for which CSR format is allowed. Similarly, edge subgraph operation could result in hypersparse graph result. It's better to hide the format decision from operator implementation except certain cases.

python/dgl/convert.py
```diff
@@ -49,6 +50,8 @@ def graph(data, ntype='_N', etype='_E', card=None, validate=True, **kwargs):
         If True, check if node ids are within cardinality, the check process may take
         some time. (Default: True)
         If False and card is not None, user would receive a warning.
+    _prefer_coo : bool or 'auto', optional
+        Internal argument for forcing the storage to be in COO format.  Do not use.
```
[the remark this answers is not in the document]
BarclayII: This is for unit test only.  I changed the docstring in the latest commits.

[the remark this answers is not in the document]
BarclayII: That's only for unit test.

src/graph/unit_graph.cc
```diff
@@ -687,7 +736,9 @@ BoolArray UnitGraph::HasVertices(dgl_type_t vtype, IdArray vids) const {
 }
 
 bool UnitGraph::HasEdgeBetween(dgl_type_t etype, dgl_id_t src, dgl_id_t dst) const {
-  if (in_csr_) {
+  if (prefer_coo_) {
+    return GetCOO()->HasEdgeBetween(etype, src, dst);
+  } else if (in_csr_) {
```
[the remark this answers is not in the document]
BarclayII: The `GetSuitableFormat` is actually quite complicated if we have a "lock format" semantic:
1. If the format is locked, return the locked format no matter what.
2. Otherwise, if the graph is hypersparse, return COO no matter what.
3. Otherwise, return the preferred format depending on the API itself (e.g. `HasEdgeBetween` would go for whatever the graph has, `InEdges` would prefer CSC, `FindEdges` would prefer COO).

Does that make sense?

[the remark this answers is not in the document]
jermainewang: There are two decisions that should not be mixed.
1. Which format to store the graph?
2. Which format to use for certain operator?

A good place for the first decision is the constructor function of a graph object, while the second decision is performed in each operator. As a result, if a graph is hypersparse, the constructor will decide to use COO and lock the format. Thus, my `GetSuitableFormat` is simpler as shown in the demo code.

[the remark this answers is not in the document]
BarclayII: > As a result, if a graph is hypersparse, the constructor will decide to use COO and lock the format.

Should we lock whenever the graph is hypersparse (and not allow converting to CSR)?  What if I want to do message passing?  I'm afraid that we would break the code that indeed does message passing on a hypersparse graph dataset.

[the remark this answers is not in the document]
jermainewang: > > As a result, if a graph is hypersparse, the constructor will decide to use COO and lock the format.
> 
> Should we lock whenever the graph is hypersparse (and not allow converting to CSR)? What if I want to do message passing? I'm afraid that we would break the code that indeed does message passing on a hypersparse graph dataset.

I c. The case is that some operators do not have implementation on a certain format while we still want to keep the code runnable, where locking a format is indeed not good. Can we change the `store_format` as `preferred_format` in the constructor, and the first argument of `GetSuitableFormat` as `allowed_format`:
(1) The `allowed_format` is a list of formats. Default is `kAny`.
(2) If `allowed_format` is specified, try match one with the stored format. Otherwise, convert and return any of the `allowed_format`.

Several cases:
* `HasEdgesBetween` provides `kAny`.
* `InEdges` provides `kCOO` and `kCSC`.
* `BinaryReduce` provides `kCSR` and `kCSC`.

What do you think?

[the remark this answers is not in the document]
jermainewang: Note that this proposal does not have the `lock_format_` member anymore. The constructor directly stores the graph in the `preferred_format`.

include/dgl/array_utils.h
```diff
@@ -0,0 +1,78 @@
...
+#include <vector>
+#include <unordered_map>
+#include <utility>
+
+namespace dgl {
+
+namespace aten {
+
+/*!
+ * \brief A hashmap that maps each ids in the given array to new ids starting from zero.
```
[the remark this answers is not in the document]
BarclayII: Moved to `src`.

python/dgl/convert.py
```diff
@@ -703,7 +713,31 @@ def to_homo(G):
...
+
+    If the edges is much less than the number of nodes, storing the graph as CSR is
+    not memory-efficient.
+
+    Parameters
+    ----------
+    urange, vrange : int
+        Number of source and destination nodes, respectively.
+    u, v : iterable of int
+        List of source and destination node IDs.
```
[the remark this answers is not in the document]
BarclayII: Moved the detection to C++.

src/graph/unit_graph.cc
BarclayII: It is lowered in C++ now.

src/array/cpu/spmat_op_impl_coo.cc
classicsong: i c. You have to handle broadcast here. So you cannot decide the size easily.
You can use vector<vector<data>> for omp

src/graph/unit_graph.cc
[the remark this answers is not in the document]
classicsong: One thing came to me is the design of preferred_format of lock_format may conflict with the target_parallel kernel design. But I think we can delay completing the target parallel kernel implementation.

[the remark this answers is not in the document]
jermainewang: It shouldn't? Kernel should specify which format it requests right?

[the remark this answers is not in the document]
classicsong: > It shouldn't? Kernel should specify which format it requests right?
Yes, kernel will specify the required format. Usually it requires all COO CSR CSC in case of forward and backward. So kernel need temp store for these formats.
Currently, kernel directly using low-level APIs of Unigraph e.g., getCSRMatrix(). So these ops will not be affected?

Another concern is which kind of kernel is good for hypersparse graph, which may be critical for GNN on large graph with sampling.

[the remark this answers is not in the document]
jermainewang: > Yes, kernel will specify the required format. Usually it requires all COO CSR CSC in case of forward and backward. So kernel need temp store for these formats.

My hunch is SPMM/SDDMM kernels probably should not require all formats. Correct me if my understanding is wrong.
* For SPMM-like by node parallelism, CSC is required.
* For SDDMM-like by edge parallelism, COO is required.
* For backward SPMM by node parallelism, CSR is required.

> Another concern is which kind of kernel is good for hypersparse graph, which may be critical for GNN on large graph with sampling.

Sampled hypersparse subgraphs will be compacted before message passing so normal SPMM and SDDMM kernels should work fine.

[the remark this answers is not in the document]
BarclayII: > I c. The case is that some operators do not have implementation on a certain format while we still want to keep the code runnable, where locking a format is indeed not good. Can we change the `store_format` as `preferred_format` in the constructor, and the first argument of `GetSuitableFormat` as `allowed_format`:
> (1) The `allowed_format` is a list of formats. Default is `kAny`.
> (2) If `allowed_format` is specified, try match one with the stored format. Otherwise, convert and return any of the `allowed_format`.
> 
> Several cases:
> 
> * `HasEdgesBetween` provides `kAny`.
> * `InEdges` provides `kCOO` and `kCSC`.
> * `BinaryReduce` provides `kCSR` and `kCSC`.
> 
> What do you think?

I adopted this solution (kind of).  Essentially we now

* have a `restrict_format_` member that restricts the format to run *graph queries* on.
* have `restrict_format_` set to COO automatically if the graph is hypersparse and constructed via COO (if from CSR we probably don't care since the indptr is already initialized anyway).
* have a private method `UnitGraph::GetFormat` which takes in a format suggestion (e.g. CSC) and return a storage pointer either as restricted, or as suggested if not restricted.
* allow materializing CSR/CSC/COO even if it is restricted, if requested explicitly (i.e. calling `GetInCSR`/`GetOutCSR`/`GetCOO` directly, bypassing `GetFormat`).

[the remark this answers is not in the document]
classicsong: > > Yes, kernel will specify the required format. Usually it requires all COO CSR CSC in case of forward and backward. So kernel need temp store for these formats.
> 
> My hunch is SPMM/SDDMM kernels probably should not require all formats. Correct me if my understanding is wrong.
> 
> * For SPMM-like by node parallelism, CSC is required.
> * For SDDMM-like by edge parallelism, COO is required.
> * For backward SPMM by node parallelism, CSR is required.
> 

consider  u_add_v none, in forward, it is edge parallel, COO is required, in backward, both lhs and rhs are node parallel, and with different target, both CSC and CSR is required. We need to take both forward and backward as a round-trip.

> > Another concern is which kind of kernel is good for hypersparse graph, which may be critical for GNN on large graph with sampling.
> 
> Sampled hypersparse subgraphs will be compacted before message passing so normal SPMM and SDDMM kernels should work fine.

I see. Compact do help.

src/array/cpu/spmat_op_impl_coo.cc
BarclayII: And I need to concatenate them afterwards, which involves another copy from different arrays.  I wonder how much gain could I get (especially for small graphs where COO implementations mainly take place).

python/dgl/convert.py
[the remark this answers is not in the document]
BarclayII: Now we have a `restrict_format` option instead.  For now the argument is intended for unit test only, and normal users should not be aware of this.

python/dgl/convert.py
```diff
@@ -122,17 +125,20 @@ def graph(data, ntype='_N', etype='_E', card=None, validate=True, **kwargs):
...
+        return create_from_edges(
+            u, v, ntype, etype, ntype, urange, vrange, validate, _prefer_coo)
     elif isinstance(data, list):
-        return create_from_edge_list(data, ntype, etype, ntype, urange, vrange, validate)
+        return create_from_edge_list(
+            data, ntype, etype, ntype, urange, vrange, validate, _prefer_coo)
     elif isinstance(data, sp.sparse.spmatrix):
         return create_from_scipy(data, ntype, etype, ntype)
     elif isinstance(data, nx.Graph):
         return create_from_networkx(data, ntype, etype, **kwargs)
```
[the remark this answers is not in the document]
BarclayII: Yes.

python/dgl/convert.py
```diff
@@ -703,7 +713,31 @@ def to_homo(G):
 # Internal APIs
 ############################################################
 
-def create_from_edges(u, v, utype, etype, vtype, urange=None, vrange=None, validate=True):
+# pylint: disable=unused-argument
+def detect_prefer_coo(urange, vrange, u, v):
+    """
+    Decides whether to prefer graph queries on COO format for a unit graph, given the
+    number of source and destination nodes, as well as the edges.
```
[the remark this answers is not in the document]
BarclayII: Moved to `COO::IsHypersparse` in C++.

src/array/cpu/spmat_op_impl_coo.cc
```diff
@@ -0,0 +1,435 @@
...
+  const IdType* col_data = static_cast<IdType*>(col->data);
+  const int64_t row_stride = (rowlen == 1 && collen != 1) ? 0 : 1;
+  const int64_t col_stride = (collen == 1 && rowlen != 1) ? 0 : 1;
+  const int64_t kmax = std::max(rowlen, collen);
+#pragma omp parallel for
+  for (int64_t k = 0; k < kmax; ++k) {
+    int64_t i = row_stride * k;
+    int64_t j = col_stride * k;
+    rst_data[k] = COOIsNonZero<XPU, IdType>(coo, row_data[i], col_data[j])? 1 : 0;
+  }
```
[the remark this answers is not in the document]
BarclayII: I left a TODO for sorted COO.

src/array/cpu/spmat_op_impl_coo.cc
```diff
@@ -0,0 +1,435 @@
...
+  return false;
+}
+
+template bool COOHasDuplicate<kDLCPU, int32_t>(COOMatrix coo);
+template bool COOHasDuplicate<kDLCPU, int64_t>(COOMatrix coo);
+
+///////////////////////////// COOGetRowNNZ /////////////////////////////
+
+template <DLDeviceType XPU, typename IdType>
+int64_t COOGetRowNNZ(COOMatrix coo, int64_t row, bool transpose) {
```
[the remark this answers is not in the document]
BarclayII: Now it's decomposed to `COOGetRowNNZ` on a `COOTranspose`.

BarclayII:
Re-requesting reviews since I also added in implementation of compaction before the first review round.

src/graph/unit_graph.cc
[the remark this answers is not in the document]
jermainewang: > * have a private method `UnitGraph::GetFormat` which takes in a format suggestion (e.g. CSC) and return a storage pointer either as restricted, or as suggested if not restricted.  All graph queries would get storage using `GetFormat`.
> * allow materializing CSR/CSC/COO even if it is restricted, if requested explicitly (i.e. calling `GetInCSR`/`GetOutCSR`/`GetCOO` directly, bypassing `GetFormat`).  This case is for message passing.

Are they conflicting with each other? The first bullet should be another API than `UnitGraph::GetFormat` which I suppose return the requested format no matter what.

src/graph/unit_graph.h
```diff
@@ -182,11 +187,29 @@ class UnitGraph : public BaseHeteroGraph {
...
    * \param coo coo
    */
-  UnitGraph(GraphPtr metagraph, CSRPtr in_csr, CSRPtr out_csr, COOPtr coo);
+  UnitGraph(GraphPtr metagraph, CSRPtr in_csr, CSRPtr out_csr, COOPtr coo,
+            SparseFormat restrict_format = SparseFormat::ANY);
 
   /*! \return Return any existing format. */
   HeteroGraphPtr GetAny() const;
 
+  /*! \return Return the given format. */
```
jermainewang: Perform format conversion if the requested format does not exist.

src/graph/unit_graph.h
```diff
@@ -34,8 +34,10 @@ class UnitGraph : public BaseHeteroGraph {
   // internal data structure
   class COO;
   class CSR;
+  class CSC;
   typedef std::shared_ptr<COO> COOPtr;
   typedef std::shared_ptr<CSR> CSRPtr;
+  typedef std::shared_ptr<CSC> CSCPtr;
```
jermainewang: remove?

src/graph/unit_graph.cc
```diff
@@ -977,4 +1132,31 @@ HeteroGraphPtr UnitGraph::GetAny() const {
...
+    return GetOutCSR();
+  else if (format == SparseFormat::CSC)
+    return GetInCSR();
+  else if (format == SparseFormat::COO)
+    return GetCOO();
+  else if (format == SparseFormat::ANY)
+    return GetAny();
+
+  LOG(FATAL) << "Unsupported format code: " << format;
+  return nullptr;
```
jermainewang: One benefit of using `enum class` is that the compiler will warn you about a missing condition when using `switch ... case ...`.

src/graph/unit_graph.cc
```diff
@@ -312,6 +366,10 @@ class UnitGraph::COO : public BaseHeteroGraph {
     return adj_;
   }
 
+  bool IsHypersparse() const {
+    return NumVertices(SrcType()) / 8 > NumEdges(EdgeType());
+  }
```
jermainewang: `(NumVertices(SrcType()) + NumVertices(DstType())) / 16 > NumEdges(EdgeType())`? BTW, you could directly use adj_ too but since these calls should be inlined, it should be fine either way.

jermainewang: Oh, I see your point here. Look like if |Src| >> |E|, we should restrict COO and CSC. If |Dst| >> |E|, we should restrict COO and CSR. Is that right?

src/graph/unit_graph.cc
```diff
@@ -656,29 +722,25 @@ bool UnitGraph::IsMultigraph() const {
 }
 
 uint64_t UnitGraph::NumVertices(dgl_type_t vtype) const {
-  if (in_csr_) {
+  const SparseFormat fmt = SelectFormat(SparseFormat::ANY);
+  const auto ptr = GetFormat(fmt);
+  // TODO(BarclayII): we have a lot of special handling for CSC.
+  // Need to have a UnitGraph::CSC backend instead.
```
jermainewang: Agree. CSC graph and CSR graph should have the same set of behaviors.

src/graph/unit_graph.cc
```diff
@@ -851,24 +974,41 @@ HeteroSubgraph UnitGraph::EdgeSubgraph(
...
+    SparseFormat restrict_format) {
+  return UnitGraph::CreateFromCOO(num_vtypes, num_src, num_dst, row, col, restrict_format);
+}
+
+HeteroGraphPtr CreateUnitGraphFromCSR(
+    int64_t num_vtypes, int64_t num_src, int64_t num_dst,
+    IdArray indptr, IdArray indices, IdArray edge_ids,
+    SparseFormat restrict_format) {
+  return UnitGraph::CreateFromCSR(
+      num_vtypes, num_src, num_dst, indptr, indices, edge_ids, restrict_format);
```
jermainewang: Maybe I missed sth here. Why need these two creator functions?

src/graph/unit_graph.cc
```diff
@@ -851,24 +974,41 @@ HeteroSubgraph UnitGraph::EdgeSubgraph(
 }
 
 HeteroGraphPtr UnitGraph::CreateFromCOO(
-    int64_t num_vtypes, int64_t num_src, int64_t num_dst, IdArray row, IdArray col) {
+    int64_t num_vtypes, int64_t num_src, int64_t num_dst, IdArray row, IdArray col,
+    SparseFormat restrict_format) {
```
jermainewang: Can we have a default `Any` value for `restrict_format` so it's easier to use it.

src/graph/unit_graph.cc
```diff
@@ -881,9 +1021,11 @@ HeteroGraphPtr UnitGraph::AsNumBits(HeteroGraphPtr g, uint8_t bits) {
     //   be fixed later.
     auto bg = std::dynamic_pointer_cast<UnitGraph>(g);
     CHECK_NOTNULL(bg);
+
     CSRPtr new_incsr = CSRPtr(new CSR(bg->GetInCSR()->AsNumBits(bits)));
     CSRPtr new_outcsr = CSRPtr(new CSR(bg->GetOutCSR()->AsNumBits(bits)));
-    return HeteroGraphPtr(new UnitGraph(g->meta_graph(), new_incsr, new_outcsr, nullptr));
+    return HeteroGraphPtr(
+        new UnitGraph(g->meta_graph(), new_incsr, new_outcsr, nullptr, SparseFormat::ANY));
```
jermainewang: This one might worth inheriting the restrict format flag.

src/graph/unit_graph.cc
```diff
@@ -897,13 +1039,26 @@ HeteroGraphPtr UnitGraph::CopyTo(HeteroGraphPtr g, const DLContext& ctx) {
   //   be fixed later.
   auto bg = std::dynamic_pointer_cast<UnitGraph>(g);
   CHECK_NOTNULL(bg);
+
   CSRPtr new_incsr = CSRPtr(new CSR(bg->GetInCSR()->CopyTo(ctx)));
   CSRPtr new_outcsr = CSRPtr(new CSR(bg->GetOutCSR()->CopyTo(ctx)));
-  return HeteroGraphPtr(new UnitGraph(g->meta_graph(), new_incsr, new_outcsr, nullptr));
+  return HeteroGraphPtr(
+      new UnitGraph(g->meta_graph(), new_incsr, new_outcsr, nullptr, SparseFormat::ANY));
```
jermainewang: Same here.

tests/compute/test_heterograph.py
```diff
@@ -62,6 +62,23 @@ def create_test_heterograph2():
...
+    wishes_nx.add_edge('u0', 'g1', id=0)
+    wishes_nx.add_edge('u2', 'g0', id=1)
+
+    follows_g = dgl.graph([(0, 1), (1, 2)], 'user', 'follows', _restrict_format='coo')
+    plays_g = dgl.bipartite(
+        [(0, 0), (1, 0), (2, 1), (1, 1)], 'user', 'plays', 'game', _restrict_format='coo')
+    wishes_g = dgl.bipartite([(0, 1), (2, 0)], 'user', 'wishes', 'game', _restrict_format='coo')
+    develops_g = dgl.bipartite(
+        [(0, 0), (1, 1)], 'developer', 'develops', 'game', _restrict_format='coo')
+    g = dgl.hetero_from_relations([follows_g, plays_g, wishes_g, develops_g])
```
jermainewang: Should test w/o restrict format but is indeed hypersparse.

python/dgl/convert.py
```diff
@@ -699,11 +721,97 @@ def to_homo(G):
...
+        The graph, or list of graphs
+
+    Returns
+    -------
+    DGLHeteroGraph or list[DGLHeteroGraph]
+        The compacted graph or list of compacted graphs
+    dict[str, Tensor]
+        Induced nodes of each type.  Contains the mapping of node IDs for each type
+        from the compacted graph(s) to the original graph(s).
+        Note that the mapping is the same for all the compacted graphs.
```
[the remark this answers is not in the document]
BarclayII: Otherwise we need to copy the features to every graph.

python/dgl/convert.py
```diff
@@ -699,11 +726,98 @@ def to_homo(G):
 
     return retg
 
+def compact_graphs(graphs):
+    """Given a list of graphs with the same set of nodes, find and eliminate the common
+    isolated nodes across all graphs.
+
+    This function requires the graphs to have the same set of nodes (i.e. the node types
+    must be the same, and the number of nodes of each node type must be the same), as
+    well as the same metagraph.
```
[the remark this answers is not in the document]
BarclayII: You're right.

python/dgl/convert.py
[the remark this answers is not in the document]
jermainewang: There won't be deep copy. Each relation graph just holds a reference to the mapping tensor.

src/graph/unit_graph.cc
BarclayII: Yes.

src/graph/unit_graph.cc
BarclayII: Should we export these creation functions?

src/graph/unit_graph.cc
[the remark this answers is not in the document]
BarclayII: I suppose we are on the same page; My `GetFormat` is essentially your "another API", which I can rename to `GetSuitableFormat(SparseFormat suggestion)`.  Your `GetFormat` would be my `GetInCSR`/`GetOutCSR`/`GetCOO`.

src/graph/unit_graph.cc
BarclayII: They do have a default value in `unit_graph.h`.

src/graph/unit_graph.cc
BarclayII: These two functions are for casting the graph index for message passing, where CSR/CSC must be materialized.  Since they are materialized anyway, the restrict format flag no longer matters.

src/graph/unit_graph.cc
jermainewang: I c. You could decide.

src/graph/unit_graph.cc
BarclayII: Besides, we don't have a GPU implementation for COOToCSR, so message passing on GPUs will fail.

src/graph/unit_graph.cc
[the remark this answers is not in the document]
BarclayII: Yes, they are named `SelectFormat` and `GetFormat` respectively.

python/dgl/convert.py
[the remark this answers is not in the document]
BarclayII: Indeed.

src/graph/unit_graph.cc
```diff
@@ -897,13 +1039,26 @@ HeteroGraphPtr UnitGraph::CopyTo(HeteroGraphPtr g, const DLContext& ctx) {
...
+UnitGraph::UnitGraph(GraphPtr metagraph, CSRPtr in_csr, CSRPtr out_csr, COOPtr coo,
+                     SparseFormat restrict_format)
   : BaseHeteroGraph(metagraph), in_csr_(in_csr), out_csr_(out_csr), coo_(coo) {
+  restrict_format_ = restrict_format;
+
+  // If the graph is hypersparse and in COO format, switch the restricted format to COO.
+  // If the graph is given as CSR, the indptr array is already materialized so we don't
+  // care about restricting conversion anyway (even if it is hypersparse).
+  if (restrict_format == SparseFormat::ANY) {
+    if (coo && coo->IsHypersparse())
```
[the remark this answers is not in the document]
BarclayII: Added in `IsHypersparse` instead.

tests/compute/test_heterograph.py
BarclayII: Added a `test_hypersparse` function.

src/graph/unit_graph.cc
jermainewang: Got it. We'll likely to revisit this when implementing #1120 where AsNumBits and CopyTo would not materialize non-existing formats.

src/graph/unit_graph.cc
jermainewang: Look like we need a more flexible way to specify the restricted formats. Any idea?

PR #1242, opened 2020-02-05
[Data format] Serialization for UnitGraph

VoVAllen:
## Description
Serialization for UnitGraph and CSRMatrix

## Checklist
- [ ] Code is well-documented


src/graph/unit_graph.cc
```diff
@@ -1152,4 +1152,38 @@ SparseFormat UnitGraph::SelectFormat(SparseFormat preferred_format) const {
...
+  aten::CSRMatrix csr_matrix;
+  CHECK(fs->Read(&csr_matrix)) << "Invalid csr_matrix";
+  SparseFormat restrict_format;
+  CHECK(fs->Read(&restrict_format)) << "Invalid restrict_format";
+  auto mg = CreateUnitGraphMetaGraph(num_vtypes);
+  CSRPtr csr(new CSR(mg, num_src, num_dst, csr_matrix.indptr, csr_matrix.indices, csr_matrix.data));
+  in_csr_ = csr;
+  meta_graph_ = mg;
+  restrict_format_ = restrict_format;
+  // HeteroGraphPtr hptr = CreateFromCSR(num_vtypes, num_src, num_dst, csr_matrix.indptr,
```
VoVAllen: The code aboves is copied from `CreateFromCSR`. I want to reuse the code in `CreateFromCSR`, but due to my limited C++ skill I don't know how to properly move `hptr` into `this`. Do I need something like a move constructor? Could you give any suggestions? @BarclayII @jermainewang

BarclayII: I would write `*this = UnitGraph(mg, csr, nullptr, nullptr, restrict_format)`.

BarclayII:
Can you also write a unit test?

VoVAllen:
Sure.

VoVAllen:
@BarclayII I didn't add the support for COO format. The main reason is that I'd like to preserve simplicity in favor of compatibility. Multiple formats need to depend on the implementation detail of the class, which may be subject to change in the future and leads to incompatibility. 
However, I leave the magic number in the serialization, which means we can extend the method in the future based on the magic number.

BarclayII:
Sounds good.

Issue #1240, opened 2020-02-04
[Bug or Feature?] Heterograph.canonical_etypes is none-deterministic between runs


classicsong:
## 🐛 Bug
When you run examples/pytorch/rgcn-hetero/ entity classification example, you can see when creating RelGraphConvHeteroEmbed in each run, the model embeddings are different from run to run. If we want to save_model for later inference and then load_model in another run, the torch model loading will crash.

## Expected behavior

Heterograph.canonical_etypes should be deterministic, so each time in entity_classify,py we build the model from run to run, the model should be identical.

## Environment

 - DGL Version (e.g., 1.0): 0.4.1

classicsong:
Fixed in pull request #1244

PR #1244, opened 2020-02-06
[Example] Fix Rgcn hetero nonedetermistic in graph and model initialization

classicsong:
## Description
RGCN-hetero model should be reloadable for further online and offline inference. But there are several none-deterministic points:
1. For rdf loader, if the graph caching is cleaned, the order of rdf triples loaded changes from run to run
2. For the model nn.Parameters (L160, entity\_classify.py), as g.canonical_etypes is none-deterministic from run to run, we should use torch.nn.ParameterDict() to store the tensors for different nodetype so that they can be loaded in inference step.
3. EntityClassify.rel_names (L222, entity\_classify.py) is not deterministic, which cause a random order of assigning basis_weight tensor for each relation type.

Also we provide an example of how to load the pre-trained model and do the inference (test) separately from traning. 

Updating the Readme.

#1240 

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/data/rdf.py
+++ b/python/dgl/data/rdf.py
@@ -142,7 +142,12 @@ class RDFGraphDataset:
         dst = []
         ntid = []
         etid = []
-        for i, (sbj, pred, obj) in enumerate(raw_tuples):
+        sorted_tuples = []
+        for t in raw_tuples:
+            sorted_tuples.append(t)
+        sorted_tuples.sort()
+
+        for i, (sbj, pred, obj) in enumerate(sorted_tuples):
             if i % self._print_every == 0:
                 print('Processed %d tuples, found %d valid tuples.' % (i, len(src)))
             sbjent = self.parse_entity(sbj)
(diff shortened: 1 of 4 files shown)


python/dgl/data/rdf.py
```diff
@@ -142,7 +142,12 @@ def process_raw_tuples(self, raw_tuples):
         dst = []
         ntid = []
         etid = []
-        for i, (sbj, pred, obj) in enumerate(raw_tuples):
+        sorted_tuples = []
+        for t in raw_tuples:
+            sorted_tuples.append(t)
+        sorted_tuples.sort()
+
+        for i, (sbj, pred, obj) in enumerate(sorted_tuples):
```
jermainewang: That's interesting. This told us that we'd better make our code deterministic if graph structure is generated from some raw input data because users might develop transductive models that are sensitive to the actual order.

classicsong: Yes, this is the case when we separate training and inference.
This should be carefully handled, otherwise, the model is meaningless.

PR #1249, opened 2020-02-08
[Feature] PinSAGE sampler

BarclayII:
## Description
This implements the Python PinSAGE neighbor sampler (for one layer), and implements the necessary components:

* `select_topk` (@jermainewang please note this since you have another implementation)
* `to_simple` for heterographs

`create_graph_from_paths` is skipped because it can be implemented with tensor operators and `dgl.graph`.

## Checklist
- [ ] Code is well-documented


BarclayII:
@zheng-da I wonder if you want to benchmark this?  Or if you would like to do that after having a PinSAGE model working end-to-end?  I was planning to make a separate PR for the PinSAGE model.

python/dgl/sampling/pinsage.py
```diff
@@ -0,0 +1,214 @@
...
+        The bidirectional bipartite graph.
+
+        The graph should only have two node types: ``ntype`` and ``other_type``.
+        The graph should only have two edge types, one connecting from ``ntype`` to
+        ``other_type``, and another connecting from ``other_type`` to ``ntype``.
+
+        PinSAGE works on a bidirectional bipartite graph where for each edge
+        going from node u to node v, there exists an edge going from node v to node u.
+    ntype : str
+        The node type for which the graph would be constructed on.
```
jermainewang: Is this the same as the starting and the ending types of the given metapath?

python/dgl/sampling/pinsage.py
```diff
@@ -0,0 +1,214 @@
...
+    metapath : list[str]
+        The metapath.
```
jermainewang: Can we make it optional so no need to provide this for homograph?

python/dgl/sampling/pinsage.py
```diff
@@ -0,0 +1,214 @@
...
+
+        Usually considered a hyperparameter.
+    num_random_walks : int
+        Number of random walks to try for each seed node.
+
+        Usually considered a hyperparameter.
+    num_neighbors : int
+        Number of neighbors to select for each seed.
+    weight_column : str, default "weights"
+        The weight of each neighbor, stored as an edge feature.
```
jermainewang: Is this weight used by randomwalk?

python/dgl/sampling/randomwalks.py
```diff
@@ -46,7 +46,7 @@ def random_walk(g, nodes, *, metapath=None, length=None, prob=None, restart_prob
         The feature tensor must be non-negative.
         If omitted, we assume the neighbors are picked uniformly.
     restart_prob : float or Tensor, optional
-        Probability to stop after each step.
+        Probability to stop before each step.
```
jermainewang: at?

python/dgl/transform.py
```diff
@@ -565,14 +581,142 @@ def partition_graph_with_halo(g, node_part, num_hops):
...
+    >>> new_g.edges(form='all', order='eid', etype='plays')
+    (tensor([0, 1]), tensor([0, 1]), tensor([0, 1]))
+    >>> new_g2.edges(form='all', order='eid', etype='plays')
+    (tensor([0, 2]), tensor([2, 3]), tensor([0, 1]))
+    """
+    return_single = False
+    if not isinstance(graphs, Iterable):
+        graphs = [graphs]
+        return_single = True
+
```
jermainewang:
```python
if len(graphs) == 0:
  return []
```

python/dgl/transform.py
```diff
@@ -565,14 +581,142 @@ def partition_graph_with_halo(g, node_part, num_hops):
...
+    # Ensure the node types are ordered the same.
+    # TODO(BarclayII): we ideally need to remove this constraint.
+    ntypes = graphs[0].ntypes
+    graph_dtype = graphs[0]._graph.dtype()
+    graph_ctx = graphs[0]._graph.ctx()
+    for g in graphs:
+        assert ntypes == g.ntypes, "Node types are not ordered the same"
```
jermainewang: "All graphs should have the same node types. Got {} and {}.".format(ntypes, g.ntypes)

src/c_api_common.h
```diff
@@ -21,7 +21,10 @@ inline bool operator == (const DLDataType& ty1, const DLDataType& ty2) {
 
 /*! \brief Output the string representation of device context.*/
 inline std::ostream& operator << (std::ostream& os, const DLDataType& ty) {
-  return os << "code=" << ty.code << ",bits=" << ty.bits << "lanes=" << ty.lanes;
+  return os <<
+    "code=" << static_cast<int>(ty.code) <<
+    ",bits=" << static_cast<int>(ty.bits) <<
+    "lanes=" << static_cast<int>(ty.lanes);
```
jermainewang: What happens here?

BarclayII: `ty.code` and `ty.bits` are unsigned char, so for instance '@' would be shown in place of 64.

include/dgl/transform.h
```diff
@@ -0,0 +1,115 @@
...
+ *   connecting simple_g.src[i] and simple_g.dst[i] in the original graph.
+ *
+ *       count[0] = [1, 2, 2, 1]
+ *
+ * * One can find the mapping between edges from the original graph to the new simple
+ *   graph.
+ *
+ *       edge_map[0] = [0, 1, 3, 1, 2, 2]
+ */
+std::tuple<HeteroGraphPtr, std::vector<IdArray>, std::vector<IdArray>>
```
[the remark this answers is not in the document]
BarclayII: Let the users know which new edge an original edge is mapped to?

include/dgl/array.h
```diff
@@ -619,7 +640,7 @@ COOMatrix COORowWiseTopk(
     COOMatrix mat,
     IdArray rows,
     int64_t k,
-    FloatArray weight,
+    NDArray weight,
```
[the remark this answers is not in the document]
BarclayII: I don't think we want to actually encourage anything here.  Both ints and floats should work for this function.

include/dgl/transform.h
```diff
@@ -0,0 +1,115 @@
...
+ * 2. The edge IDs induced from the original graph
+ *
+ *     induced_edges = [1, 2, 5, 6]
+ */
+std::pair<HeteroGraphPtr, std::vector<IdArray>> SelectTopK(
+    const HeteroGraphPtr graph,
+    const std::vector<NDArray> &weights,
+    int K,
+    bool inbound = true,
+    bool smallest = false);
```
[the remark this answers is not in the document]
BarclayII: I removed the implementation but forgot to remove the declaration...

python/dgl/sampling/pinsage.py
```diff
@@ -0,0 +1,214 @@
...
+
+    Given a heterogeneous graph, PinSAGE neighbor sampler would generate a homogeneous
+    graph where the neighbors of each node are the most commonly visited nodes of the
+    same type by random walk with restarts.  The random walk with restarts are based
+    on a given metapath, which should have the same beginning and ending node type.
+
+    Parameters
+    ----------
+    G : DGLHeteroGraph
+        The bidirectional bipartite graph.
```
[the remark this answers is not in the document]
BarclayII: My mistake.  It should be of any heterograph for `GenericPinSAGESampler`.

python/dgl/sampling/pinsage.py
BarclayII: No, but it is used for message aggregation.

include/dgl/array.h
```diff
@@ -115,6 +115,21 @@ template<typename ValueType>
...
+ * \brief Permute the elements of an array according to given indices.
+ *
+ * Equivalent to:
+ *
+ * <code>
+ *     result = np.zeros_like(array)
+ *     result[indices] = array
+ * </code>
+ */
+NDArray Scatter(NDArray array, IdArray indices);
```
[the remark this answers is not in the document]
BarclayII: Not necessarily.  It is the same as `torch.scatter`.  `IndexSelect` is actually `result = array[indices]`.

python/dgl/sampling/pinsage.py
BarclayII: You're right.

include/dgl/transform.h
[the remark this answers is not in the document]
jermainewang: In the python API, the one thing I don't like is setting a mapping to the input graph. If `dgl.NID` or `dgl.EID` is a mapping to the parent graph, then who is the parent graph in this case? I would directly return an optional mapping array.

python/dgl/sampling/pinsage.py
```diff
@@ -0,0 +1,214 @@
...
+        The node type for which the graph would be constructed on.
+    metapath : list[str]
+        The metapath.
+    random_walk_length : int
+        The maximum number of steps of random walk with restarts.
+
+        Note that here we consider traversing from ``ntype`` to ``other_type`` then back
+        to ``ntype`` as a single step (i.e. a single step consists of two hops).
+
+        Usually considered a hyperparameter.
```
jermainewang: Remove this argument? We could let the `metapath` argument have repetition.

python/dgl/sampling/pinsage.py
```diff
@@ -0,0 +1,214 @@
...
+    def __init__(self, G, ntype, metapath, random_walk_length, random_walk_restart_prob,
+                 num_random_walks, num_neighbors, weight_column='weights'):
+        self.G = G
+        self.ntype = ntype
+        self.weight_column = weight_column
+        self.num_random_walks = num_random_walks
+        self.num_neighbors = num_neighbors
+        self.random_walk_length = random_walk_length
+
+        self.metapath = metapath * random_walk_length
```
jermainewang: We could let the `metapath` argument directly have repetition. Also, make sure to check the head and tail types are the same. Another question is about the restart. What if a walk restarts (halts) at a type different than the head type?

python/dgl/sampling/pinsage.py
jermainewang: Then I guess this could be inferred from `metapath`?

python/dgl/sampling/pinsage.py
```diff
@@ -0,0 +1,214 @@
...
+import numpy as np
+
+from .. import backend as F
+from .. import convert
+from .. import transform
+from .. import function as fn
+from .randomwalks import random_walk
+from .neighbor import select_topk
+from ..base import EID
+
```
jermainewang: Add a `__all__`? You use `from pinsage import *` and I suppose the `GenericPinSAGESampler` is an internal class. Do we have other classes wish to inherit this generic class?

python/dgl/transform.py
```diff
@@ -645,4 +788,68 @@ def out_subgraph(g, nodes):
...
+    ----------
+    g : DGLHeteroGraph
+        The heterogeneous graph
+    return_counts : str, optional
+        If given, the returned graph would have a column with the same name that stores
+        the number of duplicated edges from the original graph.
+    writeback_mapping : bool, default False
+        If True, the mapping from the edge IDs of original graph to those of the returned
+        graph would be written into edge feature ``dgl.EID`` in the original graph for
+        each edge type.
```
jermainewang: Don't like this write-back behavior. Consider return the mapping optionally.

python/dgl/transform.py
```diff
@@ -645,4 +788,68 @@ def out_subgraph(g, nodes):
...
+
+    This function does not preserve node and edge features.
+
+    Parameters
+    ----------
+    g : DGLHeteroGraph
+        The heterogeneous graph
+    return_counts : str, optional
+        If given, the returned graph would have a column with the same name that stores
+        the number of duplicated edges from the original graph.
```
jermainewang: `return_counts=None` -> `counts='count'`?

src/array/cpu/array_repeat.cc
```diff
@@ -0,0 +1,48 @@
+/*!
+ *  Copyright (c) 2019 by Contributors
```
jermainewang: Remember to change copyright to 2020.

src/array/cpu/coo_coalesce.cc
```diff
@@ -0,0 +1,50 @@
...
+
+namespace dgl {
+
+namespace aten {
+
+namespace impl {
+
+template <DLDeviceType XPU, typename IdType>
+COOMatrix COOCoalesce(COOMatrix coo) {
+  CHECK(coo.row_sorted && coo.col_sorted) << "[BUG] Coalescing is only supported on sorted matrix";
```
jermainewang: How about just
```c++
if (!(coo.row_sorted && coo.col_sorted))
  coo = SortCOO(coo, true);
```

src/graph/transform/to_simple.cc
```diff
@@ -0,0 +1,100 @@
...
+
+  std::vector<IdArray> counts(num_etypes), edge_maps(num_etypes);
+  std::vector<HeteroGraphPtr> rel_graphs(num_etypes);
+
+  for (int64_t etype = 0; etype < num_etypes; ++etype) {
+    const auto vtypes = graph->GetEndpointTypes(etype);
+    const COOMatrix adj = graph->GetCOOMatrix(etype);
+    const COOMatrix sorted_adj = COOSort(adj, true);
+    const IdArray eids_shuffled = sorted_adj.data;
+    const COOMatrix coalesced_adj = COOCoalesce(sorted_adj);
```
jermainewang: It might worth implementing `Coalesce` for CSR too.

src/array/cpu/coo_coalesce.cc
```diff
@@ -0,0 +1,50 @@
...
+      new_col.push_back(curr_col);
+      new_data.push_back(1);
+      prev_row = curr_row;
+      prev_col = curr_col;
+    }
+  }
+
+  return COOMatrix{
+      coo.num_rows, coo.num_cols, NDArray::FromVector(new_row), NDArray::FromVector(new_col),
+      NDArray::FromVector(new_data), true};
```
jermainewang: I think this does not comply with our convention about the data array. For operators that shuffle/slice a sparse matrix, they should correspondingly shuffle/slice the data array if exists. If the input matrix does not have data array, we assume it to be [0,NNZ), and the same rule applies. We need to write it somewhere.

jermainewang: Hence, this operator should return the `count` array together with the new COO matrix.

src/graph/transform/to_simple.cc
```diff
@@ -0,0 +1,100 @@
...
+     * eids_remapped[i] indicates the new edge ID the old edge #i is mapped to.  The
+     * translation can simply be achieved by (in numpy code):
+     *
+     *     new_eid_for_eids_shuffled = np.range(len(count)).repeat(count)
+     *     eids_remapped = np.zeros_like(new_eid_for_eids_shuffled)
+     *     eids_remapped[eids_shuffled] = new_eid_for_eids_shuffled
+     */
+    const IdArray new_eids = Range(
+        0, coalesced_adj.row->shape[0], coalesced_adj.row->dtype.bits, coalesced_adj.row->ctx);
+    const IdArray eids_remapped = Scatter(Repeat(new_eids, count), eids_shuffled);
```
jermainewang: If `COOCoalesce` follows the convention about data array, this part could be avoided.

include/dgl/transform.h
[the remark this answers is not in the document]
BarclayII: Sure

python/dgl/sampling/pinsage.py
```diff
@@ -0,0 +1,214 @@
...
+        restart_prob = np.tile(np.array([random_walk_restart_prob, 0]), random_walk_length)
+        restart_prob[0] = 0     # allow at least one step
+        self.restart_prob = F.zerocopy_from_numpy(restart_prob)
+
+    # pylint: disable=no-member
+    def __call__(self, seed_nodes):
+        seed_nodes = F.repeat(seed_nodes, self.num_random_walks, 0)
+        paths, _ = random_walk(
+            self.G, seed_nodes, metapath=self.metapath, restart_prob=self.restart_prob)
+        src = F.reshape(paths[:, 2::2], (-1,))
```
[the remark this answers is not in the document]
BarclayII: Shoot that's a bug, I should have a test case for Generic PinSAGE.

python/dgl/sampling/pinsage.py
```diff
@@ -0,0 +1,214 @@
...
+        counts = neighbor_graph.edata[self.weight_column]
+        neighbor_graph = select_topk(neighbor_graph, self.num_neighbors, self.weight_column)
+        selected_counts = F.gather_row(counts, neighbor_graph.edata[EID])
+
+        # normalize weights
+        with neighbor_graph.local_scope():
+            neighbor_graph.edata[self.weight_column] = F.astype(selected_counts, F.float32)
+            neighbor_graph.update_all(fn.copy_e(self.weight_column, 'm'), fn.sum('m', 's'))
+            neighbor_graph.apply_edges(fn.e_div_v(self.weight_column, 's', 'w'))
+            new_weights = neighbor_graph.edata['w']
```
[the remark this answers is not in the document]
BarclayII: Currently PinSAGE aggregation takes in normalized weights.  I would do it for the users instead of letting them do it themselves.

src/array/cpu/coo_coalesce.cc
BarclayII: Sure.

src/graph/transform/to_simple.cc
BarclayII: Why can this be avoided?  We still need the mapping from the original graph's edges to those of the new simple graph.

python/dgl/sampling/pinsage.py
BarclayII: I actually don't want it to be an internal class.  For instance, one can also use `GenericPinSAGESampler` on a homogeneous graph.  `PinSAGESampler` is just a convenient wrapper on `GenericPinSAGESampler` for bidirectional bipartite graphs.

python/dgl/sampling/pinsage.py
BarclayII: Actually, `random_walk_length` controls where the random walks would restart.  In PinSAGE, random walks can only restart when a metapath traversal ends (I do need to change the docstring of `random_walk_restart_prob` though).  So I don't think I can remove the `random_walk_length` argument.

python/dgl/sampling/pinsage.py
BarclayII: See above.

src/graph/transform/to_simple.cc
jermainewang: I see. Is this part for the reverse mapping from original graph to the result graph (thus having length equal to |E_orig|)? If so, it cannot be avoided.

python/dgl/sampling/pinsage.py
jermainewang: Can we extend `PinSAGESampler` to make it work directly for homograph? I guess adding some default argument values might do the trick.

BarclayII: In fact any heterographs can have a PinSAGE sampler running given a metapath that has the same start and end node type.

src/graph/transform/to_simple.cc
BarclayII: Yes.

python/dgl/transform.py
BarclayII: It will return a dictionary of mappings by edge type though.

src/graph/transform/to_simple.cc
BarclayII: Can we do this after someone complains about the speed of `to_simple` on big graphs?

python/dgl/sampling/pinsage.py
BarclayII: Removed.

python/dgl/sampling/pinsage.py
```diff
@@ -0,0 +1,212 @@
...
+                 num_random_walks, num_neighbors, metapath=None, weight_column='weights'):
+        self.G = G
+        self.weight_column = weight_column
+        self.num_random_walks = num_random_walks
+        self.num_neighbors = num_neighbors
+        self.random_walk_length = random_walk_length
+
+        if metapath is None:
+            if len(G.ntypes) > 1 or len(G.etypes) > 1:
+                raise ValueError('Metapath must be specified if the graph is homogeneous.')
```
jermainewang: heterogeneous

python/dgl/sampling/pinsage.py
```diff
@@ -0,0 +1,212 @@
...
+            metapath = [G.canonical_etypes[0]]
+        start_ntype = G.to_canonical_etype(metapath[0])[0]
+        end_ntype = G.to_canonical_etype(metapath[-1])[-1]
+        if start_ntype != end_ntype:
+            raise ValueError('The metapath must start and end at the same node type.')
+        self.ntype = start_ntype
+
+        self.metapath_hops = len(metapath)
+        self.metapath = metapath * random_walk_length
```
jermainewang: Probably would save both the `metapath` for each step and a `full_metapath = metapath * random_walk_length`.

src/array/cpu/coo_coalesce.cc
```diff
@@ -0,0 +1,52 @@
...
+namespace impl {
+
+template <DLDeviceType XPU, typename IdType>
+std::pair<COOMatrix, IdArray> COOCoalesce(COOMatrix coo) {
+  const int64_t nnz = coo.row->shape[0];
+  const IdType* coo_row_data = static_cast<IdType*>(coo.row->data);
+  const IdType* coo_col_data = static_cast<IdType*>(coo.col->data);
+
+  if (!coo.row_sorted || !coo.col_sorted)
+    coo = COOSort(coo);
```
jermainewang: `COOSort(coo, True);`

src/array/cpu/coo_coalesce.cc
```diff
@@ -0,0 +1,52 @@
...
+
+  std::vector<IdType> new_row, new_col, new_data;
```
jermainewang: `new_data` -> `count`

src/array/cpu/coo_coalesce.cc
```diff
@@ -0,0 +1,52 @@
...
+      new_col.push_back(curr_col);
+      new_data.push_back(1);
+      prev_row = curr_row;
+      prev_col = curr_col;
+    }
+  }
+
+  COOMatrix coo_result = COOMatrix{
+      coo.num_rows, coo.num_cols, NDArray::FromVector(new_row), NDArray::FromVector(new_col),
+      NDArray(), true};
```
jermainewang: I see. So `coo_result` cannot refer back to the input due to it not being a valid mapping. It is now indistinguishable between *cannot inherit mapping* and *an identical mapping*. Any idea to fix this?

python/dgl/sampling/pinsage.py
```diff
@@ -0,0 +1,212 @@
...
+
+from .. import backend as F
+from .. import convert
+from .. import transform
+from .randomwalks import random_walk
+from .neighbor import select_topk
+from ..base import EID
+
+
+class GenericPinSAGESampler(object):
```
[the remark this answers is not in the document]
BarclayII: Good idea.

src/array/cpu/coo_coalesce.cc
BarclayII: What do you mean by "cannot inherit mapping"?  It is simply constructing a new COO with the new rows and columns.

PR #1254, opened 2020-02-10
[Data format] Serialization for Immutable Graph and HeteroGraph

VoVAllen:
## Checklist
- [ ] Code is well-documented


VoVAllen:
@BarclayII I'm not familiar with HeteroGraph enough. Could you give me some suggestions on the `HeteroGraph` part in `test_serialize.cc`? Such as how to construct more HeteroGraphs, verify edge ids and edge directions? Thanks.

BarclayII:
Could we expose the interface to Python (e.g. have `DGLHeteroGraph.save()` and `DGLHeteroGraph.load()`) and do unit test there instead?

VoVAllen:
Sure. I expect these would be used for pickling in the future.

src/graph/graph_serializer.h
```diff
@@ -0,0 +1,25 @@
+/*!
+ *  Copyright (c) 2018 by Contributors
+ * \file graph/graph_serializer.cc
+ * \brief DGL serializer APIs
+ */
+
+#pragma once
```
jermainewang: Please use the same header guard as the others.

src/graph/heterograph.h
```diff
@@ -170,7 +170,20 @@ class HeteroGraph : public BaseHeteroGraph {
...
+  /*! \return Save HeteroGraph to stream, using CSRMatrix */
+  void Save(dmlc::Stream* fs) const;
+
+
  private:
+  // To create empty class
+  friend class Serializer;
+
+  // Empty Constructor, only for serializer
+  HeteroGraph() : BaseHeteroGraph(static_cast<GraphPtr>(nullptr)) {}
```
jermainewang: If you'd like to add a default constructor, please first add one in `BaseHeteroGraph`.

src/graph/heterograph.cc
```diff
@@ -494,6 +500,39 @@ CompactGraphs(const std::vector<HeteroGraphPtr> &graphs) {
...
+  CHECK_EQ(magicNum, kDGLSerialize_HeteroGraph) << "Invalid HeteroGraph Data";
+  auto meta_grptr = new ImmutableGraph(static_cast<COOPtr>(nullptr));
+  CHECK(fs->Read(meta_grptr)) << "Invalid Immutable Graph Data";
+  uint64_t num_relation_graphs;
+  CHECK(fs->Read(&num_relation_graphs)) << "Invalid num of relation graphs";
+  std::vector<HeteroGraphPtr> relgraphs;
+  for (size_t i = 0; i < num_relation_graphs; ++i) {
+    UnitGraph* ugptr = Serializer::EmptyUnitGraph();
+    CHECK(fs->Read(ugptr)) << "Invalid UnitGraph Data";
+    relgraphs.emplace_back(dynamic_cast<BaseHeteroGraph*>(ugptr));
```
jermainewang: There is no need to up-cast a subclass pointer to its parent graph pointer.

jermainewang: Manipulating raw pointers with shared pointer as result is dangerous. Why not the following?
```c++
std::vector<HeteroGraphPtr> relgraphs;
for (...) {
  UnitGraphPtr ugptr = Serializer::EmptyUnitGraph();
  CHECK(fs->Read(ugptr.get())) << ... ;
  relgraphs.push_back(ugptr);
}
```

jermainewang: Also I really suggest implementing serializer handlers such as one [here](https://github.com/dmlc/dgl/blob/master/include/dgl/runtime/serializer.h). This automatically allows serializing type `std::vector<T>` where `T` has specialized `Handler<T>`.

src/graph/unit_graph.cc
```diff
@@ -1152,8 +1155,17 @@ SparseFormat UnitGraph::SelectFormat(SparseFormat preferred_format) const {
     return SparseFormat::COO;
 }
 
+UnitGraph* UnitGraph::EmptyGraph() {
+  auto src = NewIdArray(0);
+  auto dst = NewIdArray(0);
+  auto mg = CreateUnitGraphMetaGraph(1);
+  COOPtr coo(new COO(mg, 0, 0, src, dst));
+  return new UnitGraph(mg, nullptr, nullptr, coo);
+}
```
jermainewang: Please use shared pointer. Returning raw pointer is dangerous and it's hard to track liveness.

VoVAllen:
Thanks a lot. I'm keeping learning how to write high quality C++ codes.

src/graph/heterograph.cc
VoVAllen: > Also I really suggest implementing serializer handlers such as one [here](https://github.com/dmlc/dgl/blob/master/include/dgl/runtime/serializer.h). This automatically allows serializing type `std::vector<T>` where `T` has specialized `Handler<T>`.

Actually it's the same as the current implementation using `Load/Save` with the template class below.
https://github.com/dmlc/dmlc-core/blob/649be18a8c55c48517861d67158a45dec54992ee/include/dmlc/serializer.h#L104

PR #1257, opened 2020-02-11
[Test] Fix test script

VoVAllen:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,54 +1,143 @@
 # Deep Graph Library (DGL)
+
 [![Build Status](http://ci.dgl.ai:80/buildStatus/icon?job=DGL/master)](http://ci.dgl.ai:80/job/DGL/job/master/)
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](./LICENSE)
 
-[Documentation](https://docs.dgl.ai) | [DGL at a glance](https://docs.dgl.ai/tutorials/basics/1_first.html#sphx-glr-tutorials-basics-1-first-py) |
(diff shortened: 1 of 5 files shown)


BarclayII:
What is going on with this PR and #1256 ?

VoVAllen:
There's a weird error when `CHECK` macro failed.

http://ci.dgl.ai/blue/organizations/jenkins/DGL/detail/PR-1256/1/pipeline/133
image

A normal way of raising error should be like the figure below
image

The first figure error is usually an illegal memory access error. The reason why this happen is because the UnitGraph I created should have only 9 nodes, but I used 9 as node id in the [COO construction](https://github.com/dmlc/dgl/blob/e8ae12547ca155b80e68e48b9322f0e1de2b75ae/tests/cpp/test_serialize.cc#L62). However, if no error raised, everything works fine. The weird error only happens when `CHECK` macro failed.

BarclayII:
That indeed could happen.  An illegal memory write could wreak havoc in every possible way.  For instance, if you happen to change the `malloc` header that records the allocated chunk size, then the actual crash would happen in `malloc` or `free`.

VoVAllen:
`UnitGraph::CreatefromCOO` didn't check the validity of the COO matrix that each entry should be smaller than `num_src/num_dst`. Is it checked elsewhere?

https://github.com/dmlc/dgl/blob/23893bb0a87b98c7193ba24c5f6b9e802eeb9dea/src/graph/unit_graph.cc#L984

BarclayII:
I think it is checked in Python.  We probably need to move the check into C later.

PR #1263, opened 2020-02-14
[KVStore] New kvstore used by DGL-KE

aksnzhy:
## Description

This PR re-factoring the kvstore that can be used for DGL-kE. Compared to the previous version, this version supports backup servers that can share the same partition of data among server nodes on the same machine. For the client, it supports load-balance automatically (select backup server for communication randomly).

TODO: 

1. Add more test and examples for this PR.
2. Support async pull.

## Checklist
- [x] Code is well-documented


examples/pytorch/dis_kvstore/server.py
```diff
@@ -1,42 +1,55 @@
...
-DATA.append(th.tensor([[3.,3.,3.,],[3.,3.,3.,]]))
-DATA.append(th.tensor([[2.,2.,2.,],[2.,2.,2.,]]))
-DATA.append(th.tensor([[1.,1.,1.,],[1.,1.,1.,]]))
+    if args.server_id % args.backup_count == 0:
+        my_server.set_global2local(name='entity_embed', global2local=g2l[args.machine_id])
+        my_server.init_data(name='entity_embed', data_tensor=data[args.machine_id])
+    else:
+        time.sleep(3)
+        my_server.set_global2local(name='entity_embed', global2local=None, data_shape=tuple((8,)))
+        my_server.init_data(name='entity_embed', data_tensor=None, data_shape=tuple((2,3)))
```
zheng-da: I'm wondering if this is an easy API for people to use. In a real case, how does a user get the data shape? we cannot ask a user to hard code the shape.

aksnzhy: Yes, it could be a question. But the name-shared tensor API requires the user to provide the data shape. One solution is that the server node can communicate with each other and share the data_shape from the master server. But this can involve complicated logic on server nodes.

python/dgl/contrib/dis_kvstore.py
```diff
@@ -534,50 +495,62 @@ def __del__(self):
...
-        self._addr = self._get_local_addr()
+        # Send client address to server nodes
+        self._addr = self._get_local_usable_addr()
         client_ip, client_port = self._addr.split(':')
 
-        # find local server nodes
-        for ID, addr in self._server_namebook.items():
-            server_ip, server_port = addr.split(':')
-            if server_ip in self._ip4_addr_list():
-                self._local_server_id.add(ID)
```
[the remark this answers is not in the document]
aksnzhy: _get_machine_id() API will find local machine ID and compare it with the machine ID and server.

python/dgl/contrib/dis_kvstore.py
```diff
@@ -374,26 +340,62 @@ def start(self):
...
+            # Final message              
             elif msg.type == KVMsgType.FINAL:
-                print("Exit KVStore service, server ID: %d" % self._server_id)
+                print("Exit KVStore service %d" % self.get_id())
                 break # exit loop
             else:
                 raise RuntimeError('Unknown type of kvstore message: %d' % msg.type.value)
 
+            # garbage collection
             self._garbage_msg.append(msg)
```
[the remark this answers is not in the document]
aksnzhy: The pull() API uses zero-copy, so the memory should be release by user.

python/dgl/contrib/dis_kvstore.py
```diff
@@ -11,179 +11,106 @@
...
+        KVServer's ID (start from 0).
     server_addr : str
-        IP address and port of current KVServer node, e.g., '127.0.0.1:50051'.
+        machine ID, IP address and port of KVServer, e.g., '0|127.0.0.1:50051'.
     num_client : int
         Total number of clients connecting to server.
-    msg_queue_size : int
-        Size of message queue (2GB by default)
+    queue_size : int
+        Sise (bytes) of kvstore message queue buffer (~20 GB on default).
```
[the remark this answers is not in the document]
aksnzhy: No, it is just an upper-bound.

python/dgl/contrib/dis_kvstore.py
```diff
@@ -11,179 +11,106 @@
...
     dict
-        server namebook, e.g., 
-
-          {0:'172.31.40.143:50050',
-           1:'172.31.36.140:50050',
-           2:'172.31.47.147:50050',
-           3:'172.31.30.180:50050'}
+        server namebook. e.g.,
+
+        [server_id]:[machine_id|server_addr]
```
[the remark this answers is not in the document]
aksnzhy: I think storing all the infomation in string will be easy for uses to wirte the configuration file. The machine_id is provided by user in configuration file and I think it will be useful when debugging.

python/dgl/contrib/dis_kvstore.py
```diff
@@ -11,179 +11,106 @@
...
 
-    In practice, developers can use KVServer to hold large graph features or graph embeddings 
-    across machines in a distributed setting. User can re-wriite _push_handler and _pull_handler 
-    to support flexibale algorithms.
+    In practice, developers can use KVServer to hold large-scale graph features or 
+    graph embeddings across machines in a distributed setting. User can re-wriite _push_handler() 
+    and _pull_handler() API to support flexibale algorithms.
+
+    DGL kvstore supports backup servers. That means we can lunach many servers on the same machine and all of 
+    these servers share the same shared-memory tensor.
```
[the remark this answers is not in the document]
aksnzhy: Sure, I will change it.

python/dgl/contrib/dis_kvstore.py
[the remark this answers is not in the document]
aksnzhy: I will change the information of the server_namebook to a list.

examples/pytorch/dis_kvstore/server.py
aksnzhy: I think the master server can write the data_shape information into a tmp file and the other servers can read from that.

examples/pytorch/dis_kvstore/client.py
```diff
@@ -16,43 +21,45 @@
...
-    print(tensor_edata)
-    client.barrier()
+    my_client.connect()
 
-    print(tensor_ndata)
-    client.barrier()
+    if my_client.get_id() % args.backup_count == 0:
+        my_client.set_partition_book(name='entity_embed', partition_book=partition)
+    else:
+        time.sleep(3)
```
classicsong: why we need to sleep 3 secends?

examples/pytorch/dis_kvstore/client.py
```diff
@@ -16,43 +21,45 @@
...
+        my_client.push(name='entity_embed', id_tensor=ID[i], data_tensor=DATA[i])
 
-    tensor_edata = client.pull(name='edata', id_tensor=th.tensor([0,1,2,3,4,5,6,7]))
-    tensor_ndata = client.pull(name='ndata', id_tensor=th.tensor([0,1,2,3,4,5,6,7]))
+    my_client.barrier()
 
-    print(tensor_edata)
-    client.barrier()
+    if my_client.get_id() % args.backup_count == 0:
+        res = my_client.pull(name='entity_embed', id_tensor=th.tensor([0,1,2,3,4,5,6,7]))
```
classicsong: Is [0,1,2,3,4,5,6,7] a magic number here?

examples/pytorch/dis_kvstore/server.py
classicsong: Can we just automatically generate a config file for all these kinds of info and sent to all servers along with the partition?

python/dgl/contrib/dis_kvstore.py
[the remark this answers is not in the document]
classicsong: Does the backup servers act like replicas? All can access the data and update the data. 
How to avoid write conflicts? Or in async scenarios, we don't care w/r, w/w conflicts in kv strore?

python/dgl/contrib/dis_kvstore.py
```diff
@@ -11,179 +11,106 @@
...
         Total number of clients connecting to server.
-    msg_queue_size : int
-        Size of message queue (2GB by default)
+    queue_size : int
+        Sise (bytes) of kvstore message queue buffer (~20 GB on default).
     net_type : str
         networking type, e.g., 'socket' (default) or 'mpi' (do not support yet).
     """
-    def __init__(self, server_id, server_addr, num_client, msg_queue_size=2 * 1024 * 1024 * 1024, net_type='socket'):
+    def __init__(self, server_id, server_addr, num_client, queue_size=20000000000, net_type='socket'):
```
classicsong: I perfer 2 * 1024 * 1024 * 1024. It is more straight forward.

python/dgl/network.py
```diff
@@ -318,5 +327,4 @@ def _clear_kv_msg(garbage_msg):
     for msg in garbage_msg:
         if msg.c_ptr is not None:
             _CAPI_DeleteKVMsg(msg.c_ptr)
-    garbage_msg = []
-        
\ No newline at end of file
+    garbage_msg = []
```
classicsong: Is this necessary for a input parameter?

python/dgl/contrib/dis_kvstore.py
```diff
@@ -374,26 +340,62 @@ def start(self):
...
+                print("Exit KVStore service %d" % self.get_id())
                 break # exit loop
             else:
                 raise RuntimeError('Unknown type of kvstore message: %d' % msg.type.value)
 
+            # garbage collection
             self._garbage_msg.append(msg)
-            if len(self._garbage_msg) > 1000:
+            if len(self._garbage_msg) > GARBAGE_COLLECTION_COUNT:
                 _clear_kv_msg(self._garbage_msg)
```
classicsong: Why we use a list to store bag of msgs? Can we call _clear_ky_msg each time? The is huge performance overhead?

python/dgl/contrib/dis_kvstore.py
```diff
@@ -679,55 +701,57 @@ def pull(self, name, id_tensor):
...
                 msg = KVStoreMsg(
                     type=KVMsgType.PULL, 
                     rank=self._client_id, 
                     name=name, 
-                    id=partial_id, 
+                    id=partial_id,
                     data=None,
                     c_ptr=None)
-                _send_kv_msg(self._sender, msg, server[idx])
+                s_id = random.randint(machine[idx]*self._backup_count, (machine[idx]+1)*self._backup_count-1)
```
classicsong: Why we use random.randint here?

python/dgl/contrib/dis_kvstore.py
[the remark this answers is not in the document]
zheng-da: please write comment to explain why GC is needed.

examples/pytorch/dis_kvstore/server.py
zheng-da: another option is that the so-called "backup" server is created by the master server in the machine. In this way, the master server loads node/edge features in the file and the "backup" server gets the tensors automatically.

examples/pytorch/dis_kvstore/client.py
aksnzhy: The backup server need to wait the master server to create the shared-tensor. Both sleep(1) and sleep(2) are OK.

examples/pytorch/dis_kvstore/client.py
aksnzhy: It's only used for test.

python/dgl/contrib/dis_kvstore.py
aksnzhy: Sure, I will change that.

python/dgl/network.py
aksnzhy: I think garbage_msg is necessary for all the cases.

python/dgl/contrib/dis_kvstore.py
aksnzhy: I'm not sure which way is good. I just think that if we invoke _clear_kv_msg every time, the latency will be higher.

python/dgl/contrib/dis_kvstore.py
aksnzhy: We need to select a random server ID for communicating.

examples/pytorch/dis_kvstore/client.py
classicsong: Can we make it a check and go/fail model?

Issue #1264, opened 2020-02-15
multi_update_all with 'stack' reducer


gurdaspuriya:
## 🚀 Feature
1) Stack order is inconsistent, better to have it consistent.
2) The shape of the output is not consistent, when there is only one message-type (single frame), the same frame is returned; while for multiple message-types, an additional dimension is introduced.

BarclayII:
@gurdaspuriya We have submitted a patch.

One quick question: after you stack-reduce the per-type (reduced) messages, how would you like to use it (e.g. pass the whole thing through an MLP, or access each per-type reduced message individually)?

gurdaspuriya:
@BarclayII Thanks for taking care.
For my current use case, I would pass the whole thing through a MLP.

jermainewang:
@gurdaspuriya The patch has been merged. Please checkout the lastest master. Let us know if there are more issues.

PR #1267, opened 2020-02-17
[Bugfix] Consistent order for cross-type stack reducer

jermainewang:
## Description

Previously there is a bug in stack reducer. When stacking two type-wise tensors of shape, for example, (5, 20) and (5, 20), it accidentally expand an additional dimension, causing the result shape to be (5, 2, 1, 20). The correct shape should be (5, 2, 20).

Also use the internal edge type id as a hint for stacking order, making it consistent as long as the input graph does not change.

Issue #1264 .

## Checklist
- [x] Code is well-documented


BarclayII:
My question is that given a stack-reduced tensor, how could I get the subtensor corresponding to a given edge type?

Maybe it makes more sense to allocate a row for each edge type, even if the edge type does not exist as a neighbor?

PR #1143, opened 2019-12-25
[Example] Multi-GPU training by graph sampling

jermainewang:
## Description

Example codes for training GNNs on a single machine but multiple GPUs by sampling from a single graph data.

The example includes three models: GCN, GraphSAGE and GAT.

Results
---

Testbed: one AWS EC2 p3.16xlarge instance, 8 V100 GPUs, 64 VCPUs

**GCN:** `python main.py --gpu=0 --model=gcn --num-workers=4`

Accuracy: 93.87%

| | Speed (K samples/sec) |
| --- | --- |
| 1 GPU | 10.8 |
| 2 GPU | 24.1 |
| 4 GPU | 41.5 |
| 8 GPU | 69.1 |

**GraphSAGE:** `python main.py --gpu=0 --model=sage --num-workers=4`

Accuracy: 95.04%

| | Speed (K samples/sec) |
| --- | --- |
| 1 GPU | 10.8 |
| 2 GPU | 23.3 |
| 4 GPU | 40.5 |
| 8 GPU | 59.5 |

**GAT:** `python main.py --gpu=0 --model=gat --num-workers=4 --lr=0.01`

Accuracy: 93.7%

| | Speed (K samples/sec) |
| --- | --- |
| 1 GPU | 5.11 |
| 2 GPU | 10.4 |
| 4 GPU | 19.9 |
| 8 GPU | 33.1 |

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [x] To the my best knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change

Issue #1129, opened 2019-12-23
problem on distributed sampling example


hynsk:
Hello.

I'm testing distributed sampling of dgl according to https://github.com/dmlc/dgl/tree/master/examples/mxnet/sampling/dis_sampling

when I run
$ DGLBACKEND=mxnet python3 sampler.py --model gcn_ns --dataset cora --self-loop --num-neighbors 2 --batch-size 1000 --ip 127.0.0.1:50051 --num-sampler 1

The program is freezed.

For details, it freezed at 
https://github.com/dmlc/dgl/blob/master/python/dgl/contrib/sampling/sampler.py
line 32 : nfobjs = _CAPI_NeighborSampling

I think this is kind of parallel process problem.
Because in sampler.py code, it use SamplerPool abstract class that uses Pool for parallel process. And it isn't freezed when I modified MySamplerPool does not inherit SamplerPool class.

Can you let me know how to solve this problem?

Thank you.

aksnzhy:
@hynsk Hi, can you try to add `OMP_NUM_THREADS=1` in the head of your command line?

hynsk:
@aksnzhy 
OMP_NUM_THREADS=1 works well
Can you explain the reason and guide docs link about it?
Thank you for your help.

aksnzhy:
@hynsk This is a bug caused by openmp library. On some intel gcc compiler, the openmp library can cause a deadlock bug when using fork() for multiple processes. I will add this into the document.

Thanks so much!

hynsk:
@aksnzhy 
Thank you so much!

PR #1274, opened 2020-02-20
[WIP][Example] GraphSAGE model with 0.5 sampler and multi-GPU implementation

BarclayII:
## Description
A working end-to-end example of using the 0.5 sampler interface in a multi-GPU setting.

This is based on and compared against #1143 .

Several notes:

1. Worker processes should be created with `mp.Process` to make sure that the graph structure and data is not duplicated.  This leads to the following issue:
2. Neighbor samplers in worker processes freezes because of the long-standing deadlock bug in GNU OpenMP with `fork` (see #1129 ).  https://github.com/pytorch/pytorch/issues/17199 proposed a workaround and I adopted it in this example.  However, we need to confirm whether this also applies for MXNet and Tensorflow training (**help wanted**).

### Performance comparison

Here is a table computing the throughput of training examples processed per second.  The testbed is a p3.8xlarge AWS instance and the NodeFlow implementation is taken from #1143.

| Implementation | Average epoch time | Training example processed per second | Accuracy |
|:----:|:----:|:----:|:----:|
| NodeFlow | 33.72 | 6508 | 0.9553 |
| Current | 19.97 | 9067 | 0.9523 |
| NodeFlow, 4 GPUs | 9.093 | 24473 | 0.9604 |
| Current, 4 GPUs | 5.439 | 28342 | 0.9538 |

I believe the NodeFlow implementation can be optimized further, e.g. removing additional overheads in repeated computation of `IsMultigraph()` (@zheng-da )

### TODOs

- [ ] Unsupervised node classification example
- [ ] (Optional) Test on a dataset larger than Reddit (preferably one in Open Graph Benchmark, @mufeili any suggestions?)
- [ ] Tune the accuracy.  NodeFlow implementation and current implementation actually sample a little bit differently (for instance, the seed nodes has their neighbors resampled in the first layer in the current implementation, while the NodeFlow implementation reuses the ones in the second layer), and I wonder if this is the reason I had a minor performance drop.

## Checklist
- [ ] Code is well-documented


BarclayII:
Closing this since I'm basing on a wrong branch.

PR #1290, opened 2020-02-24
[DGL-KE] Distributed training of DGL-KE

aksnzhy:
## Description
This PR contains the main code of distributed training of DGL-KE.

TODO(chao):
1. Upload partitioned data to DGL s3 server.
2. Better launch script.

## Checklist
- [ ] Code is well-documented


apps/kg/dataloader/KGDataset.py
```diff
@@ -66,9 +66,11 @@ def __init__(self, path, name):
         self.n_entities = len(self.entity2id)
         self.n_relations = len(self.relation2id)
 
-        self.train = self.read_triple(path, 'train')
-        self.valid = self.read_triple(path, 'valid')
-        self.test = self.read_triple(path, 'test')
+        if read_triple == True:
```
zheng-da: why do you want to turn off reading triples?

apps/kg/dataloader/KGDataset.py
```diff
@@ -102,25 +104,32 @@ class KGDataset2:
...
-        with open(f_ent2id) as f_ent:
-            self.n_entities = int(f_ent.readline()[:-1])
         with open(f_rel2id) as f_rel:
             self.n_relations = int(f_rel.readline()[:-1])
 
-        self.train = self.read_triple(self.path, 'train')
-        self.valid = self.read_triple(self.path, 'valid')
-        self.test = self.read_triple(self.path, 'test')
+        if only_train == True:
+            f_ent2id = os.path.join(self.path, 'local_to_global.txt')
```
zheng-da: For distributed training, we should have a preprocessing step that converts all data in a partition into DGLGraph and serialize the data to a file.

apps/kg/distributed/fb15k_transe_l2.sh
```diff
@@ -0,0 +1,13 @@
+SERVER_ID_LOW=$1
+SERVER_ID_HIGH=$2
+
+while [ $SERVER_ID_LOW -lt $SERVER_ID_HIGH ]
+do
+    MKL_NUM_THREADS=1 OMP_NUM_THREADS=1 DGLBACKEND=pytorch python3 ../kvserver.py --model TransE_l2 --dataset FB15k \
+    --hidden_dim 400 --gamma 19.9 --lr 0.25 --total_client 64 --server_id $SERVER_ID_LOW &
+    let SERVER_ID_LOW+=1
```
zheng-da: do we want our users to invoke distributed training like this?

apps/kg/kvserver.py
```diff
@@ -0,0 +1,149 @@
...
+NUM_THREAD = 1 # Fix the number of threads to 1 on kvstore
+
+class KGEServer(KVServer):
+    """User-defined kvstore for DGL-KGE
+    """
+    def _push_handler(self, name, ID, data, target):
+        """Row-Sparse Adagrad updater
+        """
+        original_name = name[0:-6]
+        state_sum = target[original_name+'_state-data-']
```
zheng-da: how is `target` initialized?

apps/kg/train_pytorch.py
```diff
@@ -66,7 +130,10 @@ def train(args, model, train_sampler, valid_samplers=None, rank=0, rel_parts=Non
         backward_time += time.time() - start1
 
         start1 = time.time()
-        model.update(gpu_id)
+        if client is not None:
+            model.push_gradient(client)
```
zheng-da: is this asynchronous?

apps/kg/train_pytorch.py
```diff
@@ -57,6 +118,9 @@ def train(args, model, train_sampler, valid_samplers=None, rank=0, rel_parts=Non
         sample_time += time.time() - start1
         args.step = step
 
+        if client is not None:
+            model.pull_model(client, pos_g, neg_g)
```
zheng-da: is this asynchronous?

zheng-da:
my main comments are:
* we should have a preprocessing step and the distributed training only needs to deal with preprocessed data.
* can we simplify the launch script?

apps/kg/dataloader/KGDataset.py
classicsong: This local_to_global.txt is generated by the preprocessing progress?

apps/kg/dataloader/KGDataset.py
```diff
@@ -102,25 +104,32 @@ class KGDataset2:
...
+            with open(f_ent2id) as f_ent:
+                self.n_entities = len(f_ent.readlines())
+        else:
+            f_ent2id = os.path.join(self.path, 'entity2id.txt')
```
classicsong: I think, we need some docs for this entity2id.txt local2global.txt, etc.

apps/kg/distributed/luanch.sh
```diff
@@ -0,0 +1,8 @@
+SCRIPT_PATH=~/dgl/apps/kg/distributed
```
classicsong: Can you write a readme for these scripts ?

apps/kg/kvserver.py
```diff
@@ -0,0 +1,149 @@
...
+    args.format, 
+    machine_id)
+
+   # Note that the dataset doesn't ccontain the triple
+   print('n_entities: ' + str(dataset.n_entities))
+   print('n_relations: ' + str(dataset.n_relations))
+
+   model = load_model(None, args, dataset.n_entities, dataset.n_relations)
+
+   return g2l, model.entity_emb.emb, model.entity_emb.state_sum, model.relation_emb.emb, model.relation_emb.state_sum
```
classicsong: In this case, some model like transR are not supported by distributed training. 
They require additional embeddings.

apps/kg/models/general_models.py
```diff
@@ -462,3 +462,36 @@ def finish_async_update(self):
         """Terminate the async update for entity embedding.
         """
         self.entity_emb.finish_async_update()
+
+
+    def pull_model(self, client, pos_g, neg_g):
+        with th.no_grad():
+            entity_id = F.cat(seq=[pos_g.ndata['id'], neg_g.ndata['id']], dim=0)
+            relation_id = pos_g.edata['id']
+            entity_id = F.tensor(np.unique(F.asnumpy(entity_id)))
```
classicsong: I suggest we can implement F.unique by ourself.
torch has th.unique which will be efficient that converting it to numpy and back

apps/kg/train_pytorch.py
classicsong: push is async as I know.

apps/kg/train_pytorch.py
```diff
@@ -147,3 +214,128 @@ def train_mp(args, model, train_sampler, valid_samplers=None, rank=0, rel_parts=
...
+@thread_wrapped_func
+def dist_train_test(args, model, train_sampler, entity_pb, relation_pb, l2g, rank=0, rel_parts=None, cross_rels=None, barrier=None):
+    if args.num_proc > 1:
+        th.set_num_threads(args.num_thread)
+
+    client = connect_to_kvstore(args, entity_pb, relation_pb, l2g)
+    client.barrier()
+    train_time_start = time.time()
+    train(args, model, train_sampler, None, rank, rel_parts, cross_rels, barrier, client)
+    print('Total train time {:.3f} seconds'.format(time.time() - train_time_start))
```
classicsong: It is unfair to count train time without counting the barrier time.

apps/kg/train_pytorch.py
```diff
@@ -147,3 +214,128 @@ def train_mp(args, model, train_sampler, valid_samplers=None, rank=0, rel_parts=
...
+ 
+        print("Pull entity_emb ... ")
+        # split model into 100 small parts
+        start = 0
+        percent = 0
+        entity_id = F.arange(0, model_test.n_entities)
+        count = int(model_test.n_entities / 100)
+        end = start + count
+        while True:
+            print("Pull %d / 100 ..." % percent)
```
classicsong: use tqdm?

apps/kg/train_pytorch.py
```diff
@@ -147,3 +214,128 @@ def train_mp(args, model, train_sampler, valid_samplers=None, rank=0, rel_parts=
...
+            tmp_id = entity_id[start:end]
+            entity_data = client.pull(name='entity_emb', id_tensor=tmp_id)
+            model_test.entity_emb.emb[tmp_id] = entity_data
+            if end == -1:
+                break
+            start = end
+            end += count
+            percent += 1
+
+        if args.save_emb is not None:
```
classicsong: Does each client save the whole embedding on each machine?

apps/kg/train_pytorch.py
```diff
@@ -147,3 +214,128 @@ def train_mp(args, model, train_sampler, valid_samplers=None, rank=0, rel_parts=
...
+        percent = 0
+        entity_id = F.arange(0, model_test.n_entities)
+        count = int(model_test.n_entities / 100)
+        end = start + count
+        while True:
+            print("Pull %d / 100 ..." % percent)
+            if end >= model_test.n_entities:
+                end = -1
+            tmp_id = entity_id[start:end]
+            entity_data = client.pull(name='entity_emb', id_tensor=tmp_id)
```
classicsong: If the machine does not have enough memory to hold all embeddings, Can we still do the test?

apps/kg/dataloader/KGDataset.py
aksnzhy: For kvserver, we only need to know the data shape instead of the whole data triples.

apps/kg/dataloader/KGDataset.py
aksnzhy: Yes, this data should be gernerated by preprossing. We can also upload our partitioned data to S3 for users who just want to reproduce the result.

apps/kg/kvserver.py
aksnzhy: Target actually is `self.data_store`

apps/kg/train_pytorch.py
aksnzhy: Yes push is asnyc.

apps/kg/train_pytorch.py
aksnzhy: Pull is not async but I optimized it by overlapping the local computation and communication.

apps/kg/dataloader/KGDataset.py
aksnzhy: Yes, local_to_global.txt and partition_book.txt is needed to generated by preprocessing.

apps/kg/distributed/luanch.sh
aksnzhy: Sure.

apps/kg/kvserver.py
aksnzhy: Yes, it is.

apps/kg/train_pytorch.py
aksnzhy: You are right.

apps/kg/train_pytorch.py
aksnzhy: Yes

apps/kg/train_pytorch.py
aksnzhy: No, we cannot do test if the memory is not enough for that.

apps/kg/dataloader/KGDataset.py
aksnzhy: Yes, local_to_global.txt and partition_book.txt is needed to be generate by preprocessing.

apps/kg/train_pytorch.py
aksnzhy: That's a good idea. But I'm afraid that it will bring a new tqdm package installed by pip.

apps/kg/models/general_models.py
aksnzhy: I find that the numpy operation here is not a bottleneck compared with the network. So I think we can just use numpy for simplify.

PR #1295, opened 2020-02-28
[Feature] More serialization support

VoVAllen:
## Description
Fix the memory leak problem by support with shared_ptr
Now the serialization is done by writing/reading each class member, which makes the code more neat and easy to track. 

Two new api with class `Serializer` introduced in this PR, mainly for the calling of private default constructor. Default constructor is needed when reading from stream.
- `Serializer::new_object<T>()`, which returns the `*T`
- `Serializer::make_shared<T>()`, which returns `shared_ptr<T>`

## Checklist
- [ ] Code is well-documented


jermainewang:
@VoVAllen would you mind adding some PR description?

VoVAllen:
Updated

include/dgl/array.h
```diff
@@ -239,11 +242,38 @@ struct CSRMatrix {
...
+    return true;
+  }
+
+  void Save(dmlc::Stream* fs) const {
+    fs->Write(kDGLSerialize_AtenCsrMatrixMagic);
+    fs->Write(num_cols);
+    fs->Write(num_rows);
+    fs->Write(indptr);
+    fs->Write(indices);
+    fs->Write(data);
```
jermainewang: Just a question. Will this correctly handle the case of when `data` is initialized with default constructor? Note that default constructor gives undefined NDArray, whose internal pointer is null.

include/dgl/graph_serializer.h
```diff
@@ -0,0 +1,26 @@
...
+ * \file graph/graph_serializer.cc
+ * \brief DGL serializer APIs
+ */
+
+#ifndef DGL_GRAPH_SERIALIZER_H_
+#define DGL_GRAPH_SERIALIZER_H_
+
+#include <memory>
+namespace dgl {
+class Serializer {
```
jermainewang: Do you it will be easier if the default constructor of any serializable object is public?

src/graph/heterograph.cc
```diff
@@ -532,31 +535,20 @@ bool HeteroGraph::Load(dmlc::Stream* fs) {
...
-  for (size_t i = 0; i < num_relation_graphs; ++i) {
-    UnitGraph* ugptr = Serializer::EmptyUnitGraph();
-    CHECK(fs->Read(ugptr)) << "Invalid UnitGraph Data";
-    relgraphs.emplace_back(dynamic_cast<BaseHeteroGraph*>(ugptr));
-  }
-  HeteroGraph* hgptr = new HeteroGraph(GraphPtr(meta_grptr), relgraphs);
-  *this = *hgptr;
+  auto meta_imgraph = Serializer::make_shared<ImmutableGraph>();
+  CHECK(fs->Read(&meta_imgraph)) << "Invalid meta graph";
+  meta_graph_ = std::dynamic_pointer_cast<GraphInterface>(meta_imgraph);
```
jermainewang: The dynamic cast is not needed for parent class. Just `meta_graph_ = meta_imgraph` should be fine.

src/graph/heterograph.cc
```diff
@@ -4,18 +4,21 @@
...
+#include <dmlc/type_traits.h>
 #include <tuple>
 #include <utility>
+#include <vector>
 #include "../c_api_common.h"
 #include "./unit_graph.h"
-#include "graph_serializer.h"
+// TODO(BarclayII): currently CompactGraphs depend on IdHashMap implementation
+// which only works on CPU.  Should fix later to make it device agnostic.
+#include "../array/cpu/array_utils.h"
```
jermainewang: I think Quan moved `CompactGraphs` to another source file in PR #1249 . Please double check and delete these three lines.

src/graph/unit_graph.cc
```diff
@@ -401,7 +401,24 @@ class UnitGraph::COO : public BaseHeteroGraph {
            (NumVertices(SrcType()) > 1000000);
   }
 
+  bool Load(dmlc::Stream* fs) {
+    auto meta_imgraph = Serializer::make_shared<ImmutableGraph>();
+    CHECK(fs->Read(&meta_imgraph)) << "Invalid meta graph";
+    meta_graph_ = std::dynamic_pointer_cast<GraphInterface>(meta_imgraph);
```
jermainewang: ditto

src/graph/unit_graph.cc
```diff
@@ -738,7 +755,26 @@ class UnitGraph::CSR : public BaseHeteroGraph {
     return adj_;
   }
 
+  bool Load(dmlc::Stream* fs) {
+    auto meta_imgraph = Serializer::make_shared<ImmutableGraph>();
+    CHECK(fs->Read(&meta_imgraph)) << "Invalid meta graph";
+    meta_graph_ = std::dynamic_pointer_cast<GraphInterface>(meta_imgraph);
```
jermainewang: ditto

src/graph/unit_graph.cc
```diff
@@ -738,7 +755,26 @@ class UnitGraph::CSR : public BaseHeteroGraph {
...
+    CHECK(fs->Read(&sorted_)) << "Invalid boolean for sorted_";
+    CHECK(fs->Read(&adj_)) << "Invalid adj matrix";
+    return true;
+  }
+  void Save(dmlc::Stream* fs) const {
+    auto meta_graph_ptr = ImmutableGraph::ToImmutable(meta_graph());
+    fs->Write(meta_graph_ptr);
+    fs->Write(sorted_);
```
jermainewang: BTW, I don't think this `sorted_` flag is used anywhere. Would you help double check and delete it?

src/graph/unit_graph.cc
```diff
@@ -1219,46 +1255,74 @@ SparseFormat UnitGraph::SelectFormat(SparseFormat preferred_format) const {
...
+      fs->Write(GetInCSR());
+      break;
+    case SparseFormat::ANY:
+      // If not specified using CSR
+      fs->Write(GetOutCSR());
+      break;
+    default:
+      LOG(FATAL) << "unsupported format code";
+      break;
+  }
```
jermainewang: For `restrict_format_ == SparseFormat::ANY`, I think the a better solution is to use what is available instead of forcing CSR conversion. There is an interface called `SelectFormat` that returns the available format given some user preference. How about this:
```c++
fs->Write(meta_graph_ptr);
fs->Write(static_cast<int64_t>(restrict_format_));
auto avail_fmt = SelectFormat(SparseFormat::ANY);  // It assures avail_fmt != ANY
fs->Write(static_cast<int64_t>(avail_fmt));
switch (avail_fmt) {
  case SparseFormat::COO:
    // serialize COO
  case SparseFormat::CSR:
    // ...
  case SparseFormat::CSC:
    // ...
  default:
    LOG(FATAL) << ...;
}
```

src/graph/unit_graph.cc
```diff
@@ -401,7 +401,24 @@ class UnitGraph::COO : public BaseHeteroGraph {
...
+  void Save(dmlc::Stream* fs) const {
+    auto meta_graph_ptr = ImmutableGraph::ToImmutable(meta_graph());
+    fs->Write(meta_graph_ptr);
+    fs->Write(adj_);
+  }
+
  private:
+  friend class Serializer;
+
+  COO(){};
```
jermainewang: `COO() {}`
Add one space. No semicolon after a brace. There are several other places.

include/dgl/array.h
VoVAllen: It creates a new empty NDArray and set it to the current object
https://github.com/dmlc/dgl/blob/master/include/dgl/runtime/ndarray.h#L444-L464

include/dgl/graph_serializer.h
VoVAllen: Yes. Should I do so?

src/graph/heterograph.cc
VoVAllen: I'm also thinking here whether should make immutable graph as the default serialization of GraphInterface. (Add save/load function to the `GraphInterface`, which implemented using ImmutableGraph's save/load)

src/graph/heterograph.cc
VoVAllen: Removed

src/graph/unit_graph.cc
VoVAllen: This is added in https://github.com/dmlc/dgl/pull/886.
@zheng-da Is it necessary in this class?

include/dgl/graph_serializer.h
jermainewang: I think your solution is also fine, but please add more comments for this class such as code snippet of declaring it as a friend class, etc.

src/graph/heterograph.cc
jermainewang: You cannot avoid creating an explicit child class shared pointer in `Load`. It is essentially a problem of reflection. I think your solution here is fine.

src/graph/unit_graph.cc
jermainewang: Do note that the `sorted_` flag is now directly in the `CSRMatrix` structure, so technically it the one in `CSR` graph is redundant.

include/dgl/array.h
jermainewang: For `NDArray` directly created by default constructor, its internal `data_` pointer is null, so this line will fail:
https://github.com/dmlc/dgl/blob/c3a33407fe8b4938156427c701f0386dfc1a38de/include/dgl/runtime/ndarray.h#L416-L418
https://github.com/dmlc/dgl/blob/c3a33407fe8b4938156427c701f0386dfc1a38de/include/dgl/runtime/ndarray.h#L360-L362
I think the solution is to avoid such thing, which means we should not have the default argument of `data` to be `IdArray()`. I guess we could define a special ndarray to represent empty, which is equal to `NDArray({}, DLDataType{kDLInt, 32, 1}, DLContext{kDLCPU, 0})`. What do you think?

jermainewang: cc @BarclayII here too. We might need to do a pass on all the CSR/COO operators to make sure that the data array, when omitted, is correctly set to a defined empty array.

VoVAllen:
Another question:
`meta_graph` exists in both `UnitGraph` and `UnitGraph::CSR`(or other format). Are they the same?

include/dgl/array.h
BarclayII: Agree.  An undefined NDArray is probably a bad idea anyway.

In this case, it's probably better to have a static function
```C++
static NDArray NDArray::Empty(DLDataType dtype, DLContext ctx)
```

Although we already have a function named `Empty` for different purpose.

BarclayII:
They should be the same.

VoVAllen:
@jermainewang Addressed the comment above

jermainewang:
@VoVAllen for the metagraph pointers, we could avoid saving multiple same metagraphs by just storing the number of vtypes for UnitGraph::COO, UnitGraph::CSR. Upon loading, it checks the number and use the corresponding metagraph creator.

PR #1297, opened 2020-03-01
[Model][Sampler] GraphSAGE model, bipartite graph conversion & remove edges API

BarclayII:
## Description
We noticed that using `compact_graphs` for GraphSAGE on Reddit would cost much more memory than necessary.  Therefore we proposed a new `compact_as_bipartite` API, which converts any heterograph to a bipartite-structured graph.

Using the `compact_as_bipartite` API reduces GPU memory consumption from 1764MiB to 842.9 MiB on GraphSAGE/Reddit.  Training time also shrinked from 948s to 696s for 30 epochs.

The tentative changes include:

* `compact_as_bipartite()` API.
* `remove_edges()` API.
* New properties `sdata` and `ddata` for source and destination node data access on unidirectional bipartite graphs.
* New properties `srctype` and `dsttype` for accessing the name of node type for unidirectional bipartite graphs.
* Bipartite graph support for `nn.SAGEConv` module.
* GraphSAGE model.

TODO

- [x] Unit tests
- [x] Documentations
- [x] More thorough comments in example code (Maybe?  Please review)
- [ ] Tutorial (Postponed to future PR)
- [ ] Reorganize GraphSAGE example (Postponed to future PR)
- [ ] Unsupervised GraphSAGE model (Postponed to future PR)
- [x] Offline inference of GNN with full neighbors

### Example usage

These are excerpts from the `graphsage_sampling` example.

#### Sampler

```python
class NeighborSampler(object):
    def __init__(self, g, fanouts):
        self.g = g
        self.fanouts = fanouts

    def sample_frontiers(self, seeds):
        dataflow = []
        for fanout in self.fanouts:
            frontier = dgl.sampling.sample_neighbors(g, seeds, fanout)
            src, dst = frontier.all_edges()
            dataflow_piece = dgl.compact_as_bipartite(frontier, rhs_nodes=seeds, include_rhs_in_lhs=True)
            seeds = dataflow_piece.sdata[dgl.NID]
            dataflow.insert(0, dataflow_piece)
        return dataflow
```

The sampler now returns a series of bipartite graphs.  Each bipartite graph indicates the message passing dependency on the corresponding GNN layer.  Specifically, for each layer, after we sample the neighbors for the layer's "seed" nodes, we put the neighbors on the left side and the seed nodes on the right side as two separate node types.  Such bipartite structure enables us to compute representations more efficiently in terms of memory.

#### Model

```python
class SAGE(nn.Module):
    def __init__(self,
                 in_feats,
                 n_hidden, 
                 n_classes,
                 n_layers,
                 activation,
                 dropout):
        super().__init__()
        self.n_layers = n_layers
        self.layers = nn.ModuleList()
        self.layers.append(dglnn.SAGEConv(
            in_feats, n_hidden, 'mean', feat_drop=dropout, activation=activation))
        for i in range(1, n_layers - 1):
            self.layers.append(dglnn.SAGEConv(
                n_hidden, n_hidden, 'mean', feat_drop=dropout, activation=activation))
        self.layers.append(dglnn.SAGEConv(
            n_hidden, n_classes, 'mean', feat_drop=dropout))

    def forward(self, frontiers, x):
        h = x
        for layer, frontier in zip(self.layers, frontiers):
            h_dst = h[:frontier.number_of_nodes(frontier.dsttype)]
            h = layer(frontier, (h, h_dst))
        return h
```

Now `SAGEConv` supports bipartite graphs.  In this case, users have to supply two tensors: one for the source nodes, and another for destination nodes.

#### Minibatch training

```python
# sample neighbors
frontiers = sampler.sample_frontiers(seeds)
# get what are the input nodes necessary for computing the seed node representations.
induced_nodes = frontiers[0].sdata[dgl.NID]
# copy necessary inputs and labels to GPU
batch_inputs = g.ndata['features'][induced_nodes].to(dev_id)
batch_labels = labels[seeds].to(dev_id)
# forward
batch_pred = model(frontiers, batch_inputs)
# compute loss
loss = loss_fcn(batch_pred, batch_labels)
```

## Checklist
- [x] Code is well-documented


BarclayII:
@mufeili @classicsong Maybe you care about the user experience since you are working on OGB and GCMC respectively.

examples/pytorch/graphsage/graphsage_sampling.py
```diff
@@ -0,0 +1,301 @@
...
+            n_hidden, n_classes, 'mean', feat_drop=dropout))
+
+    def forward(self, blocks, x):
+        h = x
+        for layer, block in zip(self.layers, blocks):
+            # We need to first copy the representation of nodes on the RHS from the
+            # appropriate nodes on the LHS.
+            # Note that the shape of h is (num_nodes_LHS, D) and the shape of h_dst
+            # would be (num_nodes_RHS, D)
+            h_dst = h[:block.number_of_nodes(block.dsttype)]
```
jermainewang: Is this a forced convention that dsttype nodes are stored in [0, num_dst_nodes) ?

BarclayII: Yes.  It is stated in the docstring of `convert_as_bipartite`.

python/dgl/transform.py
```diff
@@ -714,6 +717,131 @@ def compact_graphs(graphs, always_preserve=None):
...
+            raise ValueError(
+                'Graph has more than one node type; please specify a dict for rhs_nodes.')
+        rhs_nodes = {g.ntypes[0]: rhs_nodes}
+
+    # rhs_nodes is now a dict
+    rhs_nodes = [rhs_nodes.get(ntype, F.tensor([], dtype=F.int64)) for ntype in g.ntypes]
+    rhs_nodes_nd = [F.zerocopy_to_dgl_ndarray(nodes) for nodes in rhs_nodes]
+
+    new_graph_index, lhs_nodes_nd, induced_edges_nd = _CAPI_DGLToBipartite(
+        g._graph, rhs_nodes_nd, include_rhs_in_lhs)
```
jermainewang: TIL a cleaner way of multiple returns.

python/dgl/transform.py
```diff
@@ -714,6 +717,131 @@ def compact_graphs(graphs, always_preserve=None):
...
+    for i, canonical_etype in enumerate(g.canonical_etypes):
+        induced_edges = F.zerocopy_from_dgl_ndarray(induced_edges_nd[i].data)
+        utype, etype, vtype = canonical_etype
+        new_canonical_etype = (utype + lhs_suffix, etype, vtype + rhs_suffix)
+        new_graph.edges[new_canonical_etype].data[EID] = induced_edges
+
+    return new_graph
+
+def remove_edges(g, edge_ids):
+    """Return a new graph with given edge IDs removed.
```
jermainewang: Does it relabel node ids?

src/array/cpu/coo_remove.cc
```diff
@@ -0,0 +1,48 @@
...
+template <DLDeviceType XPU, typename IdType>
+std::pair<COOMatrix, IdArray> COORemove(COOMatrix coo, IdArray entries) {
+  const int64_t nnz = coo.row->shape[0];
+  const IdType *row_data = static_cast<IdType *>(coo.row->data);
+  const IdType *col_data = static_cast<IdType *>(coo.col->data);
+  const IdType *eid_data = COOHasData(coo) ? static_cast<IdType *>(coo.data->data) : nullptr;
+
+  IdHashMap<IdType> eid_map(entries);
+
+  std::vector<IdType> new_rows, new_cols, new_eids;
```
jermainewang: `new_rows.reserve(nnz - entries->shape[0])`. It is the most common case and also a lower bound. Same for `new_cols` and `new_eids`.

src/array/cpu/coo_remove.cc
```diff
@@ -0,0 +1,48 @@
...
+
+  for (int64_t i = 0; i < nnz; ++i) {
+    const IdType eid = eid_data ? eid_data[i] : i;
+    if (eid_map.Contains(eid))
+      continue;
+    new_rows.push_back(row_data[i]);
+    new_cols.push_back(col_data[i]);
+    new_eids.push_back(eid);
+  }
```
jermainewang: In the most common case, the given `entries` are the positions to delete and the eid array is consecutive, which has a better implementation  when the entries array is small (sort entries + a linear scan). The current implementation has O(|NNZ|) number of hashmap queries, but when the entries array is small, the bloom filter in `IdHashMap` might perform good. Could you quickly test one such case to measure the gap?

src/graph/transform/remove_edges.cc
```diff
@@ -0,0 +1,88 @@
...
+namespace transform {
+
+std::pair<HeteroGraphPtr, std::vector<IdArray>>
+RemoveEdges(const HeteroGraphPtr graph, const std::vector<IdArray> &eids) {
+  std::vector<IdArray> induced_eids;
+  std::vector<HeteroGraphPtr> rel_graphs;
+  const int64_t num_etypes = graph->NumEdgeTypes();
+
+  for (int64_t etype = 0; etype < num_etypes; ++etype) {
+    const SparseFormat fmt = graph->SelectFormat(etype, SparseFormat::kAny);
```
jermainewang: This is a tricky. Actually, we want to pick whatever format that has consecutive eid array.

src/graph/transform/to_bipartite.cc
```diff
@@ -0,0 +1,131 @@
...
+template<typename IdType>
+std::tuple<HeteroGraphPtr, std::vector<IdArray>, std::vector<IdArray>>
+ToBipartite(
+    HeteroGraphPtr graph,
+    const std::vector<IdArray> &rhs_nodes,
+    bool include_rhs_in_lhs) {
+  std::vector<IdHashMap<IdType>> rhs_node_mappings, lhs_node_mappings;
+  const int64_t num_etypes = graph->NumEdgeTypes();
+  const int64_t num_ntypes = graph->NumVertexTypes();
+  std::vector<EdgeArray> edge_arrays;
```
jermainewang: Same for this edge_arrays.

src/graph/transform/to_bipartite.cc
```diff
@@ -0,0 +1,131 @@
...
+
+  CHECK(rhs_nodes.size() == static_cast<size_t>(num_ntypes))
+    << "rhs_nodes not given for every node type";
+
+  for (const IdArray &nodes : rhs_nodes)
+    rhs_node_mappings.emplace_back(nodes);
```
jermainewang: To make `rhs_node_mappings` local constant variables, I think this should work:
```c++
const std::vector<IdHashMap<IdType>> rhs_node_mappings(rhs_nodes.begin(), rhs_nodes.end());
```

examples/pytorch/graphsage/graphsage_sampling.py
```diff
@@ -0,0 +1,301 @@
...
+
+    def sample_blocks(self, seeds):
+        blocks = []
+        for fanout in self.fanouts:
+            # For each seed node, sample ``fanout`` neighbors.
+            frontier = dgl.sampling.sample_neighbors(g, seeds, fanout)
+            # Then we compact the frontier into a bipartite graph for message passing.
+            block = dgl.compact_as_bipartite(frontier, seeds, True)
+            # Obtain the seed nodes for next layer.
+            seeds = dataflow_piece.sdata[dgl.NID]
```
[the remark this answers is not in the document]
BarclayII: Yes.  The answer was that we may want to do further operations on the sampled frontier, such as removing the edges between positive/negative pairs.

George and Da further questioned that whether we should put an `exclude_neighbor` option in the `sample_neighbor` function instead.  What do you think of this option?

examples/pytorch/graphsage/graphsage_sampling.py
```diff
@@ -0,0 +1,301 @@
...
+            y = th.zeros(g.number_of_nodes(), self.n_hidden if l != len(self.layers) - 1 else self.n_classes)
+
+            for start in tqdm.trange(0, len(nodes), batch_size):
+                end = start + batch_size
+                batch_nodes = nodes[start:end]
+                block = dgl.compact_as_bipartite(dgl.in_subgraph(g, batch_nodes), batch_nodes, True)
+                induced_nodes = block.sdata[dgl.NID]
+
+                h = x[induced_nodes].to(device)
+                h_dst = h[:block.number_of_nodes(block.dsttype)]
```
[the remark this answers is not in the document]
BarclayII: The only place we could hide such fetches would be in NN modules.  The forward method would be responsible for judging whether the graph is
1. homogeneous (`layer(g, x)`),
2. bipartite (`layer(g, (x_src, x_dst))`)
3. something like a flow (`layer(block, x)`): `block` would have a flag indicating it is a block rather than a normal graph (e.g. by having a special attribute).

From my experience with control variate sampling the code in `forward` function per se is not very different.  The input of the module has one more argument `hbar` containing the history tensor for every layer, and the output of the module contains the results from every layer instead of one.  I can put another PR after this.  Or do you want to have it in this PR now?

src/array/cpu/coo_remove.cc
```diff
@@ -0,0 +1,48 @@
...
+
+  const COOMatrix new_coo = COOMatrix(
+      coo.num_rows, coo.num_cols,
+      IdArray::FromVector(new_rows),
+      IdArray::FromVector(new_cols));
+  return std::make_pair(new_coo, IdArray::FromVector(new_eids));
```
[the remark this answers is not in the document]
BarclayII: With our new convention `COOMatrix.data` should only contain the indices to some data, i.e. it should always be a permutation of 0..NNZ-1.  Putting `new_eids` into `data` breaks the convention.

include/dgl/transform.h
```diff
@@ -36,6 +36,55 @@ CompactGraphs(
...
+ *
+ * \param graph The graph.
+ * \param rhs_nodes Designated nodes that would appear on the right side.
+ * \param include_rhs_in_lhs Whether to always include the nodes on the right side within
+ *                           nodes on the left side.
+ *
+ * \return A triplet containing
+ *         * The bipartite-structured graph,
+ *         * The induced node from the left side for each graph,
+ *         * The induced edges.
```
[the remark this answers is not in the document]
BarclayII: How to infer from induced edges?

python/dgl/heterograph.py
```diff
@@ -542,6 +560,68 @@ def ndata(self):
...
+        See Also
+        --------
+        nodes
+        """
+        assert len(self.etypes) == 1, "Graph has more than one edge type."
+        srctype = self.canonical_etypes[0][0]
+        return HeteroNodeDataView(self, srctype, ALL)
+
+    @property
+    def ddata(self):
```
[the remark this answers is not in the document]
BarclayII: Lol.  Sure.  Will also change `sdata` to `srcdata` then.

python/dgl/nn/pytorch/conv/sageconv.py
```diff
@@ -29,6 +30,9 @@ class SAGEConv(nn.Module):
         Dropout rate on features, default: ``0``.
     aggregator_type : str
         Aggregator type to use (``mean``, ``gcn``, ``pool``, ``lstm``).
+    in_dst_feats : int, optional
+        If the layer is to be applied on a unidirectional bipartite graph,
+        specifies the input feature size on the destination nodes.
```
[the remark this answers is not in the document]
BarclayII: Good idea.

python/dgl/nn/pytorch/conv/sageconv.py
```diff
@@ -92,9 +98,11 @@ def forward(self, graph, feat):
...
         graph : DGLGraph
             The graph.
-        feat : torch.Tensor
-            The input feature of shape :math:`(N, D_{in})` where :math:`D_{in}`
-            is size of input feature, :math:`N` is the number of nodes.
+        feat : torch.Tensor or pair of torch.Tensor
+            If a torch.Tensor is given, the input feature of shape :math:`(N, D_{in})` where
+            :math:`D_{in}` is size of input feature, :math:`N` is the number of nodes.
+            If a pair of torch.Tensor is given, the pair must contain two tensors of shape
+            :math:`(N_{in}, D_{in})` and :math:`(N_{out}, D_{in})`.
```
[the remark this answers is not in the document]
BarclayII: They *should* be in different shapes in case of bipartite graphs.

The second `D_{in}` is a typo; it should be something like `D_{in_{dst}}`.  Will fix.

python/dgl/transform.py
BarclayII: No, the nodes are preserved; I should mention it.

src/array/cpu/coo_remove.cc
BarclayII: The most common case I can think of is removing the positive and negative pairs from a sampled frontier.  This is essentially done by:
```python
frontier = sample_neighbors(g, seeds)
pair_eids = g.edge_ids(heads, tails)
new_frontier = dgl.remove_edges(frontier, pair_eids)
```
So the edge IDs are not necessarily consecutive, and often times, `pair_eids` is a small array.  If we want to optimize something we should probably aim for this scenario instead.

I will still do the test though.

BarclayII: Will do.  The `eid` array being consecutive should help.

src/graph/transform/to_bipartite.cc
BarclayII: Thanks for bringing this syntax up.

examples/pytorch/graphsage/graphsage_sampling.py
```diff
@@ -0,0 +1,301 @@
...
+
+#### Neighbor sampler
+
+class NeighborSampler(object):
+    def __init__(self, g, fanouts):
+        self.g = g
+        self.fanouts = fanouts
+
+    def sample_blocks(self, seeds):
+        blocks = []
```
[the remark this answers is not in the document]
BarclayII: Then we need to define what is a flow?  Should we make the returned graph a subclass then?

examples/pytorch/graphsage/graphsage_sampling.py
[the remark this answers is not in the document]
jermainewang: Yeah, I agree with you. I don't see fusing the two operations would make it much faster and it would be more explicit in two.

src/array/cpu/coo_remove.cc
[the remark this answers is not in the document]
jermainewang: But according to L28, the `new_eids` is indeed the data to inherit from, no?

include/dgl/transform.h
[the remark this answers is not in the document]
jermainewang: Understood this better after reading the python-side code.

python/dgl/nn/pytorch/conv/sageconv.py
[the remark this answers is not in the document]
jermainewang: > The second `D_{in}` is a typo; it should be something like `D_{in_{dst}}`. Will fix.

Yeah, that's what I mean.

src/array/cpu/coo_remove.cc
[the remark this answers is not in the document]
BarclayII: I see.  Just to make sure, `COOMatrix.data` does not have to be a permutation of 0..NNZ-1, right?  As long as it contains the index of some data we are OK?

For instance, the original `COOMatrix.data` in this function is `[0, 1, 2, 3, 4]`, and `new_eids` here would return `[0, 1, 3, 4]`.  Is it still acceptable to put `new_eids` in `COOMatrix.data` even if we only have 4 entries there?

src/array/cpu/coo_remove.cc
BarclayII: Tried the setting of link prediction on Reddit.  The edges to be removed are the positive and negative training pairs.  Sorting entries consistently outperforms hash maps by around 20%.

Varied the number of negative examples, batch size, random walk window size, number of layers and number of neighbors.

mufeili:
Do we expect users to directly call `dgl.compact_as_bipartite`, i.e. a top-level API? If so, I find the function name as well as the arguments not very informative. If `compact_as_bipartite` will be a standard operation, then why not define a function like `construct_a_block`? Btw, calling it block might be more consistent with the components in NodeFlow.

examples/pytorch/graphsage/graphsage_sampling.py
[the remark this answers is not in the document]
mufeili: I'm more inclined to block, as we use block in NodeFlow.

examples/pytorch/graphsage/graphsage_sampling.py
```diff
@@ -0,0 +1,301 @@
...
+        for fanout in self.fanouts:
+            # For each seed node, sample ``fanout`` neighbors.
+            frontier = dgl.sampling.sample_neighbors(g, seeds, fanout)
+            # Then we compact the frontier into a bipartite graph for message passing.
+            block = dgl.compact_as_bipartite(frontier, seeds, True)
+            # Obtain the seed nodes for next layer.
+            seeds = block.sdata[dgl.NID]
```
classicsong: What does sdata short for? Seed data?
The naming is confusing to me.

BarclayII:
@mufeili We do expect users to directly call `dgl.compact_as_bipartite` if they were to write their own samplers.  We can rename it to `to_block` and specify that this data structure is solely for message passing purpose.

examples/pytorch/graphsage/graphsage_sampling.py
```diff
@@ -0,0 +1,301 @@
...
+    g : The entire graph.
+    inputs : The features of all the nodes.
+    labels : The labels of all the nodes.
+    val_mask : A 0-1 mask indicating which nodes do we actually compute the accuracy for.
+    batch_size : Number of nodes to compute at the same time.
+    device : The GPU device to evaluate on.
+    """
+    model.eval()
+    with th.no_grad():
+        pred = model.inference(g, inputs, batch_size, device)
```
classicsong: Why not send seed nodes from validation set into inference?

python/dgl/heterograph.py
```diff
@@ -363,6 +363,24 @@ def canonical_etypes(self):
...
+    @property
+    def srctype(self):
+        """Return the source node type if the graph has only one edge type."""
+        assert len(self.etypes) == 1
+        return self.canonical_etypes[0][0]
+
+    @property
+    def dsttype(self):
+        """Return the destination node type if the graph has only one edge type."""
+        assert len(self.etypes) == 1
```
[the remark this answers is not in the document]
classicsong: Agree. Otherwise the dsttype is confusing for heterograph.

python/dgl/nn/pytorch/conv/sageconv.py
```diff
@@ -103,29 +111,38 @@ def forward(self, graph, feat):
...
             graph.update_all(fn.copy_src('h', 'm'), fn.mean('m', 'neigh'))
-            h_neigh = graph.ndata['neigh']
+            h_neigh = graph.ddata['neigh']
         elif self._aggre_type == 'gcn':
-            graph.ndata['h'] = feat
+            graph.sdata['h'] = feat_src
+            graph.ddata['h'] = feat_dst     # same as above if homogeneous
             graph.update_all(fn.copy_src('h', 'm'), fn.sum('m', 'neigh'))
             # divide in_degrees
             degs = graph.in_degrees().float()
```
classicsong: I am just curious that if we doing sampling, how to define the degree here? the degree in bipartite graph (sampled) or the whole graph?

src/graph/transform/remove_edges.cc
```diff
@@ -0,0 +1,88 @@
...
+
+namespace transform {
+
+std::pair<HeteroGraphPtr, std::vector<IdArray>>
+RemoveEdges(const HeteroGraphPtr graph, const std::vector<IdArray> &eids) {
+  std::vector<IdArray> induced_eids;
+  std::vector<HeteroGraphPtr> rel_graphs;
+  const int64_t num_etypes = graph->NumEdgeTypes();
+
+  for (int64_t etype = 0; etype < num_etypes; ++etype) {
```
classicsong: Can we use omp here?

src/array/cpu/array_utils.h
```diff
@@ -30,10 +30,15 @@ class IdHashMap {
 
   // Construct the hashmap using the given id array.
   // The id array could contain duplicates.
+  // If the id array has no duplicates, the array will be relabeled to consecutive
+  // integers starting from 0.
```
classicsong: For hashmap. If it is expected to be a little bit large. It is better to preallocate the space for the hashtable other than let it grow during insertions.

examples/pytorch/graphsage/graphsage_sampling.py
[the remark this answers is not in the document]
jermainewang: Definitely another PR. We could leave it here and back to this once we have better solutions.

src/array/cpu/coo_remove.cc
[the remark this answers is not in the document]
jermainewang: Yes. I think we could treat it as the actual data array of a sparse matrix when we thinking about implementing an op. So remove will remove those data by the given positions but will keep the others.

examples/pytorch/graphsage/graphsage_sampling.py
BarclayII: I intended to do offline inference for all nodes.

src/graph/transform/remove_edges.cc
BarclayII: I think we would be using inter-batch (or batch-wise) multiprocessing instead of intra-batch (or op-level) multithreading.  @jermainewang found that launching OpenMP does not help much on performance presumably due to overhead of launching OpenMP contexts.

src/array/cpu/coo_remove.cc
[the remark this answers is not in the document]
BarclayII: OK.  In this case we should be careful when using CSR/COO with data to construct a graph, as the indices as edge IDs could be invalid.

python/dgl/transform.py
```diff
@@ -714,6 +717,131 @@ def compact_graphs(graphs, always_preserve=None):
 
     return new_graphs
 
+def compact_as_bipartite(g, rhs_nodes=None, include_rhs_in_lhs=False, lhs_suffix="_l",
+                         rhs_suffix="_r"):
+    """Convert a graph into a bipartite-structured graph for message passing.
+
+    Specifically, we create one node type ``ntype_l`` on the "left hand" side and another
```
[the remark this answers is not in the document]
BarclayII: We want to say that the induced nodes of each node types are a union of induced nodes from each edge type.  I don't think your statement could express this point (although I'm not sure whether mine does).

python/dgl/transform.py
```diff
@@ -714,6 +717,131 @@ def compact_graphs(graphs, always_preserve=None):
...
+    node type ``ntype_r`` on the "right hand" side for each node type ``ntype``.  The
+    nodes of type ``ntype_r`` would contain the nodes that are destination of an edge,
+    while ``ntype_l`` would contain the nodes that are the source nodes of an edge.
+
+    Edges connecting from node type ``utype`` to node type ``vtype`` are preserved,
```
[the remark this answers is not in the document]
BarclayII: Ditto.

python/dgl/transform.py
```diff
@@ -714,6 +717,131 @@ def compact_graphs(graphs, always_preserve=None):
...
+    except that the source node type and destination node type become ``utype_l`` and
+    ``vtype_r`` in the new graph.
+
+    If ``rhs_nodes`` is given, the right hand side would contain the given nodes instead.
+
+    If ``include_rhs_in_lhs`` is True, the left hand side would contain all the nodes
+    that appeared in the right hand side.
```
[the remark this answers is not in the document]
BarclayII: "Stored" doesn't look good; we are not actually storing the RHS nodes on the LHS.  I'll give an example instead.

src/graph/transform/remove_edges.cc
BarclayII: How do you recommend to do this?  Should I create another function called `SelectFormatWithConsecutiveID` or something like that?

python/dgl/nn/pytorch/conv/sageconv.py
BarclayII: The in degree would be those of destination nodes.

examples/pytorch/graphsage/graphsage_sampling.py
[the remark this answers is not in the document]
BarclayII: Changed to `to_block`.

examples/pytorch/graphsage/graphsage_sampling.py
```diff
@@ -0,0 +1,301 @@
...
+        self.g = g
+        self.fanouts = fanouts
+
+    def sample_blocks(self, seeds):
+        blocks = []
+        for fanout in self.fanouts:
+            # For each seed node, sample ``fanout`` neighbors.
+            frontier = dgl.sampling.sample_neighbors(g, seeds, fanout)
+            # Then we compact the frontier into a bipartite graph for message passing.
+            block = dgl.compact_as_bipartite(frontier, seeds, True)
```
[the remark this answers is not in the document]
BarclayII: Since I renamed to function to `to_block`, the last argument should be always True and I removed it as a result.

BarclayII:
PR is blocked by issue #1320 since blocks created from unidirectional bipartite graphs have node types with no inbound/outbound edge types.

src/graph/transform/remove_edges.cc
classicsong: I c

python/dgl/nn/pytorch/conv/sageconv.py
```diff
@@ -80,8 +93,8 @@ def _lstm_reducer(self, nodes):
         """
         m = nodes.mailbox['m'] # (B, L, D)
         batch_size = m.shape[0]
-        h = (m.new_zeros((1, batch_size, self._in_feats)),
-             m.new_zeros((1, batch_size, self._in_feats)))
+        h = (m.new_zeros((1, batch_size, self._in_src_feats)),
+             m.new_zeros((1, batch_size, self._in_src_feats)))
```
jermainewang: _in_dst_feats?

src/array/cpu/csr_remove.cc
```diff
@@ -0,0 +1,108 @@
...
+  std::sort(entry_data_sorted.begin(), entry_data_sorted.end());
+
+  int64_t k = 0;
+  new_indptr->push_back(0);
+  for (int64_t i = 0; i < csr.num_rows; ++i) {
+    for (IdType j = indptr_data[i]; j < indptr_data[i + 1]; ++j) {
+      if (k < n_entries && entry_data_sorted[k] == j) {
+        // Move on to the next different entry
+        while (k < n_entries && entry_data_sorted[k] == j)
+          ++j;
```
jermainewang: ++k?

src/graph/transform/remove_edges.cc
jermainewang: Let's put `SparseFormat::kCOO` as the preferred format for now as it usually has consecutive ids.

src/graph/unit_graph.cc
```diff
@@ -83,6 +83,7 @@ class UnitGraph::COO : public BaseHeteroGraph {
     : BaseHeteroGraph(metagraph), adj_(coo) {
     // Data index should not be inherited. Edges in COO format are always
     // assigned ids from 0 to num_edges - 1.
+    CHECK(!COOHasData(coo)) << "[BUG] COO should not contain data.";
```
jermainewang: Hmm... have you tried creating a sampled subgraph by COO?

python/dgl/nn/pytorch/conv/sageconv.py
BarclayII: The LSTM weights is of shape `(in_src_feats, in_src_feats)`.

src/graph/unit_graph.cc
BarclayII: I rewrote the subgraph construction here:
https://github.com/dmlc/dgl/blob/022ddb65464726125ac861f58d38f1fee29bfe42/src/graph/sampling/neighbor/neighbor.cc#L82-L84

PR #1299, opened 2020-03-01
[Feature] Improve sampling speed; Better pickle/unpickle; other fixes

jermainewang:
## Description

Performance improvement
* The old `RandomEngine::UniformChoice` is quite slow when the sampling number << the population size. Use `std::unorderd_set` for this case.
* The current pickle/unpickle is not bad but not great either. Specifically, if the graph is stored in CSR internally, it still converts it to COO. Improve this part by pickling what is available.
  - Expose two auxiliary data structures to achieve so: `HeteroPickleStates` and `SparseMatrix`.
  - Measured the speed of sending DGLHeteroGraph objects from one python process to another, to simulate the usage with torch's DataLoader. Testbed is EC2 x1.16xlarge. I launched 32 worker processes sending graphs to one main process. Each graph has 100K nodes and 1M edges, and are stored in int64. The result shows that current pickle/unpickle can reach 80% of pytorch's native performance. This gives a speed upper bound of sampling by multi-processing.

    Benchmark | objects / sec | edges / sec
    --- | --- | ---
    Xfer DGLHeterGraph | 169.76 | 169M
    Xfer edges in torch tensors | 175.13 | 175M
    Xfer raw torch tensor | 210.59 | 210M

New stuffs
* Add two functions `aten::NullArray` and `aten::IsNullArray`. `NullArray` is now the recommended way when specifying a placeholder NDArray (such as the `data` field in CSRMatrix and COOMatrix). Unlike `NDArray()`, NullArray is convertible to DLPack and backend tensors, causing less trouble when serializing/pickling them. Related to #1295 . @VoVAllen 
* Implemented `dgl.random.choice`.
* A new module `dgl.generators` for hosting all common graph generator functions. Currently implemented `dgl.rand_graph(num_nodes, num_edges)`.

Fixes
* Some warning fixes in object.h
* Split implementation to multiple source files for a cleaner organization.
* Changed `sample_neighbors` to have `replace` default be `False`.

## Checklist
- [x] Code is well-documented


tests/compute/test_pickle.py
```diff
@@ -262,9 +262,10 @@ def test_pickling_heterograph():
 
 
 if __name__ == '__main__':
-    test_pickling_index()
-    test_pickling_graph_index()
-    test_pickling_frame()
-    test_pickling_graph()
-    test_pickling_nodeflow()
-    test_pickling_batched_graph()
```
BarclayII: Why are we disabling these?

test.py
```diff
@@ -0,0 +1,277 @@
+import dgl
```
BarclayII: Remove this file?

python/dgl/_ffi/_ctypes/object.py
```diff
@@ -40,9 +40,10 @@ def _return_object(x):
 class ObjectBase(object):
     """Object base class"""
     __slots__ = ["handle"]
+
     # pylint: disable=no-member
     def __del__(self):
-        if _LIB is not None:
+        if _LIB is not None and hasattr(self, 'handle'):
```
VoVAllen: When would this doesn't have `handle`?

python/dgl/_ffi/object.py
```diff
@@ -52,21 +52,6 @@ def __reduce__(self):
...
         return (_new_object, (cls, ), self.__getstate__())
 
-    def __getstate__(self):
-        # TODO(minjie): TVM assumes that a Node (Object in DGL) can be serialized
-        #   to json. However, this is not true in DGL because DGL Object is meant
-        #   for runtime API, so it could contain binary data such as NDArray.
-        #   If this feature is required, please raise a RFC to DGL issue.
-        raise RuntimeError("__getstate__ is not supported for object type")
-
-    def __setstate__(self, state):
```
VoVAllen: Why remove this? The default pickle for C++ object might be only holding the handle address, which might cause other bugs.

python/dgl/heterograph.py
```diff
@@ -240,9 +235,16 @@ def __init__(self,
             frame.set_initializer(init.zero_initializer)
             self._msg_frames.append(frame)
 
-        self._is_multigraph = multigraph
+        self._is_multigraph = None
+
+    def __getstate__(self):
+        return self._graph, self._ntypes, self._etypes, self._node_frames, self._edge_frames
```
VoVAllen: We might want to rewrite the pickle implementation for `Frame` also, since it's only a dict and many members in it doesn't need to be involved in pickling

python/dgl/_ffi/_ctypes/object.py
jermainewang: Possible in a multi-processing setting. When a worker process creates an object by `__new__` but before calling `__init_handle_by_constructor` to create `handle`, it exits due to program exit, which will call `__del__` with no `handle` created.

python/dgl/heterograph.py
jermainewang: Yes but the actual performance doesn't matter too much. See the performance gap in the description to transmitting only two raw pytorch tensor.

tests/compute/test_pickle.py
jermainewang: Reverted.

test.py
jermainewang: Removed.

python/dgl/_ffi/object.py
jermainewang: reverted

python/dgl/random.py
```diff
@@ -13,4 +17,56 @@ def seed(val):
     """
     _CAPI_SetSeed(val)
 
+def choice(a, size, replace=True, prob=None):  # pylint: disable=invalid-name
+    """An equivalent to :func:`numpy.random.choice` but with more efficient implementation.
```
BarclayII: What was the performance difference between our `choice` and `np.random.choice`

src/array/cpu/array_utils.h
```diff
@@ -39,13 +39,12 @@ class IdHashMap {
...
     const IdType* ids_data = static_cast<IdType*>(ids->data);
     const int64_t len = ids->shape[0];
-    IdType newid = oldv2newv_.size();
     for (int64_t i = 0; i < len; ++i) {
       const IdType id = ids_data[i];
-      if (!Contains(id)) {
-        oldv2newv_[id] = newid++;
-        filter_[id & kFilterMask] = true;
-      }
+      // std::unorderd_map::insert assures that an insertion will not happen if the
```
BarclayII: unordered_map.

Also I think the membership check of `Contains(id)` for absent IDs should be generally faster than the one in `std::unordered_map`.  That's why we had this check in the first place.

Not sure how the speed gain was though.  Maybe @zheng-da could give some more insights on this?

src/graph/sampling/neighbor/neighbor.cc
```diff
@@ -41,14 +41,14 @@ HeteroSubgraph SampleNeighbors(
     const dgl_type_t dst_vtype = pair.second;
     const IdArray nodes_ntype = nodes[(dir == EdgeDir::kOut)? src_vtype : dst_vtype];
     const int64_t num_nodes = nodes_ntype->shape[0];
-    if (num_nodes == 0) {
-      // No node provided in the type, create a placeholder relation graph
+    if (num_nodes == 0 || fanouts[etype] == 0) {
+      // Nothing to sample for thie etype, create a placeholder relation graph
```
BarclayII: this

src/graph/sampling/neighbor/neighbor.cc
```diff
@@ -119,14 +115,14 @@ HeteroSubgraph SampleNeighborsTopk(
     const dgl_type_t dst_vtype = pair.second;
     const IdArray nodes_ntype = nodes[(dir == EdgeDir::kOut)? src_vtype : dst_vtype];
     const int64_t num_nodes = nodes_ntype->shape[0];
-    if (num_nodes == 0) {
-      // No node provided in the type, create a placeholder relation graph
+    if (num_nodes == 0 || k[etype] == 0) {
+      // Nothing to sample for thie etype, create a placeholder relation graph
```
BarclayII: this

include/dgl/array.h
```diff
@@ -31,12 +34,77 @@ typedef NDArray IntArray;
...
+  return NDArray::Empty({0}, DLDataType{kDLInt, 64, 1}, DLContext{kDLCPU, 0});
+}
+
+/*!
+ * \return Whether the input array is a null array.
+ */
+inline bool IsNullArray(NDArray array) {
+  // Although we recommend using NullArray(), sometimes people like to use
+  // an empty scalar array too.
+  return array->ndim == 0 || array->shape[0] == 0;
```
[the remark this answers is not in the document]
jermainewang: I think null arrays are internal contract. It shouldn't be visible by users. As a result, we could define the behavior. The `array->ndim == 0` is not necessary if everyone follows the contract.

src/array/cpu/array_utils.h
jermainewang: It's a very subtle difference. Previous codes could have two look ups (one in `Contains`, one in `oldv2new_[id]=newid++`, while this one only has one look up (in std::unordered_map::insert).

python/dgl/random.py
jermainewang: Good question. We should have this kind of thinking all the time during code review. Did some tests. The result is here:
```
.Draw 10% from    1000 (replace=1) | DGL  6846.37 | NP 26724.27 | DGL/NP     0.26
.Draw 10% from   10000 (replace=1) | DGL  6358.05 | NP 19804.12 | DGL/NP     0.32
.Draw 10% from  100000 (replace=1) | DGL  2554.80 | NP  7544.43 | DGL/NP     0.34
.Draw 10% from 1000000 (replace=1) | DGL   344.72 | NP  1828.84 | DGL/NP     0.19
.Draw 20% from    1000 (replace=1) | DGL  7893.40 | NP 28686.74 | DGL/NP     0.28
.Draw 20% from   10000 (replace=1) | DGL  5693.49 | NP 15070.31 | DGL/NP     0.38
.Draw 20% from  100000 (replace=1) | DGL  1512.24 | NP  4375.78 | DGL/NP     0.35
.Draw 20% from 1000000 (replace=1) | DGL   175.25 | NP   956.21 | DGL/NP     0.18
.Draw 40% from    1000 (replace=1) | DGL  7807.56 | NP 28380.26 | DGL/NP     0.28
.Draw 40% from   10000 (replace=1) | DGL  4395.09 | NP 10297.43 | DGL/NP     0.43
.Draw 40% from  100000 (replace=1) | DGL   836.02 | NP  2402.53 | DGL/NP     0.35
.Draw 40% from 1000000 (replace=1) | DGL    88.86 | NP   487.95 | DGL/NP     0.18
.Draw 80% from    1000 (replace=1) | DGL  6391.19 | NP 26715.31 | DGL/NP     0.24
.Draw 80% from   10000 (replace=1) | DGL  2982.14 | NP  6237.13 | DGL/NP     0.48
.Draw 80% from  100000 (replace=1) | DGL   441.51 | NP  1264.57 | DGL/NP     0.35
.Draw 80% from 1000000 (replace=1) | DGL    44.70 | NP   246.23 | DGL/NP     0.18
.Draw 10% from    1000 (replace=0) | DGL  6682.75 | NP 26085.69 | DGL/NP     0.26
.Draw 10% from   10000 (replace=0) | DGL  2935.02 | NP  5038.04 | DGL/NP     0.58
.Draw 10% from  100000 (replace=0) | DGL   463.65 | NP   501.47 | DGL/NP     0.92
.Draw 10% from 1000000 (replace=0) | DGL    40.91 | NP    48.42 | DGL/NP     0.85
.Draw 20% from    1000 (replace=0) | DGL  5676.86 | NP 26733.24 | DGL/NP     0.21
.Draw 20% from   10000 (replace=0) | DGL  1823.11 | NP  5029.14 | DGL/NP     0.36
.Draw 20% from  100000 (replace=0) | DGL   217.71 | NP   497.82 | DGL/NP     0.44
.Draw 20% from 1000000 (replace=0) | DGL    19.62 | NP    49.32 | DGL/NP     0.40
.Draw 40% from    1000 (replace=0) | DGL  4528.71 | NP 27357.29 | DGL/NP     0.17
.Draw 40% from   10000 (replace=0) | DGL  1016.96 | NP  5027.56 | DGL/NP     0.20
.Draw 40% from  100000 (replace=0) | DGL    99.84 | NP   499.70 | DGL/NP     0.20
.Draw 40% from 1000000 (replace=0) | DGL     9.63 | NP    50.15 | DGL/NP     0.19
.Draw 80% from    1000 (replace=0) | DGL  7678.91 | NP 27854.52 | DGL/NP     0.28
.Draw 80% from   10000 (replace=0) | DGL  4916.82 | NP  4914.69 | DGL/NP     1.00
.Draw 80% from  100000 (replace=0) | DGL  1344.87 | NP   495.89 | DGL/NP     2.71
.Draw 80% from 1000000 (replace=0) | DGL   162.35 | NP    49.88 | DGL/NP     3.26
```

Quite bad performance I would say. Will investigate. cc @classicsong .

jermainewang: Changed the report metric to average operation time, and add a test case with very small population size (10). It looks like for small operators, it is quite hard to improve because it hits the bottleneck of our FFI. It is quite interesting to see that even with `replace=True`, so the implementation is just generating several random integers, numpy is still much faster than our implementation.

jermainewang: Did another test by fixing the number of samples to draw.
```
 Draw 1000 from    1000 (replace=1) | DGL(ms)   0.1064 | NP(ms)   0.0336 | DGL/NP  3.16
.Draw 1000 from   10000 (replace=1) | DGL(ms)   0.1070 | NP(ms)   0.0445 | DGL/NP  2.40
.Draw 1000 from  100000 (replace=1) | DGL(ms)   0.1072 | NP(ms)   0.0382 | DGL/NP  2.81
.Draw 1000 from 1000000 (replace=1) | DGL(ms)   0.1061 | NP(ms)   0.0344 | DGL/NP  3.08
.Draw 1000 from    1000 (replace=0) | DGL(ms)   0.0856 | NP(ms)   0.0313 | DGL/NP  2.74
.Draw 1000 from   10000 (replace=0) | DGL(ms)   0.2853 | NP(ms)   0.1864 | DGL/NP  1.53
.Draw 1000 from  100000 (replace=0) | DGL(ms)   0.2833 | NP(ms)   1.9090 | DGL/NP  0.15
.Draw 1000 from 1000000 (replace=0) | DGL(ms)   0.2975 | NP(ms)  19.2240 | DGL/NP  0.02
```
The result is clear. **The overhead comes from memory allocation**. DGL's choice is much faster than numpy's when the population is large, probably because we use hashmap implementation while numpy uses shuffle. What do you guys think? Since the absolute operation running time is pretty small, shall we continue but remove the "more efficient implementation" claim?

classicsong: .Draw 40% from 1000000 (replace=0) | DGL(ms) 117.3515 | NP(ms)  20.4765 |
.Draw 60% from 1000000 (replace=0) | DGL(ms)  12.4088 | NP(ms)  22.1608

Why draw 40% is much slower than draw 60%?
And can we reduce the overhead of memory allocation?

include/dgl/array.h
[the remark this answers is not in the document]
BarclayII: Yeah.  I don't like the idea of making a scalar array "null".  Could we remove that part?

I would still prefer `array.Size() == 0` as a condition to make the judgement more rigorous.  Though `array->shape[0] == 0` also looks OK-ish.

python/dgl/random.py
BarclayII: > Since the absolute operation running time is pretty small, shall we continue but remove the "more efficient implementation" claim?

I would be very cautious going forward.  Even if the absolute runtime is small, stacking it a lot of times (during training etc.) would probably make the difference significant enough.

Also, it seems that DGL beats numpy on some cases of large population.

Now the question becomes: when would we want to use `dgl.choice`?  Do we have a guideline for this?

jermainewang: > Why draw 40% is much slower than draw 60%?

10% is the threshold for different implementation. <50% is hashmap-based. >50% is shuffle-based. Look like `std::unordered_set` is pretty slow when `num` is moderately large. I change the threshold to 10%, and here is the result:

```
.Draw 10% from      10 (replace=0) | DGL(ms)   0.0858 | NP(ms)   0.0176 | DGL/NP  4.87
.Draw 10% from    1000 (replace=0) | DGL(ms)   0.1113 | NP(ms)   0.0348 | DGL/NP  3.20
.Draw 10% from   10000 (replace=0) | DGL(ms)   0.3474 | NP(ms)   0.1924 | DGL/NP  1.81
.Draw 10% from  100000 (replace=0) | DGL(ms)   2.6537 | NP(ms)   1.9375 | DGL/NP  1.37
.Draw 10% from 1000000 (replace=0) | DGL(ms)  25.7389 | NP(ms)  20.2049 | DGL/NP  1.27
.Draw 20% from      10 (replace=0) | DGL(ms)   0.0867 | NP(ms)   0.0180 | DGL/NP  4.82
.Draw 20% from    1000 (replace=0) | DGL(ms)   0.1110 | NP(ms)   0.0358 | DGL/NP  3.10
.Draw 20% from   10000 (replace=0) | DGL(ms)   0.3330 | NP(ms)   0.1918 | DGL/NP  1.74
.Draw 20% from  100000 (replace=0) | DGL(ms)   2.5104 | NP(ms)   1.9390 | DGL/NP  1.29
.Draw 20% from 1000000 (replace=0) | DGL(ms)  24.1008 | NP(ms)  20.0152 | DGL/NP  1.20
.Draw 40% from      10 (replace=0) | DGL(ms)   0.0848 | NP(ms)   0.0179 | DGL/NP  4.74
.Draw 40% from    1000 (replace=0) | DGL(ms)   0.1111 | NP(ms)   0.0347 | DGL/NP  3.20
.Draw 40% from   10000 (replace=0) | DGL(ms)   0.2767 | NP(ms)   0.1926 | DGL/NP  1.44
.Draw 40% from  100000 (replace=0) | DGL(ms)   1.9557 | NP(ms)   1.9448 | DGL/NP  1.01
.Draw 40% from 1000000 (replace=0) | DGL(ms)  18.7165 | NP(ms)  20.1703 | DGL/NP  0.93
.Draw 60% from      10 (replace=0) | DGL(ms)   0.0851 | NP(ms)   0.0179 | DGL/NP  4.74
.Draw 60% from    1000 (replace=0) | DGL(ms)   0.0984 | NP(ms)   0.0351 | DGL/NP  2.80
.Draw 60% from   10000 (replace=0) | DGL(ms)   0.2058 | NP(ms)   0.1921 | DGL/NP  1.07
.Draw 60% from  100000 (replace=0) | DGL(ms)   1.2741 | NP(ms)   1.9425 | DGL/NP  0.66
.Draw 60% from 1000000 (replace=0) | DGL(ms)  11.8845 | NP(ms)  20.3029 | DGL/NP  0.59
.Draw 80% from      10 (replace=0) | DGL(ms)   0.0858 | NP(ms)   0.0178 | DGL/NP  4.82
.Draw 80% from    1000 (replace=0) | DGL(ms)   0.0925 | NP(ms)   0.0347 | DGL/NP  2.66
.Draw 80% from   10000 (replace=0) | DGL(ms)   0.1448 | NP(ms)   0.1919 | DGL/NP  0.75
.Draw 80% from  100000 (replace=0) | DGL(ms)   0.6634 | NP(ms)   1.9352 | DGL/NP  0.34
.Draw 80% from 1000000 (replace=0) | DGL(ms)   5.8648 | NP(ms)  20.5451 | DGL/NP  0.29
```
It looks much better.

> And can we reduce the overhead of memory allocation?

That might be a future work. We could try (i) implement  a memory pool or (ii) try more efficient lib like jemalloc.

jermainewang: Actually, it might be due to `RandInt` implementation. According to [Lemire's](https://lemire.me/blog/2019/09/28/doubling-the-speed-of-stduniform_int_distribution-in-the-gnu-c-library/), gnu's implementation has two division for each random which is slow and he showed that you could improve it by at least two-folds. Numpy seems to use the [same technique](https://github.com/numpy/numpy/blob/fdd8395f21a252373ff17fd43185c42e040c1b64/numpy/random/src/distributions/distributions.c#L1142). Look like a non-trivial task. Can we put a note and go? We need some end-to-end benchmark before over-optimization.

BarclayII: > DGL's choice is much faster than numpy's when the population is large, probably because we use hashmap implementation while numpy uses shuffle.

Not every case, right?  Because I saw:
* numpy consistently beats DGL in sampling with replacement.
* DGL beats numpy in sampling without replacement, but only on very small or very large samples.

BTW based on the drastic performance turn between "1000000, 40%" and "1000000, 60%", My feeling is that the crossover between "using `std::unordered_set`" and "using `TreeSampler`" is in fact much lower than 50% of the population.

src/array/cpu/array_utils.h
BarclayII: Make sense to me.

python/dgl/random.py
BarclayII: > Can we put a note and go? We need some end-to-end benchmark before over-optimization.

As long as we state when to use `dgl.choice` and when not to in docstrings I'm OK.  I imagine that in most cases (esp. in user code) we won't be using it.

jermainewang: Three cases I found it useful.
(i) Sample from a very large population and replace is False. Numpy will try to create a large range array and fail due to OOM. Our implementation avoids an explicit range array so it is runnable. It is used in `dgl.rand_graph`.
(ii) Sample from a backend tensor instead of a numpy array. In this case, we avoid copying it to numpy, performing sampling, and copying it back.
(iii) Used by our C++ implementation such as SampleNeighbors.

A relevant question is if @classicsong would like to implement an edge sampler in python, should he use this function or not? I would yes due to (ii) and we don't seem to have other options.

What do you think of the following action items?
(i) I will adjust the threshold to 10%.
(ii) I will modify the docstring to reflect these caveat.

include/dgl/array.h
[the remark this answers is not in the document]
jermainewang: sure

python/dgl/random.py
jermainewang: More benchmarks of non-uniform sampling
```
 Draw 10% from      10 (replace=1) | DGL(ms)   0.0069 | NP(ms)   0.0470 | DGL/NP  0.15
.Draw 10% from    1000 (replace=1) | DGL(ms)   0.0211 | NP(ms)   0.0700 | DGL/NP  0.30
.Draw 10% from   10000 (replace=1) | DGL(ms)   0.1804 | NP(ms)   0.2889 | DGL/NP  0.62
.Draw 10% from  100000 (replace=1) | DGL(ms)   2.7331 | NP(ms)   3.3288 | DGL/NP  0.82
.Draw 10% from 1000000 (replace=1) | DGL(ms)  37.3126 | NP(ms)  43.5458 | DGL/NP  0.86
.Draw 20% from      10 (replace=1) | DGL(ms)   0.0071 | NP(ms)   0.0481 | DGL/NP  0.15
.Draw 20% from    1000 (replace=1) | DGL(ms)   0.0304 | NP(ms)   0.0791 | DGL/NP  0.38
.Draw 20% from   10000 (replace=1) | DGL(ms)   0.3096 | NP(ms)   0.4117 | DGL/NP  0.75
.Draw 20% from  100000 (replace=1) | DGL(ms)   4.8102 | NP(ms)   5.6020 | DGL/NP  0.86
.Draw 20% from 1000000 (replace=1) | DGL(ms)  73.1789 | NP(ms)  77.1712 | DGL/NP  0.95
.Draw 40% from      10 (replace=1) | DGL(ms)   0.0075 | NP(ms)   0.0517 | DGL/NP  0.15
.Draw 40% from    1000 (replace=1) | DGL(ms)   0.0502 | NP(ms)   0.0984 | DGL/NP  0.51
.Draw 40% from   10000 (replace=1) | DGL(ms)   0.5611 | NP(ms)   0.6481 | DGL/NP  0.87
.Draw 40% from  100000 (replace=1) | DGL(ms)   8.8351 | NP(ms)   9.7920 | DGL/NP  0.90
.Draw 40% from 1000000 (replace=1) | DGL(ms) 138.1780 | NP(ms) 145.7263 | DGL/NP  0.95
.Draw 60% from      10 (replace=1) | DGL(ms)   0.0073 | NP(ms)   0.0463 | DGL/NP  0.16
.Draw 60% from    1000 (replace=1) | DGL(ms)   0.0707 | NP(ms)   0.1172 | DGL/NP  0.60
.Draw 60% from   10000 (replace=1) | DGL(ms)   0.8195 | NP(ms)   0.8847 | DGL/NP  0.93
.Draw 60% from  100000 (replace=1) | DGL(ms)  13.4447 | NP(ms)  14.1728 | DGL/NP  0.95
.Draw 60% from 1000000 (replace=1) | DGL(ms) 209.4729 | NP(ms) 202.8375 | DGL/NP  1.03
.Draw 80% from      10 (replace=1) | DGL(ms)   0.0075 | NP(ms)   0.0465 | DGL/NP  0.16
.Draw 80% from    1000 (replace=1) | DGL(ms)   0.0900 | NP(ms)   0.1359 | DGL/NP  0.66
.Draw 80% from   10000 (replace=1) | DGL(ms)   1.0746 | NP(ms)   1.1210 | DGL/NP  0.96
.Draw 80% from  100000 (replace=1) | DGL(ms)  17.6310 | NP(ms)  18.5213 | DGL/NP  0.95
.Draw 80% from 1000000 (replace=1) | DGL(ms) 274.4788 | NP(ms) 279.2041 | DGL/NP  0.98
.Draw 10% from      10 (replace=0) | DGL(ms)   0.0077 | NP(ms)   0.0841 | DGL/NP  0.09
.Draw 10% from    1000 (replace=0) | DGL(ms)   0.0234 | NP(ms)   0.1636 | DGL/NP  0.14
.Draw 10% from   10000 (replace=0) | DGL(ms)   0.2197 | NP(ms)   0.5796 | DGL/NP  0.38
.Draw 10% from  100000 (replace=0) | DGL(ms)   2.9467 | NP(ms)   5.8814 | DGL/NP  0.50
.Draw 10% from 1000000 (replace=0) | DGL(ms)  41.4497 | NP(ms)  75.7060 | DGL/NP  0.55
.Draw 20% from      10 (replace=0) | DGL(ms)   0.0073 | NP(ms)   0.0907 | DGL/NP  0.08
.Draw 20% from    1000 (replace=0) | DGL(ms)   0.0355 | NP(ms)   0.2039 | DGL/NP  0.17
.Draw 20% from   10000 (replace=0) | DGL(ms)   0.3856 | NP(ms)   0.9302 | DGL/NP  0.41
.Draw 20% from  100000 (replace=0) | DGL(ms)   5.6136 | NP(ms)  10.2525 | DGL/NP  0.55
.Draw 20% from 1000000 (replace=0) | DGL(ms)  87.4898 | NP(ms) 137.4428 | DGL/NP  0.64
.Draw 40% from      10 (replace=0) | DGL(ms)   0.0072 | NP(ms)   0.1019 | DGL/NP  0.07
.Draw 40% from    1000 (replace=0) | DGL(ms)   0.0609 | NP(ms)   0.2872 | DGL/NP  0.21
.Draw 40% from   10000 (replace=0) | DGL(ms)   0.7170 | NP(ms)   1.5686 | DGL/NP  0.46
.Draw 40% from  100000 (replace=0) | DGL(ms)  10.7577 | NP(ms)  20.2488 | DGL/NP  0.53
.Draw 40% from 1000000 (replace=0) | DGL(ms) 153.6011 | NP(ms) 287.6235 | DGL/NP  0.53
.Draw 60% from      10 (replace=0) | DGL(ms)   0.0073 | NP(ms)   0.1170 | DGL/NP  0.06
.Draw 60% from    1000 (replace=0) | DGL(ms)   0.0862 | NP(ms)   0.3583 | DGL/NP  0.24
.Draw 60% from   10000 (replace=0) | DGL(ms)   1.0501 | NP(ms)   2.3999 | DGL/NP  0.44
.Draw 60% from  100000 (replace=0) | DGL(ms)  15.9098 | NP(ms)  31.7715 | DGL/NP  0.50
.Draw 60% from 1000000 (replace=0) | DGL(ms) 245.1102 | NP(ms) 443.5592 | DGL/NP  0.55
.Draw 80% from      10 (replace=0) | DGL(ms)   0.0074 | NP(ms)   0.1356 | DGL/NP  0.05
.Draw 80% from    1000 (replace=0) | DGL(ms)   0.1099 | NP(ms)   0.4625 | DGL/NP  0.24
.Draw 80% from   10000 (replace=0) | DGL(ms)   1.3828 | NP(ms)   3.2645 | DGL/NP  0.42
.Draw 80% from  100000 (replace=0) | DGL(ms)  21.0640 | NP(ms)  45.6561 | DGL/NP  0.46
.Draw 80% from 1000000 (replace=0) | DGL(ms) 331.6010 | NP(ms) 636.4022 | DGL/NP  0.52
```
We are consistently faster than numpy.

Updated the docstring to include typical use cases. See if you guys are good with it. @classicsong @BarclayII .

BarclayII: > (i) Sample from a very large population and w/o replacement. Numpy will try to create a large range array and fail due to OOM. Our implementation avoids an explicit range array so it is runnable. It is used in `dgl.rand_graph`.

Agree.

> (ii) Sample from a backend tensor instead of a numpy array. In this case, we avoid copying it to numpy, performing sampling, and copying it back.

Just realized that MXNet does not have a sample-without-replacement function, so yes.

> (iii) Used by our C++ implementation such as SampleNeighbors.

I'm talking about the Python interface here so C++ implementation is irrelevant in this discussion.

> A relevant question is if @classicsong would like to implement an edge sampler in python, should he use this function or not? I would yes due to (ii) and we don't seem to have other options.

Agree.

> What do you think of the following action items?
> (i) I will adjust the threshold to 10%.
> (ii) I will modify the docstring to reflect these caveat.

Sounds good.

src/random/cpu/choice.cc
```diff
@@ -41,29 +42,29 @@ void RandomEngine::Choice(int64_t num, FloatArray prob, IdxType* out, bool repla
...
+void RandomEngine::UniformChoice(IdxType num, IdxType population, IdxType* out, bool replace) {
   if (!replace)
     CHECK_LE(num, population) << "Cannot take more sample than population when 'replace=false'";
   if (replace) {
-    for (int64_t i = 0; i < num; ++i)
+    for (IdxType i = 0; i < num; ++i)
       out[i] = RandInt(population);
   } else {
-    if (num < population / 2) {
+    if (num < population / 10) {
```
BarclayII: Should we leave a TODO here for future tuning?

jermainewang: Sure. Will do.

PR #1301, opened 2020-03-02
[Performance] Replace np.array with np.asarray

BarclayII:
## Description
It was observed that using `np.array` everywhere would sometimes double the memory consumption because `np.array` always copies the contents.  Similarly, `np.array(...).astype(...)` would also always copy the contents, even if the array is already a `np.ndarray` with the same dtype.

This PR replaces all `np.array` occurrences with `np.asarray`, which does not copy the contents if the array is already an `np.ndarray` with the designated dtype.

In the future, we should also establish a standard to always use `np.asarray` instead of `np.array`.

## Checklist
- [x] Code is well-documented


zheng-da:
is this good to merge?

BarclayII:
> LGTM. It looks like `asarray` is a no-brainer replacement of `array`. How about backend tensors? Does `astype` work similarly when the request dtype is already satisfied?

At least it is so for PyTorch.

PR #1303, opened 2020-03-03
[Doc] Fix the docstring of builtin functions.

zheng-da:
## Description
Provide better explanation of builtin functions.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,7 @@ endif()
 # and add set(OPTION VALUE) to override these build options.
 # Alernatively, use cmake -DOPTION=VALUE through command-line.
 dgl_option(USE_CUDA "Build with CUDA" OFF)
-dgl_option(USE_OPENMP "Build with OpenMP" OFF)
+dgl_option(USE_OPENMP "Build with OpenMP" ON)
 dgl_option(BUILD_CPP_TEST "Build cpp unittest executables" OFF)
 # Set debug compile option for gdb, only happens when -DCMAKE_BUILD_TYPE=DEBUG
 if (NOT MSVC)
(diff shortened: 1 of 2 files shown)


python/dgl/function/message.py
```diff
@@ -168,15 +168,17 @@ def copy_e(e, out):
 
 def _gen_message_builtin(lhs, rhs, binary_op):
     name = "{}_{}_{}".format(lhs, binary_op, rhs)
-    docstring = """Builtin message function that computes message by performing
-    binary operation {} between {} feature and {} feature.
+    docstring = """Builtin message function that computes a message on an edge
+    by performing element-wise {} between features of {} and {}
+    if the features have the same shape; otherwise, it first broadcasts the features
```
yzh119: I think we elaborate more on broadcasting behaviour: https://github.com/dmlc/dgl/blob/master/python/dgl/kernel.py#L75-L85

zheng-da: any reference to numpy broadcasting semantics?

CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ endif()
 # and add set(OPTION VALUE) to override these build options.
 # Alernatively, use cmake -DOPTION=VALUE through command-line.
 dgl_option(USE_CUDA "Build with CUDA" OFF)
-dgl_option(USE_OPENMP "Build with OpenMP" ON)
+dgl_option(USE_OPENMP "Build with OpenMP" OFF)
```
yzh119: Should we change this? I suppose we just need to customize this in `config.cmake`

python/dgl/function/message.py
yzh119: https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules

PR #1305, opened 2020-03-04
[Feature] Automatically determine the backend of dgl.nn

### LINKED ISSUE #1292: AttributeError: module 'dgl.nn' has no attribute 'pytorch'
## 🐛 Bug

When I try to use some of the predefined modules of dgl by the following code, I encounter an error: AttributeError: module 'dgl.nn' has no attribute 'pytorch'.

Similar problems also happen to other backends, including TensorFlow and MXNet.

## To Reproduce

```python
import dgl.nn
# or import dgl
c = dgl.nn.pytorch.conv.GraphConv(10,2)
```

## Expected behavior

The code should generate a GraphConv layer without any error.

## Environment

 - DGL Version (e.g., 1.0): 0.4.2
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): Pytorch 1.4.0
 - OS (e.g., Linux): Irrelelevent 
 - Python version: 3.7
 - CUDA/cuDNN version (if applicable): Irrelelevent 
 - GPU models and configuration (e.g. V100): Irrelelevent 

## Additional context

I read the source code and **found the reason and solution** to this problem.

### Reason:

The `__init__.py` file of `dgl/nn` is empty. Therefore, if i import dgl or dgl.nn, python cannot automatically find files of its sub-directories.

I verified it by the following code:
```python
import dgl.nn.python
c = dgl.nn.pytorch.conv.GraphConv(10,2)
```
It works fine.

### Solution:

Add 3 lines of code like `from . import pytorch` into file `dgl/nn/__init__.py` for PyTorch and the other 2 backends.
It is better to automatically detect the backend library and import the correct sub-directory.

BarclayII:
## Description

Fixes #1292 .

`import dgl` no longer imports `dgl.nn` automatically.


python/dgl/nn/__init__.py
```diff
@@ -1 +1,12 @@
...
+import importlib
+import sys
+
+def _load_backend(mod_name):
+    mod = importlib.import_module('.%s' % mod_name, __name__)
+    thismod = sys.modules[__name__]
+    for api, obj in mod.__dict__.items():
+        setattr(thismod, api, obj)
+
+_load_backend(os.environ.get('DGLBACKEND', 'pytorch').lower())
```
VoVAllen: using `dgl.backend.backend_name` instead

PR #1308, opened 2020-03-04
[Feature] add metis partitioning to DGL

zheng-da:
## Description
This adds a function to partition a DGLGraph into multiple graphs with Metis.
TODO items:
- [x] make a graph symmetric before calling Metis
- [ ] reorder nodes to have all node Ids in the same partition in a contiguous range.

## Checklist
- [ ] Code is well-documented


python/dgl/transform.py
```diff
@@ -581,6 +581,28 @@ def partition_graph_with_halo(g, node_part, num_hops):
         subg_dict[i] = subg
     return subg_dict
 
+def metis_partition(g, k):
```
VoVAllen: Is it better to set the partition id to node frame, and use another function to split into dict? Since I would assume the partition result could be used somewhere else.

zheng-da: the partition result is still preserved.

python/dgl/transform.py
```diff
@@ -575,6 +575,45 @@ def partition_graph_with_halo(g, node_part, num_hops):
...
     return subg_dict
 
+def metis_partition(g, k, num_hops=1):
+    '''
+    This is to partition a graph with Metis partitioning.
+
+    The partitioned graph is stored in DGLGraph. The DGLGraph has the `inner_node` node data
+    that indicates whether a node is inside a partition and has the `inner_edge` edge data
+    that indicates whether an edge is inside a partition. In addition, it has the `part_id`
+    node data that indicates the partition a node belongs to.
```
aksnzhy: So the `inner_node` can be used for `partition_book`? How to get the `local2global` tensor in this DGLGraph?

python/dgl/transform.py
```diff
@@ -575,6 +575,45 @@ def partition_graph_with_halo(g, node_part, num_hops):
...
+    if len(node_part) == 0:
+        return None
+
+    node_part = utils.toindex(node_part)
+    parts = partition_graph_with_halo(g, node_part, num_hops)
+    node_part = node_part.tousertensor()
+    for part_id in parts:
+        part = parts[part_id]
+        part.ndata['part_id'] = F.gather_row(node_part, part.parent_nid)
+    return parts
```
aksnzhy: In Tzy's code, if the head node and tail node belong to different partitions, the edge will be assigned to both of these two partitions. Do we have the same behavior here?

tools/partition.py
```diff
@@ -0,0 +1,56 @@
+import numpy as np
+import argparse
+import signal
+import dgl
+from dgl import backend as F
+from dgl.data.utils import load_graphs, save_graphs
+
+def main():
```
aksnzhy: I'm a little confused about the DGL_KE data preprocessing pipeline. The metis partition API needs a DGLGraph as input. However, the DGLGraph will be constructed by the dataloader. So we will invoke the dataloader before preprocessing?

python/dgl/transform.py
```diff
@@ -575,6 +575,45 @@ def partition_graph_with_halo(g, node_part, num_hops):
         subg_dict[i] = subg
     return subg_dict
 
+def metis_partition(g, k, num_hops=1):
+    '''
+    This is to partition a graph with Metis partitioning.
+
+    The partitioned graph is stored in DGLGraph. The DGLGraph has the `inner_node` node data
```
jermainewang: I feel when we say a graph is partitioned, we mean both its nodes and edges are partitioned. It looks like Metis partition only assigns nodes to partitions, so it leaves the edge assignment to our discretion. If that's the case, I think maybe the following description is better?
* The default behavior is storing in-edges to the nodes' partition.
    - The `inner_node` and `inner_edge` arrays are all one in this case.
    - You have the option to not even include them.
* Rename `num_hops` to `extra_cached_hops` and make its default be 0.
* Consider merge `inner_node` and `inner_edge` to `part_id` because for the i^th partition, the user could just use `th.nonzero(g.ndata['part_id'] == i)` to get the boolean array.

All these assume that the partitioning results in disjoint assignment, so if your actual implementation adopts node/edge mirror storage, the semantic will break.

python/dgl/transform.py
jermainewang: We should include this mapping in the `dgl.NID` and `dgl.EID` fields.

src/graph/graph_op.cc
```diff
@@ -653,4 +654,77 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLMapSubgraphNID")
     *rv = GraphOp::MapParentIdToSubgraphId(parent_vids, query);
   });
 
+#if !defined(WIN32) && !defined(_WIN32) && !defined(__WIN32__) && !defined(__NT__)
```
jermainewang: Let's move all the new codes to a separate file `graph/metis_partition.cc`

src/graph/graph_op.cc
```diff
@@ -653,4 +654,77 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLMapSubgraphNID")
...
+
+IdArray GraphOp::MetisPartition(GraphPtr g, int k) {
+  // The index type of Metis needs to be compatible with DGL index type.
+  assert(sizeof(idx_t) == sizeof(dgl_id_t));
```
jermainewang: CHECK_EQ

src/graph/graph_op.cc
```diff
@@ -653,4 +654,77 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLMapSubgraphNID")
...
+  ImmutableGraphPtr ig = std::dynamic_pointer_cast<ImmutableGraph>(g);
+  if (ig == NULL) {
```
jermainewang: CHECK(ig) << "The input graph isn't immutable graph";

src/graph/graph_op.cc
```diff
@@ -653,4 +654,77 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLMapSubgraphNID")
...
+    std::cerr << "The input graph isn't immutable graph" << std::endl;
+    return aten::NewIdArray(0);
+  }
+  // This is a symmetric graph, so in-csr and out-csr are the same.
+  CSRPtr csr = ig->GetInCSR();
+  auto mat = csr->ToCSRMatrix();
```
jermainewang: const auto& mat = ig->GetInCSR()->ToCSRMatrix();

src/graph/graph_op.cc
jermainewang: Also, is this `idx_t` from metis?

src/graph/graph_op.cc
```diff
@@ -653,4 +654,77 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLMapSubgraphNID")
...
+  idx_t nvtxs = g->NumVertices();
+  idx_t ncon = 1;        // # balacing constraints.
+  idx_t *xadj = static_cast<idx_t*>(mat.indptr->data);
+  idx_t *adjncy = static_cast<idx_t*>(mat.indices->data);
+  // TODO(zhengda) we could use this to create partitions that balance both #vertices and #edges.
+  idx_t *vwgt = NULL;
+  idx_t nparts = k;
+  IdArray part_arr = aten::NewIdArray(nvtxs);
+  idx_t objval = 0;
+  idx_t *part = static_cast<idx_t*>(part_arr->data);
```
jermainewang: Please use `const` as much as possible to distinguish between input and output arguments.

src/graph/graph_op.cc
```diff
@@ -653,4 +654,77 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLMapSubgraphNID")
...
+  // This is a symmetric graph, so in-csr and out-csr are the same.
+  CSRPtr csr = ig->GetInCSR();
+  auto mat = csr->ToCSRMatrix();
+
+  idx_t nvtxs = g->NumVertices();
+  idx_t ncon = 1;        // # balacing constraints.
+  idx_t *xadj = static_cast<idx_t*>(mat.indptr->data);
+  idx_t *adjncy = static_cast<idx_t*>(mat.indices->data);
+  // TODO(zhengda) we could use this to create partitions that balance both #vertices and #edges.
+  idx_t *vwgt = NULL;
```
jermainewang: Is this used anywhere?

src/graph/graph_op.cc
```diff
@@ -653,4 +654,77 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLMapSubgraphNID")
...
+                                NULL,        // The weights of the edges
+                                &nparts,     // The number of partitions.
+                                NULL,        // the desired weight for each partition and constraint
+                                NULL,        // the allowed load imbalance tolerance
+                                NULL,        // the array of options
+                                &objval,      // the edge-cut or the total communication volume of
+                                // the partitioning solution
+                                part);
+  printf("Partition a graph with %ld nodes and %ld edges into %d parts and get %ld edge cuts\n",
+         g->NumVertices(), g->NumEdges(), k, objval);
```
jermainewang: LOG(INFO)

src/graph/graph_op.cc
```diff
@@ -653,4 +654,77 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLMapSubgraphNID")
...
+    return part_arr;
+  } else if (ret == METIS_ERROR_INPUT) {
+    std::cerr << "Error in Metis partitioning: input error" << std::endl;
+  } else if (ret == METIS_ERROR_MEMORY) {
+    std::cerr << "Error in Metis partitioning: cannot allocate memory" << std::endl;
+  } else {
+    std::cerr << "Error in Metis partitioning: other errors" << std::endl;
+  }
+  // return an array of 0 elements to indicate the error.
+  return aten::NewIdArray(0);
```
jermainewang:
```c++
if (ret != METIS_OK) {
    switch (ret) {
      case METIS_ERROR_INPUT:
        LOG(FATAL) << "...";
      case METIS_ERROR_MEMORY:
        LOG(FATAL) << "...";
      default:
        LOG(FATAL) << "...";
    }
}
return part_arr;
```

src/graph/graph_op.cc
```diff
@@ -653,4 +654,77 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLMapSubgraphNID")
...
+    *rv = GraphOp::MetisPartition(g.sptr(), k);
+  });
+
+#else
+
+DGL_REGISTER_GLOBAL("transform._CAPI_DGLMetisPartition")
+.set_body([] (DGLArgs args, DGLRetValue* rv) {
+    GraphRef g = args[0];
+    int k = args[1];
+    LOG(WARNING) << "DGL doesn't support METIS partitioning in Windows";
```
jermainewang: Let's just throw error. No need to warn.

src/graph/graph_op.cc
```diff
@@ -653,4 +654,77 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLMapSubgraphNID")
...
+    *rv = aten::NewIdArray(0);
```
jermainewang: There is a new API `aten::NullArray()` to get an array representing null.

src/graph/graph_op.cc
jermainewang: Is it necessary to guard so many macros? In [dmlc-core](https://github.com/dmlc/dmlc-core/blob/692216e8633eacae001a023ebbdcafbad0985854/include/dmlc/filesystem.h), we only guard _WIN32.

src/graph/graph_op.cc
```diff
@@ -653,4 +654,77 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLMapSubgraphNID")
...
+                                NULL,        // The size of the vertices for computing
+                                // the total communication volume
+                                NULL,        // The weights of the edges
+                                &nparts,     // The number of partitions.
+                                NULL,        // the desired weight for each partition and constraint
+                                NULL,        // the allowed load imbalance tolerance
+                                NULL,        // the array of options
+                                &objval,      // the edge-cut or the total communication volume of
+                                // the partitioning solution
+                                part);
```
jermainewang: I wonder does METIS use a zero-based CSR or a one-based?

python/dgl/transform.py
```diff
@@ -575,6 +575,45 @@ def partition_graph_with_halo(g, node_part, num_hops):
...
+    num_hops: int
+        The number of hops a HALO node can be accessed.
+
+    Returns
+    --------
+    a dict of DGLGraphs
+        The key is the partition Id and the value is the DGLGraph of the partition.
+    '''
+    # METIS works only on symmetric graphs.
+    g = to_bidirected(g, readonly=True)
```
jermainewang: Since this might be costly. Could you add a argument `is_bidirected=False`? So if the user already has a bidirectional graph, there is no need to perform conversion again.

zheng-da: I benchmarked the one in GKLib. It's much faster. For a graph with (|V|=10000000, |E|=300000000), The GKLib implementation takes about 60 seconds with one thread. It doesn't look like it's a bottleneck.

src/graph/graph_op.cc
zheng-da: zero-based mean the first element in indptr is 0?

src/graph/graph_op.cc
zheng-da: is the exception captured in python or just crash the python process?

src/graph/graph_op.cc
zheng-da: METIS API doesn't use const. I cannot make it const here.

src/graph/graph_op.cc
zheng-da: yes

include/dgl/graph_op.h
```diff
@@ -149,6 +149,15 @@ class GraphOp {
    * \return the induced subgraph with HALO nodes.
    */
   static HaloSubgraph GetSubgraphWithHalo(GraphPtr graph, IdArray nodes, int num_hops);
+
+  /*!
+   * \brief Partition a graph with Metis.
+   * The partitioning algorithm assigns each vertex to a partition.
```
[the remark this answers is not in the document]
zheng-da: METIS only assigns nodes to partitions.

python/dgl/transform.py
```diff
@@ -575,6 +575,45 @@ def partition_graph_with_halo(g, node_part, num_hops):
         subg_dict[i] = subg
     return subg_dict
 
+def metis_partition(g, k, num_hops=1):
+    '''
```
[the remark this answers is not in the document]
zheng-da: what do you mean?

python/dgl/transform.py
zheng-da: `part_id` can be used to generate `partition_book`. `parent_nid` can be used to generate `local2global`.

zheng-da: `dgl.NID` field stores `parent_nid` in a subgraph?

python/dgl/transform.py
zheng-da: it only includes in-edges. I can change its behavior. Isn't Tzy's implementation replicates edges?

tools/partition.py
zheng-da: yes, we call data loader to construct DGLGraph and then invoke metis partitioning to partition the graph.

python/dgl/transform.py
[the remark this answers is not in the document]
jermainewang: Usually we have:
```python
def foo():
    '''Brief description
    
    More descriptions
    '''
```

PR #1309, opened 2020-03-04
[NN] Fix RelGraphConv num_bases

simonepri:
## Description
When `num_bases` is none the following exception is thrown:

```
  File "/content/code/.venv/lib/python3.7/site-packages/dgl/nn/pytorch/conv/relgraphconv.py", line 94, in __init__
    if in_feat % num_bases != 0 or out_feat % num_bases != 0:
TypeError: unsupported operand type(s) for %: 'int' and 'NoneType'
```
The same applies when `num_bases > self.num_rels` or `self.num_bases < 0`.
Instead, when `num_bases` is `0` then `in_feat // self.num_bases` it throws a modulo division by zero exception.

Also when `num_bases > self.num_rels`, `num_bases` is modified internally and the exception message does not provide enough information to adjust the value of `in_feat` or `out_feat`, so I'm also piggybacking a small change in the exception message to clarify the value of `num_bases` that is used internally.


VoVAllen:
@yzh119

simonepri:
There is also another bug:

In `__init__` we do:
```python
self.submat_in = in_feat // self.num_bases
```

then in `bdd_message_func` the following operation is done:
```python
# edge.shape = [batch_size, in_feat]
edges.src['h'].view(-1, 1, self.submat_in)
```

which throws an exception if `self.submat_in` is not a multiple of `batch_size * in_feat`.

Given that `batch_size` can change at runtime, either we add an assert inside `bdd_message_func` or we enforce `in_feat` to be a multiple of `submat_in` as well.

Should I open another PR or can I just fix everything in this one?

VoVAllen:
Thanks for your contribution!

PR #1323, opened 2020-03-06
[Bugfix] Allows node types with no inbound/outbound edges

### LINKED ISSUE #1320: [BUG] dgl.heterograph() not supporting node types with no inbound/outbound edges
## 🐛 Bug

See below:

## To Reproduce

```
In [2]: g = dgl.heterograph({('A', 'AB', 'B'): [(0, 1), (1, 2)]}, num_nodes_dict={'A': 3, 'B': 3, 'C': 3})

In [3]: g
Out[3]: 
Graph(num_nodes={'A': 3, 'B': 3},
      num_edges={('A', 'AB', 'B'): 2},
      metagraph=[('A', 'B')])
```

The node type `C` is missing.

## Expected behavior

Node type `C` should be there even if no edges connect to/from C.

## Environment

 - OS (e.g., Linux):
 - Python version:

## Additional context

A similar issue exist in the C API `CreateHeteroGraph` where we have no way to specify the number of nodes of a node type without any inbound/outbound edges.

BarclayII:
## Description
This fixes #1320 by adding new ctor:
```c++
CreateHeteroGraph(meta_graph, rel_graphs, num_nodes_per_type)
```
where the third argument is a `std::vector` containing the number of nodes for each node type.

## Checklist
- [x] Code is well-documented


include/dgl/base_heterograph.h
```diff
@@ -547,6 +550,25 @@ DGL_DEFINE_OBJECT_REF(FlattenedHeteroGraphRef, FlattenedHeteroGraph);
...
+
+HeteroGraphPtr CreateHeteroGraph(
+    GraphPtr meta_graph,
+    const std::vector<HeteroGraphPtr>& rel_graphs,
+    std::vector<int64_t>&& num_nodes_per_type);
+
+HeteroGraphPtr CreateHeteroGraph(
+    GraphPtr meta_graph,
+    const std::vector<HeteroGraphPtr>& rel_graphs,
+    IdArray num_nodes_per_type);
```
jermainewang: The second one is weird to me. Right value reference is for mutating the argument w/ no copy which quite contradicts with the one with `const vector<>&`. In terms of using vector or IdArray, I think picking either one is fine. My practice is using IdArray if the value will participate in FFI. So my suggestion is using IdArray.

src/runtime/ndarray.cc
```diff
@@ -263,6 +263,44 @@ GEN_FROMVECTOR_FOR(uint64_t, kDLInt, 64);
 GEN_FROMVECTOR_FOR(float, kDLFloat, 32);
 GEN_FROMVECTOR_FOR(double, kDLFloat, 64);
 
+template<typename T>
+std::vector<T> NDArray::ToVector(DLDataType dtype) const {
+  CHECK(data_->dl_tensor.ndim == 1) << "ToVector() only supported for 1D arrays";
+  CHECK(data_->dl_tensor.dtype.code == dtype.code &&
+        data_->dl_tensor.dtype.bits == dtype.bits &&
+        data_->dl_tensor.dtype.lanes == dtype.lanes) << "dtype mismatch";
```
jermainewang: Do we have an `==` operator overload?

python/dgl/convert.py
```diff
@@ -441,8 +453,11 @@ def heterograph(data_dict, num_nodes_dict=None):
                     nsrc = len({n for n, d in data.nodes(data=True) if d['bipartite'] == 0})
                     ndst = data.number_of_nodes() - nsrc
             elif isinstance(data, DGLHeteroGraph):
-                # Do nothing; handled in the next loop
-                continue
+                # original node type and edge type of ``data`` is ignored.
+                assert len(data.canonical_etypes) == 1, \
+                    "Relational graphs must have only one edge type."
+                nsrc = data.number_of_nodes(data.canonical_etypes[0][0])
+                ndst = data.number_of_nodes(data.canonical_etypes[0][2])
```
jermainewang: Slightly cleaner:
```python
srctype, _, dsttype = data.canonical_etypes[0]
nsrc = data.number_of_nodes(srctype)
ndst = data.number_of_nodes(dsttype)
```

python/dgl/graph_index.py
```diff
@@ -1172,8 +1172,8 @@ def from_edge_list(elist, is_multigraph, readonly):
         src, dst = elist
     else:
         src, dst = zip(*elist)
-    src = np.asarray(src)
-    dst = np.asarray(dst)
+    src = np.array(src)
+    dst = np.array(dst)
```
jermainewang: Why change it back?

include/dgl/base_heterograph.h
jermainewang: We can also extend the original interface as this right?
```c++
HeteroGraphPtr CreateHeteroGraph(
    GraphPtr meta_graph,
    const std::vector<HeteroGraphPtr>& rel_graphs,
    const IdArray num_nodes_per_type = aten::NullArray());
```

src/graph/heterograph.cc
```diff
@@ -136,30 +141,61 @@ HeteroGraph::HeteroGraph(GraphPtr meta_graph, const std::vector<HeteroGraphPtr>&
...
-      num_verts_per_type_[dsttype] = nv;
+    if (num_verts_per_type[dsttype] < 0)
+      num_verts_per_type[dsttype] = nv;
     else
-      CHECK_EQ(num_verts_per_type_[dsttype], nv)
+      CHECK_EQ(num_verts_per_type[dsttype], nv)
         << "Mismatch number of vertices for vertex type " << dsttype;
   }
+  return num_verts_per_type;
+}
```
jermainewang: This part seems to be a duplicate to some codes in the python side. Can we just use the C++ side?

src/graph/transform/union_partition.cc
```diff
@@ -12,6 +12,7 @@ HeteroGraphPtr DisjointUnionHeteroGraph(
     GraphPtr meta_graph, const std::vector<HeteroGraphPtr>& component_graphs) {
   CHECK_GT(component_graphs.size(), 0) << "Input graph list is empty";
   std::vector<HeteroGraphPtr> rel_graphs(meta_graph->NumEdges());
+  std::vector<int64_t> num_nodes_per_type(meta_graph->NumVertices());
```
jermainewang: Initialize them to 0? What if the loop over the relation graph does not set some of the num_nodes?

src/graph/transform/compact.cc
```diff
@@ -57,8 +57,11 @@ CompactGraphs(
 
   // Step 2: Relabel the nodes for each type to a smaller ID space and save the mapping.
   std::vector<IdArray> induced_nodes;
-  for (auto &hashmap : hashmaps)
+  std::vector<int64_t> num_induced_nodes;
```
jermainewang: Reserve the size of these two vectors.

jermainewang:
Also, could you please double check that the newly added member is properly handled during serialization (merged in #1295) ?

include/dgl/base_heterograph.h
BarclayII: I don't think r-value reference is for mutating the argument (for which we should always use pointers).  The r-value reference here is for implementing moves, i.e. replacing `num_verts_per_type_` member with the given one without any copying.
Also, IdArray would be converted to `std::vector` anyway because the member `num_verts_per_type_` is a `std::vector`.

src/graph/transform/union_partition.cc
BarclayII: Does `std::vector<int64_t>` initialize the contents to 0 by default?

src/runtime/ndarray.cc
VoVAllen: https://github.com/dmlc/dgl/blob/master/src/c_api_common.h#L20

BarclayII: I moved both overloads to `dgl/runtime/ndarray.h`

python/dgl/graph_index.py
BarclayII: Not intended.  Reverted.

src/graph/heterograph.cc
BarclayII: Python side removed.

BarclayII:
> Also, could you please double check that the newly added member is properly handled during serialization (merged in #1295) ?

I think so.  #1295 already serializes the number of nodes per type before I knew it.

src/graph/transform/union_partition.cc
jermainewang: No. std::vector<T> initializes values by T(), but the initial value for native int is undefined in C++ standard.

src/graph/heterograph.cc
```diff
@@ -136,30 +141,61 @@ HeteroGraph::HeteroGraph(GraphPtr meta_graph, const std::vector<HeteroGraphPtr>&
...
+  relation_graphs_ = CastToUnitGraphs(rel_graphs);
+}
+
+HeteroGraph::HeteroGraph(
+    GraphPtr meta_graph,
+    const std::vector<HeteroGraphPtr>& rel_graphs,
+    std::vector<int64_t>&& num_nodes_per_type) : BaseHeteroGraph(meta_graph) {
+  HeteroGraphSanityCheck(meta_graph, rel_graphs);
+  num_verts_per_type_ = std::move(num_nodes_per_type);
+  relation_graphs_ = CastToUnitGraphs(rel_graphs);
```
jermainewang: Remove this too?

include/dgl/base_heterograph.h
```diff
@@ -547,6 +553,15 @@ DGL_DEFINE_OBJECT_REF(FlattenedHeteroGraphRef, FlattenedHeteroGraph);
...
     GraphPtr meta_graph, const std::vector<HeteroGraphPtr>& rel_graphs);
 
+/*!
+ * \brief Create a heterograph from meta graph and a list of bipartite graph,
+ * additionally specifying number of nodes per type.
+ */
+HeteroGraphPtr CreateHeteroGraph(
+    GraphPtr meta_graph,
+    const std::vector<HeteroGraphPtr>& rel_graphs,
+    std::vector<int64_t> num_nodes_per_type);
```
jermainewang: Let's have both of them be `std::vector` and also merge it with L553 by adding a default argument?

BarclayII: If the second argument is pass-by-value then we will copy `rel_graphs` right?

PR #1327, opened 2020-03-08
[Feature] Improve the speed of to_bidirected

zheng-da:
## Description
The original implementation is very slow. GKlib is much faster.

## Checklist
- [ ] Code is well-documented


jermainewang:
@mufeili Please take a look. We changed the behavior of `to_bidirected` to that it just assures edge i->j and j->i will both present in the returned graph if *either* of them presents in the original graph. For multi-graph, we no longer enforces a well-defined behavior. The change is majorly for efficiency. Please raise your concern if there's any.

BarclayII:
Did you figure out the reason why GKLib is much faster?  If it's because GKLib manipulates CSR matrices more efficiently, we may want to update the CSR implementations in the future.

zheng-da:
@BarclayII I didn't read George's code.

mufeili:
In that case, I'd suggest we do a check for whether the graph is multigraph for the Python end and raise an error directly since multigraph is no longer supported by `to_bidirected`.

src/graph/gk_ops.cc
```diff
@@ -0,0 +1,106 @@
...
+gk_csr_t *Convert2GKCsr(CSRPtr csr, bool is_row) {
+  // TODO(zhengda) The conversion will be zero-copy in the future.
+  const aten::CSRMatrix mat = csr->ToCSRMatrix();
+  const dgl_id_t *indptr = static_cast<dgl_id_t*>(mat.indptr->data);
+  const dgl_id_t *indices = static_cast<dgl_id_t*>(mat.indices->data);
+
+  gk_csr_t *gk_csr = gk_csr_Create();
+  gk_csr->nrows = mat.num_rows;
+  gk_csr->ncols = mat.num_cols;
+  size_t nnz = csr->NumEdges();
```
classicsong: I suggest to use int64_t here.

src/graph/gk_ops.cc
```diff
@@ -0,0 +1,106 @@
...
+    num_ptrs = gk_csr->nrows + 1;
+    gk_indptr = gk_csr->rowptr = gk_zmalloc(gk_csr->nrows+1, "gk_csr_ExtractPartition: rowptr");
+    gk_indices = gk_csr->rowind = gk_imalloc(nnz, "gk_csr_ExtractPartition: rowind");
+  } else {
+    num_ptrs = gk_csr->ncols + 1;
+    gk_indptr = gk_csr->colptr = gk_zmalloc(gk_csr->ncols+1, "gk_csr_ExtractPartition: colptr");
+    gk_indices = gk_csr->colind = gk_imalloc(nnz, "gk_csr_ExtractPartition: colind");
+  }
+
+  for (size_t i = 0; i < num_ptrs; i++) {
```
classicsong: if sizeof(gk_indptr[0]) == sizeof(indptr[0])
I think we can use memcopy here at least.

src/graph/gk_ops.cc
```diff
@@ -0,0 +1,106 @@
...
+    gk_indptr[i] = indptr[i];
+  }
+  for (size_t i = 0; i < nnz; i++) {
+    gk_indices[i] = indices[i];
+  }
+  return gk_csr;
```
classicsong: Is gk_csr a shared_ptr? Or we need to free it manually?

src/graph/gk_ops.cc
zheng-da: actually, the return type of NumEdges is `uint64_t`.

src/graph/gk_ops.cc
zheng-da: at this point, they are different. In the future (when GKlib supports both int32 and int64), we'll use memory copy.

src/graph/gk_ops.cc
zheng-da: good point. we need to free gk_csr.

src/graph/gk_ops.cc
classicsong: shall we use uint64_t nnz directly?

src/graph/gk_ops.cc
classicsong: Shall we do some assertion before ahead? otherwise cast a int64 to int32 is dangerous.

zheng-da: sure. we can double check the integer types.

src/graph/gk_ops.cc
zheng-da: i did it.

src/graph/gk_ops.cc
zheng-da: actually, when i wrote code, i didn't do any casting here. The code doesn't even assume the integer type. It should work for all different combinations. In the future, we'll check the type and cast the integers when necessary.

src/graph/gk_ops.cc
```diff
@@ -0,0 +1,108 @@
...
+  const aten::CSRMatrix mat = csr->ToCSRMatrix();
+  const dgl_id_t *indptr = static_cast<dgl_id_t*>(mat.indptr->data);
+  const dgl_id_t *indices = static_cast<dgl_id_t*>(mat.indices->data);
+
+  gk_csr_t *gk_csr = gk_csr_Create();
+  gk_csr->nrows = mat.num_rows;
+  gk_csr->ncols = mat.num_cols;
+  uint64_t nnz = csr->NumEdges();
+  auto gk_indptr = gk_csr->rowptr;
+  auto gk_indices = gk_csr->rowind;
```
jermainewang: Initialize them to `nullptr` as you will override in both branches below.

src/graph/gk_ops.cc
jermainewang: If gk_csr is a transient data structure, consider using `std::unique_ptr` w/ proper destructor function.

src/graph/gk_ops.cc
```diff
@@ -0,0 +1,108 @@
...
+
+#include <dgl/graph_op.h>
+
+namespace dgl {
+
+#if !defined(_WIN32)
+
+namespace {
+
+gk_csr_t *Convert2GKCsr(CSRPtr csr, bool is_row) {
```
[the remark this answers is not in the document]
jermainewang: Two more suggestions:
* Make the two functions template functions.
* Directly use `aten::CSRMatrix` as the argument so it only depends on `array.h`.
```c++
template <typename IdxType>
gk_csr_t * Convert2GKCsr(aten::CSRMatrix csr, bool is_row);
```

include/dgl/graph_op.h
```diff
@@ -139,6 +139,13 @@ class GraphOp {
    * \return a new immutable bidirected graph.
    */
   static GraphPtr ToBidirectedImmutableGraph(GraphPtr graph);
+  /*!
+   * \brief Same as BidirectedMutableGraph except that the returned graph is immutable
+   * and call gk_csr_MakeSymmetric in GKlib. It works much faster.
```
jermainewang: .... It utilizes a more efficient implementation using Gklib. Return a null pointer if the conversion fails.

src/graph/gk_ops.cc
zheng-da: it requires special destructor.

src/graph/gk_ops.cc
zheng-da: i don't want to define the type explicitly.

src/graph/gk_ops.cc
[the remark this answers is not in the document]
zheng-da: The reason I created this file is to have all operations provided in GKLib to be called in this file.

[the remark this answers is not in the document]
jermainewang: I think we are in the progress of putting functionally related operators in one file. See `graph/partition.cc`, `graph/pickle.cc`.

src/graph/gk_ops.cc
jermainewang: std::unique supports a custom deleter function. http://www.cplusplus.com/reference/memory/unique_ptr/unique_ptr/

zheng-da: i just feel it's not very necessary. we have to explicitly manage the memory required from GKLib functions. In the future, we'll enable zero-copy. We don't need to deallocate memory.

src/graph/gk_ops.cc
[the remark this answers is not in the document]
zheng-da: yes, that's why I created this file. So all GKLib operations will be invoked in this file.

PR #1328, opened 2020-03-08
[Optimization] Optimize CompactGraph

classicsong:
## Description
When doing compact graph we use unordered_map to store parent_id to local_id mapping. 
We can let unordered_map reserve enough space to void rehash operations when inserting new nodes.

A performance profile of CompactGraph is shown as below, the performance speedup can be 10%~30% when the id space is large(1M)

*Origin*
|  | Sampled edges| Sampled edges| Sampled edges| 
|-----------|-------|-------|--------|
| node_id space | 7168 | 113664 | 1110000|
| 1000 | 0.004s | 0.038s | 0.363s |
| 10000 | 0.007s | 0.060s | 0.525s |
|100000 | 0.017s | 0.101s | 0.702s |
|1000000 | 0.017s | 0.327s | 1.627s |

*Optimized*
|  | Sampled edges| Sampled edges| Sampled edges| 
|-----------|-------|-------|--------|
| node_id space | 7168 | 113664 | 1110000|
| 1000 | 0.004s | 0.041s | 0.407s |
| 10000 | 0.006s | 0.057s | 0.562s |
| 100000 | 0.014s | 0.090s | 0.740s |
| 1000000 | 0.017s | 0.211s | 1.453s |

## Checklist
- [x] Code is well-documented


BarclayII:
Nice!  Care to test `ToBlock` as well?

classicsong:
> Nice! Care to test `ToBlock` as well?

Will do

classicsong:
> > Nice! Care to test `ToBlock` as well?
> 
> Will do

By looking at the code, toBlock has already call reserve before inserting elements.

BarclayII:
> By looking at the code, toBlock has already called reserve before inserting elements.

The reserve is only for the array of number of nodes per type.  It is not for the hash table.

classicsong:
> > By looking at the code, toBlock has already called reserve before inserting elements.
> 
> The reserve is only for the array of number of nodes per type. It is not for the hash table.

https://github.com/dmlc/dgl/blob/614acf2c28aaff37a22ac3ec9901c492ba325c8c/src/graph/transform/to_bipartite.cc#L39
here you create the IdHashMap with rhs_nodes (IdArray) which reserve enough space and then call the Update(id): https://github.com/dmlc/dgl/blob/614acf2c28aaff37a22ac3ec9901c492ba325c8c/src/array/cpu/array_utils.h#L36

So there is no need to do so.

src/graph/transform/compact.cc
```diff
@@ -36,13 +36,30 @@ CompactGraphs(
   std::vector<aten::IdHashMap<IdType>> hashmaps(num_ntypes);
   std::vector<std::vector<EdgeArray>> all_edges(graphs.size());   // all_edges[i][etype]
 
-  for (size_t i = 0; i < always_preserve.size(); ++i)
+  std::vector<int64_t> max_vertex_cnt(num_ntypes, 0);
+  for (size_t i = 0; i < graphs.size(); ++i) {
+    const HeteroGraphPtr curr_graph = graphs[i];
+    for (IdType ntype = 0; ntype < num_ntypes; ++ntype) {
+      max_vertex_cnt[ntype] += curr_graph->NumVertices(ntype);
```
BarclayII: `CompactGraphs` compacts multiple graphs with the same node space, so there is no need to repeatedly add `NumVertices()`.

Also, note that `CompactGraphs` operates on the original node space (meaning that the number of nodes would be the same as full graph).  Would reserving that many nodes cost a lot of memory?

BarclayII:
> Here you create the IdHashMap with rhs_nodes (IdArray) which reserve enough space and then call the Update(id):

I see.

src/graph/transform/compact.cc
classicsong: I have updated to get NumEdges instead of NumVertices and the numbers above are updated

src/graph/transform/compact.cc
```diff
@@ -36,13 +36,37 @@ CompactGraphs(
   std::vector<aten::IdHashMap<IdType>> hashmaps(num_ntypes);
   std::vector<std::vector<EdgeArray>> all_edges(graphs.size());   // all_edges[i][etype]
```
BarclayII: Add a comment on this section such as "reserve the space for hash maps".

tests/compute/test_transform.py
```diff
@@ -261,6 +261,10 @@ def test_metis_partition():
         assert num_inner_nodes == g.number_of_nodes()
         print(g.number_of_edges() - num_inner_edges)
 
+def test_partition():
+    test_metis_partition()
+    test_partition_with_halo()
```
BarclayII: `test_blah` should be automatically tested by unit test modules; you don't have to add this function.

classicsong: OK, I have reverted it.

PR #1333, opened 2020-03-08
[Model] Rewrite GraphSAGE example to use DataLoader

BarclayII:
## Description
Rewrote GraphSAGE example to use DataLoader for sampling.

Profiling result goes as follows:

| # GPUs | # sampler workers per GPU | Time per epoch |
|:---:|:---:|:---:|
| 1 | 1 | 29.68s | 
| 1 | 6 | 17.2401s |
| 1 | 12 | 17.1699s |
| 4 | 1 | 7.49s |
| 4 | 6 | 7.1922s |

The benefit of multiprocessed sampling also diminishes as number of workers increase:

| # GPUs | # sampler workers per GPU | Time per epoch on sampler only |
|:---:|:---:|:---:|
| 1 | 1 | 11.5548s | 
| 1 | 6 | 3.2159s |
| 1 | 12 | 4.0606s |

EDIT (Mar 9th): the reason may be that the dataset is too small as the batch size is 1000 and we only have around 200k nodes - which yields us 200 batches.  MXNet and Tensorflow both have their own solutions analogous to PyTorch DataLoader, so I expect that supporting multiprocess sampling should be straightforward enough.

Also fixes a bug on evaluation code with multi-GPU.

## Checklist
- [x] Code is well-documented


VoVAllen:
Can sampling now pipeline with the computation?

BarclayII:
> Can sampling now pipeline with the computation?

If "pipelining" means to sample and compute simultaneously then it should be naturally achieved with `DataLoader` with `num_workers > 1`.

examples/pytorch/graphsage/graphsage_sampling.py
```diff
@@ -194,7 +208,16 @@ def run(proc_id, n_gpus, args, devices, data):
...
+    sampler = NeighborSampler(g, [int(fanout) for fanout in args.fan_out.split(',')])
+
+    # Create PyTorch DataLoader for constructing blocks
+    dataloader = DataLoader(
+        dataset=train_nid.numpy(),
+        batch_size=args.batch_size,
+        collate_fn=sampler.sample_blocks,
+        shuffle=True,
+        drop_last=False,
+        num_workers=args.num_workers_per_gpu)
```
zheng-da: can we have builtin sampler for neighbor sampling?

BarclayII: What do you mean by builtin sampler?  Everything in C?

jermainewang: Quan and I think we could stay with the current implementation first and gradually putting some common samplers to DGL package once we have more examples built up. @zheng-da what do you think?

PR #1337, opened 2020-03-09
[Model] RGCN mini-batch training; Better bipartite graph support

jermainewang:
## Description

RGCN minibatch training for node classification task. See `entity_classify_mb.py`.

Better bipartite graph support:
* Upon a DGLHeteroGraph is created, it will check whether its metagraph is a uni-directional bipartite graph.
* If yes, then it will split the `ntypes` into `srctypes` and `dsttypes`. In this case, we allow two node types to have the same name as long as one belongs to `srctypes` and the other belongs to `dsttypes`.
* Following new APIs are added to ease the use of a bipartite graph:
  - `srctypes` and `dsttypes`
  - `srcdata` and `dstdata`
  - `srcnodes` and `dstnodes`
* Besides using the above APIs to distinguish nodes types in the SRC and DST groups, another way is by prepending the type name with `"SRC/"` or `"DST/"`. For example, `g.number_of_nodes("SRC/user")` returns the number of nodes of type `"user"` in the SRC group.

Other changes:
* Change the original `entity_classify.py` to directly use node embedding. Previous code use one embedding for one (srctype, etype) pair. Verified that the change does not influence accuracy.

## Checklist
- [x] Code is well-documented


BarclayII:
One side question: should we consider launching multiple CUDA streams for `multi_update_all` (or even one stream for each relation)?  I saw PyG doing this in `HeteConv`.  Not very urgent (at least for this PR) though.

python/dgl/heterograph.py
```diff
@@ -304,6 +335,24 @@ def add_edges(self, u, v, data=None, etype=None):
     # Metagraph query
     #################################################################
 
+    @property
+    def is_unibipartite(self):
+        """Return whether the graph is a uni-bipartite graph.
```
BarclayII: We may want to reconsider how we call such graph structures.  At least "unidirectional bipartite graph" refers to the graphs with only one edge type and two node types in our docstrings.

jermainewang:
> One side question: should we consider launching multiple CUDA streams for `multi_update_all` (or even one stream for each relation)? I saw PyG doing this in `HeteConv`. Not very urgent (at least for this PR) though.

From your benchmarks, the major bottleneck seems to be data movement between CPU and GPU, so we could put that as low priority.

python/dgl/heterograph.py
jermainewang: I think uni-directional bipartite graph is the word for this kind. More formally, given the definition of uni-directional bipartite graph for homogeneous graph, we can extend this definition to heterogeneous graph by considering the metagraph -- a heterograph is uni-directional bipartite iff its metagraph is uni-directional bipartite.

A block graph is a special uni-bipartite graph with nodes properly relabeled and each DST node has one mirror in the SRC group.

python/dgl/heterograph.py
```diff
@@ -479,16 +530,75 @@ def get_ntype_id(self, ntype):
...
+        """
+        if ntype is None:
+            if len(self._srctypes_invmap) != 1:
+                raise DGLError('SRC node type name must be specified if there are more than one '
+                               'SRC node types.')
+            return 0
+        ntid = self._srctypes_invmap.get(ntype, None)
+        if ntid is None:
+            raise DGLError('SRC node type "{}" does not exist.'.format(ntype))
+        return ntid
```
BarclayII: Because we sort the node types in `srctypes`, `get_srctype_id(ntype)` may not necessarily return the index of `ntype` in `srctypes`, right?  Would it cause problems?

python/dgl/heterograph.py
```diff
@@ -479,16 +530,75 @@ def get_ntype_id(self, ntype):
...
+        """
+        if ntype is None:
+            if len(self._dsttypes_invmap) != 1:
+                raise DGLError('DST node type name must be specified if there are more than one '
+                               'DST node types.')
+            return 0
+        ntid = self._dsttypes_invmap.get(ntype, None)
+        if ntid is None:
+            raise DGLError('DST node type "{}" does not exist.'.format(ntype))
+        return ntid
```
BarclayII: Same.

python/dgl/heterograph.py
```diff
@@ -3864,6 +4042,55 @@ def make_canonical_etypes(etypes, ntypes, metagraph):
...
+        not uni-bipartite.
+    """
+    src, dst, _ = metagraph.edges()
+    if set(src.tonumpy()).isdisjoint(set(dst.tonumpy())):
+        srctypes = {ntypes[tid] : tid for tid in src}
+        dsttypes = {ntypes[tid] : tid for tid in dst}
+        # handle isolated node types
+        for ntid, ntype in enumerate(ntypes):
+            if ntype not in srctypes and ntype not in dsttypes:
+                srctypes[ntype] = ntid
```
BarclayII: Add this behavior to docstring?

python/dgl/transform.py
```diff
@@ -743,38 +743,46 @@ def compact_graphs(graphs, always_preserve=None):
...
-    Otherwise, the right hand side would be determined by DGL via the rules above.
+    Since DST nodes are included in SRC nodes, a common requirement is to fetch
+    the DST node features from the SRC nodes features. To avoid expensive sparse lookup,
+    the function assures that the DST nodes in both SRC and DST sets have the same ids.
+    As a result, given the node feature tensor ``X`` of type ``utype``,
+    the following code finds the corresponding DST node features of type ``vtype``:
+
+    .. code::
+
+        X[:block.number_of_nodes('vtype')]
```
BarclayII: If we are converting a block with `'user'` as `utype` and again `'user'` as `vtype` (i.e. I'm converting a homograph to a block), would this statement work?

examples/pytorch/rgcn-hetero/entity_classify_mb.py
```diff
@@ -0,0 +1,432 @@
...
+                 num_bases,
+                 num_hidden_layers=1,
+                 dropout=0,
+                 use_self_loop=False):
+        super(EntityClassify, self).__init__()
+        self.g = g
+        self.h_dim = h_dim
+        self.out_dim = out_dim
+        self.rel_names = list(set(g.etypes))
+        self.rel_names.sort()
```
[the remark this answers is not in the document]
jermainewang: It is to keep a deterministic order so that offline evaluation could work. IMO, it can be avoided as long as we use ParameterDict rather than ParameterList. @classicsong what do you think?

examples/pytorch/rgcn-hetero/entity_classify_mb.py
```diff
@@ -0,0 +1,432 @@
...
+    if use_cuda:
+        model.cuda()
+
+    # train sampler
+    sampler = HeteroNeighborSampler(g, category, [args.fanout] * args.n_layers)
+    loader = DataLoader(dataset=list(train_idx.numpy()),
+                        batch_size=args.batch_size,
+                        collate_fn=sampler.sample_blocks,
+                        shuffle=True,
+                        num_workers=0)
```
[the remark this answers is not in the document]
jermainewang: agree

python/dgl/heterograph.py
jermainewang: Unlike `ntypes`,  `srctypes` and `dsttypes` cannot be used as a mapping from type id to type name. Thus, their results are technically set, but I still return list to keep it similar to `ntypes`. The list is sorted to make the behavior deterministic. Since the type id is retrieved from `_srctypes_invmap`, it won't cause trouble.

python/dgl/transform.py
jermainewang: It should be `X[:block.number_of_nodes('DST/vtype')]`. Will change that.

python/dgl/heterograph.py
BarclayII: OK.  I think users need to be aware of that though, since `get_srctype_id` would behave very differently from `get_ntype_id`.  Could we make them consistent?

examples/pytorch/rgcn-hetero/entity_classify_mb.py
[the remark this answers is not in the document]
BarclayII: I see.  I think using `ParameterDict` can make the code more intuitive.

examples/pytorch/rgcn-hetero/entity_classify_mb.py
```diff
@@ -0,0 +1,432 @@
...
+        g.multi_update_all(funcs, 'sum')
+
+        hs = {}
+        for ntype in g.dsttypes:
+            if 'h' in g.dstnodes[ntype].data:
+                hs[ntype] = g.dstnodes[ntype].data['h']
+        def _apply(h):
+            # apply bias and activation
+            if self.self_loop:
+                h = h + th.matmul(xs[ntype], self.loop_weight)
```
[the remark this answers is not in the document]
jermainewang: You are right. They could have different number of rows.

examples/pytorch/rgcn-hetero/entity_classify_mb.py
```diff
@@ -0,0 +1,432 @@
...
+        for layer, block in zip(self.layers, blocks):
+            h = layer(block, h)
+        return h
+
+class HeteroNeighborSampler:
+    def __init__(self, g, category, fanouts, full_neighbor=False):
+        self.g = g
+        self.category = category
+        self.fanouts = fanouts
+        self.full_neighbor = full_neighbor
```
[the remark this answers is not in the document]
jermainewang: Done

examples/pytorch/rgcn-hetero/entity_classify_mb.py
```diff
@@ -0,0 +1,432 @@
...
+
+    def sample_blocks(self, seeds):
+        blocks = []
+        seeds = {self.category : th.tensor(seeds)}
```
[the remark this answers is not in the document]
jermainewang: ddone

examples/pytorch/rgcn-hetero/entity_classify_mb.py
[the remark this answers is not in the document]
jermainewang: Done

caowencai:
#3115

PR #1343, opened 2020-03-10
[KG] make batch size compatible with neg sample size

zheng-da:
## Description
We need to allow users to specify arbitrary batch size, neg sample size. One solution is to adjust them in the training code.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/apps/kg/train.py
+++ b/apps/kg/train.py
@@ -6,7 +6,8 @@ import os
 import logging
 import time
 import json
-import math
+
+from utils import get_compatible_batch_size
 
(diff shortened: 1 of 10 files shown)


apps/kg/train.py
```diff
@@ -134,13 +133,24 @@ def get_logger(args):
     return logger
 
 
+def get_compatible_batch_size(batch_size, neg_sample_size):
```
classicsong: Can we put this into some place that three files can share the same implementation?

PR #1355, opened 2020-03-12
[Model] GraphSAGE with control variate sampling on new sampler

BarclayII:
## Description
This implements control variate sampling from https://arxiv.org/abs/1710.10568 with the new sampler interface.

@zheng-da @lingfanyu Please take a look if you are interested.

### TODO

- [x] Profile performance (speed and memory)
- [x] Verify correctness on multi-GPU
- [x] Better document the code
- [ ] ~~Refactor the code with vanilla GraphSAGE and extract common logic~~ Skipping this for now.

## Checklist
- [x] Code is well-documented


lingfanyu:
Wondering why not start with a non-distributed version?

BarclayII:
There's not much difference between a non-distributed version and a distributed one: the only differences are to initialize a distributed context and synchronize the gradients among GPUs.

examples/pytorch/graphsage/train_cv.py
```diff
@@ -0,0 +1,341 @@
...
+        inputs = g.ndata['features']
+        pred, _ = model.inference(g, inputs, batch_size, device)
+    model.train()
+    return compute_acc(pred[val_mask], labels[val_mask])
+
+def load_subtensor(g, labels, blocks, dev_id):
+    """
+    Copys features and labels of a set of nodes onto GPU.
+    """
+    blocks[0].srcdata['features'] = g.ndata['features'][blocks[0].srcdata[dgl.NID]].to(dev_id)
```
zheng-da, line 184: btw, the original paper plays a trick. it aggregates the features in the first layer in advance.

examples/pytorch/graphsage/train_cv_multi_gpu.py
```diff
@@ -0,0 +1,416 @@
...
+            old_hist = g.ndata[hist_col][ids]
+            g.ndata[hist_delta_col][ids] = h_new - old_hist
+            g.ndata[hist_col][ids] = h_new
+            # TODO: can we avoid creating a new feature called agg_new_col?
+            g.push(
+                ids,
+                fn.copy_u(hist_delta_col, 'm'),
+                fn.sum('m', agg_new_col),
+                lambda nodes: {agg_hist_col: nodes.data[agg_hist_col] + nodes.data[agg_new_col]},
+                inplace=True)
```
zheng-da, line 255: it seems you need to keep both history node embeddings for each layer as well as aggregated history for each layer. Is it right?

examples/pytorch/graphsage/train_cv_multi_gpu.py
```diff
@@ -0,0 +1,416 @@
...
+            queue.put((res, exception, trace))
+
+        start_new_thread(_queue_result, ())
+        result, exception, trace = queue.get()
+        if exception is None:
+            return result
+        else:
+            assert isinstance(exception, Exception)
+            raise exception.__class__(trace)
+    return decorated_function
```
zheng-da, line 171: can we keep this somewhere inside DGL?

zheng-da:
Why not follow the original example?

examples/pytorch/graphsage/train_cv_multi_gpu.py
lingfanyu: I think it's correct. After every batch, both history and aggregate of history will be update in `update_history` function.

But actually this is not very efficient, because updating the aggregate of history involves a `g.push`, and outgoing edges of changed nodes are going to touch a lot more nodes. In my implementation, I actually re-compute aggregate of history on demand during model forwarding, which makes the training 2-3x faster. @BarclayII

zheng-da: the original example also computes the aggregation on demand. BTW, why do you guys not use the original implementation? Do you know DGL has a control variate implementation for more than one year?

lingfanyu: This control variate serves as an example for the new sampling API, I believe.

zheng-da: it's pretty easy to adapt from the old implementation.

BarclayII:
I thought using `push` could save communication compared to using `pull`.  My bad.  Will change it.

examples/pytorch/graphsage/train_cv_multi_gpu.py
BarclayII: Totally agree.  Will put it in utils.

It is everywhere in our newly contributed examples already.

examples/pytorch/graphsage/train_cv_multi_gpu.py
BarclayII: Switched to implementation with blocks.  It is now 10 times faster.

examples/pytorch/graphsage/train_cv_multi_gpu.py
BarclayII: I'm now wondering if utils is a good place to put this function, since it depends on PyTorch multiprocessing module.

examples/pytorch/graphsage/train_cv_multi_gpu.py
```diff
@@ -0,0 +1,409 @@
...
+                g.ndata['hist_%d' % layer] = history[layer - 1]
+
+def update_history(g, blocks):
+    with th.no_grad():
+        for i, block in enumerate(blocks):
+            ids = block.dstdata[dgl.NID]
+            hist_col = 'hist_%d' % (i + 1)
+
+            h_new = block.dstdata['h_new'].cpu()
+            g.ndata[hist_col][ids] = h_new
```
jermainewang, line 248: Will there be data race issue?

jermainewang:
Other suggestions:
* Add the training results to the README.
* Does control variate based sampling converge faster? If so, you could highlight that in the README too.
* There are many code duplication in the `train_cv.py` and `train_cv_multi_gpu.py`. I can see putting everything in one script is potentially easier for our users to copy-paste. I'll leave to you to decide whether to put them in a common file.

BarclayII:
I'll leave them as is.

examples/pytorch/graphsage/train_cv_multi_gpu.py
BarclayII: Yes.  In fact this operation is not even atomic.  Doesn't matter as I observed in both Reddit and OGB though.

BarclayII: The children don't share the same frame apparently.

PR #1381, opened 2020-03-22
[Feature] Simplify shared memory graph index

zheng-da:
## Description
When creating a graph index in shared memory, users originally needed to provide a lot of information. This PR is to simplify the inputs of the creation of the shared-memory graph index.

## Checklist
- [ ] Code is well-documented


jermainewang:
The PR looks mostly good to me. Two thoughts:
* Might want to wait for PR #1389 . There will be some conflicts but should be easier to resolve on your side.
* I wonder how PyTorch implements arrays in shared memory and can we borrow some of the ideas? Their API is quite easy to use. At least, user does not need to provide a filename.
  ```python
  >>> import torch
  >>> x = torch.randn((10, 4))
  >>> x.share_memory_()   # move x to shared memory
  >>> x.is_shared()
  True
  >>> x.storage()._share_filename_()
  (b'/tmp/torch-shm-file-hh7t6N', b'/torch_2346_2986491181', 200)
  ```

jermainewang:
Another thing you might want to be aware of is this PR #1299 . Basically, the new pickling method utilizes backend tensors as much as possible, and for the backend which supports shared memory tensors (e.g., PyTorch), transmitting our graph object to another process is very efficient.

zheng-da:
Shared memory has two types: anonymous and named. I want to use named shared memory on purpose. Pytorch use anonymous shared memory. The problem with anonymous shared memory is that you have to fork processes to share data. This is inconvenient. If I create two processes separately (instead of forking one process from another process), I cannot share data between them with anonymous shared memory. Named shared memory allows us to do whatever we want.

zheng-da:
I see. Does it mean that if the backend supports shared memory tensors, transmitting graph structure can simply use pickle and it’s zero memory copy and no additional memory use?

PR #1383, opened 2020-03-22
[Feature] Distributed graph store

zheng-da:
## Description
This is the first step of adding distributed GNN training in DGL. The PR adds a distributed graph for distributed training.

## Checklist
- [ ] Code is well-documented


python/dgl/distributed/dist_graph.py
```diff
@@ -0,0 +1,506 @@
...
+            reprs[name] = 'DistTensor(shape={}, dtype={})'.format(str(shape), str(dtype))
+        return repr(reprs)
+
+
+class DistGraphServer(KVServer):
+    ''' The DistGraph server.
+
+    This DistGraph server extends KVStore server. Its main function is to load the graph data
+    and share the data in its own partition with shared memory so that the DistGraph client
+    can access the data in the local partition with shared memory.
```
jermainewang, line 257: The docstring is not self-contained. For example, what is a KVStore?

python/dgl/distributed/dist_graph.py
```diff
@@ -0,0 +1,506 @@
...
+
+    Parameters
+    ----------
+    server_id : int
+        KVServer's ID (start from 0).
```
jermainewang, line 262: Try avoid the concept of `KVStore` in the docstring unless KVStore is a user-facing API.

python/dgl/distributed/dist_graph.py
```diff
@@ -0,0 +1,506 @@
...
+           6:'[3, 172.31.30.180, 30050],
+           7:'[3, 172.31.30.180, 30051]}
+    num_client : int
+        Total number of client nodes.
+    graph_name : string
+        The name of the graph.
+    conf_file : string
+        The path of the config file generated by the partition tool.
+    '''
+    def __init__(self, server_id, server_namebook, num_client, graph_name, conf_file):
```
jermainewang, line 282: Can we have a filepath instead of a graph_name? It is not clear to me why need this name and how it is used.

python/dgl/distributed/dist_graph.py
```diff
@@ -0,0 +1,506 @@
...
+    def __repr__(self):
+        name_list = self._graph._get_all_edata_names()
+        reprs = {}
+        for name in name_list:
+            dtype, shape, _ = self._graph._client.get_data_meta(_get_edata_name(name))
+            reprs[name] = 'DistTensor(shape={}, dtype={})'.format(str(shape), str(dtype))
+        return repr(reprs)
+
+
+class DistGraphServer(KVServer):
```
jermainewang, line 252: I'm not sure if this is the correct inheritance. Suppose I have two tensors: an "h" node feature tensor and a "w" edge feature tensor. Are they stored in the same KVStore service? Does KVStore support storing distributed tensors that are partitioned differently?

python/dgl/distributed/dist_graph.py
```diff
@@ -0,0 +1,506 @@
...
+                                              num_client=num_client)
+
+        host_name = socket.gethostname()
+        host_ip = socket.gethostbyname(host_name)
+        print('Server {}: host name: {}, ip: {}'.format(server_id, host_name, host_ip))
+
+        self.client_g, node_feats, _, self.meta = load_partition(conf_file, server_id)
+        full_part_ids = self.meta[2]
+        num_nodes = self.meta[0]
+        num_edges = self.meta[1]
```
jermainewang, line 293: `full_part_ids, num_nodes, num_edges = self.meta`

python/dgl/distributed/partition.py
```diff
@@ -0,0 +1,151 @@
...
+    conf_file : str
+        The path of the partition config file.
+    part_id : int
+        The partition Id.
+    '''
+    with open(conf_file) as conf_f:
+        part_metadata = json.load(conf_f)
+    part_files = part_metadata['part-{}'.format(part_id)]
+    node_feats = pickle.load(open(part_files['node_feats'], 'rb'))
+    edge_feats = pickle.load(open(part_files['edge_feats'], 'rb'))
```
jermainewang, line 30: Let's not use pickle.

python/dgl/distributed/partition.py
```diff
@@ -0,0 +1,151 @@
...
+    client_g = load_graphs(part_files['part_graph'])[0][0]
+    node_map = pickle.load(open(part_metadata['node_map'], 'rb'))
```
jermainewang, line 32: how about `edge_map`?

python/dgl/distributed/partition.py
```diff
@@ -0,0 +1,151 @@
...
+
+    Here we load data through the normal filesystem interface. In the future, we need to support
+    loading data from other storage such as S3 and HDFS.
+
+    Parameters
+    ----------
+    conf_file : str
+        The path of the partition config file.
+    part_id : int
+        The partition Id.
```
jermainewang, line 24: Could you document the return values?

python/dgl/distributed/dist_graph.py
```diff
@@ -0,0 +1,506 @@
...
+        self.client_g = _copy_graph_to_shared_mem(self.client_g, graph_name)
+        self.meta = _move_data_to_shared_mem_array(F.tensor([num_nodes, num_edges]),
+                                                   _get_ndata_path(graph_name, 'meta'))
```
jermainewang, line 296: Why putting them to shared memory? I though each partition is owned by one server and will not be visible to the others.

python/dgl/contrib/dis_kvstore.py
```diff
@@ -830,6 +830,7 @@ def init_data(self, name, shape, dtype, target_name):
         assert len(name) > 0, 'name cannot be empty.'
         assert len(shape) > 0, 'shape cannot be empty.'
         assert len(target_name) > 0, 'target_name cannot be empty.'
+        shape = list(shape)
```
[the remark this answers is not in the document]
aksnzhy: @zheng-da I think this can be removed because I add the support of tuple type.

src/graph/network.cc
```diff
@@ -807,6 +815,9 @@ DGL_REGISTER_GLOBAL("network._CAPI_FastPull")
     // Copy local data
 #pragma omp parallel for
     for (int64_t i = 0; i < local_ids.size(); ++i) {
+      CHECK_GE(ID_size*row_size, local_ids_orginal[i] * row_size + row_size);
+      CHECK_GE(data_size, local_ids[i] * row_size + row_size);
+      CHECK_GE(local_ids[i], 0);
```
aksnzhy, line 820: Does this checking will hurt our performance?

tests/distributed/test_dist_graph_store.py
```diff
@@ -0,0 +1,101 @@
+import os
+os.environ['OMP_NUM_THREADS'] = '1'
```
aksnzhy, line 2: Will we fix OPM_THREAD to 1 in our real tasks?

tests/distributed/test_dist_graph_store.py
```diff
@@ -0,0 +1,101 @@
...
+import multiprocessing as mp
+from dgl.graph_index import create_graph_index
+from dgl.data.utils import load_graphs, save_graphs
+from dgl.distributed import DistGraphServer, DistGraph
+from dgl.distributed import partition_graph
+import backend as F
+import unittest
+import pickle
+
+server_namebook = {0: [0, '127.0.0.1', 30000, 1]}
```
aksnzhy, line 19: It's hard to do unittest in a distributed setting. This problem also bothers me.

python/dgl/distributed/dist_graph.py
```diff
@@ -0,0 +1,506 @@
...
+    name : string
+        The name of the tensor in the KVStore.
+    '''
+    def __init__(self, g, name):
+        self.kvstore = g._client
+        self.name = name
+        dtype, shape, _ = g._client.get_data_meta(name)
+        # We need to ensure that the first dim is the number of nodes in a graph.
+        shape = [s for s in shape]
+        shape[0] = g.number_of_nodes()
```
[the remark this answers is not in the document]
zheng-da: there was a bug in kvstore.

python/dgl/distributed/dist_graph.py
```diff
@@ -0,0 +1,506 @@
...
+        self._shape = tuple(shape)
+        self._dtype = dtype
+
+    def __getitem__(self, idx):
+        return self.kvstore.pull(name=self.name, id_tensor=idx)
+
+    def __setitem__(self, idx, val):
+        # TODO how do we want to support broadcast.
```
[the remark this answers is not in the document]
zheng-da: val can be a scalar or a vector.  we might need to support something like this.
```python
g.ndata['feats'][idx] = 1
```

python/dgl/distributed/partition.py
```diff
@@ -0,0 +1,151 @@
...
+    part_files = part_metadata['part-{}'.format(part_id)]
+    node_feats = pickle.load(open(part_files['node_feats'], 'rb'))
+    edge_feats = pickle.load(open(part_files['edge_feats'], 'rb'))
+    client_g = load_graphs(part_files['part_graph'])[0][0]
+    node_map = pickle.load(open(part_metadata['node_map'], 'rb'))
+    meta = (part_metadata['num_nodes'], part_metadata['num_edges'], node_map)
+
+    part_ids = node_map[client_g.ndata[NID]]
+    # TODO we need to fix this. DGL backend doesn't support boolean or byte.
+    # int64 is unnecessary.
```
jermainewang, line 37: Bool support has been added in https://github.com/dmlc/dgl/pull/1487

python/dgl/distributed/dist_graph.py
zheng-da: Do you think if we should avoid talking about KVStore in the docstring? Maybe we should explain what kvstore is.

python/dgl/distributed/partition.py
zheng-da: right now the edge assignment is implicit. Do we want to explicitly store the edge assignment? I think we should support it after we decide to partition the neighborhood of a vertex.

src/graph/network.cc
zheng-da: most likely not. if node data is large, it doesn't matter much. If the node data is small (e.g., just a float), it might hurt performance. However, it's not efficient to use memcpy to copy a scalar either.

tests/distributed/test_dist_graph_store.py
zheng-da: do we know what the problem is?

aksnzhy: It is a bug of OMP and only exists in gcc compiler.

tests/distributed/test_dist_graph_store.py
```diff
@@ -0,0 +1,101 @@
...
+
+def create_random_graph(n):
+    arr = (spsp.random(n, n, density=0.001, format='coo') != 0).astype(np.int64)
+    ig = create_graph_index(arr, readonly=True)
+    return dgl.DGLGraph(ig)
+
+def run_server(graph_name, server_id, num_clients, barrier):
```
jermainewang, line 26: Test one configuration with backup server too?

python/dgl/distributed/partition.py
```diff
@@ -0,0 +1,163 @@
+"""Functions for partitions."""
```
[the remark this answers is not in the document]
zheng-da: 1. i would prefer to expose the file names of a partition to users, instead of hard code the names in the code. it's more flexible. the json file can also tell users more information what is going on with the partitioning.
2. edge id mapping in stored EID. when you say node id mapping, do you mean edge assignment?
3. added checks for each field.

python/dgl/graph_index.py
```diff
@@ -1025,6 +1025,11 @@ def from_csr(indptr, indices, direction):
     indices : Tensor
         column index array in the CSR format
     direction : str
+
+    Returns
+    ------
+    GraphIndex
+        The graph index
```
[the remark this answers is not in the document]
zheng-da: i think the original docstring missed the return values.

python/dgl/distributed/dist_graph.py
```diff
@@ -0,0 +1,534 @@
...
+class NodeDataView(MutableMapping):
+    """The data view class when dist_graph.ndata[...].data is called.
+    """
+    __slots__ = ['_graph', '_data']
+
+    def __init__(self, g):
+        self._graph = g
+        # When this is created, the server may already load node data. We need to
+        # initialize the node data in advance.
+        names = g._get_all_ndata_names()
```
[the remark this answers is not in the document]
zheng-da: probably, that's why NodeDataView is generated when DistGraph is initialized. It's only called once.

python/dgl/distributed/dist_graph.py
```diff
@@ -0,0 +1,534 @@
...
+        super(DistGraphServer, self).__init__(server_id=server_id, server_namebook=server_namebook,
+                                              num_client=num_client)
+
+        host_name = socket.gethostname()
+        host_ip = socket.gethostbyname(host_name)
+        print('Server {}: host name: {}, ip: {}'.format(server_id, host_name, host_ip))
+
+        self.client_g, node_feats, _, self.meta = load_partition(conf_file, server_id)
+        num_nodes, num_edges, full_part_ids = self.meta
+        self.client_g = _copy_graph_to_shared_mem(self.client_g, graph_name)
```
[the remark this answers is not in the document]
zheng-da: i agree.

python/dgl/distributed/dist_graph.py
```diff
@@ -0,0 +1,534 @@
...
+        self.meta = _move_data_to_shared_mem_array(F.tensor([num_nodes, num_edges]),
+                                                   _get_ndata_path(graph_name, 'meta'))
```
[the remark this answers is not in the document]
zheng-da: i know the semantic of mxnet shared memory tensor is wrong. and none of them support named shared memroy.

python/dgl/distributed/dist_graph.py
```diff
@@ -0,0 +1,534 @@
...
+
+        self.g2l = F.zeros((num_nodes), dtype=F.int64, ctx=F.cpu())
+        self.g2l[:] = -1
+        local_nids = F.nonzero_1d(self.client_g.ndata['local_node'])
+        # The nodes that belong to this partition.
+        nids = self.client_g.ndata[NID][local_nids]
+        assert np.all(full_part_ids[nids] == server_id)
+        self.g2l[nids] = F.arange(0, len(nids))
```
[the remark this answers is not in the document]
zheng-da: I think this can be handled by the partition book, but i don't think we should postpone this PR because of this. too much dependency.

python/dgl/distributed/dist_graph.py
```diff
@@ -0,0 +1,534 @@
...
+
+        full_part_ids = F.zerocopy_from_numpy(full_part_ids)
+        if self.get_id() % self.get_group_count() == 0: # master server
+            for name in node_feats:
+                self.set_global2local(name=_get_ndata_name(name), global2local=self.g2l)
+                self.init_data(name=_get_ndata_name(name), data_tensor=node_feats[name])
+                self.set_partition_book(name=_get_ndata_name(name), partition_book=full_part_ids)
+        else:
+            for name in node_feats:
+                self.set_global2local(name=_get_ndata_name(name))
```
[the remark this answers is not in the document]
zheng-da: because all backup servers can get g2l through shared memory.

python/dgl/distributed/dist_graph.py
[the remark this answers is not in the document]
jermainewang: Sorry I mean let the later PR handle this.

PR #1385, opened 2020-03-22
[NN] Add HeteroGraphConv module for cleaner module definition

jermainewang:
## Description

Previously, we require users to directly use low-level message passing APIs to define layers operating on heterographs. This is tedious since many graph convolution modules become more and more complicated (think of `GATConv`). I proposed to add a generic wrapper module called `HeteroGraphConv` to ease this development. The basic usage is as follows:

```python
import dgl.nn.pytorch as dglnn
conv = dglnn.HeteroGraphConv({
    'plays' : dglnn.GraphConv(...),
    'follows' : dglnn.SAGEConv(...),
    'sells' : dglnn.GATConv(...)
    }, aggregate='sum')
```

It allows users to register NN modules for each edge type and the forward function will invoke them and aggregate their results automatically. Please refer to the `hetero.py` file for more explanation.

Also rewrite the RGCN code using this prototype.
- RGCN at its core is just a `HeteroGraphConv` with `GraphConv` on all its relations.
- Use one model definition for both full graph and mini-batch training.

## Checklist
- [x] Code is well-documented


examples/pytorch/rgcn-hetero/model.py
```diff
@@ -0,0 +1,199 @@
...
+        dict[str, torch.Tensor]
+            New node features for each node type.
+        """
+        g = g.local_var()
+        if self.basis:
+            weights = self.basis()
+            assert weights.shape == (len(self.rel_names), self.in_feat, self.out_feat)
+            wdict = {self.rel_names[i] : w.squeeze(0) for i, w in enumerate(th.split(weights, 1, dim=0))}
+            for rel, mod in self.conv.mods.items():
+                mod.weight = wdict[rel]
```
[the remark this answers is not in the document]
jermainewang: It is basically a form of weight sharing. I googled a little bit about what's the best practice. From the answer of the PyTorch developer (https://discuss.pytorch.org/t/how-to-create-model-with-sharing-weight/398), he suggested defining models like this:
```python
class MyModel(nn.Module):
    def __init__(self):
        self.base = ...
        self.head_A = ...
        self.head_B = ...

    def forward(self, input1, input2):
        return self.head_A(self.base(input1)), self.head_B(self.base(input2))
```
In our case, the `self.basis` instance is the `self.base` with no inputs. If adopting the above approach, I need to add an extra argument `weight` to `nn.GraphConv` for externally provided weight tensor. This complicates the `HeteroGraphConv` module since each submodule will have some extra arguments.

My proposal to this is extending the `HeteroGraphConv.forward` to allow more complicated argument specification. For example, we could have:
```python
hetconv = dglnn.HeteroGraphConv({'plays' : mod1, 'sells' : mod2, 'follows' : mod3})
node_inputs = {'user' : ..., 'game' : ...}
mod_args = {'plays' : (a1, a2), 'sells' : (b1, b2), 'follows' : (c1, c2)}
```
And the `mod_args` will be dispatched to each modules upon invocation. In this way, we can resolve your concern like the following:
```python
hetconv = dglnn.HeteroGraphConv({'plays' : mod1, 'sells' : mod2, 'follows' : mod3})
node_inputs = {'user' : ..., 'game' : ...}
h = hetconv(node_inputs, mod_args=wdict)
```

Thoughts?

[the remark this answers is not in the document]
classicsong: > It is basically a form of weight sharing. I googled a little bit about what's the best practice. From the answer of the PyTorch developer (https://discuss.pytorch.org/t/how-to-create-model-with-sharing-weight/398), he suggested defining models like this:
> 
> ```python
> class MyModel(nn.Module):
>     def __init__(self):
>         self.base = ...
>         self.head_A = ...
>         self.head_B = ...
> 
>     def forward(self, input1, input2):
>         return self.head_A(self.base(input1)), self.head_B(self.base(input2))
> ```
> 
> In our case, the `self.basis` instance is the `self.base` with no inputs. If adopting the above approach, I need to add an extra argument `weight` to `nn.GraphConv` for externally provided weight tensor. This complicates the `HeteroGraphConv` module since each submodule will have some extra arguments.
> 
> My proposal to this is extending the `HeteroGraphConv.forward` to allow more complicated argument specification. For example, we could have:
> 
> ```python
> hetconv = dglnn.HeteroGraphConv({'plays' : mod1, 'sells' : mod2, 'follows' : mod3})
> node_inputs = {'user' : ..., 'game' : ...}
> mod_args = {'plays' : (a1, a2), 'sells' : (b1, b2), 'follows' : (c1, c2)}
> ```
> 
> And the `mod_args` will be dispatched to each modules upon invocation. In this way, we can resolve your concern like the following:
> 
> ```python
> hetconv = dglnn.HeteroGraphConv({'plays' : mod1, 'sells' : mod2, 'follows' : mod3})
> node_inputs = {'user' : ..., 'game' : ...}
> h = hetconv(node_inputs, mod_args=wdict)

This one looks better. I like this one.

> ```
> 
> Thoughts?

python/dgl/nn/mxnet/conv/graphconv.py
```diff
@@ -80,52 +93,75 @@ def __init__(self,
...
+                               ' module has defined its own weight parameter. Please'
+                               ' create the module with flag weight=False.')
+        else:
+            weight = self.weight.data(feat.context)
+
         if self._in_feats > self._out_feats:
             # mult W first to reduce the feature size for aggregation.
-            feat = mx.nd.dot(feat, self.weight.data(feat.context))
-            graph.ndata['h'] = feat
+            if weight is not None:
```
classicsong: weight would not be None here.
We have handled it in the above.

python/dgl/nn/mxnet/conv/graphconv.py
```diff
@@ -80,52 +93,75 @@ def __init__(self,
...
                              fn.sum(msg='m', out='h'))
-            rst = graph.ndata.pop('h')
-            rst = mx.nd.dot(rst, self.weight.data(feat.context))
-
-        if self._norm:
+            rst = graph.dstdata.pop('h')
+            if weight is not None:
+                rst = mx.nd.dot(rst, weight)
+
+        if self._norm != 'none':
```
classicsong: We should check none-supported norm type here for self._norm other than make the assumption here.

python/dgl/nn/pytorch/conv/graphconv.py
```diff
@@ -92,42 +108,64 @@ def forward(self, graph, feat):
...
                              fn.sum(msg='m', out='h'))
-            rst = graph.ndata['h']
-            rst = th.matmul(rst, self.weight)
-
-        if self._norm:
+            rst = graph.dstdata['h']
+            if weight is not None:
+                rst = th.matmul(rst, weight)
+
+        if self._norm != 'none':
```
classicsong: Same as mxnet

python/dgl/nn/tensorflow/conv/graphconv.py
```diff
@@ -91,44 +105,68 @@ def call(self, graph, feat):
...
                              fn.sum(msg='m', out='h'))
-            rst = graph.ndata['h']
-            rst = tf.matmul(rst, self.weight)
-
-        if self._norm:
+            rst = graph.dstdata['h']
+            if weight is not None:
+                rst = tf.matmul(rst, weight)
+
+        if self._norm != 'none':
```
classicsong: same as mxnet

python/dgl/nn/mxnet/conv/graphconv.py
```diff
@@ -43,8 +43,13 @@ class GraphConv(gluon.Block):
...
         Number of output features.
-    norm : bool, optional
-        If True, the normalizer :math:`c_{ij}` is applied. Default: ``True``.
+    norm : str, optional
+        How to apply the normalizer. If is `'right'`, divide the aggregated messages
+        by each node's in-degrees, which is equivalent to averaging the received messages.
+        If is `'none'`, no normalization is applied. Default is `'both'`,
+        where the :math:`c_{ij}` in the paper is applied.
+    weight : bool, optional
+        If True, apply a linear layer after message passing.
```
BarclayII: After or before?

python/dgl/nn/mxnet/conv/graphconv.py
jermainewang: It could be. We could create graphconv as `nn.GraphConv(..., weight=False)` where `self.weight` is None. When calling its forward function, does not provide external weight. This will result in computing a graph convolution without linear projection. Actually, this is used in the first layer of RGCN.

python/dgl/nn/mxnet/conv/graphconv.py
jermainewang: Already did at https://github.com/dmlc/dgl/blob/4a4fbde8ea52affd48d7570d7024c7542795797b/python/dgl/nn/mxnet/conv/graphconv.py#L74

python/dgl/nn/mxnet/conv/graphconv.py
jermainewang: I should avoid using after or before. The math is f(AXW) so it does not matter whether it performs XW first or AX first.

PR #1393, opened 2020-03-25
[TF] TF backend fix and new logic to choose backend

VoVAllen:
## Description
- Remove tfdlpack and use `tf.experimental.dlpack` instead
- Change the backend choice logic
  - Now there's no default backend. User need to specify it, by `python -m dgl.backend.set_default_backend [BACKEND]` to do so. It writes the config to `~/.dgl/config.json`

Potential problem:
Is there any existing code rely on the behavior that the default backend is pytorch?

## Checklist
- [ ] Code is well-documented


jermainewang:
> Change the backend choice logic
> * Now there's no default backend. User need to specify it, by `python -m dgl.backend.set_default_backend [BACKEND]` to do so. It writes the config to `~/.dgl/config.json`

Can we make it more convenient? We could have an interactive way to set the default backend for them.
```bash
>>> import dgl
DGL does not detect a backend option. Which backend would you like to work with?
Backend choice (pytorch, mxnet or tensorflow): tensorflow
Set default backend to "tensorflow". You can change it in the ~/.dgl/config.json file or export the DGLBACKEND environment.
>>> print(dgl.backend.name)
tensorflow
```

python/dgl/backend/tensorflow/tensor.py
```diff
@@ -6,13 +6,12 @@
...
-import tfdlpack
 import numpy as np
-from tfdlpack import to_dlpack, from_dlpack
 
 from ... import ndarray as nd
 from ... import kernel as K
 from ...function.base import TargetCode
+from ...container import _api_internal
 
 TF_VERSION = LooseVersion(tf.__version__)
```
jermainewang: Can we check the TF version and throw warning to users? We might want to check MX and TH versions too.

VoVAllen: Sure

PR #1440, opened 2020-04-12
[Model] Unsupervised learning with GraphSAGE

BarclayII:
Unsupervised learning with GraphSAGE.  Can get to 90.05% accuracy on validation and 90.13% on test set on Reddit.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/src/array/cpu/spmat_op_impl_coo.cc
+++ b/src/array/cpu/spmat_op_impl_coo.cc
@@ -175,6 +175,7 @@ std::vector<NDArray> COOGetDataAndIndices(
     COOMatrix coo, NDArray rows, NDArray cols) {
   const int64_t rowlen = rows->shape[0];
   const int64_t collen = cols->shape[0];
+  const int64_t len = std::max(rowlen, collen);
 
   CHECK((rowlen == collen) || (rowlen == 1) || (collen == 1))
     << "Invalid row and col id array.";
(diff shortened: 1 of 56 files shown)


zheng-da:
The original GraphSage paper actually uses a random walk to generate pairs of nodes as an unsupervised signal. Do you want to reproduce that?

BarclayII:
> The original GraphSage paper actually uses a random walk to generate pairs of nodes as an unsupervised signal. Do you want to reproduce that?

I think their implementation iterates over edges instead of random walks.  Correct me if I'm wrong though.

jermainewang:
@BarclayII Let's finish this example. Some comments:
* Please add the results to the README.
* The paper reports a micro-F1 of 89.7 for GraphSAGE-mean. Does your 90.13% accuracy match the result?
* The paper used a loss based on co-occurred nodes on random walks. (See section 3.2 paragraph 2). Your implementation uses edge sampling which can be viewed as a co-occurrence loss defined on random walks of length two. Let's just document that clearly.

BarclayII:
I also tried to speed up the `edge_ids` on COO graphs with `unordered_multimap`.  The speed up can be 20x faster for large number of edges and lookups.  I also tried sorting and binary searching but found it only benefiting medium-sized graphs and lookups.  See https://gist.github.com/BarclayII/52b8a00b57e323ac1e7ad5fca0306c8b for more detailed benchmarking results.

VoVAllen:
Sorting seems unnecessary since this is equivalent to converting to CSR and finding edge ids. How about the memory consumption using `unordered_multimap` for extreme large graph?

BarclayII:
> Sorting seems unnecessary since this is equivalent to converting to CSR and finding edge ids.

The reason not to use CSR is that the graph may be hypersparse.  In fact, `edge_ids` tries to use CSR whenever possible, and will only fallback to COO if the graph is hypersparse or otherwise restricted to COO format.

> How about the memory consumption using `unordered_multimap` for extreme large graph?

This essentially doubles the COO memory.  If the graph is hypersparse this should be OK since the COO memory would take a lot less than CSR memory.

src/array/cpu/spmat_op_impl_coo.cc
```diff
@@ -190,16 +191,44 @@ std::vector<NDArray> COOGetDataAndIndices(
...
+  // medium-sized graphs and lookups, so I didn't include it.
+  if (len >= 200) {
+    // TODO(BarclayII) Ideally we would want to cache this object.  However I'm not sure
+    // what is the best way to do so since this object is valid for CPU only.
+    std::unordered_multimap<std::pair<IdType, IdType>, IdType, PairHash> pair_map;
+    pair_map.reserve(coo.row->shape[0]);
+    for (int64_t k = 0; k < coo.row->shape[0]; ++k)
+      pair_map.emplace(std::make_pair(coo_row_data[k], coo_col_data[k]), data ? data[k]: k);
+
+    for (int64_t i = 0, j = 0; i < rowlen && j < collen; i += row_stride, j += col_stride) {
```
jermainewang, line 210: Will openmp help here?

BarclayII: There are `push_back`'s in the loop body.

jermainewang: Couldn't we first do a parallel sweep to find the locations and then use a parallel look up to fetch the actual data?

BarclayII: For the current behavior we need to ensure that the returned edges are in the same order as given rows/cols.  I couldn't see benefit of parallel sweep when combining with `multimap`.  Correct me if I'm wrong though.

jermainewang: The problem can be formulated as: given an array of keys A and an array of queries B, return two arrays BB and P such that A[P[i]] is the matched key of BB[i]. If A has no duplication, we want BB to be equal to B. Otherwise, just make sure that BB contains all the elements in B. The query array B can have duplicates. If both A and B have duplicate, the result is undefined. The size of B is usually much smaller than A.

We first check whether B has duplicates, which is efficient since the size of B is usually small.
* If B has duplicates, let each openmp thread work on a segment of array B. Each thread scans its own segment and writes the position of the matched key in A to P. You can build a hashmap of A (usually pretty costly) or just do a linear scan and stop at the first hit. No need to handle the case that A has duplicates. Throw error if any query misses.
* Otherwise, let each openmp thread work on a segment of array A. Each thread scans its own segment and writes the position to B of the matched query to a temporary array PB. Missed key has a special mark. Then remove the missed key in PB and get two arrays P and BB. Then sort both P and BB according to BB to get the final answer. You can optionally apply a linear scan on BB and throw error if any query misses. If we expect the size of BB to be similar with that of B, the sorting is efficient. Besides, building a hashmap on B would be a good choice.

BarclayII: The problem is that we have two desired behaviors for `edge_id`:

Behavior 1: Return any edge that connects between `src` and `dst`, in the given order.
Behavior 2: Return all edges that connects between `src` and `dst`, potentially regardless of order.

I feel it's best implemented in two separate C++ interfaces, and I don't want the behavior selection to depend on "whether B has duplicates or not", because sometimes users may want to have behavior 2 even if B does not have duplicates (like in this example).  In fact, the behavior selection should depend on the argument "return_uv".  If "return_uv" is False, we assume the user wants behavior 1, and behavior 2 otherwise.

I would like to leave this refactor to a future PR (but of course before the next release) and for now I'm only focusing on making `edge_id` and `remove_edges` usable for the current example.  What do you think?

jermainewang: My solutions are all for behavior 2. When will behavior 1 be desired?

> I would like to leave this refactor to a future PR (but of course before the next release) and for now I'm only focusing on making edge_id and remove_edges usable for the current example.

Could you add an item to our v0.5 roadmap?

PR #1453, opened 2020-04-18
[Bugfix] Fix corrupted memory when using multi-processing

jermainewang:
## Description

When sampling with multi-processing, the sampler sometimes returns a corrupted original node ID array for destination nodes. The problem seems to be due to:
1. A PyTorch array is converted to a DGL array via DLPack.
1. It is then converted back to PyTorch array via DLPack. There is no data modification in the middle, so the conversion is actually redundant.
1. The array is then transmitted to another process via multi-processing mechanism.

The patch attempts to fix this by removing the redundant steps 1 and 2, though I haven't thoroughly verified whether the problem is completely solved.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/transform.py
+++ b/python/dgl/transform.py
@@ -898,8 +898,6 @@ def to_block(g, dst_nodes=None, include_dst_in_src=True):
 
     new_graph_index, src_nodes_nd, induced_edges_nd = _CAPI_DGLToBlock(
         g._graph, dst_nodes_nd, include_dst_in_src)
-    src_nodes = [F.zerocopy_from_dgl_ndarray(nodes_nd.data) for nodes_nd in src_nodes_nd]
-    dst_nodes = [F.zerocopy_from_dgl_ndarray(nodes_nd) for nodes_nd in dst_nodes_nd]
 
     # The new graph duplicates the original node types to SRC and DST sets.
     new_ntypes = ([ntype for ntype in g.ntypes], [ntype for ntype in g.ntypes])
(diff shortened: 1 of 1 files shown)


BarclayII:
Turns out to be a bug in PyTorch's DLPack implementation.  Submitted an issue in https://github.com/pytorch/pytorch/issues/36922.  The reproduction code goes as follows:
```python
import torch
from torch.utils.data import Dataset, DataLoader
from torch.utils.dlpack import to_dlpack, from_dlpack

def collate(a):
    a = torch.LongTensor(a)
    acap = to_dlpack(a)
    b = from_dlpack(acap)

    return a, b

loader = DataLoader(dataset=[i for i in range(10000)],
                        batch_size = 2000,
                        collate_fn=collate,
                        shuffle=True,
                        num_workers=10)

for i, (a, b) in enumerate(loader):
    assert torch.equal(a, b)
```

Probably because `a` and `b` refer to the same content but somehow PyTorch & DLPack mess it up.

jermainewang:
Note that I cannot reproduce the error again if the tensor is created from DGL:
```python
import torch
from torch.utils.data import Dataset, DataLoader
from torch.utils.dlpack import to_dlpack, from_dlpack
import dgl
import numpy as np

def collate(a):  # This causes the bug.
    a = torch.LongTensor(a)
    acap = to_dlpack(a)
    b = from_dlpack(acap)
    return a, b

def collate1(a):  # This works fine.
    a = np.array(a, dtype=np.int64)
    acap = dgl.ndarray.array(a).to_dlpack()
    b = from_dlpack(acap)
    return torch.tensor(a), b

loader = DataLoader(dataset=[i for i in range(10000)],
                    batch_size=200,
                    collate_fn=collate1,
                    shuffle=True,
                    num_workers=10)

for i, (a, b) in enumerate(loader):
    assert torch.equal(a, b)
```

PR #1454, opened 2020-04-19
[KVstore] Improve user infterface

aksnzhy:
## Description

1. The client automatically gets metadata (partition-book, data-name, dtype, shape).

```
client:
  name_list = client.get_data_name_list()
  for name in name_list:
      meta = client.get_data_meta(name)
```

2. Add an index to temporary files.

## Checklist
- [ ] Code is well-documented


python/dgl/contrib/dis_kvstore.py
```diff
@@ -189,23 +190,61 @@ def set_global2local(self, name, global2local=None):
...
-            data_shape = self._read_data_shape(name+'-g2l-shape')
+            data_shape = self._read_data_shape(name+'-g2l-shape-'+str(self._machine_id))
             shared_data = empty_shared_mem(name+'-g2l-', False, data_shape, 'int64')
             dlpack = shared_data.to_dlpack()
             self._data_store[name+'-g2l-'] = F.zerocopy_from_dlpack(dlpack)
 
         self._has_data.add(name+'-g2l-')
 
 
+    def set_partition_book(self, name, partition_book=None):
```
zheng-da, line 210: can you move the function back? I'm not sure if I should review this function.

python/dgl/contrib/dis_kvstore.py
```diff
@@ -296,6 +336,40 @@ def get_message_count(self):
         return self._msg_count
 
 
+    def get_data_name_list(self):
```
zheng-da, line 339: do we need to define this on the kvstore server?

aksnzhy: As written in our quip doc: All the information should be available in the kvstore server. So I remain this API on server.

python/dgl/contrib/dis_kvstore.py
aksnzhy: I just move this function from the client side to the server side.

python/dgl/contrib/dis_kvstore.py
zheng-da: I see. By "All the information should be available in the kvstore server", what I mean is that the information is available on the server, so the client can get the information from the server directly. I didn't mean we should provide this API on the server as well.

python/dgl/contrib/dis_kvstore.py
zheng-da: I see. My bad. I thought you just move the code around, which makes the review difficult.

HuGanghui:
@aksnzhy  a tiny typo in title: "infterface" -> "interface"

PR #1460, opened 2020-04-22
[NN] ChebConv update

Axeln78:
## Description
The ChebConv layer could be updated. 
Firstly, the use of ReLu at each convolutional layer is necessary for the model to properly work, and was missing in the original implementation. Secondly same floating point operation were repeated when their result could be stored in a variable and re-used. 

Therefore, changes have been made to be able to pass an activation function as argument, with default ReLu. As it is an optional argument existing scripts should be compatible with the changes.
The code has been simplified making it slightly faster. 

I removed the explicit initialisation of all parameters (xavier_uniform_) as this should be done automatically by torch in this implementation.

This implementation works well on my machine but I’d be grateful if someone could double check that it complies well, that would be great!

## Checklist
- [ ] Code is well-documented
- [ ] Related issue is referred in this PR - NA


yzh119:
Thanks very much for your contribution to improving ChebConv. I'll review this PR.

python/dgl/nn/pytorch/conv/chebconv.py
```diff
@@ -86,42 +92,52 @@ def forward(self, graph, feat, lambda_max=None):
...
-            return rst
+                h = unnLaplacian(X_1, D_sqrt, graph)
+                X_i = - 2 * re_norm * h + X_1 * 2 * (re_norm - 1) - X_0
+
+                Xt = torch.cat((Xt, X_i), 1)
+                X_1, X_0 = X_i, X_1
+
+            # forward pass
+            graph.ndata['h'] = Xt
+            graph.apply_nodes(func=self.apply_mod)
```
yzh119, line 141: It seems you are concatenating X_0, X_1, ... X_t and apply a linear transformation on the concatenation, however, this is not equivalent to the original paper where the weight W^k corresponding to X_k would not influence X_j(j≠k).

python/dgl/nn/pytorch/conv/chebconv.py
```diff
@@ -86,42 +92,52 @@ def forward(self, graph, feat, lambda_max=None):
...
         with graph.local_scope():
-            norm = th.pow(
-                graph.in_degrees().float().clamp(min=1), -0.5).unsqueeze(-1).to(feat.device)
+            D_sqrt = th.pow(graph.in_degrees().float().clamp(
+                min=1), -0.5).unsqueeze(-1).to(feat.device)
+
             if lambda_max is None:
-                lambda_max = laplacian_lambda_max(graph)
+                try:
+                    lambda_max = dgl.laplacian_lambda_max(graph)
```
yzh119, line 107: remove `dgl.`, we import this method via relative path.

python/dgl/nn/pytorch/conv/chebconv.py
```diff
@@ -31,37 +45,29 @@ class ChebConv(nn.Module):
...
-
+    
     def __init__(self,
                  in_feats,
                  out_feats,
                  k,
+                 activation=F.relu,
                  bias=True):
-        super(ChebConv, self).__init__()
+        super(Cheb_Conv, self).__init__()
```
yzh119, line 60:
```suggestion
        super(ChebConv, self).__init__()
```

python/dgl/nn/pytorch/conv/chebconv.py
Axeln78: Hi, I’m not sure what you mean by ‘influence’, could you please rephrase? 
In this implementation there is first a concatenation of all the polynomes making the feature at each node ( Fin*k ) dimensional. After, a linear layer with ReLu of size (Fin*k -> Fout ) is applied. As the linear layer has as one parameter for each dimension of Xt I don’t think that qualifies as influencing each other (?).

python/dgl/nn/pytorch/conv/chebconv.py
```diff
@@ -86,42 +92,52 @@ def forward(self, graph, feat, lambda_max=None):
...
-                graph.in_degrees().float().clamp(min=1), -0.5).unsqueeze(-1).to(feat.device)
+            D_sqrt = th.pow(graph.in_degrees().float().clamp(
+                min=1), -0.5).unsqueeze(-1).to(feat.device)
+
             if lambda_max is None:
-                lambda_max = laplacian_lambda_max(graph)
+                try:
+                    lambda_max = dgl.laplacian_lambda_max(graph)
+                except BaseException:
+                    # if the largest eigonvalue is not found
```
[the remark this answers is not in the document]
Axeln78: dgl_warning added in commit c852534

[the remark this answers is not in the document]
Axeln78: Actually, a CUDA error during dgl.laplacian_lambda_max is not caught by DGL so this statement is vain. Should I open an issue for that?

yzh119:
Hi @Axeln78 , would you mind giving me write access to your dgl fork so that I do some changes directly?

Axeln78:
Hi, Writing access has been added

yzh119:
@Axeln78 thanks. Actually I'm working on DGL core infrastructure refactorization, will go back and merge this PR next week.

python/dgl/nn/pytorch/conv/chebconv.py
yzh119: Sorry I misunderstood your code, you are right.

Axeln78:
Hi!
I just saw that the code provided in the DGL Documentation hadn’t changed, is that normal?

VoVAllen:
@Axeln78 Hi, please refer to the latest branch at https://docs.dgl.ai/en/latest/api/python/nn.pytorch.html#chebconv

PR #1467, opened 2020-04-24
[Refactor] Explicit dtype for HeteroGraph

VoVAllen:
## Description
Discrepancy:
- TF only supports int64 indexed sparse tensor, therefore `g.adjacency_matrix` returns an int64 sparse matrix for an int32 graph

## Checklist
- [ ] Code is well-documented


python/dgl/graph_index.py
```diff
@@ -864,6 +864,17 @@ def ctx(self):
         """
         return _CAPI_DGLGraphContext(self)
 
+    @property
+    def dtype(self):
```
jermainewang, line 868: How about `idtype`?

python/dgl/graph_index.py
```diff
@@ -864,6 +864,17 @@ def ctx(self):
...
+        """Return the index dtype
+
+        Returns
+        ----------
+        str
+            The dtype of graph index
+        """
+        return "int64"
```
jermainewang, line 876: Why it's always int64?

python/dgl/heterograph.py
```diff
@@ -1143,7 +1143,7 @@ def has_nodes(self, vids, ntype=None):
         --------
         has_node
         """
-        vids = utils.toindex(vids)
+        vids = utils.toindex(vids, self._graph.dtype)
```
jermainewang, line 1146: Add a new API `idtype` to heterograph too.

python/dgl/heterograph.py
```diff
@@ -2052,9 +2053,11 @@ def node_type_subgraph(self, ntypes):
                 edge_frames.append(self._edge_frames[i])
 
         metagraph = graph_index.from_edge_list(meta_edges, True)
+        # num_nodes_per_type doesn't need to be int32
         hgidx = heterograph_index.create_heterograph_from_relations(
-            metagraph, rel_graphs, utils.toindex(num_nodes_per_type))
-        hg = DGLHeteroGraph(hgidx, ntypes, induced_etypes, node_frames, edge_frames)
+            metagraph, rel_graphs, utils.toindex(num_nodes_per_type, "int64"))
```
jermainewang, line 2058: Why not make the dtype consistent with `rel_graphs`?

python/dgl/heterograph_index.py
```diff
@@ -149,6 +150,7 @@ def ctx(self):
         """
         return _CAPI_DGLHeteroContext(self)
 
+    @utils.cached_member(cache='_cache', prefix='nbits')
```
jermainewang, line 153: Why cache it? Actually, shouldn't we just remove it?

python/dgl/heterograph_index.py
```diff
@@ -298,7 +300,7 @@ def has_nodes(self, ntype, vids):
             0-1 array indicating existence
         """
         vid_array = vids.todgltensor()
-        return utils.toindex(_CAPI_DGLHeteroHasVertices(self, int(ntype), vid_array))
+        return utils.toindex(_CAPI_DGLHeteroHasVertices(self, int(ntype), vid_array), self.dtype)
```
jermainewang, line 303: When constructing `utils.Index` from a DGL array, we could omit the id type argument since the array contains the info.

python/dgl/runtime/scheduler.py
```diff
@@ -484,8 +484,9 @@ def schedule_pull(graph,
             schedule_apply_nodes(pull_nodes, apply_func, graph.dstframe, inplace,
                                  outframe, ntype=graph.canonical_etype[-1])
     else:
+        # TODO(Allen): Change operation to dgl operation
```
jermainewang, line 487: Which operation?

python/dgl/transform.py
```diff
@@ -463,6 +463,7 @@ def metapath_reachable_graph(g, metapath):
         A homogeneous or bipartite graph.
     """
     adj = 1
+    index_dtype = g._graph.dtype
```
jermainewang, line 466: It's better to directly use `g.idtype`. Usually we don't rely on graph index any more in those transform functions.

python/dgl/transform.py
```diff
@@ -719,13 +720,13 @@ def compact_graphs(graphs, always_preserve=None):
...
     ntypes = graphs[0].ntypes
-    graph_dtype = graphs[0]._graph.dtype()
+    graph_dtype = graphs[0]._graph.dtype
     graph_ctx = graphs[0]._graph.ctx()
     for g in graphs:
         assert ntypes == g.ntypes, \
             ("All graphs should have the same node types in the same order, got %s and %s" %
              ntypes, g.ntypes)
-        assert graph_dtype == g._graph.dtype(), "Graph data type mismatch"
+        assert graph_dtype == g._graph.dtype, "Graph data type mismatch"
```
jermainewang, line 729: Let's write a helper function for these kinds of idtype checking (and also the context checking below). One example is from torch's concat operator:
```
RuntimeError: Expected object of scalar type Long but got scalar type Int for sequence element 1 in sequence argument at position #1 'tensors'
```

python/dgl/transform.py
```diff
@@ -945,8 +946,11 @@ def remove_edges(g, edge_ids):
                 "Graph has more than one edge type; specify a dict for edge_id instead.")
         edge_ids = {g.canonical_etypes[0]: edge_ids}
 
-    edge_ids_nd = [nd.null()] * len(g.etypes)
+    edge_ids_nd = [nd.null(g._graph.dtype)] * len(g.etypes)
     for key, value in edge_ids.items():
+        if value.dtype != getattr(F, g._graph.dtype):
+            # if didn't check, this function still works, but returns wrong result
+            raise Exception("Inconsistent dtype between edge ids and graph data type")
```
jermainewang, line 953: I would like to see a helper function like this:
```python
check_idtype_dict({k : v.idtype for k,v in edge_ids}, g.idtype)
```
You could also have things like
```python
check_idtype_list(idtype_list, g.idtype)
```

python/dgl/utils.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=C0103
```
jermainewang, line 1: What is this?

python/dgl/utils.py
```diff
@@ -9,9 +10,35 @@
...
+        F.int64: 'int64'
+    }
+
+
+def STR2NPDTYPE():
+    """Convert string to np dtype"""
+    return {
+        'int32': np.int32,
+        'int64': np.int64
+    }
```
jermainewang, line 35: Why not directly define them as:
```python
STR2NPDTYPE = {
  'int32' : np.int32,
  'int64' : np.int64
}
```
Besides, you probably could just use `getattr` instead of having these dictionaries.

python/dgl/utils.py
```diff
@@ -43,18 +70,22 @@ def __getitem__(self, i):
     def _dispatch(self, data):
         """Store data based on its type."""
         if F.is_tensor(data):
-            if F.dtype(data) != F.int64:
-                raise DGLError('Index data must be an int64 vector, but got: %s' % str(data))
+            if F.dtype(data) != STR2DTYPE()[self.dtype]:
+                raise DGLError('Index data specified as %s, but got: %s' %
+                               (self.dtype, DTYPE2STR()[F.dtype(data)]))
+                # self.dtype = DTYPE2STR()[F.dtype(data)]
```
jermainewang, line 76: remove

python/dgl/utils.py
```diff
@@ -43,18 +70,22 @@ def __getitem__(self, i):
     def _dispatch(self, data):
         """Store data based on its type."""
         if F.is_tensor(data):
-            if F.dtype(data) != F.int64:
-                raise DGLError('Index data must be an int64 vector, but got: %s' % str(data))
+            if F.dtype(data) != STR2DTYPE()[self.dtype]:
+                raise DGLError('Index data specified as %s, but got: %s' %
+                               (self.dtype, DTYPE2STR()[F.dtype(data)]))
```
jermainewang, line 75: Does `str(F.dtype(data))` work?

python/dgl/utils.py
```diff
@@ -133,6 +164,7 @@ def __getstate__(self):
             return self.tousertensor()
 
     def __setstate__(self, state):
+        self.dtype = "int64"
```
jermainewang, line 167: Is this correct?

python/dgl/utils.py
```diff
@@ -155,18 +187,23 @@ def get_items(self, index):
...
-            return Index(F.gather_row(tensor, index))
+            # TODO(Allen): Change F.gather_row to dgl operation
+            return Index(F.astype(F.gather_row(tensor, index),
+                                  F.data_type_dict[self.dtype]), self.dtype)
         elif self._slice_data is None:
             # the current index is not a slice but the provided is a slice
             tensor = self.tousertensor()
             index = index._slice_data
-            return Index(F.narrow_row(tensor, index.start, index.stop))
+            # TODO(Allen): Change F.narrow_row to dgl operation
```
jermainewang, line 197: Why do you wish to use DGL operation?

python/dgl/utils.py
```diff
@@ -155,18 +187,23 @@ def get_items(self, index):
             # the provided index is not a slice
             tensor = self.tousertensor()
             index = index.tousertensor()
-            return Index(F.gather_row(tensor, index))
+            # TODO(Allen): Change F.gather_row to dgl operation
+            return Index(F.astype(F.gather_row(tensor, index),
+                                  F.data_type_dict[self.dtype]), self.dtype)
```
jermainewang, line 192: Is `F.astype` needed? The result should have the same dtype as `tensor` right?

python/dgl/view.py
```diff
@@ -277,7 +277,8 @@ def __getitem__(self, key):
 
     def __call__(self, ntype=None):
         """Return the nodes."""
-        return F.arange(0, self._graph.number_of_nodes(ntype))
+        return F.arange(0, self._graph.number_of_nodes(ntype),
+                        dtype=F.data_type_dict[self._graph._graph.dtype])
```
jermainewang, line 281: This makes me think whether we should directly return backend dtype object in `g.idtype`. This would make everything much easier.

jermainewang:
Could you also add two APIs for converting a graph between different id types?

VoVAllen:
@jeraminewang What's the difference between the API you mentioned and `g.asbits`?

python/dgl/convert.py
```diff
@@ -386,10 +387,14 @@ def hetero_from_relations(rel_graphs, num_nodes_per_type=None):
...
     ntype_dict = {ntype: i for i, ntype in enumerate(ntypes)}
     for rgrh in rel_graphs:
+        if rgrh._graph.dtype != index_dtype:
+            raise Exception("Expect relation graphs to be {}, but got {}".format(
+                index_dtype, rgrh._graph.dtype))
         stype, etype, dtype = rgrh.canonical_etypes[0]
         meta_edges_src.append(ntype_dict[stype])
         meta_edges_dst.append(ntype_dict[dtype])
         etypes.append(etype)
+    # TODO(Allen): metagraph is homograph, currently still using int64
```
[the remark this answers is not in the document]
VoVAllen: This PR didn't change `DGLGraph` part. Currently metagraph is a `DGLGraph` instead of a homo `DGLHeteroGraph`, and `DGLGraph` are all int64 graphs. Therefore currently here using int64. I think it's fine to keep it int64. How do you think about this?

python/dgl/convert.py
```diff
@@ -280,22 +280,23 @@ def bipartite(data, utype='_U', etype='_E', vtype='_V', num_nodes=None, card=Non
...
-        return create_from_networkx_bipartite(
-            data, utype, etype, vtype, restrict_format=restrict_format, **kwargs)
+        return create_from_networkx_bipartite(data, utype, etype,
+                                              vtype, restrict_format=restrict_format,
+                                              index_dtype=index_dtype, **kwargs)
     else:
         raise DGLError('Unsupported graph data type:', type(data))
 
-def hetero_from_relations(rel_graphs, num_nodes_per_type=None):
+def hetero_from_relations(rel_graphs, num_nodes_per_type=None, index_dtype='int64'):
```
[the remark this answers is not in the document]
VoVAllen: Removed `index_dtype` in `hetero_from_relations`

python/dgl/utils.py
VoVAllen: Fixed

python/dgl/utils.py
VoVAllen: Some backend operation doesn't support int32 indexing therefore manually casting to long is needed in some cases. And these operation on 1D vector should be simple to implement, changing to DGL operation might be better?

python/dgl/heterograph_index.py
VoVAllen: This is for sanity check to avoid mistake. Do you think it's necessary for DGL array to specify extra dtype when converting to `Index`?

python/dgl/heterograph_index.py
VoVAllen: Do you mean only `gidx.dtype` should be preserved?

python/dgl/heterograph.py
VoVAllen: Sure

python/dgl/utils.py
VoVAllen: STR2DTYPE() uses backend `F` which causes cyclical import. I'll refactor this part

python/dgl/utils.py
VoVAllen: Invalid constant name. Related to the function such as STR2DTYPE, will refactor those

python/dgl/view.py
VoVAllen: There are numpy dtype, backend dtype and string literal dtype here. Conversions are needed at multiple places. String literal dtype is easiest for conversion and comparison. Therefore I choose str finally.

python/dgl/graph_index.py
VoVAllen: Fixed

python/dgl/backend/tensorflow/tensor.py
```diff
@@ -77,8 +77,8 @@ def sparse_matrix(data, index, shape, force_format=False):
     if fmt != 'coo':
         raise TypeError(
             'Tensorflow backend only supports COO format. But got %s.' % fmt)
-    spmat = tf.SparseTensor(indices=tf.transpose(
-        index[1], (1, 0)), values=data, dense_shape=shape)
+    spmat = tf.SparseTensor(indices=tf.cast(tf.transpose(
```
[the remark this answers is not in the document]
VoVAllen: Added

python/dgl/convert.py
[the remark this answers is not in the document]
VoVAllen: Removed index_dtype in hetero_from_relations

python/dgl/heterograph.py
VoVAllen: Currently data related to metagraph are all int64. And the C++ `HeteroGraph` uses `std::vector<int64_t> num_verts_per_type_`. This is mainly for simplicity, otherwise lots of extra changes may needed.

python/dgl/runtime/scheduler.py
VoVAllen: I was just thinking of changing all operation on index to dgl operation, without backend operation, to avoid potential unsupported or dtype casting issue. However I'm fine with the current code.

python/dgl/utils.py
VoVAllen: Refactored

VoVAllen:
@jermainewang Do you mean `G.idtype` or `g_index.idtype`?

python/dgl/convert.py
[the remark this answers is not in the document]
jermainewang: I see. I think as long as we document the behavior clearly (e.g., in the C++ class docstring), we are fine.

python/dgl/heterograph.py
jermainewang: I see. You could put that in the C++ header docstring.

python/dgl/heterograph_index.py
jermainewang: Yes. I suppose `gidx.dtype` covers `gidx.nbits`.

jermainewang:
> @jeraminewang What's the difference between the API you mentioned and `g.asbits`?

I'm thinking of something like `GG = G.int()` and `GG = G.long()`. What do you think?

python/dgl/view.py
jermainewang: I get there are multiple options here, but which one is more convenient for the common cases? That should to be backend dtype right?

VoVAllen:
@jermainewang For `G.int()` and `G.long()`, can we return a new graph with the same Frame as the original one but with a new graph index?

tests/compute/test_heterograph.py
```diff
@@ -658,15 +671,15 @@ def check_mapping(g, fg):
             src_fg, dst_fg = fg.find_edges([i])
             # TODO(gq): I feel this code is quite redundant; can we just add new members (like
             # "induced_srcid") to returned heterograph object and not store them as features?
-            assert src_g == F.gather_row(fg.nodes[SRC].data[dgl.NID], src_fg)[0]
+            assert F.asnumpy(src_g) == F.asnumpy(F.gather_row(fg.nodes[SRC].data[dgl.NID], src_fg)[0])
             tid = F.asnumpy(F.gather_row(fg.nodes[SRC].data[dgl.NTYPE], src_fg)).item()
             assert g.canonical_etypes[etype][0] == g.ntypes[tid]
-            assert dst_g == F.gather_row(fg.nodes[DST].data[dgl.NID], dst_fg)[0]
+            assert F.asnumpy(dst_g) == F.asnumpy(F.gather_row(fg.nodes[DST].data[dgl.NID], dst_fg)[0])
```
jermainewang, line 677: Why need an extra asnumpy here?

tests/compute/test_randomwalk.py
```diff
@@ -58,9 +60,11 @@ def test_random_walk_with_restart():
             trace_diff = np.diff(F.zerocopy_to_numpy(t), axis=-1)
             assert (trace_diff % 2 == 0).all()
 
-def test_metapath_random_walk():
-    g1 = dgl.bipartite(([0, 1, 2, 3], [0, 1, 2, 3]), 'a', 'ab', 'b')
-    g2 = dgl.bipartite(([0, 0, 1, 1, 2, 2, 3, 3], [1, 3, 2, 0, 3, 1, 0, 2]), 'b', 'ba', 'a')
+# @parametrize_dtype
+@pytest.mark.parametrize("index_dtype", ["int64"])
```
jermainewang, line 64: Why only int64?

jermainewang:
> @jermainewang For `G.int()` and `G.long()`, can we return a new graph with the same Frame as the original one but with a new graph index?

That's what's in my mind.

tests/backend/__init__.py
```diff
@@ -48,12 +48,17 @@ def ones(shape, dtype=float32, ctx=_default_context):
 def randn(shape):
     return copy_to(_randn(shape), _default_context)
 
+default_dtype = None
+
+def set_default_dtype(dtype):
+    default_dtype = dtype
```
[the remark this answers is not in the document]
VoVAllen: It's not used. Will remove this

src/array/cpu/spmat_op_impl.cc
```diff
@@ -43,6 +43,8 @@ template bool CSRIsNonZero<kDLCPU, int64_t>(CSRMatrix, int64_t, int64_t);
 
 template <DLDeviceType XPU, typename IdType>
 NDArray CSRIsNonZero(CSRMatrix csr, NDArray row, NDArray col) {
+  CHECK_SAME_DTYPE(csr.indices, row);
+  CHECK_SAME_DTYPE(csr.indices, col);
```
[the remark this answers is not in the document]
VoVAllen: Usually the error would be caught when generating `Index` with specified dtype. Here's mainly for sanity check, which means error happens only when there's mistake in our implementations.

tests/compute/test_heterograph.py
VoVAllen: MXNet raises error when comparing ndarray with int. Related to https://github.com/apache/incubator-mxnet/issues/16099

src/graph/heterograph_capi.cc
```diff
@@ -431,20 +431,36 @@ DGL_REGISTER_GLOBAL("heterograph_index._CAPI_DGLHeteroDisjointUnion")
     List<HeteroGraphRef> component_graphs = args[1];
     std::vector<HeteroGraphPtr> component_ptrs;
     component_ptrs.reserve(component_graphs.size());
+    int64_t bits = component_graphs[0]->NumBits();
```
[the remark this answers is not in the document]
VoVAllen: It's checked in previous stage and makes no sense to send empty graph list to batch

python/dgl/view.py
VoVAllen: It's not convenient to convert backend dtype to np dtype(to initialize dgl ndarray, which is widely used). It's something like `getattr(np, F.reverse_data_dtype[dtype])`. I still prefer using string to represent dtypes

src/graph/heterograph_capi.cc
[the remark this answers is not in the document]
jermainewang: Then please add an assertion.

src/array/cpu/spmat_op_impl.cc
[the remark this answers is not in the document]
jermainewang: I know. The check on the python side is mainly for more informative error message.

python/dgl/view.py
jermainewang: > It's not convenient to convert backend dtype to np dtype(to initialize dgl ndarray, which is widely used). It's something like `getattr(np, F.reverse_data_dtype[dtype])`. I still prefer using string to represent dtypes

Is that visible to user? Can we allow initialize dgl ndarray with backend dtype?

python/dgl/utils.py
VoVAllen: Removed

python/dgl/view.py
VoVAllen: Which api are you referring to(visible to user)? Why do you think backend dtype is better? How about another alias?

VoVAllen:
Remaining issue:
- Flatten operation refactor using concat
  - I'm not sure how to do this. Maybe another PR to do so?
- Explcit dtype in F.arange
  - There's lots of legacy codes therefore I still keep the default value. However I added explicit dtype in heterograph 
- Metapath sampling
  - still not working on int32, trying to fix

VoVAllen:
Remaining issue:

- Flatten operation refactor using concat
I'm not sure how to do this. Maybe another PR to do so?
- Explcit dtype in F.arange
There's lots of legacy codes therefore I still keep the default value. However I added explicit dtype in heterograph

include/dgl/array.h
```diff
@@ -880,6 +892,29 @@ IdArray VecToIdArray(const std::vector<T>& vec,
...
+ *   DType *data = static_cast<DType *>(array->data);
+ * });
+ */
+#define ATEN_ID_BITS_SWITCH(bits, IdType, ...)                  \
+  do {                                                          \
+    CHECK(bits == 32 || bits == 64) << "bits must be 32 or 64"; \
+    if (bits == 32) {                                           \
+      typedef int32_t IdType;                                   \
+      { __VA_ARGS__ }                                           \
+    } else if (bits == 64) {                                    \
```
jermainewang, line 910: `bits` -> `(bits)`. Always put parenthesis on macro argument.

python/dgl/utils.py
```diff
@@ -9,9 +9,19 @@
 from . import backend as F
 from . import ndarray as nd
 
+
+class InconsistentDtypeException(DGLError):
+    """Exception class for inconsistent dtype between graph and tensor"""
+    def __init__(self, msg='', *args, **kwargs): #pylint: disable=W1113
+        prefix_message = 'DGL now requires the input tensor to have\
+            the same dtype as the graph index\'s dtype(which you can get by g.idype).'
```
jermainewang, line 17: `Require the input tensor to have the same dtype as the graph's ID type (which ...), but got {}'

src/array/cpu/array_utils.h
```diff
@@ -10,6 +10,13 @@
 #include <vector>
 #include <unordered_map>
 #include <utility>
+#include "../../c_api_common.h"
+
+#define CHECK_SAME_DTYPE(VAR1, VAR2)                                          \
+  CHECK(VAR1->dtype == VAR2->dtype)                                           \
+    << "Expected " << (#VAR2) << " to be the same type as " << (#VAR1) << "(" \
+    << VAR1->dtype << ")"                                                     \
+    << ". But got " << VAR2->dtype;
```
jermainewang, line 19: (VAR1) and (VAR2)

mufeili:
Personally I would be more inclined to return a backend data type object, which seems more natural for users from different backends.

tests/compute/test_transform.py
```diff
@@ -594,22 +615,22 @@ def test_cast():
...
+    # test_remove_self_loop()
+    # test_add_self_loop()
+    # test_partition_with_halo()
+    # test_metis_partition()
+    # test_compact()
+    # test_to_simple()
+    # test_in_subgraph("int32")
+    # test_out_subgraph()
+    test_to_block("int32")
+    # test_remove_edges()
```
yzh119, line 636: Why disabling these tests?

VoVAllen: It's s not disabled. Here is just for debug tests

PR #1470, opened 2020-04-25
[Model] GATNE-T

Linfang-He:
Add an example for GATNE-T


yzh119:
Hi @Linfang-He thanks for your contribution, would you mind packing them into a zip/tar.gz and uploading them to cloud storage?

BarclayII:
@Linfang-He You don't have to pack them to a cloud storage; just describe where to get the data and how to put it in this directory with a README and you should be good.

examples/pytorch/GATNE-T/src/main.py
```diff
@@ -0,0 +1,313 @@
...
+from numpy import random
+from torch.nn.parameter import Parameter
+import dgl
+import dgl.function as fn
+
+from utils import *
+
+
+# Build graph, treat all nodes as the same type
+def get_graph(network_data, vocab):
```
zheng-da, line 21: please provide docstring to explain what are the inputs, what is their type, etc.

Linfang-He: Thanks for your suggestion. Already add comments for it.

examples/pytorch/GATNE-T/src/main.py
```diff
@@ -0,0 +1,325 @@
...
+
+    # embs: [batch_size, embedding_size]
+    def forward(self, block):
+        input_nodes = block.srcdata[dgl.NID]
+        output_nodes = block.dstdata[dgl.NID]
+        batch_size = block.number_of_dst_nodes('user')
+        node_embed = self.node_embeddings
+        node_type_embed = []
+
+        with block.local_scope():
```
zheng-da, line 101: why do we need `local_scope` here?

examples/pytorch/GATNE-T/src/main.py
```diff
@@ -0,0 +1,325 @@
...
+            
+            node_type_embed = torch.matmul(attention, node_type_embed).view(-1, 1, self.embedding_u_size) 
+            node_embed = node_embed[output_nodes].unsqueeze(1).repeat(1, self.edge_type_count, 1) + \
+                torch.matmul(node_type_embed, trans_w).view(-1, self.edge_type_count, self.embedding_size)
+            last_node_embed = F.normalize(node_embed, dim=2)
+        
+            return last_node_embed  # [batch_size, edge_type_count, embedding_size]
+
+
+class NSLoss(nn.Module):
```
zheng-da, line 134: what is this loss? please add comments. btw, please check out the previous comment. Python comments should be written under the class or function definition.

examples/pytorch/GATNE-T/src/utils.py
```diff
@@ -0,0 +1,191 @@
...
+    pairs = []
+    skip_window = window_size // 2
+    for layer_id, walks in enumerate(all_walks):
+        for walk in walks:
+            for i in range(len(walk)):
+                for j in range(1, skip_window + 1):
+                    if i - j >= 0:
+                        pairs.append((walk[i], walk[i-j], layer_id))
+                    if i + j < len(walk):
+                        pairs.append((walk[i], walk[i+j], layer_id))
```
zheng-da, line 128: this looks very slow.

examples/pytorch/GATNE-T/src/main.py
Linfang-He: It does not matter whether we use `local_scape` or not.

examples/pytorch/GATNE-T/src/utils.py
Linfang-He: This can be regarded as part of data preprocessing, and it does not affect training speed that much.

examples/pytorch/GATNE-T/src/main.py
```diff
@@ -0,0 +1,325 @@
...
+    graphs = []
+    num_nodes = len(vocab)
+
+    for edge_type in network_data:
+        tmp_data = network_data[edge_type]
+        edges = []
+        for edge in tmp_data:
+            edges.append((vocab[edge[0]], vocab[edge[1]]))
+            edges.append((vocab[edge[1]], vocab[edge[0]]))
+        g = dgl.graph(edges, 'user', edge_type, num_nodes)
```
BarclayII, line 44: The graph in GATNE-T only has one node type.  Consider removing the name `user`.

The APIs in turn becomes something like:
```python
block.number_of_dst_nodes('user') -> block.number_of_dst_nodes()
block.srcnodes['user'].data -> block.srcdata
block.dstnodes['user'].data -> block.dstdata
```

PR #1474, opened 2020-04-27
[Refactor] Restrict sparse format for DGLHeteroGraph

yzh119:
## Description
In this PR we further restrict the behavior of DGLHeteroGraph so that we will only create a sparse matrix of one format when `restrict_format` is set to `coo`, `csr` or `csc`. This endows user the freedom to control the sparse matrix format DGL would created, in some cases it's costly to convert COO matrix to CSR matrix, and the CSR matrix is not necessary for message passing.

We provide three APIs for user:
* `restrict_format`: querying the restrict sparse format of the current graph.
*  `format_in_use`: querying the sparse format that is *in use* of the current graph.
* `to_format`: convert the graph to a new format.

Other changes:
* Now when you call `get_relation_graph`, it will create a new `HeteroGraph` class in the C++ end which holds one `UnitGraph` rather then a bare `UnitGraph` class for consistency (e.g. all APIs in DGLHeteroGraph class in the python end are now available for sliced relation graphs).
  * The kernel code was influenced where we need remove a HeteroGraph instead of UnitGraph to GPU.
* Introduce a new sparse format `auto`, the reason is as follows:
  * The meaning of `any` format is not clear, I prefer to explain it as: "all sparse format are allowed to be created(but only when necessary)" and it should be at the same level as `coo`/`csr`/`csc`. However, previous when user set restrict format to `any`, DGL may change the format to `coo` when graph is hypersparse.
  * To handle the case mentioned above, we use `auto` to denote "user does not care about which format to use, and let the system to decide the best format", `auto` is a temporary format that would be materialized once a `UnitGraph` class is created.

## Checklist
- [x] Code is well-documented


include/dgl/array.h
```diff
@@ -56,6 +56,18 @@ inline SparseFormat ParseSparseFormat(const std::string& name) {
...
+inline std::string ToStringSparseFormat(SparseFormat sparse_format) {
+  if (sparse_format == SparseFormat::kCOO)
+    return std::string("coo");
+  else if (sparse_format == SparseFormat::kCSR)
+    return std::string("csr");
+  else if (sparse_format == SparseFormat::kCSC)
+    return std::string("csc");
+  else
+    return std::string("any");
+}
```
jermainewang, line 69: Since `SparseFormat` is a new style enum class, you could define type conversion operator for it:
```c++
operator std::string() const;
```

include/dgl/base_heterograph.h
```diff
@@ -368,6 +368,27 @@ class BaseHeteroGraph : public runtime::Object {
    */
   virtual SparseFormat SelectFormat(dgl_type_t etype, SparseFormat preferred_format) const = 0;
 
+  /*!
+   * \brief Get the sparse format of the graph.
+   * 
+   * This function is only implemented for UnitGraph.
+   * \return a string representing the sparse format: 'coo'/'csr'/'csc'/'any'
+   */
```
jermainewang, line 376: The docstring should be self-contained. No need to mention UnitGraph. Also, why not return `SparseFormat` object?

include/dgl/base_heterograph.h
```diff
@@ -368,6 +368,27 @@ class BaseHeteroGraph : public runtime::Object {
...
+
+  /*!
+   * \brief Set the sparse format of the graph.
+   * 
+   * This function is only implemented for UnitGraph.
+   * \param name String representation of the sparse format: 'coo'/'csr'/'csc'/'any'.
+   */
+  virtual void SetSparseFormat(const std::string& name) {
+    LOG(FATAL) << "Not implemented.";
+  }
```
jermainewang, line 390: Is this for format conversion? I suggest implement it as a function like:
```c++
HeteroGraphPtr ToFormat(HeteroGraphPtr , SparseFormat );
```

python/dgl/heterograph.py
```diff
@@ -4110,6 +4110,32 @@ def is_homograph(self):
         """Return if the graph is homogeneous."""
         return len(self.ntypes) == 1 and len(self.etypes) == 1
 
+    def sparse_format(self, etype=None):
```
jermainewang, line 4113: `sparse_format` -> `format`

python/dgl/heterograph.py
```diff
@@ -4110,6 +4110,32 @@ def is_homograph(self):
...
+
+        Parameters
+        ----------
+        etype : str, optional
+            The edge/relation type. Can be omitted if there is only
+            one node type in the graph. (Default: None)
+
+        Returns
+        -------
+        string : 'any', 'coo', 'csr', or 'csc'
```
jermainewang, line 4124: Return `any` is quite strange to be honest. Can we have two APIs:
```python
def format(self, etype=None, return_all=False):
    """Return the sparse formats in use of the given edge/relation type.

    Returns
    --------
    string or list of string
        If return_all is True, return all the formats currently in use (could be multiple).
        Otherwise, return any one that is in use.
    """

def restrict_format(self, etype=None):
    """Return the allowed sparse formats of the given edge/relation type.
    """
```
To change the format we have:
```python
def to_format(self, format, etype=None):
    """Return a clone graph but stored in the given sparse format.

    If 'any' is given, the restrict formats of the returned graph is relaxed.

    Parameters
    -------------
    A new graph.
    """
```

src/graph/heterograph_capi.cc
```diff
@@ -452,6 +452,31 @@ DGL_REGISTER_GLOBAL("heterograph_index._CAPI_DGLHeteroDisjointPartitionBySizes")
...
+DGL_REGISTER_GLOBAL("heterograph_index._CAPI_DGLHeteroGetSparseFormat")
+.set_body([] (DGLArgs args, DGLRetValue* rv) {
+    HeteroGraphRef hg = args[0];
+    dgl_type_t etype = args[1];
+    CHECK_LE(etype, hg->NumEdgeTypes()) << "invalid edge type " << etype;
+    auto bg = std::dynamic_pointer_cast<UnitGraph>(hg.sptr());
+    if (bg != nullptr)  // hg itself is a unit graph.
+      *rv = hg->GetSparseFormat();
+    else
+      *rv = hg->GetRelationGraph(etype)->GetSparseFormat();
```
jermainewang, line 464: Will we still get a UnitGraph from Python? I thought I've patched everywhere ...

src/graph/unit_graph.cc
```diff
@@ -1269,6 +1278,12 @@ HeteroGraphPtr UnitGraph::GetAny() const {
 }
 
 HeteroGraphPtr UnitGraph::GetFormat(SparseFormat format) const {
```
jermainewang, line 1280: We probably should rename this function to be `GetGraphInFormat`.

include/dgl/array.h
yzh119: I'm not sure if it's true, https://stackoverflow.com/questions/5093460/how-to-convert-an-enum-type-variable-to-a-string it's said enum is not a class in c++.

jermainewang: SparseFormat is a scoped enum class instead of an old style enum. Scoped enum class is strongly-typed, so I could it should allow operator overloading.

python/dgl/heterograph.py
yzh119: rename format to format_in_use because format is a reserved word in python class.

yzh119:
@jermainewang this PR is ready for review.
I still cannot make the enum class conversion operator work, other concerns have been resolved.

src/graph/unit_graph.h
```diff
@@ -202,13 +204,13 @@ class UnitGraph : public BaseHeteroGraph {
   static HeteroGraphPtr CopyTo(HeteroGraphPtr g, const DLContext& ctx);
 
   /*! \return Return the in-edge CSR format. Create from other format if not exist. */
-  CSRPtr GetInCSR() const;
+  CSRPtr GetInCSR(bool inplace = true) const;
```
BarclayII, line 207: Explain the behavior of `inplace` in the docstring?

yzh119:
@BarclayII resolved.

BarclayII:
I'm good.

PR #1498, opened 2020-05-05
[BUG] Fix bug in GIN Dataset degree OHE logic

m30m:
In the previous code, the same vector was being assigned to all the nodes in each graph.
`:` was used incorrectly assuming that only one row will be selected for each column but all rows were selected for each column.

## Checklist


VoVAllen:
Thanks!

Issue #1494, opened 2020-05-02
Impossible to import dgl 0.4.2


gcorso:
I'm having this problem importing dgl 0.4.2:

DGL Version (e.g., 1.0): dgl-0.4.2
Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): pytorch 1.5.0
OS (e.g., Linux): Windows10
How you installed DGL (conda, pip, source): conda install -c dglteam dgl
Build command you used (if compiling from source):
Python version: python 3.7.6
CUDA/cuDNN version (if applicable): none
GPU models and configuration (e.g. V100): none

```
Traceback (most recent call last):
  File "main_molecules_graph_regression.py", line 9, in <module>
    import dgl
  File "C:\Users\gabri\Python\Anaconda3\envs\benchmark_gnn2\lib\site-packages\dgl\__init__.py", line 8, in <module>
    from .backend import load_backend
  File "C:\Users\gabri\Python\Anaconda3\envs\benchmark_gnn2\lib\site-packages\dgl\backend\__init__.py", line 48, in <module>
    load_backend(os.environ.get('DGLBACKEND', 'pytorch').lower())
  File "C:\Users\gabri\Python\Anaconda3\envs\benchmark_gnn2\lib\site-packages\dgl\backend\__init__.py", line 18, in load_backend
    mod = importlib.import_module('.%s' % mod_name, __name__)
  File "C:\Users\gabri\Python\Anaconda3\envs\benchmark_gnn2\lib\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "C:\Users\gabri\Python\Anaconda3\envs\benchmark_gnn2\lib\site-packages\dgl\backend\pytorch\__init__.py", line 1, in <module>
    from .tensor import *
  File "C:\Users\gabri\Python\Anaconda3\envs\benchmark_gnn2\lib\site-packages\dgl\backend\pytorch\tensor.py", line 5, in <module>
    import torch as th
  File "C:\Users\gabri\Python\Anaconda3\envs\benchmark_gnn2\lib\site-packages\torch\__init__.py", line 81, in <module>
    ctypes.CDLL(dll)
  File "C:\Users\gabri\Python\Anaconda3\envs\benchmark_gnn2\lib\ctypes\__init__.py", line 364, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: [WinError 126] Impossibile trovare il modulo specificato
[...]

PR #1507, opened 2020-05-09
[Bug?] Fixing compatibility of pickles across DGL versions

BarclayII:
## Description
Issue #1494 reported that DGL 0.4.3+ is unable to read pickles dumped in 0.4.2-.  Decided to fix this since it would be quite an obstacle for people reproducing other's works on preprocessed datasets.

## Checklist
- [ ] All changes have test coverage (How should I do unit tests?  I'm right now testing on a dataset downloaded from [Benchmarking GNNs repo](https://github.com/graphdeeplearning/benchmarking-gnns/))
- [x] Code is well-documented

### DIFF
--- a/python/dgl/graph.py
+++ b/python/dgl/graph.py
@@ -1050,6 +1050,15 @@ class DGLGraph(DGLBaseGraph):
         # set parent if the graph is a subgraph.
         self._parent = parent
 
+    def __setstate__(self, state):
+        # Compatibility with pickles from DGL 0.4.2-
+        if '_batch_num_nodes' not in state:
+            state = state.copy()
+            state.setdefault('_batch_num_nodes', None)
+            state.setdefault('_batch_num_edges', None)
+            state.setdefault('_parent', None)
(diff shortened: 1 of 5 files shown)


BarclayII:
I have tested through all datasets in the Benchmarking GNNs repo and this patch can make DGL load all those datasets.

python/dgl/graph.py
```diff
@@ -1050,6 +1050,15 @@ def __init__(self,
         # set parent if the graph is a subgraph.
         self._parent = parent
 
+    def __setstate__(self, state):
+        # Compatibility with pickles from DGL 0.4.2-
+        if '_batch_num_nodes' not in state:
```
jermainewang, line 1055: My general comment is using `dgl.__version__` for if condition. Otherwise, LGTM.

BarclayII: Decided not to do that since using `dgl.__version__` is not straightforward for nightly builds.

PR #1522, opened 2020-05-14
[Feature] Range partition

zheng-da:
## Description
To support range partitioning, this PR adds a few functions:
* new CAPI to shuffle nodes in a graph.
* graph partitioning calls the new CAPI to shuffle nodes and edges in a graph to support range partitioning when splitting a graph into multiple partitions.
* update the data loading.
* add range partition book.

## Checklist
- [ ] Code is well-documented


python/dgl/transform.py
```diff
@@ -563,25 +594,54 @@ def partition_graph_with_halo(g, node_part, num_hops):
...
         subg = g._create_subgraph(subg, subg.induced_nodes, subg.induced_edges)
         inner_node = F.zerocopy_from_dlpack(inner_node.to_dlpack())
         subg.ndata['inner_node'] = inner_node
-        inner_edge = F.zerocopy_from_dlpack(inner_edge.to_dlpack())
+        subg.ndata['part_id'] = F.gather_row(node_part, subg.parent_nid)
+        if reshuffle:
+            subg.ndata['orig_id'] = F.gather_row(g.ndata['orig_id'], subg.ndata[NID])
+            subg.edata['orig_id'] = F.gather_row(g.edata['orig_id'], subg.edata[EID])
+
+        if extra_cached_hops >= 1:
```
classicsong, line 635: Do we need to use a for loop here?

python/dgl/transform.py
```diff
@@ -613,14 +673,16 @@ def metis_partition_assignment(g, k):
         node_part = utils.toindex(node_part)
         return node_part.tousertensor()
 
-def metis_partition(g, k, extra_cached_hops=0):
+def metis_partition(g, k, extra_cached_hops=0, reshuffle=False):
```
classicsong, line 676: Can we move the reshuffle(remapping) task to a standalone func? So we do not need to implement it in each partition algorithm.

src/array/cpu/spmat_op_impl_coo.cc
```diff
@@ -427,6 +427,55 @@ template COOMatrix COOSliceMatrix<kDLCPU, int32_t>(
...
+  IdType *out_row = static_cast<IdType*>(out_row_arr->data);
+  IdType *out_col = static_cast<IdType*>(out_col_arr->data);
+  IdType *out_data = COOHasData(coo) ? static_cast<IdType*>(out_data_arr->data) : nullptr;
+
+#pragma omp parallel for
+  for (int64_t i = 0; i < nnz; i++) {
+    out_row[i] = new_row_ids[in_rows[i]];
+    out_col[i] = new_col_ids[in_cols[i]];
+    if (out_data)
+      out_data[i] = in_data[i];
```
classicsong, line 469: It seems identical, Do we really need to copy it?

python/dgl/backend/backend.py
```diff
@@ -982,6 +982,9 @@ def logical_not(input):
     """
     pass
 
+def logical_and(input1, input2):
```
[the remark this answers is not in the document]
zheng-da: we can. when data is stored in the file, it's stored in our own format.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -232,8 +231,212 @@ def get_partition(self, partid):
...
+    num_parts : int
+        number of total partitions
+    node_map : tensor
+        global node id mapping to partition id
+    edge_map : tensor
+        global edge id mapping to partition id
+    """
+    def __init__(self, part_id, num_parts, node_map, edge_map):
+        assert part_id >= 0, 'part_id cannot be a negative number.'
+        assert num_parts > 0, 'num_parts must be greater than zero.'
```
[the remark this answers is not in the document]
zheng-da: it should work for num_parts=1. this is useful for debugging.

python/dgl/distributed/dist_graph.py
```diff
@@ -41,20 +41,20 @@ def _copy_graph_to_shared_mem(g, graph_name):
...
                                                                local_edge_path)
     new_g.ndata[NID] = _move_data_to_shared_mem_array(g.ndata[NID],
                                                       _get_ndata_path(graph_name, NID))
     new_g.edata[EID] = _move_data_to_shared_mem_array(g.edata[EID],
                                                       _get_edata_path(graph_name, EID))
     return new_g
 
-FIELD_DICT = {'local_node': F.int64,
-              'local_edge': F.int64,
+FIELD_DICT = {'inner_node': F.int64,
```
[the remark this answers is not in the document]
zheng-da: these are node data. we should consider int32 for node/edge ids. i suppose we switch between them if we can check the underlying graph structure uses int32 or int64.

python/dgl/distributed/partition.py
```diff
@@ -134,24 +135,25 @@ def load_partition(conf_file, part_id):
     assert 'num_edges' in part_metadata, "cannot get the number of edges of the global graph."
     assert 'node_map' in part_metadata, "cannot get the node map."
     assert 'edge_map' in part_metadata, "cannot get the edge map."
-    node_map = np.load(part_metadata['node_map'])
-    edge_map = np.load(part_metadata['edge_map'])
-    meta = (part_metadata['num_nodes'], part_metadata['num_edges'], node_map, edge_map, num_parts)
+
+    reshuffle = isinstance(part_metadata['node_map'], list)
```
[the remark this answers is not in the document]
zheng-da: If this is a range partitioning, node map actually is stored as a list, whose elements indicate the boundary of range partitioning.

python/dgl/transform.py
zheng-da: we only need to test if the nodes and edges are inside a partition or not. we don't need to test which hop these nodes/edges are.

python/dgl/transform.py
zheng-da: reshuffling happens in `partition_graph_with_halo`. it works for different partitioning algorithms.

src/array/cpu/spmat_op_impl_coo.cc
zheng-da: probably not.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -9,12 +9,12 @@
...
 def _move_metadata_to_shared_mam(graph_name, num_nodes, num_edges, part_id,
-                                 num_partitions, node_map, edge_map):
-    ''' Move all metadata to the shared memory.
+                                 num_partitions, node_map, edge_map, is_range_part):
+    ''' Move all metadata of the partition book to the shared memory.
 
     We need these metadata to construct graph partition book.
     '''
-    meta = _to_shared_mem(F.tensor([num_nodes, num_edges,
+    meta = _to_shared_mem(F.tensor([int(is_range_part), num_nodes, num_edges,
```
classicsong, line 17: is_range_part is a bool? We need to convert it to int tensor with 0 or 1?
Both torch and tensorflow support bool tensor

python/dgl/distributed/graph_partition_book.py
```diff
@@ -27,25 +27,27 @@ def _get_shared_mem_metadata(graph_name):
     The metadata includes the number of nodes and the number of edges. In the future,
     we can add more information, especially for heterograph.
     '''
-    shape = (4,)
+    shape = (5,)
```
classicsong, line 30: What does this mean? why 5?

python/dgl/distributed/graph_partition_book.py
```diff
@@ -150,8 +154,8 @@ def shared_memory(self, graph_name):
             The graph name
         """
         self._meta, self._nid2partid, self._eid2partid = _move_metadata_to_shared_mam(
```
classicsong, line 156: Is this _move_metadata_to_shared_mem?

python/dgl/distributed/partition.py
```diff
@@ -134,24 +135,27 @@ def load_partition(conf_file, part_id):
...
-    node_map = np.load(part_metadata['node_map'], allow_pickle=True)
-    edge_map = np.load(part_metadata['edge_map'], allow_pickle=True)
-    meta = (part_metadata['num_nodes'], part_metadata['num_edges'], node_map, edge_map, num_parts)
+
+    # If this is a range partitioning, node_map actually stores a list, whose elements
+    # indicate the boundary of range partitioning. Otherwise, node_map stores a filename
+    # that contains node map in a NumPy array.
+    is_range_part = isinstance(part_metadata['node_map'], list)
+    node_map = part_metadata['node_map'] if is_range_part else np.load(part_metadata['node_map'])
+    edge_map = part_metadata['edge_map'] if is_range_part else np.load(part_metadata['edge_map'])
```
classicsong, line 144: part_metadata['node_map'] and part_metadata['edge_map'] always the same type?

tests/compute/test_transform.py
```diff
@@ -244,33 +244,96 @@ def test_partition_with_halo():
...
+        assert np.all(edge_cnts == 1)
+
         assert num_inner_nodes == g.number_of_nodes()
         print(g.number_of_edges() - num_inner_edges)
 
+@unittest.skipIf(F._default_context_str == 'gpu', reason="It doesn't support GPU")
+def test_reorder_nodes():
+    g = dgl.DGLGraph(create_large_graph_index(1000), readonly=True)
+    new_nids = np.random.permutation(g.number_of_nodes())
+    new_g = dgl.transform.reorder_nodes(g, new_nids)
```
classicsong, line 314: Can this test both COOReorder and CSRReorder?

python/dgl/distributed/graph_partition_book.py
zheng-da: yes, but the rest of the elements are integer. we have to convert it to integer as well.

python/dgl/distributed/graph_partition_book.py
aksnzhy: I think 5 is the maximal bound.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -335,6 +335,261 @@ def get_edge_size(self):
         """
         return self._edge_size
 
+
+class RangePartitionBook:
+    """GraphPartitionBook is used to store parition information.
```
aksnzhy, line 340: GraphPartitionBook -> RangePartitionBook

python/dgl/distributed/graph_partition_book.py
```diff
@@ -335,6 +335,261 @@ def get_edge_size(self):
...
+
+    Parameters
+    ----------
+    part_id : int
+        partition id of current GraphPartitionBook
```
aksnzhy, line 345: GraphPartitionBook -> RangePartitionBook

python/dgl/distributed/graph_partition_book.py
```diff
@@ -335,6 +335,261 @@ def get_edge_size(self):
...
+        """
+        return int(self._edge_map[-1])
+
+    def metadata(self):
+        """Return the partition meta data.
+
+        The meta data includes:
+
+        * The machine ID.
+        * The machine IP address.
```
aksnzhy, line 414: I think we removed the IP address already. This is my fault.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -335,6 +335,261 @@ def get_edge_size(self):
...
+        nids : tensor
+            global node IDs
+
+        Returns
+        -------
+        tensor
+            partition IDs
+        """
+        nids = utils.toindex(nids)
+        ret = np.searchsorted(self._node_map, nids.tonumpy(), side='right')
```
aksnzhy, line 446: Will searchsorted() bring much more overhead than a slice? This API would be invoked very frequent.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -335,6 +335,261 @@ def get_edge_size(self):
...
+        tensor
+            partition IDs
+        """
+        eids = utils.toindex(eids)
+        ret = np.searchsorted(self._edge_map, eids.tonumpy(), side='right')
+        ret = utils.toindex(ret)
+        return ret.tousertensor()
+
+
+    def partid2nids(self, partid):
```
aksnzhy, line 470: Actually, it seems that I never used this API

python/dgl/distributed/partition.py
zheng-da: i think so. i have a check after this line.

tests/compute/test_transform.py
zheng-da: it only works for DGLGraph right. i guess we cannot control the format.

zheng-da: i tested internally (by only enabling COOReorder). it works fine. In the future, we need to move all operations to DGLHeteroGraph.

PR #1527, opened 2020-05-15
[Performance] Reduces OpenMP atomic additions with zero inputs

pawelpiotrowicz:
This PR speeds up the computation of kernels with `fn.sum()` as reduce functions with a simple trick that skips the atomic additions if the addend is 0.  This trick is particularly important if the message has a lot of zero elements, which can be the case if working with ReLU or Dropout.

Based on my (Pawel's) observation, the GCMC code can be 50% faster on MovieLens-1m dataset.  The command I tested is as follows:

```bash
export OMP_NUM_THREADS=24
export DGLBACKEND=pytorch
time python examples/pytorch/gcmc/train_sampling.py --data_name=ml-1m --use_one_hot_fea --gcn_agg_accum=stack --gpu -1 --train_max_epoch=15
```


yzh119:
hi @pawelpiotrowicz , we understand what optimization you made, but could you please elaborate more on "training faster up to 50%" (which model on what dataset?), and in what case `grad_e` would be zero?

Thanks very much for your contribution.

pawelpiotrowicz:
Hi @yzh119 below is my script 

```
export OMP_NUM_THREADS=24
export DGLBACKEND=pytorch
time python examples/pytorch/gcmc/train_sampling.py --data_name=ml-1m --use_one_hot_fea --gcn_agg_accum=stack --gpu -1 --train_max_epoch=15
```

zheng-da:
@pawelpiotrowicz do you know which one has more 0 values? in the forward computation or the backward computation?

zheng-da:
Also, please write more description in this PR.

BarclayII:
The speed up makes a lot of sense since GCMC implementation has a Dropout probability of 0.7.  This fix would skip a lot of unnecessary atomic additions, and should often (but not always) work with any model that has Dropout or ReLU on node features or message computations.

Updated the PR description accordingly.

I love this.  Please merge ASAP.

BarclayII:
@pawelpiotrowicz I have only one question: does this trick also apply in `BackwardBinaryReduceBcast::ApplyEdge()`?
https://github.com/dmlc/dgl/blob/d6993471e08bf522253e3e7449c5fcfa1d6d418f/src/kernel/cpu/backward_binary_reduce_impl.h#L127

pawelpiotrowicz:
@BarclayII  Yes is applicable and logic says that this is a good idea but from my test the final output is 2% worse than before. Maybe there is an configuration when will be better. Anyway the setup for me is the most optimal. Thank you for update description.

BarclayII:
> @BarclayII Yes is applicable and logic says that this is a good idea but from my test the epoch time is 2% worse than before. Maybe there is an configuration when will be better. Anyway the setup for me is the most optimal. Thank you for update description.

The 2% loss looks interesting.  The only places that invokes the DGL kernels are one in https://github.com/dmlc/dgl/blob/0b902d032ab92f9c4fd6d3d8b283129851fb43d2/examples/pytorch/gcmc/model.py#L90-L91
and another in
https://github.com/dmlc/dgl/blob/0b902d032ab92f9c4fd6d3d8b283129851fb43d2/examples/pytorch/gcmc/model.py#L340

and the `BackwardBinaryReduceBcast` should not impact either of them.  Anyway, I left a comment saying that this is worth further investigation and I tend to leave the code itself as-is (and I put your name there - I hope it's OK).

pengzhao-intel:
Good job @pawelpiotrowicz !
Thanks to review @zheng-da @yzh119

PR #1529, opened 2020-05-15
[don't merge] Rpc

zheng-da:
## Checklist
- [ ] Code is well-documented


python/dgl/distributed/rpc.py
```diff
@@ -0,0 +1,487 @@
...
+        return _CAPI_DGLRPCMessageGetClientId(self)
+
+    @property
+    def server_id(self):
+        """Get server ID."""
+        return _CAPI_DGLRPCMessageGetServerId(self)
+
+    @property
+    def xxx(self):
+        return 1
```
zheng-da, line 296: what is this?

python/dgl/distributed/rpc.py
```diff
@@ -0,0 +1,487 @@
...
+    if res_cls is None:
+        raise DGLError('Got response message from service ID {}, '
+                       'but no response class is registered.'.format(msg.service_id))
+    res = deserialize_from_payload(res_cls, msg.data, msg.tensors)
+    if res.client_id != get_rank():
+        raise DGLError('Got reponse of request sent by client {}, '
+                       'different from my rank {}!'.format(res.client_id, get_rank()))
+    return res
+
+def remote_call(target_and_requests, timeout=0):
```
zheng-da, line 374: if we want to support async pull, we can still support it with the current network API, right?

src/rpc/rpc.h
```diff
@@ -0,0 +1,130 @@
...
+
+  /*!
+   * \brief Message sequence number.
+   */
+  int64_t msg_seq = 0;
+
+  /*!
+   * \brief Sender communicator.
+   */
+  std::shared_ptr<network::Sender> sender;
```
zheng-da, line 38: how does it work when it connects with many servers? it seems one process has one RPCContext.

src/rpc/server_state.h
```diff
@@ -0,0 +1,61 @@
...
+ * Each machine can have multiple server and client processes, but only one
+ * server is the *master* server while all the others are backup servers. All
+ * clients and backup servers share the state of the master server via shared
+ * memory, which means the ServerState class must be serializable and large
+ * bulk data (e.g., node/edge features) must be stored in NDArray to leverage
+ * shared memory.
+ */
+struct ServerState : public runtime::Object {
+  /*! \brief Key value store for NDArray data */
+  std::unordered_map<std::string, runtime::NDArray> kv_store;
```
zheng-da, line 42: I really think we shouldn't manage the kvstore data independently. we don't need to tie everything together.

python/dgl/distributed/rpc.py
jermainewang: Testing code. Should be deleted.

python/dgl/distributed/rpc.py
jermainewang: It will probably be another API called `remote_call_async` and it will return a list of future objects to get the responses asynchronously.

src/rpc/rpc.h
jermainewang: `network::Sender` and `network::Receiver` support multiple connections at the same time. @aksnzhy pls corrects me if i'm wrong.

aksnzhy: Yes. Both sender and receiver can support multiple connections.

PR #1544, opened 2020-05-19
[RPC] Implementation of RPC infra

aksnzhy:
## Description
Implementation of RPC infra. This pr does not handle ServerSate.

## Checklist
- [x] Code is well-documented


python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
+import time
+import socket
+
+if os.name != 'nt':
+    import fcntl
+    import struct
+
+def local_ip4_addr_list():
+    """Return a set of IPv4 address
+    """
```
jermainewang, line 17: The function is not supported in WIN since `struct` is not imported. Please raise more informative errors.

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
+
+def get_local_machine_id(server_namebook):
+    """Given server_namebook, find local machine ID
+
+    Parameters
+    ----------
+    server_namebook: dict
+        IP address namebook of server nodes, where key is the server's ID
+        (start from 0) and value is the server's machine_id, IP address,
+        port, and group_count, e.g.,
```
jermainewang, line 37: What is group_count?

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
+    Returns
+    -------
+    int
+        local machine ID
+    """
+    res = 0
+    for ID, data in server_namebook.items():
+        machine_id = data[0]
+        ip = data[1]
+        if ip in local_ip4_addr_list():
```
jermainewang, line 57: Take the result of  `local_ip4_addr_list()` out of the loop.

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
+        port, and group_count, e.g.,
+
+          {0:'[0, 172.31.40.143, 30050, 2],
+           1:'[0, 172.31.40.143, 30051, 2],
+           2:'[1, 172.31.36.140, 30050, 2],
+           3:'[1, 172.31.36.140, 30051, 2],
+           4:'[2, 172.31.47.147, 30050, 2],
+           5:'[2, 172.31.47.147, 30051, 2],
+           6:'[3, 172.31.30.180, 30050, 2],
+           7:'[3, 172.31.30.180, 30051, 2]}
```
jermainewang, line 46: What's the stype of the IP addr? string?

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
+        ip_addr = s.getsockname()[0]
+    except:
+        ip_addr = '127.0.0.1'
+    finally:
+        s.close()
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.bind(("", 0))
+    s.listen(1)
+    port = s.getsockname()[1]
+    s.close()
```
jermainewang, line 83: If I have two local processes and they both call this function to get a valid ports, will they collide?

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
+
+    return ip_addr + ':' + str(port)
+
+def connect_to_server(ip_config, queue_size=20*1024*1024*1024, net_type='socket'):
```
jermainewang, line 87: Please put all constants under `python/dgl/distributed/constants.py` and add necessary docstring.

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
     Parameters
     ----------
-    server_namebook : ...
-        Server names.
+    ip_config : str
+        Path of server IP configuration file.
+    queue_size : int
+        Size (bytes) of client queue buffer (~20 GB on default).
+        Note that the 20 GB is just an upper-bound and DGL uses zero-copy and
+        it will not allocate 20GB memory at once.
```
jermainewang, line 97: Then maybe call it `max_queue_size`?

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
+    net_type : str
+        networking type, e.g., 'socket' (on default) or 'mpi' (do not support yet).
```
jermainewang, line 99: `Networking type. Current options are: 'socket'.`

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
 
     Raises
     ------
     ConnectionError : If anything wrong with the connection.
     """
-    pass
+    assert queue_size > 0, 'queue_size (%d) cannot be a negative number.' % queue_size
+    assert net_type in ('socket', 'mpi'), \
+    'net_type (%s) can only be \'socket\' or \'mpi\'.' % net_type
```
jermainewang, line 107: Remove `mpi` since it is not supported at the moment.

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
+    for server_id, addr in server_namebook.items():
+        server_ip = addr[1]
+        server_port = addr[2]
+        rpc.add_receiver_addr(server_ip, server_port, server_id)
+    rpc.sender_connect()
+    # Get local usable IP address and port
+    ip_addr = get_local_usable_addr()
+    client_ip, client_port = ip_addr.split(':')
+    # Register client on server
+    # a temp ID because we don't assign client ID yet
```
jermainewang, line 138: `... we haven't assigned ...`

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
     ------
     ConnectionError : If anything wrong with the connection.
     """
-    pass
+    assert queue_size > 0, 'queue_size (%d) cannot be a negative number.' % queue_size
+    assert net_type in ('socket', 'mpi'), \
+    'net_type (%s) can only be \'socket\' or \'mpi\'.' % net_type
+    # Register some basic service
+    rpc.register_service(rpc.CLIENT_REGISTER,
+                         rpc.ClientRegisterReuqest,
```
jermainewang, line 110: `ClientRegisterReuqest` -> `ClientRegisterRequest`

python/dgl/distributed/rpc.py
```diff
@@ -10,12 +10,159 @@
...
+
+    Parameters
+    ----------
+    filename : str
+        Path of IP configuration file.
+
+    Returns
+    -------
+    dict
+        server namebook. The format is: {[server_id]:[machine_id, ip, port, group_count]}
```
jermainewang, line 47: The key is .... The value is ...

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
+    queue_size : int
+        Size (bytes) of client queue buffer (~20 GB on default).
+        Note that the 20 GB is just an upper-bound and DGL uses zero-copy and
+        it will not allocate 20GB memory at once.
+    net_type : str
+        networking type, e.g., 'socket' (on default) or 'mpi' (do not support yet).
 
     Raises
     ------
     ConnectionError : If anything wrong with the connection.
```
jermainewang, line 103: Have you properly handled exceptions?

python/dgl/distributed/rpc.py
```diff
@@ -10,12 +10,159 @@
...
+    except:
+        print("Error: data format on each line should be: [ip] [base_port] [server_count]")
+    return server_namebook
+
+def create_sender(msg_queue_size, net_type):
+    """Create rpc sender of this process.
+
+    Parameters
+    ----------
+    msg_queue_size : int
```
jermainewang, line 82: `max_queue_size`

python/dgl/distributed/rpc.py
```diff
@@ -10,12 +10,159 @@
...
+        Size (bytes) of network queue buffer.
+    net_type : str
+        networking type, e.g., 'socket' or 'mpi' (do not support yet).
```
jermainewang, line 85: `Networking type. Current options are: 'socket'.`

python/dgl/distributed/rpc.py
```diff
@@ -10,12 +10,159 @@
...
+
+def create_receiver(msg_queue_size, net_type):
+    """Create rpc receiver of this process.
+
+    Parameters
+    ----------
+    msg_queue_size : int
+        Size (bytes) of network queue buffer.
+    net_type : str
+        networking type, e.g., 'socket' or 'mpi' (do not support yet).
```
jermainewang, line 97: same comments as before

python/dgl/distributed/rpc.py
```diff
@@ -10,12 +10,159 @@
...
+    """Get rpc sender of this process.
+
+    Returns
+    -------
+    c handler
+        sender handler
+    """
+    return _CAPI_DGLRPCGetSender()
+
+def get_receiver():
```
jermainewang, line 111: Do we need `get_sender` and `get_receiver`? They can be retrieved directly in C++ right?

python/dgl/distributed/rpc.py
```diff
@@ -143,8 +299,38 @@ def service_id(self):
...
+
+    @client_id.setter
+    def client_id(self, client_id):
+        """Set client ID"""
+        self._client_id = client_id
+
+    @msg_seq.setter
+    def msg_seq(self, msg_seq):
+        """Set msg_seq"""
+        self._msg_seq = msg_seq
```
jermainewang, line 330: Why adding these properties?

python/dgl/distributed/rpc.py
```diff
@@ -338,6 +550,73 @@ def send_request(target, request):
...
+    Returns
+    -------
+    req : request
+        One request received from the target, or None if it times out.
+
+    Raises
+    ------
+    ConnectionError if there is any problem with the connection.
+    """
+    # TODO(chao): handle timeout
```
jermainewang, line 575: Have you handled it?

python/dgl/distributed/rpc.py
```diff
@@ -458,8 +742,7 @@ def send_rpc_message(msg):
     ------
     ConnectionError if there is any problem with the connection.
     """
-    status = _CAPI_DGLRPCSendRPCMessage(msg)
-    # TODO: handle status flag
+    _CAPI_DGLRPCSendRPCMessage(msg)
```
jermainewang, line 745: Status flag?

python/dgl/distributed/rpc.py
```diff
@@ -482,8 +765,83 @@ def recv_rpc_message(timeout=0):
     ConnectionError if there is any problem with the connection.
     """
     msg = _CAPI_DGLRPCCreateEmptyRPCMessage()
-    status = _CAPI_DGLRPCRecvRPCMessage(timeout, msg)
-    # TODO: handle status flag
+    _CAPI_DGLRPCRecvRPCMessage(timeout, msg)
```
jermainewang, line 768: Status flag?

python/dgl/distributed/server.py
```diff
@@ -2,19 +2,85 @@
...
-def start_server(clients):
-    """Start server.
+import time
 
-    This is a blocking function -- it returns only when the server receives
-    shutdown command from clients.
+from . import rpc
+from .server_state import get_server_state
+
+QUEUE_SIZE = 20*1024*1024*1024
```
jermainewang, line 10: Put all constants in one file.

python/dgl/distributed/server.py
```diff
@@ -2,19 +2,85 @@
...
     ----------
-    clients : ...
-        Client information
+    server_id : int
+        Current server ID (starts from 0).
+    ip_config : str
+        Path of IP configuration file.
+    num_clients : int
+        Total number of clients that will be connected to the server.
+        Note that, we do not support dynamic connection for now.
```
jermainewang, line 25: define what is dynamic connection

python/dgl/distributed/server.py
```diff
@@ -2,19 +2,85 @@
...
+    queue_size : int
+        Size (bytes) of server queue buffer (~20 GB on default).
+        Note that the 20 GB is just an upper-bound because DGL uses zero-copy and
+        it will not allocate 20GB memory at once.
+    net_type : str
+        networking type, e.g., 'socket' (on default) or 'mpi' (do not support yet).
```
jermainewang, line 31: same comment for the `queue_size` and `net_type` as before

src/rpc/rpc.h
```diff
@@ -97,6 +108,38 @@ enum RPCStatus {
...
+
+/*!
+ * \brief Deserialize one meta of RPCMessage from data buffer
+ *
+ * \param msg RPC message to deserilaize
+ * \param buffer buffer holding the serialized data
+ * \param size buffer size
+ * \return has_tensor
+ */
+bool DeserializeRPCMeta(RPCMessage* msg, char* buffer, int64_t size);
```
jermainewang, line 141: You should use the `Write` and `Read` interfaces.

python/dgl/distributed/client.py
zheng-da: this queue size is very large. is it the maximal queue size?

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
 
 """Functions used by client."""
```
zheng-da, line 3: i think the file should be called 'rpc_client.py'.

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
+    # Register some basic service
+    rpc.register_service(rpc.CLIENT_REGISTER,
+                         rpc.ClientRegisterReuqest,
+                         rpc.ClientRegisterResponse)
+    rpc.register_service(rpc.SHUT_DOWN_SERVER,
+                         rpc.ShutDownRequest,
+                         None)
+    server_namebook = rpc.read_ip_config(ip_config)
+    num_servers = len(server_namebook)
+    rpc.set_num_server(num_servers)
```
zheng-da, line 117: so the number of servers doesn't include the backup servers?

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
+    group_count = []
+    max_machine_id = 0
+    for server_info in server_namebook.values():
+        group_count.append(server_info[3])
+        if server_info[0] > max_machine_id:
+            max_machine_id = server_info[0]
```
zheng-da, line 123: what happens if the server ids given by a user aren't contiguous? should we throw an error? or it can be handled by the system?

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
+    num_machines = max_machine_id+1
+    machine_id = get_local_machine_id(server_namebook)
+    rpc.create_sender(queue_size, net_type)
+    rpc.create_receiver(queue_size, net_type)
+    # Get connected with all server nodes
+    for server_id, addr in server_namebook.items():
+        server_ip = addr[1]
+        server_port = addr[2]
+        rpc.add_receiver_addr(server_ip, server_port, server_id)
```
zheng-da, line 132: will it be better to call it `add_server_addr`? `add_receiver_addr` is confusing because rpc can be used for both sending and receiving?

python/dgl/distributed/client.py
```diff
@@ -2,33 +2,155 @@
...
+    rpc.sender_connect()
+    # Get local usable IP address and port
+    ip_addr = get_local_usable_addr()
+    client_ip, client_port = ip_addr.split(':')
+    # Register client on server
+    # a temp ID because we don't assign client ID yet
+    rpc.set_rank(0)
```
zheng-da, line 139: why is rank always 0?

python/dgl/distributed/rpc.py
```diff
@@ -10,12 +10,159 @@
...
+    """Read network configuration information of server from file.
+
+    The format of configuration file should be:
+
+        [ip] [base_port] [server_count]
+
+        172.31.40.143 30050 2
+        172.31.36.140 30050 2
+        172.31.47.147 30050 2
+        172.31.30.180 30050 2
```
zheng-da, line 32: network config of clients and the servers are different?

python/dgl/distributed/rpc.py
```diff
@@ -10,12 +10,159 @@
...
+    """
+    _CAPI_DGLRPCCreateSender(int(msg_queue_size), net_type)
+
+def create_receiver(msg_queue_size, net_type):
+    """Create rpc receiver of this process.
+
+    Parameters
+    ----------
+    msg_queue_size : int
+        Size (bytes) of network queue buffer.
```
zheng-da, line 95: max_queue_size

python/dgl/distributed/rpc.py
```diff
@@ -78,7 +235,6 @@ def register_service(service_id, req_cls, res_cls=None):
     if res_cls is not None:
         RESPONSE_CLASS_TO_SERVICE_ID[res_cls] = service_id
     SERVICE_ID_TO_PROPERTY[service_id] = (req_cls, res_cls)
-    # TODO: C registry
```
zheng-da, line 81: we'll need to register C callback later, right?

python/dgl/distributed/rpc.py
```diff
@@ -338,6 +550,73 @@ def send_request(target, request):
...
+    if msg.server_id != get_rank():
+        raise DGLError('Got request sent to server {}, '
+                       'different from my rank {}!'.format(msg.server_id, get_rank()))
+    return req
+
+def send_response(target, msg_seq, response):
+    """Send one response to the target client.
+
+    Serilaize the given response object to an :class`RPCMessage` and send it
+    reached the target or even have left the sender process. However,
```
zheng-da, line 596: i cannot understand the sentence.

python/dgl/distributed/rpc.py
```diff
@@ -338,6 +550,73 @@ def send_request(target, request):
     msg = RPCMessage(service_id, msg_seq, client_id, server_id, data, tensors)
     send_rpc_message(msg)
 
+def recv_request(timeout=0):
```
zheng-da, line 553: how is this different from `recv_response`?

python/dgl/distributed/rpc.py
```diff
@@ -338,6 +550,73 @@ def send_request(target, request):
...
+
+    Raises
+    ------
+    ConnectionError if there is any problem with the connection.
+    """
+    # TODO(chao): handle timeout
+    msg = recv_rpc_message(timeout)
+    if msg is None:
+        return None
+    req_cls, _ = SERVICE_ID_TO_PROPERTY[msg.service_id]
```
zheng-da, line 579: why is this different from the one in `recv_response`?

python/dgl/distributed/rpc.py
```diff
@@ -422,16 +706,16 @@ def remote_call(target_and_requests, timeout=0):
         # recv response
         msg = recv_rpc_message(timeout)
         num_res -= 1
-        if msg is None:
-            continue
```
zheng-da, line 426: this check is invalid?

python/dgl/distributed/rpc.py
zheng-da: what is the status flag?

python/dgl/distributed/rpc.py
```diff
@@ -482,8 +765,83 @@ def recv_rpc_message(timeout=0):
...
+class ClientRegisterReuqest(Request):
+    """This request will send client's ip to server.
+
+    Parameters
+    ----------
+    ip_addr : str
+        client's IP address
+    """
+    def __init__(self, ip_addr):
+        self.ip_addr = ip_addr
```
zheng-da, line 791: it doesn't require a field to indicate the type of requests?

python/dgl/distributed/server.py
```diff
@@ -2,19 +2,85 @@
...
+import time
 
-    This is a blocking function -- it returns only when the server receives
-    shutdown command from clients.
+from . import rpc
+from .server_state import get_server_state
+
+QUEUE_SIZE = 20*1024*1024*1024
+
+def start_server(server_id, ip_config, num_clients, queue_size=QUEUE_SIZE, net_type='socket'):
```
zheng-da, line 12: @jermainewang last time we discussed, you convinced me that we don't need to know the number of clients?

src/rpc/rpc.cc
```diff
@@ -7,25 +7,258 @@
...
+char* SerializeRPCMeta(const RPCMessage& msg, int64_t* size) {
+  int64_t total_size = 0;
+  total_size += sizeof(msg.service_id);
+  total_size += sizeof(msg.msg_seq);
+  total_size += sizeof(msg.client_id);
+  total_size += sizeof(msg.server_id);
+  total_size += sizeof(int64_t);  // data_size
+  total_size += msg.data.size();
+  total_size += sizeof(int8_t);   // has_tensor
+  char* buffer = new char[total_size];
```
zheng-da, line 28: Why not define a structure for this? it'll be much more readable.

src/rpc/rpc.cc
```diff
@@ -7,25 +7,258 @@
...
+      ndarray_meta_msg, msg.server_id), ADD_SUCCESS);
+    // send ndarray count
+    char* ndarray_count = new char[sizeof(int32_t)];
+    *(reinterpret_cast<int32_t*>(ndarray_count)) = msg.tensors.size();
+    network::Message ndarray_count_msg;
+    ndarray_count_msg.data = ndarray_count;
+    ndarray_count_msg.size = sizeof(int32_t);
+    ndarray_count_msg.deallocator = network::DefaultMessageDeleter;
+    CHECK_EQ(RPCContext::ThreadLocal()->sender->Send(
+      ndarray_count_msg, msg.server_id), ADD_SUCCESS);
```
zheng-da, line 127: can we put all metadata in one message?

src/rpc/rpc.cc
```diff
@@ -7,25 +7,258 @@
...
+    // send real ndarray data
+    for (auto ptr : zc_write_strm.buffer_list()) {
+      network::Message ndarray_data_msg;
+      ndarray_data_msg.data = reinterpret_cast<char*>(ptr.data);
+      ndarray_data_msg.size = ptr.size;
+      NDArray tensor = ptr.tensor;
+      ndarray_data_msg.deallocator = [tensor](network::Message*) {};
+      CHECK_EQ(RPCContext::ThreadLocal()->sender->Send(
+        ndarray_data_msg, msg.server_id), ADD_SUCCESS);
```
zheng-da, line 136: so the message doesn't own the memory of the tensors. can this cause memory issues? how do we guarantee that the memory isn't free'd before the data is actually sent out.

src/rpc/rpc.cc
```diff
@@ -7,25 +7,258 @@
...
+    std::vector<void* > buffer_list(ndarray_count);
+    for (int i = 0; i < ndarray_count; ++i) {
+      network::Message ndarray_data_msg;
+      CHECK_EQ(RPCContext::ThreadLocal()->receiver->RecvFrom(
+        &ndarray_data_msg, send_id), REMOVE_SUCCESS);
+      buffer_list[i] = ndarray_data_msg.data;
+    }
+    StringStreamWithBuffer zc_read_strm(&zerocopy_blob, buffer_list);
+    msg->tensors.resize(ndarray_count);
+    static_cast<dmlc::Stream *>(&zc_read_strm)->Read(&msg->tensors);
```
zheng-da, line 176: who owns the memory of the tensors?

python/dgl/distributed/client.py
aksnzhy: Ok, use `rpc_client` and `rpc_server`

python/dgl/distributed/client.py
aksnzhy: group_count means how many servers (main_server + bakcup_server) in total inside a machine.

python/dgl/distributed/client.py
aksnzhy: It's a string. Let's add '' for that

python/dgl/distributed/client.py
aksnzhy: The s.bind() will make sure that only one process can bind successfully. I tested this function on KE for more than 40 processes concurrently and it works well.

python/dgl/distributed/client.py
aksnzhy: It is a maximal size. I will change the name for that.

python/dgl/distributed/client.py
aksnzhy: The exception will be catched by C++ code.

python/dgl/distributed/client.py
aksnzhy: The num_servers = main_server + backup_servers

python/dgl/distributed/client.py
aksnzhy: The server id is created by read_ip_config() API. User will not set any ID of server or client.

python/dgl/distributed/client.py
aksnzhy: Both server and client will have 'sender' and 'receiver'. It's not only used in server-side. The sender and receiver is just a very low-level API and you can just treat them as mpi send and recv.

python/dgl/distributed/client.py
aksnzhy: Because in the beginning, the client doesn't have the ID and it is assigned by the server. 0 is temp ID to connect with server for some initialization.

python/dgl/distributed/rpc.py
aksnzhy: Yes, I think we can remove these two API.

python/dgl/distributed/rpc.py
aksnzhy: Yes.

python/dgl/distributed/rpc.py
aksnzhy: They will use different class for serilaize/de-serialize

python/dgl/distributed/rpc.py
aksnzhy: For now, I haven't handled the timeout.

python/dgl/distributed/rpc.py
aksnzhy: req_cls and res_cls are different classes.

python/dgl/distributed/rpc.py
aksnzhy: These properties will be used in send_response.

python/dgl/distributed/rpc.py
aksnzhy: recv_rpc_message will also return a msg. If no message is coming, the recv() will just be blocked.

src/rpc/rpc.cc
aksnzhy: Yes, I will change it to dmlc serializer.

src/rpc/rpc.cc
aksnzhy: `ndarray_data_msg.deallocator = [tensor](network::Message*) {};` this code own the handler of tensor.

src/rpc/rpc.cc
aksnzhy: The `msg->tensors` will own all the memory. I think this code is done by JJ's `Read(&msg->tensors);`

python/dgl/distributed/rpc.py
zheng-da: from the code perspective, the difference is just the name?

python/dgl/distributed/rpc.py
zheng-da: what if it times out?

src/rpc/rpc.cc
zheng-da: i'm not sure about serializer. i think it's pretty simple
```C
struct struct_name {
  int64_t data_size;
  int8_t has_tensor;
  ...
};

struct_name *o = new struct_name();
o.data_size = ...;
o.has_tensor = ...;
...
```
or you can do something like this:
```C
char *mem = new char[...];
struct_name *o = static_cast<struct_name *>(mem);
...
```
it's more readable and the compiler can help check something.

src/rpc/rpc.cc
zheng-da: ok. i can see it now.

python/dgl/distributed/rpc.py
aksnzhy: For now, I didn't handle times out. The recv will always listen to the port until recv a msg for shutdown.

python/dgl/distributed/rpc.py
aksnzhy: Here are some logic difference. For example, the recv_response will not invoke set_client_id, set_server_id, and set_msg_seq functions.

src/rpc/rpc.h
aksnzhy: @zheng-da @jermainewang Please see the new code

src/rpc/rpc.cc
aksnzhy: Changed.

python/dgl/distributed/rpc.py
aksnzhy: The status has been handled by C++ code.

python/dgl/distributed/rpc_client.py
```diff
@@ -0,0 +1,169 @@
+# pylint: skip-file
```
VoVAllen, line 1: Remove this if it's complete

src/rpc/rpc.cc
```diff
@@ -7,25 +7,150 @@
...
+  // Copy the data for now, can be optimized later
+  memcpy(rpc_meta_buffer, zerocopy_blob.data(), zerocopy_blob.size());
+  int64_t rpc_meta_size = zerocopy_blob.size();
+  network::Message rpc_meta_msg;
+  rpc_meta_msg.data = rpc_meta_buffer;
+  rpc_meta_msg.size = rpc_meta_size;
+  rpc_meta_msg.deallocator = network::DefaultMessageDeleter;
+  CHECK_EQ(RPCContext::ThreadLocal()->sender->Send(
+    rpc_meta_msg, msg.server_id), ADD_SUCCESS);
+  // send ndarray count
```
jermainewang, line 33: Can you append it to the end of the meta data buffer?

src/rpc/rpc.cc
```diff
@@ -7,25 +7,150 @@
...
+    ndarray_data_msg.deallocator = [tensor](network::Message*) {};
+    CHECK_EQ(RPCContext::ThreadLocal()->sender->Send(
+      ndarray_data_msg, msg.server_id), ADD_SUCCESS);
+  }
   return kRPCSuccess;
 }
 
 RPCStatus RecvRPCMessage(RPCMessage* msg, int32_t timeout) {
-  // TODO(???): Implement
+  // ignore timeout now
```
jermainewang, line 56: Use CHECK and raise not supported error.

src/rpc/rpc.cc
```diff
@@ -7,25 +7,150 @@
...
+  network::Message rpc_meta_msg;
+  int send_id;
+  CHECK_EQ(RPCContext::ThreadLocal()->receiver->Recv(
+    &rpc_meta_msg, &send_id), REMOVE_SUCCESS);
+  // Copy the data for now, can be optimized later
```
jermainewang, line 61: `// TODO(chao): ...`  I guess this requires @VoVAllen 's modification on the stream class right?

tests/compute/test_rpc.py
```diff
@@ -73,3 +147,22 @@ def test_rpc_msg():
     assert len(rpcmsg.data) == len(data)
     assert len(rpcmsg.tensors) == 1
     assert F.array_equal(rpcmsg.tensors[0], req.z)
+
+@unittest.skipIf(os.name == 'nt', reason='Do not support windows yet')
+def test_rpc():
+    ip_config = open("ip_config.txt", "w")
+    ip_config.write('127.0.0.1 30050 1\n')
```
jermainewang, line 154: It will be great if we could test more combinations. You could create the task in the project tracker if you don't want to do so at the moment.

python/dgl/distributed/server.py
jermainewang: I guess the `num_clients` is needed during initialization when server#0 needs to assign an ID for each client. Our discussion last time was mainly about the barrier behavior and I remember the conclusion was that full barrier will not be required at the moment.

src/rpc/rpc.cc
VoVAllen: Yes

python/dgl/distributed/rpc.py
zheng-da: but in the future, you need to deal with this case, right? maybe you can keep it so you don't need to add it back?

python/dgl/distributed/rpc_client.py
aksnzhy: Will do that

PR #1569, opened 2020-05-29
[KVStore] Re-write kvstore using DGL RPC infrastructure

aksnzhy:
## Description

Re-write kvstore using DGL RPC infrastructure. Note: `fast-pull` is not added to this PR

This PR has been tested under real distributed environment.

## Checklist
- [ ] Code is well-documented


python/dgl/distributed/rpc_server.py
```diff
@@ -70,10 +70,15 @@ def start_server(server_id, ip_config, num_clients, \
         for client_id, _ in client_namebook.items():
             register_res = rpc.ClientRegisterResponse(client_id)
             rpc.send_response(client_id, register_res)
-    server_state = get_server_state()
+    server_state = ServerState(kv_server)
```
zheng-da, line 73: i don't think you can create server state here. what should `get_server_state` returns?

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1107 @@
...
+
+from . import rpc
+from .constants import get_type_str
+
+from .. import backend as F
+from .._ffi.ndarray import empty_shared_mem
+
+KVSTORE_PULL = 901231
+
+class PartitionPolicy(object):
```
zheng-da, line 16: this should be defined in graph_partition_book.py.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1107 @@
...
+import random
+import numpy as np
+
+from . import rpc
+from .constants import get_type_str
+
+from .. import backend as F
+from .._ffi.ndarray import empty_shared_mem
+
+KVSTORE_PULL = 901231
```
zheng-da, line 14: what is this?

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1107 @@
...
+
+class PartitionPolicy(object):
+    """Wrapper for GraphPartitionBook and RangePartitionBook. 
+
+    We can extend this class to support HeteroGraph in the future.
+
+    Parameters
+    ----------
+    policy_str : str
+        partition policy string, e.g., 'edge' or 'node'.
```
zheng-da, line 24: this isn't ideal. in the future, we need to support different node types and edge types.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1107 @@
...
+    Parameters
+    ----------
+    server_id : int
+        ID of current server
+    data_tensor : tensor
+        a tensor with the same row size of data ID.
+    """
+    def __init__(self, server_id, data_tensor):
+    	self.server_id = server_id
+        self.data_tensor = data_tensor
```
zheng-da, line 112: the indent is wrong.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1107 @@
...
+    Parameters
+    ----------
+    name : str
+        data name
+    id_tensor : tensor
+        a vector storing the data ID
+    """
+    def __init__(self, name, id_tensor):
+    	self.name = name
+        self.id_tensor = id_tensor
```
zheng-da, line 132: indent.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1107 @@
...
+        for name, data in kv.data_store.items():
+            meta[name] = (F.shape(data), 
+                          get_type_str(F.dtype(data)), 
+                          kv.part_policy[name].policy_str())
+        res = GetSharedDataResponse(meta)
+        return res
+
+GET_FULL_SHAPE = 901238
+
+class GetFullShapeResponse(rpc.Response):
```
zheng-da, line 434: only shape? no dtype?

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1107 @@
...
+        func : UDF push function
+        """
+        if self._client_id == 0:
+            request = RegisterPushHandlerRequest(func)
+            # send request
+            for server_id in range(self._server_count):
+                rpc.send_request(server_id, request)
+            # wait confirmation
+            for _ in range(self._server_count):
+                response = rpc.recv_response()
```
zheng-da, line 855: the func isn't invoked in the client?

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1107 @@
...
+        """
+        request = BarrierRequest(self._client_id)
+        # send request
+        for server_id in range(self._server_count):
+            rpc.send_request(server_id, request)
+        # recv response
+        for _ in range(self._server_count):
+            response = rpc.recv_response()
+
+    def init_data(self, name, shape, dtype, policy_str, partition_book, init_func):
```
zheng-da, line 921: i think you just need to pass partition policy to kvstore.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1107 @@
...
+                                          tuple(part_shape), 
+                                          get_type_str(dtype), 
+                                          policy_str,
+                                          init_func)
+                for n in range(self._group_count):
+                    server_id = machine_id * self._group_count + n
+                    rpc.send_request(server_id, request)
+            for _ in range(self._server_count):
+                response = rpc.recv_response()
+            self.barrier() # wait all the client and server finish its job
```
zheng-da, line 960: do you still need the barrier? you already waited for all servers' response.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1107 @@
...
+
+        Parameters
+        ----------
+        name : str
+            data name
+        id_tensor : tensor
+            a vector storing the global data ID
+        data_tensor : tensor
+            a tensor with the same row size of data ID
+        """
```
zheng-da, line 1005: are they copied from the previous implementation?

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1107 @@
...
+        # TODO(chao) : remove tmp file using new shared-tensor API
+        assert len(name) > 0, 'name cannot be empty.'
+        if data_tensor is not None: # Create shared-tensor
+            data_type = get_type_str(F.dtype(data_tensor))
+            shared_data = empty_shared_mem(name+'-kvdata-', True, data_tensor.shape, data_type)
+            dlpack = shared_data.to_dlpack()
+            self._data_store[name] = F.zerocopy_from_dlpack(dlpack)
+            self._data_store[name][:] = data_tensor[:]
+            self._write_data_meta_to_file(name+'-kvmeta-'+str(self._machine_id), data_tensor)
```
zheng-da, line 702: you still write metadata to files?

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1107 @@
...
+                if (os.path.exists(name+'-kvmeta-'+str(self._machine_id))):
+                    break
+                else:
+                    time.sleep(1) # wait until the file been created
+            data_shape, data_type = self._read_data_meta_from_file(name+'-kvmeta-'+str(self._machine_id))
+            shared_data = empty_shared_mem(name+'-kvdata-', False, data_shape, data_type)
+            dlpack = shared_data.to_dlpack()
+            self._data_store[name] = F.zerocopy_from_dlpack(dlpack)
+
+    def set_partition_policy(self, name, policy_str, partition_book):
```
zheng-da, line 715: why do we need this? when data is created in kvstore, we cannot change its partition policy.

python/dgl/distributed/constants.py
```diff
@@ -2,3 +2,26 @@
 
 # Maximum size of message queue in bytes
 MAX_QUEUE_SIZE = 20*1024*1024*1024
+
+
+def get_type_str(dtype):
```
zheng-da, line 7: do you need this?
https://github.com/dmlc/dgl/blob/master/python/dgl/backend/pytorch/tensor.py#L19

VoVAllen: `F.reverse_data_type_dict[dtype]` is probably what you want

python/dgl/distributed/rpc_server.py
aksnzhy: I just put the server_state as an argument passed to start_server.

python/dgl/distributed/kvstore.py
aksnzhy: `901231` is just a unique msg_id for specified msg type.

python/dgl/distributed/kvstore.py
aksnzhy: We can get dtype in the get_shared_data() api.

python/dgl/distributed/kvstore.py
aksnzhy: You are right, it's a bug. I changed it.

python/dgl/distributed/kvstore.py
aksnzhy: I don't think so. We also need to set the partition policy on the local client. KVStore can't mapping partition_book from server to client.

python/dgl/distributed/kvstore.py
aksnzhy: Just change the logic.

python/dgl/distributed/kvstore.py
aksnzhy: It will be replaced by new shared-memory api

python/dgl/distributed/kvstore.py
aksnzhy: So you mean we should set policy inside the init_data API ?

python/dgl/distributed/constants.py
aksnzhy: Thanks!

python/dgl/distributed/graph_partition_book.py
```diff
@@ -237,3 +235,95 @@ def get_partition(self, partid):
...
+            data size
+        """
+        if self._policy_str == 'edge':
+            return len(self._partition_book.partid2eids(self._part_id))
+        elif self._policy_str == 'node':
+            return len(self._partition_book.partid2nids(self._part_id))
+        else:
+            raise RuntimeError('Cannot support policy: %s ' % self._policy_str)
+
+    # TODO(chao): more APIs?
```
zheng-da, line 328: i think you can remove it. we can add more APIs later.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1008 @@
...
+    filename : str
+        Path of file
+
+    Returns
+    -------
+    True if file exists.
+    """
+    if os.path.exists(filename):
+        return True
+    return False
```
zheng-da, line 28: this can be simply `os.path.exists(filename)`. i'm not sure if it's necessary to add a function for this.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1008 @@
...
+        kv = server_state.kv_store
+        dtype = F.data_type_dict[self.dtype]
+        if kv.is_backup_server() == False:
+            data_tensor = self.init_func(self.shape, dtype)
+            kv.init_data(name=self.name, data_tensor=data_tensor)
+        else: # backup server will read data from shared-memory
+            kv.init_data(name=self.name)
+        # Find the same partition policy from exsiting plolicy list.
+        for _, policy in kv.part_policy.items():
+            if policy.policy_str == self.policy_str:
```
zheng-da, line 175: why not make the key of the dict the policy name?

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1008 @@
...
+
+    def __setstate__(self, state):
+        self.msg = state
+
+    def process_request(self, server_state):
+        assert self.msg == BARRIER_MSG
+        kv = server_state.kv_store
+        kv.incr_barrier_count()
+        if kv.barrier_count == kv.num_clients:
+            kv.barrier_zero()
```
zheng-da, line 225: maybe you can move this logic to kvstore so you don't need to create this kind of functions.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1008 @@
...
+
+    @property
+    def barrier_count(self):
+        return self._barrier_count
+
+    def incr_barrier_count(self):
+        self._barrier_count +=1
+
+    def barrier_zero(self):
+        self._barrier_count = 0
```
zheng-da, line 589: i feel it's not necessary to expose these APIs.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1008 @@
...
+            If the data_tensor is None, KVServer will
+            read shared-memory when client invoking get_shared_data().
+        """
+        assert len(name) > 0, 'name cannot be empty.'
+        if data_tensor is not None: # Create shared-tensor
+            data_type = F.reverse_data_type_dict[F.dtype(data_tensor)]
+            shared_data = empty_shared_mem(name+'-kvdata-', True, data_tensor.shape, data_type)
+            dlpack = shared_data.to_dlpack()
+            self._data_store[name] = F.zerocopy_from_dlpack(dlpack)
+            self._data_store[name][:] = data_tensor[:]
```
zheng-da, line 643: it doesn't do anything if `data_tensor` is None?

python/dgl/distributed/kvstore.py
zheng-da: yes. if we have this function, we allow people to change partition policy later. i don't think we support it.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1008 @@
...
+        """
+        assert len(name) > 0, 'name cannot be empty.'
+        assert len(shape) > 0, 'shape cannot be empty'
+        assert policy_str in ('edge', 'node'), 'policy_str must be \'edge\' or \'node\'.'
+        shape = list(shape)
+        policy_list = []
+        if self._client_id == 0:
+            for machine_id in range(self._machine_count):
+                policy = PartitionPolicy(policy_str, machine_id, partition_book)
+                part_shape = shape.copy()
```
zheng-da, line 861: does this work for both list and tuple? i remember they have different ways to copy.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1008 @@
...
+            UDF init function
+        """
+        assert len(name) > 0, 'name cannot be empty.'
+        assert len(shape) > 0, 'shape cannot be empty'
+        assert policy_str in ('edge', 'node'), 'policy_str must be \'edge\' or \'node\'.'
+        shape = list(shape)
+        policy_list = []
+        if self._client_id == 0:
+            for machine_id in range(self._machine_count):
+                policy = PartitionPolicy(policy_str, machine_id, partition_book)
```
zheng-da, line 860: you shouldn't create a partition policy for each partition. we have discussed this. the partition book only supports for the local partition.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1008 @@
...
+            Store the partition information
+        init_func : func
+            UDF init function
+        """
+        assert len(name) > 0, 'name cannot be empty.'
+        assert len(shape) > 0, 'shape cannot be empty'
+        assert policy_str in ('edge', 'node'), 'policy_str must be \'edge\' or \'node\'.'
+        shape = list(shape)
+        policy_list = []
+        if self._client_id == 0:
```
zheng-da, line 858: why only one client creates the data in all kvservers? in this way, the client needs to access partition book of other partitions. you can avoid this by asking all clients to create data on their local server.

python/dgl/distributed/kvstore.py
zheng-da: you only need to pass the partition policy for the local partition. i think passing a partition policy should be sufficient.

python/dgl/distributed/rpc_server.py
```diff
@@ -2,9 +2,9 @@
 
 from . import rpc
 from .constants import MAX_QUEUE_SIZE
-from .server_state import get_server_state
+from .server_state import ServerState
 
-def start_server(server_id, ip_config, num_clients, \
+def start_server(server_id, ip_config, num_clients, server_state, \
```
zheng-da, line 7: why not just call get_server_state to get server state?

python/dgl/distributed/rpc_server.py
```diff
@@ -70,10 +72,14 @@ def start_server(server_id, ip_config, num_clients, \
...
             register_res = rpc.ClientRegisterResponse(client_id)
             rpc.send_response(client_id, register_res)
-    server_state = get_server_state()
     # main service loop
     while True:
         req, client_id = rpc.recv_request()
         res = req.process_request(server_state)
         if res is not None:
-            rpc.send_response(client_id, res)
+            if isinstance(res, list):
```
zheng-da, line 80: in what case `res` is a list? what happens if someone returns a tuple?

tests/compute/test_new_kvstore.py
```diff
@@ -0,0 +1,161 @@
...
+	# Init kvserver
+    kvserver = dgl.distributed.KVServer(server_id=0,
+                                        ip_config='ip_config.txt',
+                                        num_clients=1)
+    kvserver.init_data(name='data_0', data_tensor=data_0)
+    kvserver.set_partition_policy('data_0', 'node', gpb)
+    kvserver.init_data(name='data_1', data_tensor=data_1)
+    kvserver.set_partition_policy('data_1', 'edge', gpb)
+    # start server
+    server_state = dgl.distributed.ServerState(kv_store=kvserver)
```
zheng-da, line 67: how does get_server_state return thsi server state?

tests/compute/test_new_kvstore.py
```diff
@@ -0,0 +1,161 @@
...
+data_1 = F.tensor([[2.,2.],[2.,2.],[2.,2.],[2.,2.],[2.,2.],[2.,2.],[2.,2.]], F.float32)
+data_2 = F.tensor([[0.,0.],[0.,0.],[0.,0.],[0.,0.],[0.,0.],[0.,0.]], F.float32)
+
+def init_zero_func(shape, dtype):
+    return F.zeros(shape, dtype)
+
+def udf_push(target, name, id_tensor, data_tensor):
+    target[name] = F.scatter_row(target[name], id_tensor, data_tensor*data_tensor)    
+
+def test_partition_policy():
```
zheng-da, line 41: can you try partition book with multiple partitions?

tests/compute/test_new_kvstore.py
```diff
@@ -0,0 +1,161 @@
...
+def test_kv_store():
+    ip_config = open("ip_config.txt", "w")
+    ip_config.write('127.0.0.1 30050 1\n')
+    ip_config.close()
+    pid = os.fork()
+    if pid == 0:
+        start_server()
+    else:
+        time.sleep(1)
+        start_client()
```
zheng-da, line 157: can you test backup servers? i think this can still be tested in a single machine.

zheng-da:
please add description of the PR.

python/dgl/distributed/kvstore.py
aksnzhy: The key is the data's name. We may have many data using the same policy.

python/dgl/distributed/kvstore.py
aksnzhy: Ok. I think we can just use a setter.

python/dgl/distributed/kvstore.py
aksnzhy: I change the logic here. If data_tensor = None, backup_server will do nothing. The backup-server read the shared-tensor only when client invoke get_shared() API.

python/dgl/distributed/kvstore.py
aksnzhy: The input shape has already been converted to a list here.

python/dgl/distributed/kvstore.py
aksnzhy: But as we disucssed before, the shape argument for init_data() is a full shape not partitioned shape right? So you must access partition book to partition shape across servers.

python/dgl/distributed/kvstore.py
aksnzhy: Fixed it.

python/dgl/distributed/rpc_server.py
aksnzhy: get_server_state() is a c API but here we didn't use any c code.

tests/compute/test_new_kvstore.py
aksnzhy: Actually the get_server_state has not been implemented.

tests/compute/test_new_kvstore.py
aksnzhy: Sure, I will try that on real distributed environment.

tests/compute/test_new_kvstore.py
aksnzhy: Sure.

python/dgl/distributed/kvstore.py
aksnzhy: But we need the parition_book to calculate the shape of other partitions.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1072 @@
...
+        target tensor
+    name : str
+        data name
+    id_tensor : tensor
+        a vector storing the ID list.
+    data_tensor : tensor
+        a tensor with the same row size of id
+    """
+    # TODO(chao): support Tensorflow backend
+    target[name][id_tensor] = data_tensor
```
zheng-da, line 488: how to support tensorflow backend?

python/dgl/distributed/kvstore.py
zheng-da: can you put comments here?

python/dgl/distributed/kvstore.py
zheng-da: we only need to get the shape of the local partition.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -237,3 +226,95 @@ def get_partition(self, partid):
...
+
+        Returns
+        -------
+        int
+            data size
+        """
+        if self._policy_str == 'edge':
+            return len(self._partition_book.partid2eids(self._part_id))
+        elif self._policy_str == 'node':
+            return len(self._partition_book.partid2nids(self._part_id))
```
zheng-da, line 318: this operation might be expensive. it's better to add an API to get data size in a more efficient way.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1072 @@
...
+        if self._freeze:
+            raise RuntimeError("KVClient cannot create new \
+                data after invoking get_shared_data() API.")
+        shape = list(shape)
+        if self._client_id == 0:
+            for machine_id in range(self._machine_count):
+                if policy_str == 'edge':
+                    part_dim = len(partition_book.partid2eids(machine_id))
+                elif policy_str == 'node':
+                    part_dim = len(partition_book.partid2nids(machine_id))
```
zheng-da, line 856: this is implemented in partition policy. it's better to reuse the method there.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1072 @@
...
+                    rpc.send_request(server_id, request)
+            for _ in range(self._server_count):
+                response = rpc.recv_response()
+                assert response.msg == INIT_MSG
+        self.barrier()
+        # Create local shared-data
+        if policy_str == 'edge':
+            local_dim = len(partition_book.partid2eids(self._machine_id))
+        elif policy_str == 'node':
+            local_dim = len(partition_book.partid2nids(self._machine_id))
```
zheng-da, line 877: the same comment as above.

python/dgl/distributed/kvstore.py
```diff
@@ -0,0 +1,1072 @@
...
+            raise RuntimeError("Data shape %s has already exists!" % name)
+        self._part_policy[name] = PartitionPolicy(policy_str, self._part_id, partition_book)
+        shared_data = empty_shared_mem(name+'-kvdata-', False, \
+            local_shape, F.reverse_data_type_dict[dtype])
+        dlpack = shared_data.to_dlpack()
+        self._data_store[name] = F.zerocopy_from_dlpack(dlpack)
+        self._data_name_list.add(name)
+        self._full_data_shape[name] = tuple(shape)
+
+    def get_shared_data(self, partition_book):
```
zheng-da, line 896: maybe call it `map_shared_data`. `get_shared_data` sounds like it will return something.

python/dgl/distributed/rpc.py
```diff
@@ -515,7 +515,7 @@ def send_request(target, request):
     server_id = target
     data, tensors = serialize_to_payload(request)
     msg = RPCMessage(service_id, msg_seq, client_id, server_id, data, tensors)
-    send_rpc_message(msg)
+    send_rpc_message(msg, server_id)
```
zheng-da, line 518: the message already has server_id. why do you need to pass server_id again?

python/dgl/distributed/rpc_server.py
```diff
@@ -65,15 +68,20 @@ def start_server(server_id, ip_config, num_clients, \
     for client_id, addr in client_namebook.items():
         client_ip, client_port = addr.split(':')
         rpc.add_receiver_addr(client_ip, client_port, client_id)
+    time.sleep(3)
```
zheng-da, line 71: why do you need to sleep here?

tests/distributed/test_new_kvstore.py
```diff
@@ -0,0 +1,179 @@
...
+g.add_edge(2, 3) # 3
+g.add_edge(1, 1) # 4
+g.add_edge(0, 4) # 5
+g.add_edge(2, 5) # 6
+
+g.ndata[dgl.NID] = global_nid
+g.edata[dgl.EID] = global_eid
+
+gpb = dgl.distributed.GraphPartitionBook(part_id=0,
+                                         num_parts=1,
```
zheng-da, line 37: can you test the multi-partition case?

tests/distributed/test_new_kvstore.py
```diff
@@ -0,0 +1,179 @@
...
+    assert_array_equal(F.asnumpy(local_nid), F.asnumpy(F.tensor([0,1,2,3,4,5], F.int64)))
+    assert_array_equal(F.asnumpy(local_eid), F.asnumpy(F.tensor([0,1,2,3,4,5,6], F.int64)))
+    nid_partid = node_policy.to_partid(F.tensor([0,1,2,3,4,5], F.int64))
+    eid_partid = edge_policy.to_partid(F.tensor([0,1,2,3,4,5,6], F.int64))
+    assert_array_equal(F.asnumpy(nid_partid), F.asnumpy(F.tensor([0,0,0,0,0,0], F.int64)))
+    assert_array_equal(F.asnumpy(eid_partid), F.asnumpy(F.tensor([0,0,0,0,0,0,0], F.int64)))
+    assert node_policy.get_data_size() == len(node_map)
+    assert edge_policy.get_data_size() == len(edge_map)
+
+def start_server():
```
zheng-da, line 77: can you test the backup servers?

python/dgl/distributed/kvstore.py
VoVAllen: Not supported for now. Let's come back later after we fully understand what kind functions are needed here.

tests/distributed/test_new_kvstore.py
aksnzhy: Multi-partition need real distributed setting. I already tested it.

python/dgl/distributed/kvstore.py
aksnzhy: Sure

PR #1570, opened 2020-05-29
[Optimization] Optimize pickling of heterograph

soodoshll:
## Description

Implement the major part in C to avoid calling too many ``__getstate__`` functions in python.

**Time for transmitting 1000 graphs(from CoraDataset)**
- old: 2.20s
- new: 1.03s 

## Checklist
- [ ] Code is well-documented


BarclayII:
When you're finished please request a review.  Thanks for your effort!

src/graph/pickle.cc
```diff
@@ -14,85 +17,136 @@ namespace dgl {
...
+  strm->Write(fmts);
+  strm->Write(flags);
   return states;
 }
 
 HeteroGraphPtr HeteroUnpickle(const HeteroPickleStates& states) {
-  const auto metagraph = states.metagraph;
-  const auto &num_nodes_per_type = states.num_nodes_per_type;
-  CHECK_EQ(states.adjs.size(), metagraph->NumEdges());
+  char *buf = const_cast<char *>(states.meta.c_str());  // a readonly stream?
```
VoVAllen, line 58: Why not still using `MemoryStringStream`?

src/graph/pickle.cc
```diff
@@ -14,85 +17,136 @@ namespace dgl {
...
   HeteroPickleStates states;
-  states.metagraph = graph->meta_graph();
-  states.num_nodes_per_type = graph->NumVerticesPerType();
-  states.adjs.resize(graph->NumEdgeTypes());
+  dmlc::MemoryStringStream ofs(&states.meta);
+  dmlc::Stream *strm = &ofs;
+  strm->Write(ImmutableGraph::ToImmutable(graph->meta_graph()));
+  strm->Write(graph->NumVerticesPerType());
+  std::vector<SparseFormat> fmts;
+  std::vector<char> flags;
```
VoVAllen, line 25: Why does `vector<char>` needed here? You can read/write booleans separately, why you prefer first constructing a vector and then write them?

VoVAllen:
I'm a bit concerned about the way of implementation. This would only accelerate pytorch case, right? And according to pytorch, this acceleration should only be applicable for the multiprocessing scenario, which should be done by `ForkingPickler.register(DGLHeteroGraph, reduce_function)` instead of changing `setstate` and `getstate` for the original codes. @jermainewang 

My expectation would be only register this pickling method when using pytorch under multiprocessing scenario. Otherwise a naive implementation(or dgl's shared memory tensor in the future) could be fine.

src/graph/pickle.cc
soodoshll: This line is to drop the `const`. It is still necessary if using `MemoryStringStream`?

VoVAllen: You can directly use `dmlc::MemoryStringStream ifs(&states.meta);` as how you write.

jermainewang:
> I'm a bit concerned about the way of implementation. This would only accelerate pytorch case, right? And according to pytorch, this acceleration should only be applicable for the multiprocessing scenario, which should be done by `ForkingPickler.register(DGLHeteroGraph, reduce_function)` instead of changing `setstate` and `getstate` for the original codes. @jermainewang

What's the concern here? Do you foresee this implementation will slowdown pickle/unpickle in frameworks other than PyTorch?
 
> My expectation would be only register this pickling method when using pytorch under multiprocessing scenario. Otherwise a naive implementation(or dgl's shared memory tensor in the future) could be fine.

What do you mean by a naive implementation? Is the the current implementation or saving everything in one buffer?

VoVAllen:
My only concern is whether the current implementation is stable(whether there would be compatibility issue in the future). Pickle to file is different from pickle in multiprocessing, since there's no compatibility issue in the multiprocessing scenario. Therefore the purpose of multiprocessing pickle could be fast, but the purpose of file pickle could be stable.

For example, maybe we will need to turn to other methods to optimize the multiprocessing performance in Tensorflow, which might break the compatibility of file pickle in the future. Using `ForkingPickler` could be address this kind of issue.

src/graph/pickle.cc
```diff
@@ -107,8 +186,23 @@ DGL_REGISTER_GLOBAL("heterograph_index._CAPI_DGLHeteroPickle")
 DGL_REGISTER_GLOBAL("heterograph_index._CAPI_DGLHeteroUnpickle")
 .set_body([] (DGLArgs args, DGLRetValue* rv) {
     HeteroPickleStatesRef ref = args[0];
-    HeteroGraphPtr graph = HeteroUnpickle(*ref.sptr());
+    HeteroGraphPtr graph = ref->version ?
```
VoVAllen, line 189: Please use switch/if statement and check all the condition. (raise error by LOG(FATAL) if ref->version is not 0 or 1)

VoVAllen:
Thanks!

PR #1589, opened 2020-06-04
[Optimization] Improve performance in ToBlock function

IzabelaMazur:
Changed std::unordered_map to google::dense_hash_map

Time of execution for train_sampling.py from Large_script:
before this patch: 22,399326 [s]
after patch application: 15,102482 [s]

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "third_party/METIS"]
 	path = third_party/METIS
 	url = https://github.com/KarypisLab/METIS.git
+[submodule "third_party/googlehash"]
+	path = third_party/googlehash
+	url = https://github.com/sparsehash/sparsehash.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,15 @@ include_directories("third_party/dmlc-core/include")
 include_directories("third_party/minigun/minigun")
 include_directories("third_party/minigun/third_party/moderngpu/src")
 
+include(ExternalProject)
+ExternalProject_Add(googlehash
+  PREFIX googlehash
+  SOURCE_DIR ../third_party/googlehash/src
+  CONFIGURE_COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/third_party/googlehash/configure --prefix ${CMAKE_CURRENT_BINARY_DIR}/googlehash
+  BUILD_COMMAND make install
+)
+include_directories("${CMAKE_CURRENT_BINARY_DIR}/googlehash/include")
+
 # initial variables
 set(DGL_LINKER_LIBS "")
 if(MSVC OR CMAKE_SYSTEM_NAME STREQUAL "Darwin")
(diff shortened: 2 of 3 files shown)


BarclayII:
Do you need help adding Google hash map as a git submodule?

jermainewang:
Hi @IzabelaMazur , thanks for the contribution. One question I have is how does it compare with a hashmap implemented by a vector. This is viable because the key range in the ToBlock function is always `[0, num_nodes)`. By creating a vector of size `num_nodes`, it is always O(1) to relabel the IDs. The potential problem is the cost of memory allocation. Would you please also benchmark against this solution? Thx.

IzabelaMazur:
Hi @BarclayII ,
I'm in the process of adding this submodule I've met a few minor issues and it takes a while. If I have any bigger problems, I will ask you for help. Thanks!

Hi @jermainewang ,
I going to measure the memory allocations 
on the benchmark: https://github.com/dglai/dgl-benchmark/blob/master/graphsage-mp/train_sampling.py and will send you the results.

zheng-da:
@jermainewang creating a vector of num_nodes isn't feasible to scale to a graph with billions of nodes.

jermainewang:
> @jermainewang creating a vector of num_nodes isn't feasible to scale to a graph with billions of nodes.

Agree, but we should still consider such an option for medium-sized graph. Our benchmark shows that PyG is faster than DGL for cluster-based sampling due to them using this kind of hashmap.

BarclayII:
Looks like the building configuration for Windows should be modified.

BarclayII:
@IzabelaMazur I wasn't able to build google-sparsehash with Visual Studio 2017 as it threw lots of undefined names or other syntax errors.

Can we switch to e.g. https://github.com/sparsehash/sparsehash-c11 instead?  It's almost a drop-in replacement of googlehash that has C++11 support.  Also since it's a header-only library, we don't need to include it as an external CMake project; simply adding include directories should do everything.

Also I found a couple alternative implementations that claims to be more efficient than both std and googlehash in general:

* https://github.com/greg7mdp/sparsepp
* https://github.com/greg7mdp/parallel-hashmap

Wonder if you had any experience with them?  You don't have to benchmark them though.

Thanks.

zheng-da:
do we have any detailed profiling results of using the dense hashmap?

IzabelaMazur:
Hi @BarclayII I didn't know the functions you mentioned. Currently I'm checking them and comparing them with the proposed google::dense_hash_map(). From an early check it looks like phmap::flat_hash_map is less than 1 second faster. Average epoch time for google::dense_hash_map takes 19.9 seconds, and for phmap::flat_hash_map takes 19.0 seconds. The rest of the functions are slower than google::danse_hash_map.
I'm preparing a pull request with phmap::flat_hash_map(). I would still like to check the size of the allocated memory using both these functions.

BarclayII:
Seems that you fixed the Windows problem.  Thanks!

The speed up by phmap seems to be minor (only 5%) so I guess sticking with google sparsehash should be fine.  @zheng-da what do you think?

IzabelaMazur:
In case you were also interested in changing to phmap::flat_hash_map I created a new PR.
https://github.com/dmlc/dgl/pull/1662

VoVAllen:
Found an interesting benchmark of hashmap in C++
https://martin.ankerl.com/2019/04/01/hashmap-benchmarks-04-03-result-RandomFind_500000/

BarclayII:
Closing this since phmap version is integrated.

PR #1597, opened 2020-06-07
Refactor distributed graph store

zheng-da:
## Description
Refactor the distributed graph store for the new kvstore.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/graph_partition_book.py
+++ b/python/dgl/distributed/graph_partition_book.py
@@ -5,6 +5,52 @@ import numpy as np
 from .. import backend as F
 from ..base import NID, EID
 from .. import utils
+from .shared_mem_utils import _to_shared_mem, _get_ndata_path, _get_edata_path, DTYPE_DICT
+from .._ffi.ndarray import empty_shared_mem
+
+def _move_metadata_to_shared_mam(graph_name, num_nodes, num_edges, part_id,
+                                 num_partitions, node_map, edge_map):
+    ''' Move all metadata to the shared memory.
+
+    We need these metadata to construct graph partition book.
+    '''
+    meta = _to_shared_mem(F.tensor([num_nodes, num_edges,
+                                    num_partitions, part_id]),
+                          _get_ndata_path(graph_name, 'meta'))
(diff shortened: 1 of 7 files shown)


python/dgl/distributed/graph_partition_book.py
```diff
@@ -78,6 +141,18 @@ def __init__(self, part_id, num_parts, node_map, edge_map, part_graph):
...
+
+        Parameters
+        ----------
+        graph_name : str
+            The graph name
+        """
+        self._meta, self._nid2partid, self._eid2partid = _move_metadata_to_shared_mam(
+            graph_name, self.num_nodes(), self.num_edges(), self._part_id, self._num_partitions,
+            self._nid2partid, self._eid2partid)
+
```
aksnzhy, line 155: So we will not put GraphPartitionBook to ServerState, right?

python/dgl/distributed/kvstore.py
```diff
@@ -953,7 +935,6 @@ def map_shared_data(self, partition_book):
             for _ in range(self._group_count-1):
                 response = rpc.recv_response()
                 assert response.msg == SEND_META_TO_BACKUP_MSG
-        self._freeze = True
```
aksnzhy, line 956: I think we cannot just simply remove the `freeze` code here. I will change the code in a new PR to fix it.

aksnzhy:
The overall comment is: should we put all the data (except for the dynamic data tensor in kvstore) like GraphPartitionBook and meta data to ServerState and manage the shared-memory in a unified way?

zheng-da:
I, in general, prefer not to. If we want to put everything in server state, it makes the design inflexible to a future extension. If we want to add something, we need to modify the server state. what is the benefit of having everything managed by server state?

python/dgl/distributed/graph_partition_book.py
zheng-da: we can have graph partition book referenced by server state. but we don't need to have it managed by server state.

aksnzhy:
I removed the `freeze` in this pr:https://github.com/dmlc/dgl/pull/1605, and I solved the conflict in this PR

zheng-da:
i have added the reference of graph partition book to server state.

PR #1608, opened 2020-06-09
[WIP] IGMC

sneakerkg:
## Description
This PR is the implementation of the paper: Inductive Matrix Completion Based on Graph Neural Networks.
Current status, on ML-100K dataset, the reported RMSE is 0.905, what we reproduced is 0.931. The gap might come from the lack of self-loop in our model. This PR already contains the fix, but model is still training thus can't report number here.

A few TODOs to make this one ready to merge:
1.Get the same RMSE result as reported in the paper;
2.Currently we extract all the subgraphs in advance before an epoch. This is OK for ML-100K dataset, will have performance trouble when go to larger dataset as ML-1M;
3.We prefer to implement the subgraph extraction code using DGL APIs to see if the code can be cleaner;
4.We have code duplication within this PR and with code in GCMC, need further cleaning.

## Checklist
- [ ] Code is well-documented


sneakerkg:
Will deprecate due to Junfeng's PR: https://github.com/dmlc/dgl/pull/1815

PR #1616, opened 2020-06-10
[RPC] Sampling over RPC

VoVAllen:
## Description

Currently only support `dgl.graph`

## Checklist
- [ ] Code is well-documented


python/dgl/distributed/sampling.py
```diff
@@ -0,0 +1,100 @@
...
+
+def sample_neighbors(g: DistGraph, nodes, fanout, edge_dir='in', prob=None, replace=False):
+    """Sample neighbors"""
+    assert edge_dir == 'in'
+    req_list = []
+    partition_book = g.get_partition_book()
+
+    partition_id = partition_book.nid2partid(F.tensor(nodes))
+    node_id_per_partition = [[]
+                             for _ in range(partition_book.num_partitions())]
```
zheng-da, line 87: is this `[[]] * partition_book.num_partitions()`?

python/dgl/distributed/sampling.py
```diff
@@ -0,0 +1,100 @@
...
+    src_tensor = F.cat(srcs, 0)
+    dst_tensor = F.cat(dsts, 0)
+    eid_tensor = F.cat(eids, 0)
+    g = graph((src_tensor, dst_tensor),
+              restrict_format='coo', num_nodes=num_nodes)
+    g.edata[EID] = eid_tensor
+    return g
+
+
+def sample_neighbors(g: DistGraph, nodes, fanout, edge_dir='in', prob=None, replace=False):
```
zheng-da, line 79: does qidong's version implement sampling like this?

zheng-da:
can you add the sampling RPC to DistGraphServer?

python/dgl/distributed/dist_graph.py
```diff
@@ -271,7 +271,7 @@ def __init__(self, server_id, ip_config, num_clients, graph_name, conf_file):
 
         # Load graph partition data.
         self.client_g, node_feats, edge_feats, self.gpb = load_partition(conf_file, server_id)
-        self.client_g = _copy_graph_to_shared_mem(self.client_g, graph_name)
+        # self.client_g = _copy_graph_to_shared_mem(self.client_g, graph_name)
```
VoVAllen, line 274: This is commented out for test reason. Can we add an option here? @zheng-da

python/dgl/distributed/sampling.py
```diff
@@ -0,0 +1,104 @@
...
+        self.seed_nodes, self.edge_dir, self.prob, self.replace, self.fan_out = state
+
+    def __getstate__(self):
+        return self.seed_nodes, self.edge_dir, self.prob, self.replace, self.fan_out
+
+    def process_request(self, server_state):
+        local_g = server_state.graph
+        partition_book = server_state.partition_book
+        local_ids = F.astype(partition_book.nid2localnid(
+            F.tensor(self.seed_nodes), partition_book._part_id), local_g.idtype)
```
VoVAllen, line 50: Here I use a private attribute. I see it's exposed in PartitionPolicy. Do I need to add a new attribute to GraphPartitionBook & RangePartitionBook?

python/dgl/distributed/dist_graph.py
zheng-da: if you want to have three sampler server running, you need to set up the "backup" server. @aksnzhy can you provide any suggestions on how to set up backup servers?

python/dgl/distributed/rpc_server.py
```diff
@@ -78,24 +78,14 @@ def start_server(server_id, ip_config, num_clients, server_state, \
...
+        res = req.process_request(server_state)
+        if res is not None:
+            if isinstance(res, list):
+                for response in res:
+                    target_id, res_data = response
+                    rpc.send_response(target_id, res_data)
+            elif isinstance(res, str) and res == 'exit':
+                break # break the loop and exit server
+            else:
+                rpc.send_response(client_id, res)
```
zheng-da, line 91: i thought chao has modified it.

python/dgl/distributed/sampling.py
zheng-da: yes, please don't use a private attribute. we can potentially keep the partition id in the server state.

src/rpc/network/socket_communicator.h
```diff
@@ -19,7 +19,7 @@
 namespace dgl {
 namespace network {
 
-static int kMaxTryCount = 1024;    // maximal connection: 1024
+static int kMaxTryCount = 10;    // maximal connection: 1024
```
zheng-da, line 22: why modify this?

tests/distributed/test_distributed_sampling.py
```diff
@@ -0,0 +1,87 @@
...
+    print(g.number_of_nodes())
+    results = sample_neighbors(g, [0, 10, 99, 66, 1024, 2008], 3)
+    print("after sample")
+    dgl.distributed.shutdown_servers()
+    dgl.distributed.finalize_client()
+    return results
+
+
+@unittest.skipIf(os.name == 'nt', reason='Do not support windows yet')
+@unittest.skipIf(dgl.backend.backend_name != 'PyTorch', reason='Only support pytorch for now')
```
zheng-da, line 43: does it not work with mxnet?

tests/distributed/test_distributed_sampling.py
```diff
@@ -0,0 +1,87 @@
...
+    # sys.excepthook = myexcepthook
+    # partition graph
+    g = CitationGraphDataset("cora")[0]
+    g.readonly()
+    print(g.idtype)
+    num_parts = num_server
+    num_hops = 1
+
+    partition_graph(g, 'test_sampling', num_parts, tmpdir,
+                    num_hops=num_hops, part_method='metis', reshuffle=False)
```
zheng-da, line 61: please add a comment that this will be tested with range partition as well in the future.

python/dgl/distributed/rpc_server.py
VoVAllen: I tested if the program exits with an exception it will still close the sockets. No need to catch it and call finalize. And catch it without print errors/rethrow it makes it hard to debug

python/dgl/distributed/sampling.py
VoVAllen: This way of initialization is wrong
```python
In [7]: a = [[]]*10                                                                                                                                                                                                                                                             

In [8]: a                                                                                                                                                                                                                                                                       
Out[8]: [[], [], [], [], [], [], [], [], [], []]

In [9]: a[0].append(1)                                                                                                                                                                                                                                                          

In [10]: a                                                                                                                                                                                                                                                                      
Out[10]: [[1], [1], [1], [1], [1], [1], [1], [1], [1], [1]]
```

tests/distributed/test_distributed_sampling.py
VoVAllen: I heard MXNet 2.0 will break lots of backward compatibility. I think we can support MXNet later

tests/distributed/test_distributed_sampling.py
VoVAllen: How to test it with RangePartitionBook. And should RangePartitionBook & GraphPartitionBook inherits an abstract class such as `PartitionBook`?

python/dgl/distributed/dist_graph.py
```diff
@@ -319,7 +320,7 @@ class DistGraph:
     def __init__(self, ip_config, graph_name):
         connect_to_server(ip_config=ip_config)
         self._client = KVClient(ip_config)
-
+        graph_name = "{}_{}".format(graph_name, self._client.machine_id)
```
VoVAllen, line 323: I changed graph_name to graph_name+partition_id, to avoid conflict name when testing on the same machine

zheng-da: the problem is how the client knows which partition it should map?

tests/distributed/test_distributed_sampling.py
zheng-da: it can be inherited from the same class. but does it matter in Python?

zheng-da: as for how to test RangePartitionBook, you can use 'orig_id' to get the node Id in the original graph. something like below should work.
```python
sampled_graph.ndata['orig_id'] = dist_graph.ndata['orig_id'][sampled_graph.ndata[dgl.NID]]
src, dst = sampled_graph.edges()
src = sampled_graph.ndata['orig_id'][src]
dst = sampled_graph.ndata['orig_id'][dst]
assert np.all(F.asnumpy(g.has_edges_between(src, dst).bool()))
```

tests/distributed/test_distributed_sampling.py
zheng-da: but we are not using mxnet 2.0.

zheng-da: how difficult is it to make mxnet work?

tests/distributed/test_distributed_sampling.py
VoVAllen: > it can be inherited from the same class. but does it matter in Python?

Inheritance make it easy to understand that they have the same interface. (without documentation)

src/rpc/network/socket_communicator.h
VoVAllen: Max try 1024 is too much?

PR #1630, opened 2020-06-12
[Op] Farthest Point Sampler in Cpp and CUDA

hetong007:
## Description
In point cloud related applications, the **Farthest Point Sampler** is the most popular down-sample algorithm. Because of its iterative nature, it is much faster to implement in C++ than python (10x faster on input size at the ModelNet40 scale).

This PR provides FPS algorithm on CPU and CUDA, as well as the python interfaces for pytorch and mxnet, and tests.

## Checklist
- [X] Code is well-documented


yzh119:
@hetong007 any performance numbers?

hetong007:
@yzh119 Sure. My benchmark on a V100 gets this result:

```bash
$ python benchmark_pytorch.py
Using backend: pytorch
0.09509611129760742
20.0512216091156
```

The script:
```python
import dgl
import torch
import numpy as np
import time

from dgl.geometry.pytorch import FarthestPointSampler
from torch import nn

class FarthestPointSamplerPython(nn.Module):
    def __init__(self, npoints):
        super(FarthestPointSamplerPython, self).__init__()
        self.npoints = npoints

    def forward(self, pos):
        device = pos.device
        B, N, C = pos.shape
        centroids = torch.zeros(B, self.npoints, dtype=torch.long).to(device)
        distance = torch.ones(B, N).to(device) * 1e10
        farthest = torch.randint(0, N, (B,), dtype=torch.long).to(device)
        batch_indices = torch.arange(B, dtype=torch.long).to(device)
        for i in range(self.npoints):
            centroids[:, i] = farthest
            centroid = pos[batch_indices, farthest, :].view(B, 1, C)
            dist = torch.sum((pos - centroid) ** 2, -1)
            mask = dist < distance
            distance[mask] = dist[mask]
            farthest = torch.max(distance, -1)[1]
        return centroids

# benchmark
batch_size = 32
N = 1024 * batch_size
sample_points = 512
LOOP = 100

x = torch.tensor(np.random.uniform(size=(batch_size, int(N/batch_size), 3))).to("cuda").to(dtype=torch.float)
fps = FarthestPointSampler(sample_points)
fps_python = FarthestPointSamplerPython(sample_points)

tic = time.time()
for i in range(LOOP):
    res = fps(x)
print(time.time() - tic)

tic = time.time()
for i in range(LOOP):
    res = fps_python(x)
print(time.time() - tic)
```

src/array/cuda/geometry_op_impl.cu
```diff
@@ -0,0 +1,137 @@
...
+
+  // sample for each cloud in the batch
+  IdType* start_idx_data = static_cast<IdType*>(start_idx->data);
+
+  fps_kernel<<<batch_size, THREADS, 0, thr_entry->stream>>>(
+    array_data, batch_size, sample_points,
+    point_in_batch, dim, start_idx_data, dist_data, ret_data);
+}
+
+#ifdef DGL_USE_CUDA
```
BarclayII, line 120: Do we need this condition even if the file suffix is `.cu`?

src/array/array.cc
```diff
@@ -679,5 +717,17 @@ DGL_REGISTER_GLOBAL("ndarray._CAPI_DGLCreateSparseMatrix")
...
+    const NDArray data = args[0];
+    const int64_t batch_size = args[1];
+    const int64_t sample_points = args[2];
+    NDArray dist = args[3];
+    IdArray start_idx = args[4];
+    IdArray result = args[5];
+    const DLDeviceType XPU = data->ctx.device_type;
+    FarthestPointSampler(data, batch_size, sample_points, dist, start_idx, result);
+  });
+
```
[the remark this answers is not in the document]
hetong007: So here are two questions. 
1. Should we put it in `src/array` or in a separate package? @jermainewang 's opinion is to put it into array, as there won't be too many related ops and then the single-out part will just be a mini package.
2. If to be singled out, what is the name of this package? `point_cloud` sounds too narrow as it might be related to 3d applications in other data formats.

src/array/array.cc
```diff
@@ -629,6 +630,43 @@ std::pair<COOMatrix, IdArray> COOCoalesce(COOMatrix coo) {
...
 }
 
+/////////////////////////////////// Geometry Op ///////////////////////////////////
+
+void FarthestPointSampler(NDArray array, int64_t batch_size, int64_t sample_points,
+    NDArray dist, IdArray start_idx, IdArray result) {
+  CHECK_EQ(array->ctx, result->ctx) << "Array and The result should be on the same device.";
+  const DLDeviceType XPU = array->ctx.device_type;
+
+#ifdef DGL_USE_CUDA
```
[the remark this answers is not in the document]
hetong007: Switched to `ATEN_XPU_SWITHC_CUDA`, which was just merged 2 hours ago.

src/array/array.cc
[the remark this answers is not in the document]
BarclayII: Having minipackages sounds better to me as the code would be easier to navigate in e.g. source browsers.  If we put the geometry-specific code in `array`, soon we will have lots of domain-specific kernels in `array` when we extend to other domains and the code will be much harder to maintain.

If `point_cloud` sounds too narrow then how about `geometry`?

I would say that the `geometry_op_impl` files should be in the `geometry` minipackage as well.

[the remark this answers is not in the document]
hetong007: I can sure make the change if all reviewers agree. For naming, `geometry` sounds good to me (obviously).

jermainewang:
Unassigned myself since @BarclayII and @yzh119 have taken over.

python/dgl/geometry/mxnet/fps.py
```diff
@@ -0,0 +1,42 @@
...
+#pylint: disable=no-member, invalid-name
+
+from mxnet import nd
+from mxnet.gluon import nn
+import numpy as np
+
+from ..capi import farthest_point_sampler
+
+class FarthestPointSampler(nn.Block):
+    """Farthest Point Sampler
```
yzh119, line 11: Could you please elaborate more on how it works?

python/dgl/geometry/pytorch/fps.py
```diff
@@ -0,0 +1,41 @@
+"""Farthest Point Sampler for pytorch Geometry package"""
+#pylint: disable=no-member, invalid-name
+
+import torch as th
+from torch import nn
+
+from ..capi import farthest_point_sampler
+
+class FarthestPointSampler(nn.Module):
+    """Farthest Point Sampler
```
yzh119, line 10: ditto

src/geometry/cpu/geometry_op_impl.cc
```diff
@@ -0,0 +1,100 @@
...
+
+namespace dgl {
+using runtime::NDArray;
+namespace geometry {
+namespace impl {
+
+/*!
+ * \brief Farthest Point Sampler without the need to compute all pairs of distance.
+ * 
+ * The input array has shape (N, d), where N is the number of points, and d is the dimension.
```
yzh119, line 18: Copy this docstring to python end.

src/geometry/cuda/geometry_op_impl.cu
```diff
@@ -0,0 +1,133 @@
...
+      if (dist_data[array_start + j] > dist_max) {
+        dist_argmax = j;
+        dist_max = dist_data[array_start + j];
+      }
+    }
+
+    dist_max_ht[thread_idx] = dist_max;
+    dist_argmax_ht[thread_idx] = dist_argmax;
+
+    // Parallel Reduction
```
yzh119, line 79: Could you please explain how does it work? Looks like tree reduction but not exactly the same.

src/geometry/cpu/geometry_op_impl.cc
```diff
@@ -0,0 +1,100 @@
...
+ * It consists of a (flatten) batch of point clouds.
+ *
+ * In each batch, the algorithm starts with the sample index specified by ``start_idx``.
+ * Then for each point, we maintain the minimum to-sample distance.
+ * Finally, we pick the point with the maximum such distance.
+ * This process will be repeated for ``sample_points`` - 1 times.
+ */
+template <DLDeviceType XPU, typename FloatType, typename IdType>
+void FarthestPointSampler(NDArray array, int64_t batch_size, int64_t sample_points,
```
yzh119, line 27: Have a check of the validity of dist, start_idx and result.

src/geometry/cuda/geometry_op_impl.cu
hetong007: Add a short proof

src/geometry/cpu/geometry_op_impl.cc
hetong007: Is there a standard MACRO for such check?

BarclayII: `CHECK(cond) << msg`, `CHECK_LE(a, b) << msg`, `CHECK_EQ(a, b) << msg`, etc.

PR #1646, opened 2020-06-15
[KVStore] make pull/push handler per tensor.

zheng-da:
## Description
The current KVStore pull/push handler is global. It's better to make it per tensor. For example, a model contains learnable node embeddings and we want to train the model with control variate methods. We need to support push handler that updates node embeddings with sparse Adagrad as well as push handler that simply writes new data to the corresponding rows.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/kvstore.py
+++ b/python/dgl/distributed/kvstore.py
@@ -60,7 +60,7 @@ class PullRequest(rpc.Request):
         if kv_store.data_store.__contains__(self.name) is False:
             raise RuntimeError("KVServer Cannot find data tensor with name: %s" % self.name)
         local_id = kv_store.part_policy[self.name].to_local(self.id_tensor)
-        data = kv_store.pull_handler(kv_store.data_store, self.name, local_id)
+        data = kv_store.pull_handlers[self.name](kv_store.data_store, self.name, local_id)
         res = PullResponse(kv_store.server_id, data)
         return res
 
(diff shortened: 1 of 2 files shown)


python/dgl/distributed/kvstore.py
```diff
@@ -60,7 +60,7 @@ def process_request(self, server_state):
         if kv_store.data_store.__contains__(self.name) is False:
             raise RuntimeError("KVServer Cannot find data tensor with name: %s" % self.name)
```
aksnzhy, line 61: Should we add a check: `kv_store.pull_handlers.__contains__(self.name) is False:`

python/dgl/distributed/kvstore.py
```diff
@@ -98,7 +98,8 @@ def process_request(self, server_state):
...
         local_id = kv_store.part_policy[self.name].to_local(self.id_tensor)
```
aksnzhy, line 100: Should we add a check:` kv_store.push_handlers.__contains__(self.name) is False:`

aksnzhy:
Only add some easy comments. I think it is ready to merge.

python/dgl/distributed/kvstore.py
zheng-da: whenever a data is initialized, a pull and push handler is registered. i think we don't need to check.

PR #1647, opened 2020-06-16
[KVStore] Add fast-pull for kvstore

aksnzhy:
## Description
Add fast-pull for kvstore.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/rpc.py
+++ b/python/dgl/distributed/rpc.py
@@ -14,7 +14,7 @@ __all__ = ['set_rank', 'get_rank', 'Request', 'Response', 'register_service', \
 'receiver_wait', 'add_receiver_addr', 'sender_connect', 'read_ip_config', \
 'get_num_machines', 'set_num_machines', 'get_machine_id', 'set_machine_id', \
 'send_request', 'recv_request', 'send_response', 'recv_response', 'remote_call', \
-'send_request_to_machine', 'remote_call_to_machine']
+'send_request_to_machine', 'remote_call_to_machine', 'fast_pull']
 
 REQUEST_CLASS_TO_SERVICE_ID = {}
 RESPONSE_CLASS_TO_SERVICE_ID = {}
@@ -843,6 +843,55 @@ def finalize_server():
     finalize_sender()
     finalize_receiver()
     print("Server (%d) shutdown." % get_rank())
+
+def fast_pull(name, id_tensor, part_id, service_id,
+              machine_count, group_count, machine_id,
+              client_id, local_data, policy):
+    """Fast-pull api used by kvstore.
+
+    Parameters
+    ----------
+    name : str
+        data name
+    id_tensor : tensor
+        data ID
+    part_id : tensor
+        partition ID of id_tensor
+    service_id : int
+        service_id of pull request
+    machine_count : int
+        total number of machine
+    group_count : int
+        total number of server inside machine
+    machine_id : int
+        current machine ID
+    client_id : int
+        current client ID
+    local_data : tensor
+        local data tensor
+    policy : PartitionPolicy
+        store the partition information
+    """
+    msg_seq = incr_msg_seq()
+    pickle_data = bytearray(pickle.dumps(([0], [name])))
+    global_id = _CAPI_DGLRPCGetLocalPartitionData(F.zerocopy_to_dgl_ndarray(id_tensor),
+                                                  F.zerocopy_to_dgl_ndarray(part_id),
+                                                  machine_id)
+    global_id = F.zerocopy_from_dgl_ndarray(global_id)
+    g2l_id = policy.to_local(global_id)
+    res_tensor = _CAPI_DGLRPCFastPull(name,
+                                      int(machine_id),
+                                      int(machine_count),
+                                      int(group_count),
+                                      int(client_id),
+                                      int(service_id),
+                                      int(msg_seq),
+                                      pickle_data,
+                                      F.zerocopy_to_dgl_ndarray(id_tensor),
+                                      F.zerocopy_to_dgl_ndarray(part_id),
+                                      F.zerocopy_to_dgl_ndarray(g2l_id),
+                                      F.zerocopy_to_dgl_ndarray(local_data))
+    return F.zerocopy_from_dgl_ndarray(res_tensor)
 
 ############### Some basic services will be defined here #############
 
--- a/python/dgl/distributed/rpc.py
+++ b/python/dgl/distributed/rpc.py
@@ -874,9 +874,9 @@ def fast_pull(name, id_tensor, part_id, service_id,
     """
     msg_seq = incr_msg_seq()
     pickle_data = bytearray(pickle.dumps(([0], [name])))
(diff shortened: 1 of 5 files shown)


python/dgl/distributed/rpc.py
```diff
@@ -844,6 +844,55 @@ def finalize_server():
...
+        current machine ID
+    client_id : int
+        current client ID
+    local_data : tensor
+        local data tensor
+    policy : PartitionPolicy
+        store the partition information
+    """
+    msg_seq = incr_msg_seq()
+    pickle_data = bytearray(pickle.dumps(([0], [name])))
```
zheng-da, line 876: what does 0 mean here?

python/dgl/distributed/rpc.py
```diff
@@ -844,6 +844,55 @@ def finalize_server():
...
+    global_id = _CAPI_DGLRPCGetLocalPartitionData(F.zerocopy_to_dgl_ndarray(id_tensor),
+                                                  F.zerocopy_to_dgl_ndarray(part_id),
+                                                  machine_id)
```
zheng-da, line 879: what does `global_id` mean here? i thought IDs in `id_tensor` are global.

src/rpc/rpc.cc
```diff
@@ -295,5 +299,129 @@ DGL_REGISTER_GLOBAL("distributed.server_state._CAPI_DGLRPCGetServerState")
...
+//////////////////////////// KVStore ////////////////////////////
+
+DGL_REGISTER_GLOBAL("distributed.rpc._CAPI_DGLRPCGetLocalPartitionData")
+.set_body([] (DGLArgs args, DGLRetValue* rv) {
+  NDArray ID = args[0];
+  NDArray part_id = args[1];
+  int local_machine_id = args[2];
+  int64_t* ID_data = static_cast<int64_t*>(ID->data);
+  int64_t* part_id_data = static_cast<int64_t*>(part_id->data);
+  int64_t ID_size = ID.GetSize() / sizeof(int64_t);
```
zheng-da, line 311: i think we need to check if the inputs are int64. we are going to support int32 for node/edge Ids.

src/rpc/rpc.cc
```diff
@@ -295,5 +299,129 @@ DGL_REGISTER_GLOBAL("distributed.server_state._CAPI_DGLRPCGetServerState")
...
+  std::string pickle_data = args[7];
+  NDArray ID = args[8];
+  NDArray part_id = args[9];
+  NDArray g2l_id = args[10];
+  NDArray local_data = args[11];
+  // Data
+  int64_t ID_size = ID.GetSize() / sizeof(int64_t);
+  int64_t* ID_data = static_cast<int64_t*>(ID->data);
+  int64_t* part_id_data = static_cast<int64_t*>(part_id->data);
+  int64_t* g2l_id_data = static_cast<int64_t*>(g2l_id->data);
```
zheng-da, line 341: btw, why not use `dgl_id_t`, instead of int64_t?

zheng-da: this `g2l_id_data` is very confusing. My understanding is that this is the global Ids of the nodes in the local partition.

src/rpc/rpc.cc
```diff
@@ -295,5 +299,129 @@ DGL_REGISTER_GLOBAL("distributed.server_state._CAPI_DGLRPCGetServerState")
...
+  for (int i = 0; i < remote_ids.size(); ++i) {
+    if (remote_ids[i].size() != 0) {
+      RPCMessage msg;
+      msg.service_id = service_id;
+      msg.msg_seq = msg_seq;
+      msg.client_id = client_id;
+      int lower = i*group_count;
+      int higher = (i+1)*group_count-1;
+#ifndef _WIN32  // windows does not support rand_r()
+      unsigned int seed = 314;
```
zheng-da, line 385: let's not use the random number generator in C. DGL has a random number generator.

src/rpc/rpc.cc
```diff
@@ -295,5 +299,129 @@ DGL_REGISTER_GLOBAL("distributed.server_state._CAPI_DGLRPCGetServerState")
...
+  NDArray res_tensor = NDArray::Empty(local_data_shape,
+                                      local_data->dtype,
+                                      DLContext{kDLCPU, 0});
+  char* return_data = static_cast<char*>(res_tensor->data);
+  // Copy local data
+#pragma omp parallel for
+  for (int64_t i = 0; i < local_ids.size(); ++i) {
+    memcpy(return_data + local_ids_orginal[i] * row_size,
+           local_data_char + local_ids[i] * row_size,
+           row_size);
```
zheng-da, line 408: you don't want to add checks like what i did last time? i found the check was very useful for me to debug the problems. there is no guarantee that users will provide the right Ids. If they do, this will result in segfault. Segfault is very difficult to debug for users.

zheng-da: we need to carefully check if the user inputs are correct. if not here, we need to check them somewhere else.

src/rpc/rpc.cc
```diff
@@ -295,5 +299,129 @@ DGL_REGISTER_GLOBAL("distributed.server_state._CAPI_DGLRPCGetServerState")
...
+  for (int i = 0; i < msg_count; ++i) {
+    RPCMessage msg;
+    RecvRPCMessage(&msg, 0);
+    int part_id = msg.server_id / group_count;
+    char* data_char = static_cast<char*>(msg.tensors[0]->data);
+    int64_t id_size = remote_ids[part_id].size();
+    for (size_t n = 0; n < id_size; ++n) {
+      memcpy(return_data + remote_ids_original[part_id][n] * row_size,
+             data_char + n * row_size,
+             row_size);
```
zheng-da, line 420: can we parallelize the data copy here as well?

python/dgl/distributed/rpc.py
aksnzhy: 0 means the data position. For the pull message, we only have a name metadata, so the position is fixed by 0.

src/rpc/rpc.cc
aksnzhy: I changed all the `int64_t` to `dgl_id_t`

src/rpc/rpc.cc
aksnzhy: My concern is that: this for loop is very time-consuming. If we add two checks here, it actually adds two if branch, I think it could bring the overhead because of the branch prediction fault of CPU processor.

src/rpc/rpc.cc
aksnzhy: We can try that. But I don't think it can improve the performance too much in real tasks. Because the remote-id size is very small compared to the local id.

src/rpc/rpc.cc
aksnzhy: I add back these checks. I will test the performance in our benchmark.

zheng-da: in this case, memory bandwidth is usually the bottleneck. i don't think adding some check will slow down much.

PR #1654, opened 2020-06-17
[Dataset] Basic template for DGL Dataset

classicsong:
## Description
Basic template for DGL Dataset

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [x] To the my best knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR

PR #1661, opened 2020-06-18
[Dataset] Update MiniGCDataset 

hetong007:
## Description
Following https://github.com/dmlc/dgl/pull/1654, I modified the simplest `MiniGCDataset`.

The tutorial is not affected by this update.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/data/minigc.py
+++ b/python/dgl/data/minigc.py
@@ -1,13 +1,17 @@
 """A mini synthetic dataset for graph classification benchmark."""
-import math
+import math, os
 import networkx as nx
 import numpy as np
 
+from .dgl_dataset import DGLDataset
+from .. import backend as F
+from .utils import save_graphs, load_graphs, makedirs
 from ..graph import DGLGraph
+from ..graph import batch as graph_batch
 
 __all__ = ['MiniGCDataset']
 
-class MiniGCDataset(object):
+class MiniGCDataset(DGLDataset):
     """The dataset class.
 
     The datset contains 8 different types of graphs.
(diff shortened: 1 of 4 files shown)


python/dgl/data/minigc.py
```diff
@@ -33,14 +37,18 @@ class MiniGCDataset(object):
     max_num_v: int
         Maximum number of nodes for graphs
     """
-    def __init__(self, num_graphs, min_num_v, max_num_v):
-        super(MiniGCDataset, self).__init__()
+    def __init__(self, num_graphs, min_num_v, max_num_v, verbose=False, seed=None):
```
classicsong, line 40: docstring for verbose and seed

python/dgl/data/minigc.py
```diff
@@ -61,12 +69,30 @@ def __getitem__(self, idx):
         """
         return self.graphs[idx], self.labels[idx]
 
+    def save(self):
+        """save the graph list and the labels"""
+        graph_path = os.path.join(self.raw_path, 'dgl_graph.bin')
+        # this check should be adeded into save_graphs
+        makedirs(self.raw_path)
```
classicsong, line 76: I suppose it is used in download. After calling download() self.raw_path should exist and contain raw data.

hetong007: Here are two things: 
1. Synthetic datasets (like this one) may not need to download from server.
2. `save_graphs`/`load_graphs` should have had the validation of the input path in python.

classicsong: > Here are two things:
> 
> 1. Synthetic datasets (like this one) may not need to download from server.
This can be a problem.

> 2. `save_graphs`/`load_graphs` should have had the validation of the input path in python.
Can we add a checkdir() other than makdir?

python/dgl/data/minigc.py
```diff
@@ -1,13 +1,17 @@
...
+from .dgl_dataset import DGLDataset
+from .. import backend as F
+from .utils import save_graphs, load_graphs, makedirs
 from ..graph import DGLGraph
+from ..graph import batch as graph_batch
 
 __all__ = ['MiniGCDataset']
 
-class MiniGCDataset(object):
+class MiniGCDataset(DGLDataset):
```
HuXiangkun, line 14: Do we need examples in docstring?

hetong007: Eventually we do. Maybe we copy the docstring in hackmd here later, or reversely.

python/dgl/data/minigc.py
hetong007: 1. We may need to walkthrough the current design to check if any variables require `download()` to initialize. 
2. The current `data.utils.makedir()` just works as check existance + `os.makedisr`.
3. `load_graphs` needs to check the existance of the file, but `save_graphs` needs to check the existance of the path of that file. E.g. given `/home/ubuntu/.dgl/minigc/graph.bin`, `save_graphs` need to check the validity of `/home/ubuntu/.dgl/minigc/`.

PR #1664, opened 2020-06-19
[Feature][Sampler] Sort CSR by tag

soodoshll:
## Description

To implement efficient biased sampler based on tag, we need to first sort CSR matrix by tag.
Currently, it only supports homograph and bipartite graph.

## Checklist
- [ ] Code is well-documented


src/array/array.cc
```diff
@@ -350,6 +350,16 @@ void CSRSort_(CSRMatrix* csr) {
   });
 }
 
+IdArray CSRSortByTag_(CSRMatrix* csr, IdArray tag, int64_t num_tags) {
+  IdArray ret;
```
jermainewang, line 354: Please add some sanity checks. E.g., the length of the tag array should be equal to the number of columns; csr and tag should have the same context; tag array must be integer etc. You can see other operators for reference.

src/array/cpu/spmat_op_impl.cc
```diff
@@ -586,6 +586,59 @@ void CSRSort_(CSRMatrix* csr) {
 template void CSRSort_<kDLCPU, int64_t>(CSRMatrix* csr);
 template void CSRSort_<kDLCPU, int32_t>(CSRMatrix* csr);
 
+template <DLDeviceType XPU, typename IdType, typename TagType>
+IdArray CSRSortByTag_(CSRMatrix* csr, IdArray tag, int64_t num_tags) {
+  auto indptr_data = static_cast<IdType *>(csr->indptr->data);
+  auto indices_data = static_cast<IdType *>(csr->indices->data);
+  auto eid_data = static_cast<IdType *>(csr->data->data);
```
jermainewang, line 593: It is possible that the csr does not have a data array when the data is simply [0, nnz). You can use `aten::CSRHasData` for check.

src/array/cpu/spmat_op_impl.cc
```diff
@@ -586,6 +586,59 @@ void CSRSort_(CSRMatrix* csr) {
...
+  auto tag_data = static_cast<TagType *>(tag->data);
+  int64_t num_rows = csr->num_rows;
+
+  NDArray split = NewIdArray(num_rows * (num_tags + 1), csr->indptr->ctx, csr->indptr->dtype.bits);
+  auto split_data = static_cast<IdType *>(split->data);
+
+  #pragma omp parallel for
```
jermainewang, line 600: no indent

src/array/cpu/spmat_op_impl.cc
```diff
@@ -586,6 +586,59 @@ void CSRSort_(CSRMatrix* csr) {
 template void CSRSort_<kDLCPU, int64_t>(CSRMatrix* csr);
 template void CSRSort_<kDLCPU, int32_t>(CSRMatrix* csr);
 
+template <DLDeviceType XPU, typename IdType, typename TagType>
+IdArray CSRSortByTag_(CSRMatrix* csr, IdArray tag, int64_t num_tags) {
+  auto indptr_data = static_cast<IdType *>(csr->indptr->data);
```
jermainewang, line 591: `const IdType*`

src/array/cpu/spmat_op_impl.cc
```diff
@@ -586,6 +586,59 @@ void CSRSort_(CSRMatrix* csr) {
...
+      split_data[src * (num_tags + 1) + t] = split_ptr;
+      std::copy(dst_arr[t].begin(), dst_arr[t].end(), indices_ptr);
+      std::copy(eid_arr[t].begin(), eid_arr[t].end(), eid_ptr);
+      indices_ptr += dst_arr[t].size();
+      eid_ptr += eid_arr[t].size();
+      split_ptr += dst_arr[t].size();
+    }
+    split_data[src * (num_tags + 1) + num_tags] = split_ptr;
+  }
+
```
jermainewang, line 629: Remember to set `csr->sorted` to false.

src/array/cpu/spmat_op_impl.cc
```diff
@@ -586,6 +586,59 @@ void CSRSort_(CSRMatrix* csr) {
 template void CSRSort_<kDLCPU, int64_t>(CSRMatrix* csr);
 template void CSRSort_<kDLCPU, int32_t>(CSRMatrix* csr);
 
+template <DLDeviceType XPU, typename IdType, typename TagType>
+IdArray CSRSortByTag_(CSRMatrix* csr, IdArray tag, int64_t num_tags) {
+  auto indptr_data = static_cast<IdType *>(csr->indptr->data);
+  auto indices_data = static_cast<IdType *>(csr->indices->data);
+  auto eid_data = static_cast<IdType *>(csr->data->data);
+  auto tag_data = static_cast<TagType *>(tag->data);
+  int64_t num_rows = csr->num_rows;
```
jermainewang, line 595: const

BarclayII:
A high-level question: should sort_csr and sort_csc, as well as their in-place variants, be exposed to the users?

jermainewang:
> A high-level question: should sort_csr and sort_csc, as well as their in-place variants, be exposed to the users? I feel it is dedicated to BiasedSampler.

M2C:

The first question is whether we can hide this inside BiasedSampler. The answer is yes. We can sort the graph in the __init__ function and pass the returned segment position to the sampling API.

Then why should we still expose this API to the end users? The reasons are two. First, the implementation of the BiasedSampler will be in PyThon so having an API docstring for this sorting routine makes it easier for user to understand what happens internally. Second, the functionality is general enough that might help other cases besides BiasedSampler. These sorting APIs are intended for the users with knowledge of (1) sparse storage format such as CSR/CSC and (2) why sorting the adjlist helps the performance, helping them to control the internal storage of a graph.

jermainewang:
@zheng-da Let's resume this effort. The original plan was to finish the biased neighbor sampler in three PRs.
* PR#1665 is for the C++ side implementation of a general biased sampler.
* PR#1664 (this PR) is to add a python-side API to sort each node's adjacency list based on node type. It will generate a tensor storing the segment starting offsets, which will be used for the biased neighbor sampler.
* A final PR (TBD) to put everything together into a biased neighbor sampler.

There are a couple of questions
* For this PR (sorting adjacency list by node type), there are two implementations:
  1. An in-place one using compare-based sort. It is more memory-friendly but with O(N logN) cost, where N is the largest neighborhood size.
  2. An out-place one using bucket sort. It generates a new sorted graph and the complexity is O(N).
  Our question is if we want to deliver one solution, which one is more useful?
* For the last PR, what kind of user-facing interface should we implement? We could certainly implement a `dgl.sample_neighbors` -like API. Is it enough? Shall we proceed to have a data loader style interface too?

python/dgl/transform.py
```diff
@@ -2482,4 +2484,111 @@ def as_immutable_graph(hg):
                 '\tdgl.as_immutable_graph will do nothing and can be removed safely in all cases.')
     return hg
 
+def sort_out_edges(g, tag=None, tag_pos="_TAG_POS"):
+    """A copy of the given graph whose out edges are sorted.
+    After sorting, edges whose destination shares the same
+    tag will be arranged in a consecutive range. Note that this
+    will not change the edge ID. It only changes the order in the
+    internal CSR storage. As such, the graph must allow CSR storage.
```
jermainewang, line 2492: Suggested change

```python
def sort_out_edges(g, tag, tag_offset_name='_TAG_OFFSET'):

"""Return a new graph which sorts the out edges of each node.

Sort the out edges according to the given destination node tags in integer.
A typical use case is to sort the edges by the destination node types, where
the tags represent destination node types. After sorting, edges sharing
the same tag will be arranged in a consecutive range in
a node's adjacency list. Following is an example:

    Consider a graph as follows:

    0 -> 0, 1, 2, 3, 4
    1 -> 0, 1, 2

    Given node tags [1, 1, 0, 2, 0], each node's adjacency list
    will be sorted as follows:

    0 -> 2, 4, 0, 1, 3
    1 -> 2, 0, 1

The function will also returns the starting offsets of the tag
segments in a tensor of shape `(N, max_tag+2)`. For node `i`,
its out-edges connecting to node tag `j` is stored between
`tag_offsets[i][j]` ~ `tag_offsets[i][j+1]`. Since the offsets
can be viewed node data, we store it in the
`ndata` of the returned graph. Users can specify the
ndata name by the `tag_pos_name` argument.

Note that the function will not change the edge ID neither
how the edge features are stored. The input graph must
allow CSR format. Graph must be on CPU.

If the input graph is heterogenous, it must have only one edge
type and two node types (i.e., source and destination node types).
In this case, the provided node tags are for the destination nodes,
and the tag offsets are stored in the source node data.

The sorted graph and the calculated tag offsets are needed by
certain operators that consider node tags. See xxx for an example.

Examples
-----------
...

Parameters
------------
g : DGLGraph
    The input graph.
tag : Tensor
    Integer tensor of shape `(N,)`, `N` being the number of (destination) nodes.
tag_offset_name : str
    The name of the node feature to store tag offsets.

Returns
-------
g_sorted : DGLGraph
    A new graph whose out edges are sorted. The node/edge features of the
    input graph is shallow-copied over.
    - `g_sorted.ndata[tag_offset_name]` : Tensor of shape `(N, max_tag + 2)`. If
       `g` is heterogeneous, get from `g_sorted.srcdata`.
"""
```

jermainewang: No need to support None tag at the moment. We could add it later when there is a request.

src/graph/heterograph_capi.cc
```diff
@@ -657,6 +657,56 @@ DGL_REGISTER_GLOBAL("transform._CAPI_DGLAsImmutableGraph")
     *rv = GraphRef(hg->AsImmutableGraph());
   });
 
+DGL_REGISTER_GLOBAL("transform._CAPI_DGLHeteroSortOutEdges")
+.set_body([] (DGLArgs args, DGLRetValue* rv) {
+    HeteroGraphRef hg = args[0];
+    NDArray tag = args[1];
+    int64_t num_tag = args[2];
+    const auto csr = hg->GetCSRMatrix(0);
```
jermainewang, line 665: Add check for device. Raise error if it is on GPU.

jermainewang (approved):

PR #1665, opened 2020-06-19
[Sampler] BiasedChoice sampler

soodoshll:
## Description

Fundamental functions for biased sampling. 

- Adapt tree sampler to enable removing an element from a group within which elements have the same probability.
- Implement BiasedChoice based on that.

## Checklist
- [ ] Code is well-documented


yzh119:
@soodoshll is this PR ready for review?

soodoshll:
> @soodoshll is this PR ready for review?

Yeah

jermainewang (changes requested):

include/dgl/random.h
```diff
@@ -178,6 +178,42 @@ class RandomEngine {
...
+   * \brief Pick random integers with different probability for different segments.
+   *
+   * If replace is false, num must not be larger than population.
+   *
+   * \tparam IdxType Return integer type
+   * \param num Number of integers to choose
+   * \param split Split positions of different segments
+   * \param bias Weights of different segments
+   * \param out The output buffer to write selected indices.
+   * \param replace If true, choose with replacement.
```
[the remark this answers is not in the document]
soodoshll: The `split` arrays of each node are compactly stored in a N*(T+1) array, where N is the number of nodes and T is the number of tags. And the input`split` is obtained by pointer displacement.

We need the `shape` attribute of `bias` to determine how many segments there are.

src/random/cpu/sample_utils.h
```diff
@@ -279,7 +279,18 @@ class TreeSampler: public BaseSampler<Idx> {
...
+   * The parameter decrease is an array of which the length is the number of categories.
+   * Every time an element in the category x is picked, the weight of this category is subtracted
+   * by decrease[x]. It is used to support the case where a category might contains multiple
+   * candidates and decrease[x] is the weight of one candidate of the category x.
+   *
+   * When decrease == nullptr, it means there is only one candidate in each category and will
+   * directly set the weight of the chosen category as 0.
+   *
+   */
+  Idx DrawAndUpdate(const DType *decrease = nullptr) {
```
jermainewang, line 293: why not just call it `Draw`? this interface should be compatible with the old one right?

jermainewang (approved):

src/random/cpu/sample_utils.h
soodoshll: We need to also modify the interface of `BaseSampler` to make the interface consistent with the super class. But `BaseSampler` does not have template parameter `DType`.

Issue #1629, opened 2020-06-12
[RFC] Revisit RGCN model with new sampler API


classicsong:
## 🚀 Feature
We are planing to revisit RGCN example code to support new sampler API, following features will be included:
- [x] Provide a low-mem version of RelGraphConv to save memory during training
- [x] Re-organize the builtin dataset for RGCN
- [x] RGCN node classification with new sampler API
- [ ] RGCN link prediction with new sampler API
- [ ] RGCN support DRKG dataset (https://github.com/gnn4dr/DRKG)
- [x] RGCN support node features 

We will implement RGCN using homogeneous graph API first.

## Motivation
New DGL Sampler API is released for a long time. We decide to reimplement RGCN model using new API interface. 

## Alternatives
We also need provide examples using heterogeneous graph API

PR #1673, opened 2020-06-22
[Feature]heterograph.edata and heterograph.ndata support multiple ntype and etype

classicsong:
## Description
When heterograph has multiple node types or edge types, it is not allowed to use g.ndata['h'] or g.edata['h'] to get data with key 'h'.

So it is impossible to do following to get a edge subgraph:
```
train_mask = g.edata['train_mask']
train_g = g.edge_subgraph(train_mask)
```

We need to do the following:
```
train_edges = {}
for etype in g.canonical_etypes:
    mask = g.edges[etype].data['train_mask']
    train_edges[etype] = th.arange(g.number_of_edges(etype))[mask]
train_g = g.edge_subgraph(train_edges)
```

Another benefit can be, for model like GCMC:
```
for ntype in enc_g.ntypes:
    feat= enc_g.nodes[ntype].ndata['h']
    dec_g.nodes[ntype].ndata['h'] = feat
```
can be written as:
```
dec_g.ndata['h'] = enc_g.ndata['h']
```

This PR will support g.ndata['h'] or g.edata['h'] return a dictionary of node/edge types with their corresponding data under key 'h'. For example:

```
g = dgl.graph([(0, 1), (1, 2), (1, 0)], 'user', 'like', 'movie')
g.ndata['h'] = {'user': torch.zeros(2, 5), 'movie': torch.zeros(3, 5)}
```

Related issue #1629 

## Checklist
- [x] Code is well-documented


python/dgl/heterograph.py
```diff
@@ -715,7 +715,10 @@ def dstnodes(self):
     def ndata(self):
         """Return the data view of all the nodes.
 
-        **Only works if the graph has one node type.**
+        If the graph has only one node type. The g.ndata directly represents the
+        node data.
+        If the graph has multiple node type. The g.ndata is a dictionary of node
+        types and their corresponding node data.
```
mufeili, line 721: How about "If the graph has multiple node types, then g.ndata[feat][node_type]" gives the features associated with key `feat` for `node_type`-typed nodes?

python/dgl/heterograph.py
```diff
@@ -727,24 +730,41 @@ def ndata(self):
...
     @property
     def srcdata(self):
         """Return the data view of all nodes in the SRC category.
 
         Only works if the graph is either
 
-        * Uni-bipartite and has one node type in the SRC category.
+        * Uni-bipartite
+
+        * Non-uni-bipartite (in this case identical to :any:`DGLHeteroGraph.ndata`)
```
mufeili, line 762: I don't think I fully get it. By non-uni-bipartite, do you mean an arbitrary graph other than a bipartite graph?

python/dgl/view.py
```diff
@@ -345,31 +384,72 @@ def __call__(self, *args, **kwargs):
...
+        if self._multi_etype:
+            assert isinstance(val, dict), \
+                'Current HeteroEdgeDataView has multiple edge types, ' \
+                'please passing the edge type and the corresponding data through a dict.'
+
+            for (etype, data) in val.items():
+                etid = self._graph.get_etype_id(etype)
+                self._graph._set_e_repr(etid, self._edges, {key : data})
+        else:
+            self._graph._set_e_repr(self._etid, self._edges, {key : val})
```
mufeili, line 420: I think we also need to check if val is a dictionary here and raise an error if so.

python/dgl/heterograph.py
classicsong: > How about "If the graph has multiple node types, then g.ndata[feat][node_type] gives the features associated with key `feat` for `node_type`-typed nodes?"

This one is better.

python/dgl/heterograph.py
classicsong: For Uni-bipartite, src and dst are disjoint.
For Non-uni-bipartite, src and dst can have some nodes.

But I think there is no need to highlight these two kinds of graphs.

BarclayII:
So we have the following behavior comparison:
| `ndata` behavior  | one node type | multiple node types |
|:-:|:--:|:--:|
| Previously | single tensor | error |
| This PR | single tensor | dict of ntype and tensor |

Correct?

BarclayII:
Also, since `g.edge_subgraph` takes in edge IDs or dict of edge IDs but `train_mask` is actually a boolean edge feature tensor, `g.edge_subgraph(train_mask)` wouldn't work by definition anyway.

BarclayII:
If you want to change the semantics of `edge_subgraph` to also support boolean masks, I would recommend also doing that in the PR.

classicsong:
> So we have the following behavior comparison:
> 
> `ndata` behavior	one node type	multiple node types
> Previously	single tensor	error
> This PR	single tensor	dict of ntype and tensor
> Correct?

Correct

python/dgl/view.py
```diff
@@ -282,33 +282,75 @@ def __call__(self, ntype=None):
...
+            return ret
+        else:
+            return self._graph._get_n_repr(self._ntid, self._nodes)[key]
 
     def __setitem__(self, key, val):
-        self._graph._set_n_repr(self._ntid, self._nodes, {key : val})
+        if self._multi_ntype:
+            assert isinstance(val, dict), \
+                'Current HeteroNodeDataView has multiple node types, ' \
+                'please passing the node type and the corresponding data through a dict.'
```
jermainewang, line 309: Can we allow this by setting the feature to all types.

python/dgl/view.py
```diff
@@ -282,33 +282,75 @@ def __call__(self, ntype=None):
...
+
+            for (ntype, data) in val.items():
+                ntid = self._graph.get_ntype_id(ntype)
+                self._graph._set_n_repr(ntid, self._nodes, {key : data})
+        else:
+            assert isinstance(val, dict) is False, \
+                'The HeteroNodeDataView has only one node type. ' \
+                'please pass a tensor directly'
```
jermainewang, line 317: The check is strange. I think you want to check `val` is of tensor type instead.

python/dgl/heterograph.py
```diff
@@ -727,24 +730,35 @@ def ndata(self):
...
-        ntype = self.ntypes[0]
-        return HeteroNodeDataView(self, ntype, ntid, ALL)
+        if len(self.ntypes) == 1:
+            ntid = self.get_ntype_id(None)
+            ntype = self.ntypes[0]
+            return HeteroNodeDataView(self, ntype, ntid, ALL)
+        else:
+            ntids = [self.get_ntype_id(ntype) for ntype in self.ntypes]
+            ntypes = self.ntypes
+            return HeteroNodeDataView(self, ntypes, ntids, ALL)
```
jermainewang, line 751: I feel the codes can be simplified. Here, you just do
```python
ntids = [self.get_ntype_id(ntype) for ntype in self.ntypes]
ntypes = self.ntypes
return HeteroNodeDataView(self, ntypes, ntids, ALL)
```
no matter what. And in `HeteroNodeDataView`, you retrieve the data using the `ntids` and feature name. Before you return the data, check if the dictionary contains only one item and return the value if so.

python/dgl/view.py
classicsong: Different types may have different shape. How can we know who is who in this case?

python/dgl/view.py
classicsong: This assert is to give hint to someone if he pass a dict to a graph with only one edge type.

python/dgl/heterograph.py
classicsong: > ```python
> HeteroNodeDataView
> ```

This maybe in another PR. A lots of place use HeteroNodeDataView.

python/dgl/view.py
jermainewang: So the following is not allowed?
```python
g = ...  # some homograph with node type 'user'
g.ndata['h'] = {'user': user_feat}
```

python/dgl/heterograph.py
jermainewang: ok

jermainewang:
LGTM other than one small comment.

python/dgl/view.py
classicsong: Yes.
For homograph, only use tensor, otherwise this may be confusion, when there is only one node type, one can use dict and also tensor.

PR #1674, opened 2020-06-22
[Dataset] KarateClubDataset

HuXiangkun:
## Description
KarateClubDataset inherits from DGLDataset. This dataset is transformed from networkx, which is hard-coded. So no need to download and save to local disk.

## Checklist
- [x] Code is well-documented


python/dgl/data/__init__.py
```diff
@@ -11,7 +11,7 @@
 from .ppi import PPIDataset, LegacyPPIDataset
 from .tu import TUDataset, LegacyTUDataset
 from .gnn_benckmark import AmazonCoBuy, CoraFull, Coauthor
-from .karate import KarateClub
+from .karate import KarateClubDataset
```
hetong007, line 14: Doesn't it break any example/tutorial?

HuXiangkun: I searched no examples or tutorials using KarateClub class.

hetong007: In case any third-party's using this dataset, I suggest we keep the old named with a warning of the deprecation.

VoVAllen:
Same opinion as @hetong007 . Other LGTM

PR #1676, opened 2020-06-22
[Model] Pinsage example that uses sparse embedding update

### LINKED ISSUE #1671: "CUDA out of memory error" for hidden_dimension 1024 in Pinsage example
## ❓ Questions and Help
I am trying to run the Pytorch experiment on the Pinsage model. For nowplaying_rs dataset, if I use hidden dimension 1024, "CUDA out of memory" error occurs. For smaller dimensions the code runs without error however, the accuracy seems very low. How to fix the error and any idea to improve the accuracy?

BarclayII:
## Description
In our PinSAGE implementation every item has a learnable embedding vector.  When hidden state size is large the embedding matrix cannot fit into GPU.  This PR offers an alternate solution that keeps the embedding matrix on CPU, optimizing it using `torch.optim.SparseAdam`.  It conserves GPU memory but the parameter update is much slower.

Fixes #1671 .


examples/pytorch/pinsage/model_sparse.py
```diff
@@ -0,0 +1,147 @@
...
+    dataloader_test = DataLoader(
+        torch.arange(g.number_of_nodes(item_ntype)),
+        batch_size=args.batch_size,
+        collate_fn=collator.collate_test,
+        num_workers=args.num_workers)
+    dataloader_it = iter(dataloader)
+
+    # Model
+    model = PinSAGEModel(g, item_ntype, textset, args.hidden_dims, args.num_layers).to(device)
+    item_emb = nn.Embedding(g.number_of_nodes(item_ntype), args.hidden_dims, sparse=True)
```
classicsong, line 88: Use
```
nn.init.uniform_(item_emb.weight, ...) to initialize the embedding?
```

examples/pytorch/pinsage/README.md
```diff
@@ -28,4 +28,16 @@ item embeddings, which are learned as outputs of PinSAGE.
 python model.py data.pkl --num-epochs 300 --num-workers 2 --device cuda:0 --hidden-dims 64
 ```
 
+The implementation here also assigns a learnable vector to each item.  If your hidden
+state size is so large that the learnable vectors cannot fit into GPU, use this script
+for sparse embedding update (written with `torch.optim.SparseAdam`) instead:
+
+
+```
+python model_sparse.py data.pkl --num-epochs 300 --num-workers 2 --device cuda:0 --hidden-dims 1024
```
classicsong, line 37: Does sparse embedding increase the speed?

HuXiangkun: Also, maybe it's better to list the performance of it.

BarclayII: > Does sparse embedding increase the speed?

No it decreases the speed.

PR #1687, opened 2020-06-23
[Kernel] Migrate batching/unbatching on adjlist to CSR/COO

classicsong:
## Description
Migrate batching/unbatching on adjlist to CSR/COO

## Checklist
- [x] Code is well-documented


include/dgl/aten/coo.h
```diff
@@ -309,6 +309,35 @@ COOMatrix COORowWiseTopk(
...
+ * \param src_offset A list of integers recording dst vertix id offset of each Matrix in coos
+ * \return The combined COOMatrix.
+ */
+COOMatrix DisjointUnionCooGraph(
+  const std::vector<COOMatrix>& coos,
+  const std::vector<uint64_t> src_offset,
+  const std::vector<uint64_t> dst_offset);
+
+/*!
+ * \brief Split a COOMatrix into multiple disjoin components.
```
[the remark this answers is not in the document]
jermainewang: Give an example too.

src/array/cpu/coo_union_partition.cc
```diff
@@ -0,0 +1,122 @@
...
+
+      if (has_data) {
+        if (data == nullptr) {
+          result_data[edge_offset[i] + j] = edge_offset[i] + j;
+        } else {
+          result_data[edge_offset[i] + j] = edge_offset[i] + data[j];
+        }
+      }
+    }
+  }
```
jermainewang, line 58: Suggest use array operators to rewrite this. Pseudo-codes:
```python
rows = []
cols = []
nrows = 0
ncols = 0
for coo in coos:
    rows.append(coo.row + nrows)
    cols.append(coo.col + ncols)
new_row = concat(rows)
new_cols = concat(cols)
...
```

This means the implementation is actually device- and idtype-agnostic and you can directly implement it in `array.cc` without creating separate implementations for CPU and GPU.

src/array/cpu/coo_union_partition.cc
```diff
@@ -0,0 +1,122 @@
...
+    COOMatrix sub_coo = COOMatrix(
+        src_vertex_cumsum[g+1]-src_vertex_cumsum[g],
+        dst_vertex_cumsum[g+1]-dst_vertex_cumsum[g],
+        VecToIdArray(result_src, sizeof(IdType) * 8),
+        VecToIdArray(result_dst, sizeof(IdType) * 8));
+    ret[g] = sub_coo;
+  }
+
+  return ret;
+}
```
jermainewang, line 104: Similar suggestions. This means you probably need to add one `Concat` and one `Split` operators to the aten namespace. Note that their implementations are also device-agnostic, by using `runtime::Device` API for memory copy.

src/array/union_partition.cc
```diff
@@ -0,0 +1,227 @@
...
+    } else {
+      IdArray result_indptr_0 = Full(0, 1,
+                                    csr.indptr->dtype.bits,
+                                    csr.indptr->ctx);
+      result_indptr = IndexSelect(csr.indptr,
+                                  Range(src_vertex_cumsum[g] + 1,
+                                        src_vertex_cumsum[g+1] + 1,
+                                        csr.indptr->dtype.bits,
+                                        csr.indptr->ctx)) - edge_cumsum[g];
+      result_indptr = Concat(std::vector<IdArray>({result_indptr_0, result_indptr}));
```
jermainewang, line 195: How about slice one more value to avoid prepending a zero element?
```c++
result_indptr = IndexSelect(csr.indptr, src_vertex_cumsum[g], src_vertx_cumsum[g+1] + 1) - edge_cumsum[g];
```
This also means the two if branches could probably be merged.

src/graph/heterograph_capi.cc
```diff
@@ -433,6 +433,40 @@ DGL_REGISTER_GLOBAL("heterograph_index._CAPI_DGLHeteroCopyTo")
...
+    CHECK(component_graphs.size() > 0)
+      << "Expect graph list has at least one graph";
+    std::vector<HeteroGraphPtr> component_ptrs;
+    component_ptrs.reserve(component_graphs.size());
+    const int64_t bits = component_graphs[0]->NumBits();
+    for (const auto& component : component_graphs) {
+      component_ptrs.push_back(component.sptr());
+      CHECK_EQ(component->NumBits(), bits)
+        << "Expect graphs to batch have the same index dtype(int" << bits
+        << "), but got int" << component->NumBits();
```
jermainewang, line 449: Also check the device type.

src/graph/transform/union_partition.cc
```diff
@@ -8,6 +8,202 @@ using namespace dgl::runtime;
...
+    uint64_t src_offset = 0, dst_offset = 0;
+    HeteroGraphPtr rgptr = nullptr;
+
+    // do some preprocess
+    for (size_t i = 0; i < component_graphs.size(); ++i) {
+      const auto& cg = component_graphs[i];
+      CHECK_EQ(format, cg->GetRelationGraph(etype)->GetFormatInUse()) << \
+        "All batched graph should have same sparse format";
+      CHECK_EQ(restrict_format, cg->GetRelationGraph(etype)->GetRestrictFormat()) << \
+        "All batched graph should have same restrict_format format";
```
jermainewang, line 33: That might be too strict requirements for the users. Since we are promoting opaque sparse formats, it would be hard (or tedious) for users specify formats for `dgl.batch`. How about we just directly pick one format that is available?

jermainewang: Another question is what is the restricted format for the result graph? I would prefer no restriction since that is the most common case. In other word, I feel the restricted format option should stick with one graph and should not propagate to other graphs via operations.

src/graph/transform/union_partition.cc
```diff
@@ -8,6 +8,202 @@ using namespace dgl::runtime;
...
+  for (uint64_t vtype = 0; vtype < num_vertex_types; ++vtype) {
+    vertex_cumsum[vtype].push_back(0);
+    for (uint64_t g = 0; g < batch_size; ++g) {
+      // We've flattened the number of vertices in the batch for all types
+      vertex_cumsum[vtype].push_back(
+        vertex_cumsum[vtype][g] + vertex_sizes_data[vtype * batch_size + g]);
+    }
+    CHECK_EQ(vertex_cumsum[vtype][batch_size], batched_graph->NumVertices(vtype))
+      << "Sum of the given sizes must equal to the number of nodes for type " << vtype;
+  }
```
jermainewang, line 111: Just a thought. I feel any operation on heterograph is hard to implement while on homograph is much easier. I wonder is it because we need to pass all the arguments to C++ and our FFI is limited? Is it possible to write operators only for homo/bipartite graphs and just have a common utility to call them on each relation graph?

tests/cpp/test_aten.cc
```diff
@@ -224,18 +224,40 @@ TEST(ArrayTest, TestIndexSelect) {
...
+  _TestConcat<int64_t>(CPU);
+  _TestConcat<float>(CPU);
+  _TestConcat<double>(CPU);
+#ifdef DGL_USE_CUDA
+  _TestConcat<int32_t>(GPU);
+  _TestConcat<int64_t>(GPU);
+  _TestConcat<float>(GPU);
+  _TestConcat<double>(GPU);
+#endif
+}
```
jermainewang, line 263: Could you also add C++ tests for `DisjointUnionCoo|CSr` and `DisjointPartitionCoo|Csr`? They cannot be easily tested from python side.

src/array/union_partition.cc
```diff
@@ -0,0 +1,227 @@
...
+ * \file array/cpu/coo_union_partition.cc
+ * \brief COO union and partition
+ */
+#include <dgl/array.h>
+#include <vector>
+
+namespace dgl {
+namespace aten {
+///////////////////////// COO Based Operations/////////////////////////
+COOMatrix DisjointUnionCooGraph(const std::vector<COOMatrix>& coos) {
```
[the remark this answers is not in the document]
classicsong: We need to check that.

src/array/union_partition.cc
```diff
@@ -0,0 +1,227 @@
...
+}
+
+std::vector<COOMatrix> DisjointPartitionCooBySizes(
+  const COOMatrix coo,
+  const uint64_t batch_size,
+  const std::vector<uint64_t> edge_cumsum,
+  const std::vector<uint64_t> src_vertex_cumsum,
+  const std::vector<uint64_t> dst_vertex_cumsum) {
+  CHECK(IsNullArray(coo.data)) <<
+        "DisjointPartitionHeteroBySizes does not support input COOMatrix with eid mapping data";
```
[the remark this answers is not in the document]
classicsong: This one should be deleted.

src/array/union_partition.cc
classicsong: Yes. we can do that, all will - edge_cumsum[g]

PR #1693, opened 2020-06-24
[Dataset] New GNN-Benchmark dataset

HuXiangkun:
## Description
- New gnn-benchmark dataset inherits from DGLDataset. It has GNNBenchmarkDataset as base class for each dataset.
- There are 3 datasets in it: CoraFullDataset, CoauthorDataset, AmazonCoBuyDataset
- Keep the original name for each dataset in gnn-benchmark and KarateClub dataset.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/data/__init__.py
+++ b/python/dgl/data/__init__.py
@@ -10,8 +10,8 @@ from .sbm import SBMMixture
 from .reddit import RedditDataset
 from .ppi import PPIDataset, LegacyPPIDataset
 from .tu import TUDataset, LegacyTUDataset
-from .gnn_benckmark import AmazonCoBuy, CoraFull, Coauthor
-from .karate import KarateClubDataset
+from .gnn_benckmark import *
+from .karate import *
 from .gindt import GINDataset
 from .bitcoinotc import BitcoinOTC
 from .gdelt import GDELT
--- a/python/dgl/data/__init__.py
+++ b/python/dgl/data/__init__.py
@@ -10,8 +10,9 @@ from .sbm import SBMMixture
 from .reddit import RedditDataset
 from .ppi import PPIDataset, LegacyPPIDataset
 from .tu import TUDataset, LegacyTUDataset
-from .gnn_benckmark import *
-from .karate import *
+from .gnn_benckmark import AmazonCoBuyComputerDataset, AmazonCoBuyPhotoDataset, CoauthorPhysicsDataset, \
+    CoauthorCSDataset,CoraFullDataset, AmazonCoBuy, Coauthor, CoraFull
+from .karate import KarateClubDataset, KarateClub
 from .gindt import GINDataset
 from .bitcoinotc import BitcoinOTC
 from .gdelt import GDELT
--- a/python/dgl/data/dgl_dataset.py
+++ b/python/dgl/data/dgl_dataset.py
@@ -36,10 +36,12 @@ class DGLDataset(object):
         if raw_dir is None:
             self._raw_dir = get_download_dir()
         else:
-            self._raw_dir = raw_dir
+            self._raw_dir = os.path.expanduser(raw_dir)
 
         if save_dir is None:
             self._save_dir = self._raw_dir
+        else:
+            self._save_dir = os.path.expanduser(save_dir)
 
         self._load()
 
@@ -89,7 +91,7 @@ class DGLDataset(object):
         self.download()
 
     def _load(self):
-        r"""Entry point from __init__ to load the dataset. 
+        r"""Entry point from __init__ to load the dataset.
             if the cache exists:
                 Load the dataset from saved dgl graph and information files.
             else:
@@ -103,7 +105,7 @@ class DGLDataset(object):
             self._download()
             self.process(self.raw_path)
             self.save()
-    
+
     @property
     def url(self):
         r"""Get url to download the raw dataset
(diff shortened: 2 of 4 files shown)


python/dgl/data/__init__.py
```diff
@@ -10,8 +10,8 @@
 from .reddit import RedditDataset
 from .ppi import PPIDataset, LegacyPPIDataset
 from .tu import TUDataset, LegacyTUDataset
-from .gnn_benckmark import AmazonCoBuy, CoraFull, Coauthor
-from .karate import KarateClubDataset
+from .gnn_benckmark import *
```
classicsong, line 13: Better to list the names, it is easier for maintainers to find the impl.

python/dgl/data/dgl_dataset.py
```diff
@@ -36,10 +36,12 @@ def __init__(self, name, url=None, raw_dir=None, save_dir=None, force_reload=Fal
         if raw_dir is None:
             self._raw_dir = get_download_dir()
         else:
-            self._raw_dir = raw_dir
+            self._raw_dir = os.path.expanduser(raw_dir)
```
classicsong, line 39: This changes behavior of raw_dir and makes a strong hidden assumption the dir is under ~.
But it can under /mnt or /data, etc.

```suggestion
            self._raw_dir = raw_dir
```
Let use take care of the path.

python/dgl/data/dgl_dataset.py
```diff
@@ -36,10 +36,12 @@ def __init__(self, name, url=None, raw_dir=None, save_dir=None, force_reload=Fal
...
 
         if save_dir is None:
             self._save_dir = self._raw_dir
+        else:
+            self._save_dir = os.path.expanduser(save_dir)
```
classicsong, line 44: Same here:
```suggestion
            self. _save_dir = save_dir
```

python/dgl/data/gnn_benckmark.py
```diff
@@ -16,24 +19,49 @@ def eliminate_self_loops(A):
...
     _url = {}
 
-    def __init__(self, name):
-        assert name.lower() in self._url, "Name not valid"
-        self.dir = get_download_dir()
-        self.path = os.path.join(
-            self.dir, 'gnn_benckmark', self._url[name.lower()].split('/')[-1])
-        self._name = name
-        g = self.load_npz(self.path)
+    def __init__(self, **kwargs):
```
classicsong, line 26: Can we use def __init__(self, name, url, raw_dir) to explicitly define the args

python/dgl/data/gnn_benckmark.py
```diff
@@ -16,24 +19,49 @@ def eliminate_self_loops(A):
...
+        name = kwargs.get('name', '').lower()
+        assert name in self._url, "Dataset name not valid"
+        url = self._url[name]
+        raw_dir = kwargs.get('raw_dir', None)
+        if raw_dir is None:
```
classicsong, line 31: You can use DGLBuiltinDataset as parent class, it has already set the raw_dir.

python/dgl/data/gnn_benckmark.py
```diff
@@ -101,19 +161,59 @@ class Coauthor(GNNBenchmarkDataset):
...
-class AmazonCoBuy(GNNBenchmarkDataset):
-    r"""
-    Amazon Computers and Amazon Photo are segments of the Amazon co-purchase graph [McAuley
-    et al., 2015], where nodes represent goods, edges indicate that two goods are frequently bought
-    together, node features are bag-of-words encoded product reviews, and class labels are given by the
-    product category.
+    @property
+    def num_classes(self):
+        """Number of classes."""
+        if self.name == 'cs':
```
classicsong, line 204: I think we should split this dataset into 2, one for cs and one for physics.

python/dgl/data/gnn_benckmark.py
```diff
@@ -122,8 +222,67 @@ class AmazonCoBuy(GNNBenchmarkDataset):
...
+    >>> num_class = data.num_classes
+    >>> feat = g.ndata['feat']  # get node feature
+    >>> label = g.ndata['label']  # get node labels
     """
     _url = {
         'computers': "https://github.com/shchur/gnn-benchmark/raw/master/data/npz/amazon_electronics_computers.npz",
         'photo': "https://github.com/shchur/gnn-benchmark/raw/master/data/npz/amazon_electronics_photo.npz"
     }
+
+    def __init__(self, name, **kwargs):
```
classicsong, line 258: We can split for this dataset also.

python/dgl/data/gnn_benckmark.py
```diff
@@ -16,24 +20,40 @@ def eliminate_self_loops(A):
...
+        return False
+
+    def save(self):
+        graph_path = os.path.join(self.save_path, 'dgl_graph.bin')
+        save_graphs(graph_path, self.graph)
 
-    @retry_method_with_fix(_download)
-    def load_npz(self, file_name):
+    def load(self):
+        graph_path = os.path.join(self.save_path, 'dgl_graph.bin')
```
VoVAllen, line 46: Sorry for saying this late. Can we put the save_path and the filenames into the class attribute instead of instance attribute? (Just like _url). This makes it easy to identify which files are used for cache/saving. And for the root_path or other paths, are they standardized across all DGL datasets?

PR #1700, opened 2020-06-25
[Distributed] Optimize distributed sampling

zheng-da:
## Description
The server and the client can be co-located on the same machine. In this case, we should optimize sampling from the local machine. Instead of sending sampling request to the local server, we can directly sample from the local partition shared with the local server with shared memory.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/rpc.py
+++ b/python/dgl/distributed/rpc.py
@@ -752,7 +752,7 @@ def send_requests_to_machine(target_and_requests):
 
 def recv_responses(num_res, msgseq2pos, timeout=0):
     myrank = get_rank()
-    all_res = [None] * len(target_and_requests)
+    all_res = [None] * num_res
     while num_res != 0:
         # recv response
         msg = recv_rpc_message(timeout)
--- a/python/dgl/distributed/rpc.py
+++ b/python/dgl/distributed/rpc.py
@@ -731,6 +731,44 @@ def remote_call(target_and_requests, timeout=0):
         all_res[msgseq2pos[msg.msg_seq]] = res
     return all_res
 
+def send_requests_to_machine(target_and_requests):
+    msgseq2pos = {}
+    num_res = 0
+    for pos, (target, request) in enumerate(target_and_requests):
+        # send request
+        service_id = request.service_id
+        msg_seq = incr_msg_seq()
+        client_id = get_rank()
+        server_id = target
+        data, tensors = serialize_to_payload(request)
+        msg = RPCMessage(service_id, msg_seq, client_id, server_id, data, tensors)
+        send_rpc_message(msg, server_id)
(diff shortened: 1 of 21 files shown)


python/dgl/distributed/sampling.py
```diff
@@ -46,52 +68,113 @@ def __getstate__(self):
...
+    # send requests to the remote machine.
+    num_res = 0
+    msgseq2pos = None
+    if len(req_list) > 0:
+        num_res, msgseq2pos = send_requests_to_machine(req_list)
+
+    # sample neighbors for the nodes in the local partition.
+    res_list = []
+    if local_nids is not None:
+        src, dst, eids = _sample_neighbors(dist_graph.local_partition, partition_book,
```
classicsong, line 169: Local sample can only work with single process?

zheng-da: dgl.distributed.sample_neighbors will be called in multiple processes.

zheng-da:
@VoVAllen i'm not sure what you mean. `_sample_neighbors` is just a wrapper. it does the global2local and local2global mapping automatically. i just don't want to write the same code for the client and server.

python/dgl/distributed/rpc.py
```diff
@@ -777,6 +748,11 @@ def remote_call_to_machine(target_and_requests, timeout=0):
         if res_cls is not None:
             num_res += 1
             msgseq2pos[msg_seq] = pos
+    return num_res, msgseq2pos
+
+def recv_responses(num_res, msgseq2pos, timeout=0):
+    myrank = get_rank()
+    all_res = [None] * num_res
```
zheng-da, line 755: is this right? @aksnzhy

aksnzhy: I think here is no problem.

aksnzhy:
> My only question is that whether `_sample_neighbors` function is needed. Directly call `SampleRequest.process_request` on the caller machine should be fine. The only problem is how to pass the same `state` into it, if as the design before. Therefore a function of `res = invoke_on_local(request)` is probably what we want? Correct me if I'm wrong

No you mean that we should put the local request into the rpc level?

python/dgl/distributed/rpc.py
aksnzhy: Can we re-use `recv_response` and `send_request` instead of re-implement all the logic in `recv_responses` and `send_requests`?

zheng-da: i didn't reimplement these. the code of `recv_responses` and `send_requests` are copied from `remote_call_to_machine`. if we could reuse `recv_response` and `send_request`, we could reuse them in `remote_call_to_machine`.

PR #1704, opened 2020-06-26
[CUDA][Kernel] More CUDA kernels; Standardize the behavior for sorted COO/CSR

jermainewang:
## Description

Added NVIDIA/CUB as one of the dependencies. New array operators on GPU:

* aten::CumSum
* aten::CSRSliceRows
* Added operator << for NDArray. You can now print out the content of an ndarray easily from C++.
* aten::CSRIsSorted
* aten::COOIsSorted
* Inplace sort: aten::CSRSort_ and aten::COOSort_
* Outplace sort: aten::CSRSort and aten::COOSort
* More efficient implementation for aten::COOToCSR when the input COO is already sorted.

**DGLGraph now by default uses sorted CSR.** Per our offline discussion, the overhead of sorting a CSR is ok as long as (1) the conversion cost is one-time (i.e., in most full graph training scenarios) or (2) the conversion is efficient when the input COO is sorted (i.e., in most mini-batch training scenarios).

## Checklist
- [x] Code is well-documented


CMakeLists.txt
```diff
@@ -44,6 +44,8 @@ include_directories("third_party/METIS/include/")
 include_directories("third_party/dmlc-core/include")
 include_directories("third_party/minigun/minigun")
 include_directories("third_party/minigun/third_party/moderngpu/src")
+include_directories("third_party/cub/")
+include_directories("third_party/phmap/")
```
zheng-da, line 48: we didn't include phmap before?

include/dgl/aten/coo.h
```diff
@@ -195,6 +220,20 @@ bool COOHasDuplicate(COOMatrix coo);
...
 std::pair<COOMatrix, IdArray> COOCoalesce(COOMatrix coo);
 
+/*!
+ * \brief Sort the indices of a COO matrix in-place.
+ *
+ * The function sorts row indices in ascending order. If sort_column is true,
+ * col indices are sorted in ascending order too. The data array of the returned COOMatrix
+ * stores the shuffled index which could be used to fetch edge data.
+ *
+ * Complexity: O(N*log(N)), where N is the number of nonzeros.
```
zheng-da, line 230: The complexity should be lower than this.

src/array/array.cc
```diff
@@ -110,12 +114,30 @@ ValueType IndexSelect(NDArray array, uint64_t index) {
...
-template int32_t IndexSelect<int32_t>(NDArray array, uint64_t index);
-template int64_t IndexSelect<int64_t>(NDArray array, uint64_t index);
-template uint32_t IndexSelect<uint32_t>(NDArray array, uint64_t index);
-template uint64_t IndexSelect<uint64_t>(NDArray array, uint64_t index);
-template float IndexSelect<float>(NDArray array, uint64_t index);
-template double IndexSelect<double>(NDArray array, uint64_t index);
+template int32_t IndexSelect<int32_t>(NDArray array, int64_t index);
+template int64_t IndexSelect<int64_t>(NDArray array, int64_t index);
+template uint32_t IndexSelect<uint32_t>(NDArray array, int64_t index);
+template uint64_t IndexSelect<uint64_t>(NDArray array, int64_t index);
```
zheng-da, line 120: do we need to support so many?

src/array/cpu/array_cumsum.cc
```diff
@@ -0,0 +1,42 @@
...
+ * \brief Array cumsum CPU implementation
+ */
+#include <dgl/array.h>
+
+namespace dgl {
+using runtime::NDArray;
+namespace aten {
+namespace impl {
+
+template <DLDeviceType XPU, typename IdType>
```
zheng-da, line 13: do we support GPU operations on tensors yet?

src/array/cpu/spmat_op_impl_coo.cc
```diff
@@ -266,29 +266,57 @@ CSRMatrix COOToCSR(COOMatrix coo) {
...
+  if (row_sorted) {
+    // compute indptr
+    IdType* Bp = static_cast<IdType*>(ret_indptr->data);
+    Bp[0] = 0;
+    int64_t j = 0;
+    for (int64_t i = 0; i < N; ++i) {
+      const int64_t k = j;
+      for (; j < NNZ && row_data[j] == i; ++j) {}
+      Bp[i + 1] = Bp[i] + j - k;
+    }
```
zheng-da, line 291: is this branch tested?

src/array/cpu/spmat_op_impl_coo.cc
```diff
@@ -266,29 +266,57 @@ CSRMatrix COOToCSR(COOMatrix coo) {
...
   }
 
-  // cumsum
-  for (int64_t i = 0, cumsum = 0; i < N; ++i) {
-    const IdType temp = Bp[i];
-    Bp[i] = cumsum;
-    cumsum += temp;
-  }
-  Bp[N] = NNZ;
+  if (row_sorted) {
```
zheng-da, line 282: i don't understand here. if col_sorted is true, we also don't need to go through the complex computation in the else branch. it seems to me that this case isn't optimized.

src/array/cuda/coo2csr.cc
```diff
@@ -17,63 +17,43 @@ template <DLDeviceType XPU, typename IdType>
...
-        row->shape[0],
+        coo.row.Ptr<int32_t>(),
+        nnz,
         coo.num_rows,
         indptr_ptr,
         CUSPARSE_INDEX_BASE_ZERO));
 
   return CSRMatrix(coo.num_rows, coo.num_cols,
-                   indptr, col, data, false);
+                   indptr, coo.col, coo.data, col_sorted);
```
zheng-da, line 56: if CSR is always sorted, do we still need this sorted flag?

zheng-da:
did you check the speed in min-batch training?

zheng-da:
i skipped the GPU code. i'm just wondering if there are packages that have already contain all operations implemented in this PR.

CMakeLists.txt
jermainewang: Previous code uses relative include path like `include "../../third_party/phmap/parallel_hashmap/..."`. With this change, the include path is now a system one so you could use `include <parallel_hashmap/...>`.

src/array/array.cc
jermainewang: you are right, uint32_t and uint64_t are not needed.

src/array/cuda/coo2csr.cc
jermainewang: we do if the CSR is given (or generated) by some external sources (or other operators)

src/array/cpu/spmat_op_impl_coo.cc
jermainewang: I c. One thing I didn't make it clear is that the `col_sorted` flag only takes effect when the `row_sorted` is true. The `col_sorted` means each column **per row** is also sorted. Should I change another name?

tests/compute/test_heterograph.py
```diff
@@ -1741,7 +1741,7 @@ def test_format():
     # test_adj("int32")
     # test_inc()
     # test_view()
-    # test_view1("int32")
+    test_view1("int32")
```
classicsong, line 1744: comment it?

tests/cpp/common.h
```diff
@@ -29,6 +29,9 @@ inline int64_t Len(dgl::runtime::NDArray nd) {
 template <typename T>
 inline bool ArrayEQ(dgl::runtime::NDArray a1, dgl::runtime::NDArray a2) {
   if (a1->ndim != a2->ndim) return false;
+  if (a1->dtype != a2->dtype) return false;
+  if (a1->ctx != a2->ctx) return false;
+  if (a1.NumElements() == 0) return true;
```
classicsong, line 34: Also need to check a2.NumElements()

tests/compute/test_heterograph.py
jermainewang: It doesn't matter since pytest just look for `test_xxx` functions.

src/array/cpu/spmat_op_impl_coo.cc
```diff
@@ -266,29 +266,57 @@ CSRMatrix COOToCSR(COOMatrix coo) {
...
-
-  std::fill(Bp, Bp + N, 0);
-  for (int64_t i = 0; i < NNZ; ++i) {
-    Bp[row_data[i]]++;
+  bool row_sorted = coo.row_sorted;
+  bool col_sorted = coo.col_sorted;
+  if (!row_sorted) {
+    // It is possible that the flag is simply not set (default value is false),
+    // so we still perform a linear scan to check the flag.
+    std::tie(row_sorted, col_sorted) = COOIsSorted(coo);
```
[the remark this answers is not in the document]
jermainewang: It won't help because the input is a value-passed argument.

include/dgl/aten/coo.h
```diff
@@ -161,6 +172,20 @@ COOMatrix COOTranspose(COOMatrix coo);
  * the result CSR matrix stores a shuffle index for how the entries
  * will be reordered in CSR. The i^th entry in the result CSR corresponds
  * to the CSR.data[i] th entry in the input COO.
+ *
+ * Conversion complexity: O(nnz)
+ *
+ * - The function first check whether the input COO matrix is sorted
```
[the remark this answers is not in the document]
jermainewang: We shouldn't check. The flags are just hints for optimization. The operator should handle both the sorted and unsorted versions.

include/dgl/aten/csr.h
```diff
@@ -218,6 +240,20 @@ bool CSRHasDuplicate(CSRMatrix csr);
...
+/*!
+ * \brief Sort the column index at each row in ascending order.
+ *
+ * Return a new CSR matrix with sorted column indices and data arrays.
+ */
+inline CSRMatrix CSRSort(CSRMatrix csr) {
+  CSRMatrix ret(csr.num_rows, csr.num_cols,
+                csr.indptr, csr.indices.Clone(),
+                CSRHasData(csr)? csr.data.Clone() : csr.data,
+                csr.sorted);
```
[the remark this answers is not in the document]
jermainewang: done.

include/dgl/aten/coo.h
jermainewang: It should be O(N). On GPU, we use cusparse's sort API so it should be fine. On CPU, we currently use an NlogN sort. Since COOSort on CPU is not used in any critical path, I left a TODO for future improvement.

jermainewang:
> did you check the speed in min-batch training?

Using the train_sampling example under `dgl/examples/pytorch/graphsage/`.

On p3:
* PR: 19.2628 sec/epoch
* PR w/o CSRSort: 18.5343 sec/epoch
* master: 20.9780 sec/epoch

On x1:
* PR:  36.7229 sec/epoch
* PR w/o CSRSort: 36.2034 sec/epoch
* master: 37.6212 sec/epoch

src/array/cpu/array_cumsum.cc
```diff
@@ -0,0 +1,42 @@
...
+      out_d[i + 1] = out_d[i] + in_d[i];
+    return ret;
+  } else {
+    IdArray ret = aten::NewIdArray(len, array->ctx, array->dtype.bits);
+    const IdType* in_d = array.Ptr<IdType>();
+    IdType* out_d = ret.Ptr<IdType>();
+    out_d[0] = in_d[0];
+    for (int64_t i = 1; i < len; ++i)
+      out_d[i] = out_d[i - 1] + in_d[i];
+    return ret;
```
BarclayII, line 33: Just curious.  When to or not to use `std::accumulate`?

src/array/cpu/csr_sort.cc
```diff
@@ -0,0 +1,84 @@
...
+  const IdType* indices = csr.indices.Ptr<IdType>();
+  bool ret = true;
+#pragma omp parallel for shared(ret)
+  for (int64_t row = 0; row < csr.num_rows; ++row) {
+    if (!ret)
+      continue;
+    for (IdType i = indptr[row] + 1; i < indptr[row + 1]; ++i) {
+      if (indices[i - 1] > indices[i]) {
+#pragma omp critical
+        ret = false;
```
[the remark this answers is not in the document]
BarclayII: Would this introduce atomic operations?

src/array/cuda/coo_sort.cu
```diff
@@ -0,0 +1,155 @@
...
+using runtime::NDArray;
+
+namespace aten {
+namespace impl {
+
+///////////////////////////// COOSort_ /////////////////////////////
+
+template <DLDeviceType XPU, typename IdType>
+void COOSort_(COOMatrix* coo, bool sort_column) {
+  CHECK(sizeof(IdType) == 4) << "CUDA COOSort does not support int64.";
```
BarclayII, line 21: Do we plan to support int64 COOSort in the future?  If no, why?

jermainewang: Currently we use cusparse which only supports int32_t. We should support that in the future using the RadixSort routine from CUB. I'll leave a TODO there.

src/array/cpu/array_cumsum.cc
jermainewang: TIL this new function. However, it is only available in C++20. Currently, we still stick with C++11.

PR #1710, opened 2020-06-27
[Distributed] Split data to ensure data locality for multiple clients

zheng-da:
## Description
The previous version can only works for 1 client per machine. This one splits the workset further for multiple clients in a machine. It's common to have multiple clients on one machine. e.g., we run one client for each GPU or for each NUMA node.

## Checklist
- [ ] Code is well-documented


python/dgl/distributed/dist_graph.py
```diff
@@ -490,7 +497,12 @@ def rank(self):
         int
             The rank of the current graph store.
         '''
-        return self._client.client_id
+        if self._gpb is None:
+            return self._client.client_id
+        else:
+            num_client_per_part = self._num_client // self._gpb.num_partitions()
+            client_id_in_part = self._client.client_id % num_client_per_part
```
zheng-da, line 504: @aksnzhy how is the client_id decided in kvstore client?

aksnzhy: The code is right. We assume that the client id is grouped together inside a machine.

python/dgl/distributed/dist_graph.py
```diff
@@ -494,10 +495,20 @@ def rank(self):
...
         if self._g is None:
-            return self._client.client_id
+            return rpc.get_rank()
         else:
-            return self._gpb.partid
+            # If DistGraph has a local partition, we should be careful about the rank
+            # we return. We need to return a rank that node_split or edge_split can split
+            # the workload with respect to data locality.
+            num_client = rpc.get_num_client()
+            num_client_per_part = num_client // self._gpb.num_partitions()
```
aksnzhy, line 508: Do we have the case that the number of clients can be different on each machine?

python/dgl/distributed/dist_graph.py
```diff
@@ -558,7 +569,26 @@ def _get_overlap(mask_arr, ids):
         masks = F.gather_row(mask_arr.tousertensor(), ids)
         return F.boolean_mask(ids, masks)
 
-def node_split(nodes, partition_book, rank):
+def _split_local(partition_book, rank, elements, local_eles):
+    ''' Split the input element list with respect to data locality.
+    '''
+    num_clients = rpc.get_num_client()
+    num_client_per_part = num_clients // partition_book.num_partitions()
```
aksnzhy, line 576: same

python/dgl/distributed/dist_graph.py
aksnzhy: If we will meet this case in the future, we can add a new API `get_num_client_per_machine(machine_id)`

zheng-da: it sounds a very complex scenario. why do we want to run different numbers of clients on different machines?

PR #1725, opened 2020-06-30
[Kernel] Use omp flush in CSRIsSorted

classicsong:
## Description
Win CI will crash sometime on CSRIsSorted. This PR try to fix it.

## Checklist
- [ ] Code is well-documented


BarclayII:
LGTM.  The issue itself is quite strange since [one of the official VS tutorials](https://docs.microsoft.com/en-us/cpp/parallel/concrt/convert-an-openmp-loop-that-uses-cancellation?view=vs-2019) has very similar OpenMP code.

classicsong:
> LGTM. The issue itself is quite strange since [one of the official VS tutorials](https://docs.microsoft.com/en-us/cpp/parallel/concrt/convert-an-openmp-loop-that-uses-cancellation?view=vs-2019) has very similar OpenMP code.

Did you encountered CI failure due to CSRIsSorted? I suppose it is a sync problem for omp shared variable.

BarclayII:
> > LGTM. The issue itself is quite strange since [one of the official VS tutorials](https://docs.microsoft.com/en-us/cpp/parallel/concrt/convert-an-openmp-loop-that-uses-cancellation?view=vs-2019) has very similar OpenMP code.
> 
> Did you encountered CI failure due to CSRIsSorted? I suppose it is a sync problem for omp shared variable.

I did.  This is very strange as [writing the *same* value of a *primitive* type to the same memory location should work fine even if we have races](https://stackoverflow.com/questions/8315931/does-writing-the-same-value-to-the-same-memory-location-cause-a-data-race).

classicsong:
The problem is related to omp code compiling: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92672
Skip this problem by removing omp

PR #1744, opened 2020-07-03
[Kernel] Line graph with new Kernel

classicsong:
## Description
Line graph using new unigraph kernel

## Checklist
- [x] Code is well-documented


include/dgl/aten/coo.h
```diff
@@ -457,6 +457,35 @@ std::vector<COOMatrix> DisjointPartitionCooBySizes(
   const std::vector<uint64_t> &src_vertex_cumsum,
   const std::vector<uint64_t> &dst_vertex_cumsum);
 
+/*!
+ * \brief Create a LineGraph of input coo
+ * 
+ * A = [[0, 0, 1],
+ *      [1, 0, 1],
+ *      [1, 1, 0]]
+ *
```
jermainewang, line 466: The example should include the data field so we know which edge corresponds to which node in the line graph.

include/dgl/aten/coo.h
```diff
@@ -457,6 +457,35 @@ std::vector<COOMatrix> DisjointPartitionCooBySizes(
...
+ * B = COOLineGraph(A, backtracking=False)
+ *
+ * B = [[0, 0, 0, 0, 1],
+ *      [1, 0, 0, 0, 0],
+ *      [0, 0, 0, 1, 0],
+ *      [0, 0, 0, 0, 0],
+ *      [0, 1, 0, 0, 0]]
+ *
+ * C = COOLineGraph(A, backtracking=False)
```
jermainewang, line 475: `backtracking=True`?

include/dgl/aten/csr.h
```diff
@@ -452,6 +452,35 @@ std::vector<CSRMatrix> DisjointPartitionCsrBySizes(
...
+ *      [0, 0, 0, 0, 0],
+ *      [0, 1, 0, 0, 0]]
+ *
+ * C = CSRLineGraph(A, backtracking=False)
+ *
+ * C = [[0, 0, 0, 1, 1],
+ *      [1, 0, 0, 0, 0],
+ *      [0, 0, 0, 1, 1],
+ *      [1, 0, 0, 0, 0],
+ *      [0, 1, 1, 0, 0]]
```
jermainewang, line 476: same comment here

include/dgl/base_heterograph.h
```diff
@@ -465,6 +465,12 @@ class BaseHeteroGraph : public runtime::Object {
     return nullptr;
   }
 
+  /*! \brief Creat a LineGraph of self */
+  virtual HeteroGraphPtr LineGraph(bool backtracking) const {
+    LOG(FATAL) << "LineGraph not supported.";
+    return nullptr;
+  }
```
jermainewang, line 472: Could you make it as a global function instead of a member function? Overall, I will try make every graph transformation API as functional as possible.

python/dgl/heterograph.py
```diff
@@ -2576,6 +2576,55 @@ def incidence_matrix(self, typestr, ctx=F.cpu(), etype=None):
     # Alias of ``incidence_matrix``
     inc = incidence_matrix
 
+    def line_graph(self, backtracking=True):
```
jermainewang, line 2579: same thing here. I would prefer to write a global function.
```
lg = dgl.line_graph(g)
```
And if we want to also have a syntax as member function. We could have something like this:
```python
def line_graph(g, ...):
    # implementation

DGLGraph.line_graph = line_graph
```

python/dgl/heterograph.py
```diff
@@ -2576,6 +2576,55 @@ def incidence_matrix(self, typestr, ctx=F.cpu(), etype=None):
...
+        ... ndata_schemes={}
+        ... edata_schemes={})
+        >>> lg.edges()
+        ... (tensor([0, 1, 2, 4]), tensor([4, 0, 3, 1]))
+
+        """
+        assert self._graph.number_of_etypes() == 1, \
+          'line_graph only support directed homogeneous graph right now'
+        assert self._graph.number_of_ntypes() == 1, \
+          'line_graph only support directed homogeneous graph right now'
```
jermainewang, line 2622: Add a `g.is_homo()` and check that.

src/array/cpu/coo_linegraph.cc
```diff
@@ -0,0 +1,53 @@
...
+    IdType u = coo_row[i];
+    IdType v = coo_col[i];
+    for (int64_t j = 0; j < nnz; ++j) {
+      // no self-loop
+      if (i == j)
+        continue;
+
+      // succ_u == v
+      // if not backtracking succ_u != u
+      if (v == coo_row[j] && (backtracking || (!backtracking && u != coo_col[j]))) {
```
jermainewang, line 35: the `!backtracking` looks redundant

src/array/cpu/csr_linegraph.cc
```diff
@@ -0,0 +1,60 @@
...
+      IdType u = i;
+      IdType v = indices_data[j];
+
+      for (int64_t k = 0; k < csr.num_rows; ++k) {
+        for (int64_t l = indptr_data[k]; l < indptr_data[k+1]; ++l) {
+          if (k == i && j == l)
+            continue;
+
+          // succ_u == v
+          if (v == k && (backtracking || (!backtracking && u != indices_data[l]))) {
```
jermainewang, line 37: same here

src/array/cpu/csr_linegraph.cc
```diff
@@ -0,0 +1,60 @@
...
+namespace dgl {
+namespace aten {
+namespace impl {
+
+template <DLDeviceType XPU, typename IdType>
+CSRMatrix CSRLineGraph(const CSRMatrix &csr, bool backtracking) {
+  const int64_t nnz = csr.indices->shape[0];
+  const IdType *indptr_data = static_cast<IdType *>(csr.indptr->data);
+  const IdType *indices_data = static_cast<IdType *>(csr.indices->data);
+  std::vector<IdType> new_indptr = {0};
```
jermainewang, line 22: reserve space of size `nnz + 1`

src/array/cpu/csr_linegraph.cc
```diff
@@ -0,0 +1,60 @@
...
+#include <algorithm>
+#include <vector>
+#include <iterator>
+
+namespace dgl {
+namespace aten {
+namespace impl {
+
+template <DLDeviceType XPU, typename IdType>
+CSRMatrix CSRLineGraph(const CSRMatrix &csr, bool backtracking) {
```
jermainewang, line 18: One problem. The node order in the line graph is the same as the edge order in the original graph but this implementation does not keep that order. I think one quick implementation is just convert it to COO, create a line graph and then convert to CSR again. Since the operation is quite rare, the performance is not critical. We can also prefer COO format.

src/graph/unit_graph.cc
```diff
@@ -1494,6 +1494,23 @@ GraphPtr UnitGraph::AsImmutableGraph() const {
...
+  if (coo_->defined()) {
+    const auto&new_coo = aten::COOLineGraph(coo_->adj(), backtracking);
+    return CreateFromCOO(1, new_coo, restrict_format_);
+  } else if (in_csr_->defined()) {
+    const auto&new_in_csr = aten::CSRLineGraph(in_csr_->adj(), backtracking);
+    return CreateFromCSC(1, new_in_csr, restrict_format_);
+  } else if (out_csr_->defined()) {
+    const auto&new_out_csr = aten::CSRLineGraph(out_csr_->adj(), backtracking);
+    return CreateFromCSR(1, new_out_csr, restrict_format_);
+  }
```
jermainewang, line 1508: You should use the `SelectFormat` function for this.

include/dgl/base_heterograph.h
classicsong: Sure.

python/dgl/heterograph.py
classicsong: Move to transform.py

src/array/cpu/csr_linegraph.cc
classicsong: One problem of this solution is if the format is restricted to CSR? converting to COO will case extra memory cost.

include/dgl/aten/coo.h
classicsong: Yes

src/array/cpu/csr_linegraph.cc
classicsong: If backtracing is False, The space is not nnz+1.

src/array/cpu/csr_linegraph.cc
classicsong: I think we can do this. Otherwise the CSR implementation is very complex.

src/graph/unit_graph.cc
```diff
@@ -1494,6 +1494,31 @@ GraphPtr UnitGraph::AsImmutableGraph() const {
   return GraphPtr(new dgl::ImmutableGraph(in_csr_ptr, out_csr_ptr, coo_ptr));
 }
 
+HeteroGraphPtr UnitGraph::LineGraph(bool backtracking) const {
+  // TODO(xiangsx) currently we only support homogeneous graph
+  auto fmt = SelectFormat(SparseFormat::kAny);
+  switch (fmt) {
+    case SparseFormat::kCOO: {
+      return CreateFromCOO(1, aten::COOLineGraph(coo_->adj(), backtracking), restrict_format_);
```
jermainewang, line 1502: No need to propagate the restricted format flag. Just set them to any.

src/graph/unit_graph.cc
```diff
@@ -1494,6 +1494,31 @@ GraphPtr UnitGraph::AsImmutableGraph() const {
...
+    }
+    case SparseFormat::kCSR: {
+      const aten::CSRMatrix csr = GetCSRMatrix(0);
+      const aten::COOMatrix coo = aten::COOLineGraph(aten::CSRToCOO(csr, false), backtracking);
```
jermainewang, line 1506: The data_as_order flag should be set to true right?

python/dgl/transform.py
```diff
@@ -162,6 +163,53 @@ def line_graph(g, backtracking=True, shared=False):
     node_frame = g._edge_frame if shared else None
     return DGLGraph(graph_data, node_frame)
 
+def line_heterograph(g, backtracking=True):
```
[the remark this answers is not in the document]
classicsong: I do not want to break the current code. There are multiple place line_graph is called.
We can merge these two later.

src/graph/unit_graph.cc
classicsong: Changed it into SparseFormat::kAny

src/graph/unit_graph.cc
classicsong: Done

PR #1747, opened 2020-07-03
[Dataset] RDF Dataset

classicsong:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/data/minigc.py
+++ b/python/dgl/data/minigc.py
@@ -40,12 +40,12 @@ class MiniGCDataset(DGLDataset):
     seed : int, default is None
         Random seed for data generation
     """
-    def __init__(self, num_graphs, min_num_v, max_num_v, seed=None, **kwargs):
+    def __init__(self, num_graphs, min_num_v, max_num_v, seed=None):
         self.num_graphs = num_graphs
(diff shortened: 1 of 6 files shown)


python/dgl/data/minigc.py
```diff
@@ -76,8 +78,9 @@ def __getitem__(self, idx):
 
     def save(self):
         """save the graph list and the labels"""
-        graph_path = os.path.join(self.save_path, 'dgl_graph.bin')
-        save_graphs(str(graph_path), self.graphs, {'labels': self.labels})
+        if self.save_graph :
+            graph_path = os.path.join(self.save_path, 'dgl_graph.bin')
+            save_graphs(str(graph_path), self.graphs, {'labels': self.labels})
```
HuXiangkun, line 83: As has_cache() is not implemented, it always return False, so each time we call use this class is actually creating new dataset and save() and load() is not used, right?

HuXiangkun: As this is a synthetic dataset, we need to check whether the saved dataset match the given parameters. If not new dataset need to be generated.

HuXiangkun:
Others LGTM

PR #1750, opened 2020-07-04
[Distributed] fix init_data in kvstore.

zheng-da:
## Description
The previous implementation issues init_data request from one client. This doesn't work because each client only has the partition size of its own partition. We should have one client from each machine to issue the request.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/kvstore.py
+++ b/python/dgl/distributed/kvstore.py
@@ -871,7 +871,7 @@ class KVClient(object):
                                       part_policy.policy_str,
                                       init_func)
             for n in range(self._group_count):
-                server_id = machine_id * self._group_count + n
+                server_id = part_policy.part_id * self._group_count + n
                 rpc.send_request(server_id, request)
             for _ in range(self._group_count):
                 response = rpc.recv_response()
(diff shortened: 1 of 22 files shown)


python/dgl/distributed/kvstore.py
```diff
@@ -851,57 +851,42 @@ def init_data(self, name, shape, dtype, policy_str, partition_book, init_func):
...
-                                          tuple(part_shape),
-                                          F.reverse_data_type_dict[dtype],
-                                          policy_str,
-                                          init_func)
-                for n in range(self._group_count):
-                    server_id = machine_id * self._group_count + n
-                    rpc.send_request(server_id, request)
-            for _ in range(self._server_count):
+        # One of the clients in each machine will issue requests to the local server.
+        if self._client_id % part_policy.partition_book.num_partitions() == 0:
```
aksnzhy, line 865: we can use:` if self._client_id % self._machine_count`

zheng-da: it should be the same

PR #1759, opened 2020-07-07
[Distributed] Distributed multiprocess dataloader

VoVAllen:
## Checklist
- [ ] Code is well-documented


zheng-da:
please move files to the right folder.

python/dgl/distributed/dist_dataloader.py
```diff
@@ -0,0 +1,111 @@
...
+        self.recv_idxs = 0
+        self.pool = ctx.Pool(
+            num_workers, initializer=init_fn, initargs=(collate_fn, self.queue))
+        for _ in range(queue_size):
+            self._request_next_batch()
+
+    def __next__(self):
+        self._request_next_batch()
+        if self.recv_idxs < self.expected_idxs:
+            result = self.queue.get(timeout=DGL_QUEUE_TIMEOUT)
```
zheng-da, line 74: what does this timeout mean? if it time out, what is result?

python/dgl/distributed/dist_dataloader.py
```diff
@@ -0,0 +1,111 @@
...
+    def __iter__(self):
+        return self
+
+    def _request_next_batch(self):
+        next_data = self._next_data()
+        if next_data is None:
+            return None
+        else:
+            async_result = self.pool.apply_async(
+                call_collate_fn, args=(next_data, ))
```
zheng-da, line 89: i suppose this is to call the collate function in remote processes?
I'm trying to understand how communication happens here. do we need to pickle `call_collate_fn` and `next_data` every time?
my understanding is that the input collate function is pickled when the process pool is created.

tests/distributed/test_mp_dataloader.py
```diff
@@ -0,0 +1,114 @@
...
+        print(idx)
+    
+    dataloader.close()
+
+    dgl.distributed.shutdown_servers()
+    dgl.distributed.finalize_client()
+
+
+
+def main(tmpdir, num_server):
```
zheng-da, line 78: why don't you call it `test_dist_dataloader`?

tests/distributed/test_mp_dataloader.py
```diff
@@ -0,0 +1,114 @@
...
+        p.start()
+        time.sleep(1)
+        pserver_list.append(p)
+
+    time.sleep(3)
+    sampled_graph = start_client(0, tmpdir, num_server > 1, num_workers)
+    for p in pserver_list:
+        p.join()
+
+def test_dist_dataloader(tmp_dir):
```
zheng-da, line 108: can this run with nosetest?

tests/distributed/test_mp_dataloader.py
```diff
@@ -0,0 +1,114 @@
...
+        collate_fn=sampler.sample_blocks,
+        # shuffle=True,
+        drop_last=False,
+        num_workers=4) 
+    
+    dist_graph._init()
+
+    for idx, block in enumerate(dataloader):
+        print(block)
+        print(idx)
```
zheng-da, line 69: to verify the correctness of the sampling results here?

python/dgl/distributed/dist_dataloader.py
```diff
@@ -0,0 +1,111 @@
...
+            result = self.queue.get(timeout=DGL_QUEUE_TIMEOUT)
+            self.recv_idxs += 1
+            return result
+        else:
+            raise StopIteration
+
+    def __iter__(self):
+        return self
+
+    def _request_next_batch(self):
```
zheng-da, line 83: for better performance, i think you should issue multiple requests at a time. otherwise, we may a lot of overheads for synchronizations. we need to keep waking up the sampling processes.

zheng-da:
we need to get some benchmark results to compare with single-process sampling, multiprocessing sampling with pytorch dataloader and multiprocessing sampling with this dataloader.

python/dgl/distributed/dist_dataloader.py
VoVAllen: It will raise an error if doesn't get any data from queue until timeout. Usually this would not happen. Sometime maybe server/client failed which made the returned results less than expected, which could trigger this exception

python/dgl/distributed/dist_dataloader.py
VoVAllen: Your understanding is right. Obviously we need to pickle `next_data` every time since they are different. For the collate functions, it is only pickled once at the initialization process.

tests/distributed/test_mp_dataloader.py
VoVAllen: Yes. Pytest will inject a temp directory into it.

tests/distributed/test_mp_dataloader.py
VoVAllen: Because I want this file works like an example, therefore I keep the main function

python/dgl/distributed/dist_dataloader.py
zheng-da: do we need the timeout?

PR #1763, opened 2020-07-07
[bugfix] ReverseOp returns a heterograph with wrong meta graph information 

yzh119:
## Description
In the implementation of ReverseOp, we return a new heterograph which holds the same metagraph as the original heterograph, which may cause troubles, e.g.
```python
>>> import dgl
>>> g = dgl.rand_bipartite(30, 40, 100)
>>> gidx = g._graph
>>> gidx.metagraph.find_edge(0)
(0, 1)
>>> revg_idx = gidx.reverse()
>>> revg.metagraph.find_edge(0)
(0, 1)
```
Note that we depend on metagraph to find src/dst node types, if we do not reverse metagraph accordingly, we will not get correct `number_of_src_nodes`, `number_of_dst_nodes` etc.

This PR proposes a workaround (though I think it's ugly) to solve this problem, however, I have some questions:
1. Does the meta_graph field in `coo_`, `in_csr_`, `out_csr_` make sense? If not, why not removing them?
2. Note that we do not expose unitgraph c pointer to users, so I don't think the metagraph field in unitgraph matters either, and any changes on `unitgraph.cc` do not affect the program's behavior actually. What really matters is my changes on `heterograph_capi.cc`.
3. Can we guarantee that all metagraphs are stored in COO format?

@classicsong , @jermainewang , @BarclayII do you have any suggestions?

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/src/graph/heterograph_capi.cc
+++ b/src/graph/heterograph_capi.cc
@@ -5,6 +5,7 @@
  */
 #include <dgl/array.h>
 #include <dgl/packed_func_ext.h>
+#include <dgl/immutable_graph.h>
 #include <dgl/runtime/container.h>
 
 #include "../c_api_common.h"
@@ -612,6 +613,10 @@ DGL_REGISTER_GLOBAL("heterograph_index._CAPI_DGLHeteroReverse")
     }
     // node types are not changed
     const auto& num_nodes = g->NumVerticesPerType();
-    *rv = CreateHeteroGraph(hg->meta_graph(), rev_ugs, num_nodes);
+    const auto& meta_edges = hg->meta_graph()->Edges("eid");
+    const auto& rev_meta = ImmutableGraph::CreateFromCOO(
+                                                         hg->meta_graph()->NumVertices(),
+                                                         meta_edges.dst, meta_edges.src);
+    *rv = CreateHeteroGraph(rev_meta, rev_ugs, num_nodes);
   });
 }  // namespace dgl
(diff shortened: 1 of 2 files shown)


src/graph/heterograph_capi.cc
```diff
@@ -612,6 +613,10 @@ DGL_REGISTER_GLOBAL("heterograph_index._CAPI_DGLHeteroReverse")
     }
     // node types are not changed
     const auto& num_nodes = g->NumVerticesPerType();
-    *rv = CreateHeteroGraph(hg->meta_graph(), rev_ugs, num_nodes);
+    const auto& meta_edges = hg->meta_graph()->Edges("eid");
+    const auto& rev_meta = ImmutableGraph::CreateFromCOO(
+                                                         hg->meta_graph()->NumVertices(),
+                                                         meta_edges.dst, meta_edges.src);
+    *rv = CreateHeteroGraph(rev_meta, rev_ugs, num_nodes);
```
jermainewang, line 620: The metagraph does not matter and we will probably delete them later. The correct fix should be:
```c++
const auto& num_nodes = g->NumVerticesPerType();
if (num_nodes.size() == 1)
  *rv = CreateHeteroGraph(hg->meta_graph(), rev_ugs, num_nodes);
else
  *rv = CreateHeteroGraph(hg->meta_graph(), rev_ugs, {num_nodes[1], num_nodes[0]});
```

BarclayII: Why metagraph doesn't matter?  The edges in the metagraph should also be reversed, right?

yzh119: I think @jermainewang 's proposal works for bipartite graph, however for a general heterograph with multiple etypes ( > 2) we cannot simply reverse the number of nodes array.

yzh119:
closed because of #1784

PR #1765, opened 2020-07-08
[Dataset] GINDataset

hetong007:
- Update `GINDataset` with `DGLDataset` class.
- Remove the input argument `self.raw_path` in `process()` call.

### DIFF
--- a/python/dgl/data/gindt.py
+++ b/python/dgl/data/gindt.py
@@ -12,8 +12,8 @@ import numpy as np
 
 from .. import backend as F
 
-from .dgl_dataset import DGLBuiltinDataset
-from .utils import loadtxt, save_graphs, load_graphs, save_info, load_info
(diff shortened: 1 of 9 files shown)


classicsong:
I think you should merge with dgl-dataset first to make it pass the CI.

python/dgl/data/gindt.py
```diff
@@ -78,10 +86,20 @@ def __init__(self, name, self_loop, degree_as_nlabel=False):
         self.degree_as_nlabel = degree_as_nlabel
         self.nattrs_flag = False
         self.nlabels_flag = False
-        self.verbosity = False
 
-        # calc all values
-        self._load()
+        super(GINDataset, self).__init__(name=name, url=gin_url,
```
classicsong, line 90: Derived from Builtin dataset: DGLBuiltinDataset

PR #1766, opened 2020-07-08
Handle EINTR in TCP layer

mszarma:
Added handling EINTR signal by
retrying when returned value and
errno are related to EINTR interrupt
error which my appear during tcp
function execution.
Added information about EAGAIN timeout
failure in accept stage.
Added printing out errno in logs.

## Checklist
- [ ] Code is well-documented

## Changes
Added handling of EINTR in tcp_socket functions.


src/rpc/network/tcp_socket.cc
```diff
@@ -164,11 +177,23 @@ void TCPSocket::Close() {
 }
 
 int64_t TCPSocket::Send(const char * data, int64_t len_data) {
-  return send(socket_, data, len_data, 0);
+  int64_t number_send;
+
+  do {  // retry if EINTR failure appears
+    number_send = send(socket_, data, len_data, 0);
+  } while (number_send == -1 && errno == EINTR);
```
aksnzhy, line 184: If the server is crashed during the runing. Will the send() run in the loop forever?

zheng-da: it shouldn't be. in this case, the error isn't EINTR. EINTR means the syscall is interrupted by signals.

PR #1772, opened 2020-07-09
[Distributed] Distributed launching script

aksnzhy:
## Description

Usage:

1. Configuring SSH
We need password-less ssh access between the master node and the worker nodes. If users already have this in place, they can skip this section.

2. Start server and client:

```
python3 ../launch.py --workspace ~/mytask --exe_file dist_train.py --graph-name ogb-product \
--num-client 4 --conf_path data/ogb-product.json --ip_config ip_config.txt --nproc_per_node=1 --nnodes=4 \
--master_addr="172.31.40.126" --master_port=1234 \
'--num-epochs 3 --batch-size 1000 --lr 0.1'
```

User can use 'ctl+c' to exit the launching.

## Checklist
- [ ] Code is well-documented


tools/launch.py
```diff
@@ -0,0 +1,135 @@
...
+    # launch client tasks
+    for i in range(args.num_client):
+        ip, _ = hosts[int(i / client_count_per_machine)]
+        cmd = client_cmd.replace('node_rank=0', 'node_rank='+str(i))
+        cmd = 'cd ' + str(args.workspace) + '; ' + cmd
+        execute_remote(cmd, ip)
+
+    # use ctl+C to exit
+    while True:
+        time.sleep(10)
```
zheng-da, line 64: this is weird. why do you need the users to type ctrl+c to exit? can the program exit when the training job completes?

tools/launch.py
```diff
@@ -0,0 +1,135 @@
...
+
+def server_command(args):
+    server_cmd = 'python3 ' + str(args.exe_file) + ' --server'
+    server_cmd = server_cmd + ' --graph-name ' + str(args.graph_name)
+    server_cmd = server_cmd + ' --num-client ' + str(args.num_client)
+    server_cmd = server_cmd + ' --conf_path ' + str(args.conf_path)
+    server_cmd = server_cmd + ' --ip_config ' + str(args.ip_config)
+    return server_cmd
```
zheng-da, line 72: we shouldn't ask the users to write code for servers. we can set dgl into a server mode and run services automatically.
mxnet starts service like this. you can use environment variables to set dgl into the server mode.

zheng-da:
can we kill all remote jobs after the program receives ctrl+c?

tools/launch.py
```diff
@@ -0,0 +1,135 @@
...
+    client_cmd = client_cmd + ' --graph-name ' + str(args.graph_name)
+    client_cmd = client_cmd + ' --ip_config ' + str(args.ip_config)
+    client_cmd = client_cmd + ' ' + udf_args
+    return client_cmd 
+
+def main():
+    parser = argparse.ArgumentParser(description='Launch a distributed job')
+    ########## common args ####################
+    parser.add_argument('--kill',  action='store_true',
+                        help='kill all the python processes on cluster')
```
zheng-da, line 91: do we need this? ideally, when a user types ctrl+c, the remote processes should also be killed.

tools/launch.py
```diff
@@ -0,0 +1,135 @@
...
+    parser.add_argument('--workspace', type=str,
+                        help='Path of user workspace')
```
zheng-da, line 93: i don't think this is necessary.

tools/launch.py
```diff
@@ -0,0 +1,135 @@
...
+    parser.add_argument('--ip_config', type=str, default='ip_config.txt',
+                        help='The file for IP configuration')
+    parser.add_argument('--exe_file', type=str,
+                        help="Python executable file")
```
zheng-da, line 97: we should follow the launch script of mxnet. it should be something like this.
```bash
python3 [arguments] trainer_script.py [...]
```

tools/launch.py
```diff
@@ -0,0 +1,135 @@
...
+    parser.add_argument('--server', action='store_true',
+                        help='wether this is a server')
```
zheng-da, line 99: the launch script should runs the various services automatically. users don't need to invoke start services in their training script or provide this argument to turn on the server mode.

tools/launch.py
```diff
@@ -0,0 +1,135 @@
...
+                        help='Number of processes per node')
+    parser.add_argument('--nnodes', type=int, 
+                        help='Total number of nodes')
+    parser.add_argument('--master_addr', type=str, 
+                        help='IP address of master node')
+    parser.add_argument('--master_port', type=int, default=1250,
+                        help='Port of master node')
+    ####### User-defined args ###################
+    parser.add_argument('--udf_args', type=str,
+                        help='user-defined arguments.')
```
zheng-da, line 117: we don't need this.

tools/launch.py
```diff
@@ -0,0 +1,135 @@
...
+                        help='The path to the partition config file')
+    ####### Pytorch distributed ################
+    parser.add_argument('--nproc_per_node', type=int, default=1,
+                        help='Number of processes per node')
+    parser.add_argument('--nnodes', type=int, 
+                        help='Total number of nodes')
+    parser.add_argument('--master_addr', type=str, 
+                        help='IP address of master node')
+    parser.add_argument('--master_port', type=int, default=1250,
+                        help='Port of master node')
```
zheng-da, line 114: we don't need to provide `master_add` or `master_port`. we can determine them from ip_config

tools/launch.py
```diff
@@ -0,0 +1,135 @@
...
+                        help='graph name')
+    parser.add_argument('--num-client', type=int, 
+                        help='Total number of clients on cluster')
+    parser.add_argument('--conf_path', type=str, 
+                        help='The path to the partition config file')
+    ####### Pytorch distributed ################
+    parser.add_argument('--nproc_per_node', type=int, default=1,
+                        help='Number of processes per node')
+    parser.add_argument('--nnodes', type=int, 
+                        help='Total number of nodes')
```
zheng-da, line 110: we don't need users to specify this. Given the number of clients and the number of machines, we can determine them.

tools/launch.py
```diff
@@ -0,0 +1,135 @@
...
+    parser.add_argument('--ip_config', type=str, default='ip_config.txt',
+                        help='The file for IP configuration')
+    parser.add_argument('--exe_file', type=str,
+                        help="Python executable file")
+    parser.add_argument('--server', action='store_true',
+                        help='wether this is a server')
+    parser.add_argument('--graph-name', type=str, 
+                        help='graph name')
+    parser.add_argument('--num-client', type=int, 
+                        help='Total number of clients on cluster')
```
zheng-da, line 103: i think we need to provide `--num-servers`. we want to run multiple servers on one machine.

tools/launch.py
aksnzhy: Users will not access the server code. This is just for our scheduler.

tools/launch.py
aksnzhy: This is used for killing the zombie processes.

tools/launch.py
aksnzhy: Yes. This argument actually can removed.

tools/launch.py
aksnzhy: The `--num-servers` can be provided by` ip_config.txt`

tools/launch.py
aksnzhy: Sure.

tools/launch.py
zheng-da: if you want to have a thread to monitor the results of other processes, i think you can wait for all other threads to terminate, instead of waiting forever. we would like the program to exit after all training jobs are complete.

tools/launch.py
```diff
@@ -0,0 +1,97 @@
...
+                        the contents of current directory will be rsyncd')
+    parser.add_argument('--num_client', type=int, 
+                        help='Total number of client processes in the cluster')
+    parser.add_argument('--conf_path', type=str, 
+                        help='The path to the partition config file. This path can be \
+                        a remote path like s3 and dgl will download this file automatically')
+    parser.add_argument('--ip_config', type=str, 
+                        help='The file for IP configuration for server processes')
+    parser.add_argument('--udf_command', type=str,
+                        help='User-defined command line')
```
zheng-da, line 83: do we need this argument?

zheng-da:
image

zheng-da:
we should also set the environment variables.

PR #1784, opened 2020-07-10
[Transform] Reverse for heterogenous graph

classicsong:
## Description
Reverse for heterogenous graph

#1762 

## Checklist
- [x] Code is well-documented


python/dgl/transform.py
```diff
@@ -383,6 +384,107 @@ def reverse(g, share_ndata=False, share_edata=False):
         g_reversed._edge_frame = g._edge_frame
     return g_reversed
 
+def reverse_heterograph(g, share_ndata=True, share_edata=False):
```
mufeili, line 387: Why don't we directly change `dgl.reverse`?

python/dgl/heterograph.py
```diff
@@ -4751,6 +4791,8 @@ def is_unibipartite(graph):
         True if the graph is a uni-bipartite.
     """
     src, dst, _ = graph.edges()
+    print(src.tonumpy())
+    print(dst.tonumpy())
```
mufeili, line 4795: I assume these two lines are for debugging and should be removed?

python/dgl/transform.py
```diff
@@ -383,6 +384,107 @@ def reverse(g, share_ndata=False, share_edata=False):
...
+    >>> rg.edges(etype='plays')
+    (tensor([2, 1, 1]), tensor([1, 2, 1]))
+    >>> rg.nodes['game'].data['hv
+    tensor([[1.],
+            [1.],
+            [1.]])
+    >>> rg.edges['plays'].data
+    {}
+    """
+    return g.reverse(share_ndata=share_ndata, share_edata=share_edata)
```
mufeili, line 485: Can we move the implementation from `heterograph.py` to here?

python/dgl/heterograph.py
classicsong: Yes.

python/dgl/transform.py
classicsong: Theoretically yes, but as we are using self._graph, self._edge_frames. These are supposed to be internal functions?

python/dgl/heterograph.py
```diff
@@ -4701,6 +4701,46 @@ def int(self):
                               self._node_frames,
                               self._edge_frames)
 
+    def reverse(self, share_ndata=True, share_edata=False):
```
mufeili, line 4704: One possibility is to implement it directly using `dgl.heterograph` so that we can avoid accessing private properties of the graph.

tests/compute/test_heterograph.py
```diff
@@ -2092,6 +2153,6 @@ def test_reverse(index_dtype):
     # test_isolated_ntype()
     # test_bipartite()
     # test_dtype_cast()
-    # test_reverse("int32")
+    test_reverse("int32")
```
mufeili, line 2156: Copy @BarclayII and @VoVAllen for visibility. I think we don't need to pass the data type?

python/dgl/heterograph.py
```diff
@@ -4701,6 +4701,46 @@ def int(self):
...
+        r""" Same as dgl.reverse_heterograph
```
mufeili, line 4705: Let's replace `dgl.reverse_heterograph` by `:func:`dgl.reverse_heterograph`` for redirection to the doc of dgl.reverse_heterograph.

tests/compute/test_heterograph.py
classicsong: For CI, it will automatically test int32 and int64. For offline testing, we need to specify the type.

python/dgl/transform.py
classicsong: For compatablity.

mufeili: Then, why don't we create a transform branch and make the PR into dmlc:transform?

python/dgl/heterograph_index.py
```diff
@@ -991,7 +991,7 @@ def to_format(self, restrict_format):
         """
         return _CAPI_DGLHeteroGetFormatGraph(self, restrict_format)
 
-    def reverse(self):
+    def reverse(self, metagraph):
         """Reverse the heterogeneous graph adjacency
 
         The node types and edge types are not changed
```
mufeili, line 997: Add parameter description

python/dgl/heterograph.py
```diff
@@ -4701,6 +4701,46 @@ def int(self):
                               self._node_frames,
                               self._edge_frames)
 
+    def reverse(self, share_ndata=True, share_edata=False):
+        r""":func:dgl.reverse_heterograph
+        """
+
+        # TODO(0.5 release, xiangsx) need to handle BLOCK
+        # currently reversing a block results in undfined behavior
```
mufeili, line 4709: "undfined" should be "undefined"?

python/dgl/transform.py
classicsong: I move it into transform.py now

python/dgl/transform.py
```diff
@@ -383,6 +384,137 @@ def reverse(g, share_ndata=False, share_edata=False):
...
+    if share_ndata:
+        # for each ntype
+        for ntype in g.ntypes:
+            # for each data field
+            for k in g.nodes[ntype].data:
+                new_g.nodes[ntype].data[k] = g.nodes[ntype].data[k]
+
+    # handle edata
+    if share_edata:
+        # for each ntype
```
mufeili, line 509: Should be "# for each etype"

python/dgl/transform.py
```diff
@@ -383,6 +384,137 @@ def reverse(g, share_ndata=False, share_edata=False):
...
+    r"""Return the reverse of a graph
+
+    The reverse (also called converse, transpose) of a graph with edges
+    :math:`(i_1, j_1), (i_2, j_2), \cdots` is a new graph with edges
+    :math:`(j_1, i_1), (j_2, i_2), \cdots`.
+
+    For a heterograph with multiple edge types, we can treat edges corresponding to each
+    type as a separate graph and compute the reverse for each of them.
+
+    Given a :class:`DGLGraph` object, we return another :class:`DGLGraph` object
```
mufeili, line 397: For cross-page reference, I think we need to replace `:class:`DGLGraph`` by `:class:`dgl.DGLGraph``

yzh119:
@classicsong does this PR also solved the problem mentioned in #1762 ?

python/dgl/transform.py
```diff
@@ -383,6 +384,137 @@ def reverse(g, share_ndata=False, share_edata=False):
...
+    representing its reverse.
+
+    Notes
+    -----
+    * We do not dynamically update the topology of a graph once that of its reverse changes.
+      This can be particularly problematic when the node/edge attrs are shared. For example,
+      if the topology of both the original graph and its reverse get changed independently,
+      you can get a mismatched node/edge feature.
```
mufeili, line 405: I think this is no longer an issue since we change from sharing frames to sharing tensors. We can use the following sentences for the nodes instead.

    * If ``share_ndata`` or ``share_edata`` is ``True``, same tensors will be used for 
      the features of the original graph and the reversed graph. As a result, users 
      should avoid performing in-place operations on the features of the reversed 
      graph, which will corrupt the features of the original graph as well. For 
      concrete examples, refer to the ``Examples`` section below.

python/dgl/transform.py
```diff
@@ -383,6 +384,137 @@ def reverse(g, share_ndata=False, share_edata=False):
...
+    Parameters
+    ----------
+    g : dgl.DGLGraph
+        The input graph.
+    share_ndata: bool, optional
+        If True, the original graph and its reverse share memory for node features.
+        Otherwise the reverse will not have any node features. Default: True
+    share_edata: bool, optional
+        If True, the original graph and its reverse share memory for edge features.
+        Otherwise the reverse will not have any edge features.
```
mufeili, line 416: Update the docstring for ``share_ndata`` and ``share_edata`` with

```
    share_ndata: bool, optional
        If True, the node features of the reversed graph will be the same as the 
        original graph. If False, the reversed graph will not have any node features. 
        (Default: True)
    share_edata: bool, optional
        If True, the edge features of the reversed graph will be the same as the 
        original graph. If False, the reversed graph will not have any edge features. 
        (Default: False)
```

python/dgl/transform.py
```diff
@@ -383,6 +384,137 @@ def reverse(g, share_ndata=False, share_edata=False):
...
+    >>> rg.edges(etype='follows')
+    (tensor([1, 2]), tensor([0, 2]))
+    >>> rg.edges(etype='plays')
+    (tensor([2, 1, 1]), tensor([1, 2, 1]))
+    >>> rg.nodes['game'].data['hv
+    tensor([[1.],
+            [1.],
+            [1.]])
+    >>> rg.edges['plays'].data
+    {}
```
mufeili, line 483: I've updated the examples in hackmd, can you update accordingly?

tests/compute/test_heterograph.py
```diff
@@ -1944,7 +1944,7 @@ def test_reverse(index_dtype):
         ('user', 'follows', 'user'): ([0, 1, 2, 4, 3 ,1, 3], [1, 2, 3, 2, 0, 0, 1]),
     }, index_dtype=index_dtype)
     gidx = g._graph
-    r_gidx = gidx.reverse()
+    r_gidx = gidx.reverse(gidx.metagraph)
```
mufeili, line 1947: I think this file is meant to test DGLHeteroGraph, not HeteroGraphIndex. Since we now moved the implementation to transform, to add tests in test_transform.py should be sufficient.

python/dgl/transform.py
```diff
@@ -383,6 +384,137 @@ def reverse(g, share_ndata=False, share_edata=False):
...
 
+def reverse_heterograph(g, share_ndata=True, share_edata=False):
+    r"""Return the reverse of a graph
+
+    The reverse (also called converse, transpose) of a graph with edges
+    :math:`(i_1, j_1), (i_2, j_2), \cdots` is a new graph with edges
+    :math:`(j_1, i_1), (j_2, i_2), \cdots`.
+
+    For a heterograph with multiple edge types, we can treat edges corresponding to each
+    type as a separate graph and compute the reverse for each of them.
```
mufeili, line 395: Let's also mention that the reverse for edge type `(A, B, C)` will have edge type `(C, B, A)`.

tests/compute/test_transform.py
```diff
@@ -115,6 +113,66 @@ def test_reverse():
...
+    g_r = dgl.reverse_heterograph(g, share_ndata=True, share_edata=True)
+    print(g_r)
+    for etype_g, etype_gr in zip(g.canonical_etypes, g_r.canonical_etypes):
+        assert etype_g[0] == etype_gr[2]
+        assert etype_g[1] == etype_gr[1]
+        assert etype_g[2] == etype_gr[0]
+        assert g.number_of_edges(etype_g) == g_r.number_of_edges(etype_gr)
+    assert F.array_equal(g.edges['follows'].data['h'], g_r.edges['follows'].data['h'])
+    assert F.array_equal(g.edges['follows'].data['hh'], g_r.edges['follows'].data['hh'])
+
```
mufeili, line 175: Can we add tests for in-place/out-place operations when share_ndata/share_edata is True?

tests/compute/test_transform.py
```diff
@@ -115,6 +113,66 @@ def test_reverse():
     assert g.edge_id(1, 2) == rg.edge_id(2, 1)
     assert g.edge_id(2, 1) == rg.edge_id(1, 2)
```
mufeili, line 115: Can we add a test for homographs?

classicsong:
> #1762

Yes

python/dgl/transform.py
classicsong: Sure

tests/compute/test_heterograph.py
classicsong: This is more like the unitest for single graphidx object.
The test here also check the graph format.

For tests in to_transform, the tests are more like API test.

tests/compute/test_transform.py
classicsong: Yes, we can do that.
But for in-place operations, do we really want this behavior?

tests/compute/test_transform.py
classicsong: Done

python/dgl/transform.py
```diff
@@ -306,20 +307,16 @@ def khop_graph(g, k):
 
 def reverse(g, share_ndata=False, share_edata=False):
     """Return the reverse of a graph
-
     The reverse (also called converse, transpose) of a directed graph is another directed
     graph on the same nodes with edges reversed in terms of direction.
-
     Given a :class:`DGLGraph` object, we return another :class:`DGLGraph` object
```
mufeili, line 312: Replace 

```bash
:class:`DGLGraph`
```

by

```bash
:class:`dgl.DGLGraph`
```

python/dgl/heterograph.py
```diff
@@ -4701,6 +4701,7 @@ def int(self):
                               self._node_frames,
                               self._edge_frames)
 
+
```
mufeili, line 4704: I would still prefer having a `def reverse(...)` here for backward compatibility since we have `DGLGraph.reverse`.

mufeili:
Two last comments. Other than those, I'm good with this PR.

PR #1785, opened 2020-07-10
[Feature] Add reverse edges to the graph

BarclayII:
## Description
There are two behaviors of adding reverse edges to make the graph bidirectional:

1. Reverse edges have the same edge type as the original.  This works only on homogeneous graphs.
2. Reverse edges have separate edge types from the original.  This can work on both homogenous graphs and heterogeneous graphs.

Therefore `add_reverse` and `add_reverse_types` are added as two separate APIs.

This is necessary for the implementation of `EdgeDataLoader` and `LinkPredictionDataLoader` since we need to not only exclude the edge being sampled in the minibatch but also their corresponding reverse edges.

`to_bidirected` will be deprecated since currently no example is using it and the behavior is quite complex and poorly defined.

Also included is the `compose` function that adds the edges from a list of graphs with the same set of nodes into a new graph.

## Checklist
- [ ] Code is well-documented


python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
...
+    >>> g1 = dgl.graph(([0, 1, 2], [1, 0, 2]), num_nodes=5)
+    >>> g2 = dgl.graph(([1, 3, 3], [0, 2, 3]), num_nodes=5)
+    >>> g = dgl.compose([g1, g2])
+    >>> g.all_edges(order='eid')
+    (tensor([0, 1, 2, 1, 3, 3]), tensor([1, 0, 2, 0, 2, 3]))
+    """
+    # Exposes the UnionCOO/UnionCSR functions to users here
+    pass
+
+def add_reverse(g):
```
mufeili, line 1409: So I guess this will simply replace `dgl.to_bidirected`?

python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
...
+    """Return a new graph with reverse edges added to the given homogeneous graph.
```
mufeili, line 1410: Our plan is to support heterogeneous graphs as well, in which case add_reverse will be performed for all type-based homogeneous graphs in the heterograph.

python/dgl/transform.py
mufeili, line 1409: Other behavior and input arguments I would recommend:
1. `share_ndata=True`. If True, we will perform a shallow copy of the node features (tensor-level, not frame-level) to the new graph.
2. `share_edata=False`. If True, the features for the new edges (j, i) added will be the same as (i, j) in the original graph.

python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
...
+
+    For each edge with ID ``e`` connecting from node ``u`` to node ``v``, a new
+    edge with ID ``e + |E|`` connecting from node ``v`` to node ``u`` is added,
+    where ``|E|`` is the number of edges in the original graph.
+
+    Notes
+    -----
+    The graph must be homogeneous (i.e. with one node type and one edge type).
```
mufeili, line 1418: If this is to support heterographs as well, we should remove the notes.

python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
...
+    pass
+
+def add_reverse(g):
+    """Return a new graph with reverse edges added to the given homogeneous graph.
+
+    For each edge with ID ``e`` connecting from node ``u`` to node ``v``, a new
+    edge with ID ``e + |E|`` connecting from node ``v`` to node ``u`` is added,
+    where ``|E|`` is the number of edges in the original graph.
+
+    Notes
```
mufeili, line 1416: Add a note about the potential risk of applying in-place operations to node features if `share_ndata=True`.

python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
...
+    """
+    # Exposes the UnionCOO/UnionCSR functions to users here
+    pass
+
+def add_reverse(g):
+    """Return a new graph with reverse edges added to the given homogeneous graph.
+
+    For each edge with ID ``e`` connecting from node ``u`` to node ``v``, a new
+    edge with ID ``e + |E|`` connecting from node ``v`` to node ``u`` is added,
+    where ``|E|`` is the number of edges in the original graph.
```
mufeili, line 1414: In the case of multiple edge types, need to emphasis that ``E`` is the number of edges for the particular edge type in the original graph.

python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
...
+
+    Notes
+    -----
+    The graph must be homogeneous (i.e. with one node type and one edge type).
+
+    Parameters
+    ----------
+    g : DGLHeteroGraph
```
mufeili, line 1422: I remember Minjie said that they will all be called DGLGraph in the future?

python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
...
+        The graph
+
+    Returns
+    -------
+    DGLHeteroGraph
```
mufeili, line 1427: Same as above.

python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
...
+        The new graph
+
+    Examples
+    --------
+    >>> g = dgl.graph(([0, 1, 2], [0, 2, 3]))
```
mufeili, line 1432: I would add examples covering:
- The case of a homograph or a heterograph with a single edge type
- The case of a heterograph with multiple edge types
- The risk of in-place operation on node features with ``share_ndata=True``.

python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
     g.edata.update(hg.edata)
     return g
```
mufeili, line 1369: Not necessarily need to complete in this PR, but I would recommend:
1. Adding an `is_block` flag in DGLHeteroGraph
2. Adding an API like `block_to_graph` for converting a block into a graph.
2. Each transform function should check if the input graph is a block. If this is the case, an warning should be raised for pointing to `block_to_graph`.

python/dgl/transform.py
BarclayII: You mean if a heterogeneous graph has two edge types `('A', 'AA', 'A')` and `('B', 'BB', 'B')`, then `add_reverse` will add reverse edges for both edge types (while ignoring edge types such as `('A', 'AB', 'B')`)?

python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
...
 
+def compose(graphs):
```
mufeili, line 1370: Do we want to support heterographs for this function?

python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
...
+    """Return a composition of the given list of graphs.
+
+    Given the list of graphs with the same set of nodes, ``compose`` will return a new
+    graph with the same set of nodes with all the edges from the list of graphs.
+
+    All edges from all the graphs will be added.
```
mufeili, line 1376: With this sentence, "all the edges from the list of graphs" above is redundant.

python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
...
+    -------
+    DGLHeteroGraph
+        The composed graph.
+
+    Examples
+    --------
+    >>> g1 = dgl.graph(([0, 1, 2], [1, 0, 2]), num_nodes=5)
+    >>> g2 = dgl.graph(([1, 3, 3], [0, 2, 3]), num_nodes=5)
+    >>> g = dgl.compose([g1, g2])
+    >>> g.all_edges(order='eid')
```
mufeili, line 1403: Let's use `g.edges` over `g.all_edges`. Also, are we going to make `order='eid'` the default behavior?

python/dgl/transform.py
mufeili: Yes.

python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
...
+    --------
+    >>> g = dgl.graph(([0, 1, 2], [0, 2, 3]))
+    >>> g2 = dgl.add_reverse(g)
+    >>> g.all_edges(order='eid')
+    (tensor([0, 1, 2, 0, 2, 3]), tensor([0, 2, 3, 0, 1, 2]))
+    """
+    # Will replace to_bidirected whose behavior is too complicated
+    pass
+
+def add_reverse_types(g, reverse_type_names=None, reverse_type_suffix='_inv'):
```
mufeili, line 1440: I probably rather prefer:
1. Creating the reverse of a graph with `dgl.reverse`
2. Allow merging heterographs with different relations

python/dgl/transform.py
```diff
@@ -1367,4 +1367,125 @@ def as_immutable_graph(hg):
...
+    """Return a new graph with reverse edges added to the given graph as another edge type.
+
+    For each edge with ID ``e`` of type ``etype`` connecting from node ``u`` of type ``utype``
+    to node ``v`` of type ``vtype``, a new edge with ID ``e`` of type ``etype_inv`` connecting
+    from node ``v`` of type ``vtype`` to node ``u`` of type ``utype`` is added.
```
mufeili, line 1445: I feel "connecting" in this sentence can be omitted.

python/dgl/transform.py
BarclayII: Hmm because people will actually use nightly build I guess I'll keep this name for now.  When we truly merge DGLGraph and DGLHeteroGraph I guess we can simply replace all occurrences of DGLHeteroGraph to DGLGraph with `sed`.

python/dgl/transform.py
BarclayII: If feature inheritance is involved then both `share_ndata` and `share_edata` can be omitted.  I can add the third argument.

python/dgl/transform.py
BarclayII: This should be handled by feature inheritance I assume.

BarclayII:
`compose()` is removed for now because I haven't think through the behavior on node/edge features yet.

python/dgl/transform.py
BarclayII: Need to rethink merging heterographs because I haven't think through how to deal with node/edge features.

python/dgl/transform.py
mufeili: I'd still prefer making `share_edata` an optional argument as this may not be always desired.

python/dgl/transform.py
mufeili: But we still want a description about the behavior right? The docstrings for APIs should stand alone.

python/dgl/transform.py
BarclayII: By feature inheritance I don't think the feature will be materialized to a tensor.  So why it is not always desired?

python/dgl/transform.py
BarclayII: Right.  Added

```
    Node features and edge features will inherit from the original graph.  The reverse
    edges (edge types) will have the same features as that of the original edges (edge types).
```

BarclayII:
Closing this because of #1793

PR #1789, opened 2020-07-11
[Transform] New ToSimple API

classicsong:
## Description
New ToSimple API

## Checklist
- [x] Code is well-documented


classicsong:
@BarclayII 
I think you need to modify PinSage according to this API.

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
-        If given, the mapping from the edge IDs of original graph to those of the returned
-        graph would be written into edge feature with this name in the original graph for
-        each edge type.
+        If given, the returned graph would have a column with
+        the same name that stores the number of duplicated edges
+        from the original graph.
+    writeback_mapping: bool, optional
+        If True, a write back mapping is returned for each edge
+        type subgraph. If False, only the g is returned.
+        (Default: Ture)
```
BarclayII, line 1291: `True`

BarclayII:
> @BarclayII
> I think you need to modify PinSage according to this API.

No changes needed on my side since PinSage does not use `writeback_mapping` for now.

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
         ret.edges[etype].data[EID] = induced_edges[i].tousertensor()
     return ret
 
-def to_simple(g, return_counts='count', writeback_mapping=None):
-    """Convert a heterogeneous multigraph to a heterogeneous simple graph, coalescing
-    duplicate edges into one.
+def to_simple(g, return_counts='count', writeback_mapping=True, share_ndata=True):
```
mufeili, line 1258: We can still have the option of share_edata. If True, we will randomly use one of the features for duplicate edges.

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
+    r"""Convert a heterogeneous multigraph to a heterogeneous
+    simple graph, coalescing duplicate edges into one.
```
mufeili, line 1260: I think we can simply say "Convert a graph to a simple graph without duplicate edges"

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
```
mufeili, line 1261: Let's start with the case of graphs with a single edge type to make the learning curve more smooth.

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
 
-    This function does not preserve node and edge features.
+    For a heterograph with multiple edge types, we can
+    treat edges corresponding
+    to each type as a separate graph and compute the
+    `simple` for each of them.
```
mufeili, line 1265: "compute the `simple` for each of them." to "convert each of them to a simple graph"

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
+
+    Writeback mappings can be returned for each edge
+    type subgraph. A writeback mapping is a tensor
+    recording the edge ids in the `simpled` graph for
+    each edges of the original graph. If the graph has
```
mufeili, line 1270: How about "For edges left in a simple graph, a writeback mapping records their ids in the original graph."?

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
+    only one edge type, a single tensor is returned.
+    If the graph has multiple edge types, a dictionary
+    of tensor is returned using canonical edge types
+    as the key.
+
+    Given a :class:`DGLGraph` object, we return another
+    :class:`DGLGraph` object representing its simplified version.
```
mufeili, line 1277: Change from 

```bash
:class:`DGLGraph`
```

to

```bash
:class:`dgl.DGLGraph`
```

for cross-page reference.

mufeili, line 1277: Also, let's say "the simple graph corresponding to it" rather than "its simplified version"

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
     return_counts : str, optional
-        If given, the returned graph would have a column with the same name that stores
-        the number of duplicated edges from the original graph.
-    writeback_mapping : str, optional
-        If given, the mapping from the edge IDs of original graph to those of the returned
-        graph would be written into edge feature with this name in the original graph for
-        each edge type.
+        If given, the returned graph would have a column with
+        the same name that stores the number of duplicated edges
+        from the original graph.
```
mufeili, line 1287: The notion "column" is only valid with the cognition of "frames". Let's just say

If given, the count of each edge in the original graph will be stored as edge features under the name ``return_counts``. (Default: `count`)

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
+    writeback_mapping: bool, optional
+        If True, a write back mapping is returned for each edge
+        type subgraph. If False, only the g is returned.
```
mufeili, line 1290: Let's change from "only the g is returned" to "only the simple graph is returned". `g` can be ambiguous here as we just used it for the input graph.

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
+        (Default: Ture)
+    share_ndata: bool, optional
+        If True, the node features of the reversed graph will
```
mufeili, line 1293: "reversed graph" -> "simple graph"

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
+        be the same as the original graph. If False, the reversed
```
mufeili, line 1294: "reversed graph" -> "simple graph"

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
+        graph will not have any node features.
+        (Default: True)
 
     Returns
     -------
-    DGLHeteroGraph
-        The new heterogeneous simple graph.
+    DGLGraph
+        A simple graph.
+    F.tensor or dict of F.tensor
```
mufeili, line 1302: Let's just use tensor or dict of tensor. F is not clear to the users.

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
-    The returned graph would have duplicate edges connecting (1, 3) and (1, 4) removed:
-    >>> sg.all_edges(form='uv', order='eid')
-    (tensor([0, 1, 1, 2]), tensor([1, 3, 4, 2]))
+    **Homographs or Heterographs with A Single Edge Type**
 
-    If ``return_counts`` is set, the returned graph will also return how many edges
-    in the original graph are connecting the endpoints of the edges in the new graph:
-    >>> sg.edata['weights']
-    tensor([1, 2, 2, 1])
+    Create a graph to simple. In the original graph, there are multiple edges between 1 and 2
```
mufeili, line 1312: "Create a graph to simple" -> "Create a graph for demonstrating `to_simple` API"

mufeili, line 1312: "." after "1 and 2"

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
-    One can also retrieve the mapping from the edges in the original graph to edges in
-    the new graph by setting ``writeback_mapping`` and running
-    >>> g.edata['new_eid']
-    tensor([0, 1, 3, 1, 2, 2])
-
-    This tells us that the first edge in ``g`` is mapped to the first edge in ``sg``, and
-    the second and the fourth edge are mapped to the second edge in ``sg``, etc.
+    >>> import dgl
+    >>> import torch as th
+    >>> g = dgl.graph((th.tensor([0, 1, 2，1]), th.tensor([1, 2, 0，2])))
```
mufeili, line 1316: "，" should be ","

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
+    >>> g.ndata['h'] = th.tensor([[0.], [1.], [2.]])
+    >>> g.edata['h'] = th.tensor([[3.], [4.], [5.], [6.]])
+
+    to_simple the graph
```
mufeili, line 1320: "to_simple the graph" -> "Convert the graph to a simple graph."

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
+
+    >>> sg, wm = dgl.to_simple(g)
+    >>> sg.ndata['h']
+    tensor([[0.],
+            [1.],
+            [2.]])
+    >>> u, v, eid = all_edges(form='all')
```
mufeili, line 1327: Change from

```python
u, v, eid = all_edges(form='all')
```

to

```python
u, v, eid = g.edges(form='all')
```

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
+    tensor([1, 2, 0])
+    >>> eid
+    tensor([0, 1, 2])
+    >>> wm
+    tensor([0, 1, 2, 1])
+    >>> 'h' in g.edata
+    False
+
+    **In-place operations on features of one graph will be reflected on features of
+    its to_simple. Out-place operations will not be reflected.**
```
mufeili, line 1340: "its to_simple" -> "the simple graph"

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
+    **Heterographs with Multiple Edge Types**
+
+    >>> g = dgl.heterograph({
+    >>>     ('user', 'wins', 'user'): (th.tensor([0, 2, 0, 2, 2]), th.tensor([1, 1, 2, 1, 0])),
+    >>>     ('user', 'plays', 'game'): (th.tensor([1, 2, 1]), th.tensor([2, 1, 1]))
+    >>> })
+    >>> g.nodes['game'].data['hv'] = th.ones(3, 1)
+    >>> g.edges['plays'].data['he'] = th.zeros(3, 1)
+
+    The to simple operation is applied to the subgraph
```
mufeili, line 1365: "to simple" -> "``to_simple``"

python/dgl/transform.py
```diff
@@ -1255,56 +1255,139 @@ def out_subgraph(g, nodes):
...
+    corresponding to ('user', 'wins', 'user') and the
+    subgraph corresponding to ('user', 'plays', 'game').
+    The return counts is stored in the edge feature
+    `cnt`
```
mufeili, line 1369: Why don't we introduce this concept in the case of graphs with single edge types and skip it for graphs with multiple edge types?

mufeili, line 1369: `cnt` should be replaced by `'cnt'`

python/dgl/transform.py
```diff
@@ -1315,9 +1398,16 @@ def to_simple(g, return_counts='count', writeback_mapping=None):
...
+        for ntype in g.ntypes:
+            for key in g.nodes[ntype].data:
+                simple_graph.nodes[ntype].data[key] = g.nodes[ntype].data[key]
+
+    if writeback_mapping:
+        wb_map = {}
         for edge_map, canonical_etype in zip(edge_maps, g.canonical_etypes):
-            g.edges[canonical_etype].data[writeback_mapping] = edge_map
+            wb_map[canonical_etype] = edge_map
+        return simple_graph, wb_map
```
mufeili, line 1410: In the case of graphs with a single edge type, this should return a tensor rather than a dict of tensor.

tests/compute/test_transform.py
```diff
@@ -574,16 +574,20 @@ def _check(g, new_g, induced_nodes):
 @parametrize_dtype
 def test_to_simple(index_dtype):
     g = dgl.heterograph({
```
mufeili, line 576: Let's also add a test for graphs with a single edge type. In particular, writeback mapping should be a tensor in this case.

tests/compute/test_transform.py
```diff
@@ -574,16 +574,20 @@ def _check(g, new_g, induced_nodes):
...
-        ('user', 'follow', 'user'): [(0, 1), (1, 3), (2, 2), (1, 3), (1, 4), (1, 4)],
-        ('user', 'plays', 'game'): [(3, 5), (2, 3), (1, 4), (1, 4), (3, 5), (2, 3), (2, 3)]}, index_dtype=index_dtype)
-    sg = dgl.to_simple(g, return_counts='weights', writeback_mapping='new_eid')
+        ('user', 'follow', 'user'): ([0, 1, 2, 1, 1, 1], [1, 3, 2, 3, 4, 4]),
```
mufeili, line 577: 'follows' for consistency

python/dgl/transform.py
classicsong: > We can still have the option of share_edata. If True, we will randomly use one of the features for duplicate edges.

I don't like this design. This only introduces the uncertainty.

python/dgl/transform.py
classicsong: This is a good idea.

python/dgl/transform.py
mufeili: We can set this argument to False by default. Meanwhile, this can be a natural behavior if the duplicate edges will have the same feature. If I am a user who wants to copy the edge features from the original graph to the converted simple graph, then I need to first compute a mapping from the edges in the simple graph to the edges in the original graph, which can be tedious.

python/dgl/transform.py
mufeili: @classicsong , any comments?

python/dgl/transform.py
```diff
@@ -1255,56 +1255,153 @@ def out_subgraph(g, nodes):
...
-    """Convert a heterogeneous multigraph to a heterogeneous simple graph, coalescing
-    duplicate edges into one.
+def to_simple(g, return_counts='count', writeback_mapping=False, share_ndata=True):
+    r"""Convert a graph to a simple graph without duplicate edges.
 
-    This function does not preserve node and edge features.
+    For a heterograph with multiple edge types, we can
+    treat edges corresponding
+    to each type as a separate graph and convert each
+    of them to a simple graph
```
mufeili, line 1264: Missing a '.'

python/dgl/transform.py
```diff
@@ -1255,56 +1255,153 @@ def out_subgraph(g, nodes):
...
+        graph will not have any node features.
+        (Default: True)
 
     Returns
     -------
-    DGLHeteroGraph
-        The new heterogeneous simple graph.
+    DGLGraph
+        A simple graph.
+    Tensor or dict of tensor
```
mufeili, line 1303: No need to capitalize `T`

mufeili:
Three final comments. Whether we want the `share_edata` argument may need some discussion.

python/dgl/transform.py
```diff
@@ -1295,6 +1295,13 @@ def to_simple(g, return_counts='count', writeback_mapping=False, copy_ndata=True
         from the original graph. If False, the simple
         graph will not have any node features.
         (Default: True)
+    copy_edata: bool, optional
+        If True, the edge features of the simple graph are lazy-copied
+        from the original graph. If there exists duplicate edges between
+        two vertexes (u, v), the edge feature of the simpled edge is
```
mufeili, line 1301: Let's not introduce "vertexes", but stick with "nodes"

python/dgl/transform.py
```diff
@@ -1295,6 +1295,13 @@ def to_simple(g, return_counts='count', writeback_mapping=False, copy_ndata=True
...
+        randomly selected from one of the duplicate edges.
```
mufeili, line 1302: "the edge feature of the simpled edge is randomly selected from one of the duplicate edges." -> "the feature of the edge is randomly selected from one of the duplicate edges."

python/dgl/transform.py
```diff
@@ -1393,56 +1393,159 @@ def out_subgraph(g, nodes):
...
 
-def to_simple(g, return_counts='count', writeback_mapping=None):
-    """Convert a heterogeneous multigraph to a heterogeneous simple graph, coalescing
-    duplicate edges into one.
+def to_simple(g, return_counts='count', writeback_mapping=False, copy_ndata=True, copy_edata=False):
+    r"""Convert a graph to a simple graph without duplicate edges.
 
-    This function does not preserve node and edge features.
+    For a heterograph with multiple edge types, we can
+    treat edges corresponding
```
jermainewang, line 1400: ..., we ~can~ treat ...

python/dgl/transform.py
```diff
@@ -1393,56 +1393,159 @@ def out_subgraph(g, nodes):
...
+    to each type as a separate graph and convert each
+    of them to a simple graph.
+
+    Writeback mappings can be returned for each edge
+    type subgraph. For the edges in the original graph,
```
jermainewang, line 1405: When ``writeback_mapping=True``, an extra mapping is returned. For the edges ...

python/dgl/transform.py
```diff
@@ -1393,56 +1393,159 @@ def out_subgraph(g, nodes):
...
-        If given, the returned graph would have a column with the same name that stores
-        the number of duplicated edges from the original graph.
-    writeback_mapping : str, optional
-        If given, the mapping from the edge IDs of original graph to those of the returned
-        graph would be written into edge feature with this name in the original graph for
-        each edge type.
+        If given, the count of each edge in the original graph
+        will be stored as edge features under the name
+        eturn_counts.
+        (Default: count)
```
jermainewang, line 1426: (Default: "count")

python/dgl/transform.py
```diff
@@ -1393,56 +1393,159 @@ def out_subgraph(g, nodes):
...
+        mapping is returned. If the graph has only
+        one edge type, a tensor is returned. If the
+        graph has multiple edge types, a dictionary
+        of tensor is return.
+
+    If ``copy_ndata`` is ``True``, same tensors will be used for
+    the features of the original graph and the to_simpled graph. As a result, users
+    should avoid performing in-place operations on the features of the to_simpled
+    graph, which will corrupt the features of the original graph as well. For
+    concrete examples, refer to the ``Examples`` section below.
```
jermainewang, line 1459: Try to avoid passive tense. Also, I will not explicitly **suggest** anything while just describing what will happen.
```
If ``copy_ndata=True``, it shallow-copies node feature tensors to the result graph.
In-place operations on the features of the result graph may change the features
of the original graph too (depending on the shallow-copy semantics of the backend
framework in-use).
```
@mufeili  please take a look too. This paragraph will appear in many API doc.

I also suggest putting it before the `Parameters` section and surround it with `..note::` block.

tests/compute/test_transform.py
```diff
@@ -682,17 +682,56 @@ def _check(g, new_g, induced_nodes):
...
+    g.ndata['h'] = F.tensor([[0.], [1.], [2.]])
+    g.edata['h'] = F.tensor([[3.], [4.], [5.], [6.]])
+    sg, wb = dgl.to_simple(g, writeback_mapping=True)
+    u, v = g.all_edges(form='uv', order='eid')
+    u = F.asnumpy(u).tolist()
+    v = F.asnumpy(v).tolist()
+    uv = list(zip(u, v))
+    eid_map = F.asnumpy(wb)
+
+    su, sv = sg.all_edges(form='uv', order='eid')
```
jermainewang, line 696: BTW, `edges` and `all_edges` are the same function. Just let you know.

tests/compute/test_transform.py
```diff
@@ -682,17 +682,56 @@ def _check(g, new_g, induced_nodes):
...
+    for i, e in enumerate(suv):
+        assert sc[i] == sum(e == _e for _e in uv)
+    for i, e in enumerate(uv):
+        assert eid_map[i] == suv.index(e)
+    # shared ndata
+    assert F.array_equal(sg.ndata['h'], g.ndata['h'])
+    assert ('h' in sg.edata) is False
+    # new ndata to sg
+    sg.ndata['hh'] = F.tensor([[0.], [1.], [2.]])
+    assert ('hh' in g.ndata) is False
```
jermainewang, line 711: You can write `assert 'hh' not in g.ndata`

tests/compute/test_transform.py
```diff
@@ -682,17 +682,56 @@ def _check(g, new_g, induced_nodes):
 
 @parametrize_dtype
 def test_to_simple(index_dtype):
```
jermainewang, line 684: Does the function support GPU? If not, could you add a unittest skip?

classicsong: No.
Let me add skip test

python/dgl/transform.py
classicsong: I think we can refine the docstring later. How to describe copy_ndata is not well defined right now.

PR #1800, opened 2020-07-13
[Distributed] add the standalone mode in DistGraph

zheng-da:
## Description
We will have the launch script to run DistGraphServer automatically. In this case, users cannot run their training job normally. If they want to run their training script, they have to run it with the launch script. It might be difficult especially if they want to develop their model in Jupyter Notebook. This standalone mode is to help users to debug and test in a local machine.

## Checklist
- [ ] Code is well-documented


examples/pytorch/graphsage/experimental/README.md
```diff
@@ -39,11 +44,15 @@ python3 train_dist.py --server --graph-name ogb-product --id 2 --num-client 4 --
 python3 train_dist.py --server --graph-name ogb-product --id 3 --num-client 4 --conf_path data/ogb-product.json --ip_config ip_config.txt
 ```
 
+When testing in the standalone mode, we don't need to run the servers.
```
aksnzhy, line 47: I'm a little confused here. Why we don't need server here? In my opinion, the standalone mode is just distributed training with one machine.

examples/pytorch/graphsage/experimental/README.md
```diff
@@ -53,3 +62,8 @@ python3 -m torch.distributed.launch --nproc_per_node=1 --nnodes=4 --node_rank=2
 # run client on machine 3
 python3 -m torch.distributed.launch --nproc_per_node=1 --nnodes=4 --node_rank=3 --master_addr="172.31.16.250" --master_port=1234 train_dist.py --graph-name ogb-product --ip_config ip_config.txt --num-epochs 3 --num-client 4 --batch-size 1000 --lr 0.1
 ```
+
+To run the training script in the standalone mode:
+```bash
+python3 train_dist.py --graph-name ogb-product --ip_config ip_config.txt --num-epochs 3 --batch-size 1000 --conf_path data/ogb-product.json --standalone
```
aksnzhy, line 68: remove `--graph-name`?

python/dgl/distributed/fake_kvstore.py
```diff
@@ -0,0 +1,58 @@
+"""Define a fake kvstore
+
+This kvstore is used when running in the standalone mode
+"""
+
+from .. import backend as F
+
+class KVClient(object):
```
aksnzhy, line 8: Can we put the logic into the real kvstore? It is a little weird we have a `fake_kvstore.py` here.

examples/pytorch/graphsage/experimental/README.md
zheng-da: the server is launched with the launch script afterwards. the standalone mode is to allow users to run their code without launch script. For example, running it in Jupyter Notebook environment.

python/dgl/distributed/fake_kvstore.py
zheng-da: it can. but i don't want to make the logic of kvstore too complex. maybe you don't like the name "fake". I think we can rename it as standalone_kvstore.py

PR #1807, opened 2020-07-15
[Feature] Shared memory utilities

soodoshll:
## Description

Some utility functions for shared memory

- copy sparse matrices to shared memory
- python interface of copying NDArray to shared memory
- save and load heterograph to shared memory

## Checklist
- [ ] Code is well-documented


soodoshll:
@jermainewang this PR is ready for review.

VoVAllen:
I'm still thinking of using plasma store for the shared memory management. Shared memory is for sharing data between process, which always results in consistency problem to be resolved. (When it can free the memory, how to notify the producer process). Plasma store could be a solution.

Tutorial link: https://github.com/apache/arrow/blob/master/cpp/apidoc/tutorials/plasma.md

Pros of plasma:
- Can do reference counting across process (because there's a server)
- Can save shape/dtype information along with the tensor buffer (No extra effort needed to communicate)
- Easier to get the global stats (all the tensors in the shared memory for now)

Cons:
- Needs a server

soodoshll:
We now plan to use shared memory in distributed training. When the server and the client are in the same machine, they share a huge read-only graph stored in shared memory. For sampler, we still use the old pickle&unpickle interface which relies on background framework. Thus there is no need to create and destroy objects frequently, and we don't need to maintain the lifetime of objects.

src/graph/heterograph.cc
```diff
@@ -260,6 +261,145 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DLContext& ctx) {
...
 }
 
+HeteroGraphPtr HeteroGraph::CopyToSharedMem(
+      HeteroGraphPtr g, const std::string& name, const std::set<std::string>& fmts) {
+  auto hg = std::dynamic_pointer_cast<HeteroGraph>(g);
+  if (hg->shared_mem_name_ == name)
+    return g;
+  
+  std::string buf;
+  dmlc::MemoryStringStream ofs(&buf);
```
VoVAllen, line 271: You can use MemoryFixedSizeStream if the size is known (avoid one copy here)

VoVAllen:
Also please add unittest for those

src/graph/heterograph.cc
```diff
@@ -260,6 +261,147 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DLContext& ctx) {
...
+  if (hg->SharedMemName() == name)
+    return g;
+
+  std::string buf;
+  dmlc::MemoryStringStream ofs(&buf);
+  dmlc::SeekStream *strm = &ofs;
+
+  bool has_coo = fmts.find("coo") != fmts.end();
+  bool has_csr = fmts.find("csr") != fmts.end();
+  bool has_csc = fmts.find("csc") != fmts.end();
```
VoVAllen, line 280: Is this the right way to get the available formats? @yzh119 @jermainewang

src/graph/heterograph.cc
soodoshll: > You can use MemoryFixedSizeStream if the size is known (avoid one copy here)

The size of serialized meta graph is unknown

VoVAllen: But I see the file size you created is SHARED_MEM_METAINFO_SIZE_MAX, which means you can create first and directly write into it with MemoryFixedSizeStream

src/graph/heterograph.cc
```diff
@@ -260,6 +261,144 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DLContext& ctx) {
...
+  auto hg_index = std::dynamic_pointer_cast<HeteroGraph>(ret);
+  hg_index->shared_mem_ = mem;
+  return ret;
+}
+
+aten::COOMatrix CreateCOOFromSharedMem(const std::string &name, uint8_t nbits,
+    int64_t num_src, int64_t num_dst, int64_t num_edges, bool rsorted, bool csorted) {
+  DLDataType dtype = {kDLInt, nbits, 1};
+  DLContext ctx = {kDLCPU, 0};
+  NDArray row = NDArray::EmptyShared(name+"_row", {num_edges}, dtype, ctx, false);
```
VoVAllen, line 323: I would prefer the name to be `name+"_coo_row"`

src/graph/heterograph.cc
```diff
@@ -260,6 +261,144 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DLContext& ctx) {
...
+  NDArray col = NDArray::EmptyShared(name+"_col", {num_edges}, dtype, ctx, false);
```
VoVAllen, line 324: Same

src/graph/heterograph.cc
```diff
@@ -260,6 +261,144 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DLContext& ctx) {
...
+  return aten::COOMatrix(num_src, num_dst, row, col, aten::NullArray(), rsorted, csorted);
+}
+
+aten::CSRMatrix CreateCSRFromSharedMem(const std::string &name, uint8_t nbits,
+    int64_t num_src, int64_t num_dst, int64_t num_edges, bool sorted) {
+  DLDataType dtype = {kDLInt, nbits, 1};
+  DLContext ctx = {kDLCPU, 0};
+  NDArray indptr = NDArray::EmptyShared(name + "_indptr", {num_src + 1}, dtype, ctx, false);
```
VoVAllen, line 332: Same as COO

src/graph/heterograph.cc
```diff
@@ -260,6 +261,144 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DLContext& ctx) {
...
+    auto num_dst = num_verts_per_type[dsttype];
+
+    aten::COOMatrix coo;
+    aten::CSRMatrix csr, csc;
+    std::string prefix = name + "_" + std::to_string(etype);
+    if (has_coo) {
+      bool rsorted, csorted;
+      CHECK(strm->Read(&rsorted)) << "invalid rsorted";
+      CHECK(strm->Read(&csorted)) << "invalid csorted";
+      coo = CreateCOOFromSharedMem(prefix + "_coo", nbits, num_src, num_dst,
```
VoVAllen, line 377: Move "_coo" to be inside CreateCOOFromSharedMem

src/graph/heterograph.h
```diff
@@ -7,11 +7,13 @@
 #ifndef DGL_GRAPH_HETEROGRAPH_H_
 #define DGL_GRAPH_HETEROGRAPH_H_
 
+#include <dgl/runtime/shared_mem.h>
 #include <dgl/base_heterograph.h>
 #include <dgl/lazy.h>
 #include <utility>
 #include <string>
 #include <vector>
+#include <set>
```
VoVAllen, line 16: Where's set used?

src/graph/heterograph_capi.cc
```diff
@@ -434,6 +435,27 @@ DGL_REGISTER_GLOBAL("heterograph_index._CAPI_DGLHeteroCopyTo")
...
+.set_body([] (DGLArgs args, DGLRetValue* rv) {
+    HeteroGraphRef hg = args[0];
+    std::string name = args[1];
+    List<Value> fmts = args[2];
+    std::set<std::string> fmts_set;
+    for (const auto &fmt : fmts) {
+      std::string fmt_data = fmt->data;
+      fmts_set.insert(fmt_data);
+    }
+    auto hg_share = HeteroGraph::CopyToSharedMem(hg.sptr(), name, fmts_set);
```
VoVAllen, line 448: Dynamic pointer cast the pointer to shared_ptr<HeteroGraph> here, and check whether the pointer inside is null or not

src/graph/heterograph.cc
```diff
@@ -260,6 +261,144 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DLContext& ctx) {
...
+    }
+    if (has_csc) {
+      csc = hg->GetCSCMatrix(etype).CopyToSharedMem(prefix + "_csc");
+      strm->Write(csc.sorted);
+    }
+    relgraphs[etype] = UnitGraph::CreateHomographFrom(csc, csr, coo, has_csc, has_csr, has_coo);
+  }
+
+  auto ret = HeteroGraphPtr(new HeteroGraph(hg->meta_graph_, relgraphs, hg->num_verts_per_type_));
+  auto hg_index = std::dynamic_pointer_cast<HeteroGraph>(ret);
```
VoVAllen, line 314: According to https://stackoverflow.com/a/46836519, seems you can directly return shared_ptr<HeteroGraph> as HeteroGraphPtr. Could you try it? dynamic_pointer_cast is not so safe as downcasting

src/graph/heterograph.h
soodoshll: Used in the parameter `fmts` of function `CopyToSharedMem`. Maybe we can use `dgl_format_code_t` and bit operation instead?

src/graph/heterograph.cc
soodoshll: Here I need to modify the returned HeteroGraph object so I do this dynamic_pointer_cast.

VoVAllen: What I mean is to directly `std::shared_ptr<HeteroGraph>(new HeteroGraph(...))` and `std::static_pointer_cast<BaseHeteroGraph>(...)` to the returned object if necessary.

src/graph/heterograph.h
VoVAllen: But it's only needed in cc file right? I'm fine with using set

VoVAllen: Also usually people use `unordered_set` more often, but here it's not critical which I think is fine

soodoshll: It is needed in the  of the function declaration

src/graph/heterograph.h
```diff
@@ -251,6 +273,8 @@ class HeteroGraph : public BaseHeteroGraph {
   FlattenedHeteroGraphPtr FlattenImpl(const std::vector<dgl_type_t>& etypes) const;
 };
 
+const int64_t SHARED_MEM_METAINFO_SIZE_MAX = 1024 * 16;  // at most 16k
```
VoVAllen, line 276: Is this enough with list of ntypes and etypes string? Or use MemoryStringStream and create the shared memory with the size of the backup string? (Sorry for 反复横跳)

src/graph/heterograph.cc
```diff
@@ -260,6 +261,155 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DLContext& ctx) {
...
+  return ret;
+}
+
+aten::COOMatrix CreateCOOFromSharedMem(const std::string &name, uint8_t nbits,
+    int64_t num_src, int64_t num_dst, int64_t num_edges, bool rsorted, bool csorted) {
+  DLDataType dtype = {kDLInt, nbits, 1};
+  DLContext ctx = {kDLCPU, 0};
+  NDArray row = NDArray::EmptyShared(name+"_row", {num_edges}, dtype, ctx, false);
+  NDArray col = NDArray::EmptyShared(name+"_col", {num_edges}, dtype, ctx, false);
+  return aten::COOMatrix(num_src, num_dst, row, col, aten::NullArray(), rsorted, csorted);
```
VoVAllen, line 330: Where's data ndarray in COO?

src/graph/heterograph.h
soodoshll: I have some questions
>     Raise error when calling shared_memory if graph index is on gpu
it returns a copy of the graph index in shared memory, no matter where the original graph is.

>     Formats should not be specified, just saving all the materialized formats
then users have to manually materialize formats? one of the purposes of this pr is to avoid such functions like `prepare_mp` in multi gpu version of graphsage.

src/graph/heterograph.cc
soodoshll: can we assume coo matrices used in heterograph don't have data array (and csr always have data array)? it simplifies the process.

src/graph/heterograph.h
soodoshll: > Is this enough with list of ntypes and etypes string? Or use MemoryStringStream and create the shared memory with the size of the backup string? (Sorry for 反复横跳)

Now the number of etypes is limited by the number of file descriptors, which means a graph in shared memory can have at most 1k~1.6k etypes

VoVAllen: I think it's fine to add TODOs for now. And let's discuss later when the cuda_graph pr merged.

src/graph/heterograph.cc
VoVAllen: I don't think you can make such assumption.

python/dgl/heterograph.py
```diff
@@ -4651,6 +4651,30 @@ def to_format(self, restrict_format):
...
+        ----------
+        name : str
+            The name of the shared memory.
+        formats : list of str (optional)
+            Desired formats to be materialized.
+
+        Returns
+        -------
+        HeteroGraph
+            The graph index in shared memory
```
zheng-da, line 4668: the returned object isn't graph index.

python/dgl/heterograph.py
```diff
@@ -4651,6 +4651,30 @@ def to_format(self, restrict_format):
...
+        """
+        assert len(name) > 0, "The name of shared memory cannot be empty"
+        assert len(formats) > 0
+        for fmt in formats:
+            assert fmt in ("coo", "csr", "csc")
+        gidx = self._graph.shared_memory(name, self.ntypes, self.etypes, formats)
+        return DGLHeteroGraph(gidx, self.ntypes, self.etypes)
```
zheng-da, line 4675: so this operation only moves the graph index to shared memory and the returned DGLHeterograph doesn't contain node/edge data. We need to clearly define the semantics in the docstring.

python/dgl/heterograph_index.py
```diff
@@ -227,6 +227,33 @@ def copy_to(self, ctx):
...
+        """Return a copy of this graph in shared memory
+
+        Parameters
+        ----------
+        name : str
+            The name of the shared memory.
+        ntypes : list of str
+            Name of node types
+        etypes : list of str
+            Name of edge types
```
zheng-da, line 240: Why do we need to provide `ntypes` and `etypes`?

zheng-da:
It looks good to me. Just a few small comments.

python/dgl/heterograph_index.py
soodoshll: We want to store the whole heterogeneous graph object into shared memory so we can load it only using its name, which needs to store ntypes and etypes. Writing to shared memory is done in C++, through the HeteroGraphIndex interface. But a HeteroGraphIndex object does not contain ntypes and etypes. So we need to pass them as arguments.

PR #1812, opened 2020-07-16
[Example] Rgcn support ogbn-mag dataset.

classicsong:
## Description
1. Add support of ogbn-mag dataset
2. Support nodes have initial node features.
3. Suport multi-gpu, fix-gpu-cpu training.

Minibatch-RGCN for OGBN-MAG without node-feats acc 41.90%.
Minibatch-RGCN for OGBN-MAG with node-feats acc 46.22%.

#1629 

## Checklist
- [ ] Code is well-documented


examples/pytorch/rgcn/entity_classify_mp.py
```diff
@@ -247,20 +279,27 @@ def run(proc_id, n_gpus, args, devices, dataset):
 
     for epoch in range(args.n_epochs):
         model.train()
-        optimizer.zero_grad()
-        if args.sparse_embedding:
-            emb_optimizer.zero_grad()
```
zheng-da, line 252: is this a bug? we need to invoke `zero_grad` for every batch, right?

examples/pytorch/rgcn/entity_classify_mp.py
```diff
@@ -269,54 +308,107 @@ def run(proc_id, n_gpus, args, devices, dataset):
...
+                    eval_seeds = []
+                    for i in range(n_gpus):
+                        log = queue.get()
+                        val_l, val_s = log
+                        eval_logits.append(val_l)
+                        eval_seeds.append(val_s)
+                    eval_logits = th.cat(eval_logits)
+                    eval_seeds = th.cat(eval_seeds)
+                val_loss = F.cross_entropy(eval_logits, labels[eval_seeds].cpu()).item()
+                val_acc = th.sum(eval_logits.argmax(dim=1) == labels[eval_seeds].cpu()).item() / len(eval_seeds)
```
zheng-da, line 357: it's better to put the evaluation code in a function, so it can be reused. also make the code more readable.

examples/pytorch/rgcn/entity_classify_mp.py
```diff
@@ -334,37 +426,80 @@ def main(args, devices):
...
+        train_idx = th.nonzero(train_mask).squeeze()
+        test_idx = th.nonzero(test_mask).squeeze()
+        node_feats = [None] * num_of_ntype
+
+        # split dataset into train, validate, test
+        if args.validation:
+            val_idx = train_idx[:len(train_idx) // 5]
+            train_idx = train_idx[len(train_idx) // 5:]
+        else:
+            val_idx = train_idx
```
zheng-da, line 492: the same comment here. it's better to put the code in functions so the code is more readable.

python/dgl/nn/pytorch/conv/relgraphconv.py
```diff
@@ -70,7 +70,8 @@ def __init__(self,
                  activation=None,
                  self_loop=False,
                  low_mem=False,
-                 dropout=0.0):
+                 dropout=0.0,
+                 layer_norm=False):
```
zheng-da, line 74: do you benchmark the benefit of using layer norm in GNN?

classicsong: for mag, 2 or 3% performance gain.

examples/pytorch/rgcn/entity_classify_mp.py
classicsong: Basically, we should update gradient every step.
It is fixed in this PR.

examples/pytorch/rgcn/entity_classify_mp.py
classicsong: I think we only need some extra words here to tell the purpose of these 4 lines of code.

examples/pytorch/rgcn/entity_classify_mp.py
classicsong: Yes, validation and test can use same pieces of code.

examples/pytorch/rgcn/entity_classify_mp.py
```diff
@@ -269,54 +326,64 @@ def run(proc_id, n_gpus, args, devices, dataset):
...
+                    val_seeds = []
+                    for i in range(n_gpus):
+                        log = queue.get()
+                        val_l, val_s = log
+                        val_logits.append(val_l)
+                        val_seeds.append(val_s)
+                    val_logits = th.cat(val_logits)
+                    val_seeds = th.cat(val_seeds)
+                val_loss = F.cross_entropy(val_logits, labels[val_seeds].cpu()).item()
+                val_acc = th.sum(val_logits.argmax(dim=1) == labels[val_seeds].cpu()).item() / len(val_seeds)
```
zheng-da, line 354: this is a very common function for multiprocessing training and distributed training. we need to formalize it and provide an API for this.

PR #1820, opened 2020-07-17
[Performance] HugePage allocation

pawelpiotrowicz:
Performance improvement


src/runtime/cpu_device_api.cc
```diff
@@ -32,8 +35,20 @@ class CPUDeviceAPI final : public DeviceAPI {
...
     if (ptr == nullptr) throw std::bad_alloc();
 #else
-    int ret = posix_memalign(&ptr, alignment, nbytes);
-    if (ret != 0) throw std::bad_alloc();
+#ifndef _WIN32
+    constexpr size_t _HugePage2MB_ = 1<<21;
+    if (nbytes >= _HugePage2MB_) {
+       int ret = posix_memalign(&ptr, _HugePage2MB_, nbytes);
+       if (ret != 0) throw std::bad_alloc();
+       if ((ret= madvise(ptr, nbytes, MADV_HUGEPAGE )) != 0 )
```
zheng-da, line 43: if `nbytes` isn't divisable by 2M, will a portion of the memory still be allocated with huge page?
I remember you said this kind of huge allocation is only available in the latest linux kernel. do you know which version of linux starts to support something like this?

src/runtime/cpu_device_api.cc
```diff
@@ -32,8 +35,20 @@ class CPUDeviceAPI final : public DeviceAPI {
     ptr = memalign(alignment, nbytes);
     if (ptr == nullptr) throw std::bad_alloc();
 #else
-    int ret = posix_memalign(&ptr, alignment, nbytes);
-    if (ret != 0) throw std::bad_alloc();
+#ifndef _WIN32
```
classicsong, line 38: Can we add a compilation flag to control whether to turn hugepage on or off.

src/runtime/cpu_device_api.cc
```diff
@@ -32,8 +35,20 @@ class CPUDeviceAPI final : public DeviceAPI {
...
+    constexpr size_t _HugePage2MB_ = 1<<21;
+    if (nbytes >= _HugePage2MB_) {
+       int ret = posix_memalign(&ptr, _HugePage2MB_, nbytes);
+       if (ret != 0) throw std::bad_alloc();
+       if ((ret= madvise(ptr, nbytes, MADV_HUGEPAGE )) != 0 )
+       throw std::bad_alloc();
+    } else {
+#endif
```
jermainewang, line 46: `#endif  // _WIN32`
too many if/endif makes it hard to digest the code

src/runtime/cpu_device_api.cc
zheng-da: why do you want to disable hugepage?

BarclayII: I don't think hugepage is available on Mac.  It does seem better to me to have a compilation flag instead.

src/runtime/cpu_device_api.cc
BarclayII: Since 2.6.38.  I would say this is pretty safe to call for Linux.

src/runtime/cpu_device_api.cc
BarclayII: Agreed.  Seems to me that you don't have to condition on `_WIN32`, because on Windows either `_MSC_VER` (for MSVC) or `__MINGW32__` (for MinGW which we no longer supports) should be defined.

BarclayII:
So on Linux, when the requested size is greater than 2MB, the alignment will be forced to 2MB, disregarding the user-supplied `alignment`.  Correct?

src/runtime/cpu_device_api.cc
BarclayII: For the first question the answer is [yes](https://man7.org/linux/man-pages/man2/madvise.2.html), hence the check on L40.

src/runtime/cpu_device_api.cc
```diff
@@ -32,8 +35,20 @@ class CPUDeviceAPI final : public DeviceAPI {
     ptr = memalign(alignment, nbytes);
     if (ptr == nullptr) throw std::bad_alloc();
 #else
-    int ret = posix_memalign(&ptr, alignment, nbytes);
-    if (ret != 0) throw std::bad_alloc();
+#ifndef _WIN32
+    constexpr size_t _HugePage2MB_ = 1<<21;
+    if (nbytes >= _HugePage2MB_) {
```
pioy, line 40: This condition seems be insufficient. If nbytes is only a bit more than _HugePage2MB_, e.g. _HugePage2MB_+64, we are likely losing space by allocating huge table for small number of bytes and, in consequence, the time because all allocated pages have to be cleared in KMD for security reason.
What about making decision based on proportion of wasted memory to real needs?
For example:
size_t wasted = _HugePage2MB_ - (nbytes % _HugePage2MB_);
float procwasted = (float)wasted/nbytes;
if (procwated < 0.2) {
....

src/runtime/cpu_device_api.cc
zheng-da: then we should only enable hugepage in linux.

PR #1833, opened 2020-07-20
[Transfom] Mutable transform

classicsong:
## Description
Outplace support for muations.
add_edges
add_nodes
remove_edges
remove_nodes
add_self_loop
remove_self_loop

## Checklist
- [x] Code is well-documented

### DIFF
--- a/examples/pytorch/rgcn/entity_classify.py
+++ b/examples/pytorch/rgcn/entity_classify.py
@@ -22,7 +22,7 @@ from model import BaseRGCN
 
 class EntityClassify(BaseRGCN):
     def create_features(self):
-        features = torch.ones(self.num_nodes)
+        features = torch.arange(self.num_nodes)
         if self.use_cuda:
             features = features.cuda()
         return features
--- a/examples/pytorch/rgcn/entity_classify.py
+++ b/examples/pytorch/rgcn/entity_classify.py
@@ -22,7 +22,7 @@ from model import BaseRGCN
 
 class EntityClassify(BaseRGCN):
     def create_features(self):
-        features = torch.arange(self.num_nodes)
+        features = torch.ones(self.num_nodes)
         if self.use_cuda:
             features = features.cuda()
         return features
--- a/python/dgl/heterograph.py
+++ b/python/dgl/heterograph.py
@@ -699,8 +699,10 @@ class DGLHeteroGraph(object):
                 old_eids = self.edges(form='eid', order='eid', etype=c_etype)
                 # trick here, eid_0 is 0 and should be handled
                 old_eids = F.scatter_row(old_eids,
-                    F.copy_to(F.tensor(0, dtype=F.int64), F.context(old_eids)),
-                    F.copy_to(F.tensor(1, dtype=F.dtype(old_eids)),F.context(old_eids)))
+                                         F.copy_to(F.tensor(0, dtype=F.int64),
+                                                            F.context(old_eids)),
+                                         F.copy_to(F.tensor(1, dtype=F.dtype(old_eids)),
+                                                            F.context(old_eids)))
                 old_eids = F.scatter_row(old_eids, eids, F.full_1d(
                     len(eids), 0, F.dtype(old_eids), F.context(old_eids)))
                 edges[c_etype] = F.tensor(F.nonzero_1d(old_eids), dtype=F.dtype(old_eids))
@@ -793,8 +795,10 @@ class DGLHeteroGraph(object):
                 old_nids = self.nodes(c_ntype)
                 # trick here, nid_0 is 0 and should be handled
                 old_nids = F.scatter_row(old_nids,
-                    F.copy_to(F.tensor(0, dtype=F.int64), F.context(old_nids)),
-                    F.copy_to(F.tensor(1, dtype=F.dtype(old_nids)),F.context(old_nids)))
+                                         F.copy_to(F.tensor(0, dtype=F.int64),
+                                                            F.context(old_nids)),
+                                         F.copy_to(F.tensor(1, dtype=F.dtype(old_nids)),
+                                                            F.context(old_nids)))
                 old_nids = F.scatter_row(old_nids, nids, F.full_1d(
                     len(nids), 0, F.dtype(old_nids), F.context(old_nids)))
                 nodes[c_ntype] = F.tensor(F.nonzero_1d(old_nids), dtype=F.dtype(old_nids))
--- a/python/dgl/heterograph.py
+++ b/python/dgl/heterograph.py
@@ -699,8 +699,8 @@ class DGLHeteroGraph(object):
                 old_eids = self.edges(form='eid', order='eid', etype=c_etype)
                 # trick here, eid_0 is 0 and should be handled
                 old_eids = F.scatter_row(old_eids,
-                                         F.tensor(0, dtype=F.int64),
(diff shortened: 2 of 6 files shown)


python/dgl/transform.py
```diff
@@ -673,71 +676,693 @@ def metapath_reachable_graph(g, metapath):
...
+            old_eids = g.edges(form='eid', order='eid', etype=c_etype)
+            # trick here, eid_0 is 0 and should be handled
+            old_eids = F.scatter_row(old_eids,
+                                     F.copy_to(F.tensor(0, dtype=F.int64),
+                                               F.context(old_eids)),
+                                     F.copy_to(F.tensor(1, dtype=F.dtype(old_eids)),
+                                               F.context(old_eids)))
+            old_eids = F.scatter_row(old_eids, eids, F.full_1d(
+                len(eids), 0, F.dtype(old_eids), F.context(old_eids)))
+            edges[c_etype] = F.tensor(F.nonzero_1d(old_eids), dtype=F.dtype(old_eids))
```
jermainewang, line 1122: Suggest having a `utils.compensate(ids, max_range)` for computing the compensate set of the given IDs.

python/dgl/transform.py
```diff
@@ -673,71 +676,693 @@ def metapath_reachable_graph(g, metapath):
...
+    --------
+    add_nodes
+    add_edges
+    remove_nodes
+    """
+    # TODO(xiangsx): block do not support remove_edges
+    if etype is None:
+        if len(g.etypes) != 1:
+            raise DGLError('Edge type name must be specified if there are more than one ' \
+                            'edge types.')
```
jermainewang, line 1101: I feel this is a very common check that we should have a `utils.check_valid_etype(g, etype)` for that. Same for ntype.

python/dgl/transform.py
```diff
@@ -673,71 +676,693 @@ def metapath_reachable_graph(g, metapath):
...
+    eids = utils.prepare_tensor(g, eids, 'u')
+    assert g.number_of_edges(etype) > F.as_scalar(F.max(eids, dim=0)), \
+        'The input eid {} is out of the range [0:{})'.format(
+            F.as_scalar(F.max(eids, dim=0)), g.number_of_edges(etype))
```
jermainewang, line 1105: Same here, `utils.check_valid_eids(g, eids, etype)`.

python/dgl/transform.py
```diff
@@ -673,71 +676,693 @@ def metapath_reachable_graph(g, metapath):
...
+                            'Data length of {} should be {}, but got {}'.format(key,
+                                                                                len(u),
+                                                                                len(new_feats))
+                        new_g.edges[c_etype].data[key] = \
+                            F.cat([F.zeros(shape, F.dtype(val), F.context(val)),
+                                   new_feats], dim=0)
+        else:
+            for key, val in g.edges[c_etype].data.items():
+                new_g.edges[c_etype].data[key] = val
+
```
jermainewang, line 1033: Maybe a correct implementation for outplace mutation should be:
```python
def add_edges(g, u, v):
    g = g.clone()
    g.add_edges(u, v)
    return g
```
Especially if we will change some internal members in the future, this will always be safe.

python/dgl/heterograph.py
```diff
@@ -790,7 +794,11 @@ def remove_nodes(self, nids, ntype=None):
             if self.get_ntype_id(c_ntype) == ntid:
                 old_nids = self.nodes(c_ntype)
                 # trick here, nid_0 is 0 and should be handled
-                old_nids[0] += 1
+                old_nids = F.scatter_row(old_nids,
+                                         F.copy_to(F.tensor(0, dtype=F.int64),
+                                                   F.context(old_nids)),
+                                         F.copy_to(F.tensor(1, dtype=F.dtype(old_nids)),
+                                                   F.context(old_nids)))
```
jermainewang, line 801: same compensate logic here

python/dgl/transform.py
classicsong: There is lots of such codes. 
I think we should handle them in a batch later.

python/dgl/transform.py
classicsong: Review this check later when fixing test cases

jermainewang: Sure

python/dgl/transform.py
jermainewang: Sure

examples/pytorch/rgcn/entity_classify.py
```diff
@@ -22,7 +22,7 @@
 
 class EntityClassify(BaseRGCN):
     def create_features(self):
-        features = torch.arange(self.num_nodes)
+        features = torch.ones(self.num_nodes)
```
jermainewang, line 25: revert?

python/dgl/heterograph.py
```diff
@@ -696,12 +696,8 @@ def remove_edges(self, eids, etype=None):
...
             # the target edge type
             if c_etype == (u_type, e_type, v_type):
-                old_eids = self.edges(form='eid', order='eid', etype=c_etype)
-                # trick here, eid_0 is 0 and should be handled
-                old_eids[0] += 1
-                old_eids = F.scatter_row(old_eids, eids, F.full_1d(
-                    len(eids), 0, F.dtype(old_eids), F.context(old_eids)))
-                edges[c_etype] = F.tensor(F.nonzero_1d(old_eids), dtype=F.dtype(old_eids))
+                origin_eids = self.edges(form='eid', order='eid', etype=c_etype)
+                edges[c_etype] = utils.compensate(eids, origin_eids)
```
jermainewang, line 700: We don't need to create this range array right? Just an integer is enough.

python/dgl/heterograph.py
```diff
@@ -4476,6 +4347,32 @@ def cpu(self):
...
+        for s_ntype, e_type, d_ntype in self.canonical_etypes:
+            meta_edges.append((self.get_ntype_id(s_ntype), self.get_ntype_id(d_ntype)))
+
+        metagraph = graph_index.from_edge_list(meta_edges, True)
+        # rebuild graph idx
+        num_nodes_per_type = [self.number_of_nodes(c_ntype) for c_ntype in self.ntypes]
+        relation_graphs = [self._graph.get_relation_graph(self.get_etype_id(c_etype))
+                           for c_etype in self.canonical_etypes]
+        hgidx = heterograph_index.create_heterograph_from_relations(
+            metagraph, relation_graphs, utils.toindex(num_nodes_per_type, "int64"))
```
jermainewang, line 4368: It looks like we could just pass the same graphindex object to the result graph because those inplace mutation operators is actually *out-place* because a new graphindex object is created, which means it is safe to share graph index across different graphs.

examples/pytorch/rgcn/entity_classify.py
classicsong: done

python/dgl/heterograph.py
classicsong: eids is supposed to be consecutive but nids are not.
The compensate can handle none consecutive origin_ids.
I think this is not a critical path, so it is ok.

python/dgl/heterograph.py
classicsong: The graph index in C++ has lots of shared_ptr objects. Will this cause problems?

PR #1838, opened 2020-07-20
[Feature] Slice dstnode features in NN modules

BarclayII:
## Description
This PR makes all builtin NN modules slice output node features automatically when the input graph is a block.  This no longer requires the users to do it manually.

This will incur a user experience change, although previous code should also work.

Before:
```python
        for l, (layer, block) in enumerate(zip(self.layers, blocks)):
            h_dst = h[:block.number_of_dst_nodes()]
            h = layer(block, (h, h_dst))
            if l != len(self.layers) - 1:
                h = self.activation(h)
                h = self.dropout(h)
```

After:
```python
        for l, (layer, block) in enumerate(zip(self.layers, blocks)):
            h = layer(block, h)
            if l != len(self.layers) - 1:
                h = self.activation(h)
                h = self.dropout(h)
```

This change is also in line of what we have in the upcoming User Guide for minibatch training.

## TODO
- [ ] Test example models

## Checklist
- [x] Code is well-documented


yzh119:
I'm okay with this but I think you should probably describe the behavior in the docstring.

BarclayII:
@sneakerkg will add the docstrings of blocks in NN modules together in his PR.

PR #1847, opened 2020-07-22
[Distributed] update Embedding API

zheng-da:
## Description
Update the Embedding API to make it consistent with other components (`DistGraph` and `DistTensor`) and also similar to Pytorch Embedding.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/sparse_emb.py
+++ b/python/dgl/distributed/sparse_emb.py
@@ -4,7 +4,7 @@ from .. import backend as F
 from .dist_tensor import DistTensor
 from .graph_partition_book import PartitionPolicy, NODE_PART_POLICY
 
-class SparseNodeEmbedding:
+class Embedding:
     ''' Sparse embeddings in the distributed KVStore.
 
     The sparse embeddings are only used as node embeddings.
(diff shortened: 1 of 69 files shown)


python/dgl/distributed/sparse_emb.py
```diff
@@ -13,31 +13,32 @@ class SparseNodeEmbedding:
...
-        g.ndata[name] = DistTensor(g, shape, F.float32, name, part_policy, initializer)
+    def __init__(self, g, shape, name, init_func, part_policy=None):
+        if part_policy is None:
+            part_policy = PartitionPolicy(NODE_PART_POLICY, g.get_partition_book())
 
-        self._tensor = g.ndata[name]
+        self._tensor = DistTensor(g, shape, F.float32, name, init_func, part_policy)
         self._trace = []
 
     def __call__(self, idx):
```
aksnzhy, line 44: Do we need store gradient on server?

python/dgl/distributed/sparse_emb.py
```diff
@@ -13,31 +13,32 @@ class SparseNodeEmbedding:
     ----------
     g : DistGraph
```
aksnzhy, line 14: It is weird to pass the graph to Embedding. Actually the embedding does not rely on any graph information.

python/dgl/distributed/sparse_emb.py
```diff
@@ -13,31 +13,32 @@ class SparseNodeEmbedding:
...
-    name : str
-        The name of the embeddings
     shape : tuple of int
         The shape of the embedding. The first dimension should be the number of nodes.
-    initializer : callable
+    name : str
+        The name of the embeddings
+    init_func : callable
         The function to create the initial data.
+    part_policy : PartitionPolicy
```
aksnzhy, line 22: How to pass this policy to server?

python/dgl/distributed/sparse_emb.py
aksnzhy, line 44: How to delete the old gradient stored in trace?

python/dgl/distributed/sparse_emb.py
zheng-da: then what should we pass? unless we want to expose kvstore.

python/dgl/distributed/sparse_emb.py
zheng-da: i think kvstore should manage it.

python/dgl/distributed/sparse_emb.py
zheng-da: the trace is deleted in optimizer. but there is a bug if the optimizer isn't called, for example, in the inference. this is something we need to fix.

aksnzhy:
@zheng-da Is this PR ready for merge?

zheng-da:
i can be merged. But I like to have some more updates. I think it's better to rename it as DistEmbedding to make its name consistent with other classes (`DistGraph` and `DistTensor`). let's me push the changes.

PR #1861, opened 2020-07-25
[Distributed] turn off recording on embeddings in the inference.

zheng-da:
## Description
When recording data access from embeddings, we need to know whether the operation runs in the training mode or inference mode. We need to check if the system is recording all operators.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/backend/tensorflow/tensor.py
+++ b/python/dgl/backend/tensorflow/tensor.py
@@ -685,6 +685,9 @@ def grad(x):
 def is_no_grad(x):
     return cgrad.is_no_grad(x)
 
+def is_recording():
+    raise NotImplementedError("Tensorflow doesn't support is_recording")
+
 no_grad = None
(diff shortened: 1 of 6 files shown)


python/dgl/backend/tensorflow/tensor.py
```diff
@@ -685,6 +685,9 @@ def grad(x):
 def is_no_grad(x):
     return cgrad.is_no_grad(x)
 
+def is_recording():
+    raise NotImplementedError("Tensorflow doesn't support is_recording")
```
classicsong, line 689: Can you give a warning and return True so that tf can still work under train and inference.

zheng-da: a few backend APIs don't support tensorflow. we won't be able to support tensorflow for distributed training in 0.5

PR #1874, opened 2020-07-28
[Distributed] Refactor distributed training to new DGLGraph.

zheng-da:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/dist_graph.py
+++ b/python/dgl/distributed/dist_graph.py
@@ -23,7 +23,7 @@ from .rpc_server import start_server
 from .dist_tensor import DistTensor, _get_data_name
 
 def _copy_graph_to_shared_mem(g, graph_name):
-    new_g = g.shared_memory(graph_name, formats=('csc',))
+    new_g = g.shared_memory(graph_name, formats='csc')
     # We should share the node/edge data to the client explicitly instead of putting them
     # in the KVStore because some of the node/edge data may be duplicated.
     local_node_path = _get_ndata_path(graph_name, 'inner_node')
--- a/python/dgl/distributed/dist_graph.py
+++ b/python/dgl/distributed/dist_graph.py
@@ -23,7 +23,6 @@ from .rpc_server import start_server
 from .dist_tensor import DistTensor, _get_data_name
 
 def _copy_graph_to_shared_mem(g, graph_name):
-    print(type(g))
     new_g = g.shared_memory(graph_name, formats=('csc',))
     # We should share the node/edge data to the client explicitly instead of putting them
     # in the KVStore because some of the node/edge data may be duplicated.
--- a/python/dgl/distributed/dist_graph.py
+++ b/python/dgl/distributed/dist_graph.py
@@ -4,12 +4,12 @@ from collections.abc import MutableMapping
 import os
 import numpy as np
 
-from ..graph import DGLGraph
+from ..heterograph import DGLHeteroGraph
+from .. import heterograph_index
 from .. import backend as F
 from ..base import NID, EID
 from .kvstore import KVServer, KVClient
 from .standalone_kvstore import KVClient as SA_KVClient
-from ..graph_index import from_shared_mem_graph_index
 from .._ffi.ndarray import empty_shared_mem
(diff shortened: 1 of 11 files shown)


tests/distributed/test_partition.py
```diff
@@ -14,13 +14,16 @@
 
 def create_random_graph(n):
     arr = (spsp.random(n, n, density=0.001, format='coo', random_state=100) != 0).astype(np.int64)
-    ig = create_graph_index(arr, readonly=True)
+    arr = arr.tocoo()
+    row = F.tensor(arr.row, dtype=F.int64)
+    col = F.tensor(arr.col, dtype=F.int64)
+    ig = create_unitgraph_from_coo(1, n, n, row, col, ['coo', 'csr', 'csc'])
     return dgl.DGLGraph(ig)
```
VoVAllen, line 21: directly use dgl.graph(arr) should work

python/dgl/distributed/graph_services.py
```diff
@@ -133,7 +133,7 @@ def merge_graphs(res_list, num_nodes):
         dst_tensor = res_list[0].global_dst
         eid_tensor = res_list[0].global_eids
     g = graph((src_tensor, dst_tensor),
-              restrict_format='coo', num_nodes=num_nodes)
+              formats=('coo',), num_nodes=num_nodes)
```
zheng-da, line 136: do we need to specify formats here? @jermainewang the graph here is a subgraph generated from sampling.

yzh119: You can compare the speed of restricting `coo` and not restricting format after minjie finished the refactor the scheduler.

yzh119: btw `formats` also accepts a single string as input: `formats='coo'`

zheng-da: does it work?

zheng-da: i see. it doesn't work in `shared_memory`.

python/dgl/distributed/dist_graph.py
```diff
@@ -85,11 +80,10 @@ def _get_graph_from_shared_mem(graph_name):
...
     '''
-    gidx = from_shared_mem_graph_index(_get_graph_path(graph_name))
-    if gidx is None:
-        return gidx
-
-    g = DGLGraph(gidx)
+    g, ntypes, etypes = heterograph_index.create_heterograph_from_shared_memory(graph_name)
+    if g is None:
+        return None
+    g = DGLHeteroGraph(g, ntypes, etypes)
```
VoVAllen, line 86: I think you can do this by `dgl.hetero_from_shared_memory` at https://github.com/soodoshll/dgl/blob/a7dbfd1341d30125e9f740f514aad10e34325510/python/dgl/convert.py#L418

python/dgl/distributed/dist_graph.py
```diff
@@ -315,11 +309,7 @@ def __init__(self, ip_config, graph_name, gpb=None, conf_file=None):
         else:
             connect_to_server(ip_config=ip_config)
             self._client = KVClient(ip_config)
-            g = _get_graph_from_shared_mem(graph_name)
-            if g is not None:
-                self._g = as_heterograph(g)
-            else:
-                self._g = None
+            self._g = _get_graph_from_shared_mem(graph_name)
```
VoVAllen, line 312: Why is this still `_get_graph_from_shared_mem`?

python/dgl/distributed/graph_services.py
VoVAllen: Why it doesn't work with shared_memory?

python/dgl/distributed/dist_graph.py
zheng-da: it's different. i need to deal with the case that shared-memory graph index doesn't exist. also i don't want to import from convert.py.

python/dgl/distributed/dist_graph.py
zheng-da: i don't just share graph index. i also share some node/edge data. please check the function.

Issue #1776, opened 2020-07-09
[RFC] DGLDataset


classicsong:
Providing a basic template of building a Dataset object for DGL. The recommend logic of processing the Dataset is as the following:
 1) Check whether there is a cached Dataset (already processed and stored in the disk) by invoking `has_cache`
 2) If there is no cache, loading the raw dataset through `process`, otherwise goto 4
 3) Save the processed dataset into disk. We can use `dgl.data.save_graphs` to serialize DGLGraphs. The goto 5
 4) Load the cached Dataset. We can use `dgl.data.load_graphs` to deserialize saved DGLGraphs.
 5) Done

All builtin dataset will follow the above instructions.
- [x] BitcoinOTC
- [x] CoraDataset
- [x] CiteseerDataset
- [x] PubmedDataset
- [x] GDELT
- [x] GINDataset
- [x] AmazonCoBuy
- [x] Coauthor
- [x] ICEWS18
- [x] KarateClub
- [x] MiniGCDataset
- [x] PPIDataset
- [x] QM7b
- [x] AIFB
- [x] MUTAG
- [x] BGS
- [x] AM
- [x] RedditDataset
- [x] SBMMixture
- [x] SST
- [x] LegacyTUDataset/TUDataset
- [x] FB15k-237
- [x] FB15k
- [x] wn18

classicsong:
Well done everyone.

PR #1894, opened 2020-07-30
[Dataset] Add hash function to distinguish instances from the same dataset class

hetong007:
## Description
Add a `hash_key` parameter to the base dataset class. 

There are possibilities that the instances from the same dataset class could be very different, with respect to the constructor parameters. With `hash_key`, one can pass the tuple of parameters to it in order to get a hash value to distinguish instances.

https://github.com/dmlc/dgl/issues/1776

## Checklist
- [X] Code is well-documented


python/dgl/data/dgl_dataset.py
```diff
@@ -148,6 +155,9 @@ def _load(self):
             if self.verbose:
                 print('Done saving data into cached files.')
 
+    def _get_hash(self):
```
classicsong, line 158: I think, you need to put a example here.

python/dgl/data/dgl_dataset.py
```diff
@@ -39,16 +39,23 @@ class DGLDataset(object):
     save_dir : str
         Directory to save the processed dataset.
         Default: same as raw_dir
+    hash_key : tuple
+        A tuple of values as the input for the hash function.
+        Users can distinguish instances (and their caches on the disk)
+        from the same dataset class by comparing the hash values.
```
classicsong, line 45: what will be the default hash for ()?
You also need to tell the Default: ()

PR #1901, opened 2020-07-31
DistDataloader

VoVAllen:
## Checklist
- [ ] Code is well-documented


examples/pytorch/graphsage/experimental/train_dist.py
```diff
@@ -217,6 +218,16 @@ def run(args, device, data):
 def main(args):
     if not args.standalone:
         th.distributed.init_process_group(backend='gloo')
+
+    import multiprocessing, logging
+    logger = multiprocessing.log_to_stderr()
+    logger.setLevel(logging.DEBUG)
+
+    # os.environ['DGL_DIST_MODE'] = 'distributed'
```
zheng-da, line 226: are you still at the stage of debugging?

python/dgl/distributed/dist_graph.py
```diff
@@ -325,6 +320,30 @@ def __init__(self, ip_config, graph_name, gpb=None, conf_file=None):
             self._num_nodes += int(part_md['num_nodes'])
             self._num_edges += int(part_md['num_edges'])
 
+    def _init(self):
+        self._client = KVClient(self.ip_config)
+        self._g = _get_graph_from_shared_mem(self.graph_name)
+        self._gpb = get_shared_mem_partition_book(self.graph_name, self._g)
+        if self._gpb is None:
+            self._gpb = self._gpb_input
+        self._client.barrier()
```
VoVAllen, line 329: Blocked by this

examples/pytorch/graphsage/experimental/train_dist.py
```diff
@@ -265,8 +266,9 @@ def run(args, device, data):
 def main(args):
     if not args.standalone:
         th.distributed.init_process_group(backend='gloo')
+
+    dgl.distributed.init_rpc(args.ip_config, num_workers=args.num_workers)
```
zheng-da, line 270: George suggested to use a more general name for this. let's rename it as `initialize`.

python/dgl/distributed/dist_dataloader.py
```diff
@@ -0,0 +1,159 @@
...
+
+def init_fn(collate_fn, queue, sig_queue):
+    """Initialize setting collate function and mp.Queue in the subprocess"""
+    global DGL_GLOBAL_COLLATE_FN
+    global DGL_GLOBAL_MP_QUEUE
+    global DGL_SIG_QUEUE
+    DGL_SIG_QUEUE = sig_queue
+    DGL_GLOBAL_MP_QUEUE = queue
+    DGL_GLOBAL_COLLATE_FN = collate_fn
+    time.sleep(1)
```
zheng-da, line 34: why do we need to sleep here?

python/dgl/distributed/dist_dataloader.py
```diff
@@ -0,0 +1,159 @@
...
+def _exit():
+    exit_client()
+    time.sleep(1)
+
+
+def enable_mp_debug():
+    """Print multiprocessing debug information"""
+    import logging
+    logger = multiprocessing.log_to_stderr()
+    logger.setLevel(logging.DEBUG)
```
zheng-da, line 47: how does the logger work? i don't see where is used in the code.

python/dgl/distributed/dist_dataloader.py
```diff
@@ -0,0 +1,159 @@
...
+            (default: ``0``)
+        collate_fn (callable, optional): merges a list of samples to form a
+            mini-batch of Tensor(s).  Used when using batched loading from a
+            map-style dataset.
+        drop_last (bool, optional): set to ``True`` to drop the last incomplete batch,
+            if the dataset size is not divisible by the batch size. If ``False`` and
+            the size of dataset is not divisible by the batch size, then the last batch
+            will be smaller. (default: ``False``)
+        queue_size (int): Size of multiprocessing queue
+        """
```
zheng-da, line 70: we need docstring to describe `DistDataLoader`.

python/dgl/distributed/dist_dataloader.py
```diff
@@ -0,0 +1,159 @@
...
+        if not self.drop_last and len(dataset) % self.batch_size != 0:
+            self.expected_idxs += 1
+
+    def __next__(self):
+        if not self.started:
+            for _ in range(self.queue_size):
+                self._request_next_batch()
+        self._request_next_batch()
+        if self.recv_idxs < self.expected_idxs:
+            result = self.queue.get(timeout=9999)
```
zheng-da, line 115: why do we need timeout here?

python/dgl/distributed/dist_dataloader.py
```diff
@@ -0,0 +1,159 @@
...
+                return None
+            else:
+                end_pos = len(self.dataset)
+        else:
+            end_pos = self.current_pos + self.batch_size
+        ret = self.dataset[self.current_pos:end_pos]
+        self.current_pos = end_pos
+        return ret
+
+    def close(self):
```
zheng-da, line 154: do users have to call it explicitly? should we move the code to `__del__`?

python/dgl/distributed/rpc_client.py
```diff
@@ -183,30 +186,69 @@ def finalize_client():
...
+        print(e, flush=True)
+        traceback.print_exc()
+        raise e
+
+def get_sampler_pool():
+    """Return the sampler pool and num_workers"""
+    return SAMPLER_POOL, NUM_SAMPLER_WORKERS
+
+
+def init_rpc(ip_config, num_workers=0, max_queue_size=MAX_QUEUE_SIZE, net_type='socket'):
```
zheng-da, line 228: let's rename it as `initialize`.

should we move the code here to `__init__.py`? so we don't need to import kvstore in rpc_client.py? The initialization now also include kvstore.

zheng-da: we should also call `reset` in initialization.

tests/distributed/test_distributed_sampling.py
```diff
@@ -158,20 +159,20 @@ def check_rpc_sampling_shuffle(tmpdir, num_server):
 # Wait non shared memory graph store
 @unittest.skipIf(os.name == 'nt', reason='Do not support windows yet')
 @unittest.skipIf(dgl.backend.backend_name == 'tensorflow', reason='Not support tensorflow for now')
-def test_rpc_sampling_shuffle():
+@pytest.mark.parametrize("num_server", [1, 2])
+def test_rpc_sampling_shuffle(num_server):
```
zheng-da, line 163: why is this modification preferred?

tests/distributed/test_mp_dataloader.py
```diff
@@ -0,0 +1,123 @@
...
+        shuffle=True,
+        drop_last=False,
+        num_workers=4)
+
+    dist_graph._init()
+    for epoch in range(3):
+        for idx, blocks in enumerate(dataloader):
+            print(blocks)
+            print(blocks[1].edges())
+            print(idx)
```
zheng-da, line 78: can we check the sampled subgraphs instead of printing the edges? it'll be a lot of printed messages. also printing doesn't capture any errors in CI.

tests/distributed/test_mp_dataloader.py
```diff
@@ -0,0 +1,123 @@
...
+
+    dataloader.close()
+    dgl.distributed.exit_client()
+
+
+def main(tmpdir, num_server):
```
zheng-da, line 84: move all of the code here to the test function?

python/dgl/distributed/dist_dataloader.py
```diff
@@ -0,0 +1,159 @@
...
+        return self
+
+    def _request_next_batch(self):
+        next_data = self._next_data()
+        if next_data is None:
+            return None
+        else:
+            async_result = self.pool.apply_async(
+                call_collate_fn, args=(next_data, ))
+            self.send_idxs += 1
```
zheng-da, line 135: it doesn't seem `send_idxs` is used anywhere.

python/dgl/distributed/dist_dataloader.py
```diff
@@ -0,0 +1,159 @@
...
+        raise e
+    return 1
+
+
+def init_fn(collate_fn, queue, sig_queue):
+    """Initialize setting collate function and mp.Queue in the subprocess"""
+    global DGL_GLOBAL_COLLATE_FN
+    global DGL_GLOBAL_MP_QUEUE
+    global DGL_SIG_QUEUE
+    DGL_SIG_QUEUE = sig_queue
```
zheng-da, line 31: what is this used for?

python/dgl/distributed/dist_dataloader.py
```diff
@@ -0,0 +1,159 @@
...
+    def __init__(self, dataset, batch_size, shuffle=False,
+                 num_workers=1, collate_fn=None, drop_last=False,
+                 queue_size=None):
+        """
+        dataset (Dataset): dataset from which to load the data.
+        batch_size (int, optional): how many samples per batch to load
+            (default: ``1``).
+        num_workers (int, optional): how many subprocesses to use for data
+            loading. ``0`` means that the data will be loaded in the main process.
+            (default: ``0``)
```
zheng-da, line 61: this description seems not accurate.

tests/distributed/test_mp_dataloader.py
```diff
@@ -0,0 +1,123 @@
...
+    sampler = NeighborSampler(dist_graph, [5, 10],
+                              dgl.distributed.sample_neighbors)
+
+    # Create PyTorch DataLoader for constructing blocks
+    dataloader = DistDataLoader(
+        dataset=train_nid.numpy(),
+        batch_size=32,
+        collate_fn=sampler.sample_blocks,
+        shuffle=True,
+        drop_last=False,
```
zheng-da, line 70: I think both values should be tested in the unit test.

tests/distributed/test_mp_dataloader.py
```diff
@@ -0,0 +1,123 @@
...
+
+    dist_graph._init()
+    for epoch in range(3):
+        for idx, blocks in enumerate(dataloader):
+            print(blocks)
+            print(blocks[1].edges())
+            print(idx)
+
+    dataloader.close()
+    dgl.distributed.exit_client()
```
zheng-da, line 81: we shouldn't require users to call either of them explicitly in the code. We should remove them in our tests as well.

python/dgl/distributed/dist_dataloader.py
```diff
@@ -0,0 +1,159 @@
...
+    """Print multiprocessing debug information"""
+    import logging
+    logger = multiprocessing.log_to_stderr()
+    logger.setLevel(logging.DEBUG)
+
+class DistDataLoader:
+    """DGL customized multiprocessing dataloader"""
+
+    def __init__(self, dataset, batch_size, shuffle=False,
+                 num_workers=1, collate_fn=None, drop_last=False,
```
zheng-da, line 53: i think we should remove `num_workers`. we already specify it in init_rpc. later i see it reports an error if the number of workers defined here and the one defined in init_rpc don't match. to avoid confusing, we should just remove this argument.

python/dgl/distributed/dist_dataloader.py
```diff
@@ -0,0 +1,159 @@
...
+        if self.pool is None:
+            ctx = mp.get_context("spawn")
+            self.pool = ctx.Pool(num_workers)
+        else:
+            assert num_sampler_workers == num_workers, "Num workers should be the same"
+        results = []
+        for _ in range(num_workers):
+            results.append(self.pool.apply_async(
+                init_fn, args=(collate_fn, self.queue, self.sig_queue)))
+            time.sleep(0.1)
```
zheng-da, line 100: is sleep here necessary?

python/dgl/distributed/dist_dataloader.py
VoVAllen: It's for debug usage. I just want to keep it in case of debugging in the future

python/dgl/distributed/dist_dataloader.py
VoVAllen: In PyTorch, timeout is an argument for dataloader (https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataloader.py#L107) and default value is at (https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataloader.py#L808).

The reason is that the child process might fail and the main process might not get the failure message properly. Therefore infinitely waiting from queue will seems like a dead lock to user, which is not user friendly

python/dgl/distributed/dist_dataloader.py
VoVAllen: User doesn't need to call this. However, in the test the DistDataloader object is in the main process and I also `join` the server in the main process, therefore cannot avoid explicitly call exit_client and close in the test.

python/dgl/distributed/dist_dataloader.py
VoVAllen: To ensure the function is passed to all the subprocess. Because pool cannot specify which process to execute

python/dgl/distributed/rpc_client.py
VoVAllen: Moved to dist_context.py

tests/distributed/test_distributed_sampling.py
VoVAllen: When failure, we could know what's the exact argument causing failure

python/dgl/distributed/dist_dataloader.py
VoVAllen: Removed

python/dgl/distributed/dist_dataloader.py
VoVAllen: Removed

tests/distributed/test_mp_dataloader.py
VoVAllen: Same reason as above. Client is in the main process and I join the server process (wait until server exits). Therefore if I didn't explicitly call it here, it's a dead lock

python/dgl/distributed/dist_dataloader.py
VoVAllen: I met problem when remove this, however it's rare

python/dgl/distributed/dist_dataloader.py
zheng-da: you should document it in the comment.

python/dgl/distributed/dist_dataloader.py
zheng-da: i see. sleeping for 1 sec doesn't guarantee that this will work. we should try other synchronization strategies.
here you should write comments to document this.

python/dgl/distributed/dist_dataloader.py
zheng-da: what kind of problem? it sounds like there is a bug somewhere else.

python/dgl/distributed/dist_dataloader.py
zheng-da: if this is just for testing, i think it's better to remove it. DistTensor is exposed to users. we don't want users to call the API.

python/dgl/distributed/dist_context.py
```diff
@@ -0,0 +1,78 @@
...
+        connect_to_server(ip_config, max_queue_size, net_type)
+        init_kvstore(ip_config, role)
+    except Exception as e:
+        print(e, flush=True)
+        traceback.print_exc()
+        raise e
+
+
+def initialize(ip_config, num_workers=0, max_queue_size=MAX_QUEUE_SIZE, net_type='socket'):
+    """Init rpc service"""
```
zheng-da, line 39: this is exposed to users. we need more detailed descriptions of the API.

python/dgl/distributed/dist_context.py
```diff
@@ -0,0 +1,78 @@
...
+from .rpc_client import connect_to_server, shutdown_servers
+
+SAMPLER_POOL = None
+NUM_SAMPLER_WORKERS = 0
+INITIALIZED = False
+
+def set_initialized(value=True):
+    """Set the initialized state of rpc"""
+    global INITIALIZED
+    INITIALIZED = value
```
zheng-da, line 18: this one and `is_initialized` should still belong to rpc_client.

python/dgl/distributed/dist_dataloader.py
VoVAllen: https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataloader.py#L492
As mentioned in the docs above, __del__ is not guaranteed to be called.

tests/distributed/test_mp_dataloader.py
VoVAllen: Removed

python/dgl/distributed/dist_dataloader.py
VoVAllen: Removed

python/dgl/distributed/dist_dataloader.py
VoVAllen: Removed

python/dgl/distributed/dist_dataloader.py
VoVAllen: Removed

python/dgl/distributed/dist_dataloader.py
VoVAllen: I tested the exit_client function is called by atexit, in both main process and sampler process

PR #1934, opened 2020-08-05
Fix the performance issue of graph partitioning in new DGLGraph

zheng-da:
## Description
Graph partitioning in new DGLGraph is significantly slower and consumes more memory. The PR is to fix various issues.

The new version of Metis doesn't support Windows. I disable the Windows support for now.

## Checklist
- [ ] Code is well-documented


python/dgl/partition.py
```diff
@@ -175,9 +175,9 @@ def metis_partition_assignment(g, k, balance_ntypes=None, balance_edges=False):
     '''
     # METIS works only on symmetric graphs.
     # The METIS runs on the symmetric graph to generate the node assignment to partitions.
-    from .transform import to_bidirected # avoid cyclic import
     start = time.time()
-    sym_g = to_bidirected(g)
+    sym_gidx = _CAPI_DGLMakeSymmetric_Hetero(g._graph)
+    sym_g = DGLHeteroGraph(gidx=sym_gidx, ntypes=['_N'], etypes=['_E'])
```
classicsong, line 180: ntypes=['_N'] and etypes=['_E'] are the default value. We can omit it.

python/dgl/transform.py
```diff
@@ -221,8 +221,6 @@ def to_bidirected(g, readonly=None, copy_ndata=False):
                 "unidirectional bipartite graphs" \
                 ", but {} is unidirectional bipartite".format(c_etype)
 
-    assert g.is_multigraph is False, "to_bidirected only support simple graph"
```
classicsong, line 224: User should make sure the graph is a single graph.
Here we check the single graph.

zheng-da: i understand. however the flag is True by default. If I load an OGB graph, its flag is true. When the flag is true, it doesn't mean this is a multigraph.

PR #1953, opened 2020-08-06
[Refactor] Enable new kernel in all message passing APIs

jermainewang:
## Description

Removed all the old scheduler/ir/executor codes. Enabled new kernels in our message passing APIs. The main logic is in the `dgl/core.py`.

Other changes:
* Remove `FrameRef` and directly use `Frame` everywhere. Add a function called `subframe` which creates a frame lazily. The actual slicing happens when using `frame[key]` to get the tensor.
* Move all the subgraph routines into `dgl/subgraph.py` and implement feature extraction for each them.

## Checklist
- [x] Code is well-documented


python/dgl/heterograph.py
```diff
@@ -3101,16 +2699,15 @@ def set_e_initializer(self, initializer, field=None, etype=None):
...
         """Internal API to set node features.
 
         `data` is a dictionary from the feature name to feature tensor. Each tensor
         is of shape (B, D1, D2, ...), where B is the number of nodes to be updated,
         and (D1, D2, ...) be the shape of the node representation tensor. The
         length of the given node ids must match B (i.e, len(u) == B).
 
-        All update will be done out of place to work with autograd unless the
-        inplace flag is true.
+        All update will be done out of place to work with autograd.
```
BarclayII, line 2710: updates

python/dgl/heterograph.py
```diff
@@ -3381,7 +2892,7 @@ def apply_nodes(self, func, v=ALL, ntype=None, inplace=False):
             The node type. Can be omitted if there is only one node type
             in the graph. (Default: None)
         inplace : bool, optional
-            If True, update will be done in place, but autograd will break.
+            **DEPRECATED**. If True, update will be done in place, but autograd will break.
```
BarclayII, line 2895: **DEPRECATED**.  Must be False.

python/dgl/heterograph.py
```diff
@@ -3427,7 +2939,7 @@ def apply_edges(self, func, edges=ALL, etype=None, inplace=False):
             The edge type. Can be omitted if there is only one edge type
             in the graph. (Default: None)
         inplace: bool, optional
-            If True, update will be done in place, but autograd will break.
+            **DEPRECATED**. If True, update will be done in place, but autograd will break.
```
BarclayII, line 2942: **DEPRECATED**.  Must be False.

python/dgl/heterograph.py
```diff
@@ -3612,7 +3014,7 @@ def send_and_recv(self,
...
```
BarclayII, line 3017: **DEPRECATED**.  Must be False.

python/dgl/heterograph.py
```diff
@@ -3444,149 +2956,39 @@ def apply_edges(self, func, edges=ALL, etype=None, inplace=False):
...
-        receive them.
-
-        Optionally, apply a function to update the node features after "receive".
-
-        This is a convenient combination for performing
-        :mod:`send <dgl.DGLHeteroGraph.send>` along the ``edges`` and
-        :mod:`recv <dgl.DGLHeteroGraph.recv>` for the destinations of the ``edges``.
+        This is a convenient combination for performing
+        :mod:`send <dgl.DGLHeteroGraph.send>` along the ``edges`` and
+        :mod:`recv <dgl.DGLHeteroGraph.recv>` for the destinations of the ``edges``.
```
BarclayII, line 2991: Probably this should be rewritten since `send` and `recv` are removed.

python/dgl/heterograph.py
```diff
@@ -3825,7 +3098,7 @@ def pull(self,
             The edge type. Can be omitted if there is only one edge type
             in the graph. (Default: None)
         inplace: bool, optional
-            If True, update will be done in place, but autograd will break.
+            **DEPRECATED**. If True, update will be done in place, but autograd will break.
```
BarclayII, line 3101: **DEPRECATED**.  Must be False.

python/dgl/heterograph.py
```diff
@@ -5251,4 +4378,72 @@ def __repr__(self):
...
+        The input graph.
+    u : Tensor
+        Src nodes.
+    v : Tensor
+        Dst nodes.
+    eid : Tensor
+        Edge IDs.
+    recv_nodes : Tensor
+        Nodes that receive messages. If None, it is equal to unique(v).
+        Otherwise, it must be a super set of v and can contain nodes
```
BarclayII, line 4408: "super set" -> "superset".

Also, if the dst nodes must have at least one incoming edge, can `recv_nodes` still be a superset of `v`?

python/dgl/heterograph.py
```diff
@@ -5251,4 +4378,72 @@ def __repr__(self):
             return ret.format(
                 srcnode=nsrcnode_dict, dstnode=ndstnode_dict, edge=nedge_dict, meta=meta)
 
+
+def _create_compute_graph(graph, u, v, eid, recv_nodes=None):
+    """Create a computation graph from the given edges.
+
+    The compute graph is a uni-directional bipartite graph with only
+    one edge type. There is no orphan destination nodes, i.e., all the
+    dst nodes have at least one in-coming edge.
```
BarclayII, line 4387: in-coming -> incoming

python/dgl/subgraph.py
```diff
@@ -0,0 +1,519 @@
...
+    else:
+        edge_frames = []
+        for i, ind_edges in enumerate(induced_edges):
+            subf = parent._edge_frames[i].subframe(ind_edges)
+            subf[EID] = ind_edges
+            edge_frames.append(subf)
+
+    hsg = DGLHeteroGraph(sgi.graph, parent.ntypes, parent.etypes,
+                         node_frames, edge_frames)
+    return hsg
```
BarclayII, line 517: I'm wondering if it is possible to have a utility function that let a user automatically copy-on-read features from a parent graph given the induced NID and EID.  For instance, we would also want to auto-slice features in various transformation functions like `sample_neighbors`, `select_topk`, `compact_graphs`, and `to_block`.

python/dgl/core.py
```diff
@@ -0,0 +1,293 @@
...
+        assert len(eid_bkt) == deg * len(node_bkt)
+        msgdata_bkt = msgdata.subframe(eid_bkt)
+        # reshape all msg tensors to (num_nodes_bkt, degree, feat_size)
+        maildata = {}
+        for k, msg in msgdata_bkt.items():
+            newshape = (len(node_bkt), deg) + F.shape(msg)[1:]
+            maildata[k] = F.reshape(msg, newshape)
+        # invoke udf
+        nbatch = NodeBatch(graph, orig_nid_bkt, ntype, ndata_bkt, msgs=maildata)
+        bkt_rsts.append(func(nbatch))
```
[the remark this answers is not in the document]
jermainewang: I don't think that will matter a lot since the bottleneck is usually kernel launching. Also it is non-trivial to achieve so in a framework-neural way, i.e., need to add extra abstraction of calling operators with different streams.

python/dgl/frame.py
```diff
@@ -53,108 +50,122 @@ def infer_scheme(tensor):
...
+        The row index to the feature data storage. None means an
+        identity mapping.
 
     Attributes
     ----------
+    storage : Tensor
+        Feature data storage.
     data : Tensor
-        The data of the column.
+        Feature data of this column.
```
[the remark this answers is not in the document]
jermainewang: > In this case, I wonder whether distinguishing storage and data as two public attributes (and `index` as a public attribute) is necessary, especially since `storage` doesn't seem to be accessed anywhere outside `Column`.

I think your question is more about can we have only one public `data` attribute? For the moment, yes. But exposing them is also fine as long as we document them clearly. Will update the docstring.

python/dgl/core.py
```diff
@@ -0,0 +1,293 @@
...
+
+    # prepare a result frame
+    retf = Frame(num_rows=len(nodes))
+    retf._initializers = dstdata._initializers
+    retf._default_initializer = dstdata._default_initializer
```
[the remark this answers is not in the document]
jermainewang: No, the result frame should have the same initializer as the graph's because the data is supposed to be written to the graph's frame.

python/dgl/frame.py
```diff
@@ -189,44 +227,41 @@ def __repr__(self):
...
-        update on one will not reflect to the other. The inplace update will
-        be seen by both. This follows the semantic of python's container.
-    num_rows : int, optional [default=0]
+        update on one will not reflect to the other.
+    num_rows : int, optional
         The number of rows in this frame. If ``data`` is provided and is not empty,
         ``num_rows`` will be ignored and inferred from the given data.
     """
-    def __init__(self, data=None, num_rows=0):
+    def __init__(self, data=None, num_rows=None):
```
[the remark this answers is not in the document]
jermainewang: > Can `data` and `num_rows` be both None?

Not at the moment but it could be when later we create a frame using `Frame()` and append another frame to it.

> Would it cause problems?

No

python/dgl/frame.py
```diff
@@ -168,18 +179,45 @@ def extend(self, feats, feat_scheme=None):
...
+
+        The operation triggers index selection.
+        """
         return Column(F.clone(self.data), self.scheme)
 
+    def subcolumn(self, rowids):
+        """Return a subcolumn.
+
+        If the self column has an index tensor, it will create a new index instead of
+        performing index selection.
```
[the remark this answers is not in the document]
jermainewang: Hmm, that description is a little bit confusing, will fix that.

python/dgl/heterograph.py
jermainewang: Yes, I will change them all later.

python/dgl/subgraph.py
jermainewang: > I'm wondering if it is possible to have a (public) utility function that let a user automatically copy-on-read features from a parent graph given the induced NID and EID.

Can we instead implement them using subgraph routines such as `edge_subgraph`?

python/dgl/core.py
```diff
@@ -0,0 +1,293 @@
...
+from .frame import Frame
+from .udf import NodeBatch, EdgeBatch
+from . import ops
+
+def is_builtin(func):
+    """Return true if the function is a DGL builtin function."""
+    return isinstance(func, fn.BuiltinFunction)
+
+def invoke_node_udf(graph, ntype, func, ndata, *, orig_nid=None):
+    """Invoke user-defined node function on all the nodes in the graph.
```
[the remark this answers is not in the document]
jermainewang: It actually invokes the user-defined node function on the given ndata.

python/dgl/core.py
```diff
@@ -0,0 +1,293 @@
...
+    bucketor : callable[Tensor -> list[Tensor]]
+        A bucketing function that splits the given tensor data as the same
+        way of how the :attr:`val` tensor is grouped.
+    """
+    sorted_val, idx = F.sort_1d(val)
+    unique_val = F.asnumpy(F.unique(sorted_val))
+    bkt_idx = []
+    for v in unique_val:
+        eqidx = F.nonzero_1d(F.equal(sorted_val, v))
+        bkt_idx.append(F.gather_row(idx, eqidx))
```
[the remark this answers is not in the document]
jermainewang: There are some issues with bool data type in MXNet, so I try to avoid it as much as possible.

python/dgl/heterograph.py
jermainewang: It could but there is no such case. The function is called by `send_and_recv`, `pull` and `update_all`. Only `pull` will set the `recv_nodes` and could contain zero-degree nodes.

python/dgl/ops/sddmm.py
```diff
@@ -41,6 +42,21 @@ def gsddmm(g, op, lhs_data, rhs_data, lhs_target='u', rhs_target='v'):
...
+        lhs_shape = F.shape(lhs_data)
+        rhs_shape = F.shape(rhs_data)
+        if len(lhs_shape) != len(rhs_shape):
+            max_ndims = max(len(lhs_shape), len(rhs_shape))
+            lhs_pad_ndims = max_ndims - len(lhs_shape)
+            rhs_pad_ndims = max_ndims - len(rhs_shape)
+            new_lhs_shape = (lhs_shape[0],) + (1,) * lhs_pad_ndims + lhs_shape[1:]
+            new_rhs_shape = (rhs_shape[0],) + (1,) * rhs_pad_ndims + rhs_shape[1:]
+            lhs_data = F.reshape(lhs_data, new_lhs_shape)
+            rhs_data = F.reshape(rhs_data, new_rhs_shape)
```
BarclayII, line 59: Maybe make an internal utility function for broadcasting (for this and gSpMM)?

python/dgl/ops/spmm.py
```diff
@@ -41,7 +44,32 @@ def gspmm(g, op, reduce_op, lhs_data, rhs_data):
...
+            new_lhs_shape = (lhs_shape[0],) + (1,) * lhs_pad_ndims + lhs_shape[1:]
+            new_rhs_shape = (rhs_shape[0],) + (1,) * rhs_pad_ndims + rhs_shape[1:]
+            lhs_data = F.reshape(lhs_data, new_lhs_shape)
+            rhs_data = F.reshape(rhs_data, new_rhs_shape)
+    if reduce_op == 'mean':
+        ret = gspmm_internal(g._graph, op, 'sum', lhs_data, rhs_data)
+        ret_shape = F.shape(ret)
+        deg = g.in_degrees()
+        if F.as_scalar(F.min(deg, dim=0)) == 0:
+            dgl_warning('Zero-degree node encountered in mean reducer.')
```
BarclayII, line 67: `'Zero-degree node encountered in mean reducer.  Setting the mean to 0.'`

python/dgl/subgraph.py
BarclayII: I partially take my claim back: we probably don't need a public utility function, but I think it's still better to have an internal utility function, since `compact_graphs` and `to_block` are not generating subgraphs but they need the copy-on-read functionality.

`sample_neighbors` and `select_topk` generates subgraphs and they can probably reuse `_create_hetero_subgraph`.

What PinSage samples is not a subgraph, but I guess it's OK to just update it with `neighbor_graph.ndata.update(self.G.nodes[ntype].data)` since they have the same node space.

python/dgl/transform.py
```diff
@@ -43,8 +43,6 @@
     'to_block',
     'to_simple',
     'to_simple_graph',
-    'in_subgraph',
-    'out_subgraph',
     'as_immutable_graph',
     'as_heterograph']
```
BarclayII, line 47: Can we add a TODO at the beginning: remove `copy_ndata` and `copy_edata` arguments?  Since copy-on-read is implemented I don't think they are necessary any more.

python/dgl/udf.py
```diff
@@ -95,33 +94,33 @@ def __len__(self):
         return self.batch_size()
 
     @property
-    def canonical_etype(self):
+    def etype(self):
         """Return the canonical edge type (i.e. triplet of source, edge, and
-        destination node type) for this edge batch, if available."""
-        return self._canonical_etype
+        destination node type) for this edge batch."""
+        return self._etype
```
BarclayII, line 100: We no longer have `canonical_etype`?

python/dgl/frame.py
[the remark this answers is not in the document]
BarclayII: K.

python/dgl/core.py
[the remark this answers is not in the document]
BarclayII: Then what does `orig_nid` do?  It's probably better to make the docstring consistent with `invoke_edge_udf` below where you say `on the given edges`.

python/dgl/frame.py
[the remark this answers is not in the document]
BarclayII: But when would we use `Frame()`?  When we don't have any data and the number of nodes/edges is 0?  If that's the only case then I would rather explicitly specify the number of nodes to be 0.

I was asking this because previously I have seen bugs in subgraph creation where the frame is incorrectly created with `Frame()` and therefore features cannot be assigned later.

A minor issue though.

python/dgl/heterograph.py
BarclayII: If there is the case when `recv_nodes` contains zero-degree nodes, then the description `There is no orphan destination nodes...` above should be removed.

python/dgl/function/base.py
```diff
@@ -15,9 +15,9 @@ class TargetCode(object):
     EDGE = 2
 
     CODE2STR = {
-        0: "src",
-        1: "dst",
-        2: "edge",
+        0: "u",
```
yzh119, line 18: Does this mean we no longer support `src_mul_edge`?

python/dgl/udf.py
yzh119: The `_etype` field is initialized with canonical etypes.

python/dgl/utils/checks.py
```diff
@@ -63,6 +63,34 @@ def prepare_tensor_dict(g, data, name):
     return {key : prepare_tensor(g, val, '{}["{}"]'.format(name, key))
             for key, val in data.items()}
 
+def parse_edges_arg_to_eid(g, edges, etid, argname='edges'):
+    """Parse edges argument and return edge IDs.
```
yzh119, line 67: Does this function return the edge ids with dtype: g.dtype and device: g.device of given edges?

python/dgl/core.py
```diff
@@ -0,0 +1,293 @@
...
+        y = alldata[mfunc.rhs][mfunc.rhs_field]
+        op = getattr(ops, '{}_{}'.format(mfunc.name, rfunc.name))
+        z = op(graph, x, y)
+    else:
+        x = alldata[mfunc.target][mfunc.in_field]
+        op = getattr(ops, '{}_{}'.format(mfunc.name, rfunc.name))
+        z = op(graph, x)
+    return {rfunc.out_field : z}
+
+def message_passing(g, mfunc, rfunc, afunc):
```
yzh119, line 245: I wonder how do we fuse multiple message functions and multiple reduce functions.
e.g. 
```
msg_func = [fn.copy_u('x', 'm1'), fn.u_mul_e('x', 'e', 'm2')]
reduce_func = [fn.sum('m1', 'y1'), fn.mean('m2', 'y2')]
```

python/dgl/_deprecate/frame.py
```diff
@@ -0,0 +1,1018 @@
+"""Columnar storage for DGLGraph."""
```
yzh119, line 1: Is this module still in use?

python/dgl/ops/sddmm.py
BarclayII: Sure.

python/dgl/utils/internal.py
```diff
@@ -708,3 +707,42 @@ def compensate(ids, origin_ids):
                          ids,
                          F.full_1d(len(ids), 0, F.dtype(ids), F.context(ids)))
     return F.tensor(F.nonzero_1d(mask), dtype=F.dtype(ids))
+
+def relabel(x):
```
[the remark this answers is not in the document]
jermainewang: I don't understand your question. This `relabel` function is an internal utility for converting IDs to consecutive range. When converting a networkx graph to DGLGraph, we use `networkx.convert_node_labels_to_integers`.

tests/tensorflow/test_nn.py
```diff
@@ -209,31 +209,6 @@ def test_edge_softmax():
     # checkout gradient
     assert F.allclose(grads[0], grad_score)
     print(grads[0][:10], grad_score[:10])
-
```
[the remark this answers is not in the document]
jermainewang: `group_apply_edges` is removed

python/dgl/function/base.py
jermainewang: We still do. `src_mul_edge` is just an alias for `u_mul_e`.

python/dgl/utils/checks.py
jermainewang: Yes and have clarified this in the docstring.

python/dgl/core.py
jermainewang: They are deprecated. To reuse the results of a message function for multiple reducers, call `apply_edges` first and use `copy_e` as the message function later.

python/dgl/_deprecate/frame.py
jermainewang: Will delete these `_deprecate` files later.

python/dgl/heterograph.py
jermainewang: done

python/dgl/transform.py
jermainewang: I scanned the functions that have `copy_ndata` and `copy_edata`.
- `to_bidirected`
- `add_reverse_edges`
- `reverse`
- `to_block`
- `to_simple`

I feel many of their `copy_xdata` semantics are quite specific to the operator semantics. Maybe you can have a discussion with @classicsong and @mufeili on which arguments should be kept?

python/dgl/udf.py
jermainewang: reverted

python/dgl/subgraph.py
jermainewang: Made a utility function called `utils.extract_subframes`.

python/dgl/core.py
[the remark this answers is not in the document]
jermainewang: Done.

python/dgl/ops/sddmm.py
jermainewang: Since it is only used by two functions (and very likely stays so), I feel inlining the codes in each function makes it easier to read.

python/dgl/transform.py
BarclayII: I feel no function would require `copy_ndata` and `copy_edata` since they are all `index_copy`ing from the original tensor.

python/dgl/ops/sddmm.py
BarclayII: I feel it's the opposite since they are essentially duplicate code.  But it's your choice anyway.

Issue #1947, opened 2020-08-06
Pinsage sampler.py meets 'method' object is not subscriptable


Woooosz:
## To Reproduce

1. python process_movielens1m.py ./ml-1m ./data.1m.pkl
2. python model.py data.1m.pkl --num-epochs 300 --num-workers 2 --device cuda:0 --hidden-dims 128
<img width="298" alt="截屏2020-08-06 上午11 21 22" src="https://user-images.githubusercontent.com/16368530/89487014-0af52580-d7d7-11ea-896d-5e5ca95a2756.png">

## Environment

 - dgl-cu100          0.5a200805
 - torch              1.2.0
 - torchtext          0.4.0
 - tornado            6.0.4
 - Centos 7
 - How you installed DGL (`conda`, `pip`, source): pip
 - Build command you used (if compiling from source):
 - Python version:3.6
 - CUDA/cuDNN version (if applicable): 100

PR #1957, opened 2020-08-06
[bugfix] Fix a bunch of examples to be compatible with dgl 0.5

yzh119:
## Description
As described in the title.

## Checklist
- [x] Code is well-documented

## Changes
- gat example
- gcn example
- graphsage example
- gcmc example
- [ ]


examples/pytorch/gat/train_ppi.py
```diff
@@ -60,7 +60,7 @@ def main(args):
     g = train_dataset[0]
     n_classes = train_dataset.num_labels
     num_feats = g.ndata['feat'].shape[1]
-    g = g.to(device)
+    g = g.int().to(device)
```
BarclayII, line 63: We still need to cast it to int32?

yzh119: I think so.

BarclayII: Hmm how much was the performance gain?  It's OK if you didn't benchmark it though; I'm just curious.  We may need to benchmark it later after 0.5.  I could imagine that newcomers will definitely ask why `g.int()` is necessary and how much the benefit is.

VoVAllen:
Also this https://github.com/dmlc/dgl/issues/1947 @BarclayII

BarclayII:
Should also fix #1947

examples/pytorch/gat/train_ppi.py
yzh119: For GATConv there is no performance gain, for SAGEConv and GraphConv where cusparse is invoked the performance gap is around 30% and I think the gap would be reduced after we merge tvm kernel.

PR #1976, opened 2020-08-08
[Distributed] Set the number of threads correctly to speed up

zheng-da:
## Description
When running all processes (trainer, sampler and server) together on the same set of machines, every process is competing for CPU resources. We need to set up the threads correctly to reduce the conflicts on CPU. The sampler processes and the server processes should use fewer threads.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/dist_context.py
+++ b/python/dgl/distributed/dist_context.py
@@ -1,6 +1,5 @@
 """Initialize the distributed services"""
 
-import os
 import multiprocessing as mp
 import traceback
 import atexit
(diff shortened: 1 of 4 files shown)


python/dgl/distributed/dist_context.py
```diff
@@ -29,6 +31,8 @@ def _init_rpc(ip_config, max_queue_size, net_type, role):
     ''' This init function is called in the worker processes.
     '''
     try:
+        os.environ['OMP_NUM_THREADS'] = '1'
```
BarclayII, line 34: Changing `OMP_NUM_THREADS` in a running process doesn't change the number of OMP threads.
https://software.intel.com/content/www/us/en/develop/documentation/mkl-linux-developer-guide/top/managing-performance-and-memory/improving-performance-with-threading/changing-the-number-of-openmp-threads-at-run-time.html

zheng-da: so we should create an CAPI to change the number of OMP threads?

BarclayII: I guess so.

Issue #1969, opened 2020-08-07
Support for Cuda 11


bjascob:
## 🚀 Feature
Cuda 11 in now supported by pytorch 1.6 (at least for Ubuntu via `pip3 install torch`).  I'm wondering if there are plans to support this in DGL?

The current binaries won't install because of missing cuda .so files but I was able to build and and run DGL with a few small mods.  Here's specifics in case someone else wants to try this...

**System:**
Ubuntu 20.04.1
Python 3.8 
Cuda 11
torch 1.6
GPU: GTX Titan X (5.2 compute capability)

**Mods starting from DGL master 8/7/2020**
1 - In `cmake/modules/CUDA.cmake`  change 2 instances of `sm_30` to something newer like `sm_52`.  Cuda 11 doesn't support `sm_30` anymore and fails on this argument.

2 - Cuda sparse functions `cusparseScsrmm2` and `cusparseDcsrmm2` no longer available.  These are replaced with `cusparseSpMM`.  The arguments are different and I didn't look into how to fix this but if you're not using sparse vectors these functions can be commented out in 3 different files.

- In `src/array/cuda/spmm.cu` and `src/kernel/cuda/binary_reduce_sum.cu` comment out the 2 functions that use the above.
- In `src/array/cuda/csr_transpose.cc`, comment out the code where cusparseScsr2csc is on about line 76

With those changes I was able to build and install.  I haven't tried to run any unit tests but the GraphWriter example is able to train.

yzh119:
Yes we do have plan to support CUDA 11, I'm take over the job.
For cusparse, they have several new APIs and we need to set up a benchmark to decide which one to use in practice.

yzh119:
Hi @bjascob , I've finished the CUDA 11 support in #1979 , I'm not sure if it would be merged into the master branch recently but you can try this feature by installing from source on [this branch](https://github.com/yzh119/dgl/tree/cu11).

bjascob:
Nice.  Everything built, installed, and at least started training like it should.  You even got rid of the C++14 warning messages.  
I'll let you close this out when you think it's appropriate (maybe after merge?)

[...]

PR #1979, opened 2020-08-08
[Feature] Use new cusparse API to support CUDA 11.

yzh119:
## Description
CuSparse have deprecated its APIs `cusparseScsrmm2`/`cusparseDcsrmm2` and turn to use `cusparseSpMM`in CUDA 11, thus CUDA11 users could not build our repo because of the API breaking changes(#1969 ).
This PR replaces old API calls with new ones when CUDA version >=11 is detected. I've verified its correctness/performance on some examples such as GCN/GAT.

This is the first step towards utilizing CUDA 11, they also have some new features (e.g. SDDMM and float16), we can utilize them in later PRs.

This PR may not be a part of DGL v0.5, but users who are using CUDA 11 could try installing DGL from source on this branch.

## Checklist
- [x] Code is well-documented


src/array/cuda/csr_transpose.cc
```diff
@@ -47,7 +47,7 @@ CSRMatrix CSRTranspose<kDLGPU, int32_t>(CSRMatrix csr) {
   int32_t* t_indices_ptr = static_cast<int32_t*>(t_indices->data);
   void* t_data_ptr = t_data->data;
 
-#if __CUDA_API_VERSION >= 10010
+#if CUDART_VERSION >= 10010
```
yzh119, line 50: Previously this branch was not activated.

cmake/modules/CUDA.cmake
```diff
@@ -43,15 +43,15 @@ set(CUDA_gpu_detect_output "")
       #find vcvarsall.bat and run it building msvc environment
       get_filename_component(MY_COMPILER_DIR ${CMAKE_CXX_COMPILER} DIRECTORY)
       find_file(MY_VCVARSALL_BAT vcvarsall.bat "${MY_COMPILER_DIR}/.." "${MY_COMPILER_DIR}/../..")
-      execute_process(COMMAND ${MY_VCVARSALL_BAT} && ${CUDA_NVCC_EXECUTABLE} -arch sm_30 --run  ${__cufile}
+      execute_process(COMMAND ${MY_VCVARSALL_BAT} && ${CUDA_NVCC_EXECUTABLE} -arch sm_35 --run  ${__cufile}
```
yzh119, line 46: sm30 is no longer supported.

CMakeLists.txt
```diff
@@ -77,9 +77,9 @@ if(MSVC)
   endif()
 else(MSVC)
   include(CheckCXXCompilerFlag)
-  check_cxx_compiler_flag("-std=c++11"    SUPPORT_CXX11)
+  check_cxx_compiler_flag("-std=c++14"    SUPPORT_CXX14)
```
yzh119, line 80: @BarclayII I'm using the latest version of cuda-gcc (9.3.0) so I must set standard to c++14 to make it work.
Could we make it a variable depending on the compiler version?

BarclayII: This crashes on CentOS 7 where GCC 4.8.5 doesn't support C++14.

Does CUDA-GCC 9.3.0 really require C++14?  What was the error message it gave you if you use C++11?  My Fedora is using CUDA-GCC 8.3.0 and it worked fine with C++11 albeit without CUDA 11.

yzh119: Sorry, I have confirmed that C++14 is not necessarily required but thrust throws a warning encouraging me to compile DGL with C++11.
```
/opt/cuda/include/thrust/detail/config/cpp_dialect.h:104:13: warning: Thrust requires C++14. Please pass -std=c++14 to your compiler. Define THRUST_IGNORE_DEPRECATED_CPP_DIALECT to suppress this message.
```

BarclayII: I see.  So thrust requires C++14.  You can switch the flag by checking CMake Variable `CUDA_VERSION_MAJOR`.

src/kernel/cuda/binary_reduce_sum.cu
```diff
@@ -125,6 +165,7 @@ void CusparseCsrmm2(
       static_cast<int32_t*>(csr.indptr->data),
       static_cast<int32_t*>(csr.indices->data),
       B_data, n, &beta, trans_out, m));
```
BarclayII, line 167: missing `CUSPARSE_CALL(cusparseDestroyMatDescr(descr));`

BarclayII (approved):
LGTM

PR #1982, opened 2020-08-09
no tuple input for heteroconv and each submodule call will take tuple

sneakerkg:
## Description

Current HeteroConv implementation seems not updated with some graph code change. Put the fix here:
1.Now we have g.is_block, HeteroConv will never receive tuple as input now;
2.For the nn module call for each relation, we will force each input is a bipartite graph and its tuple features.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/nn/pytorch/hetero.py
+++ b/python/dgl/nn/pytorch/hetero.py
@@ -135,38 +135,25 @@ class HeteroGraphConv(nn.Module):
         if mod_kwargs is None:
             mod_kwargs = {}
         outputs = {nty : [] for nty in g.dsttypes}
-        if isinstance(inputs, tuple) or g.is_block:
-            if isinstance(inputs, tuple):
-                src_inputs, dst_inputs = inputs
-            else:
-                src_inputs = inputs
-                dst_inputs = {k: v[:g.number_of_dst_nodes(k)] for k, v in inputs.items()}
 
-            for stype, etype, dtype in g.canonical_etypes:
-                rel_graph = g[stype, etype, dtype]
-                if rel_graph.number_of_edges() == 0:
-                    continue
-                if stype not in src_inputs or dtype not in dst_inputs:
-                    continue
-                dstdata = self.mods[etype](
-                    rel_graph,
-                    (src_inputs[stype], dst_inputs[dtype]),
-                    *mod_args.get(etype, ()),
-                    **mod_kwargs.get(etype, {}))
-                outputs[dtype].append(dstdata)
+        if g.is_block:
+            src_inputs = inputs
+            dst_inputs = {k: v[:g.number_of_dst_nodes(k)] for k, v in inputs.items()}
         else:
-            for stype, etype, dtype in g.canonical_etypes:
-                rel_graph = g[stype, etype, dtype]
-                if rel_graph.number_of_edges() == 0:
-                    continue
-                if stype not in inputs:
-                    continue
-                dstdata = self.mods[etype](
-                    rel_graph,
-                    inputs[stype],
-                    *mod_args.get(etype, ()),
-                    **mod_kwargs.get(etype, {}))
-                outputs[dtype].append(dstdata)
+            src_inputs = dst_inputs = inputs
+
+        for stype, etype, dtype in g.canonical_etypes:
+            rel_graph = g[stype, etype, dtype]
+            if rel_graph.number_of_edges() == 0:
+                continue
+            if stype not in src_inputs or dtype not in dst_inputs:
+                continue
+            dstdata = self.mods[etype](
+                rel_graph,
+                (src_inputs[stype], dst_inputs[dtype]),
+                *mod_args.get(etype, ()),
+                **mod_kwargs.get(etype, {}))
+            outputs[dtype].append(dstdata)
         rsts = {}
         for nty, alist in outputs.items():
             if len(alist) != 0:


sneakerkg:
@BarclayII @yzh119 @jermainewang may I bother you have a check for HeteroConv if the change is correct? Thanks.

sneakerkg:
@BarclayII I think I made a mistake. I used to thought heteroconv call will never receive a tuple, only its submodule will receive tuple. But if the submodule is SAGEConv, the input for heteroconv has to be tuple, for example:

```
    conv = nn.HeteroGraphConv({
        'follows': nn.SAGEConv(2, 3, 'mean'),
        'plays': nn.SAGEConv((2, 4), 4, 'mean'),
        'sells': nn.SAGEConv(3, 4, 'mean')},
        agg)

    h = conv(g, ({'user': uf}, {'user' : uf, 'game' : gf}))
```

So my change is not correct.

The real bug happens at, for each submodule, the input graph is a bipartite graph, but the input feature is not a tuple. For the above example, the input is just {'user': uf}.

Then it will fall into the `else` branch, send inputs[stype] as feature to the submodule. The submodule is checking it is bipartite or not using input type, since it's not tuple, feat_src and feat_dst will be the same. If the submodule is SAGEConv, and the src and dst type have the same feature dimension and number of nodes. Then the bug is triggered. We will use feat_src as feat_dst while the graph is actually bipartite.

So the real fix should be in each submodule, if the graph input is bipartite, the input feature has to be tuple.

BarclayII:
> @BarclayII I think I made a mistake. I used to thought heteroconv call will never receive a tuple, only its submodule will receive tuple. But if the submodule is SAGEConv, the input for heteroconv has to be tuple, for example:
> 
> ```
>     conv = nn.HeteroGraphConv({
>         'follows': nn.SAGEConv(2, 3, 'mean'),
>         'plays': nn.SAGEConv((2, 4), 4, 'mean'),
>         'sells': nn.SAGEConv(3, 4, 'mean')},
>         agg)
> 
>     h = conv(g, ({'user': uf}, {'user' : uf, 'game' : gf}))
> ```
> 
> So my change is not correct.
> 
> The real bug happens at, for each submodule, the input graph is a bipartite graph, but the input feature is not a tuple. For the above example, the input is just {'user': uf}.
> 
> Then it will fall into the `else` branch, send inputs[stype] as feature to the submodule. The submodule is checking it is bipartite or not using input type, since it's not tuple, feat_src and feat_dst will be the same. If the submodule is SAGEConv, and the src and dst type have the same feature dimension and number of nodes. Then the bug is triggered. We will use feat_src as feat_dst while the graph is actually bipartite.
> 
> So the real fix should be in each submodule, if the graph input is bipartite, the input feature has to be tuple.

You are talking about the test case in `tests/pytorch/test_nn.py`?  The test case doesn't make sense anymore because for full graphs why can't I just do
```python
h = conv(g, {'user': uf, 'game': gf})
```
The test case was there simply because we would like to quickly test whether `HeteroGraphConv` supports a pair as input, and that's only because we lacked the ability of testing if the input graph as a block.  Now that we are able to test if the graph is a block, you can just remove that test case.

sneakerkg:
@BarclayII @jermainewang @yzh119 @mufeili After some offline discussion, we plan to have a more strict way to call heteroconv to avoid undefined behavior or potential bug. Here's a summary. Correct me if I misunderstand something:

1.No tuple input for both heteroconv and each nn submodule; For heteroconv, the input is a dict, and when we call submodule, the input is also a dict. 
2.We will check whether the graph type and input feature matches. If it's homo, input is a tensor; if it's hetero, input is a dict. Each module will handle whether we give the correct dict, by using graph's src key and dst key. Then heteroconv module no need to handle the difference between GCN and SAGEConv;
3.When we call heteroconv, all relations will need to be run; We will not do any type checking and filtering in heteroconv, all the check is in submodules. If you only want to run part of the heterograph, please extract subgraph.

Since this is a breaking change, we may not merge this in v05 release.

BarclayII:
> 1.No tuple input for both heteroconv and each nn submodule; For heteroconv, the input is a dict, and when we call submodule, the input is also a dict.

So this is not allowed?

```python
conv = dglnn.SAGEConv(10, 20, 'mean')
conv(g['user', 'plays', 'game'], (g.nodes['user'].data['h'], g.nodes['game'].data['h']))
```

And it should be changed to

```python
conv = dglnn.SAGEConv(10, 20, 'mean')
conv(g['user', 'plays', 'game'], {'user': g.nodes['user'].data['h'], 'game': g.nodes['game'].data['h']})
```

Like this?

> 2.We will check whether the graph type and input feature matches. If it's homo, input is a tensor; if it's hetero, input is a dict. Each module will handle whether we give the correct dict, by using graph's src key and dst key. Then heteroconv module no need to handle the difference between GCN and SAGEConv;

> 3.When we call heteroconv, all relations will need to be run; We will not do any type checking and filtering in heteroconv, all the check is in submodules. If you only want to run part of the heterograph, please extract subgraph.

sneakerkg:
> > 1.No tuple input for both heteroconv and each nn submodule; For heteroconv, the input is a dict, and when we call submodule, the input is also a dict.
> 
> So this is not allowed?
> 
> ```python
> conv = dglnn.SAGEConv(10, 20, 'mean')
> conv(g['user', 'plays', 'game'], (g.nodes['user'].data['h'], g.nodes['game'].data['h']))
> ```
> 
> And it should be changed to
> 
> ```python
> conv = dglnn.SAGEConv(10, 20, 'mean')
> conv(g['user', 'plays', 'game'], {'user': g.nodes['user'].data['h'], 'game': g.nodes['game'].data['h']})
> ```
> 
> Like this?
> 
> > 2.We will check whether the graph type and input feature matches. If it's homo, input is a tensor; if it's hetero, input is a dict. Each module will handle whether we give the correct dict, by using graph's src key and dst key. Then heteroconv module no need to handle the difference between GCN and SAGEConv;
> 
> > 3.When we call heteroconv, all relations will need to be run; We will not do any type checking and filtering in heteroconv, all the check is in submodules. If you only want to run part of the heterograph, please extract subgraph.

Correct. No tuple input any more. However, I'm not quite sure how to get the src dst name from graph since the class is evolving, in my previous assumption, we don't need to provide g['user', 'plays', 'game'], g can somehow knows the src name is `user`. @mufeili might have the idea.

PR #1984, opened 2020-08-10
Remove stale partition code and fix #1949

### LINKED ISSUE #1949: Split tensor equally for multi gpu
When we run this example demo https://github.com/dmlc/dgl/blob/master/examples/pytorch/graphsage/train_sampling_multi_gpu.py to enable multi gpu graphsage training, we found there exists a mirror error that is located at line148. 
图片
Actually, its split result would drop some training data. For Example, if we have a tensor and the split logic is same as the example demo:
图片
The split result would be as follow and as you can see the final tensor element 8 is dropped.
图片

What we expected is to change the split logic as follow:
图片
Now the split result is:
图片
This time, all the training tensor would be loaded to its corresponding GPU to be trained.

Our Environment：
- DGL Version (master branch)

VoVAllen:
## Description
Fix #1949

## Checklist
- [ ] Code is well-documented


VoVAllen:
I think C++ codes would be deprecated with other GraphOP codes together

PR #1986, opened 2020-08-10
[Feature] Lazy copy ndata, edata to device

classicsong:
## Checklist
- [ ] Code is well-documented


python/dgl/frame.py
```diff
@@ -105,8 +106,15 @@ def shape(self):
     def data(self):
         """Return the feature data. Perform index selecting if needed."""
         if self.index is not None:
+            # copy index is usually cheaper than copy data
+            self.index = F.copy_to(self.index, F.context(self.storage))
```
jermainewang, line 110: Maybe we should check that the device of `index` is equal to `self.device` if it is not None or equal to the device of `storage` in the `subcolumn` function for a better error message. This also simplifies the implementation because there is no need to copy the index tensor to the storage device.

classicsong: Here we use copy to is to handle following case:
```
g = dgl.graph(([0,1,2], [2,3,1]))
g.ndata['h'] = th.tensor([1,1,1,2])
g = g.to(0)
print(g.ndata['h'])
# now ndata['h'] is in gpu

# go back to CPU
g = g.to(th.device('cpu'))
sg = dgl.node_subgraph(g, [0,1,2])
# now storage is still in GPU but index is in CPU
print(sg.ndata['h'])
```

jermainewang: i c

PR #2017, opened 2020-08-13
[WIP][Kernel] Set the built-in reduce result of zero-degree nodes to 0 in C

BarclayII:
## Description
We have decided to set the built-in reduce result of zero-degree nodes to 0 regardless of the reducer.  This is currently done in Python.  I'm wondering whether we can lower the logic to C which should make things faster.

For sum and mean reducers, we do nothing.

For max and min reducers, we simply replace all -infinity and infinity entries to 0.

## Checklist
- [x] Code is well-documented


src/array/cpu/spmm.h
```diff
@@ -104,8 +104,10 @@ void SpMMSumCoo(
       const DType* lhs_off = Op::use_lhs? X + rid * lhs_dim + lhs_add : nullptr;
       const DType* rhs_off = Op::use_rhs? W + eid * rhs_dim + rhs_add : nullptr;
       const DType val = Op::Call(lhs_off, rhs_off);
+      if (val != 0) {
```
yzh119, line 107: Why do we need this judgement?

BarclayII: See #1527

src/array/cpu/spmm.h
```diff
@@ -315,7 +328,10 @@ template <typename DType> constexpr bool CopyRhs<DType>::use_rhs;
 //////////////////////////////// Reduce operators on CPU ////////////////////////////////
 template <typename DType>
 struct Max {
-  static constexpr DType zero = std::numeric_limits<DType>::lowest();
+  static constexpr DType zero = (
+      std::numeric_limits<DType>::has_infinity ?
+      -std::numeric_limits<DType>::infinity() :
+      std::numeric_limits<DType>::lowest());
```
yzh119, line 334: Why don't we just use `lowest` and `max`?

BarclayII: Now I'm only using infinity.

PR #2028, opened 2020-08-15
[Doc] API Doc update for mxnet and tf, remove some degree check

sneakerkg:
## Description
1.API Doc update for mxnet and tf.
2.Remove degree check for nnconv and ginconv.
3.GAT has two fc layers for bipartite.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/nn/mxnet/conv/gatconv.py
+++ b/python/dgl/nn/mxnet/conv/gatconv.py
@@ -6,12 +6,17 @@ from mxnet.gluon import nn
 from mxnet.gluon.contrib.nn import Identity
 
 from .... import function as fn
+from ....base import DGLError
 from ....ops import edge_softmax
 from ....utils import expand_as_pair
 
 #pylint: enable=W0235
 class GATConv(nn.Block):
-    r"""Apply `Graph Attention Network <https://arxiv.org/pdf/1710.10903.pdf>`__
+    r"""
+
+    Description
+    -----------
+    Apply `Graph Attention Network <https://arxiv.org/pdf/1710.10903.pdf>`__
     over an input signal.
 
     .. math::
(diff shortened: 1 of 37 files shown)


python/dgl/nn/mxnet/conv/gatconv.py
```diff
@@ -53,16 +146,26 @@ def __init__(self,
...
         self._in_src_feats, self._in_dst_feats = expand_as_pair(in_feats)
         self._in_feats = in_feats
         self._out_feats = out_feats
+        self._allow_zero_in_degree = allow_zero_in_degree
         with self.name_scope():
-            self.fc = nn.Dense(out_feats * num_heads, use_bias=False,
-                               weight_initializer=mx.init.Xavier(magnitude=math.sqrt(2.0)),
-                               in_units=in_feats)
+            if isinstance(in_feats, tuple):
+                self.fc_src = nn.Dense(out_feats * num_heads, use_bias=False,
```
yzh119, line 159: If we change the behavior to two separate linear projection layers.
The `test_nn` and clusterGAT example have to be refactored accordingly.

python/dgl/nn/mxnet/conv/gatconv.py
```diff
@@ -84,33 +187,69 @@ def __init__(self,
...
+                                   'calling `g = dgl.add_self_loop(g)` will resolve '
+                                   'the issue. Setting ``allow_zero_in_degree`` '
+                                   'to be `True` when constructing this module will '
+                                   'suppress the check and let the code run.')
+
             if isinstance(feat, tuple):
                 h_src = self.feat_drop(feat[0])
                 h_dst = self.feat_drop(feat[1])
-                feat_src = self.fc(h_src).reshape(
+                feat_src = self.fc_src(h_src).reshape(
```
yzh119, line 250: Considering there are cases we apply GAT-on-homograph to bipartite graph, I suggest add a if-statement here:
```suggestion
                if not has_attr(self, fc_src): self.fc_src, self.fc_dst = self.fc, self.fc
                feat_src = self.fc_src(h_src).reshape(
```

yzh119:
GAT test: https://github.com/dmlc/dgl/blob/master/tests/pytorch/test_nn.py#L385-L392
ClusterGAT example: https://github.com/dmlc/dgl/blob/master/examples/pytorch/ogb/cluster-gat/main.py (note the difference between forward and inference function).

python/dgl/nn/mxnet/conv/gatconv.py
sneakerkg: When will we apply a GAT-on-homo to bipartite graph? As discussed before, for sample-based inference, we will just is_block() and  graph.number_of_dst_nodes() to get the dst nodes, no need to make the input tuple.

sneakerkg:
> GAT test: https://github.com/dmlc/dgl/blob/master/tests/pytorch/test_nn.py#L385-L392
> ClusterGAT example: https://github.com/dmlc/dgl/blob/master/examples/pytorch/ogb/cluster-gat/main.py (note the difference between forward and inference function).

Fixed ClusterGAT by sending a tensor not tuple when g is a block. 

For the test case, if no fc_src, we will go for fc. Still not quite sure whether we should provide this flexibility.

PR #2030, opened 2020-08-16
[Distributed] Fix sparse adagrad

zheng-da:
## Description
The sparse adagrad implementation is different from Pytorch. This leads to lower performance in trained models. This PR is to make sparse adagrad behave the same as Pytorch.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/sparse_emb.py
+++ b/python/dgl/distributed/sparse_emb.py
@@ -1,5 +1,7 @@
 """Define sparse embedding and optimizer."""
 
+import torch as th
+
 from .. import backend as F
 from .. import utils
 from .dist_tensor import DistTensor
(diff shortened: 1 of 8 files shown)


python/dgl/distributed/sparse_emb.py
```diff
@@ -136,6 +139,15 @@ def step(self):
                     # We'll need to scale them with the state sum on the kvstore server
                     # after we push them.
                     grads = F.cat(grads, 0)
-                    kvstore.push(name, idxs, grads)
+
+                uniq_idxs, inverse_idxs = F.unique(idxs, return_inverse=True)
```
classicsong, line 143: If it is a distributed system, shall we gather grad from all machines?

zheng-da: this requires significant modification. let's have this for now. we need to test if it's necessary to merge all gradients first.

PR #2046, opened 2020-08-18
[Feature] Change the Default Value of transpose in DGLGraph.adj

mufeili:
## Description
It was promised in the warning of 0.4.3.

## Checklist
- [x] Code is well-documented


yzh119:
My concern is that should we change the behavior in C++ end?
https://github.com/dmlc/dgl/blob/master/src/graph/unit_graph.cc#L1054-L1072

mufeili:
> My concern is that should we change the behavior in C++ end?
> https://github.com/dmlc/dgl/blob/master/src/graph/unit_graph.cc#L1054-L1072

I see. I don't feel compelled to re-flip the `transpose` flag in the API exposed to users. Another possibility is that we re-flip the flag internally and change the behavior in C++ end. Copy @jermainewang .

PR #2113, opened 2020-08-27
[Doc] New style of code blocks with prompts; copy button

BarclayII:
## Description
@classicsong brought up an issue on a code cell with manually numbered lines.  I don't like the appearance so I'm replacing it with `:lineno:` tag.

Also @classicsong mentioned that people wants to copy the code without prompts and outputs.  I found a solution using `sphinx-copybutton` extension.  Please check it out.

In the future developer's guide, we should put together another section settling the documentation style as well.


classicsong (approved):

BarclayII:
@VoVAllen Could you install `sphinx_copybutton` on the doc server?

VoVAllen:
Done

PR #2120, opened 2020-08-28
[Hotfix] Disable memory optimization for hypersparse graphs

BarclayII:
Currently hypersparse graph formats are forced to be COO to conserve memory, which requires all COO operators to be supported on GPU, which is not the case.

This is to temporarily disable this optimization.


yzh119:
An alternative is to comment these two lines:
https://github.com/dmlc/dgl/blob/44089c8b4d4db4ca71e816e0de50dca972dbabdb/src/graph/unit_graph.cc#L1449-L1450

Issue #2125, opened 2020-08-29
[Patch Release] 0.5.1


BarclayII:
This is a patch release including the following bug fixes and minor features

**Documentation fixes**

* #2081 Reorganize user guide and split chapters into multiple pages
* #2085 #2090 Fix links
* #2091 #2114 User guide on distributed training
* #2096 #2097 Other documentation fixes
* #2086
* #2123 Temporarily remove SSE MXNet tutorial

**Bug fixes**

* #2100 add_edges() crashes if the input tensor is empty
* #2084 Fix distributed GraphSage running with GPU
* #2107 Building with HDFS previously fails
* #2087 Cannot load the PTC dataset via `dgl.data.GINDataset`
* #2076 Empty cuda graph raise error in create_formats_
* #2121 Disable hypersparse memory optimization due to incomplete COO graph support on GPU
* #2115 Fallback to CPU for graph traversal functions on GPU graphs
* #2108 Force `num_workers` and `num_samplers` to be the same for distributed training
* #2118 (#2127 )

**Bug fixes in examples**

* #2119 Bug in using Layer Normalization in RGCN

**New features**

* #2102 5 utility functions that handle raw data features
* #1979 #2117 CUDA 11 support - We will not release binary builds of CUDA 11 for now

**Release changes**

* Linux now requires GCC 5 to build DGL from source.
* DGL now supports Mac 10.9+.

jermainewang:
> #2076 Empty cuda graph raise error in create_formats_

Shall we rename this API to `create_formats_`? The naming is inconsistent with `DGLGraph.formats`. @yzh119 

Since there are many changes, will you just merge the master to 0.5.x branch or do cherry-pick one-by-one?

BarclayII:
If we don't exclude commits I'll just do a merge.

yzh119:
> > #2076 Empty cuda graph raise error in create_formats_
> 
> Shall we rename this API to `create_formats_`? The naming is inconsistent with `DGLGraph.formats`. @yzh119
> 
> Since there are many changes, will you just merge the master to 0.5.x branch or do cherry-pick one-by-one?

I'm okay.

BarclayII:
0.5.1 has been released.

PR #2126, opened 2020-08-29
rename create_format_ to create_formats_

BarclayII:
As per requested in #2125 .


examples/pytorch/graphsage/train_sampling_unsupervised.py
```diff
@@ -298,7 +298,7 @@ def main(args, devices):
     val_mask = g.ndata['val_mask']
     test_mask = g.ndata['test_mask']
     g.ndata['features'] = features
-    g.create_format_()
+    g.create_formats_()
```
jermainewang, line 301: I'm surprised that there are so many examples using this API. The API should be used quite rarely. Even for mini-batch training, whether materializing all formats in the data loading process is beneficial is actually not clear. I feel we should avoid using this API in the examples unless absolutely sure. cc @classicsong for opinion.

Issue #2137, opened 2020-09-01
[Bug] 0.5.x taking too much shared memory during multiprocess training


BarclayII:
## 🐛 Bug

After upgrading to 0.5.x, training with multiprocessing now requires much larger shared memory, like, **much** larger, especially compared to 0.4.x.

Also the consumption of shared memory scales linearly with the number of worker processes, but that's also a problem in 0.4.x.

## To Reproduce

1. Take `examples/pytorch/graphsage/train_sampling.py` as an example.
1. Replace the node features with a much bigger dimension, like 50000.
1. Train and observe the shared memory consumption via `df`.  The shared memory consumption can be found as `/dev/shm`.

Here is my result on Reddit with feature size of 50000:

| Version | 2 workers | 4 workers |
|:-:|:-:|:-:|
| 0.4.x branch | 39M | 67M |
| master | 92G | 184G |

## Environment

 - DGL Version (e.g., 1.0):
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):
 - OS (e.g., Linux):
 - How you installed DGL (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version (if applicable):
 - GPU models and configuration (e.g. V100):
 - Any other relevant information:

jermainewang:
The 0.5 version copies the feature tensor of the parent graph to shared memory while 0.4.3 does not. It is due to the lazy-feature-copy behavior -- when a subgraph/block is created, it will hold a reference to the feature tensor of the parent graph and slice out subfeatures upon feature access. When the subgraph/block is transmitted to another process, the current behavior will copy the parent graph feature to shared memory, causing an excessive amount of shared memory usage. Three viable solutions:
1. For lazy features, instead of copying the parent graph feature to shared memory, first slice out subfeatures and then copy the subfeatures to shared memory. However, this may cause many unnecessary data transmission.
1. Try to deduce that the parent graph feature is transmitted back to the main process so there is no need to put it in shared memory. It is not clear whether it is feasible as the serializer/pickler has no knowledge about the receiver of the data.
1. In the high-level interfaces (e.g., NodeDataLoader and EdgeDataLoader), do not pass the parent features to the sampler process so that the sampled subgraphs/blocks will have no features. Once the main process gets the subgraphs/blocks, restore  the lazy features.

It looks like No.3 is a better solution.

VoVAllen:
[...]

PR #2139, opened 2020-09-02
[Bug] Fixes issues in feature pickling and transmission for subgraphs and blocks

### LINKED ISSUE #2135: [Bug] Pickling a subgraph stores the feature of the original graph
## 🐛 Bug

See reproducing code below.

This affects (1) pickling a graph as a file ~~, and (2) multiprocess neighborhood sampling.~~

## To Reproduce

```python
import dgl
import torch
import pickle
g = dgl.graph(([], []), num_nodes=10000000)
g.ndata['x'] = torch.randn(10000000, 100)
sg = g.subgraph([0])
with open('a.pkl', 'wb') as f:
    pickle.dump(sg, f)     # Gives a pickle file of 4GB
sg.ndata['x']              # Reads the feature
with open('b.pkl', 'wb') as f:
    pickle.dump(sg, f)     # Gives a pickle file of 3KB
```

The reason is in the implementation of lazy feature slicing for node and edge features.

## Expected behavior

Pickle file of `sg` should take a small amount of space.

BarclayII:
## Description
Fixes #2135 and #2137 by allowing the `storage` of `Column` to be `None`.

For subprocesses in PyTorch node and edge DataLoaders, we first check if the feature comes from the original graph by checking whether the column storage of the subgraph/block is identical to that of the original graph.  If so, we set the storage to be `None`.

After the subprocesses transmit the subgraphs/blocks, we then check in the main process if the column storage is None.  If so, we restore the storage to be that with the same name in the original graph.

### DIFF
--- a/python/dgl/dataloading/pytorch/__init__.py
+++ b/python/dgl/dataloading/pytorch/__init__.py
@@ -79,7 +79,6 @@ def _pop_blocks_storage(blocks, g):
 def _restore_subframe_storage(subframe, frame):
     for key, col in subframe._columns.items():
         if col.storage is None:
-            assert key in frame._columns
             col.storage = frame._columns[key].storage
 
 def _restore_subgraph_storage(subg, g):
--- a/python/dgl/dataloading/pytorch/__init__.py
+++ b/python/dgl/dataloading/pytorch/__init__.py
@@ -34,10 +34,14 @@ def _pop_subframe_storage(subframe, frame):
 
 def _pop_subgraph_storage(subg, g):
     for ntype in subg.ntypes:
+        if ntype not in g.ntypes:
+            continue
         subframe = subg._node_frames[subg.get_ntype_id(ntype)]
         frame = g._node_frames[g.get_ntype_id(ntype)]
         _pop_subframe_storage(subframe, frame)
     for etype in subg.canonical_etypes:
+        if etype not in g.canonical_etypes:
+            continue
         subframe = subg._edge_frames[subg.get_etype_id(etype)]
         frame = g._edge_frames[g.get_etype_id(etype)]
         _pop_subframe_storage(subframe, frame)
@@ -45,14 +49,20 @@ def _pop_subgraph_storage(subg, g):
 def _pop_blocks_storage(blocks, g):
     for block in blocks:
         for ntype in block.srctypes:
+            if ntype not in g.ntypes:
+                continue
             subframe = block._node_frames[block.get_ntype_id_from_src(ntype)]
             frame = g._node_frames[g.get_ntype_id(ntype)]
             _pop_subframe_storage(subframe, frame)
         for ntype in block.dsttypes:
+            if ntype not in g.ntypes:
+                continue
             subframe = block._node_frames[block.get_ntype_id_from_dst(ntype)]
             frame = g._node_frames[g.get_ntype_id(ntype)]
             _pop_subframe_storage(subframe, frame)
         for etype in block.canonical_etypes:
+            if etype not in g.canonical_etypes:
+                continue
             subframe = block._edge_frames[block.get_etype_id(etype)]
             frame = g._edge_frames[g.get_etype_id(etype)]
             _pop_subframe_storage(subframe, frame)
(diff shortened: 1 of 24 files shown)


python/dgl/dataloading/pytorch/__init__.py
```diff
@@ -13,6 +13,129 @@ def _remove_kwargs_dist(kwargs):
         print('Distributed DataLoader does not support pin_memory')
     return kwargs
 
+# The following code is a fix to the PyTorch-specific issue in
+# https://github.com/dmlc/dgl/issues/2137
+#
```
jermainewang, line 18: ...

The sampled blocks/subgraphs contain features extracted from the parent graph. In DGL, the sub-feature
data structure holds a reference to the parent graph feature tensor and an index tensor. When transmitting
the sampled block/subgraph to the main process, it is unnecessary to transmit the parent graph feature
back.

To achieve that, in the collator ....

python/dgl/dataloading/pytorch/__init__.py
```diff
@@ -13,6 +13,129 @@ def _remove_kwargs_dist(kwargs):
...
+        for etype in block.canonical_etypes:
+            subframe = block._edge_frames[block.get_etype_id(etype)]
+            frame = g._edge_frames[g.get_etype_id(etype)]
+            _restore_subframe_storage(subframe, frame)
+
+class _NodeCollator(NodeCollator):
+    def collate(self, items):
+        input_nodes, output_nodes, blocks = super().collate(items)
+        _pop_blocks_storage(blocks, self.g)
+        return input_nodes, output_nodes, blocks
```
jermainewang, line 95: Why not directly having `_pop_blocks_storage` in `NodeCollator`?

python/dgl/dataloading/pytorch/__init__.py
```diff
@@ -13,6 +13,129 @@ def _remove_kwargs_dist(kwargs):
...
+            input_nodes, pair_graph, blocks = super().collate(items)
+            _pop_subgraph_storage(pair_graph, self.g)
+            _pop_blocks_storage(blocks, self.g_sampling)
+            return input_nodes, pair_graph, blocks
+        else:
+            input_nodes, pair_graph, neg_pair_graph, blocks = super().collate(items)
+            _pop_subgraph_storage(pair_graph, self.g)
+            _pop_subgraph_storage(neg_pair_graph, self.g)
+            _pop_blocks_storage(blocks, self.g_sampling)
+            return input_nodes, pair_graph, neg_pair_graph, blocks
```
jermainewang, line 109: same here. I feel the logic can go to `EdgeCollator` directly.

python/dgl/dataloading/pytorch/__init__.py
```diff
@@ -13,6 +13,129 @@ def _remove_kwargs_dist(kwargs):
...
+            frame = g._node_frames[g.get_ntype_id(ntype)]
+            _pop_subframe_storage(subframe, frame)
+        for ntype in block.dsttypes:
+            subframe = block._node_frames[block.get_ntype_id_from_dst(ntype)]
+            frame = g._node_frames[g.get_ntype_id(ntype)]
+            _pop_subframe_storage(subframe, frame)
+        for etype in block.canonical_etypes:
+            subframe = block._edge_frames[block.get_etype_id(etype)]
+            frame = g._edge_frames[g.get_etype_id(etype)]
+            _pop_subframe_storage(subframe, frame)
```
jermainewang, line 58: What if the block/subgraph has new node/edge types that do not exist in the original graph?

python/dgl/dataloading/pytorch/__init__.py
```diff
@@ -13,6 +13,129 @@ def _remove_kwargs_dist(kwargs):
...
+
+def _restore_subframe_storage(subframe, frame):
+    for key, col in subframe._columns.items():
+        if col.storage is None:
+            assert key in frame._columns
+            col.storage = frame._columns[key].storage
```
jermainewang, line 64: This check is only valid when a column's `storage` can never be `None`. Maybe add one sentence in `Column`'s docstring to emphasize that?

jermainewang: I feel the assertion at L63 is redundant, gives no extra information and incurs an extra key lookup. I'd suggest remove it.

python/dgl/dataloading/pytorch/__init__.py
BarclayII: What if you write your own iterator that gets the blocks with NodeCollator (like in Tensorflow)?  The features of the blocks will be removed in this case.

python/dgl/dataloading/pytorch/__init__.py
BarclayII: I left one paragraph in the docstring of `Column`.  Does that suffice?

python/dgl/dataloading/pytorch/__init__.py
BarclayII: K.

python/dgl/dataloading/pytorch/__init__.py
BarclayII: Right.  Added a check.

python/dgl/dataloading/pytorch/__init__.py
BarclayII: Or we can put them in the node collator and have an argument `pop_storage` which defaults to False.

PR #2148, opened 2020-09-03
OMP pragma was added to support hetero_graph loop.

mozga-intel:
The OMP flag was added to support/obtain the inputs and outputs format of the matrices. This pull-request gives us the ability to get better performance in comparison with the master branch (approx. +5,6%). For brevity, the input and output format are triggered in a parallel way.


yzh119:
Could we only enable this optimization when the graph is stored on CPU? I don't think the optimization works for GPU as well.

mozga-intel:
@yzh119 Of course, done!

yzh119 (approved):
I'm okay with this PR.

mozga-intel:
@yzh119 Thanks for your approval, could you please tell me whether it could be merged into the master branch?

PR #2155, opened 2020-09-05
[performance] Batch DGLGraph in C++ end. 

### LINKED ISSUE #2145: [Regression] dgl.batch() in 0.5.x is slower than 0.4.x
## 🐛 Bug

Not sure if this is a bug, but here is the result:

## To Reproduce

Benchmarking script:
```python
import dgl
import numpy as np
import time
import itertools

for E, B in itertools.product([1000, 10000, 100000], [10, 100, 1000]):
    gs = [dgl.DGLGraph((np.random.randint(0, 100, E), np.random.randint(0, 100, E)), readonly=True) for _ in range(B)]
    t0 = time.time()
    for _ in range(20):
        g = dgl.batch(gs)
    tt = time.time()
    print(E, B, tt - t0)
```

Result on a g3.16x instance (although the code works on CPU):

|E|B|0.5.x|0.4.x|
|:-:|:-:|:-:|:-:|
| 1000 | 10 | 0.0637 | 0.00382 |
| 1000 | 100 | 0.447 | 0.0276 |
| 1000 | 1000 | 4.44 | 0.353 |
| 10000 | 10 | 0.0843 | 0.0138 |
| 10000 | 100 | 0.790 | 0.194 |
| 10000 | 1000 | 10.0 | 2.85 |
| 100000 | 10 | 0.376 | 0.120 |
| 100000 | 100 | 4.11 | 2.38 |
| 100000 | 1000 | 41.0 | 26.2 |

## Context

Vijay and I noticed a significant slowdown on the training of GATs for graph regression on ZINC in Benchmarking GNNs.  The slowdown goes as much as 100% (depending on the machine).  With `CUDA_LAUNCH_BLOCKING=1`, I further figured out that fetching a batched graph from the dataloader increases from 13ms to 89ms.  The forward propagation also increases from 29ms to 41ms per iteration, and the backward propagation increases from 31ms to 43ms, but that's a separate issue.

yzh119:
## Description
To fix #2145 

## Checklist
- [x] Code is well-documented


src/graph/transform/union_partition.cc
```diff
@@ -18,7 +18,7 @@ HeteroGraphPtr JointUnionHeteroGraph(
   for (dgl_type_t etype = 0; etype < meta_graph->NumEdges(); ++etype) {
     auto pair = meta_graph->FindEdge(etype);
     const dgl_type_t src_vtype = pair.first;
-    const dgl_type_t dst_vtype = pair.second;
+    const dgl_type_t dst_vtype = pair.second;
```
yzh119, line 21: @classicsong I'm wondering which functions are (not) in use in this file?

BarclayII:
Turns out that this implementation is still twice as slow as the one in 0.4.2:
image

I also tried batching with features (`g.ndata['h'] = torch.randn(100, 50); g.edata['h'] = torch.randn(E, 20)`) and the result is similar.

It seems that the reason is that `number_of_nodes()` is also more than 10 times slower.  Is it possible to get around by not using `number_of_nodes()` during batching (for instance, by directly operating on frames)?

yzh119:
What configuration you are using? (E=?, B=?)
To disable data frame concatenation, you have to explicitly set `dgl.batch(graphs, ndata=None, edata=None)`.

Cache the `num_nodes` and `num_edges` on python end somewhat helps, but not enough IMO.

BarclayII:
I tried a grid of E = [1000, 10000, 100000] and B = [10, 100, 1000].

Can we change `_batch_feat_dicts` to accept empty dictionaries, so that we don't have to call `number_of_nodes/edges` (although I don't know if this is a sane solution)?  `g.ndata` returns an empty dictionary if no feature is assigned to it.  Also, assigning features to an empty graph is also allowed so you don't have to handle empty graphs separately:
```python
g = dgl.graph(([], []), num_nodes=0)
g.ndata['x'] = torch.randn(0, 4)
g.ndata    # {'x': tensor([], size=(0, 4))}
```

yzh119:
@jermainewang , @BarclayII , use `cython` instead of `ctypes` greatly reduce the ffi overhead.

This PR w/ cython
```
1000 10 0.0059814453125
1000 100 0.03812980651855469
1000 1000 0.44299983978271484
10000 10 0.011900186538696289
10000 100 0.1617600917816162
10000 1000 2.3301467895507812
100000 10 0.08980894088745117
100000 100 2.0322461128234863
100000 1000 21.827609300613403
```

this PR w/o cython
```
1000 10 0.013225317001342773
1000 100 0.09706544876098633
1000 1000 1.0208745002746582
10000 10 0.022931337356567383
10000 100 0.26139378547668457
10000 1000 3.2776715755462646
100000 10 0.10930180549621582
100000 100 2.4994430541992188
100000 1000 22.912336587905884
```

dgl 0.5
```
1000 10 0.031327247619628906
1000 100 0.16253089904785156
1000 1000 1.6365175247192383
10000 10 0.06584453582763672
10000 100 0.6080005168914795
10000 1000 7.752443552017212
100000 10 0.5082671642303467
100000 100 5.105277061462402
100000 1000 52.8203604221344
```

dgl 0.4.3
```
1000 10 0.004023075103759766
1000 100 0.041051626205444336
1000 1000 0.25679492950439453
10000 10 0.006957292556762695
10000 100 0.06652617454528809
10000 1000 1.7245545387268066
100000 10 0.0612950325012207
100000 100 1.5573503971099854
100000 1000 16.25874972343445
```

yzh119:
Tensorflow GPU bug still not fixed.

Issue #2128, opened 2020-08-30
cannot request out_edges() for empty node sets on cuda


chwan-rice:
## 🐛 Bug

I'm using the latest DGL version. The following code throws a DGLError: Check failed: dim != 0 (0 vs. 0) : 

``` Python
import dgl
import torch

g = dgl.DGLGraph(([0, 1], [1, 2])).to('cuda:0')
idx = torch.LongTensor([]).to('cuda:0')
g.out_edges(idx)

```

However, if the graph runs on CPU, the code works well. I'm not sure whether this is a bug.

VoVAllen:
Empty tensor should not be allowed in `out_edges` function. Do you have any expected behavior here?

chwan-rice:
Actually, I just feel it's strange that `out_edges` has different behavior on CPU and GPU. It would be better if it returns an empty edge set.

chwan-rice:
A similar problem is found in `out_degrees`

BarclayII:
Should be fixed in master.  Will go into 0.5.2 release.

PR #2179, opened 2020-09-11
[Bug] fix cumsum on an empty array with prepend_zero returning an empty array

BarclayII:
## Description
Currently `aten::CumSum` will return an empty array if the input is empty, regardless of the value of `prepend_zero`.  This PR makes it return `[0]` if `prepend_zero` is True.

This is another reason of the bug #2128 .


tests/compute/test_heterograph.py
```diff
@@ -365,6 +365,40 @@ def _test(g):
...
 
+@parametrize_dtype
+def test_empty_query(idtype):
+    g = dgl.graph(([1, 2, 3], [0, 4, 5]), idtype=idtype, device=F.ctx())
+    g.add_nodes(0)
+    g.add_edges([], [])
+    g.remove_edges([])
+    g.remove_nodes([])
+    assert F.shape(g.has_nodes([])) == (0,)
+    assert F.shape(g.has_edges_between([], [])) == (0,)
```
yzh119, line 376: MXNet do not support summation on a empty array: https://github.com/dmlc/dgl/blob/master/python/dgl/heterograph.py#L2568, resulting in this error.

yzh119:
Updated, this should work for now.

PR #2180, opened 2020-09-11
[Do not merge] The set of data is shuffled or is not shuffled

mozga-intel:
The dataloader_shuffle_data flag was added to enable/disable the shuffle data.

### DIFF
--- a/examples/pytorch/gcmc/train_sampling.py
+++ b/examples/pytorch/gcmc/train_sampling.py
@@ -174,6 +174,7 @@ def config():
                         help='The dataset name: ml-100k, ml-1m, ml-10m')
     parser.add_argument('--data_test_ratio', type=float, default=0.1) ## for ml-100k the test ration is 0.2
     parser.add_argument('--data_valid_ratio', type=float, default=0.1)
+    parser.add_argument('--dataloader_shuffle_data', type=bool, default=True)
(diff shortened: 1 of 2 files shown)


BarclayII:
When should we set it to False?

IzabelaMazur:
It should be False when you want to have a constant set of data for tests. Not to randomize the data but to have the same dataset all the time. To be able to compare performance correctly.

BarclayII:
If you look at Line 224 in the GCMC example, you will see that the EdgeDataLoader works on `train_enc_graph`, which consists of the edges from only the training set.  The test set is separately handled in another object called `test_dataloader`.

GraphSAGE example is similar.

IzabelaMazur:
I am sorry for the incorrect expression. I didn't mean the test during the training, but to run the whole benchmark. We use it to check if performance changes in the dgl code bring any benefits. That's why we need a constant set of data to be able to compare results before and after the changes. If we have the data randomization enabled once the training time will be longer once shorter and we don't know whether this is due to our changes or to selection of other data.

BarclayII:
Got it.  I would prefer to keep this change in your fork for benchmarking purposes, instead of merging it into our codebase since fixing the iteration order does not bring any particular benefit from the modeling perspective.

PR #2197, opened 2020-09-15
[Distributed] Change rsync to scp of copy_files.py

aksnzhy:
## Description
Change rsync to scp of copy_files.py because rsync is very slow for big files.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/tools/copy_files.py
+++ b/tools/copy_files.py
@@ -9,9 +9,9 @@ import logging
 import json
 import copy
 
-def copy_file(file_name, ip, workspace):
+def copy_file(file_name, ip, workspace, param):
     print('copy {} to {}'.format(file_name, ip + ':' + workspace + '/'))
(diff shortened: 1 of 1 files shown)


VoVAllen:
It's weird rsync is slower than scp. How much slower did you find?

aksnzhy:
> It's weird rsync is slower than scp. How much slower did you find?

If the file is already in the remote machine, sync will do diff before sending the file.

VoVAllen:
You can remove the `-c` option, which will only use file and creation time to diff, without calculating the checksum

aksnzhy:
> You can remove the `-c` option, which will only use file and creation time to diff, without calculating the checksum

It seems that we don't use `-c` option here?

VoVAllen:
The option here is ` -arvc`, removing c should be `-arv`

PR #2220, opened 2020-09-21
[Bug] fix multiple bugs in JTNN example

BarclayII:
Several bugs in previous refactor of JTNN led to significant performance regression:

1. `pull(v, msg_func, reduce_func, apply_func)` should be refactored into `pull(msg_func, reduce_func); apply_nodes(apply_func, v=v)`.
2. I still missed updates from `g.edata` to `lg.ndata` or vice versa in a couple places.

@mufeili I ran a couple iterations and saw training accuracies (word, topo, assm, stereo) matching the numbers from 0.4.  Also compared and verified the closeness of outputs against 0.4 on the same initialization and inputs.  Please give this PR a try.


yzh119:
@BarclayII Any speed difference?

BarclayII:
DGL 0.4 takes only 55% of the time of DGL 0.5.  The reason seems to be in multiple places.

Here is the result of 0.5 when running 100 iterations:
image

And that from 0.4:
image

I would say to merge the PR first if the code is correct.  Then we take time to look into why it becomes that slow.

PR #2228, opened 2020-09-23
Fix segment_reduce() ignoring tailing 0 segments

mjwen:
## Description

[segment_reduce()](https://github.com/dmlc/dgl/blob/7e0107c395b24fa55e080dee3903552430008b48/python/dgl/ops/segment.py#L9) supports zeor-length segments. But it fails if the zero-length segments is at the end of `seglen`. 

For example,
```python
import dgl
import torch

val = torch.ones(10, 3)
seg = torch.tensor([1, 0, 5, 4, 0])
x = dgl.ops.segment.segment_reduce(seg, val, "sum")
```

yields a value for `x` as 
```
tensor([[1., 1., 1.],
        [0., 0., 0.],
        [5., 5., 5.],
        [4., 4., 4.]])
```
but it should be

```
tensor([[1., 1., 1.],
        [0., 0., 0.],
        [5., 5., 5.],
        [4., 4., 4.],
        [0., 0., 0.]])
```

This PR fixes this.

### DIFF
--- a/python/dgl/ops/segment.py
+++ b/python/dgl/ops/segment.py
@@ -50,7 +50,8 @@ def segment_reduce(seglen, value, reducer='sum'):
     if len(u) != len(v):
         raise DGLError("Invalid seglen array:", seglen,
                        ". Its summation must be equal to value.shape[0].")
-    g = convert.heterograph({('_U', '_E', '_V'): (u, v)})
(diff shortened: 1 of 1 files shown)


BarclayII:
Thanks for your fix!

PR #2234, opened 2020-09-25
[Example]Add sparse embedding in GATNE-T example to support big graph

sangyx:
## Description

Hi, I have modified the code of GATNE-T to support a big graph.

## Changes

- [x] Use multiprocessing to read data.
- [x] Remove duplicate pairs and store them in `np.ndarray` to save memory.
- [x] Modify the `num_workers` and `pin_memory` to accelerate training.
- [x] Use sparse embedding and sparse optimizer to avoid `cuda out of memory` in backward.
- [x] Format code with black.

I have tested the code for the example dataset and it can achieve the performance previously reported in the readme.


sangyx:
I have also added some code to support training with multiple GPUs. But it is worth noting that DistributedDataParallel will cause more Cuda memory consumption and a certain loss of performance (About 3% in AUC in the example dataset).

yzh119:
Actually, DDP should not cause more GPU memory consumption, would you mind elaborating more on the GPU statistics?
(e.g. if the memory footprint on a single GPU is x GB, when training in a multi-GPU setting, will the GPU memory consumption on each GPU be greater then x? ).

sangyx:
When training on example dataset, 1134MiB for a single GPU(`run_example_sparse.sh`), 1178MiB for two GPU(`run_example_sparse_multi_gpu.sh`). Batch_size is 256, GPU is Tesla V100-SXM2. The difference between them will increase with the increase of the data set size and batch_size.

BarclayII:
This looks fine to me.  @yzh119 What do you think?

yzh119 (approved):
LGTM.

PR #2262, opened 2020-10-06
[bugfix] Fix crash due to unfound attribute

Samiisd:
## Description

An unexpected `AttributeError` exception is launched whenever the user defines `save_dir` on `DGLDataset` instantiation and uses `self._save_dir`  within `has_cache` method.

## Changes

Initializes the attribute `self._save_dir` in all cases.


BarclayII:
Thanks!

PR #2263, opened 2020-10-06
[GPU][Allocations] Use shared workspace for normal GPU allocations

nv-dlasalle:
## Description
Repeated `cudaMalloc()` and `cudaFree()` calls can be expensive, and in addition, cudaFree() requires synchronizing with the device, which can prevent overlapping CPU and GPU work. This PR makes it so all CUDA allocations go through a `WorkspacePool::Pool` object, ensuring we re-use allocations many times, amortizing the cost of `cudaMalloc()` and `cudaFree()`, and avoid synchronizing with the device unnecessarily. These changes are only enabled through the environment variable `DGL_USE_CUDA_MEMORY_POOL`, otherwise the regular allocation path is taken.

In the worst case, this allocator uses 4x the memory necessary, by having half of the allocations in use, and all of those only using 50% of the allocated space. However, this has little impact on the overall memory usage, as the graph index arrays (which is almost all of the data allocated DGL), are much smaller than the weights and feature data allocated by the backend.

I could refactor this PR around the "TensorDispatcher" in https://github.com/dmlc/dgl/pull/2328, and have the memory pool here simply be another allocator that can be selected.

## Checklist
- [x] All changes have test coverage (looking at the current testing code, I believe both the CPU and GPU allocations are executed, but I can't say I'm 100% sure).
- [x] Code is well-documented

## Changes

This PR adds two new methods to `DeviceAPI`: `AllocRawDataSpace()` and `FreeRawDataSpace()` to explicitly by-pass workspace allocations, where as the `AllocDataSpace()` and `FreeDataSpace()` methods in the `CUDADeviceAPI` class now uses a shared workspace to make all allocations.

Below are two profiles from Nvidia Nsight Systems, showing the GPU utilization in the forward pass of RGCN (this is the CSR generation and SpMM operations):

Master branch:
cuda_without_workspace_profile

This PR:
cuda_workspace_profile

### DIFF
--- a/include/dgl/runtime/device_api.h
+++ b/include/dgl/runtime/device_api.h
@@ -82,8 +82,7 @@ class DeviceAPI {
   virtual void* AllocDataSpace(DGLContext ctx,
                                size_t nbytes,
                                size_t alignment,
-                               DGLType type_hint)
-  {
+                               DGLType type_hint) {
     return AllocRawDataSpace(ctx, nbytes, alignment, type_hint);
   }
   /*!
@@ -97,8 +96,7 @@ class DeviceAPI {
    * \param ctx The device context to perform operation.
    * \param ptr The data space.
    */
-  virtual void FreeDataSpace(DGLContext ctx, void* ptr)
-  {
+  virtual void FreeDataSpace(DGLContext ctx, void* ptr) {
       FreeRawDataSpace(ctx, ptr);
   }
   /*!
--- a/include/dgl/runtime/device_api.h
+++ b/include/dgl/runtime/device_api.h
@@ -58,6 +58,19 @@ class DeviceAPI {
    */
   virtual void GetAttr(DGLContext ctx, DeviceAttrKind kind, DGLRetValue* rv) = 0;
   /*!
+   * \brief Allocate a data space on device using system mechanisms.
+   * \param ctx The device context to perform operation.
+   * \param nbytes The number of bytes in memory.
+   * \param alignment The alignment of the memory.
+   * \param type_hint The type of elements. Only needed by certain backends such
+   * as OpenGL, as nbytes & alignment are sufficient for most backends.
+   * \return The allocated device pointer.
+   */
+  virtual void* AllocRawDataSpace(DGLContext ctx,
+                               size_t nbytes,
+                               size_t alignment,
+                               DGLType type_hint) = 0;
+  /*!
    * \brief Allocate a data space on device.
    * \param ctx The device context to perform operation.
    * \param nbytes The number of bytes in memory.
@@ -69,13 +82,25 @@ class DeviceAPI {
   virtual void* AllocDataSpace(DGLContext ctx,
                                size_t nbytes,
                                size_t alignment,
-                               DGLType type_hint) = 0;
+                               DGLType type_hint)
+  {
+    return AllocRawDataSpace(ctx, nbytes, alignment, type_hint);
+  }
   /*!
-   * \brief Free a data space on device.
+   * \brief Free a data space on device using system mechanisms.
    * \param ctx The device context to perform operation.
    * \param ptr The data space.
    */
-  virtual void FreeDataSpace(DGLContext ctx, void* ptr) = 0;
+  virtual void FreeRawDataSpace(DGLContext ctx, void* ptr) = 0;
+  /*!
+   * \brief Free a data space on the device.
+   * \param ctx The device context to perform operation.
+   * \param ptr The data space.
+   */
+  virtual void FreeDataSpace(DGLContext ctx, void* ptr)
+  {
+      FreeRawDataSpace(ctx, ptr);
+  }
   /*!
    * \brief copy data from one place to another
    * \param from The source array.
--- a/src/runtime/cuda/cuda_device_api.cc
+++ b/src/runtime/cuda/cuda_device_api.cc
@@ -14,32 +14,28 @@
 namespace dgl {
 namespace runtime {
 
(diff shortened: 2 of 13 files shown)


src/runtime/cuda/cuda_device_api.cc
```diff
@@ -8,11 +8,38 @@
...
+    return pool_.AllocWorkspace(ctx, size);
+  }
+
+  void FreeWorkspace(DGLContext ctx, void* ptr) {
+    std::lock_guard<std::mutex> guard(mtx_);
+    pool_.FreeWorkspace(ctx, ptr);
+  }
+
+ private:
+  WorkspacePool pool_;
```
zheng-da, line 37: is this CUDA workspace pol?

nv-dlasalle: Right, this workspace pool only gets created with `kDLGPU` and `CUDADeviceAPI::Global()`.

include/dgl/runtime/device_api.h
```diff
@@ -57,6 +57,19 @@ class DeviceAPI {
...
+   * \param type_hint The type of elements. Only needed by certain backends such
+   * as OpenGL, as nbytes & alignment are sufficient for most backends.
+   * \return The allocated device pointer.
+   */
+  virtual void* AllocRawDataSpace(DGLContext ctx,
+                               size_t nbytes,
+                               size_t alignment,
+                               DGLType type_hint) = 0;
   /*!
    * \brief Allocate a data space on device.
```
BarclayII, line 74: Maybe state the difference between this and `AllocRawDataSpace`?  Something like "Allocate a data space on device, potentially with a global workspace caching allocator".

include/dgl/runtime/device_api.h
```diff
@@ -69,13 +82,23 @@ class DeviceAPI {
...
+  }
+  /*!
+   * \brief Free a data space on device using system mechanisms.
+   * \param ctx The device context to perform operation.
+   * \param ptr The data space.
+   */
+  virtual void FreeRawDataSpace(DGLContext ctx, void* ptr) = 0;
   /*!
-   * \brief Free a data space on device.
+   * \brief Free a data space on the device.
```
BarclayII, line 95: Same here

BarclayII (commented):
Went through the PR.  It essentially replaces `AllocDataSpace` and `FreeDataSpace` with a `WorkspacePool`-based allocator.  The only place that calls `Alloc/FreeDataSpace` is `NDArray::Empty` which is a frequent operation especially for temporary array allocations.  Presumably this is why this PR brings better utilization.

I'm wondering if we can make the intermediate array allocations base on `Alloc/FreeWorkspace` instead of `Alloc/FreeDataSpace`.  This way we can potentially have thread-local allocators and avoid locking, but it will require lots of changes in C++ code.

I'm personally good with this PR.  cc @jermainewang for your opinion.

jermainewang:
The potential concern is the efficiency of the workspace pool, as is stated in the comment:
https://github.com/dmlc/dgl/blob/7d19b33c2446e2b36d95c9d8e35309c100ffa448/src/runtime/workspace_pool.h#L15-L25

Current memory pool strategy is quite naive. It maintains a sorted list of free memory chunks in ascending order of the chunk size. The cost for insertion (upon `Free`) and deletion (upon `Alloc`) is O(N), where N is the free list size. The cost will be quite high when there are a lot of allocations and this is why it is previously used **only** in temporary workspace memory. The PR essentially changes the behavior of all `AllocDataSpace` to this pool-based allocator, so I'm afraid that may cause a large pool size.

@nv-dlasalle Could you also measure the maximum size of `free_list_` before and after the PR for a typical workload like GraphSAGE?

Another potential issue is since DGL's memory pool is not visible to the memory management of backend framework, it may result in more memory consumption than actually needed. The problem has already existed before this PR but the PR will amplify it since more memory allocation will go through the pool-based allocator.

zheng-da:
It's going to be very difficult to manage two memory pools that don't know each other. can we allocate memory from Pytorch's memory pool directly?

BarclayII:
> It's going to be very difficult to manage two memory pools that don't know each other. can we allocate memory from Pytorch's memory pool directly?

That will be hard if we do not compile and link against the backend library themselves.  Take PyTorch as an example, the canonical way to allocate a PyTorch tensor is to call `at::empty()` which invokes the underlying methods in `CUDADeviceAllocator` or `CPUAllocator`.  MXNet and Tensorflow should have similar mechanisms since they all support writing C++ operators.

nv-dlasalle:
@jermainewang You're right, this poorly utilizes the memory in graphsage (and likely many other cases):

```
Unused entries: 1(0 bytes) / 21(26976256 bytes)
Unused bytes: 22274047 / 26976256
```

While all of the allocations get used at once during an epoch (the 1 entry is the null entry at the start of the list), they are sized poorly (22 of the 27 MB allocated is space at the end of an allocation).

jermainewang:
@nv-dlasalle It looks like you've added an env var for the pool-based allocator. Thanks! Have you done the change?

nv-dlasalle:
@jermainewang I have, but I still need to add unit tests to cover the memory pool, now that it's not used in the normal tests. Where do you think such tests should be placed within the existing unit tests?

jermainewang:
You could add a c++ test here: https://github.com/dmlc/dgl/tree/master/tests/cpp.

jermainewang (approved):

jermainewang:
A similar solution has been implemented in #2328 .

PR #2265, opened 2020-10-07
Fix typo

chwan-rice:
I found some typoes in DGL.


BarclayII:
Is this PR finished?  It seems that the history is messed up somehow (which is fine)

chwan-rice:
Yes, it's finished. This branch was developed on another branch. You can check the changed files.

python/dgl/heterograph.py
```diff
@@ -68,7 +68,7 @@ def __init__(self,
             raise DGLError('The input is already a DGLGraph. No need to create it again.')
         if not isinstance(gidx, heterograph_index.HeteroGraphIndex):
             dgl_warning('Recommend creating graphs by `dgl.graph(data)`'
-                        ' instead of `dgl.DGLGraph(data)`.')
```
jermainewang, line 71: This is intended because DGLHeteroGraph and DGLGraph are the same. We actually have an alias in the root package. We will definitely change the class name in the future.

jermainewang:
@chwan-rice Sorry for taking this long. Could you address my comment and I will merge it.

chwan-rice:
Thanks for the comment. I have fixed it.

PR #2266, opened 2020-10-07
[Performance] In HeteroNodeView, build arange on target device, instead of on CPU and copying it

nv-dlasalle:
## Description
Create the arange on the target device directly using the backend, rather than creating it on the CPU, and copying it the target device.

## Checklist
- [x] All changes have test coverage (covered by test_basics.py's `test_update_routines()`, `test_update_0deg()`, etc. )
- [x] Code is well-documented

## Changes
The `arange` method in the backends was modified to accept a device context, and the HeteroNodeView was modified to make use of this, to avoid generating the range on the CPU, and then synchronously copying it up to the GPU.

### DIFF
--- a/python/dgl/backend/tensorflow/tensor.py
+++ b/python/dgl/backend/tensorflow/tensor.py
@@ -421,8 +421,10 @@ def sort_1d(input):
     return tf.sort(input), tf.cast(tf.argsort(input), dtype=tf.int64)
 
 
-def arange(start, stop, dtype=tf.int64):
-    with tf.device("/cpu:0"):
+def arange(start, stop, dtype=tf.int64, ctx=None):
+    if not ctx:
+        ctx = "/cpu:0"
+    with tf.device(ctx):
         t = tf.range(start, stop, dtype=dtype)
     return t
 
--- a/python/dgl/backend/tensorflow/tensor.py
+++ b/python/dgl/backend/tensorflow/tensor.py
@@ -454,7 +454,8 @@ def zerocopy_to_dgl_ndarray(data):
         elif data.dtype == tf.int64:
             data = tf.cast(data, tf.uint64)
         return nd.cast_to_signed(nd.from_dlpack(zerocopy_to_dlpack(data)))
-    return nd.from_dlpack(zerocopy_to_dlpack(data))
+    else:
+        return nd.from_dlpack(zerocopy_to_dlpack(data))
 
 
 def zerocopy_to_dgl_ndarray_for_write(input):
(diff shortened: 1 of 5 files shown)


BarclayII:
How was the speed up?  Just curious; not knowing the number is fine.

nv-dlasalle:
The overall speed up is pretty small, but in the region immediately following its in invocation in the graphsage example, we see a reduction from 1.8ms to 547us (> 3x), as a result of not having to wait for the H2D copy:

Before this change:
original_arange

With this change:
this_branch_arange

But as more synchronization points are removed, the impact could be larger. In the above images, synchronizing to check if the COO is sorted is what limits the benefit.

python/dgl/backend/tensorflow/tensor.py
```diff
@@ -444,13 +446,15 @@ def zerocopy_from_numpy(np_array):
...
+        # NOTE: TF doesn't keep signed tensors on GPU due to legacy issues with
+        #   shape inference. Convert it to unsigned and cast it back afterwards.
+        if data.dtype == tf.int32:
+            data = tf.cast(data, tf.uint32)
+        elif data.dtype == tf.int64:
+            data = tf.cast(data, tf.uint64)
         return nd.cast_to_signed(nd.from_dlpack(zerocopy_to_dlpack(data)))
-    else:
-        return nd.from_dlpack(zerocopy_to_dlpack(data))
+    return nd.from_dlpack(zerocopy_to_dlpack(data))
```
zheng-da, line 457: why do we need this modification?

BarclayII: I think the effect is the same.  Do you prefer keeping the old code?  It doesn't look like a big deal to me.

BarclayII (approved):
I'm good after Da's comment is resolved.

python/dgl/backend/tensorflow/tensor.py
VoVAllen: It's a potential bug(feature?) in tensorflow dlpack, the device of int32 tensor is not accurate. I think this is fixed in the latest tf, but we need to keep it for backward compatibility

nv-dlasalle: I reverted the change the else clause in 2abb841.

With respect as to why this method needed to be modified, in the unit tests for int64_t I ran into the issue that the pointer in the tensor was silently converted to a CPU pointer rather than a GPU pointer. The code was already handling the in32_t case, so I expanded it to also handle the int64_t case. This was with TensorFlow 2.3.1.

This wasn't an issue in the unit tests before this PR, as the tensor was generated on the CPU, and then explicitly copied to the GPU.

PR #2281, opened 2020-10-15
[Feature][WIP] JAX Backend

yuanqing-wang:
## Description
Attempt to implement JAX backend.

## Checklist
- [ x ] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] Code is well-documented


python/dgl/backend/jax/sparse.py
```diff
@@ -0,0 +1,79 @@
...
+
+def _addsub(op, x):
+    return -x if op == 'sub' else x
+
+
+def _expand(x, shape):
+    return x.expand(-1, *shape)
+
+def gspmm(gidx, op, reduce_op, X, Y):
+    out, (argX, argY) = _gspmm(gidx, op, reduce_op, X, Y)
```
yzh119, line 63: Not that `_gspmm` and `_gsddmm` do not handle gradients.
In PyTorch backend we use `torch.autograd.Function` to manipulate gradients: https://github.com/dmlc/dgl/blob/master/python/dgl/backend/pytorch/sparse.py#L70-L118

yuanqing-wang: Yes! I was thinking about making sure all the forward passes are working properly before worrying about the gradients. One thing that's a bit tricky is that if we use things like `jax.custom_vjp` we'll have to think about how to cache and reuse stuff from the forward pass like this one: https://github.com/dmlc/dgl/blob/da9f14dff38c9939958ab1126438b7be2c3c39fc/python/dgl/backend/pytorch/sparse.py#L65

jermainewang:
Since this will be a major update, how about we setting up a meeting on how to move forward?

yuanqing-wang:
@jermainewang I'd love to chat as well! When would be a good time to chat?

yuanqing-wang:
Hi @yzh119 just wanted to catch your attention on this again since we had spoken two years ago. 

I've spent some time on this and it looks like this backend is fully functional, passing all unittests. I've also implemented corresponding NN modules. Let me know if you would be interested in taking another look at it!

yzh119:
Thank @yuanqing-wang for the effort, this is huge.

And I love the idea of having a JAX backend, I encourage some full-time DGL team members to take look at this and think about the future of the new backend: @mufeili @Rhett-Ying @BarclayII @jermainewang .

I can help review this PR as well.

PR #2284, opened 2020-10-15
[Dataloading] Add class for copying tensors to/from the GPU on a non-default stream

nv-dlasalle:
## Description
The purpose of this PR is to allow overlapping copying data to/from the GPU, with performing computations on the GPU.

## Checklist
- [x] Code is well-documented

## Changes
This adds the AsyncTransferer class to `dgl.sampling`, as it is to be used by data loaders.


include/dgl/runtime/ndarray.h
```diff
@@ -139,11 +139,14 @@ class NDArray {
...
    * \brief Copy data content from another array.
    * \param other The source array to be copied from.
+   * \param stream The stream to perform the copy on.
    * \note The copy may happen asynchrously if it involves a GPU context.
    *       DGLSynchronize is necessary.
    */
-  inline void CopyFrom(DLTensor* other);
-  inline void CopyFrom(const NDArray& other);
+  inline void CopyFrom(DLTensor* other,
+      DGLStreamHandle stream = nullptr);
```
zheng-da, line 147: how does this work for CPU?

nv-dlasalle: @zheng-da I expanded the comment for the stream parameter to:
```
   * \param stream The stream to perform the copy on if it involves a GPU
   *        context, otherwise this parameter is ignored.
```

in 286ec68 .

src/dataloading/async_transferer.cc
```diff
@@ -0,0 +1,137 @@
...
+
+  return id;
+}
+
+NDArray AsyncTransferer::Wait(
+    const TransferId id) {
+  auto iter = transfers_.find(id);
+  CHECK(iter != transfers_.end()) << "Unknown transfer: " << id;
+
+  Transfer t = std::move(iter->second);
```
zheng-da, line 96: is the first transmission request always completed before the second one?

zheng-da (approved):

zheng-da:
please fix CI

VoVAllen:
You need to change `F.ones([100,75,25], dtype=F.int32).cpu()` to `F.copy_to(F.ones([100,75,25], dtype=F.int32), F.cpu())`, since `.cpu()` is not supported on mxnet

python/dgl/dataloading/async_transferer.py
```diff
@@ -0,0 +1,85 @@
+""" API for transferring data to/from the GPU over second stream.A """
```
jermainewang, line 1: Could you add some module docstring on code examples of `Transfer` and `AsyncTransferer`? Please also add it to the API doc page [here](https://github.com/dmlc/dgl/blob/master/docs/source/api/python/dgl.dataloading.rst).

jermainewang:
Good with the RP overall. One comment about documentation.

python/dgl/dataloading/async_transferer.py
nv-dlasalle: Added in b1765ce

docs/source/api/python/dgl.dataloading.rst
```diff
@@ -48,3 +48,17 @@ to generate negative edges.
...
+
+Data can be copied from the CPU to the GPU, or from the GPU to the CPU,
+while the GPU is being used for
+computation, using the :class:`AsyncTransferer`.
+
+.. autoclass:: AsyncTransferer
+    :members: __init__, async_copy
+
+.. autoclass:: async_transferer.Transfer
+    :members: wait
```
zheng-da, line 64: should these APIs exposed to users?

jermainewang: generally I feel people are reading all the public codes so there is hardly a line between internal and external APIs. I would well-doc it assuming people will use it.

zheng-da: once it is documented, i would consider that we officially announce it as part of DGL API. Then it's our responsibility to make the API stable.

PR #2285, opened 2020-10-16
[Bugfix][Frame] Fix typos

ZenoTan:
## Checklist
- [ ] Code is well-documented

## Changes

- Fix some typos in `python/dgl/frame.py`

### DIFF
--- a/python/dgl/frame.py
+++ b/python/dgl/frame.py
@@ -168,7 +168,7 @@ class Column(object):
         Tensor
             The feature data
         """
-        return F.gather_row(self.data, rwoids)
+        return F.gather_row(self.data, rowids)
 
     def __setitem__(self, rowids, feats):
         """Update the feature data given the index.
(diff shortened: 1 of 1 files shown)


python/dgl/frame.py
```diff
@@ -183,7 +183,7 @@ def __setitem__(self, rowids, feats):
         feats : Tensor
             New features.
         """
-        self.update(idx, feats)
```
zheng-da, line 186: why this kind of bug hasn't been triggered before?

ZenoTan: Maybe previously code in `_deprecate` is used? I think we should add some tests on this.

yzh119: I think they are used in the old `send_and_recv` functions which are no longer supported.

yzh119 (approved):
Thanks for your contribution.

python/dgl/frame.py
ZenoTan: Will it be triggered if someone updates graph feature manually?

PR #2291, opened 2020-10-20
[Performance] Dynamic cpu kernel for SpMMSumCsr 

pawelpiotrowicz:
* support AVX512

* env DGL_CPU_INTEL_KERNEL_ENABLED=1

* env DGL_CPU_INTEL_KERNEL_LOG=1

* no specialization

## Description
This feature allows to use dynamic kernels.  
Depends on dimension you can create a dedicated assembly code without unnecessary checks and conditions – it helps CPU to be faster.  
Currently there is only one default code for all types of dimensions, specializations are in progress.
The engine uses AVX512 to compute element_wise(). 
To activate the feature set an environment variable DGL_CPU_INTEL_KERNEL_ENABLED=1 
The cache with the code is assigned to main thread and distributed between OMP pool threads, 
if dimension will be changed and a specialization exists then a new kernel will be generated and assigned as a main kernel.

### DIFF
--- a/include/intel/cpu_support.h
+++ b/include/intel/cpu_support.h
@@ -1,0 +1,156 @@
+/*!
+ *  Copyright (c) 2019 by Contributors
+ * \file intel/cpu_support.h
+ * \brief Intel CPU support
+ */
+#ifndef INTEL_CPU_SUPPORT_H_
+#define INTEL_CPU_SUPPORT_H_
+#include <memory>
+#include <type_traits>
+#include "xbyak/xbyak.h"
+#include "xbyak/xbyak_util.h"
+namespace intel {
+#define log_intel(x) if (IntelKernel<T>::log_enabled()) { std::cout << x << std::endl; }
+#ifndef log_intel
+#define log_intel(x)
+#endif
+template<class T>
+using Uptr = std::unique_ptr<T>;
+
+template<class T>
+struct IntelKernel {
+      static int64_t getValue() {
+        int64_t v = 0;
+        const char *label = "DGL_CPU_INTEL_KERNEL_ENABLED";
+        const char *ptr = std::getenv(label);
+        if (ptr) {
+          v = atoll(ptr);
+          log_intel(label << "=>" << v);
+        }
+        return v;
+      }
+
+      static int64_t enabled() {
+        static int64_t r = IntelKernel<T>::getValue();
+        return r;
+      }
+
+      static int log_enabled() {
+        static int r = (std::getenv("DGL_CPU_INTEL_KERNEL_LOG")) ? 1 : 0;
+        return r;
+      }
+};
+
+template <class T, int has_specialization = 0>
+class elem_wise_add_update : public Xbyak::CodeGenerator {
+    typedef elem_wise_add_update<T> self;
+    /* [performance] skip static check*/
+    int64_t size;
+
+    /* [functional] Does kernel is applicable on this machine ? */
+    bool applicable;
+
+    /* [performance-functional] use kernel when input size >= min_size */
+    int64_t min_size;
+
+ public:
+     void prolog() {
+        push(r8);
+        push(r9);
+     }
+
+      void epilog() {
+        pop(r9);
+        pop(r8);
+      }
+    explicit elem_wise_add_update(int64_t _size) :
+     size(_size), applicable(false), min_size(IntelKernel<T>::enabled()) {
+      static Xbyak::util::Cpu current_cpu;
+      /* input => { src=RSI , dst=RDI , size=RDX } */
+
+      /* Default case for all */
+      if (current_cpu.has(Xbyak::util::Cpu::tAVX512F)) {
+        prolog();
+        /* prepare REMAINDER */
+        mov(r8, rdx);  // rdx => size
+        and_(r8, 0xf);  // r8_modulo = size/(sizeof(zmm)/sizeof(float))
+        xor_(r9, r9);  // reset r9
+        cmp(rdx, 0x10);  // if ( size < 16 ) {  }
+        jl("remainder");
+
+        /*  decrease  divident */
+        sub(rdx, r8);  // prepare alignment chunks
+        cmp(rdx, 0);  // do we have any full chunks ?
+        jz("remainder");
+
+        L("for_i");
+        /* load first part of mem to zmm0*/
+        vmovups(zmm0, ptr[rdi + r9 * 4]);
+        /* load second part of mem to zmm1*/
+        vmovups(zmm1, ptr[rsi + r9 * 4]);
+        /* zmm2 = zmm0 + zmm1 */
+        vaddps(zmm2, zmm0, zmm1);
+        /* save output to dst*/
+        vmovups(ptr[rdi + r9 * 4], zmm2);
+        add(r9, 16);  // r9+=sizeof(zmm)/sizeof(float)
+        cmp(rdx, r9);  // more full chunks ?
+        jnz("for_i");
+
+        L("remainder");
+        cmp(r8, 0);  //  do we have a remainder ?
+        jz("done");
+        xor_(rax, rax);
+        /* prepare a bitmask for k1 */
+        mov(rax, 1);
+        mov(rcx, r8);
+        sal(rax, cl);
+        dec(rax);  // k1= (1 << r8 )-1
+        kmovw(k1, eax);  // set bitmask
+        /* same logic as above but with mask */
+        /* vmovups zmm0{k1},ZMMWORD PTR [rdi+r9*4] */
+        const uint8_t ptr_3[7] = {0x62, 0xb1, 0x7c, 0x49, 0x10, 0x04, 0x8f};
+        db(ptr_3, sizeof(ptr_3) / sizeof(uint8_t));
+        /* vmovups zmm1{k1},ZMMWORD PTR [rsi+r9*4] */
+        const uint8_t ptr_2[7] = {0x62, 0xb1, 0x7c, 0x49, 0x10, 0x0c, 0x8e};
+        db(ptr_2, sizeof(ptr_2) / sizeof(uint8_t));
+        /*  zmm2 = zmm0 + zmm1 */
+        vaddps(zmm2, zmm0, zmm1);
+        /* vmovups ZMMWORD PTR [rdi+r9*4]{k1},zmm2  */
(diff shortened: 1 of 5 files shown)


src/array/cpu/spmm.h
```diff
@@ -57,9 +60,90 @@ void SpMMSumCsr(
...
+ * \param csr The Csr matrix.
+ * \param ufeat The feature on source nodes.
+ * \param efeat The feature on edges.
+ * \param out The result feature on destination nodes.
+ * \note it uses node parallel strategy, different threads are responsible
+ *       for the computation of different nodes.
+ */
+#if !defined(_WIN32)
+template <class IdType>
+struct SpMMSumCsrCore<IdType, float, dgl::aten::cpu::op::CopyLhs<float>> {
```
jermainewang, line 81: Does it support double?

src/array/cpu/spmm.h
```diff
@@ -57,9 +60,90 @@ void SpMMSumCsr(
...
+          /* Prepare an assembler code for this paricular dimension */
+          thread_local intel::Uptr<ElemWiseUpd> th_cpu_spec((intel::IntelKernel<DType>::enabled())
+           ? new ElemWiseUpd(dim) : nullptr);
+          /* Check if the code has been created */
+          ElemWiseUpd *cpu_spec = (th_cpu_spec && th_cpu_spec->is_applicable())
+           ? th_cpu_spec.get() : nullptr;
+          if (cpu_spec && cpu_spec->require_new_instance<bool>(dim)) {
+            th_cpu_spec.reset(new ElemWiseUpd(dim));
+            cpu_spec = (th_cpu_spec && th_cpu_spec->is_applicable()) ? th_cpu_spec.get() : nullptr;
+          }
```
jermainewang, line 104: Please change to use our own ThreadLocalStore. You could refer to the example here: https://github.com/dmlc/dgl/blob/master/src/runtime/cuda/cuda_common.h#L75

1. In the header file `intel_cpu_support.h`, define a structure such as `IntelCPUThreadEntry` with a static function returning a pointer of `IntelCPUThreadEntry`.
    ```c++
    struct IntelCPUThreadEntry {
      // any members you'd like to include
      // get the threadlocal instance
      static IntelCPUThreadEntry* ThreadLocal();
    };
    ```
1. In the a separate .cc file, include `<dmlc/thread_local.h>` and define
    ```c++
    typedef dmlc::ThreadLocalStore<IntelCPUThreadEntry> IntelCPUThreadStore;
    IntelCPUThreadEntry* IntelCPUThreadEntry::ThreadLocal() {
      return IntelCPUThreadStore::Get();
    }
    ```
    See example [here](https://github.com/dmlc/dgl/blob/master/src/runtime/cuda/cuda_device_api.cc#L200).
1. To use the thread local object, just include `intel_cpu_support.h` and call `ThreadLocal()`. See example [here](https://github.com/dmlc/dgl/blob/2c04ecb55a121587f6fc6beb6ce1bb593f7bccc1/src/array/cuda/csr2coo.cu#L25).

The logic for compiling new kernels should be as follows:
1. Check whether the CPU supports AVX512 (i.e., `is_applicable`). The check should not be related to the feature dim.
1. Check whether `dim` is larger than `min_dim`.
1. Check whether kernel for `dim` has been generated. If not, compile the kernel.

I can see 1 and 2 being implemented but not 3. Does `Xbyak::CodeGenerator` automatically cache the kernel of the same `dim`? If not, we should maintain a dictionary in the `IntelCPUThreadEntry` ourselves.

src/array/cpu/spmm.h
```diff
@@ -57,9 +60,90 @@ void SpMMSumCsr(
...
+    const IdType row_start = indptr[rid], row_end = indptr[rid + 1];
+    DType *out_off = O + rid * dim;
+    std::fill(out_off, out_off + dim, 0);
+    for (IdType j = row_start; j < row_end; ++j) {
+      const IdType cid = indices[j];
+      if (!bcast.use_bcast && cpu_spec) {
+          /* Forward to cpu asm kernel */
+          cpu_spec->run(out_off, X + cid * lhs_dim , dim);
+          continue;
+      }
```
jermainewang, line 117: Can we lift this branch out of the loop because it does not depend on the loop variable.

src/array/cpu/spmm.h
```diff
@@ -57,9 +60,90 @@ void SpMMSumCsr(
...
+      for (int64_t k = 0; k < dim; ++k) {
+        const int64_t lhs_add = bcast.use_bcast ? bcast.lhs_offset[k] : k;
+        typedef dgl::aten::cpu::op::CopyLhs<DType> Op;
+        const DType *lhs_off =
+        Op::use_lhs ? X + cid * lhs_dim + lhs_add : nullptr;
+        out_off[k] += Op::Call(lhs_off, nullptr);
+        }
```
jermainewang, line 124: I suppose this part of code is *identical* to the original SpMMCsr which is now in `SpMMCsrCore` w/o specialization. This will cause maintenance burden. I think the easiest way is to not use specialization.
```c++
template <typename IdType, typename DType, typename Op>
void SpMMSumCsr(
    const BcastOff& bcast,
    const CSRMatrix& csr,
    NDArray ufeat, NDArray efeat,
    NDArray out) {
  if (!bcast.use_bcast && AVX512_is_enabled() &&
      operator_is_supported<Op>() && workload_is_suitable(ufeat, efeat, out)) {
     kernel = generate_kernel_for_workload_if_needed(ufeat, efeat, out);
     invoke_kernel(kernel)
  } else {
     // original kernel logic
     ...
  }
}
```
This automatically resolves the issue that those checks are inside the loop.

jermainewang (changes requested):

jermainewang:
Other general problems to fix:
* Please follow the google coding style for naming conventions and indentations. The [clang-format](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) could be helpful. I will try it out and tell you the instructions to autoformat them.
* Please add some basic unit tests. I'm thinking of testing the results when this optimization is on and off and making sure the outcomes are the same.

VoVAllen:
My clang-format setting for dgl is `{ BasedOnStyle: Google, IndentWidth: 2,  ContinuationIndentWidth: 2 }`. FYI

include/intel/cpu_support.h
```diff
@@ -0,0 +1,156 @@
...
+#define INTEL_CPU_SUPPORT_H_
+#include <memory>
+#include <type_traits>
+#include "xbyak/xbyak.h"
+#include "xbyak/xbyak_util.h"
+namespace intel {
+#define log_intel(x) if (IntelKernel<T>::log_enabled()) { std::cout << x << std::endl; }
+#ifndef log_intel
+#define log_intel(x)
+#endif
```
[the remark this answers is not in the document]
pawelpiotrowicz: done , in a new version

include/intel/cpu_support.h
```diff
@@ -0,0 +1,156 @@
...
+template<class T>
+using Uptr = std::unique_ptr<T>;
```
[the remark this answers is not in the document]
pawelpiotrowicz: done

include/intel/cpu_support.h
```diff
@@ -0,0 +1,156 @@
...
+ public:
+     void prolog() {
+        push(r8);
+        push(r9);
+     }
+
+      void epilog() {
+        pop(r9);
+        pop(r8);
+      }
```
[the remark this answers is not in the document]
pawelpiotrowicz: done

include/intel/cpu_support.h
```diff
@@ -0,0 +1,156 @@
...
+
+      static int log_enabled() {
+        static int r = (std::getenv("DGL_CPU_INTEL_KERNEL_LOG")) ? 1 : 0;
+        return r;
+      }
+};
+
+template <class T, int has_specialization = 0>
+class elem_wise_add_update : public Xbyak::CodeGenerator {
```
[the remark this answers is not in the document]
pawelpiotrowicz: We don't have to override any method

include/intel/cpu_support.h
```diff
@@ -0,0 +1,156 @@
...
+        push(r8);
+        push(r9);
+     }
+
+      void epilog() {
+        pop(r9);
+        pop(r8);
+      }
+    explicit elem_wise_add_update(int64_t _size) :
+     size(_size), applicable(false), min_size(IntelKernel<T>::enabled()) {
```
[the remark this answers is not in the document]
pawelpiotrowicz: N/A in next release

include/intel/cpu_support.h
```diff
@@ -0,0 +1,156 @@
...
+    bool is_applicable() const {
+      return applicable;
+    }
+
+    /*[functional] generate a new kernel for this size*/
+    template<class R>
+    typename std::enable_if<has_specialization, R>::type
+    require_new_instance(int64_t _size) {
+      return  (_size != size) && (_size >= min_size);
+    }
```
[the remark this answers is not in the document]
pawelpiotrowicz: bool was used as workaround to enforce SFINAE

src/array/cpu/spmm.h
pawelpiotrowicz: the new release supports all Ops {CopyLhs, CopyRhs , Sum ... } with all types { float , doble }

src/array/cpu/spmm.h
pawelpiotrowicz: N/A  the concept  has been changed

src/array/cpu/spmm.h
pawelpiotrowicz: done

src/array/cpu/spmm.h
pawelpiotrowicz: done

include/intel/cpu_support.h
```diff
@@ -0,0 +1,156 @@
...
+#define log_intel(x)
+#endif
+template<class T>
+using Uptr = std::unique_ptr<T>;
+
+template<class T>
+struct IntelKernel {
+      static int64_t getValue() {
+        int64_t v = 0;
+        const char *label = "DGL_CPU_INTEL_KERNEL_ENABLED";
```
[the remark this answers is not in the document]
pawelpiotrowicz: done

pawelpiotrowicz:
@jermainewang  @VoVAllen  Thank you very much for your review.  I added your remarks to next release – it’s almost finished.

jermainewang:
Updated in #2309

PR #2306, opened 2020-10-27
[kernel] Select GE-SpMM when feature size is large.

yzh119:
## Description
GE-SpMM is a SC2020 paper targets improving the performance of SpMM when dealing with scalar shape edge features.
The performance of GE-SpMM is better than cusparse in some settings. This PR implements GE-SpMM in DGL.

## TODO
* [x] Our profiling result suggests GE-SpMM's performance is not as good as cusparse when feature size is small (<=64) in ogbn-arxiv, DGL need to switch between cusparse's implementation and GE-SpMM for better performance.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/src/array/cuda/ge_spmm.cuh
+++ b/src/array/cuda/ge_spmm.cuh
@@ -33,7 +33,6 @@ __global__ void GESpMMSumKernel(
     const Idx* __restrict__ indices,
     const int64_t num_rows, const int64_t num_cols,
     const int64_t feat_len) {
-  /*
   extern __shared__ char smem[];
   Idx* col = nullptr;
   DType* val = nullptr;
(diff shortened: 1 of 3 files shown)


yzh119:
Some observations:
1. we have to switch the order of dimension x and dimension y in the grid, so that the blocks would be scheduled along the "nodes" dimension, in this way the L2 cache utilization was improved by 2x (compared to scheduling along the feature dimension).
2. The effect of shared memory is negligible (at least on my V100 machine).

yzh119:
The observation 1 indicates there are still room for us to improve the kernel (e.g. reordering the nodes to improve L2 utilization).

When blocks are scheduled along "feature" dimension, no memory access to node features was reused (different block access to different regions of node features). 
When blocks are scheduled along "nodes" dimension, the cache locality highly depends on how many neighbor nodes are overlapped among the blocks being executed simultaneously in SMs. If the graph structure is ordered nicely (nodes with similar IDs have similar neighbor sets) we can expect higher L2 cache locality.

However, the behavior of block scheduler is undefined (i.e. it's not sure if different architectures would use the same strategy).

cc @xuyifangreeneyes as you might be interested.

PR #2323, opened 2020-11-06
[BugFix][Dataloading] Disable AsyncTransferer copying to anywhere but the GPU

nv-dlasalle:
## Description
Disable using AsyncTransferer for copying to anywhere but the GPU, to avoid the failing unit. Will diagnose and create new PR re-enabling functionality once issue is identified.

## Checklist
- [x] Code is well-documented

## Changes

Made AsyncTransferer.copy_async() raise a ValueError if the destination context is not the GPU.

Change the unit test to expect an a ValueError when the destination is not a GPU.

### DIFF
--- a/python/dgl/dataloading/async_transferer.py
+++ b/python/dgl/dataloading/async_transferer.py
@@ -1,6 +1,5 @@
-""" API for transferring data to/from the GPU over second stream.A """
+"""API for transferring data to the GPU over second stream."""
 
-import dgl
 from .. import backend as F
 from .. import ndarray
 from .. import utils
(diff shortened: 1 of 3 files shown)


python/dgl/dataloading/async_transferer.py
```diff
@@ -1,6 +1,6 @@
 """ API for transferring data to/from the GPU over second stream.A """
 
-
+import dgl
```
jermainewang, line 3: remove

nv-dlasalle: Removed in aada9b9

PR #2333, opened 2020-11-10
[Bug] Multiple fixes for CUDA 11 support

BarclayII:
## Description
This PR makes sure to include CUB/Thrust in our submodules rather than the system ones if CUDA version is no greater than 11.0.  It also removes a redundant NonZero implementation.

### DIFF
--- a/cmake/util/FindCUDA.cmake
+++ b/cmake/util/FindCUDA.cmake
@@ -41,20 +41,6 @@ macro(find_cuda use_cuda)
         ${CUDA_TOOLKIT_ROOT_DIR}/lib/x64
         ${CUDA_TOOLKIT_ROOT_DIR}/lib/Win32)
     else(MSVC)
-      find_library(_CUDA_CUDA_LIBRARY cuda
-        PATHS ${CUDA_TOOLKIT_ROOT_DIR}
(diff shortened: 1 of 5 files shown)


jermainewang:
Do you have a separate PR to the master branch?

BarclayII:
This is to the master branch.

cmake/util/FindCUDA.cmake
```diff
@@ -41,20 +41,14 @@ macro(find_cuda use_cuda)
         ${CUDA_TOOLKIT_ROOT_DIR}/lib/x64
         ${CUDA_TOOLKIT_ROOT_DIR}/lib/Win32)
     else(MSVC)
+      # The CI seems to have trouble locating the CUDA library so I'm manually doing that.
       find_library(_CUDA_CUDA_LIBRARY cuda
         PATHS ${CUDA_TOOLKIT_ROOT_DIR}
         PATH_SUFFIXES lib lib64 targets/x86_64-linux/lib targets/x86_64-linux/lib/stubs lib64/stubs
```
BarclayII, line 47: CI fails if I remove the stubs directory.  But the same Docker image passes on my machine @jermainewang

yzh119:
Shall we release dgl 0.5.3 now?

PR #2345, opened 2020-11-13
[Fix] Enable mini-batch r-gcn for CPU

mszarma:
## Description

Fix for enabling r-gcn training for CPU device in entity_classify_mp.py
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/examples/pytorch/rgcn/entity_classify_mp.py
+++ b/examples/pytorch/rgcn/entity_classify_mp.py
@@ -185,7 +185,7 @@ def evaluate(model, embed_layer, eval_loader, node_feats):
 
 @thread_wrapped_func
 def run(proc_id, n_gpus, args, devices, dataset, split, queue=None):
-    dev_id = devices[proc_id]
+    dev_id = devices[proc_id] if devices[proc_id] != 'cpu' else -1
     g, node_feats, num_of_ntype, num_classes, num_rels, target_idx, \
         train_idx, val_idx, test_idx, labels = dataset
     if split is not None:
--- a/examples/pytorch/rgcn/model.py
+++ b/examples/pytorch/rgcn/model.py
@@ -78,7 +78,7 @@ class RelGraphEmbedLayer(nn.Module):
                  sparse_emb=False,
                  embed_name='embed'):
         super(RelGraphEmbedLayer, self).__init__()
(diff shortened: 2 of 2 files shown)


examples/pytorch/rgcn/entity_classify_mp.py
```diff
@@ -185,7 +185,7 @@ def evaluate(model, embed_layer, eval_loader, node_feats):
 
 @thread_wrapped_func
 def run(proc_id, n_gpus, args, devices, dataset, split, queue=None):
-    dev_id = devices[proc_id]
```
classicsong, line 188: For cpu runs, the input is already ['cpu']. Why we need to change here?

mszarma: Hi @classicsong thanks for comment.
Regarding you questions please look at line 261 - there is a problem with that condition and str vs int comparison.

examples/pytorch/rgcn/model.py
```diff
@@ -118,13 +118,14 @@ def forward(self, node_ids, node_tids, type_ids, features):
             embeddings as the input of the next layer
         """
         tsd_ids = node_ids.to(self.node_embeds.weight.device)
-        embeds = th.empty(node_ids.shape[0], self.embed_size, device=self.dev_id)
+        device_id = self.dev_id if self.dev_id >= 0 else 'cpu'
```
classicsong, line 121: We can make self.dev_id = th.device(dev_id if dev_id >= 0 else 'cpu') at __init__ function.
Then there is no need to do following changes.

examples/pytorch/rgcn/entity_classify_mp.py
classicsong: I c.
Good point.
Can we just change the L514 to send in [-1] ?

classicsong (commented):
Some suggestions.
Overall LGTM.

examples/pytorch/rgcn/entity_classify_mp.py
BarclayII: I don't think so.  For instance `RelGraphEmbedLayer` uses `dev_id` rather than `device`.

PR #2356, opened 2020-11-19
[Performance]: Remove indptr correction for CooToCSR

IzabelaMazur:
Remove the loop used to improve indptr in COOToCSR(), instead use CreateView() to enter the correct values at once.

### DIFF
--- a/src/array/cpu/spmat_op_impl_coo.cc
+++ b/src/array/cpu/spmat_op_impl_coo.cc
@@ -341,6 +341,7 @@ CSRMatrix COOToCSR(COOMatrix coo) {
   } else {
     // compute indptr
     IdType* Bp = static_cast<IdType*>(ret_indptr->data);
+    *(Bp++) = 0;
     std::fill(Bp, Bp + N, 0);
     for (int64_t i = 0; i < NNZ; ++i) {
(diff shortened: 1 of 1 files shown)


src/array/cpu/spmat_op_impl_coo.cc
```diff
@@ -339,8 +340,9 @@ CSRMatrix COOToCSR(COOMatrix coo) {
     ret_indices = coo.col;
     ret_data = coo.data;
   } else {
+    static_cast<IdType*>(ret_indptr->data)[0] = 0;
     // compute indptr
-    IdType* Bp = static_cast<IdType*>(ret_indptr->data);
+    IdType* Bp = static_cast<IdType*>(ret_indptr_view->data);
```
jermainewang, line 345: We could further simplify it as:
```c++
} else {
  IdType* Bp = static_cast<IdType*>(ret_indptr->data);
  *(Bp++) = 0;
  ...  // the rest part is the same as your change
}
```
It seems that there is no need to create a view at all.

IzabelaMazur: I followed the suggestion. Thank you.

PR #2391, opened 2020-12-03
[Performance] Improve COO to CSR, and sort columns of CSR only when necessary.

nv-dlasalle:
## Description
Avoid unnecessarily sort CSR columns, reduces the amount of work done when rows are sorted, and avoids synchronizing to double check if a COO is not sorted. It also parallelizes the conversion of an unsorted COO to CSR on the CPU, which can significantly reduce startup time (when the formats are generated for the global graph).

On `examples/pytorch/rgcn/entity_classify_mp.py` with the `ogb-mag` dataset, using the master branch with the PyTorch allocator, sorting on the GPU looks like:
sorting_master

And here it is with this PR:
sorting_opt_pr

With this PR, you can see, the check for if the matrix is sorted is removed, and the number of radix sort iterations is significantly reduced.

## Checklist
- [x] Code is well-documented

## Changes

Changes UnitGraph to return unsorted CSRs for `GetInCSR()` and `GetOutCSR()`

Changes COOToCSR to not check if the matrix is sorted, if it claims to be unsorted. This eliminates a D2H copy, which forces synchronization between the CPU and GPU.

Changes `COOSort()` to respect the `sort_columns` flag, and merges the separate int32_t and int64_t version into a single one which uses CUB and takes advantage of the small number of rows relative to the width of IdType. That is, if we have 100,000 nodes in our block, we will only perform a radix sort of width 17, rather than the full 32 or 64 bits of the IdType.

Changes the CPU version of COOToCSR to generate the CSR in parallel, by doing two passes over the graph, one to count the number of edges per vertex, and the second to insert the edges in the correct positions.

This also expands the unit tests in `tests/cpp/test_spmat_csr.cc` to ensure unsorted CSRs are not problematic.

### DIFF
--- a/include/dgl/aten/array_ops.h
+++ b/include/dgl/aten/array_ops.h
@@ -314,7 +314,16 @@ IdArray NonZero(NDArray array);
  * is always in int64.
  *
  * \param array Input array.
- * \param num_bits The number of bits used by the range of values in the array.
+ * \param num_bits The number of bits used by the range of values in the array,
+ * or 0 to use all bits of the type. This is currently only used when sort
+ * arrays on the GPU. 
+ * \param num_bits The number of bits used in key comparison. The bits are
+ * right aligned. For example, setting `num_bits` to 8 means using bits from
+ * `sizeof(IdType) * 8 - num_bits` (inclusive) to `sizeof(IdType) * 8`
+ * (exclusive). Setting it to a small value could speed up the sorting if the
+ * underlying sorting algorithm is radix sort (e.g., on GPU). Setting it to
+ * value of zero, uses full number of bits of the type (sizeof(IdType)*8).
+ * On CPU, it currently has no effect.
  * \return A pair of arrays: sorted values and sorted index to the original position.
  */
 std::pair<IdArray, IdArray> Sort(IdArray array, int num_bits = 0);
--- a/include/dgl/aten/array_ops.h
+++ b/include/dgl/aten/array_ops.h
@@ -314,9 +314,10 @@ IdArray NonZero(NDArray array);
  * is always in int64.
  *
  * \param array Input array.
+ * \param num_bits The number of bits used by the range of values in the array.
  * \return A pair of arrays: sorted values and sorted index to the original position.
  */
-std::pair<IdArray, IdArray> Sort(IdArray array);
+std::pair<IdArray, IdArray> Sort(IdArray array, int num_bits = 0);
 
 /*!
  * \brief Return a string that prints out some debug information.
--- a/src/array/cpu/array_sort.cc
+++ b/src/array/cpu/array_sort.cc
@@ -161,7 +161,7 @@ namespace aten {
 namespace impl {
 
 template <DLDeviceType XPU, typename IdType>
-std::pair<IdArray, IdArray> Sort(IdArray array) {
+std::pair<IdArray, IdArray> Sort(IdArray array, int num_bits) {
   const int64_t nitem = array->shape[0];
   IdArray val = array.Clone();
   IdArray idx = aten::Range(0, nitem, 64, array->ctx);
@@ -181,8 +181,8 @@ std::pair<IdArray, IdArray> Sort(IdArray array) {
   return std::make_pair(val, idx);
 }
 
-template std::pair<IdArray, IdArray> Sort<kDLCPU, int32_t>(IdArray);
-template std::pair<IdArray, IdArray> Sort<kDLCPU, int64_t>(IdArray);
+template std::pair<IdArray, IdArray> Sort<kDLCPU, int32_t>(IdArray, int num_bits);
+template std::pair<IdArray, IdArray> Sort<kDLCPU, int64_t>(IdArray, int num_bits);
 
 }  // namespace impl
 }  // namespace aten
--- a/src/array/cpu/array_sort.cc
+++ b/src/array/cpu/array_sort.cc
@@ -161,7 +161,7 @@ namespace aten {
 namespace impl {
 
 template <DLDeviceType XPU, typename IdType>
-std::pair<IdArray, IdArray> Sort(IdArray array, int num_bits) {
+std::pair<IdArray, IdArray> Sort(IdArray array, int /* num_bits */) {
   const int64_t nitem = array->shape[0];
   IdArray val = array.Clone();
   IdArray idx = aten::Range(0, nitem, 64, array->ctx);
--- a/src/array/cpu/spmat_op_impl_coo.cc
+++ b/src/array/cpu/spmat_op_impl_coo.cc
@@ -3,6 +3,7 @@
  * \file array/cpu/spmat_op_impl.cc
  * \brief CPU implementation of COO sparse matrix operators
  */
+#include <omp.h>
 #include <vector>
 #include <unordered_set>
 #include <unordered_map>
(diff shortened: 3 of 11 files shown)


include/dgl/aten/array_ops.h
```diff
@@ -314,9 +314,10 @@ IdArray NonZero(NDArray array);
  * is always in int64.
  *
  * \param array Input array.
+ * \param num_bits The number of bits used by the range of values in the array.
```
jermainewang, line 317:
```c++
\param num_bits The number of bits used in key comparison. The bits are right aligned.
                For example, setting `num_bits` to 8 means using bits from `sizeof(IdType) * 8 - num_bits` (inclusive)
                to `sizeof(IdType) * 8` (exclusive). Setting it to a small value could speed up the sorting
                if the underlying sorting algorithm is radix sort (e.g., on GPU). On CPU, it currently has no effect.
```

tests/cpp/test_spmat_coo.cc
```diff
@@ -121,7 +121,7 @@ void _TestCOOToCSR(DLContext ctx) {
   auto tcsr = aten::COOToCSR(coo);
   ASSERT_EQ(coo.num_rows, csr.num_rows);
   ASSERT_EQ(coo.num_cols, csr.num_cols);
-  ASSERT_TRUE(ArrayEQ<IDX>(csr.indptr, tcsr.indptr));
+  ASSERT_TRUE(ArrayEQ<IDX>(csr.indptr, tcsr.indptr)) << "csr.indptr = " << csr.indptr << ", tcsr.indptr = " << tcsr.indptr;
```
jermainewang, line 124: revert?

tests/cpp/test_spmat_csr.cc
```diff
@@ -30,14 +30,38 @@ aten::CSRMatrix CSR2(DLContext ctx = CTX) {
   //  [0, 0, 1, 1, 0],
   //  [0, 0, 0, 0, 0]]
   // data: [0, 2, 5, 3, 1, 4]
-  return aten::CSRMatrix(
+   return aten::CSRMatrix(
```
jermainewang, line 33: revert?

src/array/cpu/spmat_op_impl_coo.cc
```diff
@@ -296,14 +297,14 @@ template COOMatrix COOTranspose<kDLCPU, int64_t>(COOMatrix coo);
 
 ///////////////////////////// COOToCSR /////////////////////////////
 
-// complexity: time O(NNZ), space O(1)
+// complexity: time O(NNZ), space O(NNZ)
```
jermainewang, line 300: The space complexity is `O(num_threads * N)` right? Could you run the recently merged [benchmark code](https://github.com/dmlc/dgl/blob/master/benchmarks/benchmarks/api/bench_format_conversion.py) and compare the memory consumption and speed up with the previous implementation (especially on those large graphs)? You could copy-paste out the benchmark code and run it locally without the asv pipeline.

jermainewang (changes requested):
The PR is mostly in good shape. Some comments.

src/array/cuda/coo_sort.cu
```diff
@@ -16,107 +17,97 @@ namespace impl {
 
 ///////////////////////////// COOSort_ /////////////////////////////
 
-template <DLDeviceType XPU, typename IdType>
-void COOSort_(COOMatrix* coo, bool sort_column) {
-  LOG(FATAL) << "Unreachable codes";
+template <typename IdType>
+__global__ void _COOMergeEdgesKernel(
```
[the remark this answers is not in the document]
nv-dlasalle: In 5e8cb08 I added full doxygen comments to both, and renamed them to `_COOEncodeEdgesKernel` and `_COODecodeEdgesKernel`.

src/array/cuda/array_sort.cu
```diff
@@ -14,36 +14,40 @@ namespace aten {
...
-std::pair<IdArray, IdArray> Sort(IdArray array) {
+std::pair<IdArray, IdArray> Sort(IdArray array, int num_bits) {
   const auto& ctx = array->ctx;
   auto device = runtime::DeviceAPI::Get(ctx);
   const int64_t nitems = array->shape[0];
-  IdArray orig_idx = Range(0, nitems, 64, ctx);
+  IdArray orig_idx = Range(0, nitems, sizeof(IdType)*8, ctx);
   IdArray sorted_array = NewIdArray(nitems, ctx, array->dtype.bits);
-  IdArray sorted_idx = NewIdArray(nitems, ctx, 64);
+  IdArray sorted_idx = NewIdArray(nitems, ctx, sizeof(IdType)*8);
```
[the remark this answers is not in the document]
nv-dlasalle: Reverted this changes in 8364d72

tests/cpp/test_spmat_coo.cc
nv-dlasalle: Reverted in 9cfabb9

tests/cpp/test_spmat_csr.cc
nv-dlasalle: Reverted in 9cfabb9

include/dgl/aten/array_ops.h
nv-dlasalle: Added this documentation in e9edbb4

src/array/cpu/array_sort.cc
```diff
@@ -161,7 +161,7 @@ namespace aten {
 namespace impl {
 
 template <DLDeviceType XPU, typename IdType>
-std::pair<IdArray, IdArray> Sort(IdArray array) {
+std::pair<IdArray, IdArray> Sort(IdArray array, int num_bits) {
```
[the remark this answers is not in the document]
nv-dlasalle: Added documentation in e9edbb4  and commented out the variable name.

src/array/cpu/spmat_op_impl_coo.cc
nv-dlasalle: I updated the complexity in d95a08d to:
```
complexity: time O(NNZ), space O(1) if the coo is row sorted,
time O(NNZ/p + N), space O(NNZ + N*p) otherwise, where p is the number of
threads.
```

I ran the benchmark by just reverting the 

For reddit, the previous implementation averaged 6.63s, and used 6.20 GB of memory.
With the new implementation using OMP_NUM_THREADS=1, it averaged 6.5s and used 6.20 GB of memory.
With OMP_NUM_THREADS=12, it averaged 1.83s and took 6.22 GB of memory.

For ogb-products the previous implementation averaged 11.9s and used 7.15GB of memory.
With the new implementation  using OMP_NUM_THREADS=1, it averaged 12.8s and used 7.15 GB of memory.
With OMP_NUM_THREADS=12, it averaged 2.07s and used 7.36 GB of memory.

For friendster, the previous implementation averaged 13.1s and used 58.6 GB of memory.
With the new implementation using OMP_NUM_THREADS=1, it averaged 22s and used 87.8 GB of memory.
With OMP_NUM_THREADS=12, it averaged 7.32s and used 98 GB of memory.
Note, friendster is sorted, and thus the original implementation takes advantage of re-using arrays to reduce memory consumption.

To get the memory usage I ran the commands with `/usr/bin/time -v`, and the above is the `Maximum resident set size`.

jermainewang: Cool! Thanks for the benchmarks. It clears my concern.

jermainewang (approved):
LGTM

Issue #2408, opened 2020-12-08
[feature request] Wrap OpenMP's API for better resource control


yzh119:
## 🚀 Feature
DGL should provide a wrapped `parallel_for` API; instead of calling `#pragma omp parallel for` directly, all requests to parallelize a loop must go through DGL's `parallel_for` API to determine whether to use multi-threading and limit the number threads it creates.

## Motivation

We found that graph creation and transformation (coo2csr, etc.) creates too many threads. The main reason is we simply uses OpenMP's `parallel for` intrinsic and do not consider the resource we own or the loop requires. For example, we use multi-threading for IdArray's element-wise operations:

https://github.com/dmlc/dgl/blob/501b2b75a59aab3dd9a3f56240b83fb27962b184/src/array/cpu/array_op_impl.cc#L52

which significantly slows down the dataloading phase, as observed in @vijaydwivedi75 's experience on running several benchmarking-gnn script on ZINC dataset in parallel on a machine with limited CPU cores.

## Reference

PyTorch has wrapped a `parallel_for` API: https://github.com/pytorch/pytorch/blob/93973ee6993c97c8b9d60c4f720423bc625073ea/aten/src/ATen/ParallelOpenMP.h#L15

which did the following:
1. checks whether the loop requires multi-threading (`end - start >= grain_size`).
2. checks whether the current task is enclosed by an active parallel region (already in parallel).
3. controls number of threads used in current loop.
4. split loop into segments for better cache locality.

We can use similar approach to control where/how to use OpenMP's parallel for, and replace all `#pragma omp parallel` with a `parallel_for` call controlled by DGL.

## Additional Context

Also we need to set up regression test to avoid performance degradation in some examples.

cc @jermainewang @BarclayII @VoVAllen @zheng-da

PR #2412, opened 2020-12-10
[Performance][Hotfix] Disable openmp in arithmetic operation

BarclayII:
## Description
Currently we parallelize DGL's arithmetic operations using OpenMP.  However the workload size of those operations are usually very light:
* For Benchmarking GNNs, the workload size is around 10k for SBM-PATTERN 100k experiments, and less than 100 for ZINC 100k experiments.
* For OGB products, the workload size is usually 1.

Using OpenMP to parallelize these operations will incur lots of overhead in thread spawning and scheduling, which is the main reason why Benchmarking GNNs experiments slowed down in 0.5+, especially in parallel experiments.

Here is the comparison of training time (seconds per epoch) on a g3.16x AWS instance between removing and keeping the OpenMP routines in Benchmarking GNNs experiments and OGB experiments:

```
		0.4.2	0.5.3	0.5.3 No OMP
ZINC
GCN		2.1130	2.4221	2.1997
GAT		3.3316	3.2240	2.9484
SAGE		2.3838	2.4674	2.2154
GIN		3.0113	3.2034	3.0296
GGCN		4.1356	6.2027	4.2853
MoNet		2.9530	3.0451	2.8517

SBM
GCN		29.5327	18.6368	16.6338
GAT		84.6075	33.3924	23.4785
SAGE		38.2151	17.8404	14.2021
GIN		32.2307	25.8041	19.7009
GGCN		207.929	177.431	175.144
MoNet		110.498	32.5692	31.5179

OGB-Products
SAGE (4 workers)	37.6392	40.5491
GAT (4 workers)		216.969	198.433
SAGE (0 workers)	56.5372	59.1538
GAT (0 workers)		131.085	117.392

OGB-Arxiv
GCN			0.8707	0.9196
```

Later on, we should replace all OpenMP parallel loops with PyTorch's `parallel_for` construct as mentioned in #2408 .

### DIFF
--- a/src/array/cpu/array_op_impl.cc
+++ b/src/array/cpu/array_op_impl.cc
@@ -48,8 +48,9 @@ IdArray BinaryElewise(IdArray lhs, IdArray rhs) {
   const IdType* lhs_data = static_cast<IdType*>(lhs->data);
   const IdType* rhs_data = static_cast<IdType*>(rhs->data);
   IdType* ret_data = static_cast<IdType*>(ret->data);
-  // TODO(minjie): we should split the loop into segments for better cache locality.
-#pragma omp parallel for
+  // TODO(BarclayII): this usually incurs lots of overhead in thread spawning, scheduling,
+  // etc., especially since the workload is very light.
+//#pragma omp parallel for
   for (int64_t i = 0; i < lhs->shape[0]; ++i) {
     ret_data[i] = Op::Call(lhs_data[i], rhs_data[i]);
   }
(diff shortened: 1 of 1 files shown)


src/array/cpu/array_op_impl.cc
```diff
@@ -84,8 +85,9 @@ IdArray BinaryElewise(IdArray lhs, IdType rhs) {
   IdArray ret = NewIdArray(lhs->shape[0], lhs->ctx, lhs->dtype.bits);
   const IdType* lhs_data = static_cast<IdType*>(lhs->data);
   IdType* ret_data = static_cast<IdType*>(ret->data);
-  // TODO(minjie): we should split the loop into segments for better cache locality.
-#pragma omp parallel for
+  // TODO(BarclayII): this usually incurs lots of overhead in thread spawning, scheduling,
+  // etc., especially since the workload is very light.
```
jermainewang, line 89: `TODO(BarclayII): Comment out .... due to .... . Need to replace it with ... in the future.`

jermainewang (approved):
LGTM

PR #2415, opened 2020-12-10
[Test] Basic regression test setup.

jermainewang:
## Description

Include a basic regression test suite setup. Compared with the previous setup, this one:
* Easy to run locally. Just by `asv run -n -e --python=same` under the `benchmarks` folder.
* Use ASV's native environment mechanism instead of using subprocess to workaround it.
* Re-organize the benchmark folder and add a readme for detailed instruction on how-to-add-new-ones.
* Make benchmark device an environment variable `DGL_BENCH_DEVICE` instead of the benchmark parameter.

Please see the `README.md` for instructions.

This PR does NOT cover
* Complete benchmark suites such as kernel benchmarks.
* Resource scheduler script such as assigning jobs to specific hardware.

## Checklist
- [x] Code is well-documented


VoVAllen (approved):
LGTM. One question is about specifying the machine name for the publishing. Since we will use on-demand launching instances in the future, the machine name will change under default setting(using hostname) for every run. Probably we will use aws instance type as name in the future.

PR #2418, opened 2020-12-12
[RPC] Make the port reusable after the socket is closed.

zheng-da:
## Description
Currently, when a socket that binds to a port is closed, the socket cannot be reused for a few minutes. This is because the closed socket is in a wait status in the Linux kernel. More details can be found here: https://hea-www.harvard.edu/~fine/Tech/addrinuse.html
The fix is to make the socket reusable after the socket is closed.

## Checklist
- [ ] Code is well-documented


Lyken17:
Tested on >20 runs and the port unreleased problem is fixed. Please merge.

PR #2419, opened 2020-12-13
delete shared memory when receiving signals.

zheng-da:
## Description
Currently, the shared memory is deleted when the process exits normally, but cannot be deleted if the process is killed by signals. This PR is to clean up shared memory when the process is killed by SIGINT (interrupt from Ctrl+C) and SIGTERM (kill).

## Checklist
- [ ] Code is well-documented


Lyken17:
I have tested on ogb-product, everything works well. I will merge it after testing on ogb-paper100M. 


TODO: fix lint errors 
```
=====140/142 cpp-source files passed check=====

src/runtime/resource_manager.cc: 1 Errors of 1 Categories map={'build': 1}

src/runtime/shared_mem.cc: 1 Errors of 1 Categories map={'runtime': 1}
```

zheng-da:
i'll fix lint.

VoVAllen:
Will the `shm_unlink` be called twice? Also why the unit test failed?

zheng-da:
`shm_unlink` should only be called once. If the shared-memory file is deleted explicitly, the resource is removed from the hashmap, its destroy function won't be called. otherwise, the destroy function is called by the signal handler.

VoVAllen (approved):
LGTM. The ultimate solution might be using a separate process to monitor the lifecycle of shared memory (such as what PyTorch and Plasma Store did). However, I believe the solution is enough for our current scenario

PR #2430, opened 2020-12-17
[feature] APIs for manually set batch_num_nodes and batch_num_edges

### LINKED ISSUE #2422: Add subgraph support for batched graphs
## 🚀 Feature
Add subgraph support for batched graphs.

## Motivation

Taking the subgraph of batched DGLGraphs (e.g. `node_subgraph`) will remove the `batch_num_nodes` and `batch_num_edges` attributes, leading to incorrect behavior when need batch information in the following operations.

## Alternatives

One can reset these two attributes manully after the subgraph operation, however the side-effect of this alternative method is unknown.

## Pitch

```python
# assert the graph is batched
graph = dgl.node_subgraph(graph, sub_idx)

# then we can use batch_num_nodes here
foo(graph, graph.batch_num_nodes())
```

yzh119:
## Description
<del>To fix #2422 , it's still work in progress.</del>
Expose `batch_num_nodes` and `batch_num_edges` to user to address #2422.

## TODO List
- [x] <del>Support the feature for remaining subgraph functions.</del>
- [ ] <del>Raise an warning when the given node/edge index is not sorted because in this case the `batch_num_nodes` and `batch_num_edges` no longer makes sense.</del>
- [ ] <del>Implement a more efficient `_get_subgraph_batch_info` function.</del>

## Checklist
- [x] Code is well-documented

cc @lygztq WDYT?


lygztq:
Great!

python/dgl/heterograph.py
```diff
@@ -1257,7 +1257,16 @@ def batch_num_nodes(self, ntype=None):
...
+        """Manually set the number of nodes for each graph in the batch with the specified node
+        type.
+
+        Parameters
+        ----------
+        val : Tensor or Mapping[str, Tensor]
+            The dictionary storing number of nodes for each graph in the batch for all node types.
+            If the graph has only one node type, ``val`` can also be a single array indicating the
+            number of nodes per graph in the batch.
+        """
```
mufeili, line 1269: Add examples

python/dgl/heterograph.py
```diff
@@ -1326,7 +1335,16 @@ def batch_num_edges(self, etype=None):
...
+        """Manually set the number of edges for each graph in the batch with the specified edge
+        type.
+
+        Parameters
+        ----------
+        val : Tensor or Mapping[str, Tensor]
+            The dictionary storing number of edges for each graph in the batch for all edge types.
+            If the graph has only one edge type, ``val`` can also be a single array indicating the
+            number of edges per graph in the batch.
+        """
```
mufeili, line 1347: Ditto

mufeili:
Do we want to automatically handle batch information in `node_subgraph` and `edge_subgraph`? If so, do we also want to do that for graph transformations? @hetong007 @jermainewang

yzh119:
@mufeili I've looked into https://github.com/dmlc/dgl/blob/5f44a4efce7a064aeba9b50e759a0f27cb2ccf75/python/dgl/transform.py
and found there is no usage of `subgraph` API.

The feature would not influence the speed of `subgraph` in most cases because we rarely apply `subgraph` on batched graphs (however, this is the common case for graph pooling layers).

mufeili:
> @mufeili I've looked into https://github.com/dmlc/dgl/blob/5f44a4efce7a064aeba9b50e759a0f27cb2ccf75/python/dgl/transform.py
> and found there is no usage of `subgraph` API.
> 
> The feature would not influence the speed of `subgraph` in most cases because we rarely apply `subgraph` on batched graphs (however, this is the common case for graph pooling layers).

Ok

yzh119:
After discussion we decide not to create batch information in `node_subgraph`/`edge_subgraph`, but to expose `set_batch_num_nodes` and `set_batch_num_edges` to user and let user process the batch information manually.

python/dgl/heterograph.py
```diff
@@ -1257,7 +1257,78 @@ def batch_num_nodes(self, ntype=None):
...
+        The following example uses PyTorch backend.
+
+        >>> import dgl
+        >>> import torch
+
+        Create a homogeneous graph.
+        >>> g = dgl.graph(([0, 1, 2, 3, 4, 5], [1, 2, 0, 4, 5, 3]))
+
+        Manually set batch information
+        >>> g.set_batch_num_nodes(torch.tensor([3, 3])
```
hetong007, line 1287: If there are edges linking nodes from two specified subgraphs, what will the `unbatch` do?

tests/compute/test_batched_graph.py
```diff
@@ -206,13 +206,86 @@ def test_batch_no_edge(idtype):
...
+    subg2 = dgl.edge_subgraph(g2, list(range(50, 100)), preserve_nodes=True)
+    assert subg_e1.num_nodes() == subg1.num_nodes()
+    assert subg_e2.num_nodes() == subg2.num_nodes()
+
+
 if __name__ == '__main__':
-    test_batch_unbatch()
-    test_batch_unbatch1()
-    test_batch_unbatch_frame()
+    #test_batch_unbatch()
```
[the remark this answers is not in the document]
yzh119: Only of debugging on my machine, pytest will collect all functions starts with `test_` and it doesn't matter whether we put them under `if __name__ == "__main__"` or not.

tests/compute/test_batched_graph.py
```diff
@@ -206,13 +206,86 @@ def test_batch_no_edge(idtype):
     g3.add_nodes(1)  # no edges
     g = dgl.batch([g1, g3, g2]) # should not throw an error
 
+def _get_subgraph_batch_info(keys, induced_indices_arr, batch_num_objs):
```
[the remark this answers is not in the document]
yzh119: currently no.

python/dgl/heterograph.py
yzh119: We do not handle this case.

hetong007: Then what is the behavior, will `unbatch` ignore those edges?

A follow up question: if `set_batch_num_nodes` and `set_batch_num_edges` conflict with each other, how does it affect `unbatch` and other related APIs' behavior?

I suggest we either add an API to check that, or add related details in the doc of these two new APIs to warn users.

yzh119: User must guarantee that there will be no invalid (cross-graph) links, but checking validity is too costly and I don't think it necessary to add an API to do so.
I prefer mentioning the behaviour in the docstring.

tests/compute/test_batched_graph.py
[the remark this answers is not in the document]
hetong007: I assume in the end you'll uncomment these lines?

python/dgl/heterograph.py
hetong007: Then would you mind to add a Note/Warning style block in the docstring? 

This note:

- defines *invalid input*, i.e. cross-subgraph links;
- informs users of the potential risks (if called by other APIs like `unbatch`).

yzh119: Done.

python/dgl/heterograph.py
```diff
@@ -1409,8 +1411,10 @@ def set_batch_num_edges(self, val):
 
         Notes
         -----
-        This API is always used together with ``set_batch_num_nodes`` to specify batching
-        information of a graph.
+        This API is always used together with ``set_batch_num_edges`` to specify batching
```
hetong007, line 1414: Seems `set_batch_num_edges` should be `set_batch_num_nodes`

PR #2442, opened 2020-12-22
[Doc][Org] move edge_softmax to functional

BarclayII:
## Description
This PR moves `edge_softmax` to `dgl.nn.functional` since `dgl.ops` is much more low-level than `edge_softmax`.

@jermainewang @yzh119


yzh119:
I wonder why `dgl.ops` is much low-level?
I prefer regarding `dgl.ops` as a framework-agnostic operator module defined on DGLGraphs like [jax.ops](https://jax.readthedocs.io/en/latest/jax.ops.html).

The interface of `edge_softmax` is not different from other functions inside this module.

BarclayII:
The question arises when I am writing a tutorial recommending what functions or classes we recommend people to use when customizing their GNN module.  When it comes to `edge_softmax` (which is common), if we tell them to use `dgl.ops.edge_softmax`, people will naturally look into the other functions in `dgl.ops` (which include `u_add_v_max` etc.).  Then people will wonder whether they should use `update_all` or `dgl.ops.u_add_v_max` to do stuff.  I would like to avoid this assumption.

Besides, the namespace `dgl.ops` did not have any usage in the examples or tutorials.

yzh119:
I suggest keeping the function definitions in both locations such as `torch.tann` and `torch.nn.funtional.tanh` so that previous code would not be broken.

jermainewang:
Re Zihao. Consider backward compatibility. We should still keep `dgl.ops.edge_softmax`.

Issue #1697, opened 2020-06-24
Is it a trick to speed up the training to add the features of all nodes to the GPU before training?


maqy1995:
## ❓ Questions and Help

When running the GraphSAGE example(the code I use is 'dgl/examples/pytorch/graphsage/train_sampling.py '), I found that passing the features of all nodes into the GPU before starting training can improve the training speed, like this:

```
    # passing the features of all nodes into the GPU before starting training
    g.ndata['features'] = g.ndata['features'].to(device)
    # line 180
    # Training loop
    avg = 0
    iter_tput = []
    for epoch in range(args.num_epochs):
        tic = time.time()

        # Loop over the dataloader to sample the computation dependency graph as a list of
        # blocks.
        for step, blocks in enumerate(dataloader):
            tic_step = time.time()
        .......
        .......
```
    
  This can reduce the epoch training time from 20s to about 11s, at the same time, the increased gpu memory uses about 1G. 
  This seems to be very cost-effective. Are there other hidden dangers of this method?

mufeili:
To move all features to GPU at the beginning avoids repeating it for each mini-batch, which can be time-consuming. In general this should be preferred if your GPU can hold all node features simultaneously.

maqy1995:
Thanks for your reply, I will close this issue.

cfangplus:
I think this could be created as a feature, for instance, we could provide an option to user and it's up to the user to decide whether to move features to GPU before training or not. @mufeili

cfangplus:
I also have a question, we run the code 'dgl/examples/pytorch/graphsage/train_sampling.py' and after only two epoches, the loss seems to have convergented and accuracy reached around 95%, I am really confused, could you help to explain that? @mufeili

mufeili:
> I think this could be created as a feature, for instance, we could provide an option to user and it's up to the user to decide whether to move features to GPU before training or not. @mufeili

I think it depends on if GPU memory can be a bottleneck here. For the particular example we are talking about, can you check the amount of GPU memory occupied? If that's not very large, then I don't think it will be an issue for most users.

mufeili:
> I also have a question, we run the code 'dgl/examples/pytorch/graphsage/train_sampling.py' and after only two epoches, the loss seems to have convergented and accuracy reached around 95%, I am really confused, could you help to explain that? @mufeili

Reddit is an easy dataset to tackle. With sampling-based training, we have a lot more iterations with two epochs than the case of full graph training.

[...]

PR #2453, opened 2020-12-24
[Example] Move Data to GPU before Minibatch Training

mufeili:
## Description
Move data to GPU before minibatch training to save time spent on data copy from CPU to GPU in each iteration.

For a reference, we record the training time(s) per epoch before and after the changes. For single-gpu training, we use g4dn.4xlarge. For multi-gpu training, we use g4dn.12xlarge.

|    | before | after |
| - | ------  | ----  |
| GraphSAGE on Reddit                 | 18  | 4   |
| GraphSAGE on Reddit (4 gpus) | 12  | 5   |
| GraphSAGE on ogbn-products  | 28 | 16 |
| ClusterGAT on ogbn-products  | 24  | 22 |
| GAT on ogbn-products               | 112 | 99 |
| RGCN on AM                                | 2    | 1.8 |

## Checklist
- [ ] Code is well-documented


classicsong:
For examples under ogb, I agree with you that we can put everything in GPU.
For other examples, my opinion is to add an option (which is the default config) as putting everything in GPU as the example maybe modified to train other datasets which may not be able to fit in GPU memory.

examples/pytorch/graphsage/train_sampling_multi_gpu.py
```diff
@@ -137,7 +131,20 @@ def run(proc_id, n_gpus, args, devices, data):
     th.cuda.set_device(dev_id)
 
     # Unpack data
-    in_feats, n_classes, train_g, val_g, test_g = data
+    n_classes, train_g, val_g, test_g = data
+
+    if args.inductive:
+        train_nfeat = train_g.ndata.pop('features').to(dev_id)
```
classicsong, line 137: Will it cause any problem in multi-GPU environment?

mufeili:
> For examples under ogb, I agree with you that we can put everything in GPU.
> For other examples, my opinion is to add an option (which is the default config) as putting everything in GPU as the example maybe modified to train other datasets which may not be able to fit in GPU memory.

@jermainewang What do you think?

examples/pytorch/graphsage/train_sampling_multi_gpu.py
mufeili: I tried running the script and it worked fine.

jermainewang:
I agree. We should expose that option to users. For example, `--pin-feat-on-cpu`; default is False.

maqy1995:
No offense, adding Feature to GPU before training seems to have been discussed in the previous issue? 
Maybe a link should be given to https://github.com/dmlc/dgl/issues/1697 in this PR?

mufeili:
> No offense, adding Feature to GPU before training seems to have been discussed in the previous issue?
> Maybe a link should be given to #1697 in this PR?

Sure. Thanks for pointing out that.

Issue #2436, opened 2020-12-20
[RFC] Support heterogeneous graphs in distributed training


zheng-da:
# Motivation
Starting from version 0.5, DGL supports distributed training. The detailed design of distributed training support in DGL can be found [here](https://arxiv.org/abs/2010.05337). In short, DGL splits a large graph into multiple partitions with METIS and each machine is responsible for a partition. DGL partitions node/edge attributes of the graph according to the METIS partition as well. DGL provides a distributed KVStore to serve node/edge attributes in the distributed setting, which provides a pull API to read nodes/edges attributes and a push API to update node embeddings. The keys for accessing the attributes are node Ids and edge Ids. DGL provides `DistGraph` as a unified interface to access partitioned graph data in the cluster. The main functions provided by `DistGraph` are sampling neighbors of seed nodes (`sample_neighbors`) and accessing node/edge data (`ndata` and `edata`).

Currently, DGL only supports homogeneous graphs in distributed training. However, many graphs in the industry are heterogeneous, in which nodes/edges of different types have different feature dimensions and some nodes/edges may not have features. Unfortunately, we cannot simply extend the current heterogeneous graph design to the distributed setting for the sake of performance. The current heterogeneous graph design uses a separate CSR for each relation and invokes the graph kernel on each relation separately. This doesn't cause much overhead in full-batch/full-graph training but results in inefficient mini-batch computation because a mini-batch is usually small and splitting the small graph structure further into smaller pieces and invoking the graph kernels on them separately usually result in suboptimal performance (any Python overhead and parallelization overhead become more noticeable). In the context of distributed mini-batch training, the current graph structure design will also result in many small sampling requests to the cluster of machines.

# Proposal
The main proposal is to define a graph structure format that works well for distributed heterogeneous graphs to address the problems above. We use a homogeneous graph structure to support distributed sampling and revise the graph partition algorithm to convert any heterogeneous graphs into this proposed format. We revise the distributed version of `sample_neighbors` to run on the proposed graph structure and allows outputting the sampled results in both a homogeneous graph format and a heterogeneous graph format. We extend the KVStore to support the storage of node/edge attributes of different types.

## Graph structure
[...]

PR #2457, opened 2020-12-26
[Distributed] Heterogeneous graph support

zheng-da:
## Description
This PR implements heterogeneous graph support. The design is described in https://github.com/dmlc/dgl/issues/2436

## Checklist
- [ ] Code is well-documented


python/dgl/distributed/graph_partition_book.py
```diff
@@ -73,11 +75,27 @@ def get_shared_mem_partition_book(graph_name, graph_part):
...
-    is_range_part, part_id, num_parts, node_map, edge_map = _get_shared_mem_metadata(graph_name)
+    is_range_part, part_id, num_parts, node_map_data, edge_map_data = \
+            _get_shared_mem_metadata(graph_name)
     if is_range_part == 1:
-        return RangePartitionBook(part_id, num_parts, node_map, edge_map)
+        # node Id ranges and edge Id ranges are stored in the order of node type Ids
+        # and edge type Ids.
+        node_map = {}
+        ntypes = {}
+        node_map_data = pickle.loads(bytes(F.asnumpy(node_map_data).tolist()))
```
jermainewang, line 85: Don't understand this line. Why convert it to bytes and load it again?

python/dgl/distributed/graph_partition_book.py
```diff
@@ -408,101 +494,238 @@ class RangePartitionBook(GraphPartitionBook):
         partition id of current partition book
     num_parts : int
         number of total partitions
-    node_map : tensor
-        map global node id to partition id
-    edge_map : tensor
-        map global edge id to partition id
+    node_map : a dict of tensors
```
jermainewang, line 497: dict[str, Tensor]

python/dgl/distributed/graph_partition_book.py
```diff
@@ -408,101 +494,238 @@ class RangePartitionBook(GraphPartitionBook):
...
+        map global node id to partition id for each node type.
+        The key of the dict is the node type and the value is a 2-column matrix.
+        Each row represents the Id range of a partition, the first element is the starting Id
+        and the second element is the ending Id.
+    edge_map : a dict of tensors
```
jermainewang, line 502: dict[str, Tensor]

python/dgl/distributed/graph_partition_book.py
```diff
@@ -408,101 +494,238 @@ class RangePartitionBook(GraphPartitionBook):
...
+        map global edge id to partition id for each edge type.
+        The key of the dict is the edge type and the value is a 2-column matrix.
+        Each row represents the Id range of a partition, the first element is the starting Id
+        and the second element is the ending Id.
+    ntypes : a dict
```
jermainewang, line 507: dict[str, int]

python/dgl/distributed/graph_partition_book.py
```diff
@@ -408,101 +494,238 @@ class RangePartitionBook(GraphPartitionBook):
...
+        map ntype strings to ntype Ids.
+    etypes : a dict
```
jermainewang, line 509: dict[str, int]

python/dgl/distributed/graph_partition_book.py
```diff
@@ -408,101 +494,238 @@ class RangePartitionBook(GraphPartitionBook):
...
-        if not isinstance(edge_map, np.ndarray):
-            edge_map = F.asnumpy(edge_map)
-        self._node_map = node_map
-        self._edge_map = edge_map
+        self._ntypes = [None] * len(ntypes)
+        self._etypes = [None] * len(etypes)
+        for ntype in ntypes:
+            ntype_id = ntypes[ntype]
+            self._ntypes[ntype_id] = ntype
+        assert all([ntype is not None for ntype in self._ntypes]), \
```
jermainewang, line 522: You could just use `all(self.ntypes)`. Boolean and between None and objects become None and None evaluates to be False.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -408,101 +494,238 @@ class RangePartitionBook(GraphPartitionBook):
...
+                "The node types have invalid Ids."
+        for etype in etypes:
+            etype_id = etypes[etype]
+            self._etypes[etype_id] = etype
+        assert all([etype is not None for etype in self._etypes]), \
```
jermainewang, line 527: same here

jermainewang (changes requested):

python/dgl/distributed/dist_graph.py
```diff
@@ -61,26 +64,21 @@ def _copy_graph_to_shared_mem(g, graph_name):
...
+
+    new_g.edata['inner_edge'] = _to_shared_mem(g.edata['inner_edge'],
+                                               _get_edata_path(graph_name, 'inner_edge'))
     new_g.edata[EID] = _to_shared_mem(g.edata[EID], _get_edata_path(graph_name, EID))
     return new_g
 
-FIELD_DICT = {'inner_node': F.int64,
-              'inner_edge': F.int64,
+FIELD_DICT = {'inner_node': F.int32,
+              'inner_edge': F.int32,
```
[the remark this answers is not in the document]
zheng-da: these are not Ids. They are flags indicating whether nodes/edges are inside a partition.

examples/pytorch/rgcn/experimental/entity_classify_dist.py
```diff
@@ -181,22 +201,41 @@ def forward(self, node_ids, node_tids, features):
...
+    def __init__(self, g, feat_name, num_feats, dev_id):
+        self.g = g
+        self.feat_name = feat_name
+        self.num_feats = num_feats
+        self.dev_id = dev_id
+        self.etype_id_map = {i:etype for i, etype in enumerate(g.etypes)}
+
+    def __call__(self, edge_ids, etype_ids):
+        data = th.empty(edge_ids.shape[0], self.num_feats, device=self.dev_id)
+        for etype_id in th.unique(etype_ids):
```
BarclayII, line 224: Same; do not iterate over tensors directly.

BarclayII (changes requested):
Example code reviews

python/dgl/distributed/dist_graph.py
```diff
@@ -119,29 +117,57 @@ def _get_graph_from_shared_mem(graph_name):
...
+
+    g.edata['inner_edge'] = _get_shared_mem_edata(g, graph_name, 'inner_edge')
     g.edata[EID] = _get_shared_mem_edata(g, graph_name, EID)
     return g
 
+NodeSpace = namedtuple('NodeSpace', ['data'])
+EdgeSpace = namedtuple('EdgeSpace', ['data'])
+
+class HeteroNodeView(object):
+    """A NodeView class to act as G.nodes for a DGLHeteroGraph."""
```
BarclayII, line 132:
```
A NodeView class to act as G.nodes for a DistDGLGraph.
```

python/dgl/distributed/dist_graph.py
```diff
@@ -119,29 +117,57 @@ def _get_graph_from_shared_mem(graph_name):
...
+
+    def __init__(self, graph):
+        self._graph = graph
+
+    def __getitem__(self, key):
+        assert isinstance(key, str)
+        return NodeSpace(data=NodeDataView(self._graph, key))
+
+class HeteroEdgeView(object):
+    """A NodeView class to act as G.nodes for a DGLHeteroGraph."""
```
BarclayII, line 143: Ditto.  Also change node to edge.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -658,9 +910,94 @@ def get_size(self):
...
+    def __init__(self, partition_book, etype='_E'):
+        super(EdgePartitionPolicy, self).__init__(EDGE_PART_POLICY + ':' + etype, partition_book)
+
+class HeteroDataName(object):
+    ''' The data name in a heterogeneous graph.
+
+    A unique data name has three components:
+    * indicate it's node data or edge data.
+    * indicate the node/edge type.
+    * the name of the data.
```
BarclayII, line 938: What does `data` precisely mean?  I assume it's features?

Also the term `data name` and the class name `HeteroDataName` is extremely vague, and may bring confusion with code like `HeteroDataName.get_name()` as you will be getting a name from a name.  Consider renaming it.

I would rename the class to something like `FeatureIdentifier` and put the docstring to something like:
```
A unique identifier for features on heterogeneous graphs, indicating

* Whether the feature is a node feature or edge feature.
* The name of the node/edge type the feature is on.
* The feature's name.
```

python/dgl/distributed/dist_graph.py
```diff
@@ -697,26 +836,43 @@ def barrier(self):
         '''
         self._client.barrier()
 
-    def _get_all_ndata_names(self):
+    def _split_type_name(self, data_names, types):
```
BarclayII, line 839: Where is this function used?

python/dgl/distributed/graph_partition_book.py
```diff
@@ -73,11 +75,27 @@ def get_shared_mem_partition_book(graph_name, graph_part):
...
+        node_map = {}
+        ntypes = {}
+        node_map_data = pickle.loads(bytes(F.asnumpy(node_map_data).tolist()))
+        for i, (ntype, nid_range) in enumerate(node_map_data):
+            ntypes[ntype] = i
+            node_map[ntype] = nid_range
+
+        edge_map = {}
+        etypes = {}
+        edge_map_data = pickle.loads(bytes(F.asnumpy(edge_map_data).tolist()))
```
BarclayII, line 92: Same question as Minjie's.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -149,63 +167,71 @@ def metadata(self):
...
-    def eid2partid(self, eids):
+    def eid2partid(self, eids, etype):
         """From global edge IDs to partition IDs
 
         Parameters
         ----------
         eids : tensor
             global edge IDs
+        etype : str
+            The edge type
```
BarclayII, line 194: Just to make sure: what to do with canonical edge types (i.e. triplets)?  It's OK not handling them though.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -247,6 +277,16 @@ def partid(self):
...
 
+    @property
+    def ntypes(self):
+        """Get the list of node types
+        """
+
+    @property
+    def etypes(self):
+        """Get the list of edge types
+        """
```
BarclayII, line 288: raise NotImplementedError?

BarclayII (changes requested):
Another set (not complete yet).

python/dgl/distributed/dist_graph.py
```diff
@@ -119,29 +117,57 @@ def _get_graph_from_shared_mem(graph_name):
...
         # initialize the node data in advance.
-        names = g._get_all_ndata_names()
-        policy = PartitionPolicy(NODE_PART_POLICY, g.get_partition_book())
-        self._data = {}
+        names = g._get_ndata_names(ntype)
+        self._data = g._ndata_store
         for name in names:
-            name1 = _get_data_name(name, policy.policy_str)
-            dtype, shape, _ = g._client.get_data_meta(name1)
+            assert name.is_node()
```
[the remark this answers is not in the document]
zheng-da: this is a sanity check.

python/dgl/distributed/dist_graph.py
```diff
@@ -119,29 +117,57 @@ def _get_graph_from_shared_mem(graph_name):
...
+            policy = PartitionPolicy(name.policy_str, g.get_partition_book())
```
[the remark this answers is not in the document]
zheng-da: is it too expensive to create?
I see `HeteroNodeView` is created every time as well. Should we buffer them?

python/dgl/distributed/graph_partition_book.py
zheng-da: # node_map_data and edge_map_data were serialized with pickle and converted into a list of bytes and then stored in a numpy array before being placed in shared memory. To deserialize, we need to reverse the process.

zheng-da: i added comments to explain the process.

examples/pytorch/rgcn/experimental/entity_classify_dist.py
```diff
@@ -138,32 +138,52 @@ class DistEmbedLayer(nn.Module):
...
+                        part_policy = dgl.distributed.NodePartitionPolicy(g.get_partition_book(), ntype)
+                        self.node_embeds[ntype] = dgl.distributed.DistEmbedding(g.number_of_nodes(ntype),
+                                self.embed_size,
+                                embed_name + '_' + ntype,
+                                init_emb,
+                                part_policy)
             else:
-                self.node_embeds = th.nn.Embedding(g.number_of_nodes(), self.embed_size, sparse=self.sparse_emb)
-                nn.init.uniform_(self.node_embeds.weight, -1.0, 1.0)
+                self.node_embeds = nn.ModuleDict()
```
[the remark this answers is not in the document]
zheng-da: this is developed for users to test embedding methods.

examples/pytorch/rgcn/experimental/entity_classify_dist.py
```diff
@@ -181,22 +201,41 @@ def forward(self, node_ids, node_tids, features):
...
             embeddings as the input of the next layer
         """
-        embeds = th.empty(node_ids.shape[0], self.embed_size)
-        for ntype in range(self.num_of_ntype):
-            assert features[ntype] is None, 'Currently Dist RGCN only support non input feature'
-            loc = node_tids == ntype
-            embeds[loc] = self.node_embeds(node_ids[loc])
-
+        embeds = th.empty(node_ids.shape[0], self.embed_size, device=self.dev_id)
+        for ntype_id in th.unique(ntype_ids):
```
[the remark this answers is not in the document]
zheng-da: how about converting it to numpy array or a list?

python/dgl/distributed/graph_partition_book.py
```diff
@@ -408,101 +494,238 @@ class RangePartitionBook(GraphPartitionBook):
...
+            assert node_map[key].shape == (num_parts, 2)
+            self._typed_nid_range[key] = node_map[key].reshape(-1, 2)
+            # This is used for per-node-type lookup.
+            self._typed_node_map[key] = np.cumsum(self._typed_nid_range[key][:, 1]
+                                                  - self._typed_nid_range[key][:, 0])
+            # This is used for homogeneous node Id lookup.
+            max_node_map = np.maximum(self._typed_nid_range[key][:, 1], max_node_map)
+        # This is a vector that indicates the last node Id in each partition.
+        # The Id is the global Id in the homogeneous representation.
+        self._node_map = node_map = max_node_map
```
jermainewang, line 552: Same. Everything is `xxx_map`. The name is not informative.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -408,101 +494,238 @@ class RangePartitionBook(GraphPartitionBook):
...
         number of total partitions
-    node_map : tensor
-        map global node id to partition id
-    edge_map : tensor
-        map global edge id to partition id
+    node_map : a dict of tensors
+        map global node id to partition id for each node type.
+        The key of the dict is the node type and the value is a 2-column matrix.
+        Each row represents the Id range of a partition, the first element is the starting Id
+        and the second element is the ending Id.
```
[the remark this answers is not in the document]
jermainewang: After reading the implementation, I found my understanding maybe wrong. Look like the `node_map` stores global ID ranges instead of local ID ranges. If so, pls doc it clearly.

python/dgl/id_map.py
```diff
@@ -0,0 +1,68 @@
...
+from . import utils
+
+class IdMap:
+    '''This maps node/edge Ids in the homogeneous form to per-type form.
+
+    It stores the Id ranges for each partition and for each type. The Ids are in
+    the homogeneous form (i.e., all nodes of different types have unique Ids). For each type,
+    all Ids in a partition fall in a contiguous range. The Id ranges are stored in
+    a matrix, where each row has two elements to represent the beginning and the end of
+    the range.
```
jermainewang, line 15: Even with the help of your doc, it is very hard to decipher this description. I could only make the following guess:
1. The input `id_ranges` is an object of type `dict[str, Tensor]`. It stores the same information as the `node_map` argument to `GraphPartitionBook`.
2. The input ranges are global ranges, meaning that it refers to ID ranges w.r.t. the entire graph.
3. The goal of this `IdMap` class is to parse a global node ID into its type ID and also the offset within the type (or called per-type ID). The function should support vectorization, i.e., process multiple ID conversions in parallel.

If my understanding is correct, we could formulate the problem as the following. Given a list of sorted values `A` and a list of arbitrary values `B`, return an index `i` for each element in `B` such that `A[i] <= the_element < A[i + 1]`. If we replace `A` with the cumsum array of the number of nodes of each type globally and `B` with the IDs to convert, the result is the type ID of each input. This is exactly [`numpy.searchsorted`](https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html). PyTorch added this op in 1.6.0. Getting the offset (per-type ID) of each element is also achievable using array operations:
```python
A = ...  # zero-based cumsum(num_nodes_per_type)
B = ...  # IDs to convert
I = searchsorted(A, B)  # type ID of each element in B
per_type_id = B - A[I]
```

python/dgl/distributed/graph_partition_book.py
jermainewang: Shadowing the input `node_map` variable here is really bad. It makes the program quite hard to read. Why not directly use `max_node_map`? Also, if `self._node_map` and `self._type_nid_range` really represent the same thing, why save two members?

python/dgl/distributed/graph_partition_book.py
```diff
@@ -408,101 +494,238 @@ class RangePartitionBook(GraphPartitionBook):
...
+        """
+        return self._eid_map(ids)
+
+    def map_to_homo_nid(self, ids, ntype):
+        """Map per-node-type Ids to global node Ids in the homogeneous format.
+        """
+        ids = utils.toindex(ids).tousertensor()
+        partids = self.nid2partid(ids, ntype)
+        end_diff = F.tensor(self._typed_node_map[ntype])[partids] - ids
+        # TODO(zhengda) make everything operate on Pytorch tensors.
```
jermainewang, line 665: Is  this TODO still needed?

python/dgl/distributed/graph_services.py
```diff
@@ -286,22 +288,53 @@ def sample_neighbors(g, nodes, fanout, edge_dir='in', prob=None, replace=False):
 
         For sampling without replacement, if fanout > the number of neighbors, all the
         neighbors are sampled. If fanout == -1, all neighbors are collected.
+    output_format : str, optional
+        The format of the sampled subgraph. A user can specify the output subgraph to be
+        stored in a homogeneous graph format or in a heterogeneous graph format. The valid
+        values are "homogeneous" and "heterogeneous". This argument is only valid if
+        the input graph is heterogeneous.
```
jermainewang, line 295: Do we support `heterogeneous` format now?

python/dgl/partition.py
```diff
@@ -46,6 +46,59 @@ def reorder_nodes(g, new_node_ids):
 def _get_halo_heterosubgraph_inner_node(halo_subg):
     return _CAPI_GetHaloSubgraphInnerNodes_Hetero(halo_subg)
 
+def reshuffle_graph(g, node_part=None):
+    '''Reshuffle node ids and edge Ids of a graph.
+
+    We need to reshuffle them so that all nodes/edges of the same type have contiguous Ids.
+    In the case that a graph is partitioned, all nodes/edges in a partition should
+    get contiguous Ids; within a partition, all nodes/edges of the same type have contigous Ids.
+    '''
```
jermainewang, line 55: Is this a user-facing API? If so, missing docstring for argument and return type. It also should include an example section.

src/graph/transform/partition_hetero.cc
```diff
@@ -236,6 +236,13 @@ DGL_REGISTER_GLOBAL("partition._CAPI_DGLPartitionWithHalo_Hetero")
...
   });
 
+template<class IdType>
+struct edge_property {
+  IdType eid;
+  int64_t idx;
+  int part_id;
+};
+
 // TODO(JJ): What's this?
```
jermainewang, line 246: What is this TODO @VoVAllen ?

src/graph/transform/partition_hetero.cc
```diff
@@ -245,26 +252,54 @@ DGL_REGISTER_GLOBAL("partition._CAPI_DGLReassignEdges_Hetero")
...
+        std::sort(indexed_eids.begin(), indexed_eids.end(), comp);
+      }
+      for (int64_t new_eid = 0; new_eid < num_edges; new_eid++) {
+        int64_t orig_idx = indexed_eids[new_eid].idx;
+        typed_new_data[new_eid] = typed_data[orig_idx];
+        typed_data[orig_idx] = new_eid;
       }
     });
+    ugptr->InvalidateCSR();
+    ugptr->InvalidateCOO();
```
jermainewang, line 302: Why need to invalidate them?

jermainewang (changes requested):

python/dgl/distributed/graph_partition_book.py
zheng-da: I think "feature" isn't very general for the data stored on nodes and edges. We can store labels, masks, etc on nodes/edges. Do we refer to them as features as well?

zheng-da: I notice that the doc string of DGLHeteroGraph uses "feature" a lot. Does it cause confusion?

python/dgl/distributed/graph_partition_book.py
zheng-da: actually, one TODO I have is that the current API doesn't support canonical edge types. Will it be quick to support it?

python/dgl/distributed/graph_partition_book.py
zheng-da: what is the practice in other places?

python/dgl/distributed/graph_partition_book.py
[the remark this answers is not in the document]
zheng-da: yes, `node_map` stores the global IDs. I have corrected it.

We cannot store the number of nodes for each parition. these are global IDs. The ID ranges of a node type in different partitions are connected.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -408,101 +494,238 @@ class RangePartitionBook(GraphPartitionBook):
...
+        # This stores the node Id map for per-node-type Ids in each partition.
+        # The key is the node type, the value is a NumPy vector which indicates
+        # the last node Id in a partition.
+        self._typed_node_map = {}
+        max_node_map = np.zeros((num_parts,), dtype=np.int64)
+        for key in node_map:
+            if not isinstance(node_map[key], np.ndarray):
+                node_map[key] = F.asnumpy(node_map[key])
+            assert node_map[key].shape == (num_parts, 2)
+            self._typed_nid_range[key] = node_map[key].reshape(-1, 2)
```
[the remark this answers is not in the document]
zheng-da: It looks like that. I'll remove the reshape.

For the second question, I think so.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -408,101 +494,238 @@ class RangePartitionBook(GraphPartitionBook):
...
+            # This is used for per-node-type lookup.
+            self._typed_node_map[key] = np.cumsum(self._typed_nid_range[key][:, 1]
+                                                  - self._typed_nid_range[key][:, 0])
```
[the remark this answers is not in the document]
zheng-da: this is ID map is similar to the input ID map. The main difference is that it stores the per-type IDs, so we don't need to store the starting and ending IDs. It's sufficient to store the ending IDs.

src/graph/transform/partition_hetero.cc
zheng-da: this is not required by distributed heterogeneous graphs.
I found a bug in the original code. When we reassign edge IDs, we need to invalidate other data structures. otherwise, there will be problems. I can move the bug fix to another PR.

python/dgl/partition.py
zheng-da: i think this function is used internally. For example, if you want to develop the sampling algorithm with type-specific sampling probability. I have added arguments and returns.

python/dgl/distributed/graph_services.py
zheng-da: In the end of the function, I tried to convert a sampled subgraph into a heterogeneous graph. so it does support heterogeneous graph. It's kind of slow. The question is whether this is necessary.

python/dgl/distributed/graph_partition_book.py
zheng-da: fixed for overwriting `node_map`.
`_node_map` and `_type_nid_range` are diferent. `_node_map` only stores the ending IDs. This is used by `np.searchsorted` for range search. otherwise, we have write a customized version of the range search.

python/dgl/distributed/graph_partition_book.py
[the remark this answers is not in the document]
zheng-da: rename `xxx_map` with `max_xxx_ids`

python/dgl/distributed/graph_partition_book.py
BarclayII: I think it'll be raising NotImplementedError if you don't use abstract classes.

examples/pytorch/rgcn/experimental/entity_classify_dist.py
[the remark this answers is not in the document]
BarclayII: Sure numpy array works.

examples/pytorch/rgcn/experimental/entity_classify_dist.py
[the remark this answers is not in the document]
BarclayII: If this is for distributed training, do we still want to give users this freedom?

python/dgl/distributed/graph_partition_book.py
BarclayII: Let's stick to the name "feature" for now and change it once we hear someone complaining.

python/dgl/distributed/graph_partition_book.py
BarclayII: Not sure.  You can leave it for now.

python/dgl/distributed/graph_services.py
```diff
@@ -286,22 +288,53 @@ def sample_neighbors(g, nodes, fanout, edge_dir='in', prob=None, replace=False):
...
+        eid_per_type_id, eid_type = gpb.map_to_per_etype(subg.edata[dgl.EID])
+        subg_edges = {}
+        for etype in g.canonical_etypes:
+            etype_id = g.get_etype_id(etype[1])
+            edge_idx = eid_type == etype_id
+            subg_edges[etype] = (F.boolean_mask(src_per_type_id, edge_idx),
+                                 F.boolean_mask(dst_per_type_id, edge_idx),
+                                 F.boolean_mask(eid_per_type_id, edge_idx))
+        num_nodes = {ntype:g.number_of_nodes(ntype) for ntype in g.ntypes}
+        subg = heterograph(subg_edges, num_nodes_dict=num_nodes)
```
BarclayII, line 336: Missing import.  Also does `heterograph` support taking in a triplet?

Also it seems that unit test did not cover this branch.  Maybe just remove this functionality?

python/dgl/distributed/partition.py
```diff
@@ -55,15 +86,46 @@ def load_partition(part_config, part_id):
...
+    for etype in etypes:
+        etype_id = etypes[etype]
+        # graph.edata[EID] are global homogeneous edge Ids.
+        eids = F.boolean_mask(graph.edata[EID], _get_inner_edge_mask(graph, etype_id))
+        partids1 = gpb.eid2partid(eids)
+        _, per_type_eids = gpb.map_to_per_etype(eids)
+        partids2 = gpb.eid2partid(per_type_eids, etype)
+        assert np.all(F.asnumpy(partids1 == part_id)), 'load a wrong partition'
+        assert np.all(F.asnumpy(partids2 == part_id)), 'load a wrong partition'
+    return graph, node_feats, edge_feats, gpb, graph_name, ntypes, etypes
```
BarclayII, line 128: Missing docstring for returned ntypes and types.  The main text may also need to change accordingly to include them.

python/dgl/distributed/partition.py
```diff
@@ -187,15 +287,39 @@ def partition_graph(g, graph_name, num_parts, out_path, num_hops=1, part_method=
     If node IDs and edge IDs are not shuffled to ensure that all nodes/edges in a partition
     fall into a contiguous ID range, DGL needs to store node/edge mappings (from
     node/edge IDs to partition IDs) in separate files (node_map.npy and edge_map.npy).
-    The node/edge mappings are stored in numpy files.
+    The node/edge mappings are stored in numpy files. **Note**: this format is
```
BarclayII, line 290: Use note with the following format:

```
.. note::

   <main-text>
```

Also it's probably better to use `.. warning::` since the interface will be deprecated.

python/dgl/distributed/partition.py
```diff
@@ -187,15 +287,39 @@ def partition_graph(g, graph_name, num_parts, out_path, num_hops=1, part_method=
...
+    deprecated and will not be supported by the next release. In other words, the future
+    release will always shuffle node Ids and edge Ids when partitioning a graph.
+
+    If node Ids and edge Ids are shuffled, `node_map` and `edge_map` contains the information
```
BarclayII, line 294: `Id` -> `ID` and `Ids` -> `IDs`.  Should be consistent.

Please scan other places as well.

Also, `If node IDs and edge IDs are shuffled` ->
```
If ``reshuffle = True``
```

python/dgl/distributed/partition.py
```diff
@@ -187,15 +287,39 @@ def partition_graph(g, graph_name, num_parts, out_path, num_hops=1, part_method=
...
+    is as follows:
+
+    .. code-block:: none
+
+        {
+            "node_type": [ [ part1_start, part1_end ],
+                           [ part2_start, part2_end ],
+                           ... ],
+            ...
+        },
```
BarclayII, line 307: remove comma

python/dgl/distributed/partition.py
```diff
@@ -177,8 +276,9 @@ def partition_graph(g, graph_name, num_parts, out_path, num_hops=1, part_method=
     * `part_method` is the method used to assign nodes to partitions.
       Currently, it supports "random" and "metis".
     * `num_parts` is the number of partitions.
-    * `halo_hops` is the number of HALO nodes we want to include in a partition.
+    * `halo_hops` is the number of hops of nodes we include in a partition as HALO nodes.
```
BarclayII, line 279: Use either
```
``halo_hops``
```
or
```
:attr:`halo_hops`
```
for monospace fonts.  Single backquotes won't render correctly.

Please scan for other places as well as there are a lot of misuses..

python/dgl/distributed/partition.py
```diff
@@ -258,26 +383,115 @@ def partition_graph(g, graph_name, num_parts, out_path, num_hops=1, part_method=
     >>> g, node_feats, edge_feats, gpb, graph_name = dgl.distributed.load_partition(
     ...                                 'output/test.json', 0)
     '''
+    def get_homogeneous(g, balance_ntypes):
```
BarclayII, line 386: Any particular reason to put this inside a function?

python/dgl/distributed/partition.py
```diff
@@ -187,15 +287,39 @@ def partition_graph(g, graph_name, num_parts, out_path, num_hops=1, part_method=
...
+    .. code-block:: none
+
+        {
+            "node_type": [ [ part1_start, part1_end ],
+                           [ part2_start, part2_end ],
+                           ... ],
+            ...
+        },
+
+    Essentially, the data in `node_map` and `edge_map` is a dictionary. The key is node type
```
BarclayII, line 309: `node/edge type`

Please also check for similar occurrences.

python/dgl/distributed/partition.py
```diff
@@ -187,15 +287,39 @@ def partition_graph(g, graph_name, num_parts, out_path, num_hops=1, part_method=
...
+    and the value is partitioning information of the node type. The partition information
+    is stored as a list of two-element tuples. The length of the list is the number of
+    partitions; each element in the list is a tuple that stores the start and the end of
+    an Id range for a particular node/edge type in the partition.
```
BarclayII, line 313: My rephrasing:
```
``node_map`` and ``edge_map`` are both dictionaries.  The keys are node/edge types.
The values are lists of pairs containing the start and end of the ID interval for the
corresponding types, one pair for each partition.
```

BarclayII (changes requested):
Python code finished reviews

src/graph/graph_op.cc
```diff
@@ -719,4 +719,61 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLMapSubgraphNID")
...
+  const IdType *ids_data = static_cast<IdType *>(ids->data);
+  const IdType *typed_map_data = static_cast<IdType *>(typed_map->data);
+  IdType *types_data = static_cast<IdType *>(ret->data);
+  IdType *per_type_ids_data = static_cast<IdType *>(ret->data) + num_ids;
+#pragma omp parallel for
+  for (int64_t i = 0; i < ids->shape[0]; i++) {
+    IdType id = ids_data[i];
+    auto it = std::lower_bound(range_end_data, range_end_data + num_ranges, id);
+    // The range must exist.
+    CHECK(it != range_end_data + num_ranges);
```
BarclayII, line 740: If it fails because of a bug, consider `BUG_ON`.  Otherwise you need an error message.

src/graph/graph_op.cc
```diff
@@ -719,4 +719,61 @@ DGL_REGISTER_GLOBAL("graph_index._CAPI_DGLMapSubgraphNID")
...
+    size_t range_id = it - range_end_data;
+    int type_id = range_id % num_types;
+    types_data[i] = type_id;
+    int part_id = range_id / num_types;
+    assert(part_id < num_parts);
```
BarclayII, line 745: Ditto.

src/graph/transform/partition_hetero.cc
```diff
@@ -236,7 +236,15 @@ DGL_REGISTER_GLOBAL("partition._CAPI_DGLPartitionWithHalo_Hetero")
     *rv = ret_list;
   });
 
-// TODO(JJ): What's this?
+template<class IdType>
+struct edge_property {
```
BarclayII, line 240: `EdgeProperty`

BarclayII (changes requested):
C++ code reviews.

python/dgl/distributed/graph_services.py
jermainewang: If there is no strong motivation, we should consider omit it in this release, especially if later we found a better heterograph solution.

python/dgl/id_map.py
zheng-da: I think it's more complex than that because we need to deal with partitions as well. Here is an example:
```
+---------+-----+----------
  range   | type | partition
(0, 100)  |   0    |    0
(100,200) |   1    |    0
(200,300) |   0    |    1
(300,400) |   1    |    1
```
With binary search, we can get its type, but getting per-type ID is a little more difficult.
I guess it's still doable with pure tensor operations. let me try

python/dgl/distributed/partition.py
zheng-da: this should be right. there are other items I skipped in the dict.

python/dgl/distributed/partition.py
zheng-da: no particular reasons. this function is just very specific to `partition_graphs`.

python/dgl/distributed/dist_tensor.py
```diff
@@ -147,7 +140,9 @@ def __init__(self, shape, dtype, name=None, init_func=None, part_policy=None,
             # Thus, all of them should have the same Ids.
             name = 'anonymous-' + get_role() + '-' + str(DIST_TENSOR_ID)
             DIST_TENSOR_ID += 1
-        self._name = _get_data_name(name, part_policy.policy_str)
+        assert isinstance(name, str), 'name {} is type {}'.format(name, type(name))
```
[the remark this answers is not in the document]
zheng-da: `name` is an argument provided by users. we need to make sure it's a string.

python/dgl/distributed/graph_partition_book.py
zheng-da: I'm not sure. the data serialized to the byte array can have a fairly complex data structure.

python/dgl/distributed/graph_partition_book.py
zheng-da: you mean both `None` will be considered as `False`?

examples/pytorch/rgcn/experimental/README.md
```diff
@@ -78,7 +78,7 @@ python3 ~/dgl/tools/launch.py \
 --num_samplers 4 \
```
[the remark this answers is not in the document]
zheng-da: this should be explained in the user guide and tutorials.

examples/pytorch/rgcn/experimental/README.md
```diff
@@ -98,5 +98,5 @@ python3 partition_graph.py --dataset ogbn-mag --num_parts 1
 
 ### Step 2: run the training script
 ```bash
-python3 entity_classify_dist.py --graph-name ogbn-mag  --dataset ogbn-mag --fanout='25,25' --batch-size 256 --n-hidden 64 --lr 0.01 --eval-batch-size 8 --low-mem --dropout 0.5 --use-self-loop --n-bases 2 --n-epochs 3 --layer-norm --ip-config ip_config.txt --conf-path 'data/ogbn-mag.json' --standalone
+python3 entity_classify_dist.py --graph-name ogbn-mag  --dataset ogbn-mag --fanout='25,25' --batch-size 512 --n-hidden 64 --lr 0.01 --eval-batch-size 128 --low-mem --dropout 0.5 --use-self-loop --n-bases 2 --n-epochs 3 --layer-norm --ip-config ip_config.txt --conf-path 'data/ogbn-mag.json' --standalone  --sparse-embedding  --sparse-lr 0.06 --node-feats
 ```
```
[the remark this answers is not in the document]
zheng-da: we will need to specify the path to `data/ogbn-mag.json`. When partitioning a graph, we actually allow users to specify the path.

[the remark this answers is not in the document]
zheng-da: I have update README for your first comment.

examples/pytorch/rgcn/experimental/entity_classify_dist.py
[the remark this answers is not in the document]
zheng-da: this is originally implemented by xiang to provide users options of using different embedding implementations.

examples/pytorch/rgcn/experimental/entity_classify_dist.py
```diff
@@ -181,22 +201,41 @@ def forward(self, node_ids, node_tids, features):
...
+            ntype = self.ntype_id_map[int(ntype_id)]
+            loc = ntype_ids == ntype_id
+            if feat_name in self.g.nodes[ntype].data:
+                embeds[loc] = self.node_projs[ntype](self.g.nodes[ntype].data[feat_name][node_ids[ntype_ids == ntype_id]].to(self.dev_id))
```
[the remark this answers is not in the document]
zheng-da: our mini-batch is stored in a homogeneous graph. We have to make all node data in one tensor as input.

examples/pytorch/rgcn/experimental/entity_classify_dist.py
```diff
@@ -274,29 +315,32 @@ def sample_blocks(self, seeds):
...
         ntypes = []
         seeds = th.LongTensor(np.asarray(seeds))
-        cur = seeds
+        gpb = self.g.get_partition_book()
+        cur = gpb.map_to_homo_nid(seeds, 'paper')
         for fanout in self.fanouts:
-            frontier = self.sample_neighbors(self.g, cur, fanout, replace=True)
-            etypes = self.g.edata[dgl.ETYPE][frontier.edata[dgl.EID]]
-            norm = self.g.edata['norm'][frontier.edata[dgl.EID]]
+            frontier = self.sample_neighbors(self.g, cur, fanout, replace=False)
```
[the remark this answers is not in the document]
zheng-da: yes, the mini-batch is stored as homogeneous graphs.

python/dgl/distributed/graph_partition_book.py
jermainewang: Yes, that's python standard.

python/dgl/id_map.py
```diff
@@ -0,0 +1,91 @@
...
+        ranges = np.zeros((self.num_parts * self.num_types, 2), dtype=np.int64)
+        typed_map = []
+        id_ranges = list(id_ranges.values())
+        id_ranges.sort(key=lambda a: a[0, 0])
+        for i, id_range in enumerate(id_ranges):
+            ranges[i::self.num_types] = id_range
+            map1 = np.cumsum(id_range[:, 1] - id_range[:, 0])
+            typed_map.append(map1)
+        print(ranges)
+        print(typed_map)
```
jermainewang, line 57: remove

python/dgl/distributed/partition.py
BarclayII: I would move this outside the function and make it private (prefix with an underscore).  Otherwise the function definition will be run every time you call it.  Although that may not be a big deal.

examples/pytorch/rgcn/experimental/README.md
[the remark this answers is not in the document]
BarclayII: My question is, how do a user know the path?  And why is the path `data/ogbn-mag.json`?  Is it always `data/<graph-name>.json`?  If so, we should state it, preferably earlier (e.g. in the `This script generates...` you have updated).

examples/pytorch/rgcn/experimental/entity_classify_dist.py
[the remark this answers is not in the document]
BarclayII: @classicsong Do you think this freedom is necessary?  I'm good if it's indeed the case.

examples/pytorch/rgcn/experimental/entity_classify_dist.py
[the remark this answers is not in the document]
BarclayII: This is a discrepancy and makes the code hard to understand, especially since `DistEmbedLayer` and `NeighborSampler` are initialized with the same `g`.  Do we explain the mechanic somewhere (e.g. where are the conversions from/to heterogeneous graph taking place)?

examples/pytorch/rgcn/experimental/README.md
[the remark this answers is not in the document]
zheng-da: i changed the readme to explain which directory stores the output data.

examples/pytorch/rgcn/experimental/entity_classify_dist.py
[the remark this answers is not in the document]
zheng-da: if we should modify it, let's do it in another PR. this is PR is already too large.

examples/pytorch/rgcn/experimental/entity_classify_dist.py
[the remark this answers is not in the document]
zheng-da: this is a temporary solution. we need to fix heterograph mini-batch training first before we can fix this problem.
We'll have to use a doc to explain what is happening here.

examples/pytorch/rgcn/experimental/entity_classify_dist.py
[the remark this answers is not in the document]
BarclayII: Sure

python/dgl/distributed/id_map.py
```diff
@@ -0,0 +1,89 @@
...
+    This class is to map any homogeneous ID to a per-type ID and its type. For example,
+    homogeneous node ID 90 is mapped to type 0 and per-type ID 90; homogeneous node ID
+    201 is mapped to type 0 and per-type ID 101.
+
+    Parameters
+    ----------
+    id_ranges : dict[str, Tensor].
+        The ID ranges of each partition for each type. It stores the same information
+        as the `node_map` argument in `RangePartitionBook`.
+    '''
```
jermainewang, line 44:
```
'''A hash map for converting node/edge IDs to their type IDs and type-wise IDs.

    For a heterogeneous graph, DGL assigns an integer ID to each node/edge type;
    node and edge of different types have independent IDs starting from zero.
    Therefore, a node/edge can be uniquely identified by an ID pair,
    ``(type_id, type_wise_id)``. To make it convenient for distributed processing,
    DGL further encodes the ID pair into one integer ID, which we refer to
    as *homogeneous ID*.

    DGL arranges nodes and edges so that all nodes of the same type have contiguous
    homogeneous IDs. If the graph is partitioned, the nodes/edges of the same type
    within a partition have contiguous homogeneous IDs.

    Below is an example adjancency matrix of an unpartitioned heterogeneous graph
    stored using the above ID assignment. Here, the graph has two types of nodes
    (``T0`` and ``T1``), and four types of edges (``R0``, ``R1``, ``R2``, ``R3``).
    There are a total of 400 nodes in the graph and each type has 200 nodes. Nodes
    of type 0 have IDs in [0,200), while nodes of type 1 have IDs in [200, 400).

    ```
        0 <- T0 -> 200 <- T1 -> 400
     0  +-----------+------------+
        |           |            |
     ^  |    R0     |     R1     |
     T0 |           |            |
     v  |           |            |
    200 +-----------+------------+
        |           |            |
     ^  |    R2     |     R3     |
     T1 |           |            |
     v  |           |            |
    400 +-----------+------------+
    ```

    Below shows the adjacency matrix after the graph is partitioned into two.
    Note that each partition still has two node types and four edge types,
    and nodes/edges of the same type have contiguous IDs.

    ```
                partition 0              partition 1

        0 <- T0 -> 100 <- T1 -> 200 <- T0 -> 300 <- T1 -> 400
     0  +-----------+------------+-----------+------------+
        |           |            |                        |
     ^  |    R0     |     R1     |                        |
     T0 |           |            |                        |
     v  |           |            |                        |
    100 +-----------+------------+                        |
        |           |            |                        |
     ^  |    R2     |     R3     |                        |
     T1 |           |            |                        |
     v  |           |            |                        |
    200 +-----------+------------+-----------+------------+
        |                        |           |            |
     ^  |                        |    R0     |     R1     |
     T0 |                        |           |            |
     v  |                        |           |            |
    100 |                        +-----------+------------+
        |                        |           |            |
     ^  |                        |    R2     |     R3     |
     T1 |                        |           |            |
     v  |                        |           |            |
    200 +-----------+------------+-----------+------------+
    ```

    The following table is an alternative way to represent the above ID assignments.
    It is easy to see that the homogeneous ID range [0, 100) is used for nodes of type 0
    in partition 0, [100, 200) is used for nodes of type 1 in partition 0, and so on.
    ```
    +---------+------+----------
      range   | type | partition
    [0, 100)  |   0  |    0
    [100,200) |   1  |    0
    [200,300) |   0  |    1
    [300,400) |   1  |    1
    ```

    The goal of this class is to, given a node's homogenous ID, convert it into the
    ID pair ``(type_id, type_wise_id)``. For example, homogeneous node ID 90 is mapped
    to (0, 90); homogeneous node ID 201 is mapped to (0, 101).

    Parameters
    ----------
    id_ranges : dict[str, Tensor].
        Node ID ranges within partitions for each node type. The key is the node type
        name in string. The value is a tensor of shape :math:`(K, 2)`, where :math:`K` is
        the number of partitions. Each row has two integers: the starting and the ending IDs
        for a particular node type in a partition. For example, all nodes of type ``"T"`` in
        partition ``i`` has ID range ``id_ranges["T"][i][0]`` to ``id_ranges["T"][i][1]``.
        It is the same as the `node_map` argument in `RangePartitionBook`.
    '''
```

python/dgl/distributed/id_map.py
```diff
@@ -0,0 +1,89 @@
...
+
+        Parameters
+        ----------
+        ids : 1D tensor
+            The homogeneous ID.
+
+        Returns
+        -------
+            type_ids, per_type_ids
+        '''
```
jermainewang, line 74:
```
'''Convert the given homogenous node IDs into (type_id, type_wise_id) pairs.

Parameters
------------
ids : Tensor
    The homogeneous node IDs stored in a 1D tensor.

Returns
--------
type_ids : Tensor
    Type IDs of each node.
type_wise_ids : Tensor
    Type-wise IDs of each node.
'''
```

jermainewang (approved):

jermainewang:
@zheng-da Uploaded my take on explaining the ID shenanigans. I feel the description should be included in the user guide too. Putting it in the docstring of an internal `IdMap` class just hides it from the users. I also suggest formally define
* *homogeneous ID* as the ID that encodes type ID and type-wise ID.
* Use *type-wise ID* instead of *per-type ID*.

Otherwise, I'm good with this PR.

PR #2462, opened 2020-12-27
[Test] Regression speed test for GAT neighbor sampler

classicsong:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/benchmarks/benchmarks/model_speed/bench_gat_ns.py
+++ b/benchmarks/benchmarks/model_speed/bench_gat_ns.py
@@ -151,4 +151,4 @@ def track_time(data):
 
     t1 = time.time()
 
-    return t1 - t0
+    return (t1 - t0) / num_epochs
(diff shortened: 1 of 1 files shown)


benchmarks/benchmarks/model_speed/bench_gat_ns.py
```diff
@@ -0,0 +1,154 @@
...
+
+    num_epochs = 20
+    num_hidden = 16
+    num_heads = 8
+    num_layers = 2
+    fan_out = '10,25'
+    batch_size = 1024
+    lr = 0.003
+    dropout = 0.5
+    num_workers = 4
```
jermainewang, line 93: We should test a variety of configurations later but also would like to keep the benchmarking cost in a manageable range. Thoughts?

jermainewang (approved):
LGTM

benchmarks/benchmarks/model_speed/bench_gat_ns.py
classicsong: The range can be selected based on the performance bottlenecks. In config0, the system can be CPU bounded and in config1, the system can be memory bounded, etc.

PR #2464, opened 2020-12-28
docs: fix simple typo, tranfer -> transfer

timgates42:
There is a small typo in python/dgl/dataloading/async_transferer.py.

Should read `transfer` rather than `tranfer`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


jermainewang:
Thanks for the fix!

PR #2465, opened 2020-12-29
[Bugfix] Handle a Corner Case of Batching after Removing Nodes/Edges

### LINKED ISSUE #2409: Batching fails when remove_edges is called on part of the graphs
## 🐛 Bug

In my graph dataloader, `g.remove_edges(zero_eid)` is called on each graph before batching, but for some graphs `zero_eid` has no elements. When calling `dgl.batch`, I got the following error:

> dgl._ffi.base.DGLError: Expect all graphs to have the same schema on nodes["ao"].data, but graph 6 got
>         {'n': Scheme(shape=(), dtype=torch.int64), 'l': Scheme(shape=(), dtype=torch.int64)}
> which is different from
>         {'n': Scheme(shape=(), dtype=torch.int64), 'l': Scheme(shape=(), dtype=torch.int64), '_ID': Scheme(shape=(), dtype=torch.int64)}.

## Suggested changes

I used the following workaround that changes [L680-690 of dgl/heterograph.py](https://github.com/dmlc/dgl/blob/master/python/dgl/heterograph.py#L680) to:
```
         if len(eids) > 0:
             assert self.number_of_edges(etype) > F.as_scalar(F.max(eids, dim=0)), \
                 "The input eid {} is out of the range [0:{})".format(
                     F.as_scalar(F.max(eids, dim=0)), self.number_of_edges(etype))

         # edge_subgraph
         edges = {}
         u_type, e_type, v_type = self.to_canonical_etype(etype)
         for c_etype in self.canonical_etypes:
             # the target edge type
             if c_etype == (u_type, e_type, v_type) and len(eids) > 0:
                 origin_eids = self.edges(form='eid', order='eid', etype=c_etype)
                 edges[c_etype] = utils.compensate(eids, origin_eids)
             else:
                 edges[c_etype] = self.edges(form='eid', order='eid', etype=c_etype)
```

which enforce `self.edge_subgraph` to be called when the number of edges to be removed is zero. I can make a PR if this is a desirable solution. Thanks!

## Environment

 - DGL Version: 0.5.3
 - Backend Library & Version: PyTorch 1.7.0
 - OS: Linux
 - How you installed DGL: pip
 - Python version: 3.7
 - CUDA/cuDNN version (if applicable): 10.2
 - GPU models and configuration: V100

mufeili:
## Description
Fix #2409 

@zrqiao With this PR, `remove_nodes`/`remove_edges` no longer store the original node/edge IDs. See if this is good to you.

## Checklist
- [x] Code is well-documented


mufeili:
@BarclayII Any idea for the error reported in CI regarding neighbor sampling?

python/dgl/subgraph.py
```diff
@@ -52,6 +49,10 @@ def node_subgraph(graph, nodes):
         If the graph is homogeneous, one can directly pass the above formats.
         Otherwise, the argument must be a dictionary with keys being node types
         and values being the nodes.
+    store_raw_ids : bool, optional
+        If True, it will store the raw IDs of the extracted nodes and edges in the ``ndata``
+        and ``edata`` of the resulting graph under name ``dgl.NID`` and ``dgl.EID``,
+        respectively.
```
jermainewang, line 55: `store_raw_ids` or `store_ids`?

jermainewang (approved):

BarclayII:
The code that threw error is:
```python
# (BarclayII) If frontier already has a EID field and located_eids is empty,
# the returned graph will keep EID intact.  Otherwise, EID will change
# to the mapping from the new graph to the old frontier.
# So we need to test if located_eids is empty, and do the remapping ourselves.
if len(located_eids) > 0:
    frontier = transform.remove_edges(frontier, located_eids)
    frontier.edata[EID] = F.gather_row(parent_eids, frontier.edata[EID])      
```
What the code does is to remove the edges in `frontier` that has its `EID` in the given `exclude_eids`.

If `remove_edges` no longer returns the edge ID of the parent graph, how should I know the edge correspondence between the original graph and the new graph?

zrqiao (commented):
Thanks! The changes look good to me, and I also suggest clarify the behaviors when perform batching on subgraphs.

- [ ] Consider adding docs to `dgl.batch()`

zrqiao:
@mufeili Regarding that check error - I am not familiar with your CI workflow, but do we need the original NID/EIDs to be accessible from subgraphs in other functions? Perhaps the `store_raw_ids` should be `True` by default.

python/dgl/subgraph.py
mufeili: Done.

mufeili:
@BarclayII I've made the update based on our discussion. Can you take a look?

mufeili:
@zrqiao I've chatted with @BarclayII regarding the issue and we decided to add a flag to `remove_nodes`/`remove_edges` for storing raw ids and set it to False by default. Internally it's easy to set the flag to True when necessary.

Regarding the doc, I think it might be better to add doc for `remove_nodes/remove_edges`. With the flag default to False, do you still prefer a note in the doc?

zrqiao (approved):

zrqiao:
@mufeili Thanks! I would also prefer the flag default to False, as long as that does not break other utilities. Approved.

BarclayII:
I'm good.

PR #2468, opened 2020-12-30
[Performance][Test] New low memory implementation for RGCN and related regression tests

jermainewang:
## Description

Implement a new low memory implementation for RGCN which should be faster than the old one. The trick is to sort the graph by the edge types first to avoid random memory access when calculating messages for each edge type. To avoid sorting the input graph every time (which is a problem for full graph training), the PR extends the interface of RGCN's forward function. The `etypes` argument now accepts a *count array*. It is an integer list where each element is the number of edges of that edge type. Under this input format, the code assumes the graph has already been sorted.

How to sort a graph based on edge types and generate the count array is not obvious to many users. In fact, the code is also complicated and depends on some specific operator just added to PyTorch (see the code [here](https://github.com/dmlc/dgl/blob/49b331cd21d508e3507e3ea5c00f361e8dde4997/python/dgl/nn/pytorch/conv/relgraphconv.py#L315)). Fortunately, such type of graphs are usually results from `to_homogeneous`, and the API already produces graph with this property. Therefore, this PR formally introduces this behavior in `to_homogeneous` and adds two flags:
  - `store_type`: If True, store type information as the `dgl.NTYPE` and `dgl.ETYPE` features in the returned graph. The default is True and is our original behavior.
  - `return_count`: If True, return type information as an integer list; the i^th element corresponds to the number of nodes/edges of the i^th type.

Performance diff:
Dataset: AIFB
Hardware: V100
Baseline: master build
Metric: Epoch training time in seconds
* `low_mem=False`: the performance is the same
* `low_mem=True` and `etypes` argument is Tensor: master is 0.0989s, this PR is 0.0227s, (4.35x faster)
* `low_mem=True` and `etypes` argument is the count array: this PR is 0.0206s, (4.8x faster)

Other improvements:
- [x] Unittest.
- [x] Add two parameters "lowmem" and "sortgraph" to the RGCN benchmarks.
- [x] Improved the README of benchmark pipeline.

Related changes for future PRs:
- [ ] MXNet and TensorFlow implementation.
- [ ] Update the RGCN example

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/convert.py
+++ b/python/dgl/convert.py
@@ -605,17 +605,20 @@ def to_homogeneous(G, ndata=None, edata=None, store_type=True, return_count=Fals
     -------
     DGLGraph
         A homogeneous graph.
-    ntype_count : list[int]
-        Number of nodes of each type.
-    etype_count : list[int]
-        Number of edges of each type.
+    ntype_count : list[int], optional
+        Number of nodes of each type. Return when ``return_count`` is True.
+    etype_count : list[int], optional
+        Number of edges of each type. Return when ``return_count`` is True.
 
-    Note
-    ----
-    Calculating type information may introduce noticeable cost. Setting both ``store_type``
-    and ``return_count`` to False can avoid such cost if type information is not needed.
-    Otherwise, DGL recommends to use ``store_type=False`` and ``return_count=True`` due
-    to its memory efficiency.
+    Notes
+    -----
+
+    * Calculating type information may introduce noticeable cost. Setting both ``store_type``
+      and ``return_count`` to False can avoid such cost if type information is not needed.
+      Otherwise, DGL recommends to use ``store_type=False`` and ``return_count=True`` due
+      to its memory efficiency.
+    * The ``ntype_count`` and ``etype_count`` lists can help speed up some operations.
+      See :class:`~dgl.nn.pytorch.conv.RelGraphConv` for such an example.
 
     Examples
     --------
--- a/python/dgl/convert.py
+++ b/python/dgl/convert.py
@@ -598,7 +598,7 @@ def to_homogeneous(G, ndata=None, edata=None, store_type=True, return_count=Fals
         If True, store type information as the ``dgl.NTYPE`` and ``dgl.ETYPE`` features
         in the returned graph.
     return_count : bool, optional
-        If True, return type information as an integer list; the i^th corresponds to
+        If True, return type information as an integer list; the i^th element corresponds to
         the number of nodes/edges of the i^th type.
 
     Returns
--- a/python/dgl/convert.py
+++ b/python/dgl/convert.py
@@ -564,16 +564,21 @@ def to_hetero(G, ntypes, etypes, ntype_field=NTYPE, etype_field=ETYPE,
     return to_heterogeneous(G, ntypes, etypes, ntype_field=ntype_field,
                             etype_field=etype_field, metagraph=metagraph)
 
-def to_homogeneous(G, ndata=None, edata=None):
+def to_homogeneous(G, ndata=None, edata=None, store_type=True, return_count=False):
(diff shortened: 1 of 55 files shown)


jermainewang:
@hetong007 Please review whether the API changes make sense to you (i.e., `RelGraphConv` and `to_homogeneous`).
@BarclayII Please help on the review too.

python/dgl/convert.py
```diff
@@ -564,16 +564,21 @@ def to_hetero(G, ntypes, etypes, ntype_field=NTYPE, etype_field=ETYPE,
     return to_heterogeneous(G, ntypes, etypes, ntype_field=ntype_field,
                             etype_field=etype_field, metagraph=metagraph)
 
-def to_homogeneous(G, ndata=None, edata=None):
+def to_homogeneous(G, ndata=None, edata=None, store_type=True, return_count=False):
```
hetong007, line 567: 1. Is there any real use cases asking for `return_count`? 
2. What if we provide a `return_count` API for hetero graphs, other than just a param in `to_homogeneous`?

jermainewang: > 1. Is there any real use case asking for `return_count`?

It is for this PR. Feeding `return_count` to the forward function of `RelGraphConv` can avoid the cost of sorting the graph. 

> 2. What if we provide a `return_count` API for hetero graphs, other than just a param in `to_homogeneous`?

The count array is useful when the user use a homogeneous graph to represent a heterograph. It can speed up some operations. Right now it only affects `RelGraphConv`, but we will add more later.

Look like I should add some descriptions about using scenarios of this count array.

hetong007: Providing the senarios sounds good. 

Besides, even we'll consider adding a standalone `return_count` for hetero graph, it doesn't conflict with this `to_homogeneous(return_count)` : we can simply replace the internal implementation with that new API, but keep the input param as is.

python/dgl/convert.py
```diff
@@ -589,11 +594,28 @@ def to_homogeneous(G, ndata=None, edata=None):
...
+        the number of nodes/edges of the i^th type.
 
     Returns
     -------
     DGLGraph
         A homogeneous graph.
+    ntype_count : list[int]
+        Number of nodes of each type.
+    etype_count : list[int]
+        Number of edges of each type.
```
BarclayII, line 611: Does `ntype_count` and `etype_count` return if `return_count = False`?

python/dgl/convert.py
```diff
@@ -633,28 +655,38 @@ def to_homogeneous(G, ndata=None, edata=None):
...
-    eids = []
-    ntype_ids = []
     nids = []
+    eids = []
+    if store_type:
+        ntype_ids = []
+        etype_ids = []
+    if return_count:
+        ntype_count = []
+        etype_count = []
```
BarclayII, line 665: Can we not use a Python list?  We can create an `np.array` beforehand.

python/dgl/nn/pytorch/conv/relgraphconv.py
```diff
@@ -292,3 +347,12 @@ def forward(self, g, feat, etypes, norm=None):
                 node_repr = self.activation(node_repr)
             node_repr = self.dropout(node_repr)
             return node_repr
+
+def _searchsorted(sorted_sequence, values):
+    # searchsorted is introduced to PyTorch in 1.6.0
+    if getattr(th, 'searchsorted', None):
```
BarclayII, line 353: Should we cache this flag?

python/dgl/convert.py
jermainewang: The count array is usually used together with `torch.split` to compute messages one type at a time. `torch.split` requires python list: https://pytorch.org/docs/stable/generated/torch.split.html.

PR #2480, opened 2021-01-02
[Example] Add a TensorFlow example for SGC

joshcarty:
## Description
This is a TensorFlow implementation of Simple Graph Convolution.

#### Keras model interface
I've tried for a more TensorFlow-native approach by implementing it as a `keras.Model`. I recognise this is different to the other examples in `examples/tensorflow` where the training loop is not part of the model and I'm keen to get your feedback.

On one hand, I see the advantage of some consistency between the backends. On the other, it would be great to play to a framework's strengths. One key strength in leaning into TensorFlow's Keras API would be in making these models more accessible. Having to write a training loop from scratch can be a real barrier to users and I'm not sure it's always necessary.

Let me know what you think. I'm happy to change this to the existing style. Alternatively, I would love to discuss how we can best offer a Keras interface for DGL. To start, I'd happily support with migrating the existing TensorFlow examples.

#### Weight decay
Separately, I've used AdamW in TensorFlow Addons as Adam in TensorFlow doesn't support weight decay. I've read that the implementation differs to Adam in PyTorch and this results in worse performance in evaluation. I can confirm in my tests that this is the cause and that the results are similar to the PyTorch SGC when setting the loss like [this](https://github.com/dmlc/dgl/blob/master/examples/tensorflow/gcn/train.py#L99). Do you think it's worth including a custom optimizer to ensure the results are comparable to the other backends?

#### GPU
Given the simplicity of the model, I didn't show how use the model with a GPU. Do you think we should also include that?

@yzh119 I saw you reviewed #658. Would you mind taking a look at this?

Thanks!

## Checklist

## Changes
- [x] Add a TensorFlow example for SGC


yzh119 (approved):
@joshcarty thanks for your contribution!
This PR looks good to me, however the accuracy seems lower than our PyTorch's implementation. I was wondering if it's because of AdamW optimizer?

joshcarty:
Thanks @yzh119!

Yeah, if you implement weight decay like [this](https://github.com/dmlc/dgl/blob/master/examples/tensorflow/gcn/train.py#L100), you do get results that are more similar to the paper and [PyTorch](https://github.com/dmlc/dgl/tree/master/examples/pytorch/sgc) implementation:

```bash
# Cora citation network dataset
$ python sgc.py --dataset cora --lr 0.2 --n-epochs 100 --weight-decay 5e-6
...
Epoch 100/100
1/1 [==============================] - 0s 37ms/step - loss: 0.1724 - accuracy: 1.0000 - val_loss: 0.8172 - val_accuracy: 0.7900
Test Accuracy: 79.9%

# Citeseer citation network dataset
$ python sgc.py --dataset citeseer --lr 0.2 --n-epochs 150 --bias --weight-decay 5e-5
...
Epoch 150/150
1/1 [==============================] - 0s 63ms/step - loss: 0.7747 - accuracy: 0.9583 - val_loss: 1.2750 - val_accuracy: 0.7280
Test Accuracy: 72.6%

# Pubmed citation network dataset
$ python sgc.py --dataset pubmed --lr 0.2 --n-epochs 100 --bias --weight-decay 5e-5
...
Epoch 100/100
1/1 [==============================] - 0s 53ms/step - loss: 0.3104 - accuracy: 0.9833 - val_loss: 0.6409 - val_accuracy: 0.8140
Test Accuracy: 79.7%
```

| Dataset  | Accuracy | Paper | PyTorch
|----------|----------|-------|-------|
| Cora     | 79.9%    | 81.0% | 81.9% |
| Citeseer | 72.6%    | 71.9% | 72.5% | 
| Pubmed   | 79.7%    | 78.9% | 78.8% |

I don't love that this isn't handled by the optimizer but I don't think I have the knowledge to replicate how it works in PyTorch. Instead, I have to pass the decay parameter into the fitting method.

Do you think that would be preferable for more comparable results? Or do we feel comfortable just knowing that this is the cause of the difference.

yzh119:
@joshcarty thanks for the update, optimizers in PyTorch has a `weight_decay` parameter that did similar thing to https://github.com/dmlc/dgl/blob/master/examples/tensorflow/gcn/train.py#L100 , would you mind mentioning the cause of difference in the README?

PR #2487, opened 2021-01-04
[Doc] add compute_loss example in link prediction user guide

BarclayII:


docs/source/guide/minibatch-link.rst
```diff
@@ -137,6 +137,11 @@ above.
 
 .. code:: python
 
+    def compute_loss(pos_score, neg_score):
+        # an example hinge loss
+        n = pos_score.shape[0]
+        return (neg_score.view(n, -1) - pos_score.view(n, -1) + 1).clamp(min=0).mean()
```
jermainewang, line 143: You mentioned the number of negative samples may not match the number of positive samples. What about this example?

BarclayII: Each positive example may receive k negative examples.  So this code should work fine.

jermainewang: I see. Then `neg_score.view(n, -1) - pos_score.view(n, -1)` has broadcasting. This also means the edges of positive graphs and negative graphs must be aligned so that a reshape works here. Is there any guarantee on this? It looks like this is ensured because the negative sampler is as follows:
```python
class NegativeSampler(object):
    def __init__(self, g, k):
        # caches the probability distribution
        self.weights = g.in_degrees().float() ** 0.75
        self.k = k

    def __call__(self, g, eids):
        src, _ = g.find_edges(eids)
        src = src.repeat_interleave(self.k)
        dst = self.weights.multinomial(len(src), replacement=True)
        return src, dst
```
If the user does not use `src.repeat_interleave(self.k)` but other ways to generate `src, dst` pair, will the above loss function still work?

BarclayII: No.  The loss function above works only with the case where a positive example has a fixed number of negative examples.

jermainewang (approved):

PR #2491, opened 2021-01-05
[Examples] add venue label filepath; delete redundant code

Pein-hub:
## Description
the code of examples/pytorch/metapath2vec/test.py has 2 bugs

## Changes
Fix1,  add the venue label filepath, it's `file_1` that miss in code

Fix2, there are some redundant code for split the train and test data； 
         annotate the print in loop of authors which is much more print information；


BarclayII:
Thank you so much!

PR #2496, opened 2021-01-06
[Feature] Add GraphDataLoader implementation 

hetong007:
## Description
This PR adds `GraphDataLoader` for graph classification/regression tasks, and saves users' effort to repeatedly implement their own collate function to batch the graphs.

After this PR's merged, I'll update the relevant examples/tutorials with this new `GraphDataLoader`.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [X] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [X] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage
- [X] Code is well-documented
- [ ] To the my best knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR

PR #2504, opened 2021-01-07
[Doc] Update GraphDataLoader in our docs

hetong007:
## Description
This is a follow up of https://github.com/dmlc/dgl/pull/2496.
It adds `GraphDataLoader` to our tutorials and user guides en/cn.

I'm not sure if updating the examples is necessary, as users using pre-0.6 versions may experience an API break.

## Checklist
- [ ] Code is well-documented


BarclayII:
We can update the examples.  The release branches also contain the examples for that specific version so everything should be fine.

hetong007:
I've done my change on examples as well. Please kindly review that part, thanks!

python/dgl/dataloading/pytorch/__init__.py
```diff
@@ -427,7 +427,7 @@ class GraphDataLoader:
 
     Parameters
     ----------
-    collate : Function, default is None
+    collate_fn : Function, default is None
```
BarclayII, line 430: You probably don't need this argument explicitly; `DataLoader` also has an argument called `collate_fn`.

BarclayII (approved):

python/dgl/dataloading/pytorch/__init__.py
hetong007: I put it here in order to have an easier control over `self.collate`.

Issue #2528, opened 2021-01-13
When an ntype is empty, g.add_nodes() overwrites prior features instead of initializing them


noncomputable:
## 🐛 Bug

If there are no nodes or edges of a particular type, adding them with g.add_nodes() or g.add_edges() *with data* will result in the feature schema for that type being erased and overwritten.

## To Reproduce

```python
g = dgl.graph(([], []))
g.ndata["x"] = torch.tensor([])
g.ndata["y"] = torch.tensor([])
g.ndata
> {'x': tensor([]), 'y': torch.tensor([])}
g.add_nodes(1, {'y': torch.tensor([0])})
g.ndata
{'y': tensor([0])
```

Same thing for edges.

## Expected behavior

If a graph specifies features for an ntype or etype, even if there are no nodes or edges yet, new nodes and edges added to the graph should initialize those features instead of erasing them.

## Environment

 - DGL Version (e.g., 1.0): 0.5.3
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): PyTorch 1.6.0
 - How you installed DGL (`conda`, `pip`, source): conda
 - Python version: 3.7

PR #2527, opened 2021-01-12
[Bug fix] [Feature] added option for batching empty data

noncomputable:
## Description

Fix for bug https://github.com/dmlc/dgl/issues/2526. Instead of strictly changing the behavior of dgl.batch(), the fix is implemented as an optional argument, *just in case* any projects have ended up working around the assumption that empty frames get erased.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage
- [x] Code is well-documented
- [x] To the my best knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [x] Related issue is referred in this PR

PR #2529, opened 2021-01-13
[Bug fix] Stop erasing schemas of frames with no data rows

noncomputable:
## Description
Solving issue #2528. There was code explicitly causing the issue, with a comment motivating it:

> if no rows in current frame; append is equivalent to
> directly updating columns.

But this assumption is not true to the docs and doesn't justify erasing the existing features. Any defined features, even if they don't have rows yet, should be initialized for added nodes and edges.

## Checklist
- [x] Code is well-documented


BarclayII (approved):
Good catch!  And thanks for the unit tests as well!

BarclayII:
I also noticed that you have another PR #2527 which has the same code.  Should I close that one?

noncomputable:
> I also noticed that you have another PR #2527 which has the same code. Should I close that one?

No I've just made a change to that branch after a discussion with Minjie.

noncomputable:
And no problem, happy to help!

jermainewang:
@noncomputable Would you please add yourself to https://github.com/dmlc/dgl/blob/master/CONTRIBUTORS.md ?

BarclayII:
> @noncomputable Would you please add yourself to https://github.com/dmlc/dgl/blob/master/CONTRIBUTORS.md ?

As a TODO item we should scan the contributor list and add them later ourselves.

noncomputable:
> @noncomputable Would you please add yourself to https://github.com/dmlc/dgl/blob/master/CONTRIBUTORS.md ?

Yep, got it. https://github.com/dmlc/dgl/pull/2538

python/dgl/batch.py
```diff
@@ -62,6 +62,8 @@ def batch(graphs, ndata=ALL, edata=ALL, *,
         Node features to batch.
     edata : list[str], None, optional
         Edge features to batch.
+    include_empty_data : bool, optional
+        Whether to copy a frame even if no nodes or edges exist for it.
```
jermainewang, line 66: remove

PR #2550, opened 2021-01-20
[Performance] Use ALG2 for SpMM in cuSparse

nv-dlasalle:
## Description
@yzh119 identified that SPMM was running slower in CUDA11 using `cusparseSpMM()` + `CUSPARSE_SPMM_CSR_ALG1,` than using `cusparseScsrmm2()`.

This PR uses CUSPARSE_SPMM_CSR_ALG2 in SPMM when using CUDA11 instead of CUSPARSE_SPMM_CSR_ALG1. This results in a large reduction in runtime as ALG2 operates on row-major matrices, and can take advantage of the memory layout for larger numbers of features. This also avoids transposing the output.

For https://github.com/dglai/dgl-0.5-benchmark/blob/master/end-to-end/full-graph/node-classification/main_dgl_reddit_sage.py I have seen a ~4x reduction in the SpMM kernel time on a Titan V.

alg1
~1.84s for ALG1

alg2
~0.459s for ALG2

## Checklist
- [x] Code is well-documented

## Changes

The tests in `tests/compute/test_sparse.py` cover these changes.

In addition to changing which algorithm is used, the matrices B and C are stored in row-major order instead of column-major order when the operation is performed.

### DIFF
--- a/src/array/cuda/spmm.cu
+++ b/src/array/cuda/spmm.cu
@@ -122,8 +122,6 @@ void CusparseCsrmm2(
     CUSPARSE_CALL(cusparseCreate(&(thr_entry->cusparse_handle)));
   }
   CUSPARSE_CALL(cusparseSetStream(thr_entry->cusparse_handle, thr_entry->stream));
-  // allocate matrix for temporary transposed output
-  DType* trans_out = static_cast<DType*>(device->AllocWorkspace(ctx, m * n * sizeof(DType)));
   // all one data array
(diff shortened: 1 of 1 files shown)


yzh119 (approved):
LGTM, thanks @nv-dlasalle for the help.

I notice that ALG2 `May produce slightly different results during different runs with the same input parameters`, is this the behavior the same as `cusparseScsrmm2` in CUDA 10?

PR #2559, opened 2021-01-23
[Example] Add Graph Cross Net (GXN) example for pytorch backend 

lygztq:
## Description
Add Graph Cross Net (GXN) example for pytorch backend.

## Checklist
- [x] Code is well-documented

@hetong007


examples/pytorch/gxn/README.md
```diff
@@ -0,0 +1,108 @@
...
+- NumGraphs: 5000
+- AvgNodesPerGraph: 74.49
+- AvgEdgesPerGraph: 2457.78
+- NumFeats: -
+- NumClasses: 3
+
+
+How to run example files
+--------------------------------
+If you want to reproduce the author's result, in the GXN-DGL folder, run
```
hetong007, line 58: where does this `GXN-DGL` refer to?

examples/pytorch/gxn/README.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+where
+- dataset_name: Dataset name used in this experiment. Could be DD', 'PROTEINS', 'ENZYMES', 'IMDB-BINARY', 'IMDB-MULTI' and 'COLLAB'.
+- device_id: ID of computation device. -1 for pure CPU computation. For example if you only have single GPU, set this value to be 0.
+- num_trials: How many times does the experiment conducted.
+- print_training_log_every: Print training log every ? epochs. -1 for silent training.
+
+
+NOTE: If your have problem when using 'IMDB-BINARY', 'IMDB-MULTI' and 'COLLAB', it could be caused by a bug in `LegacyTUDataset`/`TUDataset` in DGL (see [here](https://github.com/dmlc/dgl/pull/2543)). Try update your DGL or
```
hetong007, line 77: update DGL to which version?

examples/pytorch/gxn/README.md
```diff
@@ -0,0 +1,108 @@
...
+**Accuracy**
+
+**NOTE**: Different from our implementation, the author uses fixed dataset split. Thus there may be difference between our result and the author's result. **To compare our implementation with the author's, we follow the setting in the author's implementation that performs model-selection on testset**. We also try early-stop with patience equals to 1/5 of the total number of epochs for some datasets. The result of `Author's Code` in the table below are obtained using first-ford data as the test dataset.
+
+|                   | DD           | PROTEINS    | ENZYMES     | IMDB-BINARY | IMDB-MULTI | COLLAB     |
+| ------------------| ------------ | ----------- | ----------- | ----------- | ---------- | ---------- |
+| Reported in Paper | 82.68(4.1 )  | 79.91(4.1)  | 57.50(6.1)  | 78.60(2.3)  | 55.20(2.5) | 78.82(1.4) |
+| Author's Code     | 82.05        | 72.07       | 58.33       | 77.00       | 56.00      | 80.40      |
+| DGL               | 82.97(3.0)   | 78.21(2.0)  | 57.50(5.5)  | 78.70(4.0)  | 52.26(2.0) | 80.58(2.4) |
+| DGL(early-stop)   | 78.66(4.3)   | 73.12(3.1)  | 39.83(7.4)  | N/A         | N/A        | N/A        |
```
hetong007, line 94: Can you specify the reason of the last three columns being `N/A` ?

examples/pytorch/gxn/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+Device: 
+- CPU: Intel(R) Xeon(R) CPU E5-2686 v4 @ 2.30GHz
+- GPU: Tesla V100-SXM2 16GB
+
+In seconds
+
+|               | DD    | PROTEINS | ENZYMES | IMDB-BINARY | IMDB-MULTI | COLLAB |
+| ------------- | ----- | -------- | ------- | ----------- | ---------- | ------ |
+| Author's Code | 25.32 | 2.93     | 1.53    | 2.42        | 3.58       | 19.78  |
```
hetong007, line 107: Is there explanation on DGL being slower on COLLAB?

examples/pytorch/gxn/networks.py
```diff
@@ -0,0 +1,305 @@
...
+        self.num_cross_layers = num_cross_layers
+
+        # build network
+        self.start_gcn_scale1 = GraphConvWithDropout(in_dim, hidden_dim)
+        self.start_gcn_scale2 = GraphConvWithDropout(hidden_dim, hidden_dim)
+        self.end_gcn = GraphConvWithDropout(2 * hidden_dim, out_dim)
+
+        self.index_select_scale1 = IndexSelect(pool_ratios[0], hidden_dim, act="prelu", dist=dist)
+        self.index_select_scale2 = IndexSelect(pool_ratios[1], hidden_dim, act="prelu", dist=dist)
+        self.start_pool_s12 = GraphPool(hidden_dim)
```
hetong007, line 55: Is it possible to use loop and lists for a more flexible implementation? In that case it is easy to control the number of pool/unpool operations.

lygztq: Actually, the author only uses models with the number of graph cross layers less than 3. Also, there is some difference between these layers. For example line 121 and line 140 in this file.

examples/pytorch/gxn/README.md
lygztq: I may need to check the author's code. As you can see, the training time on DD, which is a dataset with fewer edges (~716/graph) than COLLAB (~2500/graph), is longer than the time on COLLAB.

examples/pytorch/gxn/networks.py
hetong007: That seems to me is just adding skip connections with regard to `num_cross_layers`, and can still be generalized. But anyway, if no one's requesting a flexible version, we can keep it as is for now.

examples/pytorch/gxn/README.md
lygztq: This is because of the bug in DGL's TUDataset, I need to modify code in `main.py`. I will conduct experiments on these datasets later.

hetong007:
In general LGTM, would like to wait for the investigation on COLLAB efficiency and the early-stopping results before merging.

examples/pytorch/gxn/README.md
lygztq: It seems that CPU is faster on small batch size.

PR #2576, opened 2021-01-27
[Distributed] Distributed METIS partition

zheng-da:
## Description
This script is to convert METIS partitions into DGLGraphs.

## Checklist
- [ ] Code is well-documented


examples/pytorch/rgcn/experimental/README.md
```diff
@@ -88,6 +88,58 @@ We can get the performance score at the second epoch:
...
+More details about the four steps are explained in our
+[user guide](https://doc.dgl.ai/guide/distributed-preprocessing.html).
+
+### Step 1: write the graph into files.
+
+The graph structure should be written as a node file and an edge file. The node features and edge features
+can be written as DGL tensors. `write_mag.py` shows an example of writing the OGB MAG graph into files.
+
+```bash
+python3 write_mag.py
```
zhjwy9343, line 103: Run the command will get an error:

  File "write_mag.py", line 42, in <module>
    node_data = th.stack(node_data, 1)
RuntimeError: Expected object of scalar type long int but got scalar type bool for sequence element 1.

tools/convert_partition.py
```diff
@@ -0,0 +1,232 @@
+import os
+import json
+import time
+import argparse
+import numpy as np
+import dgl
+import torch as th
+import pyarrow
+from pyarrow import csv
```
jermainewang, line 9: Remember to add package dependencies in the README.

zhjwy9343:
I tested the rgcn/experimental codes about parMETIS. In the 2nd round, all python scripts pasted tests.

In addition, have two requirements:
1. For users who do not know what "weights" mean or do not care about the performance of different weights, but just want to run distributed partitioning, we can allow them not setting weight. If no weights, DGL partition code can have a default weights. If our users care about performance of using weights, they can set weights for nodes and edges, overwriting the default setting in DGL.
2. The convert_partition.py should be packed into DGL, otherwise the relative path used in Step 3 will fail if users use pip to install DGL.

zhjwy9343 (commented):
See my comments in general.

tools/convert_partition.py
zheng-da: in which README?

zheng-da:
For the first comment, I need to wait for George to change ParMETIS.

For the second comment, do we still have time to pack the script into DGL? @jermainewang We should also pack the launch script to DGL.

jermainewang (approved):

PR #2578, opened 2021-01-27
[Bug] Heterogeneous graph convolution bugfix

BarclayII:
## Description
`HeteroGraphConv` does not consider node features on destination node types in full graph convolution.  This is wrong for convolution modules like `SAGEConv`.

### DIFF
--- a/python/dgl/nn/pytorch/hetero.py
+++ b/python/dgl/nn/pytorch/hetero.py
@@ -169,7 +169,7 @@ class HeteroGraphConv(nn.Module):
                     continue
                 dstdata = self.mods[etype](
                     rel_graph,
-                    inputs[stype],
+                    (inputs[stype], inputs[dtype]),
                     *mod_args.get(etype, ()),
                     **mod_kwargs.get(etype, {}))
                 outputs[dtype].append(dstdata)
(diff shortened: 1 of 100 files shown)


python/dgl/nn/pytorch/hetero.py
```diff
@@ -169,7 +169,7 @@ def forward(self, g, inputs, mod_args=None, mod_kwargs=None):
                     continue
                 dstdata = self.mods[etype](
                     rel_graph,
-                    inputs[stype],
+                    (inputs[stype], inputs[dtype]),
```
sneakerkg, line 172: Test failed. If we want to add this change, all dst node feature should be there in the dict when calling forward. This might be a breaking change, we may also need to test on heterograph examples. If we don't want to break existing code, we may instead using an option, when user use graphSAGE, the error message will inform the option. If we think it's not a huge break on existing code, we can go on with this change and just fix test.

sneakerkg (changes requested):

BarclayII:
Removed the test cases where a pair of dictionaries are fed for full graph updates.  Added a heterogeneous block case for that purpose.  Please take a look at the test cases.

sneakerkg (approved):

PR #2608, opened 2021-02-02
[Performance] Fix Cython not being imported after switching to RTLD_LOCAL

BarclayII:
In #2328 I switched from `RTLD_GLOBAL` to `RTLD_LOCAL`.  This would make Cython FFI fail and fallback to ctypes since it cannot locate the DGL C symbols in the global namespace.  This PR fixes the issue by linking the Cython FFI library to `libdgl.so`.


yzh119 (approved):
LGTM.

PR #2610, opened 2021-02-02
[bugfix] Solve the boundary issue in backward function of segment sum

yzh119:
## Description
When apply `segment_sum` on segments with trailing zeros. The backward function would trigger out-of-boundary issue because we tried to access the `m`'th element of indices, which does not exist (the largest available index in `indices` is `m-1`).

The right solution would be adding a pseudo element in `indices` array and remove it after `scatter` and `cumsum`.

This PR also adds a unittest for this case.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/backend/mxnet/sparse.py
+++ b/python/dgl/backend/mxnet/sparse.py
@@ -346,10 +346,10 @@ class SegmentReduce(mx.autograd.Function):
         offsets = self.offsets
         m = offsets[-1].asscalar()
         if self.op == 'sum':
-            offsets_np = asnumpy(offsets[1:-1])
-            indices_np = np.zeros((m,), dtype=offsets_np.dtype)
+            offsets_np = asnumpy(offsets[1:])
+            indices_np = np.zeros((m + 1,), dtype=offsets_np.dtype)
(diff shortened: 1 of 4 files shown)


BarclayII (approved):
Don't understand why this will cause a segfault though since everything is in Python.  I'm good with the PR.  Could you add a comment in the changes referring to this PR?

yzh119:
@BarclayII it does not trigger segment fault. Just python-side index error.

yzh119:
There is no related issue, a user found this bug when he tried to apply `sum_nodes` on batched graphs.

IceCapriccio:
Fortunately I encounter this bug in dgl-0.5.3, and I find it was fixed in the next version.
There I will provide a simple program to reproduce this code
```python
import dgl.backend as F
data = torch.ones(6, dtype=torch.float32, requires_grad=True)
seglen = torch.tensor([1, 2, 3, 0], dtype=torch.int64)  # when trailing zeros exist in seglen, error will occur
print(F.context(seglen))
offsets = F.cumsum(F.cat([F.zeros((1,), torch.int64, 'cpu'), seglen], 0), 0)
rst = F.segment_reduce('sum', data, offsets).sum()
rst.backward()
```

Issue #2600, opened 2021-02-01
[feature] Aggregate duplicate edge weights for `to_simple`


yzh119:
## 🚀 Feature
Our `to_simple` API did not handle the case where there might be duplicate edges in the graph, DGL randomly select the edge for one of them but we should allow user to aggregate these entries (e.g. summation).

cc @smilexuhc

PR #2623, opened 2021-02-04
[Feature] Support aggregate multiple edge features in to_simple.

yzh119:
## Description
Implement #2600 .

## Checklist
- [x] Code is well-documented


BarclayII:
Random brainstorming:

A related functionality is to add two weighted adjacency matrices together.  What is our recommended way?  If we do "concat" then "to_simple", would it be too heavy both in user code and in system performance?

yzh119:
@BarclayII what do you mean by `concat` then `to_simple`?
I think adding two adjacency matrix is basically sparse-sparse matrix addition?

BarclayII:
> @BarclayII what do you mean by `concat` then `to_simple`?
> I think adding two adjacency matrix is basically sparse-sparse matrix addition?

Right, but we don't have an API for this kind of thing.

python/dgl/transform.py
```diff
@@ -1886,12 +1888,19 @@ def to_simple(g,
...
-        from one of the duplicate edges.
+        two nodes (u, v), the feature of the edge is the aggregation
+        of edge feature of duplicate edges.
 
         If False, the simple graph will not have any edge features.
 
         (Default: False)
+    aggregator: str, optional
+        Indicate how to aggregate edge feature of duplicate edges.
+        If ``random``, randomly select one of the duplicate edges' feature.
```
BarclayII, line 1899: I realized that it's not really "random".  I would use "arbitrary" instead ("Arbitrary" means "at one's discretion").

python/dgl/transform.py
```diff
@@ -1991,13 +2000,32 @@ def to_simple(g,
...
+                    data = col.data
+                    new_data = F.scatter_add(data, feat_idx, _num_rows)
+                    if aggregator == 'mean':
+                        norm = F.astype(counts[i], F.dtype(data))
+                        norm = F.reshape(norm, (F.shape(norm)[0],) + (1,) * (F.ndim(data) - 1))
+                        new_data /= norm
+                    _data[key] = new_data
+
+                newf = Frame(data=_data, num_rows=_num_rows)
+                edge_frames.append(newf)
```
BarclayII, line 2028: Can we functionize each of the branches?

python/dgl/transform.py
```diff
@@ -1991,13 +2000,32 @@ def to_simple(g,
...
+        else:  # 'sum'/'mean'
+            edge_frames = []
+            for i in range(len(g.canonical_etypes)):
+                feat_idx = edge_maps[i]
+                _, indices = np.unique(F.asnumpy(feat_idx), return_index=True)
+                _num_rows = len(indices)
+                _data = {}
+                for key, col in g._edge_frames[i]._columns.items():
+                    data = col.data
+                    new_data = F.scatter_add(data, feat_idx, _num_rows)
```
BarclayII, line 2020: How does this compare to `torch.scatter_add_`?  I'm wondering if we should replace the usage of `scatter_add_` in GSpMM backward function if the current one is more efficient.

BarclayII (approved):
I'm personally OK with extending the behavior of to_simple.  @jermainewang Do you have any more comments?

python/dgl/transform.py
yzh119: This one is different from `torch.scatter_add_`, our `scatter_add` only requires the indices on the first dimension `(n,)` but torch's `scatter_add_` requires the indices on first two dimensions: `(n,d)`.

python/dgl/transform.py
yzh119: yep

python/dgl/transform.py
yzh119: We use torch's `scatter_add_` in GSpMM backward function because the argmax/min at each feature dimension is different.

Issue #2158, opened 2020-09-08
[feature request] API to query DGL allocated memory


yzh119:
## 🚀 Feature
`dgl.memory_stats(device)` that returns the memory allocated via DGL on a given device. 

## Motivation
DGL has its own memory allocation mechanism which is seperated from backend frameworks, in PyTorch, the `torch.cuda.max_memory_cached` API cannot get accurate GPU memory usage of a model written in DGL.

Reference:
* PyTorch's memory_stats function: https://pytorch.org/docs/stable/cuda.html#torch.cuda.memory_stats

## Implementation

Maintain the size of total allocated memory in `AllocWorkspace` function and `FreeWorkspace` function: https://github.com/dmlc/dgl/blob/870da747eaee96313aa798806a22573ea1e1e8eb/src/runtime/workspace_pool.cc.

BarclayII:
This doesn't count the memory allocated by `AllocDataSpace` or `NDArray::Empty`.  I guess a better solution if possible would be using the framework-specific allocator.

jermainewang:
Closed. DGL now can use backend frameworks' allocators so it also supports querying memory usage using backend APIs.

PR #2651, opened 2021-02-11
[Performance][Memory] When cuda11.2 is detected, use cudaMallocAsync/cudaFreeAsync

nv-dlasalle:
## Description
This enables use of the new `cudaMallocAsync` and `cudaFreeAsync` functions in CUDA 11.2:
https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__HIGHLEVEL.html#group__CUDART__HIGHLEVEL_1ga31efcffc48981621feddd98d71a0feb

This accomplishes essentially what #2328 provided, but is backend agnostic, and does not require linking against external frameworks.

## Checklist
- [x] Code is well-documented

## Changes

Adds `GetStream()` to the `DeviceAPI`.

When `DeviceAPI::SetStream()` is called, the previous stream is synchronized with, to ensure no race condition is created between works scheduled before the call, and work scheduled after the call.

Makes `cudaMemcpy` invocations asynchronous, unless they are `DeviceToHost`. This is because any device-side accesses to the copied memory will be on the associated stream (default or not), and thus will wait for it to complete, while the CPU can move on to schedule more work.

Replaces calls to `cudaMalloc` and `cudaFree` with `cudaMallocAsync` and `cudaFreeAsync`, when the cuda version >= 11.2.

### DIFF
--- a/include/dgl/runtime/device_api.h
+++ b/include/dgl/runtime/device_api.h
@@ -125,6 +125,14 @@ class DeviceAPI {
    * \param stream The stream to be set.
    */
   virtual void SetStream(DGLContext ctx, DGLStreamHandle stream) {}
+  /**
+   * \brief Get the underlying stream.
+   *
+   * \param ctx The context to get the stream of.
+   *
+   * \return The stream.
+   */
+  virtual DGLStreamHandle GetStream(DGLContext ctx) = 0;
   /*!
    * \brief Synchronize 2 streams of execution.
    *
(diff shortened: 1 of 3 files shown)


yzh119:
@nv-dlasalle thanks for bringing this up, we are glad to merge this PR However #2328 solves several other issues besides performance:
1. DGL cannot reuse PyTorch's cached memory, and sometimes incurs out-of-memory issue.
2. DGL's allocated memory is not counted in PyTorch's memory stats: #2158 .

nv-dlasalle:
Thanks @yzh119. I'll open an issue for this (sorry for not doing that in the first place), and we can discuss it there to see if this change makes sense.

PR #2658, opened 2021-02-13
[hotfix] A bunch of fix for SetTransformer

### LINKED ISSUE #2657: The paddings in multihead attention model
## ❓ Questions and Help

When I use set_transformer model on graphs implemented by DGL, the model will output NaN just after one training epoch. After reading source codes for a few hours, I think I find where the mistakes happen.

 # padding to (B, max_len_x/mem, num_heads, d_head)
        queries = F.pad_packed_tensor(queries, lengths_x, 0)
        keys = F.pad_packed_tensor(keys, lengths_mem, 0)
        values = F.pad_packed_tensor(values, lengths_mem, 0)

# attention score with shape (B, num_heads, max_len_x, max_len_mem)
        e = th.einsum('bxhd,byhd->bhxy', queries, keys)
        # normalize
        e = e / np.sqrt(self.d_head)

        # generate mask
        mask = th.zeros(batch_size, max_len_x, max_len_mem).to(e.device)
        for i in range(batch_size):
            mask[i, :lengths_x[i], :lengths_mem[i]].fill_(1)
        mask = mask.unsqueeze(1)
        e.masked_fill_(mask == 0, -float('inf'))

The above are the codes that I think finally cause this problem. Since we need to pad neighbors of every node so that they have the same number of neighbors, there will be multiple rows consist of zeros only. And then we will use "-inf" to mask Q,K,V. So the rows consist of only zeros will consist of "-inf" only. When these rows are sent to softmax function later, they will generate many rows consist of "NaN". And we need to multiply several matrix, so after several epochs we will only have matrix consists of NaN. So how can I solve this problem, using a different number to pad the sequence?

yzh119:
## Description
To fix #2657 and #2660.

## Checklist
- [x] Code is well-documented


python/dgl/nn/pytorch/glob.py
```diff
@@ -852,6 +892,8 @@ class PMALayer(nn.Module):
     def __init__(self, k, d_model, num_heads, d_head, d_ff, dropouth=0., dropouta=0.):
         super(PMALayer, self).__init__()
         self.k = k
+        if k == 1:
+            dgl_warning("if k is set to 1, W_q and W_k would not be updated during training.")
```
mufeili, line 896: It's not clear what are `W_q` and `W_k` and why they will not be updated in the case of k=1. Perhaps you can include a more detailed note in the doc.

python/dgl/nn/pytorch/glob.py
```diff
@@ -785,6 +821,8 @@ class InducedSetAttentionBlock(nn.Module):
     def __init__(self, m, d_model, num_heads, d_head, d_ff, dropouth=0., dropouta=0.):
         super(InducedSetAttentionBlock, self).__init__()
         self.m = m
+        if m == 1:
+            dgl_warning("if m is set to 1, W_q and W_k would not be updated during training.")
```
mufeili, line 825: Ditto

python/dgl/nn/pytorch/glob.py
yzh119: DONE

yzh119:
@mufeili are you ok with the changes.

mufeili (approved):

PR #2669, opened 2021-02-17
Fix bug in EdgeConv implementation

### LINKED ISSUE #2668: EdgeConv message passing
https://github.com/dmlc/dgl/blob/ff64bd0de83f865a8076cf50f69525549c32ca87/python/dgl/nn/pytorch/conv/edgeconv.py#L112

According to the equation, the node h_i itself should be the input of `phi`. Shouldn't this be `edges.dst['x']` instead of `edges.src['x']` ?(I am assuming the message travels from src to dst, and thus dst is the node i which receives messages from neighbours)

BarclayII:
Fixes #2668 (Thanks @pmorerio)

### DIFF
--- a/python/dgl/nn/mxnet/conv/edgeconv.py
+++ b/python/dgl/nn/mxnet/conv/edgeconv.py
@@ -25,6 +25,12 @@ class EdgeConv(nn.Block):
     where :math:`\mathcal{N}(i)` is the neighbor of :math:`i`.
     :math:`\Theta` and :math:`\Phi` are linear layers.
 
+    .. note::
+
+       The original formulation includes a ReLU inside the maximum operator.
+       This is equivalent to first applying a maximum operator then applying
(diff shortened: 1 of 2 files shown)


pmorerio (approved):

yzh119 (approved):
LGTM. btw @BarclayII could we use builtin for EdgeConv?

python/dgl/nn/mxnet/conv/edgeconv.py
```diff
@@ -118,7 +124,7 @@ def message(self, edges):
         r"""The message computation function
         """
         theta_x = self.theta(edges.dst['x'] - edges.src['x'])
-        phi_x = self.phi(edges.src['x'])
```
yzh119, line 121: Also removing the message function in MXNet impl.

yzh119 (approved):
LGTM

PR #2673, opened 2021-02-18
[Doc] Clarify link prediction tutorial

### LINKED ISSUE #2670: Clarify link prediction docs
## 📚 Documentation

 Tutorial part

On your new tutorial about [link prediction](https://docs.dgl.ai/en/latest/new-tutorial/4_link_predict.html), you stated that it followed [SEAL model](https://papers.nips.cc/paper/2018/file/53f0d7c537d99b3824f0f99d62ea2428-Paper.pdf), but what was represented there is just simply applying score function after message passing, which is quite different from SEAL. More particular, SEAL model does not use node feature but node labeling instead, which allows it to apply to inductive setting. Node labeling also requires constructing a local subgraph for each pair of nodes beforehand, then message passing on these graphs, which is really computationally inefficient, and quite difficult to implement.

BarclayII:
In the new link prediction tutorial we said it follows the practice of SEAL.  This is not quite accurate since we only took their idea of treating it as a binary classification problem, but not how they build node features in their model.

Fixes #2670.


BarclayII:
@nguyenhungquang I rephrased it a bit.  Please let me know what you think.  Thanks!

nguyenhungquang:
Thank for your quick response, it's clearer now.

new-tutorial/4_link_predict.py
```diff
@@ -37,9 +37,7 @@
 # either citing or being cited, between two papers exists in a citation
 # network.
 #
-# This tutorial follows a relatively simple practice from
-# `SEAL <https://papers.nips.cc/paper/2018/file/53f0d7c537d99b3824f0f99d62ea2428-Paper.pdf>`__.
-# It formulates the link prediction problem as a binary classification
+# This tutorial formulates the link prediction problem as a binary
```
jermainewang, line 40: binary classification problem

jermainewang (approved):

PR #2689, opened 2021-02-22
[Example] Recommend NFS over copy files

Lyken17:
## Description
Current files are synchronized via manually copying, which is hard to use (users need to specify the path mapping) and time consuming (each change on codes requires a new round of copy and there is no increamental diff). A shared network filesystem is a better way to take the role and this has been used widely in many distributed DL tasks. 

## Checklist
- [ ] Code is well-documented


examples/pytorch/rgcn/experimental/README.md
```diff
@@ -1,62 +1,64 @@
 ## Distributed training
 
-This is an example of training RGCN node classification in a distributed fashion. Currently, the example train RGCN graphs with input node features. The current implementation follows ../rgcn/entity_claasify_mp.py.
+This is an example of training RGCN node classification in a distributed fashion. Currently, the example only support training RGCN graphs with no input features. The current implementation follows [../rgcn/entity_claasify_mp.py](../rgcn/entity_claasify_mp.py).
```
zheng-da, line 3: it supports node features now.

examples/pytorch/rgcn/experimental/README.md
```diff
@@ -1,62 +1,64 @@
...
 
-Before training, please install some python libs by pip:
+Before training, install python libs by pip:
 
 ```bash
-sudo pip3 install ogb
-sudo pip3 install pyinstrument
-sudo pip3 install pyarrow
+pip3 install ogb pyinstrument
```
zheng-da, line 8: we need to install pyarrow.

examples/pytorch/rgcn/experimental/README.md
```diff
@@ -1,62 +1,64 @@
...
+sudo apt-get install nfs-kernel-server
+``` 
 
-Users need to make sure that the master node (node-0) has right permission to ssh to all the other nodes.
+Then create a folder for sharing and (optional) remove all restrictions
+```bash
+sudo mkdir -p /NFS
+# Change ownership and remove restrictions
+sudo chown -R nobody:nogroup /NFS
+sudo chmod 755 /NFS
```
zheng-da, line 28: does this require root permission to write `/NFS`? I tried it and didn't find a way to write data to the folder with a regular account.

examples/pytorch/rgcn/experimental/README.md
```diff
@@ -1,62 +1,64 @@
...
 
-In this example, we partition the ogbn-mag graph into 4 parts with Metis. The partitions are balanced with respect to
-the number of nodes, the number of edges and the number of labelled nodes.
+The server's internal ip can be checked  via `ifconfig` or `ip`. If the ip does not begin with `192.168`, then you may use
 ```bash
-python3 partition_graph.py --dataset ogbn-mag --num_parts 4 --balance_train --balance_edges
+# for ip range 10.0.0.0 – 10.255.255.255	
+/NFS  10.0.0.0/8(rw,sync,no_subtree_check)
+# for ip range 172.16.0.0 – 172.31.255.255	
+/NFS  172.16.0.0/12(rw,sync,no_subtree_check)
```
zheng-da, line 44: after modifying /etc/exports, do we need to restart the server?

we need to provide instructions of mounting NFS from the client.

examples/pytorch/rgcn/experimental/README.md
```diff
@@ -1,62 +1,64 @@
...
-DGL provides a script for copying partitioned data to the cluster. Before that, copy the training script to a local folder:
+
+DGL provides a script for copying partitioned data to the cluster. 
+Before that, copy the training script to a local folder:
 
 
 ```bash
 mkdir ~/dgl_code
-cp ~/dgl/examples/pytorch/rgcn/experimental/entity_classify_dist.py ~/dgl_code
+cp /home/ubuntu/dgl/examples/pytorch/rgcn/experimental/entity_classify_dist.py ~/dgl_code
```
zheng-da, line 59: why do you change it to use `/home/ubuntu`?

examples/pytorch/rgcn/experimental/README.md
```diff
@@ -75,90 +77,55 @@ python ~/dgl/tools/copy_files.py --ip_config ip_config.txt \
...
+"python3 dgl_code/entity_classify_dist.py --graph-name ogbn-mag --dataset ogbn-mag --fanout='25,25' --batch-size 512  --n-hidden 64 --lr 0.01 --eval-batch-size 16  --low-mem --dropout 0.5 --use-self-loop --n-bases 2 --n-epochs 3 --layer-norm --ip-config ip_config.txt  --num-workers 4 --num-servers 1 --sparse-embedding  --sparse-lr 0.06 --node-feats"
 ```
 
-### Step 5: Verify the partition result (Optional)
-
-```bash
-python3 verify_mag_partitions.py 
+We can get the performance score at the second epoch:
+```
+Val Acc 0.4323, Test Acc 0.4255, time: 128.0379
```
zheng-da, line 128: the original README shows how to partition with ParMETIS. can you copy the instructions of using ParMETIS?

examples/pytorch/rgcn/experimental/README.md
Lyken17: Updated

examples/pytorch/rgcn/experimental/README.md
Lyken17: Usually NFS setup is performed by an admin account, so I assume that the user has sudo access. 

It is possible to use a common directory, says a folder under user's home, as a NFS point (without sudo access). However, in this case, they will not be able to edit `/etc/exports` and `/etc/fstab` to enable NFS.

examples/pytorch/rgcn/experimental/README.md
Lyken17: it is a typo. Fixed

examples/pytorch/rgcn/experimental/README.md
Lyken17: Added

examples/pytorch/rgcn/experimental/README.md
zheng-da: NFS setup should be done by admin. but when using NFS (e.g., writing data to NFS), we need to allow a normal user account. otherwise, it'll be very inconvenient.

we should show an example of using a folder under user's home directory. NFS setup can be done by admin. It's fine.

examples/pytorch/rgcn/experimental/README.md
```diff
@@ -75,15 +111,41 @@ python ~/dgl/tools/copy_files.py --ip_config ip_config.txt \
...
+```
+
 ### Step 3: Launch distributed jobs
 
 DGL provides a script to launch the training job in the cluster. `part_config` and `ip_config`
 specify relative paths to the path of the workspace.
 
-The command below launches one training process on each machine and each training process has 4 sampling processes.
-**Note**: There is a known bug in Python 3.8. The training process hangs when running multiple sampling processes for each training process.
-Please set the number of sampling processes to 0 if you are using Python 3.8.
```
zheng-da, line 85: we have to notify users of how to set up the sampling processes correctly. we'll fix the bug in the next release.

zheng-da (approved):

PR #2710, opened 2021-03-01
[Performance] Linear UniformChoice optimization

pawelpiotrowicz:
UniformChoice function is a significant part of libdgl usage in GraphSage. 

UniformChoiceMaster

The overhead of std::unordered_set can be replaced by linear search implementation - is more cheaper for CPU, besides it removes final std::copy to desired output.
In real for the small range of random set such (1 < n < 64) we can expect 4 times performance gain or more for entire random output. An internal benchmark showed that up to 128 output numbers it make sense to use linear implementation but for sure I set limit up to 64. In fact I see that GraphSage uses either 10 or 25 as desired random output so there is quite good speedup here.

Below is gain for 3 epochs.

UniformChoiceDiff

Moreover total overhead of libdgl.so has been reduced from **4.3% to 3.9%** .


src/random/cpu/choice.cc
```diff
@@ -26,73 +26,90 @@ IdxType RandomEngine::Choice(FloatArray prob) {
...
+        auto b = out + 1;
+        auto e = b + num - 1;
+        while (b != e) {
+          *b = RandInt(population);
+          auto it = std::find(out, b, *b);
+          if (it != b) continue;
+          ++b;
+        }
+
+      } else {
```
yzh119, line 81: Looks good to me, would you minding mention this PR in the comments?

yzh119 (approved):

src/random/cpu/choice.cc
pawelpiotrowicz: @yzh119 done

yzh119: perfect

Issue #2514, opened 2021-01-11
Putting the graph on GPU will arise AssertionError: all the graphs must be on CPU


YankeeTsao:
When I trained unsupervised graphsage model, too much time was wasted on I/O between CPU and GPU. Therefore, I put the whole graph on GPU.

```
def run(proc_id, n_gpus, args, devices, data):
    device = devices[proc_id]
    in_feats, g = data
    g = g.to(device)
```

And got the following error：
File "dgl/transform.py", line 1526, in compact_graphs
assert all(g.device == F.cpu() for g in graphs), 'all the graphs must be on CPU'

mufeili:
If you use the example DGL code for unsupervised GraphSage, it performs mini-batch training and you don't need to put the whole graph on GPU.

YankeeTsao:
> If you use the example DGL code for unsupervised GraphSage, it performs mini-batch training and you don't need to put the whole graph on GPU.

It cost too much time on I/O between CPU and GPU. I tried to tune the parameter  'num_workers' of DataLoader, but got another error: shm-size is sufficient(Running in docker). That's why I want to put the whole graph on GPU.

mufeili:
@BarclayII Have you profiled this example before? Any suggestions?

VoVAllen:
@YankeeTsao You can manually increase the shared memory size inside docker by `--shm-size 4G` to increase it to 4G

YankeeTsao:
> @YankeeTsao You can manually increase the shared memory size inside docker by `--shm-size 4G` to increase it to 4G

Exactly! However, I don't have permission to modify the docker. Is there any other chance to decrease the time wasted on I/O between CPU and GPU? Thanks for help!

BarclayII:
The tricky thing is that currently `EdgeDataLoader` will require the graph to reside on CPU, since it uses the function `dgl.compact_graphs` and `dgl.to_block`, both of which currently do not have GPU implementation.

One possible workaround is to put the features on GPU while putting the graph topology itself on CPU, so you don't have to spend time copying node features from CPU to GPU.  That is, instead of calling `g.to(device)`, add a `.cuda()` call to the following statement:
https://github.com/dmlc/dgl/blob/5d4995e6e8381ecc3c6b9b3e6e3a8eb9d0bf72c8/examples/pytorch/graphsage/train_sampling_unsupervised.py#L179
while keeping `g` unchanged.

YankeeTsao:
[...]

PR #2716, opened 2021-03-02
[Performance][GPU] Enable GPU uniform edge sampling

nv-dlasalle:
## Description
This PR implements only uniform edge sampling on a single GPU (`CSRRowWiseSamplingUniform()`).

For graphsage with sampling on ogbn-products, this takes the per-epoch time on a V100 from 5.75 seconds with 4 worker processes performing sampling on the CPU, to 1.69 seconds with no workers when performing sampling on the GPU.

## Checklist
- [X] Code is well-documented

## Changes

This implements `CSRRowWiseSamplingUniform` for the GPU.

It also adds a `_ScalarDataLoader` which returns a tensor of node IDs, rather than list, which significantly reduces overhead in the dataloader. This is used only when the number of worker processes is zero, as it does not implement multiple workers.

Overhead of converting python arrays to DGL NDArrays was reduced by caching fanout and probability information in `MultiLayerNeighborSampler`, rather than converting the arrays at each iteration.

The sample `/examples/pytorch/graphsage/train_sampling.py` was modified to support sampling via the GPU.

Finally, the sampling unit test was split, such that uniform edge sampling could be tested on the GPU, while other sampling would only be tested on the CPU.


examples/pytorch/graphsage/train_sampling.py
```diff
@@ -234,11 +248,6 @@ def run(args, device, data):
         train_nfeat = train_nfeat.to(device)
         train_labels = train_labels.to(device)
 
-    # Create csr/coo/csc formats before launching training processes with multi-gpu.
-    # This avoids creating certain formats in each sub-process, which saves momory and CPU.
-    train_g.create_formats_()
-    val_g.create_formats_()
-    test_g.create_formats_()
```
nv-dlasalle, line 241: Since this is a single-gpu example, creating these formats before-hand is unnecessary.

nv-dlasalle:
This should fix #2514

examples/pytorch/graphsage/train_sampling.py
yzh119: Exactly.

examples/pytorch/graphsage/train_sampling.py
```diff
@@ -214,6 +225,8 @@ def run(args, device, data):
 
     if args.dataset == 'reddit':
         g, n_classes = load_reddit()
+    elif args.dataset == 'ogb-products':
```
zheng-da, line 228: can we use its original name 'ogbn-products'?

nv-dlasalle: Changed to 'ogbn-products' in fc37138

nv-dlasalle:
@zheng-da I'm stumped as to why this is failing the CI. The unit tests that run prior to `tests/distributed` appear to be running in the same amount of time as successful runs, and this shouldn't impact anything in the `tests/distributed` set of unit tests, particularly since that is only run with a CPU. It passes on my local machine fine (in 390.72s on my local machine vs 373.43s on a successful CI run). Given that the CI kills it after 15 minutes, it certainly appears to be hanging in that test for some reason (https://github.com/dmlc/dgl/blob/master/tests/distributed/test_new_kvstore.py#L298).

src/array/cuda/rowwise_sampling.cu
```diff
@@ -0,0 +1,389 @@
...
+      }
+    }
+
+    out_row += gridDim.x*BLOCK_ROWS;
+  }
+}
+
+/**
+* @brief Perform row-wise sampling on a CSR matrix, and generate a COO matrix,
+* without replacement.
```
isratnisa, line 191: I think you meant with replacement.

isratnisa (approved):

examples/pytorch/graphsage/train_sampling.py
```diff
@@ -9,7 +9,9 @@
 import argparse
 import tqdm
 
-from load_graph import load_reddit, inductive_split
+from torch.cuda import nvtx
```
zheng-da, line 12: i think it's better to remove nvtx in the example code.

python/dgl/dataloading/pytorch/__init__.py
```diff
@@ -242,9 +297,25 @@ def __init__(self, g, nids, block_sampler, device='cpu', **kwargs):
...
+                self.dataloader = DataLoader(
+                    self.collator.dataset,
+                    collate_fn=self.collator.collate,
+                    **dataloader_kwargs)
+            else:
+                if isinstance(self.collator.dataset, th.Tensor):
+                    self.dataloader = _ScalarDataLoader(
+                        self.collator.dataset,
+                        collate_fn=self.collator.collate,
+                        **dataloader_kwargs)
```
zheng-da, line 313: there might be compatibility issues. not all pytorch DataLoader's arguments are supported here.

python/dgl/dataloading/pytorch/__init__.py
```diff
@@ -242,9 +297,25 @@ def __init__(self, g, nids, block_sampler, device='cpu', **kwargs):
...
+            if "num_workers" in dataloader_kwargs and dataloader_kwargs["num_workers"] > 0:
+                assert g.device == th.device('cpu'), \
+                    'Only graphs on the cpu are supported for NodeDataLoader ' \
+                    'in the case of num_workers > 0.'
+                self.dataloader = DataLoader(
+                    self.collator.dataset,
+                    collate_fn=self.collator.collate,
+                    **dataloader_kwargs)
+            else:
+                if isinstance(self.collator.dataset, th.Tensor):
```
zheng-da, line 309: if this is more efficient, how about we convert `dataset` to `th.Tensor`. so users don't see weird performance difference.

python/dgl/sampling/neighbor.py
```diff
@@ -129,18 +128,24 @@ def sample_neighbors(g, nodes, fanout, edge_dir='in', prob=None, replace=False,
         else:
             nodes_all_types.append(nd.array([], ctx=nd.cpu()))
 
-    if not isinstance(fanout, dict):
-        fanout_array = [int(fanout)] * len(g.etypes)
+    if isinstance(fanout, nd.NDArray):
```
zheng-da, line 131: why is `fanout` `nd.NDArray`?

python/dgl/sampling/neighbor.py
```diff
@@ -129,18 +128,24 @@ def sample_neighbors(g, nodes, fanout, edge_dir='in', prob=None, replace=False,
...
-            raise DGLError('Fan-out must be specified for each edge type '
-                           'if a dict is provided.')
-        fanout_array = [None] * len(g.etypes)
-        for etype, value in fanout.items():
-            fanout_array[g.get_etype_id(etype)] = value
-    fanout_array = F.to_dgl_nd(F.tensor(fanout_array, dtype=F.int64))
-
-    if prob is None:
+        if not isinstance(fanout, dict):
+            fanout_array = [int(fanout)] * len(g.etypes)
```
jermainewang, line 135: same issue of `None` type `fanout` argument as above

jermainewang (changes requested):
Sorry for the late review. Comments on python codes.

python/dgl/sampling/neighbor.py
nv-dlasalle: `_CAPI_DGLSampleNeighbors()` takes an `nd.NDArray` as the argument for fanout. Creating it each step is expensive relative to sampling, so I added the option in `MultiLayerNeighborSampler()` to construct it once, and keep passing it in here each step.

python/dgl/sampling/neighbor.py
nv-dlasalle: The Doc String says:
```
fanout : int or dict[etype, int]
```

It looks like calling code should convert `None`s to `-1`.
```
If -1 is given for a single edge type, all the neighboring edges with that edge
        type will be selected.
```

python/dgl/dataloading/neighbor.py
```diff
@@ -63,6 +65,10 @@ def __init__(self, fanouts, replace=False, return_eids=False):
         self.fanouts = fanouts
         self.replace = replace
 
+        # used to cache computations and memory allocations
+        self.fanout_arrays = []
```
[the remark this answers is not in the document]
nv-dlasalle: Updated in 
51bf37e

examples/pytorch/graphsage/train_sampling.py
```diff
@@ -122,13 +124,22 @@ def run(args, device, data):
     val_nid = th.nonzero(val_g.ndata['val_mask'], as_tuple=True)[0]
     test_nid = th.nonzero(~(test_g.ndata['train_mask'] | test_g.ndata['val_mask']), as_tuple=True)[0]
 
+    dataloader_device = th.device('cpu')
+    if args.sample_gpu:
+        train_nid = train_nid.to(device)
+        # copy only the csr to the GPU
```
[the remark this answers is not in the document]
nv-dlasalle: Fixed in 
a89a0db

examples/pytorch/graphsage/train_sampling.py
```diff
@@ -149,7 +160,9 @@ def run(args, device, data):
         # Loop over the dataloader to sample the computation dependency graph as a list of
         # blocks.
         tic_step = time.time()
+        nvtx.range_push("dataloader")
```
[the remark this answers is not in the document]
nv-dlasalle: Sorry -- removed in 
8ffcbbf

examples/pytorch/graphsage/train_sampling.py
nv-dlasalle: Sorry -- removed in 
8ffcbbf

python/dgl/dataloading/pytorch/__init__.py
nv-dlasalle: I changed it to the pytorch prescribed method of implementing an IterableDataset, and using that to control the batches in 
89d8fce

src/array/cuda/rowwise_sampling.cu
nv-dlasalle: Fixed in 
a89a0db

python/dgl/sampling/neighbor.py
jermainewang: I was confused by the docstring. Fan-out value of each edge type cannot be None, but the whole `fanout` argument could be None. I think we should improve the docstring later. No more question on this.

jermainewang (approved):
I'm good except the `ScalarDataLoader`. I think the optimization should belong to PyTorch (plus the implementation is not complete either). Otherwise, LGTM.

python/dgl/dataloading/pytorch/__init__.py
```diff
@@ -1,10 +1,65 @@
...
+        return batch
+
+class _ScalarDataLoader:
+    """Custom DataLoader to return mini-batches as tensors, rather than as
+    lists. When used inside of the NodeDataLoader, this significantly reduces
+    the overhead. For the case of a batch size of 1024, instead of giving a
+    list of 1024 tensors to the collator, a single tensor of 1024 dimensions
+    is passed in.
+    This implementation supports only minimum set of features.
+    """
```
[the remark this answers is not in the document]
nv-dlasalle: I've switched it to be an IterableDataset wrapper that manually does the batching (https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset) in 
89d8fce, instead of using a custom dataloader.

This is quite necessary, as it's about a 3x speedup as compared to using the default path:
dataloader_issue

zheng-da (approved):

python/dgl/dataloading/pytorch/__init__.py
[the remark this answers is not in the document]
nv-dlasalle: @jermainewang Talking with some of the PyTorch folks, one workaround for this is in 0a04381, which drops the cost of converting to a tensor on the GPU from ~16.5ms to ~2.5ms. However, using the IterableDataset as I currently am, avoids the cost of conversion entirely, and is ~25% faster end-to-end (bs=1024, hidden=256, fanout=30,30).

python/dgl/dataloading/pytorch/__init__.py
nv-dlasalle: In 
d7c8619 if it's a DGL tensor, it will get converted to PyTorch tensor.

PR #2730, opened 2021-03-08
[Example][Bug Fix] Improve DiffPool

lygztq:
## Description

* Fix bug in Diffpool: the wrong `assign_dim` param
* Improve efficiency of DiffPool, make the model independent of batch size. Remove redundant computation.

Details:

**efficiency**

On V100-SXM2 16GB

| Experiments       | Train time per epoch(s) | Memory usage(MB) |
| ----------------- | -----------------------: | ----------------: |
| Original-DD-bs=20 | OOM                     | OOM              |
| Original-DD-bs=10 | 21.302                  | 5274.620         |
| Original-ENZYMES  | 1.749                   | 25.685           |
| Improved-DD-bs=20 | 44.682                  | 10088.889        |
| Improved-DD-bs=10 | 17.282                  | 2928.568         |
| Improved-ENZYMES  | 1.685                   | 21.909           |

The accuracy is not affected by this improvement. I will list accuracy results later

Each experiment with improved model is only conducted once, thus the result may has noise.

| Experiments      | Acc    |
| ---------------- | ------: |
| Original-DD      | 79.31% |
| Original-ENZYMES | 63.33% |
| Improved-DD      |  training...      |
| Improved-ENZYMES | 68.33% |

## Checklist
- [ ] Code is well-documented

@hetong007

### DIFF
--- a/examples/pytorch/diffpool/README.md
+++ b/examples/pytorch/diffpool/README.md
@@ -37,22 +37,23 @@ DD 79.31% (with early stopping)
 
 On V100-SXM2 16GB
 
-| Experiments       | Train time per epoch(s) | Memory usage(MB) |
-| ----------------- | ----------------------: | ---------------: |
-| Original-DD-bs=20 | OOM                     | OOM              |
-| Original-DD-bs=10 | 21.302                  | 5274.620         |
-| Original-ENZYMES  | 1.749                   | 25.685           |
-| Improved-DD-bs=20 | 44.682                  | 10088.889        |
-| Improved-DD-bs=10 | 17.282                  | 2928.568         |
-| Improved-ENZYMES  | 1.685                   | 21.909           |
+|                    | Train time/epoch (original) (s) | Train time/epoch (improved) (s) |
+| ------------------ | ------------------------------: | ------------------------------: |
+| DD (batch_size=10) |                          21.302 |                      **17.282** |
+| DD (batch_size=20) |                             OOM |                      **44.682** |
+| ENZYMES            |                           1.749 |                       **1.685** |
+
+|                    | Memory usage (original) (MB) | Memory usage (improved) (MB) |
+| ------------------ | ---------------------------: | ---------------------------: |
+| DD (batch_size=10) |                     5274.620 |                 **2928.568** |
+| DD (batch_size=20) |                          OOM |                **10088.889** |
+| ENZYMES            |                       25.685 |                   **21.909** |
 
 **Accuracy**
 
 Each experiment with improved model is only conducted once, thus the result may has noise.
 
-| Experiments      | Acc    |
-| ---------------- | -----: |
-| Original-DD      | 79.31% |
-| Original-ENZYMES | 63.33% |
-| Improved-DD      | 78.33% |
-| Improved-ENZYMES | 68.33% |
+|         |   Original |   Improved |
+| ------- | ---------: | ---------: |
+| DD      | **79.31%** |     78.33% |
+| ENZYMES |     63.33% | **68.33%** |
--- a/examples/pytorch/diffpool/README.md
+++ b/examples/pytorch/diffpool/README.md
@@ -16,7 +16,7 @@ How to run
 ----------
 
 ```bash
-python train.py --dataset ENZYMES --pool_ratio 0.10 --num_pool 1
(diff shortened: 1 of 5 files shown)


examples/pytorch/diffpool/README.md
```diff
@@ -16,8 +16,8 @@ How to run
 ----------
 
 ```bash
-python train.py --dataset ENZYMES --pool_ratio 0.10 --num_pool 1
-python train.py --dataset DD --pool_ratio 0.15 --num_pool 1  
+python train.py --dataset ENZYMES --pool_ratio 0.10 --num_pool 1 --epochs 1000
```
hetong007, line 19: Can we merge the commands into a single run?

examples/pytorch/diffpool/model/dgl_layers/gnn.py
```diff
@@ -101,27 +102,13 @@ def __init__(self, input_dim, assign_dim, output_feat_dim,
...
-            block_diag(
-                *
-                assign_tensor_masks)).to(
-            device=device)
-        assign_tensor = masked_softmax(assign_tensor, mask,
-                                       memory_efficient=False)
+        assign_tensor = self.pool_gc(g, h)  # size = (sum_N, N_a), N_a is num of nodes in pooled graph.
+        assign_tensor = F.softmax(assign_tensor, dim=1)
+        assign_tensor = torch.split(assign_tensor, g.batch_num_nodes().tolist())
+        assign_tensor = torch.block_diag(*assign_tensor)  # size = (sum_N, batch_size * N_a)
```
hetong007, line 110: Seems `torch.block_diag` returns a dense tensor.

Another strategy is to split `assign_tensor` and `feat`, loop the `matmul`, and finally concatenate back.

Is this strategy much slower than the call in L112? If the speed is comparable I'd say we avoid the creation of a large block tensor.

hetong007 (commented):
Also, your claim as `The accuracy is not affected by this improvement` is incorrect. This improvement reduces model parameters in `pool_gc`, so the `pool_gc` layer sees more data samples thus should be trained better.

examples/pytorch/diffpool/README.md
```diff
@@ -16,8 +16,8 @@ How to run
...
+python train.py --dataset DD --pool_ratio 0.15 --num_pool 1  --batch-size 10
 ```
 Performance
 -----------
```
hetong007, line 23: Please include a short description of the improvement of both memory and accuracy, and add statistics in README as supportive evidence.

Also, if applicable, please also include acc/memory from the original implementation.

yzh119:
#2730 might further help accelerate diffpool.

examples/pytorch/diffpool/README.md
lygztq: Why? These two commands are for two different datasets.

examples/pytorch/diffpool/model/dgl_layers/gnn.py
lygztq: It is worth trying this method on node features. However, this method is expensive for pooling edges because we can only get one big matrix from graphs in dgl (no matter dense or sparse).

examples/pytorch/diffpool/README.md
hetong007: Oh I see, never mind.

examples/pytorch/diffpool/model/dgl_layers/gnn.py
hetong007: OK that makes sense, so that means the computation on L114 - L115 are needing this diagonal.

hetong007:
@yzh119 Did you want to mention something else in this PR, or mention this PR in somewhere else?

examples/pytorch/diffpool/README.md
```diff
@@ -25,5 +25,34 @@ ENZYMES 63.33% (with early stopping)
...
-## Dependencies
+## Update (2021-03-09)
 
+**Changes:**
+
+* Fix bug in Diffpool: the wrong `assign_dim` parameter
+* Improve efficiency of DiffPool, make the model independent of batch size. Remove redundant computation.
+
+
+**Efficiency:**
```
hetong007, line 36: Do we observe speed difference?

hetong007 (approved):

PR #2733, opened 2021-03-09
[Distributed] Distributed node embedding and sparse optimizer

classicsong:
## Description
This is a draft right now

## Checklist
- [x] Code is well-documented


python/dgl/distributed/__init__.py
```diff
@@ -18,7 +18,7 @@
 from .dist_tensor import DistTensor
 from .partition import partition_graph, load_partition, load_partition_book
 from .graph_partition_book import GraphPartitionBook, PartitionPolicy
-from .sparse_emb import SparseAdagrad, DistEmbedding
```
zheng-da, line 21: do you completely hide DistEmbedding?

python/dgl/distributed/kvstore.py
```diff
@@ -1119,6 +1119,25 @@ def get_data_meta(self, name):
         part_policy = self._part_policy[name]
         return (data_type, data_shape, part_policy)
 
+    def get_partid(self, name, id_tensor):
```
zheng-da, line 1122: do we need this API? we can get the same information from the graph partition book.

python/dgl/distributed/sparse_emb.py
```diff
@@ -3,9 +3,11 @@
 from .. import backend as F
 from .. import utils
 from .dist_tensor import DistTensor
+import torch.distributed as dist
+import torch as th
 
-class DistEmbedding:
-    '''Distributed embeddings.
+class NodeEmbedding:
```
zheng-da, line 9: For compatibility, we should still expose DistEmbedding. We can deprecate the name and remove it in the next release.

python/dgl/distributed/sparse_emb.py
```diff
@@ -58,148 +60,101 @@ class DistEmbedding:
...
-    the forward computation, users have to invoke py:meth:`~dgl.distributed.SparseAdagrad.step`
+    When a ``NodeEmbedding``  object is used when the deep learning framework is recording
+    the forward computation, users have to invoke py:meth:`~dgl.distributed.optim.SparseAdagrad.step`
     afterwards. Otherwise, there will be some memory leak.
     '''
     def __init__(self, num_embeddings, embedding_dim, name=None,
                  init_func=None, part_policy=None):
         self._tensor = DistTensor((num_embeddings, embedding_dim), F.float32, name,
-                                  init_func, part_policy)
+                                  init_func=init_func)
```
zheng-da, line 70: we don't need to provide partition policy? how does the distributed tensor know which partition policy to use?

python/dgl/distributed/sparse_emb.py
```diff
@@ -58,148 +60,101 @@ class DistEmbedding:
...
         self._trace = []
-
-    def __call__(self, idx):
+        self._name = name
+        self._num_embeddings = num_embeddings
+        self._embedding_dim = embedding_dim
+
+        # Check whether it is multi-gpu training or not.
```
zheng-da, line 76: it's not just multi-gpu training. it's any distributed training.

python/dgl/distributed/sparse_emb.py
```diff
@@ -58,148 +60,101 @@ class DistEmbedding:
...
+        if th.distributed.is_initialized():
+            self._rank = th.distributed.get_rank()
+            self._world_size = th.distributed.get_world_size()
+        else:
+            assert 'th.distributed shoud be initialized'
+        self._optm_state = None # track optimizer state
+        self._part_policy = part_policy
+
+    def __call__(self, idx, device=th.device('cpu')):
```
zheng-da, line 85: let's not provide `device` here for now. users can move data explicitly.

python/dgl/distributed/sparse_emb.py
```diff
@@ -58,148 +60,101 @@ class DistEmbedding:
...
+        """Return rank of current process.
 
-        This function runs on the KVStore server. It updates the gradients by scaling them
-        according to the state sum.
+        Returns
+        -------
+        int
+            The rank of current process.
+        """
+        return self._rank
```
zheng-da, line 111: this rank is from pytorch distributed. it's confusing because DistGraph exposes the rank of DGL distributed.

python/dgl/distributed/optim/pytorch/sparse_optim.py
```diff
@@ -0,0 +1,221 @@
...
+        self._shared_cache = {}
+        self._clean_grad = False
+        self._opt_meta = {}
+
+        for emb in params:
+            assert isinstance(emb, NodeEmbedding), \
+                'DGL DistSparseOptimizer only supports dgl.distributed.NodeEmbedding'
+
+            if self._rank is None:
+                self._rank = emb.rank
```
zheng-da, line 37: why do you not get the rank from Pytorch directly?

python/dgl/distributed/optim/pytorch/sparse_optim.py
```diff
@@ -0,0 +1,221 @@
...
+                self._world_size = emb.world_size
+            else:
+                assert self._rank == emb.rank, \
+                    'MultiGPU rank for each embedding should be same.'
```
zheng-da, line 41: "MultiGPU" => "Distributed"

python/dgl/distributed/optim/pytorch/sparse_optim.py
```diff
@@ -0,0 +1,221 @@
...
+                assert self._world_size == emb.world_size, \
+                    'MultiGPU world_size for each embedding should be same.'
```
zheng-da, line 43: the same here.

python/dgl/distributed/optim/pytorch/sparse_optim.py
```diff
@@ -0,0 +1,221 @@
...
+
+                # will send grad to each corresponding trainer
+                if self._world_size > 1:
+                    # get idx split from kvstore
+                    idx_split = kvstore.get_partid(name, idics)
+                    idx_split_size = []
+                    idics_list = []
+                    grad_list = []
+                    # split idx and grad first
+                    for i in range(kvstore.num_servers):
```
zheng-da, line 75: I wonder if this is a good way to split the gradients. When we get more machines, this for loop may become slow.
KVStore actually tried to optimize this kind of data distribution.

python/dgl/distributed/optim/pytorch/sparse_optim.py
```diff
@@ -0,0 +1,221 @@
...
+                    # Note: If we have GPU nccl support, we can use all_to_all to sync information here
+                    gather_list = list(th.empty([self._world_size], dtype=th.int64).chunk(self._world_size))
+                    alltoall_cpu(self._rank, self._world_size, gather_list, idx_split_size)
+                    # use cpu until we have GPU alltoallv
+                    idx_gather_list = [th.empty((int(num_emb),), dtype=idics.dtype) for num_emb in gather_list]
+                    alltoallv_cpu(self._rank, self._world_size, idx_gather_list, idics_list, idics.dtype)
+                    local_indics[name] = idx_gather_list
+                    grad_gather_list = [th.empty((int(num_emb), grads.shape[1]), dtype=grads.dtype) for num_emb in gather_list]
+                    alltoallv_cpu(self._rank, self._world_size, grad_gather_list, grad_list, idics.dtype)
+                    local_grads[name] = grad_gather_list
```
zheng-da, line 102: ideally, the entire thing should be done in C. but let's see how much overhead it has first.

zheng-da:
it seems you remove the async gradient updates. I think we should still keep async gradient updates.

python/dgl/distributed/__init__.py
classicsong: It is now renamed to NodeEmbedding

python/dgl/distributed/kvstore.py
classicsong: Which API should I use?

python/dgl/distributed/sparse_emb.py
classicsong: You mean we keep the entire DistEmbedding implementation?

python/dgl/distributed/sparse_emb.py
```diff
@@ -58,148 +60,101 @@ class DistEmbedding:
...
+                                  init_func=init_func)
         self._trace = []
-
-    def __call__(self, idx):
+        self._name = name
+        self._num_embeddings = num_embeddings
+        self._embedding_dim = embedding_dim
+
+        # Check whether it is multi-gpu training or not.
+        if th.distributed.is_initialized():
```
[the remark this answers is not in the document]
classicsong: Right now.  We focusing on PyTorch.
Maybe I need to move it into some where else.

python/dgl/distributed/sparse_emb.py
classicsong: The optimizer may be confused if user moves data around. Let's discuss later.

python/dgl/distributed/optim/pytorch/sparse_optim.py
classicsong: As everything is in GPU, it should be fast.
But, we can design some new API to replace this for loop.

classicsong:
> it seems you remove the async gradient updates. I think we should still keep async gradient updates.

Async update is another story. You mean we need to provide async AdaOptimizer?

zheng-da:
> > it seems you remove the async gradient updates. I think we should still keep async gradient updates.
> 
> Async update is another story. You mean we need to provide async AdaOptimizer?

we should support both sync and async update.

examples/pytorch/graphsage/experimental/train_dist_unsupervised_transductive.py
```diff
@@ -0,0 +1,328 @@
...
+    loss_fcn = CrossEntropyLoss()
+    loss_fcn = loss_fcn.to(device)
+    optimizer = optim.Adam(model.parameters(), lr=args.lr)
+    sparse_optimizer = SparseAdagrad([emb_layer.emb], lr=args.lr)
+
+    # Training loop
+    epoch = 0
+    #profiler = Profiler()
+    #if g.rank() == 0:
+    #    profiler.start()
```
zheng-da, line 159: maybe remove the commented code.

python/dgl/distributed/kvstore.py
zheng-da: i see. you want to have a global place for mapping IDs to part IDs. GraphPartitionBook requires to know if it's mapping node IDs or edge IDs.

python/dgl/distributed/optim/pytorch/sparse_optim.py
zheng-da: are there any profiling results on this piece of code?

python/dgl/distributed/optim/pytorch/sparse_optim.py
```diff
@@ -0,0 +1,233 @@
...
+    ----------
+    shape : tuple of ints
+        The shape of the state tensor
+    dtype : torch dtype
+        The data type of the state tensor
+    """
+    arr = th.zeros(shape, dtype=dtype)
+    return arr
+
+class SparseAdagrad(DistSparseGradOptimizer):
```
zheng-da, line 161: this is synchronized update. we should keep the async update so that we can make a comparison between them.

python/dgl/distributed/sparse_emb.py
```diff
@@ -58,148 +59,147 @@ class DistEmbedding:
...
+        Tensor
+            The requested node embeddings
+        """
         idx = utils.toindex(idx).tousertensor()
-        emb = self._tensor[idx]
+        emb = self._tensor[idx].to(device, non_blocking=True)
         if F.is_recording():
             emb = F.attach_grad(emb)
-            self._trace.append((idx, emb))
+            self._trace.append((idx.to(device, non_blocking=True), emb))
```
zheng-da, line 101: what was the decision for async read?

python/dgl/distributed/sparse_emb.py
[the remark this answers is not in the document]
zheng-da: I see you get rank and world size in the optimizer. We don't need to get them here any more.

python/dgl/distributed/sparse_emb.py
```diff
@@ -58,148 +59,147 @@ class DistEmbedding:
...
-def _init_state(shape, dtype):
-    return F.zeros(shape, dtype, F.cpu())
+        Returns
+        -------
+        int
+            The world size of the pytorch distributed training env.
+        """
+        return self._world_size
 
-class SparseAdagrad:
```
zheng-da, line 126: can we keep this async optimizer? we should make a performance comparison between the sync version and async version.

python/dgl/distributed/sparse_emb.py
classicsong: The idx is not used right after this call. Only emb is used immediately. Thus  we can use async to here.

examples/pytorch/graphsage/experimental/train_dist_unsupervised_transductive.py
classicsong: Sure

python/dgl/distributed/sparse_emb/pytorch/sparse_emb.py
```diff
@@ -0,0 +1,208 @@
...
+
+        Returns
+        -------
+        PartitionPolicy
+            partition policy
+        """
+        return self._part_policy
+
+    @property
+    def rank(self):
```
zheng-da, line 121: i don't understand why we should keep this field in Embedding? is it used in any place? I think it'll confuse users because this rank is different from the one in DistGraph.

python/dgl/distributed/sparse_emb.py
zheng-da: this is still very tricky when used with the async sampling pipeline. We cannot perform any GPU operations in the sampling thread. otherwise, it'll interfere with GPU operations in the training thread. I guess we can only perform this operation in the training thread.

python/dgl/distributed/sparse_emb/pytorch/sparse_emb.py
classicsong: They are not used. I will delete them.

python/dgl/distributed/sparse_emb.py
classicsong: There will be some other issue.
The gradient of emb will be moved to CPU by default in pytorch.

examples/pytorch/graphsage/experimental/train_dist_transductive.py
```diff
@@ -0,0 +1,355 @@
...
+        batch_labels = load_subtensor(self.g, seeds, input_nodes, "cpu")
+        blocks[-1].dstdata['labels'] = batch_labels
+        return blocks
+
+class TransDistSAGE(DistSAGE):
+    def __init__(self, in_feats, n_hidden, n_classes, n_layers,
+                 activation, dropout):
+        super(TransDistSAGE, self).__init__(in_feats, n_hidden, n_classes, n_layers, activation, dropout)
+
+    def inference(self, standalone, g, x, batch_size, device):
```
zheng-da, line 66: is this function different for inductive and transductive models?

classicsong: transductive do not need to copy node features.

examples/pytorch/graphsage/experimental/train_dist.py
```diff
@@ -51,8 +52,9 @@ def sample_blocks(self, seeds):
 
         input_nodes = blocks[0].srcdata[dgl.NID]
         seeds = blocks[-1].dstdata[dgl.NID]
-        batch_inputs, batch_labels = load_subtensor(self.g, seeds, input_nodes, "cpu")
-        blocks[0].srcdata['features'] = batch_inputs
+        batch_inputs, batch_labels = load_subtensor(self.g, seeds, input_nodes, "cpu", self.load_feat)
+        if self.load_feat:
+            blocks[0].srcdata['features'] = batch_inputs
```
zheng-da, line 57: if we don't load features, what data are used in message passing in the input layer?

examples/pytorch/graphsage/experimental/train_dist_transductive.py
```diff
@@ -0,0 +1,319 @@
...
+
+    def forward(self, idx):
+        # embeddings are stored in cpu
+        idx = idx.cpu()
+        if self.dgl_sparse_emb:
+            return self.sparse_emb(idx, device=self.dev_id)
+        else:
+            return self.sparse_emb(idx).to(self.dev_id)
+
+def load_embs(standalone, emb_layer, g):
```
zheng-da, line 107: this is to convert a distributed embedding to a distributed tensor?

examples/pytorch/graphsage/experimental/train_dist_transductive.py
```diff
@@ -0,0 +1,319 @@
...
+    inputs : The features of all the nodes.
+    labels : The labels of all the nodes.
+    val_nid : the node Ids for validation.
+    batch_size : Number of nodes to compute at the same time.
+    device : The GPU device to evaluate on.
+    """
+    model.eval()
+    emb_layer.eval()
+    with th.no_grad():
+        inputs = load_embs(standalone, emb_layer, g)
```
zheng-da, line 143: can we just call `emb_layer.emb_tensor` for `NodeEmbedding` or something like that for Pytorch embedding?

examples/pytorch/graphsage/experimental/train_dist_transductive.py
```diff
@@ -0,0 +1,319 @@
...
+    return arr
+
+class DistEmb(nn.Module):
+    def __init__(self, num_nodes, emb_size, dgl_sparse_emb=False, dev_id='cpu'):
+        super().__init__()
+        self.dev_id = dev_id
+        self.emb_size = emb_size
+        self.dgl_sparse_emb = dgl_sparse_emb
+        if dgl_sparse_emb:
+            self.sparse_emb = NodeEmbedding(num_nodes, emb_size, name='sage', init_func=initializer)
```
zheng-da, line 94: if `dgl_sparse_emb` is enabled, the NN module has no model weights from the perspective of pytorch. is this OK?

python/dgl/distributed/nn/pytorch/sparse_emb.py
```diff
@@ -0,0 +1,194 @@
...
+
+        Returns
+        -------
+        tuple of torch.Tensor
+            The optimizer related state.
+        """
+        return self._optm_state
+
+    @property
+    def weight(self):
```
zheng-da, line 186: i think the non-distributed version of NodeEmbedding has a field called `emb_tensor`. do you want to be consistent?

python/dgl/distributed/nn/pytorch/sparse_emb.py
```diff
@@ -0,0 +1,194 @@
...
+        self._num_embeddings = num_embeddings
+        self._embedding_dim = embedding_dim
+
+        # Check whether it is multi-gpu/distributed training or not
+        if th.distributed.is_initialized():
+            self._rank = th.distributed.get_rank()
+            self._world_size = th.distributed.get_world_size()
+        else:
+            assert 'th.distributed shoud be initialized'
+        self._optm_state = None # track optimizer state
```
zheng-da, line 82: it's interesting that you choose to keep track of the optimizer state in NodeEmbedding, instead of hiding it inside the optimizer. Is it necessary to keep track of it in the embedding?

tests/distributed/test_dist_graph_store.py
```diff
@@ -148,6 +147,67 @@ def run_client(graph_name, part_id, server_count, num_clients, num_nodes, num_ed
...
+        if num_clients == 1:
+            assert np.all(F.asnumpy(grad_sum[nids]) == np.ones((len(nids), 1)) * num_clients)
+        assert np.all(F.asnumpy(grad_sum[rest]) == np.zeros((len(rest), 1)))
+
+        emb = NodeEmbedding(g.number_of_nodes(), 1, 'emb2', emb_init)
+        with F.no_grad():
+            feats1 = emb(nids)
+        assert np.all(F.asnumpy(feats1) == 0)
+
+        optimizer = SparseAdagrad([emb], lr=lr)
```
zheng-da, line 192: this is to test two-embedding case?

zheng-da:
can you change the document of `DistEmbedding` and indicate that it is going to be deprecated?
I think you should add `NodeEmbedding` and optimizer to doc build system so that their doc will show up in the doc website.

zheng-da (approved):
everything looks good to me.

examples/pytorch/graphsage/experimental/train_dist.py
classicsong: If we don't load features, nodes will be transductive nodes.

examples/pytorch/graphsage/experimental/train_dist_transductive.py
classicsong: Yes.

examples/pytorch/graphsage/experimental/train_dist_transductive.py
classicsong: I think it is better not to access emb_layer.emb_tensor directly.

examples/pytorch/graphsage/experimental/train_dist_transductive.py
classicsong: Yeah. We have special logic to deal with it.

python/dgl/distributed/nn/pytorch/sparse_emb.py
classicsong: non-distributed version now have emb_tensor and weight and we will deprecate emb_tensor later.

python/dgl/distributed/nn/pytorch/sparse_emb.py
classicsong: Good point. Actually, we also store the status in the optimizer. We do not need it here.

PR #2737, opened 2021-03-10
[README] Update CUDA 11.1 instructions

BarclayII:


jermainewang (approved):
Maybe change them in a way that we don't need to update them everytime in the future release?

BarclayII:
This would only happen if new CUDA release comes out so I think it's OK.  Besides I think using wildcards can still confuse people if they couldn't get it working.

Issue #2603, opened 2021-02-01
New GNN model: Boosted Graph Neural Network (BGNN)


nd7141:
## 🚀 Feature
A proposal of a new GNN model coming from ICLR 2021: https://openreview.net/forum?id=ebS5NUfoMKL
The model is implemented with DGL: https://github.com/nd7141/bgnn

## Motivation

This is the first GNN designed to work well on graphs when node features are heterogeneous. Heterogeneous means that each feature has some individual meaning. For example, in a social network each person can have age, income, gender, graduation date, etc. as features. On the other hand, previous GNNs perform well when the node features are homogeneous. For example, node features are pretrained word embeddings or bag-of-words features. 

## Pitch

What I proposed in the paper is not a new GNN layer, but a whole model, rather a combination of GNN and GBDT models. Ideally, the end user will do something like:

```
from dgl.models import BGNN

bgnn = BGNN(*params)
bgnn.fit(graph, node_features, target_labels)
```

## Additional context

The model currently works for node classification and regression tasks. 

The model is implemented with DGL: https://github.com/nd7141/bgnn
dgl_cu92==0.5.3 is tested. 

Importantly, BGNN relies on installed GBDT package. I tested it with CatBoost package (https://catboost.ai/). I feel like it's easier to install CatBoost across all OS platforms rather than LightGBM or XGBoost (those are possible options too, but I haven't tested them). So CatBoost can be part of optional dependencies for users who want to use BGNN.

jermainewang:
Hi Sergei, awesome work and really appreciate the support! Overall, three ways to add a new model to DGL.
* Add it to `dgl.nn` as one of the built-in NN modules. It is common for work that propose new GNN layers and mainly aims at providing building blocks for new architecture.
* Add it to be one of the examples under the `examples` folder. The examples are more diverse than the NN modules. They are python scripts that train models end-to-end. They aims at providing users a base for trying out different ideas on existing work or benchmarking the model on different datasets. As a result, we recommend to follow the practice [here](https://github.com/zhjwy9343/MVP4ModelExample) to make them more consistent.
* Add it to a model zoo such as the one in [DGL-LifeSci](https://github.com/awslabs/dgl-lifesci). They usually include a set of pretrained parameters and could produce an accurate prediction using one line of code.

[...]

PR #2740, opened 2021-03-11
[Model] Add BGNN example

nd7141:
## Description
A proposal of a new GNN model coming from ICLR 2021: https://openreview.net/forum?id=ebS5NUfoMKL
The model is implemented with DGL: https://github.com/nd7141/bgnn

Reviewers: @jermainewang @BarclayII 

Discussion thread: https://github.com/dmlc/dgl/issues/2603

This is initial proposal of the model. There are two files: `BGNN.py` which implements BGNN model and `run.py` which shows how to use it. 

I also include datasets I used in the paper for the sake of reproducibility (although I think that datasets should be included in DGL as a separate request). 

Current results match those of the papers (although I have not done thorough hyperparameter tuning). 

Here are images showing performance of BGNN, that is similar to the paper (https://openreview.net/forum?id=ebS5NUfoMKL): 
image
image
image
image
image
image

I think `BGNN.py` should be part of the DGL, in the sense that an end user can just do `import dgl.BGNNPredictor`. On the other hand, `run.py` can be kept in the example. 

A few notes: 
* Currently `BGNN.py` includes as well plotting, which may not be needed for the final version.
* I evaluate for train, valid, and test masks during training, which can slow down training. These parts could be excluded, but should be discussed as it impacts early stopping. 
* BGNN relies on CatBoost package (https://catboost.ai/), so it should be installed if an end user uses BGNN. 
* There are some packages which could be installed as well but are not necessary (e.g. tqdm, sklearn). 

I am happy to answer your questions and discuss how to integrate BGNN in DGL in the optimal way. Please let me know your thoughts.


BarclayII:
Thanks a lot!  It's a brand new model contribution so I put an item in the model index README.

Could you also add a README with instructions on how to download the dataset (once you have uploaded them)?

nd7141:
@BarclayII @jermainewang I updated the code in the example and provided instructions how you can download datasets. Please let me know if something else is needed from me?

examples/pytorch/bgnn/BGNN.py
```diff
@@ -0,0 +1,356 @@
...
+                 gnn_model,
+                 task = 'regression',
+                 loss_fn = None,
+                 trees_per_epoch = 10,
+                 backprop_per_epoch = 10,
+                 lr=0.01,
+                 append_gbdt_pred = True,
+                 train_input_features = False,
+                 gbdt_depth=6,
+                 gbdt_lr=0.1):
```
jermainewang, line 25: Could you pls add docstring for these? We follow the [numpy doc style](https://numpydoc.readthedocs.io/en/latest/format.html). [Code example in DGL](https://github.com/dmlc/dgl/blob/master/python/dgl/nn/pytorch/conv/graphconv.py#L141). It's good to have some code examples for usage.

examples/pytorch/bgnn/BGNN.py
```diff
@@ -0,0 +1,356 @@
...
+                catboost_loss_fn = 'MultiClass'
+            else: # we predict the gradients for each class at epochs > 0
+                catboost_model_obj = CatBoostRegressor
+                catboost_loss_fn = 'MultiRMSE'
+
+        return catboost_model_obj(iterations=num_epochs,
+                                  depth=self.gbdt_depth,
+                                  learning_rate=self.gbdt_lr,
+                                  loss_function=catboost_loss_fn,
+                                  random_seed=0,
```
jermainewang, line 57: expose `random_seed` to the init function of `BGNNPredictor`?

examples/pytorch/bgnn/BGNN.py
```diff
@@ -0,0 +1,356 @@
...
+
+        if self.task == 'regression':
+            self.out_dim = y.shape[1]
+        elif self.task == 'classification':
+            self.out_dim = len(set(y.iloc[test_mask, 0]))
+        self.in_dim = self.out_dim + X.shape[1] if self.append_gbdt_pred else self.out_dim
+
+        gbdt_X_train = X.iloc[train_mask]
+        gbdt_y_train = y.iloc[train_mask]
+        gbdt_alpha = 1
```
jermainewang, line 268: Is this a tunable hyperparameter? Consider expose it to the `__init__` function with default value.

examples/pytorch/bgnn/BGNN.py
```diff
@@ -0,0 +1,356 @@
...
+            train_mask, val_mask, test_mask,
+            cat_features = None,
+            num_epochs=100,
+            patience=10,
+            logging_epochs=1,
+            metric_name='loss',
+            normalize_features=False,
+            replace_na=True,
+            ):
+        '''
```
jermainewang, line 230: Same. Pls follow the numpy docstring style.  DGL also has some conventions for docing the argument types:
* Use `dgl.DGLGraph` for DGLGraph.
* Use `numpy.array` for numpy arrays and describe the shape in the description.
* Use `list[int]` for list of integers
* Use python type names for python objects (e.g., `int`, `float`, ...)

examples/pytorch/bgnn/BGNN.py
```diff
@@ -0,0 +1,356 @@
...
+
+        :param graph: dgl graph
+        :param X: numpy matrix of input features. Can be composed of categorical features with missing values.
```
jermainewang, line 233: describe how missing values are represented in the matrix

examples/pytorch/bgnn/BGNN.py
```diff
@@ -0,0 +1,356 @@
...
+        :param y: numpy matrix of target labels. Can be multidimensional targets.
+        :param train_mask: list of integers (rows in X that correspond to train set)
```
jermainewang, line 235: describe the element of the mask array (e.g., 0-1 array?)

examples/pytorch/bgnn/BGNN.py
```diff
@@ -0,0 +1,356 @@
...
+        :param val_mask: list of integers (rows in X that correspond to valid set)
+        :param test_mask: list of integers (rows in X that correspond to test set)
+        :param cat_features: list of integers (columns in X that correspond to categorical features)
+        :param num_epochs: number of epochs to run
+        :param patience: number of epochs to wait until early stopping
+        :param logging_epochs: log every n epoch
+        :param metric_name: metric to use for early stopping
```
jermainewang, line 242: I wonder how different early stopping metric performs in practice? Is it critical? Can we remove the option and use the default value `'loss'`?

examples/pytorch/bgnn/BGNN.py
```diff
@@ -0,0 +1,356 @@
...
+        :param normalize_features: normalize original input features X (column wise)
+        :param replace_na: replace NA in X
```
jermainewang, line 244: I wonder if these options (`normalize_features` and `replace_na`) should be here. They look like data preprocessing steps that may go to the example code.

jermainewang (changes requested):

examples/pytorch/bgnn/BGNN.py
nd7141: Updated.

examples/pytorch/bgnn/BGNN.py
nd7141: I think we better keep as it is, as we want to look at accuracy and not at the loss, if accuracy if our final goal.

examples/pytorch/bgnn/BGNN.py
nd7141: Moved them to run.py for preprocessing.

nd7141:
Hey @jermainewang 

I updated the code documentation and also moved the preprocessing to run.py (that requires adding one additional parameter for fit function, as GBDT still want to operate on non-preprocessed features). That worked on my side. I hope it works for you too.

jermainewang (approved):
LGTM. Thanks for the contribution!

jermainewang:
@BarclayII do you have other comments?

BarclayII:
I'm good.  Thanks!

PR #2742, opened 2021-03-12
[BugFix] Return a list of ntypes/etypes

chwan-rice:
## Description

The comment of function `load_partition` shows that two return variables `ntype` and `etype` are lists. After running the code, I notice that the actual types are dictionaries. I guess returning lists is preferred, otherwise the comment should be fixed.

Note that there is no test that is related to this change.

## Checklist
- [x] Code is well-documented


zheng-da:
It seems there is some inconsistency between the comments and the code. Thanks for fixing the problem.
By checking how `load_partition` and `load_partition_book` are used, I think it's better to make the returned results by these two functions consistent. Please check out how these two functions are used [here](https://github.com/dmlc/dgl/blob/master/python/dgl/distributed/dist_graph.py#L302-L311).
I think our option is either returning dicts in both functions and changing the docstring of `load_partition`, or returning lists and changing the docstring of `load_partition_book`.

zheng-da (approved):

PR #2758, opened 2021-03-17
[Bugfix] Wrap cub with CUB_NS_PREFIX and remove dependency on Thrust to linking issues with Torch 1.8

nv-dlasalle:
## Description
This wraps all cub usages with CUB_NS_PREFIX, and removes the one use of thrust. This prevents symbol collision when linking against PyTorch 1.8.

## Checklist
- [x] Code is well-documented

## Changes
This adds `/src/array/cuda/dgl_cub.cuh`, which should be included in place of `cub/cub.cuh`, and it wraps it in the appropriate namespace.

This also changes the implementation of NonZero to use [cub::DeviceSelect::If](https://nvlabs.github.io/cub/structcub_1_1_device_select.html#abec3c5afa3f1ca08e7b4fe3fd8990c39) instead of thrust.

### DIFF
--- a/src/runtime/cuda/cuda_hashtable.cu
+++ b/src/runtime/cuda/cuda_hashtable.cu
@@ -4,11 +4,11 @@
  * \brief Device level functions for within cuda kernels.
  */
 
-#include <cub/cub.cuh>
 #include <cassert>
 
 #include "cuda_hashtable.cuh"
 #include "../../kernel/cuda/atomic.cuh"
+#include "../../array/cuda/dgl_cub.cuh"
 
 using namespace dgl::kernel::cuda;
 
--- a/src/array/cuda/array_cumsum.cu
+++ b/src/array/cuda/array_cumsum.cu
@@ -4,9 +4,9 @@
  * \brief Array cumsum GPU implementation
  */
 #include <dgl/array.h>
-#include <cub/cub.cuh>
 #include "../../runtime/cuda/cuda_common.h"
 #include "./utils.h"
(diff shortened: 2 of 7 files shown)


src/runtime/cuda/cuda_hashtable.cu
```diff
@@ -4,11 +4,11 @@
  * \brief Device level functions for within cuda kernels.
  */
 
-#include <cub/cub.cuh>
 #include <cassert>
 
 #include "cuda_hashtable.cuh"
 #include "../../kernel/cuda/atomic.cuh"
```
yzh119, line 10: The `src/kernel` folder would be deprecated, please include `src/array/cuda/atomic.cuh` instead.

BarclayII:
So should we avoid using thrust altogether in DGL later on?

Also a side question: if I write a PyTorch CUDA extension that uses thrust (or linking/dlopening two libraries both using thrust), would it also have symbol duplication?

nv-dlasalle:
@BarclayII I think until https://github.com/NVIDIA/thrust/issues/1401 is resolved, it will be simplest to avoid using it in DGL.

yzh119:
@BarclayII I think this PR fix the low-accuracy issue of dgl + cu112.

src/runtime/cuda/cuda_hashtable.cu
jermainewang: I will remove the legacy codes and fix this include later.

jermainewang (approved):
LGTM. Thanks for the fix.

PR #2767, opened 2021-03-19
[Feature] Add kd-tree implementation (CPU) for kNN

lygztq:
## Description
Add kd-tree implementation (CPU) for kNN. The kd-tree implementation works well on 3D point cloud data. However, experiments show that kd-tree has poor performance compared with vector-sorting implementation when the feature dimension goes higher.

The kd-tree implementation is from [nanoflann](https://github.com/jlblancoc/nanoflann)

@hetong007 

## Checklist
- [x] Code is well-documented

## Experiment:

**Script**

```python
from networkx.classes.function import non_neighbors
import torch as th
from dgl.transform import knn_graph
import matplotlib.pyplot as plt
from time import time

from sklearn.neighbors import NearestNeighbors

def generate_point_cloud(num_points, num_features, device=th.device("cpu")):
    return th.randn((num_points, num_features), device=device)

num_point_list = [10, 100, 1000, 10000]
num_features_list = [3, 30, 300]
num_k_list = [2, 4, 8, 16]
torch_time_list = []
dgl_time_list = []
sklearn_time_list = []

device = th.device("cpu")
for n_p in num_point_list:
    for n_f in num_features_list:
        for k in num_k_list:
            if k > n_p: continue
            x = generate_point_cloud(n_p, n_f)
            s_time = time()
            _ = knn_graph(x, k)
            torch_time_list.append(time() - s_time)
            s_time = time()
            _ = knn_graph(x, k, algorithm='kd-tree')
            dgl_time_list.append(time() - s_time)
            x = x.numpy()
            s_time = time()
            nbrs = NearestNeighbors(n_neighbors=k, algorithm="auto").fit(x)
            _, _ = nbrs.kneighbors(x)
            sklearn_time_list.append(time() - s_time)
            print("P = {:5d}, F = {:4d}, k = {:2d}, dgl_time: {:.4f}, torch_time: {:.4f}, sklearn_time: {:.4f}".format(
                n_p, n_f, k, dgl_time_list[-1], torch_time_list[-1], sklearn_time_list[-1]))
```

**Result:**
```
P: num of points, F: num of features, results are presented in seconds, sklearn_time(auto) means let sklearn choose which implementation to use (kd-tree, ball-tree or brute-force)
------------------------------------
P =    10, F =    3, k =  2, kdtree_time(openmp): 0.0010, kdtree_time: 0.0010, torch_time: 0.0029, sklearn_time: 0.0012, sklearn_time(auto): 0.0012
P =    10, F =    3, k =  4, kdtree_time(openmp): 0.0007, kdtree_time: 0.0008, torch_time: 0.0013, sklearn_time: 0.0009, sklearn_time(auto): 0.0009
P =    10, F =    3, k =  8, kdtree_time(openmp): 0.0007, kdtree_time: 0.0008, torch_time: 0.0013, sklearn_time: 0.0009, sklearn_time(auto): 0.0008
P =    10, F =   30, k =  2, kdtree_time(openmp): 0.0008, kdtree_time: 0.0008, torch_time: 0.0013, sklearn_time: 0.0009, sklearn_time(auto): 0.0006
P =    10, F =   30, k =  4, kdtree_time(openmp): 0.0009, kdtree_time: 0.0009, torch_time: 0.0013, sklearn_time: 0.0010, sklearn_time(auto): 0.0006
P =    10, F =   30, k =  8, kdtree_time(openmp): 0.0007, kdtree_time: 0.0008, torch_time: 0.0013, sklearn_time: 0.0009, sklearn_time(auto): 0.0006
P =    10, F =  300, k =  2, kdtree_time(openmp): 0.0008, kdtree_time: 0.0008, torch_time: 0.0013, sklearn_time: 0.0010, sklearn_time(auto): 0.0007
P =    10, F =  300, k =  4, kdtree_time(openmp): 0.0008, kdtree_time: 0.0009, torch_time: 0.0012, sklearn_time: 0.0011, sklearn_time(auto): 0.0007
P =    10, F =  300, k =  8, kdtree_time(openmp): 0.0008, kdtree_time: 0.0008, torch_time: 0.0012, sklearn_time: 0.0010, sklearn_time(auto): 0.0007
P =   100, F =    3, k =  2, kdtree_time(openmp): 0.0007, kdtree_time: 0.0008, torch_time: 0.0018, sklearn_time: 0.0010, sklearn_time(auto): 0.0010
P =   100, F =    3, k =  4, kdtree_time(openmp): 0.0008, kdtree_time: 0.0009, torch_time: 0.0016, sklearn_time: 0.0010, sklearn_time(auto): 0.0011
P =   100, F =    3, k =  8, kdtree_time(openmp): 0.0008, kdtree_time: 0.0009, torch_time: 0.0016, sklearn_time: 0.0010, sklearn_time(auto): 0.0010
P =   100, F =    3, k = 16, kdtree_time(openmp): 0.0008, kdtree_time: 0.0010, torch_time: 0.0016, sklearn_time: 0.0011, sklearn_time(auto): 0.0011
P =   100, F =   30, k =  2, kdtree_time(openmp): 0.0009, kdtree_time: 0.0012, torch_time: 0.0016, sklearn_time: 0.0014, sklearn_time(auto): 0.0023
P =   100, F =   30, k =  4, kdtree_time(openmp): 0.0009, kdtree_time: 0.0012, torch_time: 0.0016, sklearn_time: 0.0014, sklearn_time(auto): 0.0027
P =   100, F =   30, k =  8, kdtree_time(openmp): 0.0009, kdtree_time: 0.0012, torch_time: 0.0016, sklearn_time: 0.0015, sklearn_time(auto): 0.0027
P =   100, F =   30, k = 16, kdtree_time(openmp): 0.0009, kdtree_time: 0.0013, torch_time: 0.0017, sklearn_time: 0.0016, sklearn_time(auto): 0.0028
P =   100, F =  300, k =  2, kdtree_time(openmp): 0.0017, kdtree_time: 0.0035, torch_time: 0.0018, sklearn_time: 0.0061, sklearn_time(auto): 0.0027
P =   100, F =  300, k =  4, kdtree_time(openmp): 0.0017, kdtree_time: 0.0035, torch_time: 0.0017, sklearn_time: 0.0060, sklearn_time(auto): 0.0047
P =   100, F =  300, k =  8, kdtree_time(openmp): 0.0017, kdtree_time: 0.0036, torch_time: 0.0018, sklearn_time: 0.0061, sklearn_time(auto): 0.0033
P =   100, F =  300, k = 16, kdtree_time(openmp): 0.0020, kdtree_time: 0.0036, torch_time: 0.0018, sklearn_time: 0.0063, sklearn_time(auto): 0.0037
P =  1000, F =    3, k =  2, kdtree_time(openmp): 0.0017, kdtree_time: 0.0015, torch_time: 0.0079, sklearn_time: 0.0028, sklearn_time(auto): 0.0057
P =  1000, F =    3, k =  4, kdtree_time(openmp): 0.0015, kdtree_time: 0.0019, torch_time: 0.0076, sklearn_time: 0.0033, sklearn_time(auto): 0.0036
P =  1000, F =    3, k =  8, kdtree_time(openmp): 0.0023, kdtree_time: 0.0027, torch_time: 0.0081, sklearn_time: 0.0045, sklearn_time(auto): 0.0046
P =  1000, F =    3, k = 16, kdtree_time(openmp): 0.0032, kdtree_time: 0.0040, torch_time: 0.0095, sklearn_time: 0.0063, sklearn_time(auto): 0.0060
P =  1000, F =   30, k =  2, kdtree_time(openmp): 0.0171, kdtree_time: 0.0348, torch_time: 0.0070, sklearn_time: 0.0627, sklearn_time(auto): 0.0356
P =  1000, F =   30, k =  4, kdtree_time(openmp): 0.0100, kdtree_time: 0.0351, torch_time: 0.0067, sklearn_time: 0.0630, sklearn_time(auto): 0.0394
P =  1000, F =   30, k =  8, kdtree_time(openmp): 0.0101, kdtree_time: 0.0362, torch_time: 0.0069, sklearn_time: 0.0637, sklearn_time(auto): 0.0276
P =  1000, F =   30, k = 16, kdtree_time(openmp): 0.0107, kdtree_time: 0.0375, torch_time: 0.0103, sklearn_time: 0.0657, sklearn_time(auto): 0.0419
P =  1000, F =  300, k =  2, kdtree_time(openmp): 0.0687, kdtree_time: 0.2648, torch_time: 0.0096, sklearn_time: 0.6036, sklearn_time(auto): 0.0367
P =  1000, F =  300, k =  4, kdtree_time(openmp): 0.0687, kdtree_time: 0.2655, torch_time: 0.0095, sklearn_time: 0.6034, sklearn_time(auto): 0.0391
P =  1000, F =  300, k =  8, kdtree_time(openmp): 0.0693, kdtree_time: 0.2657, torch_time: 0.0095, sklearn_time: 0.6075, sklearn_time(auto): 0.0382
P =  1000, F =  300, k = 16, kdtree_time(openmp): 0.0696, kdtree_time: 0.2669, torch_time: 0.0121, sklearn_time: 0.6059, sklearn_time(auto): 0.0376
P = 10000, F =    3, k =  2, kdtree_time(openmp): 0.0054, kdtree_time: 0.0104, torch_time: 0.3677, sklearn_time: 0.0299, sklearn_time(auto): 0.0302
P = 10000, F =    3, k =  4, kdtree_time(openmp): 0.0069, kdtree_time: 0.0154, torch_time: 0.3695, sklearn_time: 0.0383, sklearn_time(auto): 0.0379
P = 10000, F =    3, k =  8, kdtree_time(openmp): 0.0098, kdtree_time: 0.0231, torch_time: 0.3727, sklearn_time: 0.0510, sklearn_time(auto): 0.0512
P = 10000, F =    3, k = 16, kdtree_time(openmp): 0.0157, kdtree_time: 0.0377, torch_time: 0.3858, sklearn_time: 0.0951, sklearn_time(auto): 0.0702
P = 10000, F =   30, k =  2, kdtree_time(openmp): 1.3666, kdtree_time: 5.3891, torch_time: 0.3869, sklearn_time: 6.8717, sklearn_time(auto): 2.1261
P = 10000, F =   30, k =  4, kdtree_time(openmp): 1.3784, kdtree_time: 5.4362, torch_time: 0.3791, sklearn_time: 6.8522, sklearn_time(auto): 2.3850
P = 10000, F =   30, k =  8, kdtree_time(openmp): 1.3761, kdtree_time: 5.4214, torch_time: 0.3854, sklearn_time: 6.7284, sklearn_time(auto): 2.7572
P = 10000, F =   30, k = 16, kdtree_time(openmp): 1.3817, kdtree_time: 5.4239, torch_time: 0.3988, sklearn_time: 6.5370, sklearn_time(auto): 2.5295
P = 10000, F =  300, k =  2, kdtree_time(openmp): 7.1159, kdtree_time: 29.8202, torch_time: 0.5762, sklearn_time: 54.1735, sklearn_time(auto): 2.4005
P = 10000, F =  300, k =  4, kdtree_time(openmp): 7.1116, kdtree_time: 28.1571, torch_time: 0.5664, sklearn_time: 54.5195, sklearn_time(auto): 2.8144
P = 10000, F =  300, k =  8, kdtree_time(openmp): 7.1247, kdtree_time: 28.1930, torch_time: 0.5681, sklearn_time: 55.2666, sklearn_time(auto): 2.6391
P = 10000, F =  300, k = 16, kdtree_time(openmp): 7.1283, kdtree_time: 28.1779, torch_time: 0.5779, sklearn_time: 53.4221, sklearn_time(auto): 2.9298
```


src/graph/transform/kdtree_ndarray_adaptor.h
```diff
@@ -0,0 +1,115 @@
...
+  ~KDTreeNDArrayAdaptor() {
+    delete m_index;
+  }
+
+  index_type * GetIndex() {
+    return m_index;
+  }
+
+  /*!
+   * \brief Query for the \a num_closest points to a given point
```
hetong007, line 57: What is `\a` here?

src/graph/transform/kdtree_ndarray_adaptor.h
```diff
@@ -0,0 +1,115 @@
...
+   *  Note that this is a short-cut method for GetIndex()->findNeighbors().
+   * \note nChecks_IGNORED is ignored but kept for compatibility with the
+   *  original FLANN interface.
+   */
+  void query(const FloatType * query_pt, const size_t num_closest,
+             IdType * out_idxs, FloatType * out_dists, const int nChecks_IGNORED=10) const {
```
hetong007, line 63: I suggest to remove `nChecks_IGNORED`. A not being used argument sometimes can be confusing.

src/graph/transform/knn.cc
```diff
@@ -0,0 +1,117 @@
...
+namespace transform {
+namespace impl {
+
+/*! \brief The kd-tree implementation of K-Nearest Neighbors */
+template <typename FloatType, typename IdType>
+void KdTreeKNN(const NDArray data_points, const IdArray data_offsets,
+               const NDArray query_points, const IdArray query_offsets,
+               const int k, IdArray result) {
+  int64_t batch_size = data_offsets->shape[0] - 1;
+  int64_t feature_size = data_points->shape[1];
```
hetong007, line 26: Can we replace by `IdType`?

src/graph/transform/knn.cc
```diff
@@ -0,0 +1,117 @@
...
+        result_data[curr_out_offset] = q + q_offset; curr_out_offset++;
+        result_data[curr_out_offset] = out_buffer[i] + d_offset; curr_out_offset++;
+      }
+    }
+  }
+}
+}  // namespace impl
+
+template <DLDeviceType XPU, typename FloatType, typename IdType>
+void KNN(const NDArray data_points, const IdArray data_offsets,
```
hetong007, line 66: I suppose this will be the entry point for every KNN algorithm that we are going to support right?

python/dgl/transform.py
```diff
@@ -208,22 +222,102 @@ def segmented_knn_graph(x, k, segs):
...
+    points in the same segment in :attr:`x`.
+
+    This function allows multiple point sets with different capacity. The points
+    from different sets are stored contiguously in the :attr:`x` and :attr:`y` tensor.
+    :attr:`x_segs` and :attr:`y_segs` specifies the number of points in each point set.
+
+    Parameters
+    ----------
+    x : Tensor
+        The point coordinates in x. It can be either on CPU or GPU (must be the
```
hetong007, line 258: Do we support GPU now? If not then we shouldn't claim like this in the doc. Also, we need to add corresponding checks below.

hetong007 (changes requested):

python/dgl/transform.py
```diff
@@ -59,7 +59,7 @@ def pairwise_squared_distance(x):
     return x2s + F.swapaxes(x2s, -1, -2) - 2 * x @ F.swapaxes(x, -1, -2)
 
 #pylint: disable=invalid-name
-def knn_graph(x, k):
+def knn_graph(x, k, use_dgl_impl=False):
```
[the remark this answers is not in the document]
BarclayII: Is nanoflann's KD-tree KNN an *approximate* KNN method?

If so, I agree with Tong's suggestion to rename this (and other similar places') argument.

If not (meaning that nanoflann's implementation is *exact*), I would suggest benchmarking both algorithms and see if C++ is universally more efficient than Python, and simply replace the implementation if this is the case.

Also, did you compare with scikit-learn's [nearest neighbor transformers](https://scikit-learn.org/stable/modules/neighbors.html#nearest-neighbors-transformer)?  If nanoflann is the same as scikit-learn's then I would suggest using that instead.

python/dgl/transform.py
```diff
@@ -208,22 +222,102 @@ def segmented_knn_graph(x, k, segs):
...
+    out = F.zeros((F.shape(y)[0] * k, 2), F.dtype(x_segs), F.context(x_segs))
+
+    # if use cosine distance, normalize input points first
+    # thus we can use euclidean distance to find knn equivalently.
+    if dist == "cosine":
+        l2_norm = lambda v : F.sqrt(F.sum(v * v, dim=1, keepdims=True))
+        x = x / (l2_norm(x) + 1e-5)
+        y = y / (l2_norm(y) + 1e-5)
+
+    _CAPI_DGLKNN(F.zerocopy_to_dgl_ndarray(x),
```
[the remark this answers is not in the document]
BarclayII: Well I would prefer prefixing with `DGL` to stay consistent to other functions' namings.

python/dgl/transform.py
```diff
@@ -208,22 +222,102 @@ def segmented_knn_graph(x, k, segs):
...
+                 F.zerocopy_to_dgl_ndarray(x_offset),
```
BarclayII, line 314: You can use `F.to_dgl_nd` and `F.from_dgl_nd` just in case you don't like typing a lot.

src/graph/transform/kdtree_ndarray_adaptor.h
```diff
@@ -0,0 +1,115 @@
+/*!
+ *  Copyright (c) 2021 by Contributors
+ * \file graph/transform/kdtree_ndarray_adaptor.h
+ * \brief NDArray adaptor for nanoflann, without
```
BarclayII, line 4: `adapter` - this spelling is more common in our codebase.
Please also scan for other occurrences (including class names such as `KDTreeNDArrayAdapter`.

src/graph/transform/kdtree_ndarray_adaptor.h
BarclayII: This is for typesetting typewriter fonts in doxygen I suppose.

src/graph/transform/kdtree_ndarray_adaptor.h
```diff
@@ -0,0 +1,115 @@
...
+      << "template argument.";
+    m_index = new index_type(static_cast<int>(dims), *this, nanoflann::KDTreeSingleIndexAdaptorParams(leaf_max_size));
+    m_index->buildIndex();
+  }
+
+  ~KDTreeNDArrayAdaptor() {
+    delete m_index;
+  }
+
+  index_type * GetIndex() {
```
BarclayII, line 52: Either `index_type* GetIndex()` or `index_type *GetIndex()` looks good (our codebase prefers the former).  `index_type * GetIndex()` looks ~~heretic~~ like a multiplication and therefore should be avoided.

Same for references (e.g. `self_type& derived()`)

src/graph/transform/knn.cc
BarclayII: You don't want to; shapes are indeed `int64_t`.

src/graph/transform/knn.cc
```diff
@@ -0,0 +1,117 @@
...
+    IdArray result = args[5];
+
+    aten::CheckContiguous(
+      {data_points, data_offsets, query_points, query_offsets, result},
+      {"data_points", "data_offsets", "query_points", "query_offsets", "result"});
+    aten::CheckCtx(
+      data_points->ctx, {data_offsets, query_points, query_offsets, result},
+      {"data_offsets", "query_points", "query_offsets", "result"});
+
+    ATEN_XPU_SWITCH_CUDA(data_points->ctx.device_type, XPU, "KNN", {
```
BarclayII, line 106: Seems that we don't support CUDA for now.  Change to `ATEN_XPU_SWITCH` so it will fail on GPU tensors.

src/graph/transform/knn.cc
```diff
@@ -0,0 +1,117 @@
...
+template <typename FloatType, typename IdType>
+void KdTreeKNN(const NDArray data_points, const IdArray data_offsets,
+               const NDArray query_points, const IdArray query_offsets,
+               const int k, IdArray result) {
+  int64_t batch_size = data_offsets->shape[0] - 1;
+  int64_t feature_size = data_points->shape[1];
+  IdType * data_offsets_data = static_cast<IdType*>(data_offsets->data);
+  IdType * query_offsets_data = static_cast<IdType*>(query_offsets->data);
+  IdType * result_data = static_cast<IdType*>(result->data);
+  FloatType * query_points_data = static_cast<FloatType*>(query_points->data);
```
BarclayII, line 30: Missing `const` qualifiers (except `result_data`).  In general, we would prepend all variables that shouldn't be mutated with `const` qualifiers.  Please check for other places as well.

We also no longer use `static_cast` in new code.  Please use `query_offsets.Ptr<IdType>()` to do the same thing.

src/graph/transform/knn.cc
```diff
@@ -0,0 +1,117 @@
...
+
+  for (int64_t b = 0; b < batch_size; ++b) {
+    auto d_offset = data_offsets_data[b];
+    auto d_length = data_offsets_data[b + 1] - d_offset;
+    auto q_offset = query_offsets_data[b];
+    auto q_length = query_offsets_data[b + 1] - q_offset;
+    auto out_offset = 2 * k * q_offset;
+
+    // create view for each segment
+    NDArray current_data_points = const_cast<NDArray*>(&data_points)->CreateView(
```
BarclayII, line 40: Why `const_cast`?

src/graph/transform/knn.cc
```diff
@@ -0,0 +1,117 @@
...
+      {d_length, feature_size}, data_points->dtype, d_offset * feature_size * sizeof(FloatType));
+    FloatType * current_query_pts_data = query_points_data + q_offset * feature_size;
+
+    KDTreeNDArrayAdaptor<FloatType, IdType> kdtree(feature_size, current_data_points);
+
+    // query
+#pragma omp parallel for
+    for (int64_t q = 0; q < q_length; ++q) {
+      std::vector<IdType> out_buffer(k);
+      std::vector<FloatType> out_dist_buffer(k);
```
BarclayII, line 50: Can we avoid allocating `out_buffer` and `out_dist_buffer` per iteration?

You can try OMP's `firstprivate` clause.

src/graph/transform/knn.cc
```diff
@@ -0,0 +1,117 @@
...
+
+using namespace dgl::runtime;
+using namespace dgl::transform::knn_utils;
+namespace dgl {
+namespace transform {
+namespace impl {
+
+/*! \brief The kd-tree implementation of K-Nearest Neighbors */
+template <typename FloatType, typename IdType>
+void KdTreeKNN(const NDArray data_points, const IdArray data_offsets,
```
BarclayII, line 22: `const NDArray&`.  Please check for similar occurrences as well.

python/dgl/transform.py
```diff
@@ -129,29 +132,37 @@ def knn_graph(x, k):
...
+            x_seg = x_size[0] * [x_size[1]]
+        else:
+            x_seg = [F.shape(x)[0]]
+        return segmented_knn_graph(x, k, x_seg, use_dgl_impl=True)
+    else:
+        if F.ndim(x) == 2:
+            x = F.unsqueeze(x, 0)
+        n_samples, n_points, _ = F.shape(x)
+
+        dist = pairwise_squared_distance(x)
```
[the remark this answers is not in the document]
BarclayII: Does nanoflann support bruteforcing?  If so this could be done in this PR.  If not and we still want to have two algorithms then I suggest to do it in a separate PR.

python/dgl/transform.py
```diff
@@ -129,29 +132,37 @@ def knn_graph(x, k):
...
-    return convert.from_scipy(adj)
+    if use_dgl_impl:
+        if F.ndim(x) == 3:
+            x_size = tuple(F.shape(x))
+            x = F.reshape(x, (x_size[0] * x_size[1], x_size[2]))
+            x_seg = x_size[0] * [x_size[1]]
+        else:
+            x_seg = [F.shape(x)[0]]
+        return segmented_knn_graph(x, k, x_seg, use_dgl_impl=True)
+    else:
```
BarclayII, line 143: I suggest packing the `else` block into another function (like `_knn_graph_bruteforce` or something).

python/dgl/transform.py
```diff
@@ -208,22 +222,102 @@ def segmented_knn_graph(x, k, segs):
...
-    dst = F.reshape(dst, (-1,))
-    src = F.reshape(src, (-1,))
-    adj = sparse.csr_matrix((F.asnumpy(F.zeros_like(dst) + 1), (F.asnumpy(dst), F.asnumpy(src))))
-
-    return convert.from_scipy(adj)
+    if use_dgl_impl:
+        out = knn(x, segs, x, segs, k).t().contiguous()
+        row, col = out[1], out[0]
+        return convert.graph((row, col))
+    else:
```
BarclayII, line 229: Same here.

python/dgl/transform.py
```diff
@@ -129,29 +132,37 @@ def knn_graph(x, k):
...
-
-    return convert.from_scipy(adj)
+    if use_dgl_impl:
+        if F.ndim(x) == 3:
+            x_size = tuple(F.shape(x))
+            x = F.reshape(x, (x_size[0] * x_size[1], x_size[2]))
+            x_seg = x_size[0] * [x_size[1]]
+        else:
+            x_seg = [F.shape(x)[0]]
+        return segmented_knn_graph(x, k, x_seg, use_dgl_impl=True)
```
BarclayII, line 142: I personally would directly call `knn` here.  Otherwise when I read the code I will spend a few seconds wondering why I call `segmented_knn_graph` in one branch but not in the other.

src/graph/transform/kdtree_ndarray_adaptor.h
```diff
@@ -0,0 +1,115 @@
...
+   *
+   */
+  template <typename BBOX>
+  bool kdtree_get_bbox(BBOX & /* bb */) const {
+    return false;
+  }
+
+private:
+  index_type * m_index;    // The kd tree index
+  const NDArray m_data;  // data points
```
BarclayII, line 108: We usually suffix the private members with an underscore (like `index_`) instead of prefixing with an `m_`.

BarclayII (changes requested):

python/dgl/transform.py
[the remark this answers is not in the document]
lygztq: The nanoflann is exact ("No support for approximate NN is provided"). I have tried sklearn's implementation and add experiment results above. It looks like that sklearn's kd-tree performs worse than nanoflann.

python/dgl/transform.py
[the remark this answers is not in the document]
lygztq: nanoflann is a kd-tree implementation, it does not support sorting (brute force).

python/dgl/transform.py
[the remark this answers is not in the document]
hetong007: Based on the above benchmark on `num_points` and `feature_dim`, I think bruteforce and kd-tree algorithms have their own preferred data distribution. 

Besides, when k << n, the current implementation in python that computes an n-by-n distance matrix can be suboptimal, as one can easily avoid the memory consumption by implementing the same algorithm in c++. To claim one is universally more efficient than another, we might need a comprehensive study on various implementations over as many data distributions as possible.

python/dgl/transform.py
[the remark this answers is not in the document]
hetong007: Yep I agree to have it in another PR, and in that PR we should support both CPU and GPU.

python/dgl/transform.py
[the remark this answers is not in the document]
lygztq: By the way, I'm not sure whether to add algorithm option (e.g. "brute", "kd-tree" and "ball-tree") because on GPU we don't use kd-tree and ball-tree.

python/dgl/transform.py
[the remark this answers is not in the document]
hetong007: OK I agree then for the consistency.

python/dgl/transform.py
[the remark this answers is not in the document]
hetong007: Well we never know, like there are advanced approximation algorithm, they might be added to DGL in the future.

src/graph/transform/knn.cc
lygztq: You are right. If we want to improve KNN (for CPU) in future, we can add new implementations (e.g. ball tree) here

python/dgl/transform.py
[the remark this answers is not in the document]
lygztq: Maybe we can add a check here like "if you want to use kd-tree, your data must on CPU".

src/graph/transform/knn.cc
lygztq: I add an `algorithm` option here

python/dgl/transform.py
lygztq: Currently, if the input tensor is not on the CPU, we will perform a copy operation.

src/graph/transform/knn.cc
lygztq: Because `CreateView` is not a const member function.

python/dgl/transform.py
hetong007: This behavior should definitely let user know. Either we ask users to copy to CPU themselves, or at least print a warning as the notification. GPU to CPU copy is slow, and I don't see the code copying the result back to GPU -- this will also break user experience.

lygztq: `knn_graph` and `segmented_knn_graph` return cpu graphs, regardless of the context of inputs. I also add a warning in `knn` if input tensors are not on cpu

python/dgl/transform.py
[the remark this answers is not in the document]
BarclayII: How does KD-tree compare against brute-force (the current implementation) on CPU?  If KD-tree is still better than brute-force then I would say always use KD-tree for CPU and brute-force for GPU, and not provide the algorithm argument.

python/dgl/transform.py
BarclayII: So the input features are copied to CPU before brute-force?  If so we can just do the same and always use KD-tree once we prove that KD-tree is better than brute-force.

python/dgl/transform.py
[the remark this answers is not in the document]
hetong007: @BarclayII Please check the result above. KD-tree is slower on high-dimensional data, which is common for learned embeddings.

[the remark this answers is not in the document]
lygztq: KD-tree is better than brute-force provided that the feature dimension `p` is small. When high dimensional data points are used, the KD-tree has poor performance compared to brute-force.

python/dgl/transform.py
lygztq: If use GPU and brute-force, kNN results will be first computed, then the results will be copied to CPU for graph construction.

src/graph/transform/knn.cc
BarclayII: I see.

As `current_data_points` are not mutated you need a `const` qualifier before.

@jermainewang I feel that we need to provide a `const NDArray NDArray::CreateView(...) const` member function.  What do you prefer?  Or you think `const_cast` is OK?

python/dgl/transform.py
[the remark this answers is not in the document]
BarclayII: I see.  We can either decide which algorithm to use for the users by thresholding, or giving an argument to specify which algorithm to use.

Also, would you mind comparing against `torch.cdist`?  For Minkowski distances the implementation in `pairwise_square_distance` is not the best I would say, especially since we require 1.5.0+ already.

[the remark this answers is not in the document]
lygztq: Actually the current API only support Euclidean distance (i.e. p=2). I will try the `torch.cdist` to see whether it has better performance typically when p=2.

BarclayII (approved):
I'm good with this PR.  How did `torch.cdist` go?

lygztq:
> I'm good with this PR. How did `torch.cdist` go?

The function signature `torch.cdist(x1, x2, p=2.0, compute_mode='use_mm_for_euclid_dist_if_necessary')`

I did some research on this function and found these facts:
- `torch.cdist` has a `compute_mode` option. Its default behavior is to try to use matrix multiplication (mm) approach to calculate euclidean distance (p = 2) if `|x1| > 25` or `|x2| > 25`. Otherwise it will use a pairwise bruteforce method for computation. (see [doc](https://pytorch.org/docs/master/generated/torch.cdist.html#torch-cdist) and [implementation](https://github.com/pytorch/pytorch/blob/2309173143abf3795e44a8472803d68849668434/aten/src/ATen/native/Distance.cpp#L44))
- When the size of input data is small (e.g. `|x| < 25`), `torch.cdist` is faster than the mm method. For example, if we have 10 points, `F=300` and `k=2`, `cdist` uses 0.0013s and mm method uses 0.0015s.
- When the size of input data is large, `torch.cdist` has less memory usage. For example, if we have 10 * 10000 points (in batch), `F=300` and `k=8`, the mm method will out of memory but `cdist` outputs the result in 0.2123s.
- In other cases, they are the same. (current `knn` in dgl assumes we only need the Euclidean distance)
- I think the mm method is better if we want to support `knn` for mxnet or TensorFlow in the future.

hetong007 (approved):

PR #2783, opened 2021-03-25
[Bug] Disable cpu fp16

BarclayII:
Currently when passed a float16 CPU array, we are using `float` pointers due to lack of native CPU fp16 support.  This can give wrong results.

Disabling CPU fp16 support for now.  GPU fp16 is not affected.


src/array/cpu/spmm_binary_ops.h
```diff
@@ -148,7 +148,9 @@ constexpr DType Min<DType>::zero;
 
 #define SWITCH_BITS(bits, DType, ...)                           \
   do {                                                          \
-    if ((bits) == 16 || (bits) == 32) {                         \
+    if ((bits) == 16) {                                         \
+      LOG(FATAL) << "FP16 not supported on CPU";                \
```
yzh119, line 152: I'm afraid it will fail compilation if you do not define DType here.

BarclayII: It won't because we don't have `__VA_ARGS__` in this branch.

yzh119: okay it's fine.

yzh119 (approved):

PR #2793, opened 2021-03-29
[Performance] Prefer parallelized conversion to CSC from COO instead of transposing CSR

BarclayII:
This makes `g.create_formats_()` significantly faster for large graphs such as OGB-MAG240M.


jermainewang (approved):

jermainewang:
Approved but would you mind adding some comments to these methods for what is the preference order of different formats?

BarclayII:
Do you think this is the proper solution?  I'm not sure if this will have any side effects such as regressions on say GPU conversions.

jermainewang:
@VoVAllen This is an example PR that needs benchmark tests triggered. How to test it right now?

BarclayII:
I would like to have this PR merged since #2810 depends on this PR performance-wise.

PR #2796, opened 2021-03-29
[Bug] Fix HGT example where relation-specific value tensors are overwritten

### LINKED ISSUE #2785: HGT implementation bug?
https://github.com/dmlc/dgl/blob/master/examples/pytorch/hgt/model.py#L81

The implemented hgt set value of nodes multiple times when calculating by different edge attention, which will cause the `nodes["v"]` of earlier etype be overwritten by later one.

BarclayII:
Fixes #2785 .


yzh119 (approved):
LGTM

PR #2798, opened 2021-03-30
[Distributed] Bind to the correct NIC

Tonny-Gu:
## Description
On HPC clusters, each node may have one Ethernet NIC and one Infiniband Adapter. If the default route rule uses Ethernet NIC to transfer data, whatever the content of `ip_config.txt` is, either the address of Ethernet or that of Infiniband, DGL will **only use Ethernet NIC** to communicate since the function `get_local_usable_addr` always returns the local address of Ethernet NIC.
This modification could shift most of the traffic to the expected network interface, but there is a small amount of uncontrollable traffic.


zheng-da:
Thanks for your contribution. It tries to connect to one of the servers to get local addresses. With the current implementation, one of the machines still tries to connect to the local server to get the local address. Does this cause problems in your multiple NIC setting?

I'm curious why "there is a small amount of uncontrollable traffic"?

Tonny-Gu:
> Thanks for your contribution. It tries to connect to one of the servers to get local addresses. With this implementation, one of the machines still tries to connect to the local server to get the local address. Does this cause problems in your multiple NIC setting? @Tonny-Gu
>
> I'm curious why "there is a small amount of uncontrollable traffic"?

Here is the configuration of one of our nodes. As you might notice, there are two different types of NIC supporting TCP/IP stack.

```bash
# Node 8

# ifconfig
eno1: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 10.1.1.8  netmask 255.255.0.0  broadcast 10.1.255.255
...
ib0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 2044
        inet 10.2.1.8  netmask 255.255.0.0  broadcast 10.2.255.255
...

# route
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
default         mn01            0.0.0.0         UG    0      0        0 eno1
10.1.0.0        0.0.0.0         255.255.0.0     U     0      0        0 eno1
10.2.0.0        0.0.0.0         255.255.0.0     U     0      0        0 ib0
```

Note that there are two different subnets, `10.1.0.0/16` and `10.2.0.0/16`. According to the route table, the network traffic will go through `eno1` by default.

> ```python
> # Original Implementation
> sock.connect(('10.255.255.255', 1))
> ip_addr = sock.getsockname()[0]
> ```

The original implementation will try to connect to a hard-coded IP address `10.255.255.255`. Of course, it will fail to match any more specific route rule like `10.2.0.0/16 via ib0`. Thus, it will always request to this unreachable IP address using `eno1`, and return the local IP address of `eno1`.

```python
>>> import socket
>>> s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
>>> s.connect(('10.255.255.255', 1))
>>> s.getsockname()[0]
'10.1.1.8'
>>> s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
>>> s.connect(('10.1.0.2', 1))
>>> s.getsockname()[0]
'10.1.1.8'
>>> s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
>>> s.connect(('10.2.0.2', 1))
>>> s.getsockname()[0]
'10.2.1.8'
```

But the problem is I cannot choose the NIC to communicate among nodes by rewriting `ip_config.txt`. I expect to utilize the Ethernet if I wrote `10.1.1.8` in `ip_config.txt`, and utilize the Infiniband if I wrote `10.2.1.8`. What's worse, DGL will always select the Ethernet on our cluster, which degrades the performance significantly. 

As for the "uncontrollable traffic", my idea is even I patch the code, there is still a small amount of traffic always passing through the Ethernet, which is about 20Mbps when training distributed GraphSAGE. I did not figure out the cause so far, and I have no idea whether this traffic will affect the performance.

zheng-da:
I see. Thanks

zheng-da (approved):

PR #2801, opened 2021-03-30
[Bugfix] Fix processing of QM9EdgeDataset

milesial:
## Description

When QM9EdgeDataset was downloaded and processed for the first time (not cached), there was a typo in `self.process` that caused a `KeyError`.

```python
from dgl.data import QM9EdgeDataset
ds = QM9EdgeDataset(force_reload=True)
```
```
KeyError: 'target is not a file in the archive'
```
This PR fixes that and also removes duplicate code so this can't happen in the future.

## Checklist

## Changes
- [x] Changed `self.process` to call `self.load`

### DIFF
--- a/python/dgl/data/qm9_edge.py
+++ b/python/dgl/data/qm9_edge.py
@@ -135,7 +135,7 @@ class QM9EdgeDataset(DGLDataset):
                  raw_dir=None, 
                  force_reload=False, 
                  verbose=True):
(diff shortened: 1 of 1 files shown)


hengruizhang98:
@mufeili

mufeili (approved):
LGTM. Thank you for the fix.

PR #2812, opened 2021-04-04
[Distributed] Automatically setting the number of OMP threads for trainers

zheng-da:
## Description
When running more than one trainer in a machine, the number of OMP thread will be set to 1 by default. This makes CPU training very slow. This PR adds an argument to the launch script to customize the number of OMP threads in trainers.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/tools/launch.py
+++ b/tools/launch.py
@@ -145,7 +145,7 @@ def main():
     if args.num_omp_threads is None:
         # Here we assume all machines have the same number of CPU cores as the machine
         # where the launch script runs.
-        args.num_omp_threads = multiprocessing.cpu_count() // 2 // args.num_trainers
+        args.num_omp_threads = max(multiprocessing.cpu_count() // 2 // args.num_trainers, 1)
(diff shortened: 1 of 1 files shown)


tools/launch.py
```diff
@@ -137,6 +142,12 @@ def main():
             'A user has to specify a partition configuration file with --part_config.'
     assert args.ip_config is not None, \
             'A user has to specify an IP configuration file with --ip_config.'
+    if args.num_omp_threads is None:
+        # Here we assume all machines have the same number of CPU cores as the machine
+        # where the launch script runs.
+        args.num_omp_threads = multiprocessing.cpu_count() // 2 // args.num_trainers
```
classicsong, line 148: What will happen when it is 0? (if multiprocessing.cpu_count() // 2 < args.num_trainers.)

classicsong (approved):
LGTM.
Left one comment.

tools/launch.py
zheng-da: good point. i'll set a minimum value.

PR #2818, opened 2021-04-06
[Feature][Performance] Add NCCL implementation for NodeEmbedding and SparseOptimizer

nv-dlasalle:
## Description
This PR implements NCCL all-to-all-v, for performing GPU communication of the sparse embbeding during forward, and communicating the gradients during backward.

This PR is quite large, so I let me know if you want me to split it up (e.g., a NCCL PR, and then a PR the sparse embedding and a sparse optimizer).

There are two changes that are needed in the future that are not part of this PR:
1. Allowing more flexible partitioning. That is, rather than only using `remainder` to assign indexes to processors, allow a range based partition, where a vector part_ptr[N_PROCS+1] could be passed in, and each index would be assigned to a processor based on that.
2. Allowing using NCCL for communication in the SparseOptimizer, while storing the embedding the CPU in shared memory.

## Checklist
- [X] Code is well-documented

## Changes

This adds NCCL wrappers in /python/dgl/cuda/nccl.py, and /src/runtime/cuda/nccl_api.{h,cu}. This also incorporates it into the cmake files.

This adds an extra `device` parameter to the NodeEmbedding, which allows storing it the on the GPU and using the NCCL wrapper.

This modifies the SparseOptimizer, to check of the embedding has a nccl communicator, and if it does, utilizes that for communicating the gradients.

This modifies the /examples/pytorch/rgcn/entity_classify_mp.py example to accept the `--embedding-gpu` flag to make use of the new code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,6 +157,10 @@ list(APPEND DGL_SRC ${DGL_SRC_1})
 if(USE_CUDA)
   dgl_config_cuda(DGL_CUDA_SRC)
   list(APPEND DGL_SRC ${DGL_CUDA_SRC})
+
+  include(cmake/util/FindNccl.cmake)
+  include_directories(${NCCL_INCLUDE_DIR})
+  list(APPEND DGL_LINKER_LIBS ${NCCL_LIBRARY})
 endif(USE_CUDA)
 
 if(USE_CUDA)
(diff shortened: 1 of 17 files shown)


nv-dlasalle:
Closing in favor of #2824 and #2825

PR #2819, opened 2021-04-07
Fix discrepancy in dgl.dataloading.EdgeDataLoader

### LINKED ISSUE #2780: reverse_edge_ids  in EdgeDataLoader versus reverse_eids in EdgeCollator
Thanks for the great library!

dgl.dataloading.pytorch.EdgeDataLoader accepts the keyword argument **reverse_edge_ids**. 
https://docs.dgl.ai/api/python/dgl.dataloading.html

dgl.dataloading.EdgeCollator however requires **reverse_eids**.
https://docs.dgl.ai/api/python/dgl.dataloading.html#dgl.dataloading.EdgeCollator

Maybe you can adjust this naming discrepancy. Moreover, I would suggest updating the documentation of the "reverse_eids". It sounds like it requires a mapping "eid to reverse eid", while in fact, it takes a list of reverse eids which should be in the same order as the non-reversed eids.

Cheers and keep up the great work!

BarclayII:
Fixes #2780.


mufeili:
Can you also address this "Moreover, I would suggest updating the documentation of the "reverse_eids". It sounds like it requires a mapping "eid to reverse eid", while in fact, it takes a list of reverse eids which should be in the same order as the non-reversed eids."

mufeili (approved):
LGTM

PR #2824, opened 2021-04-07
[Feature] Allow using NCCL for communication in dgl.NodeEmbedding and dgl.SparseOptimizer

nv-dlasalle:
## Description
This PR implements keeping embedding weights split across GPUs, and using NCCL to communicate those embeddings during forward, and communicate the gradients during backwards.

There are two changes that are needed in the future that are not part of this PR:
- Allowing more flexible partitioning. That is, rather than only using remainder to assign indexes to processors, allow a range based partition, where a vector part_ptr[N_PROCS+1] could be passed in, and each index would be assigned to a processor based on that.
- Allowing using NCCL for communication in the SparseOptimizer, while storing the embedding the CPU in shared memory.

This PR does not include the actual NCCL implementation, as that has been split into another PR.

## Checklist
- [x] Code is well-documented

## Changes
This adds an extra device parameter to the NodeEmbedding, which allows storing it the on the GPU and using the NCCL wrapper.

This modifies the SparseOptimizer, to check of the embedding has a nccl communicator, and if it does, utilizes that for communicating the gradients.

This modifies the /examples/pytorch/rgcn/entity_classify_mp.py example to accept the `--embedding-gpu` flag to make use of the new code.


nv-dlasalle:
@classicsong

examples/pytorch/rgcn/entity_classify_mp.py
```diff
@@ -541,7 +543,7 @@ def main(args, devices):
     n_gpus = len(devices)
     n_cpus = mp.cpu_count()
     # cpu
-    if devices[0] == -1:
```
classicsong, line 544: Why we need n_gpus == 1? This is for CPU training. The device is set to ['cpu']

python/dgl/cuda/nccl.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+
+    def sparse_all_to_all_push(self, idx, value, mode):
+        """ Perform an all-to-all-v operation, where by all processors send out
+            a set of indices and corresponding values.
+
+            Parameters
+            ----------
+            idx : IdArray
+                The 1D set of indices to send to other processors.
```
classicsong, line 38: Is this store the indices of the processors?

python/dgl/cuda/nccl.py
```diff
@@ -0,0 +1,95 @@
...
+                The set of recieved indices.
+            NDArray
+                The set of recieved values.
+
+        """
+        raise NotImplementedError
+
+    def sparse_all_to_all_pull(self, req_idx, value, mode):
+        """ Perform an all-to-all-v operation, where by all processors request
+            the values corresponding to ther set of indices.
```
classicsong, line 57: Typo: ther -> the

python/dgl/nn/pytorch/sparse_emb.py
```diff
@@ -109,10 +169,19 @@ def __call__(self, node_ids, device=th.device('cpu')):
         device : th.device
             Target device to put the collected embeddings.
         """
-        emb = self._tensor[node_ids].to(device)
+        if not self._comm or self._comm.size() == 1:
+            emb = self._tensor[node_ids].to(device)
+        else:
+            if self.world_size > 0:
+                emb = self._comm.sparse_all_to_all_pull(
```
classicsong, line 176: How can the comm know the partition/split of the range?

classicsong (commented):
I have a general question, what will happen if the embeddings can not be stored in GPU memory? At this time, we need to store the embeddings in CPU, but we still can use alltoallv to sync gradients during step.

examples/pytorch/rgcn/entity_classify_mp.py
nv-dlasalle: I was doing some multi-socket CPU only testing (`--gpu="-1,-1"`), and this would break. I can remove it if you want.

python/dgl/cuda/nccl.py
nv-dlasalle: This is the indices corresponding to the values. That is, if we have 2 processors, and processor 0 calls this with idx=[0,1,2,3,4,5] and mode='remainder'.

Then it would send `1,3,5` and the corresponding values to processor 1, and receive any even indices from processor 1.

nv-dlasalle:
@classicsong I haven't implemented that part yet. I see two options:
1. We could add a flag to the sparse optimizer's constructor, telling it exchange gradients on the GPU despite the embedding being the CPU. It would then proceed to create it's own NCCL communicator.
2. In the first step of the optimizer, it could see if the gradients it received are on the GPU, create a NCCL communicator there, and the proceed to exchange the gradients.
Let me know which you would prefer, and whether you want the change in this PR, or a follow up PR. Thanks.

python/dgl/cuda/nccl.py
nv-dlasalle: Fixed in 
42e9d49

python/dgl/cuda/nccl.py
nv-dlasalle: I also expanded the documentation of this function in 
5c805fa to make this more clear.

classicsong:
> @classicsong I haven't implemented that part yet. I see two options:
> 
> 1. We could add a flag to the sparse optimizer's constructor, telling it exchange gradients on the GPU despite the embedding being the CPU. It would then proceed to create it's own NCCL communicator.
> 2. In the first step of the optimizer, it could see if the gradients it received are on the GPU, create a NCCL communicator there, and the proceed to exchange the gradients.
>    Let me know which you would prefer, and whether you want the change in this PR, or a follow up PR. Thanks.

I prefer the second. The gradients exchange can be decoupled with gradients update. If the gradients are in GPU at the very beginning, we use NCCL communicator to exchange gradients.

examples/pytorch/rgcn/entity_classify_mp.py
classicsong: I c.
My suggestion is keepping it unchanged.

python/dgl/cuda/nccl.py
classicsong: The expanded doc is much clear.

examples/pytorch/rgcn/entity_classify_mp.py
nv-dlasalle: Reverted in 
01c653f

nv-dlasalle:
> I prefer the second. The gradients exchange can be decoupled with gradients update. If the gradients are in GPU at the very beginning, we use NCCL communicator to exchange gradients.

I've implemented the second option now in sparse_optim.py.

VoVAllen:
Did those tests contain gpu related test yet?

python/dgl/nn/pytorch/sparse_emb.py
```diff
@@ -70,32 +82,75 @@ def __init__(self, num_embeddings, embedding_dim, name,
...
+            if rank == 0: # the master gpu process
+                for _ in range(1, world_size):
+                    # send embs
+                    self._store.set(name, name)
+            elif rank > 0:
+                # receive
+                self._store.wait([name])
+                emb = get_shared_mem_array(name, (num_embeddings, embedding_dim), th.float32)
+            self._tensor = emb
+        else:
```
classicsong, line 114:
```suggestion
        else: # embeddings is stored in GPU memory.
```

classicsong (commented):
Can you add a save_tensor or get_tensor API for NodeEmbedding?
This is useful when someone wants to save the initial embedding while the embedding is actually store in multiple GPUs.

nv-dlasalle:
> 
> 
> Can you add a save_tensor or get_tensor API for NodeEmbedding?
> This is useful when someone wants to save the initial embedding while the embedding is actually store in multiple GPUs.
Do you have any thoughts on the strategy for this? Should we allocate a shared memory tensor, and then have all GPUs copy their sections to that tensor? How would this work for DistDGL where the tensor can't find into a single machines memory?

examples/pytorch/rgcn/model.py
```diff
@@ -97,14 +101,15 @@ def __init__(self,
             if isinstance(input_size[ntype], int):
                 if dgl_sparse:
                     self.node_embeds[str(ntype)] = dgl.nn.NodeEmbedding(input_size[ntype], embed_size, name=str(ntype),
-                        init_func=initializer)
+                        init_func=initializer, device=self.dev_id)
```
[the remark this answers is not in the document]
nv-dlasalle: Exactly.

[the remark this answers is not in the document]
nv-dlasalle: However, in the case where we store the embedding on the CPU but do the optimization on the GPU, we still have to slice the embedding state on the CPU, copy the sliced portion to the GPU, calculate the updates, copy those back down to the CPU, and scatter them back into the embedding.

python/dgl/backend/pytorch/tensor.py
```diff
@@ -332,12 +332,12 @@ def zerocopy_to_dgl_ndarray_for_write(input):
...
 
 def zerocopy_from_dgl_ndarray(data):
-    if data.shape == (0,):
+    if len(data.shape) == 0 or builtins.min(data.shape) == 0:
         # NOTE: PyTorch v1.5 does not accept DLPack object representing empty CUDA tensor.
         #  Related issue: https://github.com/pytorch/pytorch/issues/41182
         #  The issue will be fixed in v1.6 and later.
-        return th.tensor([], dtype=getattr(th, data.dtype),
-                         device=to_backend_ctx(data.ctx))
+        return th.empty(data.shape, dtype=getattr(th, data.dtype),
```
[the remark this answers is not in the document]
nv-dlasalle: I've reverted this in a8ad4c8 , so that only behavior which changes, is that which previous resulted in error (e.g., a tensor with shape [0,5]).

classicsong:
> > Can you add a save_tensor or get_tensor API for NodeEmbedding?
> > This is useful when someone wants to save the initial embedding while the embedding is actually store in multiple GPUs.
> 
> @classicsong Do you have any thoughts on the strategy for this? Should we allocate a shared memory tensor, and then have all GPUs copy their sections to that tensor? How is this currently done in DistDGL?

We can allocate a shared memory tensor to have all GPUs copy their sections to that tensor.
For DistDGL, we need some extra discussion about it.

examples/pytorch/rgcn/entity_classify_mp.py
```diff
@@ -532,6 +513,14 @@ def main(args, devices):
     train_idx.share_memory_()
     val_idx.share_memory_()
     test_idx.share_memory_()
+
+    # make a mapping from target type back to global id
```
classicsong, line 517: This implementation is a little bit tricky. But I think this is the only choice we have if we want to use dgl.dataloading.NodeDataLoader.

Can we add more docstring here to tell what we are doing here?

examples/pytorch/rgcn/model.py
```diff
@@ -76,6 +78,7 @@ class RelGraphEmbedLayer(nn.Module):
     """
     def __init__(self,
                  dev_id,
+                 out_dev_id,
```
classicsong, line 81: I think we need some comments about what is dev_id and what is out_dev_id.

src/partition/cuda/partition_op.cu
```diff
@@ -56,6 +56,22 @@ __global__ void _MapLocalIndexByRemainder(
...
+    const IdType * const in,
+    IdType * const out,
+    const int part_id,
+    const int64_t num_items,
+    const int comm_size) {
+  const int64_t idx = threadIdx.x+blockDim.x*blockIdx.x;
+
+  assert(part_id < comm_size);
+
+  if (idx < num_items) {
```
classicsong, line 70: Do we need a for loop here?

classicsong (approved):
Only little comments
Overall LGTM.

examples/pytorch/rgcn/entity_classify_mp.py
nv-dlasalle: Added a docstring here to explain how it's used for the labels, and inside of the training loop where we do `inv_target[seeds]`, in 
50a9b26.

examples/pytorch/rgcn/model.py
nv-dlasalle: I expanded the comments for parameters `dev_id` and `out_dev_id`, and I also changed the name of `dev_id` to `storage_dev_id` to make it more clear it it's used for, in 
50a9b26.

src/partition/cuda/partition_op.cu
nv-dlasalle: No, this is always launched with enough threads to cover the `num_items`. I've added assertions to all of these kernels to ensure this is the case (when run in debug mode), in 
50a9b26.

PR #2825, opened 2021-04-07
[Feature][Performance] Implement NCCL wrapper for communicating NodeEmbeddings and sparse gradients.

nv-dlasalle:
## Description
This PR implements NCCL all-to-all-v at a high-level for performing GPU communication in NodeEmbedding and the SparseOptimizer. It accompanies #2824.

Communication is split into 1) a sparse all-to-all push, where each processor has a set of indices and corresponding values to distribute among the processors (used in gradient synchronization), and into 2) a sparse all-to-all pull, where each processor sends out a set of indices it is requesting the corresponding values to, and then all processors fulfill these requests (used in communicating node embeddings).

In the future, we will want to allow more flexible partitioning. That is, rather than only using remainder to assign indexes to processors, allow a range based partition, where a vector part_ptr[N_PROCS+1] could be passed in, and each index would be assigned to a processor based on that.

## Checklist
- [x] Code is well-documented

## Changes
This implements a NCCL wrapper at /src/runtime/cuda/nccl_api.cuh, and at /python/dgl/cuda/nccl.py.

It also adds a build time dependency of NCCL, and accompanying CMake module to find it on the system.

I don't know of a good way to test this in the CI, as it looks like that environment only has a single GPU. As a result, the current testing only covers a fraction of the code.


nv-dlasalle:
@classicsong

CMakeLists.txt
```diff
@@ -157,6 +157,10 @@ list(APPEND DGL_SRC ${DGL_SRC_1})
 if(USE_CUDA)
   dgl_config_cuda(DGL_CUDA_SRC)
   list(APPEND DGL_SRC ${DGL_CUDA_SRC})
+
+  include(cmake/util/FindNccl.cmake)
```
classicsong, line 161: Is this nccl included in cuda package?

python/dgl/cuda/nccl.py
```diff
@@ -0,0 +1,143 @@
...
+        mode_id = _COMM_MODES_MAP[mode]
+
+        out_idx, out_value = _CAPI_DGLNCCLSparseAllToAllPush(
+            self.get(), F.zerocopy_to_dgl_ndarray(idx),
+            F.zerocopy_to_dgl_ndarray(value),
+            mode_id)
+        return (F.zerocopy_from_dgl_ndarray(out_idx),
+                F.zerocopy_from_dgl_ndarray(out_value))
+
+    def sparse_all_to_all_pull(self, req_idx, value, mode):
```
classicsong, line 90: The API will be changed if we introduce distributed training.
Can we break this API into multiple APIs including 1) get req_idx split according to mode; 2) play alltoallv to synchronize tensors; 
Then when we support distributed all_to_all_pull and all_to_all_push, we only need to provide the distributed version of req_idx split (according to mode and the partition book).

classicsong (commented):
The only concern is that the distributed version will break some APIs like all-to-all-v-pull and push.

CMakeLists.txt
nv-dlasalle: No, by default NCCL is not part of the cuda package, but it does require cuda. Most frameworks will also require NCCL.

src/runtime/cuda/nccl_api.cu
```diff
@@ -0,0 +1,1102 @@
...
+      // comm_size is not a power of 2
+      _MapProcByRemainder<<<grid, block, 0, stream>>>(
+          in_idx,
+          num_in,
+          comm_size,
+          proc_id_in);
+      CUDA_CALL(cudaGetLastError());
+    } else {
+      // comm_size is a power of 2
+      _MapProcByMaskRemainder<<<grid, block, 0, stream>>>(
```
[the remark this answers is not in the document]
nv-dlasalle: Right, this uses `&` instead of `%`, so it's a faster path.

src/runtime/cuda/nccl_api.cu
```diff
@@ -0,0 +1,1102 @@
...
+  int64_t * send_prefix = static_cast<int64_t*>(
+      device->AllocWorkspace(ctx, sizeof(int64_t)*(comm_size+1)));
+  {
+    size_t prefix_workspace_size;
+    CUDA_CALL(cub::DeviceScan::ExclusiveSum(nullptr, prefix_workspace_size,
+        send_sum, send_prefix, comm_size+1));
+
+    void * prefix_workspace = device->AllocWorkspace(
+        ctx, prefix_workspace_size);
+    CUDA_CALL(cub::DeviceScan::ExclusiveSum(prefix_workspace, prefix_workspace_size,
```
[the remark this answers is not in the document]
nv-dlasalle: The first call gets the amount of workspace needed, and the second call actually performs the summation.
https://nvlabs.github.io/cub/structcub_1_1_device_scan.html#a02b2d2e98f89f80813460f6a6ea1692b
```
When d_temp_storage is NULL, no work is done and the required allocation size is returned in temp_storage_bytes.
```

python/dgl/cuda/nccl.py
nv-dlasalle: We could just add an assignment vector as your saying so the method would look like:
```
def sparse_all_to_all_pull(self, proc_id, req_idx, value):
```
Where `proc_id` and `req_idx` have the same dimensions.

Another option would be, we could change the `mode` parameter into something like a `Partitioning` object that would have a C++ method the communicator could call:
```
def sparse_all_to_all_pull(self, req_idx, value, partitioning):
```
```
template<typename IdType>
void Partitioning::Map(const IdType * idx, int64_t num, IdType * proc, cudaStream_t stream);
```

That way we could have tighter control of stream and memory management, as well as avoid the overheads associated with going back and forth with python.

src/runtime/cuda/nccl_api.cu
```diff
@@ -0,0 +1,1102 @@
...
+
+  BlockHistogram(temp_storage).Histogram(thread_vals, local_counts);
+
+  __syncthreads();
+
+  // write local histogram back to global memory
+  for (int i = threadIdx.x; i < num_counts; i+=BLOCK_SIZE) {
+    const int64_t val = local_counts[i];
+    if (val > 0) {
+      AtomicAdd(counts+i, val);
```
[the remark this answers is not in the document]
nv-dlasalle: In 38d7efe  I switched to using cub's histogram implementation, which should be quite performant.

nv-dlasalle:
@classicsong In my latest changes I've added an `NDArrayPartition` object, which for now only has a `RemainderPartition` as a sub-class, but in the future we can add a `RangePartition` sublcass to support the distributed case (and possibly a `RangeAndRemainderPartition` where we want to use graph partitioning across machines, but only remainder between GPUs on a machine).

src/runtime/cuda/nccl_api.cu
```diff
@@ -0,0 +1,832 @@
...
+}
+template<> ncclDataType_t NCCLType<float>() {
+    return ncclFloat32;
+}
+template<> ncclDataType_t NCCLType<double>() {
+    return ncclFloat64;
+}
+
+
+template<typename IdType> __global__ void _MapProcByRemainder(
```
classicsong, line 76: Same function as in src/partition/cuda/partition_op.cu.
Shall we remove this one?

src/runtime/cuda/nccl_api.cu
```diff
@@ -0,0 +1,832 @@
...
+    IdType * const proc_id) {
+  const int64_t idx = blockDim.x*static_cast<int64_t>(blockIdx.x)+threadIdx.x;
+
+  if (idx < num_index) {
+    proc_id[idx] = index[idx] % num_proc;
+  }
+}
+
+template<typename IdType>
+__global__ void _MapProcByMaskRemainder(
```
classicsong, line 89: Same here.

src/runtime/cuda/nccl_api.cu
```diff
@@ -0,0 +1,832 @@
...
+  }
+}
+
+template<typename IdType, typename DType>
+std::pair<IdArray, NDArray> SparsePush(
+    NCCLCommunicatorRef comm,
+    IdArray in_idx,
+    NDArray in_value,
+    NDArrayPartitionRef part) {
+  CHECK_EQ(in_idx->shape[0], in_value->shape[0]);
```
classicsong, line 219: Can we add some error str here?

src/runtime/cuda/nccl_api.cu
```diff
@@ -0,0 +1,832 @@
...
+
+  const auto& ctx = in_idx->ctx;
+  CHECK_EQ(ctx, in_value->ctx);
```
classicsong, line 222: Error string here.

src/partition/cuda/partition_op.cu
```diff
@@ -0,0 +1,185 @@
...
+          num_in,
+          static_cast<IdType>(num_parts-1),  // bit mask
+          proc_id_in);
+      CUDA_CALL(cudaGetLastError());
+    }
+  }
+
+  // then create a permutation array that groups processors together by
+  // performing a radix sort
+  IdType * proc_id_out = static_cast<IdType*>(
```
classicsong, line 108: Where to free this object?

src/runtime/cuda/nccl_api.cu
```diff
@@ -0,0 +1,832 @@
...
+    void * prefix_workspace = device->AllocWorkspace(
+        ctx, prefix_workspace_size);
+    CUDA_CALL(cub::DeviceScan::ExclusiveSum(prefix_workspace, prefix_workspace_size,
+        recv_sum, recv_prefix, comm_size+1));
+    device->FreeWorkspace(ctx, prefix_workspace);
+  }
+  device->FreeWorkspace(ctx, recv_sum);
+
+  // finally copy the prefixsum sum down to the host
+  std::vector<int64_t> recv_prefix_host(comm_size+1);
```
classicsong, line 321: Do we need to copy recv_predix into CPU?

src/runtime/cuda/nccl_api.cu
```diff
@@ -0,0 +1,832 @@
...
+        perm,
+        num_in,
+        num_feat,
+        send_idx,
+        send_value);
+    CUDA_CALL(cudaGetLastError());
+  }
+
+  // compute the prefix sum of the send values
+  int64_t * send_prefix = static_cast<int64_t*>(
```
classicsong, line 270: Where to release send_prefix? Or it is released automatically?

src/runtime/cuda/nccl_api.cu
```diff
@@ -0,0 +1,832 @@
...
+    void * prefix_workspace = device->AllocWorkspace(
+        ctx, prefix_workspace_size);
+    CUDA_CALL(cub::DeviceScan::ExclusiveSum(prefix_workspace, prefix_workspace_size,
+        recv_sum, response_prefix, comm_size+1, stream));
+    device->FreeWorkspace(ctx, prefix_workspace);
+  }
+  device->FreeWorkspace(ctx, recv_sum);
+
+  // finally copy the prefixsum sum down to the host
+  std::vector<int64_t> response_prefix_host(comm_size+1);
```
classicsong, line 475: Do we need to copy prefix sum to CPU?

src/runtime/cuda/nccl_api.cu
```diff
@@ -0,0 +1,832 @@
...
+    aten::impl::IndexSelectSingleKernel<<<grid, block, 0, stream>>>(
+        static_cast<const IdType*>(req_idx->data),
+        perm,
+        num_in,
+        send_idx);
+    CUDA_CALL(cudaGetLastError());
+  }
+
+  // compute the prefix sum of the indexes this process is requesting
+  int64_t * request_prefix = static_cast<int64_t*>(
```
classicsong, line 425: Where to release request_prefix?

src/runtime/cuda/nccl_api.cu
nv-dlasalle: This is needed for the call to NCCL, to specify which portions of the array need to be sent to which processors. See https://github.com/dmlc/dgl/pull/2825/files#diff-5d6e33a4f7ce35295db947aff3e3cb6462f630d6a7b85629c3b2a3eb5c32e98dR82

src/runtime/cuda/nccl_api.cu
nv-dlasalle: This is needed for the call to NCCL, to specify which portions of the array need to be sent to which processors. See https://github.com/dmlc/dgl/pull/2825/files#diff-5d6e33a4f7ce35295db947aff3e3cb6462f630d6a7b85629c3b2a3eb5c32e98dR82

src/runtime/cuda/nccl_api.cu
nv-dlasalle: Added in
4fe379c

src/runtime/cuda/nccl_api.cu
nv-dlasalle: Added in 
4fe379c

src/runtime/cuda/nccl_api.cu
nv-dlasalle: Removed in 
4a9a3c0

src/runtime/cuda/nccl_api.cu
nv-dlasalle: Removed in 
4a9a3c0

python/dgl/cuda/nccl.py
```diff
@@ -0,0 +1,141 @@
...
+            value : NDArray
+                The multi-dimension set of values that can be requested from
+                this processor.
+            partition : NDArrayPartition
+                The object containing information for assigning indices to
+                processors.
+
+            Returns
+            -------
+            NDArray
```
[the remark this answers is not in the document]
nv-dlasalle: Changed in 
720a0f2

python/dgl/cuda/nccl.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def sparse_all_to_all_pull(self, req_idx, value, partition):
+        """ Perform an all-to-all-v operation, where by all processors request
+            the values corresponding to ther set of indices.
+
+            Parameters
+            ----------
+            req_idx : IdArray
+                The set of indices this processor is requesting.
+            value : NDArray
```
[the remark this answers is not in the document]
nv-dlasalle: Changed in 
720a0f2

python/dgl/cuda/nccl.py
```diff
@@ -0,0 +1,141 @@
...
+        return (F.zerocopy_from_dgl_ndarray(out_idx),
+                F.zerocopy_from_dgl_ndarray(out_value))
+
+    def sparse_all_to_all_pull(self, req_idx, value, partition):
+        """ Perform an all-to-all-v operation, where by all processors request
+            the values corresponding to ther set of indices.
+
+            Parameters
+            ----------
+            req_idx : IdArray
```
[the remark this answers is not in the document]
nv-dlasalle: Changed in 
720a0f2

python/dgl/cuda/nccl.py
```diff
@@ -0,0 +1,141 @@
...
+                The 0th dimension must match that of `idx`.
+            partition : NDArrayPartition
+                The object containing information for assigning indices to
+                processors.
+
+            Returns
+            -------
+            IdArray
+                The set of recieved indices.
+            NDArray
```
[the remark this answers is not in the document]
nv-dlasalle: Changed in 
720a0f2

python/dgl/cuda/nccl.py
```diff
@@ -0,0 +1,141 @@
...
+            value : NDArray
+                The multi-dimension set of values to send to other processors.
+                The 0th dimension must match that of `idx`.
+            partition : NDArrayPartition
+                The object containing information for assigning indices to
+                processors.
+
+            Returns
+            -------
+            IdArray
```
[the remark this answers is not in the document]
nv-dlasalle: Changed in 
720a0f2

python/dgl/cuda/nccl.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def sparse_all_to_all_push(self, idx, value, partition):
+        """ Perform an all-to-all-v operation, where by all processors send out
+            a set of indices and corresponding values.
+
+            Parameters
+            ----------
+            idx : IdArray
+                The 1D set of indices to send to other processors.
+            value : NDArray
```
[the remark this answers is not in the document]
nv-dlasalle: Changed in 
720a0f2

python/dgl/cuda/nccl.py
```diff
@@ -0,0 +1,141 @@
...
+        self._rank = rank
+        self._size = size
+
+    def sparse_all_to_all_push(self, idx, value, partition):
+        """ Perform an all-to-all-v operation, where by all processors send out
+            a set of indices and corresponding values.
+
+            Parameters
+            ----------
+            idx : IdArray
```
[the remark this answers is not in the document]
nv-dlasalle: Changed in 
720a0f2

python/dgl/cuda/nccl.py
```diff
@@ -0,0 +1,141 @@
...
+            Parameters
+            ----------
+            size : int
+                The number of processes in the communicator.
+            rank : int
+                The rank of the current process in the communicator.
+            unique_id : NCCLUniqueId
+                The unique id of the root process (rank=0).
+        """
+        assert rank < size
```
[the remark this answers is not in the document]
nv-dlasalle: Added in 
720a0f2

src/partition/cuda/partition_op.cu
nv-dlasalle: I switched all workspace allocations to use RAII via https://github.com/dmlc/dgl/pull/2825/files#diff-128b2340cfe395c91243eab36571e6f38af6b4eb0d0c3e01eaaba552995d8828 in 
69daf52, to prevent leaking workspace.

src/runtime/cuda/nccl_api.cu
nv-dlasalle: I switched all workspace allocations to use RAII via https://github.com/dmlc/dgl/pull/2825/files#diff-128b2340cfe395c91243eab36571e6f38af6b4eb0d0c3e01eaaba552995d8828 in 
69daf52, to prevent leaking workspace.

src/runtime/cuda/nccl_api.cu
nv-dlasalle: I switched all workspace allocations to use RAII via https://github.com/dmlc/dgl/pull/2825/files#diff-128b2340cfe395c91243eab36571e6f38af6b4eb0d0c3e01eaaba552995d8828 in 
69daf52, to prevent leaking workspace.

python/dgl/cuda/nccl.py
```diff
@@ -0,0 +1,141 @@
...
+            self.get(), F.zerocopy_to_dgl_ndarray(idx),
+            F.zerocopy_to_dgl_ndarray(value),
+            partition.get())
+        return (F.zerocopy_from_dgl_ndarray(out_idx),
+                F.zerocopy_from_dgl_ndarray(out_value))
+
+    def sparse_all_to_all_pull(self, req_idx, value, partition):
+        """ Perform an all-to-all-v operation, where by all processors request
+            the values corresponding to ther set of indices.
+
```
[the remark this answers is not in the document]
nv-dlasalle: I added an example of the code to use it, as well as what the resulting arrays would look like in 
a31a237

python/dgl/cuda/nccl.py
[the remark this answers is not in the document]
nv-dlasalle: In 
a31a237 changed the doc string to include examples, and to explicitly call out that idx and value the belong to the current processor will be copied to the output.

classicsong (approved):
LGTM

PR #2863, opened 2021-04-23
[Doc]Fix examples on negative graph sampler

Theheavens:
## Description
In heterogeneous graph, the old example may build a negative graph without edges. In dgl.dataloading.dataloader line637, it shows the key of neg_edges dict should be item in g.canonical_etypes.


BarclayII (approved):

BarclayII:
Thanks!

PR #2867, opened 2021-04-25
[NN] Fix GATConv for Broadcasting with Residual Connections

mufeili:
## Description
Previously the forward computation of `GATConv` fails if `in_feats=out_feats` and `residual=True`.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/nn/pytorch/conv/gatconv.py
+++ b/python/dgl/nn/pytorch/conv/gatconv.py
@@ -304,7 +304,8 @@ class GATConv(nn.Module):
             rst = graph.dstdata['ft']
             # residual
             if self.res_fc is not None:
-                resval = self.res_fc(h_dst).view(h_dst.shape[0], self._num_heads, self._out_feats)
+                # Use -1 rather than self._num_heads to handle broadcasting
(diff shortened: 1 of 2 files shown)


python/dgl/nn/pytorch/conv/gatconv.py
```diff
@@ -304,7 +304,8 @@ def forward(self, graph, feat, get_attention=False):
             rst = graph.dstdata['ft']
             # residual
             if self.res_fc is not None:
-                resval = self.res_fc(h_dst).view(h_dst.shape[0], self._num_heads, self._out_feats)
+                # Use -1 rather than self._num_heads to handle broadcasting
+                resval = self.res_fc(h_dst).view(h_dst.shape[0], -1, self._out_feats)
```
yzh119, line 309: Could you please explain why does it work?

mufeili: If we use `self._num_heads`, then we will get an error with `view`. If we use -1, it will be 1 in this case. We can add a tensor of shape `(A, 1, C)` to a tensor of shape `(A, B, C)`.

BarclayII (approved):

PR #2881, opened 2021-04-28
[Distributed] Kill training jobs in distributed training

zheng-da:
## Description
Right now if training fails for any reason, we have to login to each machine to kill the training jobs. This is very inconvenient. This PR adds the functionality in the launch script so that when a user kills the training script with ctrl+c, the launch script will kill all training jobs in the cluster.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/tools/launch.py
+++ b/tools/launch.py
@@ -29,6 +29,7 @@ def cleanup_proc(remote_pids, conn):
         # Otherwise, we need to ssh to each machine and kill the training jobs.
         for (ip, port), pids in remote_pids.items():
             kill_process(ip, port, pids)
+    print('cleanup process exits')
 
 def kill_process(ip, port, pids):
     '''ssh to a remote machine and kill the specified processes.
@@ -66,6 +67,14 @@ def get_remote_pids(ip, port, cmd_regex):
         res = re.search(cmd_regex, p)
         if res is not None:
             pids.append(l[1])
+
+    print('get from ps:', pids)
+    pid_str = ','.join([str(pid) for pid in pids])
+    ps_cmd = 'ssh -o StrictHostKeyChecking=no -p ' + str(port) + ' ' + ip + ' \'pgrep -P {}\''.format(pid_str)
+    res = subprocess.run(ps_cmd, shell=True, stdout=subprocess.PIPE)
+    pids1 = res.stdout.decode('utf-8').split('\n')
+    pids = list(set(pids + pids1))
+    print('merge with pgrep:', pids)
     return pids
 
 remote_pids = {}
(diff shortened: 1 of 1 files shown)


tools/launch.py
```diff
@@ -9,10 +9,70 @@
...
+    # to Python's process pool. After sorting, we always kill parent processes first.
+    pids.sort()
+    for pid in pids:
+        assert curr_pid != pid
+        print('kill process {} on {}:{}'.format(pid, ip, port), flush=True)
+        kill_cmd = 'ssh -o StrictHostKeyChecking=no -p ' + str(port) + ' ' + ip + ' \'kill {}\''.format(pid)
+        subprocess.run(kill_cmd, shell=True)
+        killed_pids.append(pid)
+    # It's possible that some of the processes are not killed. Let's try again.
+    for i in range(3):
```
classicsong, line 51: Why it is 3? Can we do it like `while get_killed_pids(xxx) != 0:` ?

tools/launch.py
```diff
@@ -102,8 +205,25 @@ def submit_jobs(args, udf_command):
...
+    # Start a cleanup process dedicated for cleaning up remote training jobs.
+    conn1,conn2 = multiprocessing.Pipe()
+    func = partial(get_all_remote_pids, hosts, args.ssh_port, udf_command)
+    process = multiprocessing.Process(target=cleanup_proc, args=(func, conn1))
+    process.start()
+
+    def signal_handler(signal, frame):
+        logging.info('Stop launcher')
+        # We need to tell the cleanup process to kill remote training jobs.
+        conn2.send('cleanup')
```
classicsong, line 217: Who will receive this 'cleanup'?

tools/launch.py
```diff
@@ -102,8 +205,25 @@ def submit_jobs(args, udf_command):
...
+        sys.exit(0)
```
classicsong, line 218: Do we need to wait for clean_up_proc to exit?

classicsong (approved):

tools/launch.py
zheng-da: 3 is kind of a random number. i prefer to not run the loop forever in case we cannot kill some of the processes, although this is most likely not happen because i use "kill -9 pid" to kill the processes.

tools/launch.py
zheng-da: the cleanup process receive the message.

tools/launch.py
zheng-da: even if we don't call `sys.exit`, the launch process will still be killed by the signal.

PR #2882, opened 2021-04-28
Add heterograph support in C kernels

isratnisa:
## Description
Added heterograph support for C SpMM and SDDMM kernel. Functions are named as *Hetero for now to distinguish from the original codes for homogeneous graphs. Eventually both functions will be merged to one. This code avoids multiple python invocation for each relation type. Inside the C kernel, it iterates over the relation types and calls the SpMM (or SDDMM) function sequentially. This is a basic solution and can be improved. 

## Checklist
- [ ] Code is well-documented


zheng-da:
does it support both SpMM and SDDMM?

src/array/cpu/spmm.cc
```diff
@@ -40,6 +40,49 @@ void SpMMCsr(const std::string& op, const std::string& reduce,
...
+  //     SWITCH_OP(op, Op, {
+  //       if (reduce == "max")
+  //         cpu::SpMMCmpCsr<IdType, DType, Op, cpu::op::Max<DType>>(
+  //             bcast, csr, ufeat, efeat, out, out_aux[0], out_aux[1]);
+  //       else
+  //         cpu::SpMMCmpCsr<IdType, DType, Op, cpu::op::Min<DType>>(
+  //             bcast, csr, ufeat, efeat, out, out_aux[0], out_aux[1]);
+  //     });
+  //   });
+  // }
```
zheng-da, line 80: does the code above not work?

src/array/cpu/spmm.cc
```diff
@@ -40,6 +40,60 @@ void SpMMCsr(const std::string& op, const std::string& reduce,
...
+template <int XPU, typename IdType, int bits>
+void SpMMCsrHetero(const std::string& op, const std::string& reduce,
+             const BcastOff& bcast,
+             const std::vector<CSRMatrix>& vec_csr,
+             std::vector<NDArray> vec_ufeat,
+             std::vector<NDArray> vec_efeat,
+             std::vector<NDArray> vec_out,
+             std::vector<NDArray> out_aux,
+             const std::vector<dgl_type_t> ufeat_eid,
+             const std::vector<dgl_type_t> out_eid) {
```
zheng-da, line 53: the names of the two arguments is kind of confusing. it seems they are eid, but they are actual node type ID of the source nodes and destination nodes, but the length of the vector is the number of edge type.

zheng-da: can we just use something like `ufeat_node_tids` and `out_node_tids`?

src/array/cpu/spmm.cc
```diff
@@ -40,6 +40,60 @@ void SpMMCsr(const std::string& op, const std::string& reduce,
...
+          CSRMatrix csr = vec_csr[etype];
+          NDArray ufeat = (vec_ufeat.size() == 0) ? NullArray() : vec_ufeat[src_id];
+          NDArray efeat = (vec_efeat.size() == 0) ? NullArray() : vec_efeat[etype];
+          NDArray out = vec_out[dst_id];
+          if (reduce == "max")
+            cpu::SpMMCmpCsr<IdType, DType, Op, cpu::op::Max<DType>>(
+                bcast, csr, ufeat, efeat, out, out_aux[0], out_aux[1]);
+          else
+            cpu::SpMMCmpCsr<IdType, DType, Op, cpu::op::Min<DType>>(
+                bcast, csr, ufeat, efeat, out, out_aux[0], out_aux[1]);
```
zheng-da, line 86: it looks the code in the two branches are most or less the same. why do we not use if condition to check `reduce` to decide the template arguments?

src/array/kernel.cc
```diff
@@ -52,6 +52,53 @@ void SpMM(const std::string& op, const std::string& reduce,
...
+        if (format == SparseFormat::kCSC) {
+          SpMMCsrHetero<XPU, IdType, bits>(
+              op, reduce, bcast, vec_graph,
+              ufeat_vec, efeat_vec, out, out_aux,
+              ufeat_eid, out_eid);
+        } //else if (format == SparseFormat::kCOO) {
+        //   SpMMCoo<XPU, IdType, bits>(
+        //       op, reduce, bcast, graph->GetCOOMatrix(0),
+        //       ufeat, vec_efeat, out, out_aux);
+        // }
```
zheng-da, line 92: why is this commented out?

src/array/kernel.cc
```diff
@@ -83,6 +130,50 @@ void SDDMM(const std::string& op,
...
+/*! \brief Generalized Sampled Dense-Dense Matrix Multiplication. */
+void SDDMMHetero(const std::string& op,
+           HeteroGraphPtr graph,
+           std::vector<NDArray> lhs,
+           std::vector<NDArray> rhs,
+           std::vector<NDArray> out,
+           int lhs_target,
+           int rhs_target) {
+  // TODO(Israt): change it to COO_CODE
+  SparseFormat format = graph->SelectFormat(0, CSR_CODE);
```
zheng-da, line 143: this is to change the format of the graph for relation type 0?

src/array/kernel.cc
```diff
@@ -235,6 +326,52 @@ DGL_REGISTER_GLOBAL("sparse._CAPI_DGLKernelSpMM")
...
+    const std::string reduce_op = args[2];
+    List<Value> list_U = args[3];
+    List<Value> list_E = args[4];
+    List<Value> list_V = args[5];
+    NDArray ArgU = args[6];
+    NDArray ArgE = args[7];
+    std::vector<NDArray> U_vec;
+    std::vector<NDArray> V_vec;
+    std::vector<NDArray> E_vec;
+    // CHECK_EQ(list_U.size(), list_V.size());
```
zheng-da, line 342: do you want to keep the check or remove it? let's not comment out code. if you don't want it, you should remove it directly.

src/array/kernel.cc
```diff
@@ -235,6 +326,52 @@ DGL_REGISTER_GLOBAL("sparse._CAPI_DGLKernelSpMM")
...
+      NDArray E = (E_vec.size() == 0) ? NullArray() : E_vec[etype];
+      CheckCtx(graph->Context(), {U, E, V_vec[dst_id], ArgU, ArgE},
+          {"U_data", "E_data", "out", "Arg_U", "Arg_E"});
+      CheckContiguous({U, E, V_vec[dst_id], ArgU, ArgE},
+          {"U_data", "E_data", "out", "Arg_U", "Arg_E"});
+      // CheckShape(
+      //     {graph->NumVertices(src_id), graph->NumEdges(etype), graph->NumVertices(dst_id)},
+      //     {0, 1, 2, 2, 2},
+      //     {U_vec[src_id], E, V_vec[dst_id], ArgU, ArgE},
+      //     {"U_data", "E_data", "out", "Arg_U", "Arg_E"});
```
zheng-da, line 370: the same comment here.

src/array/kernel.cc
```diff
@@ -235,6 +326,52 @@ DGL_REGISTER_GLOBAL("sparse._CAPI_DGLKernelSpMM")
...
+      U_vec.push_back(val->data);
+    }
+    for (Value val : list_V) {
+      V_vec.push_back(val->data);
+    }
+    for (Value val : list_E) {
+      E_vec.push_back(val->data);
+    }
+
+    for (dgl_type_t etype = 0; etype < graph->NumEdgeTypes(); ++etype) {
```
zheng-da, line 356: this format is to verify the metadata of the input graph?

src/array/kernel.cc
isratnisa: We don't support COO for now.

src/array/cpu/spmm.cc
isratnisa: We can. I followed the original code structure.

src/array/kernel.cc
isratnisa: Yes. We don't expect different relation types to have different graph formats. So, checking only the first one should be fine.

src/array/kernel.cc
isratnisa: Do you mean why we are converting from list to vector?

src/array/cpu/spmm.cc
isratnisa: Done!

src/array/kernel.cc
isratnisa: Done!

src/array/cpu/spmm.cc
zheng-da: i don't see the names changed.

src/array/kernel.cc
zheng-da: i don't understand the comment above the line.

zheng-da (approved):

src/array/cpu/spmm.cc
```diff
@@ -18,28 +18,106 @@ void SpMMCsr(const std::string& op, const std::string& reduce,
...
+    });
+  } else if (reduce == "max" || reduce == "min") {
+    SWITCH_BITS(bits, DType, {
+      SWITCH_OP(op, Op, {
+         // TODO(Israt): Ideally the for loop should go over num_ntypes
+        for (dgl_type_t etype = 0; etype < ufeat_node_tids.size(); ++etype) {
+          DType *out_off = vec_out[out_node_tids[etype]].Ptr<DType>();
+          IdType* argX = Op::use_lhs ? static_cast<IdType*>(out_aux[0]->data) : nullptr;
+          IdType* argW = Op::use_rhs ? static_cast<IdType*>(out_aux[1]->data) : nullptr;
+          std::fill(out_off, out_off + dim, cpu::op::Max<DType>::zero);
```
zheng-da, line 94: you will get the same problem as above.

src/array/cpu/spmm.cc
```diff
@@ -18,28 +18,106 @@ void SpMMCsr(const std::string& op, const std::string& reduce,
...
     SWITCH_BITS(bits, DType, {
       SWITCH_OP(op, Op, {
-        if (reduce == "max")
+        DType *out_off = out.Ptr<DType>();
+        IdType* argX = Op::use_lhs ? static_cast<IdType*>(out_aux[0]->data) : nullptr;
+        IdType* argW = Op::use_rhs ? static_cast<IdType*>(out_aux[1]->data) : nullptr;
+        if (Op::use_lhs) std::fill(argX, argX + csr.num_rows * dim, 0);
+        if (Op::use_rhs) std::fill(argW, argW + csr.num_rows * dim, 0);
+        if (reduce == "max") {
+          std::fill(out_off, out_off + csr.num_rows * dim, cpu::op::Max<DType>::zero);
```
zheng-da, line 39: why do you need to do this while the original code doesn't need it?

src/array/kernel_decl.h
```diff
@@ -114,19 +136,18 @@ std::pair<CSRMatrix, NDArray> CSRMM(
 
 /*!
  * \brief Sparse-sparse matrix summation.
- *
- * \param A The sparse matrices with the same size.
- * \param A_weights The weights of each sparse matrix as a 1D tensor.
- *
- * \note GPU implementation will cast the indices to 32 bit.
- * \note The zero entries in the result are not removed.
- * \note The CSR matrix should not have duplicate entries.
```
zheng-da, line 123: why do you remove the doc string here?

src/array/kernel_decl.h
```diff
@@ -96,14 +125,7 @@ void BackwardSegmentCmp(NDArray feat,
...
  * \brief Sparse-sparse matrix multiplication
  *
- * \param A The left operand.
- * \param A_weights The weights of matrix as a 1D tensor.
- * \param B The right operand.
- * \param B_weights The weights of matrix as a 1D tensor.
- *
- * \note GPU implementation will cast the indices to 32 bit.
- * \note The zero entries in the result are not removed.
- * \note The CSR matrix should not have duplicate entries.
```
zheng-da, line 106: here as well.

PR #2887, opened 2021-05-05
[NN] Small bug fix for out_feat = 1 with self loops

Vincentx15:
When using self_loops with out_feat=1, the shapes are as follow : 
node_repr : (graph_size)
loop_message : (graph_size, 1)
So the addition uses broadcasting and the output is of size (graph_size, graph_size)

I just added a check to add an axis when out_feat=1


BarclayII:
Hi @Vincentx15 , would you mind putting how you reproduced the error here?  I checked with `RelGraphConv(3, 1, 2)` and I found the shapes correct.

jermainewang:
Closed due to lack of activity.

PR #2891, opened 2021-05-06
[Feature] add udf support for `cross_reducer`

Secbone:
## Checklist
- [x] Code is well-documented

## Changes
- add udf support for `cross_reducer`


BarclayII:
Is there any reason to use a custom UDF function instead of using `stack` cross-type reducer and process the features later?

Secbone:
@BarclayII I just update the PR, actually, we don't want the `stack` function in the reducing stage because the `stack` function requires the features has the same shape, we can do this in a UDF reduce function.

BarclayII:
I see.  Could you add the signature of the callable function in the docstring?  (Like what are the input arguments and what are the outputs)

Secbone:
@BarclayII I have updated the docstring of the callable function, please have a look

python/dgl/heterograph.py
```diff
@@ -4742,8 +4742,10 @@ def multi_update_all(self, etype_dict, cross_reducer, apply_node_func=None):
                 An optional apply function to further update the node features
                 after the message reduction. It must be a :ref:`apiudf`.
 
-        cross_reducer : str
-            Cross type reducer. One of ``"sum"``, ``"min"``, ``"max"``, ``"mean"``, ``"stack"``.
+        cross_reducer : str or callable function
+            Cross type reducer. One of ``"sum"``, ``"min"``, ``"max"``, ``"mean"``, ``"stack"``
+            or a callable function. If a callable function is provided, the input arguments is
+            a list of tensors from cross types, and the output of function must be a single tensor
```
BarclayII, line 4748:
```suggestion
            or a callable function. If a callable function is provided, the input argument must be
            a single list of tensors containing aggregation results from each edge type, and the
            output of function must be a single tensor.
```

python/dgl/heterograph.py
```diff
@@ -5914,8 +5916,10 @@ def reduce_dict_data(frames, reducer, order=None):
     ----------
     frames : list[dict[str, Tensor]]
         Input tensor dictionaries
-    reducer : str
-        One of "sum", "max", "min", "mean", "stack"
+    reducer : str or callable function
+        One of "sum", "max", "min", "mean", "stack" or a callable function.
+        If a callable function is provided, the input arguments is a list of tensors
+        from cross types, and the output of function must be a single tensor
```
BarclayII, line 5922:
```suggestion
        If a callable function is provided, the input arguments must be a single list
        of tensors containing aggregation results from each edge type, and the
        output of function must be a single tensor.
```

BarclayII (approved):
Looks good.  Minor edits.

PR #2911, opened 2021-05-14
[Feature] Add a use_ddp flag to PyTorch Node/Edge/GraphDataLoaders for multi-GPU training

### LINKED ISSUE #2539: Current Node/EdgeDataLoader usage in examples may yield different number of batches per epoch for multiple GPUs.
## 🐛 Bug

Currently the DataLoaders for multiple GPUs works like this:

1. Divide the list of training seed nodes into the same number of chunks as the number of GPUs using `torch.split()` (`torch.chunk()` triggers the same bug):
   ```
   train_nid = th.split(train_nid, math.ceil(len(train_nid) / n_gpus))[proc_id]
   ```
2. Iterate over the list of training nodes with `Node/EdgeDataLoader`, which iterates over the list of seed nodes using `DataLoader` with a given batch size.

This method sometimes will yield an uneven number of batches for each GPU.  For instance, given 4097 nodes and 4 GPUs, a batch size of 1024 will yield 2 batches (one complete batch and another batch with only one element) for 3 GPUs, and only one batch (a single batch with 1022 elements) for another GPU.  With `DistributedDataParallel` that requires gradient from all 4 workers, this will cause a deadlock.

## To Reproduce

1. Run `examples/pytorch/graphsage/train_sampling_multi_gpu.py`, but only keep the first few 4001 training nodes:
   ```python
   new_train_nids = g.ndata['train_mask'].nonzero(as_tuple=True)[0][:4001]
   g.ndata['train_mask'].zero_()
   g.ndata['train_mask'][new_train_nids] = True
   ```
   You will notice that the program hangs with all 4 GPUs' utilization at 100%. 

## Expected behavior

At least the code should not freeze.

## Environment

 - OS (e.g., Linux):
 - Python version:

## Additional context

PyTorch's DataLoader and DDP also has this problem, and they have a `torch.utils.data.distributed.DistributedSampler` object to handle that.  See also https://yangkky.github.io/2019/07/08/distributed-pytorch-tutorial.html.  `DistributedSampler` addresses the problem by cycling to the head of the dataset:
```python
# If the dataset length is evenly divisible by # of replicas, then there
# is no need to drop any data, since the dataset will be split equally.
if self.drop_last and len(self.dataset) % self.num_replicas != 0:  # type: ignore
    # Split to nearest available length that is evenly divisible.
    # This is to ensure each rank receives the same amount of data when
    # using this Sampler.
    self.num_samples = math.ceil(
        # `type:ignore` is required because Dataset cannot provide a default __len__
        # see NOTE in pytorch/torch/utils/data/sampler.py
        (len(self.dataset) - self.num_replicas) / self.num_replicas  # type: ignore
    )
else:
    self.num_samples = math.ceil(len(self.dataset) / self.num_replicas)  # type: ignore
self.total_size = self.num_samples * self.num_replicas
# ...
# add extra samples to make it evenly divisible
padding_size = self.total_size - len(indices)
if padding_size <= len(indices):
    indices += indices[:padding_size]
else:
    indices += (indices * math.ceil(padding_size / len(indices)))[:padding_size]
```

Should be an easy fix for single machine training (@jermainewang we may need two additional optional arguments `num_replicas` and `rank`, following `DistributedSampler`).  However, I'm not sure what changes are needed in `DistDataLoader` (@zheng-da @VoVAllen @aksnzhy).

Credit to @nv-dlasalle

BarclayII:
This PR adds a new flag `use_ddp` to the DataLoaders to handle shuffling and splitting using PyTorch's `DistributedSampler`.  Otherwise deadlocks might occur under some particular configurations of number of GPUs and batch size.

Also fixes a bunch of example models to use PyTorch's `DistributedSampler`.

Fixes #2539.

### DIFF
--- a/python/dgl/dataloading/pytorch/__init__.py
+++ b/python/dgl/dataloading/pytorch/__init__.py
@@ -2,6 +2,7 @@
 import inspect
 import torch as th
 from torch.utils.data import DataLoader
+from torch.utils.data.distributed import DistributedSampler
 from ..dataloader import NodeCollator, EdgeCollator, GraphCollator
 from ...distributed import DistGraph
 from ...distributed import DistDataLoader
@@ -272,6 +273,12 @@ class NodeDataLoader:
     device : device context, optional
         The device of the generated MFGs in each iteration, which should be a
         PyTorch device object (e.g., ``torch.device``).
+    use_ddp : boolean, optional
+        If True, tells the DataLoader to split the training set for each
+        participating process appropriately using
+        :mod:`torch.utils.data.distributed.DistributedSampler`.
+
+        Overrides the :attr:`sampler` argument of :class:`torch.utils.data.DataLoader`.
     kwargs : dict
         Arguments being passed to :py:class:`torch.utils.data.DataLoader`.
 
@@ -288,6 +295,21 @@ class NodeDataLoader:
     >>> for input_nodes, output_nodes, blocks in dataloader:
     ...     train_on(input_nodes, output_nodes, blocks)
 
+    **Using with Distributed Data Parallel**
+
+    If you are using PyTorch's distributed training (e.g. when using
+    :mod:`torch.nn.parallel.DistributedDataParallel`), you can train the model by turning
+    on the `use_ddp` option:
+
+    >>> sampler = dgl.dataloading.MultiLayerNeighborSampler([15, 10, 5])
+    >>> dataloader = dgl.dataloading.NodeDataLoader(
+    ...     g, train_nid, sampler, use_ddp=True,
+    ...     batch_size=1024, shuffle=True, drop_last=False, num_workers=4)
+    >>> for epoch in range(start_epoch, n_epochs):
+    ...     dataloader.set_epoch(epoch)
+    ...     for input_nodes, output_nodes, blocks in dataloader:
+    ...         train_on(input_nodes, output_nodes, blocks)
+
     Notes
     -----
     Please refer to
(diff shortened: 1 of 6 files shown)


python/dgl/dataloading/pytorch/__init__.py
```diff
@@ -347,10 +369,21 @@ def __init__(self, g, nids, block_sampler, device='cpu', **kwargs):
                         dataloader_kwargs['shuffle'] = False
                         dataloader_kwargs['drop_last'] = False
 
+            self.use_ddp = use_ddp
+            if use_ddp:
+                self.dist_sampler = DistributedSampler(
+                    dataset,
+                    shuffle=dataloader_kwargs['shuffle'],
+                    drop_last=dataloader_kwargs['drop_last'])
+                dataloader_kwargs['shuffle'] = False
```
VoVAllen, line 378: What's the point of resetting kwargs here?

python/dgl/dataloading/pytorch/__init__.py
```diff
@@ -553,12 +629,27 @@ def __init__(self, g, eids, block_sampler, device='cpu', **kwargs):
...
                 'EdgeDataLoader does not support DistGraph for now. ' \
                 + 'Please use DistDataLoader directly.'
+
+        self.use_ddp = use_ddp
+        if use_ddp:
+            self.dist_sampler = DistributedSampler(
+                dataset,
+                shuffle=dataloader_kwargs['shuffle'],
+                drop_last=dataloader_kwargs['drop_last'])
+            dataloader_kwargs['shuffle'] = False
```
VoVAllen, line 644: Same question

VoVAllen (approved):

python/dgl/dataloading/pytorch/__init__.py
BarclayII: PyTorch spec requires that `shuffle` must not be specified (i.e. False) when `sampler` is specified.

PR #2918, opened 2021-05-17
[Distributed] add distributed in-degree and out-degree.

zheng-da:
## Description
Add `in_degrees` and `out_degrees` in `DistGraph`.
This PR also fixes the issue that the out-degree computed from the partitioned graph is incorrect. To fix this issue, we need to add out-going edges between partitions on the source nodes.

## Checklist
- [ ] Code is well-documented


python/dgl/distributed/dist_graph.py
```diff
@@ -745,6 +747,104 @@ def num_edges(self, etype=None):
                 return sum([self._gpb._num_edges(etype) for etype in self.etypes])
         return self._gpb._num_edges(etype)
 
+    def out_degrees(self, u=ALL):
+        """Return the out-degree(s) of the given nodes.
+
+        It computes the out-degree(s) w.r.t. to the edges of the given edge type.
```
classicsong, line 753: So where is the edge type?

src/graph/transform/partition_hetero.cc
```diff
@@ -98,6 +98,32 @@ HaloHeteroSubgraph GetSubgraphWithHalo(std::shared_ptr<HeteroGraph> hg,
     }
   }
 
+  if (num_hops > 0) {
+    EdgeArray out_edges = hg->OutEdges(0, nodes);
```
classicsong, line 102: This is going to add halo nodes for out edges?

classicsong (approved):
A general question: when we add out-edges in each partitation for out_degree(), what is the memory and storage cost of doing so?

Overall, the PR LGTM.

src/graph/transform/partition_hetero.cc
zheng-da: yes

zheng-da:
this will save the out-edges that are cut by the partitioning algorithm with a partition. if METIS has many edge cuts, it shouldn't cause too much edge duplication.

classicsong (approved):
LGTM

Issue #2921, opened 2021-05-17
GATConv mismatched input and output dimensions when setting residual to True and using multiple heads


xieweiyi:
## 🐛 Bug
[In GATConv code](https://github.com/dmlc/dgl/blob/master/python/dgl/nn/pytorch/conv/gatconv.py)
```
        if residual:
            if self._in_dst_feats != out_feats:
                self.res_fc = nn.Linear(
                    self._in_dst_feats, num_heads * out_feats, bias=False)
            else:
                self.res_fc = Identity()
```
The output dimension is num_heads * out_feats. however, the code is only checking if the input dimension equals 'out_feats'.

## To Reproduce

1. setting num_heads > 1 and residual to True, e.g. num_heads=2
1. setting in_feats = out_feats. e.g, all set to 8.

`RuntimeError: shape '[1, 2, 8]' is invalid for input of size 8`

## Environment

 - DGL Version (e.g., 1.0): dgl-cuda10.1 0.6.1
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):
 - OS (e.g., Linux): Linux
 - How you installed DGL (`conda`, `pip`, source): pip install dgl-cu101
 - Python version: 3.6
 - CUDA/cuDNN version (if applicable): 10.1

PR #2922, opened 2021-05-17
Fixing issue #2921

xieweiyi:
A linear layer (res_fc) is used to reshape features to the shape of the output feature of GATConv (in dimensions of out_feats * num_heads) for the residual connection. However, the previous conditional check failed to cover this.


BarclayII (approved):

BarclayII:
Thanks!

PR #2925, opened 2021-05-19
Add heterograph support in CUDA kernels (SpMM, SDDMM)

isratnisa:
## Description
Added heterograph support in CUDA kernels for SpMM and SDDMM operators. Functions are named as *Hetero for now to distinguish from the original codes for homogeneous graphs. Eventually both functions will be merged to one.  Different relation types of the heterograph are handled inside the cuda kernels. Kernels for relation types can be launched sequentially or in parallel using CUDA streams. This is an initial solution.

## Checklist
- [ ] Code is well-documented


src/array/cuda/sddmm.cu
```diff
@@ -91,6 +91,43 @@ void SDDMMCsr(const std::string& op,
...
+              std::vector<NDArray> vec_rhs,
+              std::vector<NDArray> vec_out,
+              int lhs_target,
+              int rhs_target,
+              const std::vector<dgl_type_t> lhs_eid,
+              const std::vector<dgl_type_t> rhs_eid) {
+  int maxstrm = 16;
+  cudaStream_t stream[maxstrm];
+  for (int i = 0; i < maxstrm; i++)
+    cudaStreamCreate(&(stream[i]));
```
nv-dlasalle, line 112: These streams should be destroyed at the end of this function, and this call should be wrapped in CUDA_CALL() in case it fails.

src/array/cuda/sddmm.cu
```diff
@@ -91,6 +91,43 @@ void SDDMMCsr(const std::string& op,
...
+void SDDMMCsrHetero(const std::string& op,
+              const BcastOff& bcast,
+              const std::vector<CSRMatrix>& vec_csr,
+              std::vector<NDArray> vec_lhs,
+              std::vector<NDArray> vec_rhs,
+              std::vector<NDArray> vec_out,
+              int lhs_target,
+              int rhs_target,
+              const std::vector<dgl_type_t> lhs_eid,
+              const std::vector<dgl_type_t> rhs_eid) {
```
nv-dlasalle, line 108: We may want to pass all of these `std::vector`s as `const std::vector<>&`, since we don't want to copy them.

src/array/cuda/sddmm.cu
```diff
@@ -91,6 +91,43 @@ void SDDMMCsr(const std::string& op,
   });
 }
 
+/*!
+ * \brief CUDA implementation of g-SDDMM on heterograph using 
+    Csr format.
+ */
+template <int XPU, typename IdType, int bits>
```
nv-dlasalle, line 98: Does bits need to be a template parameter? Can we just parse it from `vec_lhs`, `vec_rhs`, `vec_out` or `vec_csr`?

src/array/cuda/spmm.cu
```diff
@@ -282,6 +282,107 @@ void CusparseCsrmm2(
...
+  // allocate cusparse handle if needed
+  if (!thr_entry->cusparse_handle) {
+    CUSPARSE_CALL(cusparseCreate(&(thr_entry->cusparse_handle)));
+  }
+  CUSPARSE_CALL(cusparseSetStream(thr_entry->cusparse_handle, strm_id));
+  // all one data array
+  DType* valptr = nullptr;
+  if (!A_data) {
+    valptr = static_cast<DType*>(device->AllocWorkspace(ctx, nnz * sizeof(DType)));
+    _Fill(valptr, nnz, static_cast<DType>(1.));
```
nv-dlasalle, line 320: This kernel is launching on the default stream rather than strm_id.

src/array/cuda/spmm.cu
```diff
@@ -282,6 +282,107 @@ void CusparseCsrmm2(
...
+  auto* thr_entry = runtime::CUDAThreadEntry::ThreadLocal();
+  // allocate cusparse handle if needed
+  if (!thr_entry->cusparse_handle) {
+    CUSPARSE_CALL(cusparseCreate(&(thr_entry->cusparse_handle)));
+  }
+  CUSPARSE_CALL(cusparseSetStream(thr_entry->cusparse_handle, strm_id));
+  // all one data array
+  DType* valptr = nullptr;
+  if (!A_data) {
+    valptr = static_cast<DType*>(device->AllocWorkspace(ctx, nnz * sizeof(DType)));
```
nv-dlasalle, line 319: Using a workspace allocation on stream different than the one associated with that context (which currently is always stream 0), could potentially cause race conditions between streams, as the workspace will get freed before the GPU is done with it, and it may be given to a different stream while the other stream is still using it.

nv-dlasalle: In cuda 11.2 there is a stream based allocator which would solve this issue nicely, but for now I think you'll have to allocate all of the memory before creating these streams, and free it only after destroying the streams.

src/array/cuda/spmm.cu
```diff
@@ -282,6 +282,107 @@ void CusparseCsrmm2(
...
+      dtype, CUSPARSE_SPMM_CSR_ALG2,
+      workspace));
+  device->FreeWorkspace(ctx, workspace);
+
+  CUSPARSE_CALL(cusparseDestroySpMat(matA));
+  CUSPARSE_CALL(cusparseDestroyDnMat(matB));
+  CUSPARSE_CALL(cusparseDestroyDnMat(matC));
+#else
+  // allocate matrix for temporary transposed output
+  DType* trans_out = static_cast<DType*>(device->AllocWorkspace(ctx, m * n * sizeof(DType)));
```
nv-dlasalle, line 362: Same issue here with allocations potentially used by multiple streams.

src/array/cuda/spmm.cu
```diff
@@ -282,6 +282,107 @@ void CusparseCsrmm2(
...
+      CUSPARSE_OPERATION_NON_TRANSPOSE,
+      CUSPARSE_OPERATION_TRANSPOSE,
+      m, n, k, nnz, &alpha,
+      descr, (valptr)? valptr : A_data,
+      static_cast<int32_t*>(csr.indptr->data),
+      static_cast<int32_t*>(csr.indices->data),
+      B_data, n, &beta, trans_out, m));
+  CUSPARSE_CALL(cusparseDestroyMatDescr(descr));
+  // transpose the output matrix
+  _Transpose(trans_out, C_data, n, m);
```
nv-dlasalle, line 379: This is running on the default stream, rather than strm_id.

src/array/cuda/spmm.cu
```diff
@@ -400,6 +501,109 @@ void SpMMCsr(const std::string& op, const std::string& reduce,
...
+template <int XPU, typename IdType, int bits>
+void SpMMCsrHetero(const std::string& op, const std::string& reduce,
+             const BcastOff& bcast,
+             const std::vector<CSRMatrix>& vec_csr,
+             std::vector<NDArray> vec_ufeat,
+             std::vector<NDArray> vec_efeat,
+             std::vector<NDArray> vec_out,
+             std::vector<NDArray> out_aux,
+             const std::vector<dgl_type_t> ufeat_eid,
+             const std::vector<dgl_type_t> out_eid) {
```
nv-dlasalle, line 518: Same thing here about passing the vectors by value instead of `const std;:vector<>&`.

src/array/cuda/spmm.cu
```diff
@@ -400,6 +501,109 @@ void SpMMCsr(const std::string& op, const std::string& reduce,
   }
 }
 
+/*!
+ * \brief CUDA implementation of g-SpMM on Csr format.
+ * \note use cusparse if the reduce operator is `sum` and there is
+ *       no broadcast, use dgl's kernel in other cases.
+ */
+template <int XPU, typename IdType, int bits>
```
nv-dlasalle, line 509: Again, can we infer the bits from the input arrays, rather templating on it?

src/array/cuda/spmm.cu
```diff
@@ -400,6 +501,109 @@ void SpMMCsr(const std::string& op, const std::string& reduce,
...
+             const std::vector<dgl_type_t> ufeat_eid,
+             const std::vector<dgl_type_t> out_eid) {
+  int64_t feat_len = bcast.out_len;
+  bool is_scalar_efeat = vec_efeat.size() != 0; //efeat.NumElements() == csr.indices->shape[0];
+  bool use_efeat = op != "copy_lhs";
+  const DLContext& ctx = DLContext{kDLGPU, 0};
+  int maxstrm = 16;
+  cudaStream_t stream[maxstrm];
+  for (int i = 0; i < 16; i++)
+    cudaStreamCreate(&(stream[i]));
```
nv-dlasalle: Are these streams getting destroyed?

src/array/cuda/spmm.cu
```diff
@@ -400,6 +501,109 @@ void SpMMCsr(const std::string& op, const std::string& reduce,
...
+             std::vector<NDArray> out_aux,
+             const std::vector<dgl_type_t> ufeat_eid,
+             const std::vector<dgl_type_t> out_eid) {
+  int64_t feat_len = bcast.out_len;
+  bool is_scalar_efeat = vec_efeat.size() != 0; //efeat.NumElements() == csr.indices->shape[0];
+  bool use_efeat = op != "copy_lhs";
+  const DLContext& ctx = DLContext{kDLGPU, 0};
+  int maxstrm = 16;
+  cudaStream_t stream[maxstrm];
+  for (int i = 0; i < 16; i++)
```
nv-dlasalle: should this 16 be `maxstrm`?

src/array/cuda/spmm.cu
```diff
@@ -400,6 +501,109 @@ void SpMMCsr(const std::string& op, const std::string& reduce,
...
+             std::vector<NDArray> vec_ufeat,
+             std::vector<NDArray> vec_efeat,
+             std::vector<NDArray> vec_out,
+             std::vector<NDArray> out_aux,
+             const std::vector<dgl_type_t> ufeat_eid,
+             const std::vector<dgl_type_t> out_eid) {
+  int64_t feat_len = bcast.out_len;
+  bool is_scalar_efeat = vec_efeat.size() != 0; //efeat.NumElements() == csr.indices->shape[0];
+  bool use_efeat = op != "copy_lhs";
+  const DLContext& ctx = DLContext{kDLGPU, 0};
```
nv-dlasalle, line 522: We should be getting the device from the input parameters (otherwise it won't work in a multi-GPU setting).

src/array/cuda/spmm.cu
```diff
@@ -400,6 +501,109 @@ void SpMMCsr(const std::string& op, const std::string& reduce,
...
+          for (int i = 1; i < nd_ufeat->ndim; ++i){
+            x_length *= nd_ufeat->shape[i]; //TODO:: All ufeat shares same shape?
+          }
+          cusparse::CusparseCsrmm2Hetero<DType, IdType>(
+              ctx, csr,
+              static_cast<DType*>(vec_ufeat[src_id]->data),
+              nullptr,
+              static_cast<DType*>(vec_out[dst_id]->data),
+              x_length,
+              stream[etype % maxstrm]);
```
nv-dlasalle, line 547: This seems to be the only operation here that's running on a non-default stream. Shouldn't everything in here run on stream[etype % maxstrm]?

nv-dlasalle (changes requested):

src/array/cuda/spmm.cu
isratnisa: Allocating memory for all the streams might cause OOM. Is there a way to keep it flexible? For example, if there's not enough resource, limit the number of streams and only allocate memory for those streams?

src/array/cuda/spmm.cu
isratnisa: Not sure how to choose the right max stream.

src/array/cuda/spmm.cu
nv-dlasalle: Right now, I don't think there is. You would need to know whether the WorkspacePool has an allocation that can be used, or a new one would need to be made. Same thing with the TensorAdapter.

In general though, I think restricting parallelism based on the amount of available memory will be problematic, both in terms of code complexity and performance. I think OOM error, and forcing the user to choose a different batch size/configuration, is better than running slower. We could maybe make an environment variable to control the maximum number of stream, and thus let the user control the amount of parallelism. But this would still require stream-based allocations to limit the concurrent memory usage.

jermainewang:
@nv-dlasalle I feel we could leave the multi-stream optimization to future so to unblock this PR. @isratnisa for now, just use the default stream to avoid possible data race issues.

src/array/cuda/sddmm.cu
nv-dlasalle: Sounds good.

src/array/cuda/spmm.cu
```diff
@@ -400,6 +501,117 @@ void SpMMCsr(const std::string& op, const std::string& reduce,
...
+             const std::vector<dgl_type_t>& out_eid) {
+  int64_t feat_len = bcast.out_len;
+  bool is_scalar_efeat = vec_efeat.size() != 0;
+  bool use_efeat = op != "copy_lhs";
+  // TODO(Israt): 1:Resolve PR-https://github.com/dmlc/dgl/issues/2995
+  // to use maxstream > 1
+  int maxstrm = 1;
+  std::vector<cudaStream_t> stream(maxstrm);
+  for (int i = 0; i < maxstrm; i++) {
+    CUDA_CALL(cudaStreamCreate(&(stream[i])));
```
nv-dlasalle, line 527: We still shouldn't create a stream here, as we could run into a race condition with the default stream (technically I think the default stream should block around it by default, but I'd consider it dangerous). I think for now, it would be better to do:
```
const int maxstrm = 1;
std::vector<cudaStream_t> stream{static_cast<cudaStream_t>(0)};
```
Such that everything will be run on the default stream.

src/array/cuda/spmm.cu
```diff
@@ -282,6 +282,107 @@ void CusparseCsrmm2(
...
+
+/*! Cusparse implementation of SpMM on Csr format. */
+template <typename DType, typename IdType>
+void CusparseCsrmm2Hetero(
+    const DLContext& ctx,
+    const CSRMatrix& csr,
+    const DType* B_data, const DType* A_data,
+    DType* C_data,
+    int x_length,
+    cudaStream_t strm_id) {
```
nv-dlasalle, line 294: Since with the current implementation, this function will only work correctly if this is the default stream, I think it would be good to add something like:
```
CHECK_EQ(strm_id, 0) << "CusparseCsrmm2Hetero is only implemented for the default stream currently.";
```

src/array/cuda/spmm.cu
```diff
@@ -282,6 +282,107 @@ void CusparseCsrmm2(
...
+  // We use csrmm2 to perform following operation:
+  // C = A x B, where A is a sparse matrix in csr format, B is the dense matrix for node
+  // feature tensor. However, since cusparse only supports column-major, while our tensor
+  // is stored in row-major, the actual computation is:
+  // C = trans(A x trans(B)).
+  // Currently, we use cublasXgeam to implement transposition and allocate intermediate
+  // workspace memory for this.
+  const int m = csr.num_rows;
```
nv-dlasalle, line 302: Because things like csr.num_rows and csr.indices->shape[0] are int64_t, I think we should have some checks that we are not exceeding `std::numeric_limits<int>::max` before performing the conversion:
```
CHECK_GE(std::numeric_limits<int>::max, csr.num_rows);
CHECK_GE(std::numeric_limits<int>::max, csr.num_cols);
CHECK_GE(std::numeric_limits<int>::max, csr.indices->shape[0]);
```

src/array/cuda/spmm.cu
```diff
@@ -282,6 +282,107 @@ void CusparseCsrmm2(
...
+  CUSPARSE_CALL(cusparseSetMatType(descr, CUSPARSE_MATRIX_TYPE_GENERAL));
+  CUSPARSE_CALL(cusparseSetMatIndexBase(descr, CUSPARSE_INDEX_BASE_ZERO));
+  CUSPARSE_CALL(Xcsrmm2<DType>(
+      thr_entry->cusparse_handle,
+      CUSPARSE_OPERATION_NON_TRANSPOSE,
+      CUSPARSE_OPERATION_TRANSPOSE,
+      m, n, k, nnz, &alpha,
+      descr, (valptr)? valptr : A_data,
+      static_cast<int32_t*>(csr.indptr->data),
+      static_cast<int32_t*>(csr.indices->data),
```
nv-dlasalle, line 375: Is this safe? If `IdType == int64_t`, do we need to perform a `AsBits()` call?

src/array/cuda/spmm.cu
```diff
@@ -282,6 +282,107 @@ void CusparseCsrmm2(
...
 }
+
+/*! Cusparse implementation of SpMM on Csr format. */
+template <typename DType, typename IdType>
+void CusparseCsrmm2Hetero(
+    const DLContext& ctx,
+    const CSRMatrix& csr,
+    const DType* B_data, const DType* A_data,
+    DType* C_data,
+    int x_length,
```
nv-dlasalle, line 293: This should probably be of type `int64_t` so we can check its size internally.

nv-dlasalle (changes requested):

src/array/cuda/spmm.cu
isratnisa: Line 293?

src/array/cuda/spmm.cu
isratnisa: @BarclayII You might know this piece of code pretty well. I borrowed it from homogeneous counterpart. What's your opinion on this?

nv-dlasalle (approved):
Aside from the potential `int32_t*` conversion issue in CUDA 10, LGTM.

PR #2929, opened 2021-05-20
[API]  Standardize Subgraph APIs

mufeili:
## Description
- Provide a standard `store_ids` argument whenever nodes/edges can be re-labeled in subgraph extraction
- Change `g` to `graph` for standardization

## Checklist
- [x] Code is well-documented


VoVAllen (approved):
LGTM. Only question is why you want to change the `_create_hetero_subgraph` function signature, since it didn't change the behavior?

python/dgl/subgraph.py
```diff
@@ -296,11 +294,12 @@ def _process_edges(etype, e):
...
     sgi = graph._graph.edge_subgraph(induced_edges, preserve_nodes)
     induced_nodes = sgi.induced_nodes
-    return _create_hetero_subgraph(graph, sgi, induced_nodes, induced_edges, store_ids)
+    return _create_hetero_subgraph(graph, sgi, induced_nodes, induced_edges,
+                                   store_nids=store_ids, store_eids=store_ids)
 
 DGLHeteroGraph.edge_subgraph = utils.alias_func(edge_subgraph)
 
-def in_subgraph(g, nodes):
+def in_subgraph(graph, nodes, store_ids=True):
```
jermainewang, line 302: * Should the default be true or false?
* Should you have two options `store_nids` and `store_eids`?

mufeili:
> LGTM. Only question is why you want to change the `_create_hetero_subgraph` function signature, since it didn't change the behavior?

Because not in all cases we want to store both the original node IDs and edge IDs. For example, a subgraph extracted with `in_subgraph` has the same set of nodes as the original graph whether they are isolated or not.

python/dgl/subgraph.py
mufeili: - I set the default to be `True` because storing the original IDs was the default behavior before and this is also consistent with other subgraph extraction APIs like `node_subgraph` and `edge_subgraph`.
- Storing the original node IDs seems to be redundant for `in_subgraph` as a subgraph extracted with this API will have the same set of nodes as the original graph, regardless of whether some nodes will be isolated in the subgraph.

jermainewang: > Storing the original node IDs seems to be redundant for in_subgraph as a subgraph extracted with this API will have the same set of nodes as the original graph, regardless of whether some nodes will be isolated in the subgraph. This is consistent with the behavior before.

This behavior is not consistent with `dgl.subgraph`. I think this PR will fix this kind of discrepancy.

mufeili: You mean having the same set of nodes as the original graph?

mufeili:
@jermainewang 
1. `node_subgraph`, `edge_subgraph`, `in_subgraph`, `out_subgraph` now have standardized arguments as we discussed.
2. `relabel_nodes=False` is not supported yet for `node_subgraph` and an error will be raised in `VertexSubgraph`.
3. We have replaced `preserve_nodes` with `relabel_nodes`, raised a deprecation error and modified the code using `preserve_nodes=True` with `edge_subgraph`.
4. As we discussed, if `relabel_nodes=False`, the original node IDs will not be stored. Note that this will be a breaking change for `edge_subgraph`. Also cc @BarclayII in case you have any concerns.
5. For `in_subgraph` and `out_subgraph`, the isolated nodes will be removed if `relabel_nodes=True`. Alternatively, we can also preserve the specified destination/source nodes even if they are isolated. 
6. To implement `InEdgeGraphRelabelNodes` and `OutEdgeGraphRelabelNodes`, we first preprocess induced edges and then use `EdgeSubgraph`.

VoVAllen:
@dgl-bot run r5.12xlarge subgraph dmlc/master

python/dgl/subgraph.py
```diff
@@ -276,7 +274,10 @@ def edge_subgraph(graph, edges, preserve_nodes=False, store_ids=True):
     --------
     node_subgraph
     """
-    if graph.is_block and not preserve_nodes:
+    if len(deprecated_kwargs) != 0:
+        raise DGLError("Key word argument preserve_nodes is deprecated. "
+                       "Use relabel_nodes instead.")
```
jermainewang, line 279: Suggest raise deprecation warning for this release to have a smoother transition.

python/dgl/subgraph.py
```diff
@@ -294,44 +295,47 @@ def _process_edges(etype, e):
...
-    An edge-induced subgraph is equivalent to creating a new graph
-    with the same number of nodes using the given edges.  In addition to extracting
-    the subgraph, DGL conducts the following:
-
-    * Copy the features of the extracted nodes and edges to the resulting graph.
-      The copy is *lazy* and incurs data movement only when needed.
-
-    * Store the IDs of the extracted edges in the ``edata``
-      of the resulting graph under name ``dgl.EID``.
+    An edge-induced subgraph is equivalent to creating a new graph using the given edges.
```
jermainewang, line 308: edge-induced?

python/dgl/subgraph.py
```diff
@@ -386,47 +405,43 @@ def in_subgraph(g, nodes):
...
```
jermainewang, line 433: wrong description

src/graph/heterograph_capi.cc
```diff
@@ -379,6 +379,8 @@ DGL_REGISTER_GLOBAL("heterograph_index._CAPI_DGLHeteroVertexSubgraph")
 .set_body([] (DGLArgs args, DGLRetValue* rv) {
     HeteroGraphRef hg = args[0];
     List<Value> vids = args[1];
+    bool relabel_nodes = args[2];
+    CHECK(relabel_nodes) << "Only node relabeling is supported.";
```
jermainewang, line 383: `"Node subgraph currently only supports relabel_nodes=True.`

mufeili:
@jermainewang I've addressed the comments above. `InEdgeGraphRelabelNodes` and `OutEdgeGraphRelabelNodes` now depend on `EdgeSubgraph`.

jermainewang (approved):

PR #2930, opened 2021-05-21
[Performance] Only copy and slice index's in subcolumn when data is used.

nv-dlasalle:
## Description
When message passing is performed on a subgraph, an edge subgraph is created from that, which results in calls to `Column.subcolumn()`. This results in the node ids of the subgraph, being copied form the GPU to the CPU, and then a slice being performed on the CPU. However, the resulting `Column` object is never used.

In the below picture, I ran the following to generate a Nsight-Systems profile from `/examples/pytorch/rgcn/` which highlights the issue:
```
python3 entity_classify_mp.py --num-workers=0 --low-mem --batch-size=4096 --fanout=16,16 --n-hidden=64 --dgl-sparse --dataset ogbn-mag
```
rgcn_slice_index

This PR changes the `Column` class, to allow lazy evaluation of multiple indexing/slicing. Below is the same run, but with this change:
rgcn_lazy_index

This results in a significant performance improvement in the forward pass (about 2x faster for this highlighted region of the forward pass).

## Checklist
- [x] Code is well-documented

## Changes
Added the `_LazyIndex` object to allow layering multiple index/slice operations without evaluating them.

Used the `_LazyIndex` class inside of `Column`, when `subcolumn` is called on a Column which already has an index.

Adds a unit test in `test_frame.py`.


BarclayII:
> This results in the node ids of the subgraph, being copied form the GPU to the CPU

This is very strange.

BarclayII:
I tried profiling the RGCN code with `pyinstrument `but I could not find either the `subcolumn` call nor the subgraph creation you have mentioned in the forward pass (with the same command as you provided).  Could you locate the place where the subgraph is created?

VoVAllen:
Do we need to check in this for v0.7?

nv-dlasalle:
@BarclayII 
> I tried profiling the RGCN code with `pyinstrument `but I could not find either the `subcolumn` call nor the subgraph creation you have mentioned in the forward pass (with the same command as you provided). Could you locate the place where the subgraph is created?

I was mistaken, it's not a subgraph, but an explicit subframe here https://github.com/dmlc/dgl/blob/master/python/dgl/core.py#L81, which is on top of the subframe taken for the DGLBlock's construction during sampling. Since the merging of other PRs, you have to explicitly force the sampling to happen on the CPU to replicate it (in this case by having a worker):
```
python3 entity_classify_mp.py --num-workers=1 --low-mem --batch-size=4096 --fanout=16,16 --n-hidden=64 --dgl-sparse --dataset ogbn-mag --n-epochs 1
```

I've added nvtx markers to highlight where it's taking place in these screenshots:
updated_lazy_frame

On the CPU, `heterograph.update_all` is ~10x faster due to not having to synchronize, and on the GPU it is ~4x faster.

BarclayII:
I did my own stepping and here is what I found:
* After fetching the blocks from the dataloader, each column will have its storage the same as the original features on CPU, as well as indices on CPU.  The only exception is the `dgl.NID` column which does not have an index.
* In the model implementation, there is a `to('cuda')` call, which makes the device of all columns as GPU.  The storages are still on CPU at this point.
* In the `RelGraphConv` layer, after performing an edge subgraph, the feature storage are still on CPU and no actual data transfer / feature slicing occur.
* With custom message function, the `invoke_edge_udf` will extract a subframe for the node features, with a GPU index yielded by the `edges()` method (to find the source and destination nodes) and GPU storage for new feature computed, but CPU storage for old columns that already exist in the original graph (including `_ID` and `_TYPE`).  The subframe creation will iterate over all the columns, even for `_ID` and `_TYPE` as well as the features from the original graph.  If there are indices in the parent column, DGL will copy the GPU subframe row IDs to CPU where the parent column indices resides, and perform a slicing, even if they are actually never used.

Similar problem should also exist for `apply_edges` on a GPU subgraph created from a CPU parent graph.

Popping out the old node features should clearly resolve the problem.  Though I feel there should be a simpler way handling this...

BarclayII (approved):

PR #2934, opened 2021-05-22
[Bugfix] Include NCCL as a submodule

### LINKED ISSUE #2933: NCCL dependency is not handled well.
## 🐛 Bug

After merging #2825, when building on a system without NCCL development headers and the flag `-DUSE_CUDA=ON`, the result is:
```
...
-- Build with OpenMP.
-- Build with AVX optimization.
-- -fopenmp -O2 -Wall -fPIC -std=c++11  -DUSE_AVX -DIDXTYPEWIDTH=64 -DREALTYPEWIDTH=32
-- Running GPU architecture autodetection
nvcc warning : The 'compute_35', 'compute_37', 'compute_50', 'sm_35', 'sm_37' and 'sm_50' architectures are deprecated, and may be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).
-- Found CUDA arch 7.5 7.5 7.5 7.5
-- CUDA flags: -Xcompiler ,-fopenmp,-O2,-Wall,-fPIC,,,-DUSE_AVX,-DIDXTYPEWIDTH=64,-DREALTYPEWIDTH=32;--expt-relaxed-constexpr;-gencode;arch=compute_75,code=sm_75;--expt-extended-lambda;-Wno-deprecated-declarations;-std=c++14
CMake Error at cmake/util/FindNccl.cmake:61 (file):
  file STRINGS file
  "/home/ubuntu/workspace/dgl/NCCL_INCLUDE_DIR-NOTFOUND/nccl.h" cannot be
  read.
Call Stack (most recent call first):
  CMakeLists.txt:162 (include)
```

## To Reproduce

1. Build dgl on a system without NCCL in the system path while enabling `USE_CUDA`:
```
cmake .. -DCMAKE_VERBOSE_MAKEFILE=1 -DUSE_CUDA=ON
```

## Expected behavior

At the very least the error message should be improved (the user needs to install NCCL and/or specify NCCL_ROOT), we could also make NCCL support toggle-able, but this may increase the chance someone would mis-configure and get poor performance when evaluating DGL.

nv-dlasalle:
## Description
Include NCCL as a submodule, and by default use that. This fixes #2933.

## Checklist
- [X] Related issue is referred in this PR

## Changes
This adds NCCL as a submodule.

Adds a cmake module for building NCCL from source.

This adds the option USE_SYSTEM_NCCL to cmake, which will cause the system's NCCL library to be used (previous behavior), rather than the submodule.


BarclayII:
Would it require the users to install NCCL runtime even if they don't need multi-GPU training?

BarclayII (approved):

PR #2948, opened 2021-05-26
[Distributed] Specify the graph format for distributed training

zheng-da:
## Description
Previously, the graph structure in each partition is stored as CSC. When users call APIs such as find_edges and out_degrees, DGL generates corresponding graph format for these APIs. In the setting of distributed training, the graph structure is shared among all trainers and servers. If an API triggers the construction of another graph format, the graph structure of the new format is stored in local memory. If every trainer and server in a machine does the same thing, a graph will be replicated by many times in the machine.

This PR allows users to specify the graph formats during the launch time. Once the graph format is constructed, any graph API can no longer construct a new graph format during the runtime. If a graph format is required but not created during the launch time, an error is reported.

## Checklist
- [ ] Code is well-documented


python/dgl/distributed/dist_context.py
```diff
@@ -93,11 +93,14 @@ def initialize(ip_config, num_servers=1, num_workers=0,
                 'Please define DGL_NUM_CLIENT to run DistGraph server'
         assert os.environ.get('DGL_CONF_PATH') is not None, \
                 'Please define DGL_CONF_PATH to run DistGraph server'
+        formats = os.environ.get('DGL_GRAPH_FORMAT', 'csc')
+        formats = [f.strip() for f in formats]
```
classicsong, line 97: formats will be 'csc,csr,coo'.
Does formats = [f.strip() for f in formats] work ?

classicsong (approved):
Overall LGTM

classicsong (approved):

PR #2953, opened 2021-05-27
Speed up random number generation.

zheng-da:
## Description
reservoir algorithm is memory efficient, but not computationally efficient. This PR uses a new sampling algorithm, which is strictly faster than reservoir algorithm in terms of computation complexity, to speed up random number generation in the most common case in neighbor sampling.

when `num`=10, `population`=90, this new sampling algorithm gets roughly 7.5x speedup on a single core of Intel(R) Xeon(R) Platinum 8259CL.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/src/random/cpu/choice.cc
+++ b/src/random/cpu/choice.cc
@@ -108,7 +108,7 @@ void RandomEngine::UniformChoice(IdxType num, IdxType population, IdxType* out,
       // complexity O(num). In the case of `num >= population/10`, we don't need to worry about
       // memory complexity because `num` is usually small. So is `population`. Allocating a small
       // piece of memory is very efficient.
-	  std::vector<IdxType> seq(population);
+      std::vector<IdxType> seq(population);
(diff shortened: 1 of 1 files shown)


BarclayII:
How did it impact real datasets (like GraphSAGE in OGB-products or things like that)?

zheng-da:
I just modify the case of `num >= population/10`. `num` is the number of neighbors to sample. if it's a large-degree node, the code usually doesn't go to this branch unless we want to sample a large number of neighbors from a large-degree node. In other words, if the code goes to this branch, it usually doesn't need to allocate a large array. Allocating a small piece of memory is pretty cheap.

Suppose someone really wants to sample a large number of neighbors from a large-degree node. How large it needs to be to hit the overhead of allocating large memory? usually, allocating memory of 1MB triggers mmap. This means that a node has a degree of 256K and a user wants to sample 25K neighbors out of this neighbor list. It's very rare that this scenario happens.

nv-dlasalle (approved):

PR #2947, opened 2021-05-26
fixbug gatconv residual dim mis-match

jxgu1016:
## Checklist
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [ ] To the my best knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

PR #2962, opened 2021-05-31
[Bug] Fix shape mismatch with block inputs for multiple modules

BarclayII:
## Description
This is a reopen of #2947 (originally from @jxgu1016).

The test case was spelled wrong, and as a result some of cases are never tested.


VoVAllen (approved):
LGTM

Issue #2944, opened 2021-05-25
[RFC] add the `exclude_edges` argument in `sample_neighbors`


zheng-da:
In link prediction tasks, we sample a set of edges, compute the embeddings of the endpoint nodes of the edges and compute the node similarity to predict the existence of edges. It is possible that the edges to be predicted are also used to compute node embeddings with message passing, which causes information leak and potentially decreases the model performance. 

Thus, we propose the extension of `sample_neighbors` by adding a new argument called `exclude_edges`. The interface of the extended `sample_neighbors` is:
```
dgl.sampling.sample_neighbors(g, nodes, fanout, edge_dir='in', prob=None, replace=False, copy_ndata=True,
                                                       copy_edata=True, _dist_training=False, exclude_edges=None)
```
The argument has a value of an edge ID array. By default, its value is `None`. If we want to exclude certain edges in the neighbor sampling for the seed nodes, we should pass a 1D array that contains the edge IDs.

zheng-da:
@jennyzhang0215 can you provide your opinions on this proposal here?

jennyzhang0215:
*exclude_edges* as a 1D array can meet many cases.  A more complicated situation is that for different seed nodes the edges to be excluded are different. In this case, *exclude_edges* should be in the format of a list of lists. Each list corresponds to a seed node in the *nodes*.

jennyzhang0215:
If the *exclude_edges* is a 1D array, it means that all the seed nodes exclude the same set of edges when sampling neighbors. So what if different seed nodes need to exclude different sets of edges? For example, seed node *n1* needs to exclude edge *e1* and node *n2* needs to exclude *e2* and *e3*. In this case, if *nodes=[n1, n2]*, then * exclude_edges* can be *[[e1], [e2,e3]]*, which is a list of lists. (We cannot use a 2D tensor here, because the number of edges to be excluded is different.)

However, I cannot show a good application for the above case. So we can just regard *exclude_edges* as a 1D array (for homograph) and a dict (for heterograph) at the current stage.

PR #2971, opened 2021-06-02
[Feature] Exclude edges in sample_neighbors

mszarma:
## Description
Extending sample_neighbors API to support
exclude_edges parameter.

exclude_edges support tensor and dict data
Feature enable excluding certain edges
during neighborhood sampling
Exclude_edges contains EID's of edges
which will be excluded
during neighbor picking for seed nodes.

Added test case for heterograph and homograph
RFC issue id: https://github.com/dmlc/dgl/issues/2944

## Checklist
- [ ] Code is well-documented

## Changes
[Feature] Exclude edges in sample_neighbors
- Extending sample_neighbors API to support exclude_edges parameter.
 - Added test case for heterograph and homograph


python/dgl/sampling/neighbor.py
```diff
@@ -52,6 +52,8 @@ def sample_neighbors(g, nodes, fanout, edge_dir='in', prob=None, replace=False,
         The features must be non-negative floats, and the sum of the features of
         inbound/outbound edges for every node must be positive (though they don't have
         to sum up to one).  Otherwise, the result will be undefined.
+    exclude_edges: tensor or dict
+        Edge IDs to exclude during sampling neighbors for the seed nodes.
```
zheng-da, line 56: is it right that a tensor is for homogeneous graph and dict is for heterogeneous graph? if so, could you make it more explicit?

python/dgl/sampling/neighbor.py
```diff
@@ -155,8 +181,21 @@ def sample_neighbors(g, nodes, fanout, edge_dir='in', prob=None, replace=False,
...
+        if not isinstance(exclude_edges, dict):
+            if len(g.etypes) > 1:
+                raise DGLError("Must specify etype type when the graph is not homogeneous.")
+            exclude_edges = {g.canonical_etypes[0] : exclude_edges}
+        exclude_edges = utils.prepare_tensor_dict(g, exclude_edges, 'edges')
+        for etype in g.canonical_etypes:
+            if etype in exclude_edges:
+                excluded_edges_all_t.append(F.to_dgl_nd(exclude_edges[etype]))
+            else:
+                excluded_edges_all_t.append(nd.array([], ctx=nd.cpu()))
```
zheng-da, line 195: does it generate a right list of excluded edges? the list should be indexed by edge type IDs, but you iterate over the edge type based its order in the dict. There is no guarantee that the order is the same as indicated by edge type IDs.

src/array/cpu/coo_remove.cc
```diff
@@ -81,9 +81,9 @@ COOMatrix COORemove(COOMatrix coo, IdArray entries) {
     return coo;
 
   std::vector<IdType> new_rows, new_cols, new_eids;
-  new_rows.reserve(nnz - n_entries);
-  new_cols.reserve(nnz - n_entries);
-  new_eids.reserve(nnz - n_entries);
+  new_rows.reserve(nnz > n_entries ? nnz - n_entries : nnz);
```
zheng-da, line 84: if nnz < n_entries, return nnz? why?

src/graph/sampling/neighbor/neighbor.cc
```diff
@@ -90,6 +91,8 @@ HeteroSubgraph SampleNeighbors(
         default:
           LOG(FATAL) << "Unsupported sparse format.";
       }
+      if (!exclude_edges.empty() && exclude_edges[etype].GetSize() != 0)
+        sampled_coo = COORemove(sampled_coo, exclude_edges[etype]);
```
zheng-da, line 95: what is the complexity of using this function? `SampleNeighbors` runs on the CSC format. adding this function requires to generate the COO format?

zheng-da:
did you benchmark how fast is this implementation?

mszarma:
> did you benchmark how fast is this implementation?

I benchmarked second version on synthetic bench, here are some examples of the results:

OLD = remove_edges + neighbor_sampling call
NEW = neighbor_sampling with excluded edges call
==================================================
old=0.0427086 , new=0.0160954 , diff=0.0266132 seconds, diff=62.31 percentage , faster

GRAPH num_nodes: 1019999  num_edges: 29855 
sampled_nodes_amount: 174  fanout: 3 excluded edges amount: 168 
SUBGRAPH num_nodes: 1019999  num_edges: 231 
==================================================
==================================================
old=12.5179353 , new=6.0757782 , diff=6.4421570 seconds, diff=51.46 percentage , faster

GRAPH num_nodes: 80002000  num_edges: 239989054 
sampled_nodes_amount: 689396  fanout: 6 excluded edges amount: 1990128 
SUBGRAPH num_nodes: 80002000  num_edges: 4136376
==================================================
==================================================
old=8.8966670 , new=6.6231253 , diff=2.2735417 seconds, diff=25.55 percentage , faster

GRAPH num_nodes: 80020000  num_edges: 239989189 
sampled_nodes_amount: 94441  fanout: 6 excluded edges amount: 76725 
SUBGRAPH num_nodes: 80020000  num_edges: 566646
==================================================
==================================================
old=0.0245998 , new=0.0203795 , diff=0.0042202 seconds, diff=17.16 percentage , faster

GRAPH num_nodes: 299992  num_edges: 149993 
sampled_nodes_amount: 8109  fanout: 178 excluded edges amount: 15545 
SUBGRAPH num_nodes: 299992  num_edges: 5420
==================================================
==================================================
old=0.0080178 , new=0.0049868 , diff=0.0030310 seconds, diff=37.80 percentage , faster

GRAPH num_nodes: 30000  num_edges: 30000 
sampled_nodes_amount: 7653  fanout: 9 excluded edges amount: 16394 
SUBGRAPH num_nodes: 30000  num_edges: 3677
==================================================

mszarma:
> did you benchmark how fast is this implementation?

Yes, i bench second version on synthetic benchmark , here are some examples of results:
OLD = remove_edges + neighbor_sampling call
NEW = neighbor_sampling with excluded edges call
==================================================
old=0.0038526 , new=0.0017021 , diff=0.0021505 seconds, diff=55.82 percentage , faster

GRAPH num_nodes: 2997  num_edges: 1909 
sampled_nodes_amount: 735  fanout: 151 excluded edges amount: 1165 
SUBGRAPH num_nodes: 2997  num_edges: 198
==================================================
==================================================
old=0.0061331 , new=0.0036194 , diff=0.0025136 seconds, diff=40.99 percentage , faster

GRAPH num_nodes: 29998  num_edges: 29797 
sampled_nodes_amount: 1126  fanout: 196 excluded edges amount: 684 
SUBGRAPH num_nodes: 29998  num_edges: 1775
==================================================
==================================================
old=0.0427086 , new=0.0160954 , diff=0.0266132 seconds, diff=62.31 percentage , faster

GRAPH num_nodes: 1019999  num_edges: 29855 
sampled_nodes_amount: 174  fanout: 3 excluded edges amount: 168 
SUBGRAPH num_nodes: 1019999  num_edges: 231 
==================================================

==================================================
old=12.5179353 , new=6.0757782 , diff=6.4421570 seconds, diff=51.46 percentage , faster

GRAPH num_nodes: 80002000  num_edges: 239989054 
sampled_nodes_amount: 689396  fanout: 6 excluded edges amount: 1990128 
SUBGRAPH num_nodes: 80002000  num_edges: 4136376
==================================================
==================================================
old=8.8966670 , new=6.6231253 , diff=2.2735417 seconds, diff=25.55 percentage , faster

GRAPH num_nodes: 80020000  num_edges: 239989189 
sampled_nodes_amount: 94441  fanout: 6 excluded edges amount: 76725 
SUBGRAPH num_nodes: 80020000  num_edges: 566646
==================================================

zheng-da:
how did you benchmark it? i think you can just take one dataset (e.g., OGBN-products) and benchmark it with and without using this PR for edge sampling.

mszarma (commented):
> how did you benchmark it? i think you can just take one dataset (e.g., OGBN-products) and benchmark it with and without using this PR for edge sampling.

@zheng-da i made a simple bench test - used API calls and compare it with one new call.
OLD = remove_edges + neighbor_sampling call
NEW = neighbor_sampling with excluded edges call

 I think to benchmark it like you suggest master vs PR it's not available out of the box. Probably afaiu, someone with good understanding of this GNN use case would need to help to adapt upper layers code/calls to make it possible. Who could help with that? Jenny?

src/graph/sampling/neighbor/neighbor.cc
```diff
@@ -104,6 +95,39 @@ HeteroSubgraph SampleNeighbors(
...
+    << "Number of fanout values must match the number of edge types.";
+  CHECK_EQ(prob.size(), hg->NumEdgeTypes())
+    << "Number of probability tensors must match the number of edge types.";
+
+  if (!exclude_edges.empty()) {
+    std::vector<IdArray> remain_edges(hg->NumEdgeTypes());
+    for (dgl_type_t etype = 0; etype < hg->NumEdgeTypes(); ++etype) {
+     EdgeArray edges = hg->Edges(etype);
+     remain_edges[etype] =
+            edges.excludeCertainEids<decltype(hg->NumEdges(etype))>(exclude_edges[etype]);
```
BarclayII, line 120: `NumEdges` is always `int64_t`.  So this function will fail if the graph is `int32`.

You can dispatch to different primitive types using `ATEN_ID_TYPE_SWITCH` (similarly there are other switches).  It works like this:
```C++
ATEN_ID_TYPE_SWITCH(hg->DataType(), IdType, {
  remain_edges[etype] =
      edges.excludeCertainEids<IdType>(exclude_edges[etype]);
});
```

mszarma: Done in newest commit version.

src/graph/sampling/neighbor/neighbor.cc
```diff
@@ -104,6 +95,41 @@ HeteroSubgraph SampleNeighbors(
...
+    << "Number of node ID tensors must match the number of node types.";
+  CHECK_EQ(fanouts.size(), hg->NumEdgeTypes())
+    << "Number of fanout values must match the number of edge types.";
+  CHECK_EQ(prob.size(), hg->NumEdgeTypes())
+    << "Number of probability tensors must match the number of edge types.";
+
+  if (!exclude_edges.empty()) {
+    std::vector<IdArray> remain_edges(hg->NumEdgeTypes());
+    for (dgl_type_t etype = 0; etype < hg->NumEdgeTypes(); ++etype) {
+     EdgeArray edges = hg->Edges(etype);
```
zheng-da, line 118: the indent is not correct here.

src/graph/sampling/neighbor/neighbor.cc
```diff
@@ -104,6 +95,41 @@ HeteroSubgraph SampleNeighbors(
   return ret;
 }
 
+HeteroSubgraph SampleNeighbors(
```
zheng-da, line 98: this implementation is not efficient on a large graph. you are remove the edges from the input graph. you should remove edges from the sampled subgraph.

zheng-da:
you need to benchmark your implementation on a large graph.

zheng-da:
you can take a look at the implementation of `EdgeDataLoader`: https://doc.dgl.ai/api/python/dgl.dataloading.html#dgl.dataloading.pytorch.EdgeDataLoader
it excludes certain edges.

python/dgl/dataloading/dataloader.py
```diff
@@ -248,33 +253,8 @@ def sample_blocks(self, g, seed_nodes, exclude_eids=None):
...
-                    # so I need to keep track of the edge IDs left one by one.
-                    new_eids = parent_eids.copy()
-                    for k, v in located_eids.items():
-                        if len(v) > 0:
-                            frontier = transform.remove_edges(
-                                frontier, v, etype=k, store_ids=True)
-                            new_eids[k] = F.gather_row(parent_eids[k], frontier.edges[k].data[EID])
-                    frontier.edata[EID] = new_eids
+
+            frontier = self.sample_frontier(block_id, g, seed_nodes_in, exclude_eids=exclude_eids)
```
BarclayII, line 257: This is moving the edge ID exclusion into `sample_frontier`.  The reason we have edge ID exclusion after `sample_frontier` is that we would like to let user write their own neighbor samplers without taking care of edge exclusion.  Removing it like this will potentially break existing code since the edges are no longer automatically removed and users might do extra work.  @zheng-da do you think it's better to straight move it into `sample_frontier`?

python/dgl/sampling/neighbor.py
BarclayII: Is this resolved?  I saw it is now iterating over `g.canonical_etypes`.

src/graph/sampling/neighbor/neighbor.cc
```diff
@@ -17,12 +17,57 @@ using namespace dgl::aten;
 namespace dgl {
 namespace sampling {
 
+HeteroSubgraph ExcludeCertainEdges(
```
zheng-da, line 20: how is this different from calling remove_edges in python?

zheng-da: actually, i'm kind of surprised that this solution has speedup over the existing solution.

python/dgl/dataloading/dataloader.py
BarclayII: After discussion with Minjie we decide it's better to keep the original behavior so existing code will not break.  To have the best of both worlds we can add a flag in `BlockSampler`, say `exclude_edges_in_sample_frontier`.  When it is True, BlockSampler assumes that the edges must be excluded in `sample_frontier`, and `sample_blocks` no longer removes the edges.  Otherwise, it resorts to the current behavior where a frontier is first sampled then with edges removed.  The argument defaults to False, and could be set to True in `MultiLayerNeighborSampler` so `sample_neighbors` with `exclude_edges` can still be utilized.

BarclayII:
Noticed another problem: now that we are putting `exclude_eids` in `sample_neighbors`, `exclude_eids_in_frontier` may not work by itself since we need to support the GPU version as well.  Will fix this later.

BarclayII (approved):

PR #2972, opened 2021-06-02
[CPU Parallel] implementation of parallel_for with grain size

tpatejko:
This PR implements `parallel_for`, a function wrapper that allows controlling number of threads executing OpenMP parallel region by setting grain size, a number of iterations a single OpenMP thread will execute.


BarclayII (approved):

BarclayII:
Parallel reduction depends on this PR so I'll merge this PR for now as it doesn't really impact any routine's performance yet.  @tpatejko

PR #2973, opened 2021-06-02
[Performance][Feature] Add `src_nodes` paramter to `to_block()` to avoid cost running unique() when available.

nv-dlasalle:
## Description
This adds the parameter `src_nodes` to `to_block()`, which can save time when available. On the GPU, this is particularly important, as it avoid the synchronous copy of the number of unique source nodes from the GPU to the CPU to size the tensor.

## Checklist
- [x] Code is well-documented

## Changes
This updates both the CPU and GPU implementation of to_block.

This extends the existing unit test cover specifying the list of unique source nodes.


src/graph/transform/to_bipartite.cc
```diff
@@ -15,8 +15,7 @@
 #include <dgl/runtime/container.h>
 #include <vector>
 #include <tuple>
-// TODO(BarclayII): currently ToBlock depend on IdHashMap<IdType> implementation which
-// only works on CPU.  Should fix later to make it device agnostic.
```
nv-dlasalle, line 19: Given that we have to_block implemented on the GPU, this comment didn't seem relevant/correct anymore.

python/dgl/transform.py
```diff
@@ -2097,15 +2103,36 @@ def to_block(g, dst_nodes=None, include_dst_in_src=True):
...
+    src_node_ids_nd = None
+    if src_nodes is not None and not isinstance(src_nodes, Mapping):
+        # src_nodes is a Tensor, check if the g has only one type.
+        if len(g.ntypes) > 1:
+            raise DGLError(
+                'Graph has more than one node type; please specify a dict for src_nodes.')
+        src_nodes = {g.ntypes[0]: src_nodes}
+        src_node_ids = [
+            utils.toindex(src_nodes.get(ntype, []), g._idtype_str).tousertensor(
+                ctx=F.to_backend_ctx(g._graph.ctx))
```
BarclayII, line 2116: I feel you can just use `F.copy_to(F.tensor(...), F.to_backend_ctx(...))`.  If the input tensor is already on the input device it does not actually perform a copy.

src/graph/transform/to_bipartite.h
```diff
@@ -19,7 +19,7 @@ namespace transform {
 template<DLDeviceType XPU, typename IdType>
 std::tuple<HeteroGraphPtr, std::vector<IdArray>, std::vector<IdArray>>
 ToBlock(HeteroGraphPtr graph, const std::vector<IdArray> &rhs_nodes,
-        bool include_rhs_in_lhs);
+        bool include_rhs_in_lhs, std::vector<IdArray> lhs_nodes);
```
BarclayII, line 22: Seems that `lhs_nodes` should be mutable and is always the same object as the second element in the returned tuple?  If that's the case I will no longer return `lhs_nodes` but instead make `lhs_nodes` an output parameter (`std::vector<IdArray> &`).

BarclayII:
I'm also wondering when you will have `src_nodes` available without calling unique.

zheng-da (approved):

python/dgl/transform.py
nv-dlasalle: Changed to `F.copy_to()` in 33b5d14.

src/graph/transform/to_bipartite.h
nv-dlasalle: Switched to an in/out parameter in 
add691b. However, using a pointer as the linter doesn't like non-const references.

PR #2983, opened 2021-06-04
[Doc] Revised Two Issues in Message Passing Tutorial

kayzliu:
## Description

1. Although the 1D weight vector works in DGL built-in message function, it doesn't fit in Numpy broadcasting rule, and it result in error if we replace the built-in message function with user defined message function. Converting the weight into 2D matrix can solve this problem.
2. Previous sections in the tutorial use mean as reduce function, but in customized message function example use sum. It's better to keep a consistent reduce message function.
@zhjwy9343 

## Checklist
- [ ] Code is well-documented


zhjwy9343 (approved):
LGTM

PR #2989, opened 2021-06-07
[Feature] Add Heterograph support on Python for builtin unary msg functions (copy_u, copy_e)

isratnisa:
## Description
Add Python API to support builtin message functions in heterogeneous graphs. Currently supports forward and backward operators for unary builtin message functions: `copy_u`, `copy_e`

TO DO:
1. Fix backward operator for binary functions 
2.  Support reduce functions: `max`, `min`, and `mean`. Currently supports `sum`.

## Checklist
- [ ] Code is well-documented


python/dgl/backend/__init__.py
```diff
@@ -7,7 +7,8 @@
 
 from . import backend
 from .set_default_backend import set_default_backend
-
+from .pytorch.sparse import gspmm_hetero
+from .pytorch.sparse import gsddmm_hetero
```
zheng-da, line 11: we should import any function in pytorch backend here.

python/dgl/core.py
```diff
@@ -291,6 +291,17 @@ def message_passing(g, mfunc, rfunc, afunc):
             msgdata = invoke_edge_udf(g, ALL, g.canonical_etypes[0], mfunc, orig_eid=orig_eid)
         # reduce phase
         if is_builtin(rfunc):
+            # Converting msgdata from tuple of tensors to dict of tensors for heterographs
+            mdict = dict()
+            if g._graph.number_of_etypes() > 1:
+                msgdata_dict = {}
+                key = list(msgdata.keys())[0]
```
zheng-da, line 298: what does msgdata contain?

python/dgl/heterograph.py
```diff
@@ -4799,6 +4799,95 @@ def multi_update_all(self, etype_dict, cross_reducer, apply_node_func=None):
             if apply_node_func is not None:
                 self.apply_nodes(apply_node_func, ALL, self.ntypes[dtid])
 
+    #################################################################
+    # New Message passing on heterograph // will replace *update_all
+    #################################################################
+    def update_all_new(self,
+                       mfunc,
+                       rfunc,
+                       afunc=None):
```
zheng-da, line 4808: we should use `update_all`

python/dgl/heterograph.py
```diff
@@ -4799,6 +4799,95 @@ def multi_update_all(self, etype_dict, cross_reducer, apply_node_func=None):
...
+            An optional apply function after the messages are reduced both
+            type-wisely and across different types.
+            It must be a :ref:`apiudf`.
+
+        Notes
+        -----
+        DGL recommends using DGL's bulit-in function for the message_func
+        and the reduce_func in the type-wise message passing arguments,
+        because DGL will invoke efficient kernels that avoids copying node features to
+        edge features in this case.
```
zheng-da, line 4846: we should explicitly what is allowed and what is not enabled in this release.

python/dgl/heterograph.py
```diff
@@ -4799,6 +4799,95 @@ def multi_update_all(self, etype_dict, cross_reducer, apply_node_func=None):
...
+        >>> g.nodes['user'].data['h']
+        tensor([[0.],
+                [4.]])
+        """
+        if rfunc.name in ['max', 'min', 'mean']:
+            raise NotImplementedError("Reduce op \'" + rfunc.name + "\' is not supported in "
+                                      "the new heterograph API. Use multi_update_all().")
+        # if mfunc.name not in ['copy_u', 'copy_e']:
+        #     raise NotImplementedError("Op \'" + mfunc.name + "\' is not supported in "
+        #                               "the new heterograph API. Use multi_update_all().")
```
zheng-da, line 4876: remove this

python/dgl/heterograph.py
```diff
@@ -4799,6 +4799,95 @@ def multi_update_all(self, etype_dict, cross_reducer, apply_node_func=None):
...
+        out_tensor_tuples = all_out[key]
+
+        dst_tensor = {}
+        for _, _, dsttype in g.canonical_etypes:
+            dtid = g.get_ntype_id(dsttype)
+            dst_tensor[key] = out_tensor_tuples[dtid]
+            self._node_frames[dtid].update(dst_tensor)
+        # TODO (Israt): Add apply function in heterograph
+        # if apply_node_func is not None:
+        #     self.apply_nodes(apply_node_func, ALL, self.ntypes[dtid])
```
zheng-da, line 4889: why is this not enabled?

python/dgl/ops/sddmm.py
```diff
@@ -3,10 +3,32 @@
...
+        The right operand, could be None if it's not required by op.
+    """
+    lhs_shape = F.shape(lhs_data)
+    rhs_shape = F.shape(rhs_data)
+    if len(lhs_shape) != len(rhs_shape):
+        max_ndims = max(len(lhs_shape), len(rhs_shape))
+        lhs_pad_ndims = max_ndims - len(lhs_shape)
+        rhs_pad_ndims = max_ndims - len(rhs_shape)
+        new_lhs_shape = (lhs_shape[0],) + (1,) * lhs_pad_ndims + lhs_shape[1:]
+        new_rhs_shape = (rhs_shape[0],) + (1,) * rhs_pad_ndims + rhs_shape[1:]
```
zheng-da, line 28: can you give an example of the output shape? it seems you add a new dimension in the tensor, but the value of the added dimension isn't 1.

python/dgl/ops/sddmm.py
```diff
@@ -43,24 +65,41 @@ def gsddmm(g, op, lhs_data, rhs_data, lhs_target='u', rhs_target='v'):
...
+            lhs_shape = F.shape(lhs_data)
+            rhs_shape = F.shape(rhs_data)
+            if len(lhs_shape) != len(rhs_shape):
+                max_ndims = max(len(lhs_shape), len(rhs_shape))
+                lhs_pad_ndims = max_ndims - len(lhs_shape)
+                rhs_pad_ndims = max_ndims - len(rhs_shape)
+                new_lhs_shape = (lhs_shape[0],) + (1,) * lhs_pad_ndims + lhs_shape[1:]
+                new_rhs_shape = (rhs_shape[0],) + (1,) * rhs_pad_ndims + rhs_shape[1:]
+                lhs_data = F.reshape(lhs_data, new_lhs_shape)
+                rhs_data = F.reshape(rhs_data, new_rhs_shape)
```
zheng-da, line 83: can you call reshape function here?

python/dgl/ops/spmm.py
```diff
@@ -2,10 +2,32 @@
...
+    lhs_shape = F.shape(lhs_data)
+    rhs_shape = F.shape(rhs_data)
+    if len(lhs_shape) != len(rhs_shape):
+        max_ndims = max(len(lhs_shape), len(rhs_shape))
+        lhs_pad_ndims = max_ndims - len(lhs_shape)
+        rhs_pad_ndims = max_ndims - len(rhs_shape)
+        new_lhs_shape = (lhs_shape[0],) + (1,) * lhs_pad_ndims + lhs_shape[1:]
+        new_rhs_shape = (rhs_shape[0],) + (1,) * rhs_pad_ndims + rhs_shape[1:]
+        lhs_data = F.reshape(lhs_data, new_lhs_shape)
+        rhs_data = F.reshape(rhs_data, new_rhs_shape)
```
zheng-da, line 29: this reshape function seems the same as the one above.

python/dgl/ops/spmm.py
```diff
@@ -43,29 +65,63 @@ def gspmm(g, op, reduce_op, lhs_data, rhs_data):
...
+            lhs_shape = F.shape(lhs_data)
+            rhs_shape = F.shape(rhs_data)
+            if len(lhs_shape) != len(rhs_shape):
+                max_ndims = max(len(lhs_shape), len(rhs_shape))
+                lhs_pad_ndims = max_ndims - len(lhs_shape)
+                rhs_pad_ndims = max_ndims - len(rhs_shape)
+                new_lhs_shape = (lhs_shape[0],) + (1,) * lhs_pad_ndims + lhs_shape[1:]
+                new_rhs_shape = (rhs_shape[0],) + (1,) * rhs_pad_ndims + rhs_shape[1:]
+                lhs_data = F.reshape(lhs_data, new_lhs_shape)
+                rhs_data = F.reshape(rhs_data, new_rhs_shape)
```
zheng-da, line 85: the same comment. call the reshape function here.

python/dgl/sparse.py
```diff
@@ -179,6 +180,86 @@ def _gspmm(gidx, op, reduce_op, u, e):
...
+                e = F.unsqueeze(e, -1)
+                expand_e = True
+        ctx = F.context(u) if use_u else F.context(e) # TODO(Israt): Put outside of loop
+        dtype = F.dtype(u) if use_u else F.dtype(e) # TODO(Israt): Put outside of loop
+        u_shp = F.shape(u) if use_u else (0,)
+        e_shp = F.shape(e) if use_e else (0,)
+        if use_u:
+            list_u[src_id] = u if use_u else None
+        if use_e:
+            list_e[etid] = e if use_e else None
```
zheng-da, line 221: why don't you put everything in one `if use_u` condition and `if use_e` condition?

python/dgl/sparse.py
```diff
@@ -179,6 +180,86 @@ def _gspmm(gidx, op, reduce_op, u, e):
...
+        v_shp = (gidx.number_of_nodes(dst_id), ) +\
+            infer_broadcast_shape(op, u_shp[1:], e_shp[1:])
+        list_v[dst_id] = F.zeros(v_shp, dtype, ctx)
```
zheng-da, line 224: can `v` be also used as input data?

tests/compute/test_new_update_all_hetero.py
```diff
@@ -0,0 +1,173 @@
...
+
+        if not F.allclose(r1, r2):
+            _print_error(r1, r2)
+        assert F.allclose(r1, r2)
+        if not F.allclose(n_grad1, n_grad2):
+            print('node grad')
+            _print_error(n_grad1, n_grad2)
+        assert(F.allclose(n_grad1, n_grad2))
+
+    _test(fn.copy_u, fn.sum)
```
zheng-da, line 95: do you support other message functions?

python/dgl/core.py
isratnisa: The output of message phase.

tests/compute/test_new_update_all_hetero.py
isratnisa: This PR supports the two unary functions available in dgl: copy_u, copy_e

python/dgl/backend/__init__.py
isratnisa: Putting it inside the pytorch folder doesn't work. Where should I put it?

python/dgl/heterograph.py
isratnisa: My bad. This should be enabled for this PR.

python/dgl/ops/sddmm.py
isratnisa: We can. But I didn't want to touch the original code written for homogeneous code.

python/dgl/backend/pytorch/sparse.py
```diff
@@ -145,6 +145,84 @@ def backward(ctx, dZ):
         return None, None, None, dX, dY
 
 
+class GSpMM_hetero(th.autograd.Function):
+    @staticmethod
+    @custom_fwd(cast_inputs=th.float16)
+    def forward(ctx, g, op, reduce_op, *X): # X = X + Y
```
jermainewang, line 151: maybe use `feats` instead of `X`?

python/dgl/backend/pytorch/sparse.py
```diff
@@ -145,6 +145,84 @@ def backward(ctx, dZ):
...
+        ctx.backward_cache = g, op, reduce_op
+        ctx.save_for_backward(*X, argX, argY)
+        return out
+
+    @staticmethod
+    @custom_bwd
+    def backward(ctx, *dZ):
+        g, op, reduce_op = ctx.backward_cache 
+        X = ctx.saved_tensors[:-2]
+        argX=ctx.saved_tensors[-2]
```
jermainewang, line 162:
```suggestion
        argX = ctx.saved_tensors[-2]
```

python/dgl/backend/pytorch/sparse.py
```diff
@@ -145,6 +145,84 @@ def backward(ctx, dZ):
...
+        ctx.save_for_backward(*X, argX, argY)
+        return out
+
+    @staticmethod
+    @custom_bwd
+    def backward(ctx, *dZ):
+        g, op, reduce_op = ctx.backward_cache 
+        X = ctx.saved_tensors[:-2]
+        argX=ctx.saved_tensors[-2]  
+        argY=ctx.saved_tensors[-1]
```
jermainewang, line 163:
```suggestion
        argY = ctx.saved_tensors[-1]
```

python/dgl/backend/pytorch/sparse.py
```diff
@@ -145,6 +145,84 @@ def backward(ctx, dZ):
...
+        out, (argX, argY) = _gspmm_hetero(g, op, reduce_op, X)
+        ctx.backward_cache = g, op, reduce_op
+        ctx.save_for_backward(*X, argX, argY)
+        return out
+
+    @staticmethod
+    @custom_bwd
+    def backward(ctx, *dZ):
+        g, op, reduce_op = ctx.backward_cache 
+        X = ctx.saved_tensors[:-2]
```
jermainewang, line 161: same. use `feats` instead of `X` to avoid confusion?

python/dgl/backend/pytorch/sparse.py
```diff
@@ -145,6 +145,84 @@ def backward(ctx, dZ):
...
+        argX=ctx.saved_tensors[-2]  
+        argY=ctx.saved_tensors[-1]
+        num_ntypes = g._graph.number_of_ntypes()
+        X ,Y = X[:num_ntypes], X[num_ntypes:]
+
+
+        if op != 'copy_rhs' and any([x is not None for x in X]):
```
jermainewang, line 168: I see only a certain branch (i.e., when `op == 'copy_lhs'`). I suppose you will update the rest in future PRs. If that's the case, please have an explicit check before. Something like:
```python
if g._graph.number_of_etypes() > 1 and (op, reduce_op) not in [('copy_lhs', 'sum'), ('copy_rhs', 'sum')]:
    raise NotImplementedError('... some error msg ...')

# begin the main logic ...
if op != 'copy_rhs' and any([x is not None for x in X]):
```

python/dgl/backend/pytorch/sparse.py
```diff
@@ -206,6 +284,71 @@ def backward(ctx, dZ):
...
+        ctx.save_for_backward(*X)
+        return out
+
+    @staticmethod
+    @custom_bwd
+    def backward(ctx, *dZ):
+        g, op, lhs_target, rhs_target = ctx.backward_cache
+        gidx = g._graph
+        X = ctx.saved_tensors
+        if op != 'copy_rhs' and ctx.needs_input_grad[2]:
```
jermainewang, line 302: same suggestions as in `GSPMM`

python/dgl/core.py
jermainewang: This also means the `message_passing` function should return a value of type `Dict[str, List[Tensor]]` accordingly.

python/dgl/ops/sddmm.py
jermainewang: Is this code copy-pasted from https://github.com/dmlc/dgl/blob/fc69f6cecad20335d076bf1c64cf03b2be5f6d37/python/dgl/ops/sddmm.py#L47-L60 ?

If so, please also call into this function inside the `if g._graph.number_of_etypes() == 1` branch of the `gsddmm` and `gspmm` functions to avoid duplication. Please also keep the original comment about the shape example per Da's suggestion.

python/dgl/ops/sddmm.py
```diff
@@ -43,24 +65,41 @@ def gsddmm(g, op, lhs_data, rhs_data, lhs_target='u', rhs_target='v'):
...
+        rhs_data_dict = rhs_data
+        lhs_list = [None] * g._graph.number_of_ntypes()
+        rhs_list = [None] * g._graph.number_of_ntypes()
+        for srctype, _, dsttype in g.canonical_etypes:
+            src_id = g.get_ntype_id(srctype)
+            dst_id = g.get_ntype_id(dsttype)
+            lhs_data = lhs_data_dict[srctype]
+            rhs_data = rhs_data_dict[dsttype]
+            if op not in ['copy_lhs', 'copy_rhs']:
+                lhs_data, rhs_data = reshape_lhs_rhs(lhs_data, rhs_data)
```
jermainewang, line 97: Since you only allow 'copy_u' and 'copy_e', this branch will never happen right? Consider delete it.

python/dgl/ops/sddmm.py
```diff
@@ -43,24 +65,41 @@ def gsddmm(g, op, lhs_data, rhs_data, lhs_target='u', rhs_target='v'):
...
+                lhs_data = F.reshape(lhs_data, new_lhs_shape)
+                rhs_data = F.reshape(rhs_data, new_rhs_shape)
+        return gsddmm_internal(
+            g._graph, op, lhs_data, rhs_data, lhs_target, rhs_target)
+    else:
+        lhs_data_dict = lhs_data
+        rhs_data_dict = rhs_data
+        lhs_list = [None] * g._graph.number_of_ntypes()
+        rhs_list = [None] * g._graph.number_of_ntypes()
+        for srctype, _, dsttype in g.canonical_etypes:
```
jermainewang, line 91: My understanding is that this for-loop is to convert the `dict[str, Tensor]` type argument to `list[Tensor]` type. As I have suggested above, you could handle this in `core.invoke_gspmm` and `core.invoke_gsddmm` so all the functions in the deeper stack will use `list[Tensor]`.

python/dgl/ops/spmm.py
```diff
@@ -43,29 +66,64 @@ def gspmm(g, op, reduce_op, lhs_data, rhs_data):
...
+        # Replace infinity with zero for isolated nodes when reducer is min/max
+        # if reduce_op in ['min', 'max']:
+        #     # TODO (Israt): Following code "often" throws 'free(): invalid pointer
+        #     ret = list(ret)
+        #     for i in range(g._graph.number_of_ntypes()):
+        #         if(len(ret[i]) > 0):
+        #             ret[i] = F.replace_inf_with_zero(ret[i])
+        #     ret = tuple(ret)
+
+    # TODO (Israt): Add support for 'mean' in heterograph
```
jermainewang, line 126: One-level 'mean' aggregation is not well-defined in heterograph. You could skip it and throw error if users specify it.

jermainewang (changes requested):

python/dgl/backend/backend.py
```diff
@@ -1452,6 +1452,45 @@ def gspmm(gidx, op, reduce_op, lhs_data, rhs_data):
...
+
+    .. math::
+        x_v = \psi_{(u, v, e)\in \mathcal{G}}(\rho(x_u, x_e))
+
+    where :math:`x_v` is the returned feature on destination nodes, and :math`x_u`,
+    :math:`x_e` refers to :attr:`u`, :attr:`e` respectively. :math:`\rho` means binary
+    operator :attr:`op` and :math:`\psi` means reduce operator :attr:`reduce_op`,
+    :math:`\mathcal{G}` is the graph we apply gspmm on: :attr:`g`.
+
+    Note that this function does not handle gradients.
```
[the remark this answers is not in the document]
isratnisa: The original 'spmm' function notes it as 'does not'. I am not sure from which perspective it says that. I followed the similar description.

python/dgl/backend/pytorch/sparse.py
isratnisa: We have this check at the beginning of message passing function in python/dgl/heterograph.py. We can consider deleting the part since the code will never reach this point.

python/dgl/backend/backend.py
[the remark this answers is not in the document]
BarclayII: This function indeed only handles forward propagation; backward propagation is handled in its respective framework-specific operator.

python/dgl/heterograph.py
```diff
@@ -4799,6 +4799,95 @@ def multi_update_all(self, etype_dict, cross_reducer, apply_node_func=None):
             if apply_node_func is not None:
                 self.apply_nodes(apply_node_func, ALL, self.ntypes[dtid])
 
+    #################################################################
+    # New Message passing on heterograph // will replace *update_all
+    #################################################################
+    def update_all_new(self,
+                       message_func,
+                       reduce_func,
+                       apply_node_func=None):
```
BarclayII, line 4808: Can we merge this function into `update_all` in this release?  It's OK not supporting every combination of message and reduce functions for the case of heterogeneous graphs.

python/dgl/ops/sddmm.py
BarclayII: I would rather leave it here since we will be adding this branch back later anyway when we introduce other builtin message functions.

python/dgl/core.py
```diff
@@ -255,6 +255,19 @@ def invoke_gspmm(graph, mfunc, rfunc, *, srcdata=None, dstdata=None, edata=None)
...
+            rhs_list = [None] * graph._graph.number_of_etypes()
+            if mfunc.name == "copy_u":
+                for srctype, _, _ in graph.canonical_etypes:
+                    src_id = graph.get_ntype_id(srctype)
+                    lhs_list[src_id] = x[srctype]
+            elif mfunc.name == "copy_e":
+                for rel in graph.canonical_etypes:
+                    etid = graph.get_etype_id(rel)
+                    rhs_list[etid] = x[rel]
+            x = tuple(lhs_list + rhs_list)
```
jermainewang, line 270: Probably we want to add a utility function for this so later we could reuse it in many places. Sth like:
```python
def data_dict_to_tuple(data_dict, name, types):
    """Get node or edge feature data of the given name for all the types.

    Parameters
    -------------
    data_dict : dict[str, dict[str, Tensor]] or dict[str, dict[(str, str, str), Tensor]]
        Node or edge data stored in DGLGraph. The key of the outer dictionary is the feature name,
        while the key of the inner one is the node type name or edge type tuple.
    name : str
        Feature name.
    types : list[str] or list[(str, str, str)]
        Node type names or edge type tuples.

    Returns
    --------
    data_tuple : tuple(Tensor)
        Feature data stored in tuple of tensors. The i^th tensor stores the feature
        data of type ``types[i]``.
    """
    pass
```

For implementation, you could avoid calling `get_ntype_id` and `get_etype_id` because the type names you retrieved from DGLGraph are already in the type ID order. Simply use:
```python
for stid, srctype in enumerate(graph.srctypes):
    # stid is already source type ID
    ...

for etid, etype in enumerate(graph.canonical_etypes):
    # etid is already edge type ID
    ...
```

python/dgl/ops/spmm.py
```diff
@@ -43,31 +68,30 @@ def gspmm(g, op, reduce_op, lhs_data, rhs_data):
...
+                                    'sum' if reduce_op == 'mean' else reduce_op,
+                                    *lhs_and_rhs_tuple)
+
+    # TODO (Israt): Add support for 'max', 'min', 'mean' in heterograph
 
     # divide in degrees for mean reducer.
     if reduce_op == 'mean':
+        if g._graph.number_of_etypes() > 1:
+            raise NotImplementedError("Reduce op 'mean' is not supported in "
+                                      "the new heterograph API. Use multi_update_all().")
```
jermainewang, line 94: Users will not understand what this *new* API is referring to in the future. `"Cannot set both intra-type and inter-type reduce operators as 'mean' using update_all. Please use multi_update_all instead."`. That being said, such error should be raised inside the `update_all` function instead of here. The reason is that `gspmm` might be invoked by other high-level APIs than `update_all` and this error message will be out-of-context.

python/dgl/sparse.py
jermainewang: For spmm, no.

jermainewang (changes requested):

python/dgl/core.py
isratnisa: I don't think the following two segments of code are same:
`for stid, srctype in enumerate(graph.srctypes):
    # stid is already source type ID
    ...`
 `for srctype, _, _ in graph.canonical_etypes:
       stid = graph.get_ntype_id(srctype)
       ...`
E.g., a graph has total #ntypes=3, #etypes=2 (ntype 0 -> ntype 1, ntype 2-> ntype 1). I want the srctypes to be 0 and 2. I think enumeration will always assign them as 0 and 1.

jermainewang: I see. You are right.

python/dgl/heterograph.py
```diff
@@ -4694,13 +4698,58 @@ def update_all(self,
...
+            out_tensor_tuples = all_out[key]
+
+            dst_tensor = {}
+            for _, _, dsttype in g.canonical_etypes:
+                dtid = g.get_ntype_id(dsttype)
+                dst_tensor[key] = out_tensor_tuples[dtid]
+                self._node_frames[dtid].update(dst_tensor)
+            # apply
+            if apply_node_func is not None:
+                self.apply_nodes(apply_node_func, ALL, self.ntypes[dtid])
```
jermainewang, line 4751: `core.message_passing` will call apply too. This will call it an extra time.

jermainewang (approved):
LGTM

jermainewang:
Did you just change `spmm.cu`?

BarclayII (approved):
That's a huge effort.  Thanks @isratnisa !

PR #2990, opened 2021-06-07
[Performance][Kernel] Eliminating ctor&dtor zand IsNullArray overhead in random walks

AjayBrahmakshatriya:
Added a special implementation for MetapathBasedRandomWalkStep for the Uniform randomwalk case

## Description

There is a big performance issue with the implementation of the `MetapathRandomWalkStep` function. Specifically with the lines - 

```
FloatArray prob_etype = prob[etype];
  IdxType idx = 0;
  if (IsNullArray(prob_etype)) {
```

The copy constructor and destructors of type FloatArray update the reference count for the underlying data buffer. This is an atomic operation since FloatArray is supposed to be thread safe. But when `MetapathRandomWalkStep` function is called from multiple threads potentially running on multiple cores, the same integer value (refcount) is read and written concurrently from all the threads. This leads to cache ping-ponging and a severe performance hit. The first change in this PR removes the call to the copy constructor and destructor by making `prob_etype` a reference instead of a copy. 

This only solves the problem partially. `IsNullArray` check performed only every step for every randomwalk also hurts the performance. This PR creates a separate version of the `MetapathRandomWalkStep` to be used in the uniform case - `MetapathRandomWalkStepUniform`. The `IsNullArray` check is hoisted to the point where the step lambda is created. 

The implementation before and after is tested with a simple randomwalk invocation like - 

```
num_walks = 1000
num_seeds = 1000
seeds = np.random.choice(max_src_id + 1, num_seeds, replace=False)
seeds = np.repeat(seeds, num_walks)
start = time.time()
paths = dgl.sampling.random_walk(g, seeds, length=6)[0]
end = time.time()
```

When tested on the [ml-1m](https://grouplens.org/datasets/movielens/1m/) on a 48 core system the baseline performance is - 
> Using backend: pytorch
> random walk: 0.8552s

With the changes in the PR the new performance numbers for the uniform case are - 
> Using backend: pytorch
> random walk: 0.0348s

A similar performance improvement is also seen on other graphs. The performance on the Non-Uniform case should also be improved because of eliminating the call to the copy constructor and destructor. 

## Checklist
- [x] Code is well-documented

### DIFF
--- a/src/graph/sampling/randomwalks/metapath_randomwalk.h
+++ b/src/graph/sampling/randomwalks/metapath_randomwalk.h
@@ -116,6 +116,7 @@ std::pair<dgl_id_t, bool> MetapathRandomWalkStep(
  * \param terminate Predicate for terminating the current random walk path.
  *
  * \return A pair of ID of next successor (-1 if not exist), as well as whether to terminate.
+ * \note This function is called only if all the probability arrays are null.
  */
 template<DLDeviceType XPU, typename IdxType>
 std::pair<dgl_id_t, bool> MetapathRandomWalkStepUniform(
--- a/src/graph/sampling/randomwalks/metapath_randomwalk.h
+++ b/src/graph/sampling/randomwalks/metapath_randomwalk.h
@@ -74,7 +74,7 @@ std::pair<dgl_id_t, bool> MetapathRandomWalkStep(
   if (size == 0)
     return std::make_pair(-1, true);
 
-  // Use a reference to the original array instead of copying 
+  // Use a reference to the original array instead of copying
   // This avoids updating the ref counts atomically from different threads
   // and avoids cache ping-ponging in the tight loop
   const FloatArray &prob_etype = prob[etype];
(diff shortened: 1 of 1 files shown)


src/graph/sampling/randomwalks/metapath_randomwalk.h
```diff
@@ -74,7 +74,10 @@ std::pair<dgl_id_t, bool> MetapathRandomWalkStep(
   if (size == 0)
     return std::make_pair(-1, true);
 
-  FloatArray prob_etype = prob[etype];
+  // Use a reference to the original array instead of copying
+  // This avoids updating the ref counts atomically from different threads
+  // and avoids cache ping-ponging in the tight loop
+  const FloatArray &prob_etype = prob[etype];
   IdxType idx = 0;
   if (IsNullArray(prob_etype)) {
```
BarclayII, line 82: With your PR this branch will never be called I assume?

BarclayII: Makes sense.

BarclayII (approved):

PR #3004, opened 2021-06-10
[CPU, Parallel] parallel_for with default grain size

tpatejko:
This PR introduces implementation of `parallel_for` function with default grain size.

Default grain size is set to 1 in which case `parallel_for` acts as OpenMP parallel for pragma with static scheduling, or can be globally set to a value controlled with environment variable `DGL_PARALLEL_FOR_GRAIN_SIZE`.

The idea behind this implementation is that user could start parallelizing code with `parallel_for` with default grain size, and set grain size either through environment variable or by using `parallel_for` with grain size parameter when performance requirements are not satisfied.

### DIFF
--- a/include/dgl/runtime/parallel_for.h
+++ b/include/dgl/runtime/parallel_for.h
@@ -8,6 +8,8 @@
 
 #include <dmlc/omp.h>
 #include <algorithm>
+#include <string>
+#include <cstdlib>
 
 namespace {
 int64_t divup(int64_t x, int64_t y) {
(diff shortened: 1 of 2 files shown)


BarclayII:
One question for the `parallel_for` implementation: I noticed that the callback function takes in a single argument:
```C++
      for (auto i = begin_tid; i < std::min(end, chunk_size + begin_tid); i++) {
        f(i);
      }
```
If the workload is very lightweight (say adding two vectors), would the function call itself (like pushing and poping from stack) cause a lot of overhead?

I asked this question because I saw PyTorch not looping over the indices; instead the callback function is responsible for looping.  See https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/ParallelOpenMP.h#L57

BarclayII (approved):

PR #3016, opened 2021-06-12
[Performance] Perform to_block on the GPU when the dataloader is created with a GPU `device`.

nv-dlasalle:
## Description
to_block is a very expensive operation in preparing mini-batches, this PR makes it so that if the device for the dataloader is specified at a GPU and the graph is the CPU, the frontier will be sampled on the CPU, but then the sampled edges will be copied to the GPU, and to_block will be run there.

This has a large impact on performance where there is not sufficient CPU power to generate mini-batches for the available GPUs. This happens when either the network is relatively light weight (e.g. GraphSage), or we have multiple GPUs. When running graphsage on 8x V100s, using a fanout of `16,16`, batch size of `4096`, and a hidden dimension of `64`, this resulted in greater than a 4x speedup per-epoch.

## Checklist
- [x] Code is well-documented

These changes are covered by the unit test in `tests/pytorch/test_dataloader.py`, `test_node_dataloader()`.

This also modifies the RGCN and GraphSage examples to run to_block on the GPU by default, when`num_workers == 0`.


BarclayII (approved):
How would you suggest changing the tutorials?  Namely the ones in `tutorials/large` and `tutorials/multi`.

zheng-da (approved):

Issue #2986, opened 2021-06-04
Calling `to_block()` results in duplicate node types -- breaking several parts of heterograph


nv-dlasalle:
## 🐛 Bug

Calling `to_block()` results in duplicate node types, breaking things like edge functions, as assigning features via `ndata` only operates on the `srcnodes`, and not the `dstnodes`.

At least, the implementation of `DGLHeteroGraph.get_ntype_id()` is broken by this:
```
def get_ntype_id(self, ntype):
  ...
  ntid = self._srctypes_invmap.get(ntype, self._dsttypes_invmap.get(ntype, None))
```
As it will only ever choose the source node type, and not the destination node type. There may be other such things that are broken as well.

## To Reproduce

```
import dgl
import torch as th
g = dgl.heterograph({('user','follows','user'): [(0,1), (1,0), (0,2)]})
print("graph ntypes = {}".format(g.ntypes))
b = dgl.to_block(g)
print("block ntypes = {}".format(b.ntypes))
```

Results in:
```
Using backend: pytorch
graph ntypes = ['user']
block ntypes = ['user', 'user']
```

## Expected behavior

My understanding of how heterograph is implemented, is for it to work properly, we would expect one edge type in the block:
```
Using backend: pytorch
graph ntypes = ['user']
block ntypes = ['user']
```

## Environment

 - DGL Version (e.g., 1.0): master branch
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): pytorch 1.8
 - OS (e.g., Linux): ubuntu 18.04
 - How you installed DGL (`conda`, `pip`, source): source
 - Build command you used (if compiling from source): cmake .. -DUSE_CUDA=ON -DCMAKE_BUILD_TYPE=Debug
 - Python version: 3.6
 - CUDA/cuDNN version (if applicable): 11.3
 - GPU models and configuration (e.g. V100): TitanV
 - Any other relevant information:

## Additional context

This is currently a blocker for edge property prediction on a homogeneous graph with sampling.

jermainewang:
This is an expected behavior in current DGL, which I agree is not ideal. The main motivation of this design is to make it easier to port models written for full graph training to codes fit for mini-batch training. For example, consider a model with an loop over the relation types (such as in R-GCN):

```python
def forward(g, feats):
    # g is a DGLGraph; feats is a dictionary of feature tensors
    for srctype, etype, dsttype in g.canonical_etypes:
        # type wise message passing
[...]

PR #3021, opened 2021-06-15
[Feature][Performance][GPU] Multi-gpu tensor

nv-dlasalle:
## Description
Adds support for storing graph features partitioned in GPU memory. This doesn't implement new functionality, but rather better exposes that which was already implemented in the NCCL based NodeEmbedding.

This enables running GraphSage on ogbn-papers100M with the node features entirely in GPU memory, on 8x 16GB GPUs (such as a P3 instance).

## Checklist
- [x] Code is well-documented

## Changes
This add the dgl.distributed.MultiGPUTensor class. It does not implement the standard tensor interface, as all reads from it must be invoked from all GPUs to participate in the collective communication operation, and thus cannot be transparently used.

This changes the NodeEmbedding to use the MultiGPUTensor class, rather than manage the data itself.

This adds a unit test for the MultiGPUTensor class.

This modifies the `/examples/pytorch/graphsage/train_sampling_multi_gpu.py` example to make use of the MultiGPUTensor, and adds ogbn-papers100M as a dataset option.


lmeyerov:
Super cool!

In discussion, two use cases came up:

1. Including edge features -- seemingly more of a docs issue?

2. Interop with cudf/dask-cudf, where etl/cleaning is on-gpu, and we need to do some sort of dlpack call: https://docs.rapids.ai/api/cudf/stable/10min-cudf-cupy.html#Converting-a-cuDF-DataFrame-to-a-CuPy-Array

python/dgl/contrib/multi_gpu_datastore.py
```diff
@@ -0,0 +1,156 @@
...
+    ...                               dev_id,
+    ...                               nccl_comm)
+    >>> split_data.set_global(large_cpu_tensor)
+
+    Then, once it is stored across GPU memory, during training, feature for
+    mini-batches can be fetched via the `get_global()` method. If we have the
+    tensor of mini-batch nodes `input_nodes` we can use the following code to
+    fetch the features for the mini-batch:
+    ...
+    >>> batch_features = split_data.all_gather_row(input_nodes)
```
classicsong, line 42: Do we need to call the `split_data.all_gather_row(input_nodes)` from all the processes simultaneously?

python/dgl/contrib/multi_gpu_datastore.py
```diff
@@ -0,0 +1,156 @@
...
+    Then, once it is stored across GPU memory during training, features for
+    mini-batches can be fetched via the `get_global()` method. If we have the
+    tensor of mini-batch nodes `input_nodes` we can use the following code to
+    fetch the features for the mini-batch:
+    ...
+    >>> batch_features = split_data.all_gather_row(input_nodes)
+    """
+    def __init__(self, shape, dtype, device, comm, partition):
+        """ Create a new Tensor stored across multiple GPUs according to
+        `partition`. This function must be called by all processes.
```
classicsong, line 46:
```suggestion
        `partition`. This function must be called by all processes at the same time.
```

python/dgl/contrib/multi_gpu_datastore.py
```diff
@@ -0,0 +1,156 @@
...
+            The shape of the tensor. The tensor will be partitioned across its
+            first dimension. As a result, dimensionless tensors are not
+            allowed.
+        dtype : DType
+            The backend data type.
+        device : Device
+            The current backend device.
+        comm : nccl.Communicator
+            The NCCL communicator to use.
+        partition : NDArrayPartition
```
classicsong, line 60: The example code in the above does not pass in a partition object.

python/dgl/contrib/pytorch/multi_gpu_dataloader.py
```diff
@@ -0,0 +1,245 @@
...
+            node_frames = []
+            # have to handle src and dst nodes separately in a block
+            for ntype in block.srctypes:
+                index = block.srcnodes[ntype].data[dgl.NID]
+                frame = Frame(num_rows=len(index))
+                for k, v in self._n_feat[ntype].items():
+                    data = _gather_row(v, index)
+                    frame.update_column(k, data)
+                node_frames.append(frame)
+            for ntype in block.dsttypes:
```
classicsong, line 70: What if an ntype appears in both block.srctypes and block.dsttypes?

python/dgl/partition.py
```diff
@@ -497,5 +514,35 @@ def map_to_global(self, idxs, part_id):
...
+        # This mode makes use of locality, so we want to preserve the same
+        # locality for edges. To do this, we find the number of edges
+        # per-partition and use the prefix-sum of this for the range
+        rng = [0]
+        for part in range(graph.num_parts()):
+            idx = partition.get_local_indices(part, ctx=F.cpu())
+            rng.append(rng[-1] + \
+                F.as_scalar(F.sum(graph.out_degrees(u=idx), dim=0)))
+        return NDArrayPartition(array_size=graph.number_of_edges(),
+                                num_parts=partition.num_parts(),
```
classicsong, line 544: What if each node has only one partition but has more than one GPUs?

examples/pytorch/graphsage/train_sampling_multi_gpu.py
```diff
@@ -73,32 +91,48 @@ def run(proc_id, n_gpus, args, devices, data):
...
-        batch_size=args.batch_size,
-        shuffle=True,
-        drop_last=False,
-        num_workers=args.num_workers)
+
+    nccl_comm = None
+    if nccl_id:
+        nccl_comm = nccl.Communicator(n_gpus, proc_id, nccl_id)
+
+    if not args.data_cpu:
```
classicsong, line 111: Do you put the node feature into GPUTensor in this example?

classicsong (commented):
Is the definition of MultiGPUTensor in another PR?

python/dgl/contrib/multi_gpu_datastore.py
nv-dlasalle: Yes, the call cannot return until all processes have called it.

python/dgl/contrib/pytorch/multi_gpu_dataloader.py
nv-dlasalle: This is the case for blocks. In the homogeneous case, we'll have a `_N` for src and dst. See #2986.

I essentially took this from https://github.com/dmlc/dgl/blob/master/python/dgl/utils/internal.py#L793. I had originally just iterated `block.ntypes`, but this breaks for homogeneous graphs due to having the duplicated type.

examples/pytorch/graphsage/train_sampling_multi_gpu.py
nv-dlasalle: Yes, the MultiGPUNodeDataLoader stores all node features split on the GPUs behind the scenes.

python/dgl/partition.py
nv-dlasalle: The number of GPUs must be equal to the number of partitions. Or are you asking if the number of nodes in the graph is less than the number of partitions?

python/dgl/contrib/multi_gpu_datastore.py
```diff
@@ -0,0 +1,156 @@
...
+    nodes. When reading memory from oehr GPUs, the call must be from all
+    GPUs storing the data.
+
+    Example:
+    --------
+
+    It can be created via the shape and data type information from the tensor
+    to split across gpus `large_cpu_tensor` in the below code:
+
+    >>> split_data = MultiGPUTensor(large_cpu_tensor.shape,
```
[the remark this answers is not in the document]
nv-dlasalle: I had forgotten to update this example. The new method is `all_set_global()` to make it clear that it must be called by all processes at the same time.

I did not make it take a tensor as input in the constructor, as a user may also want to only set the local partition's portion of the tensor (e.g. in the case of using DistDGL).

python/dgl/contrib/multi_gpu_datastore.py
nv-dlasalle: Sorry, this is left over from the previous iteration. I've updated the docs to have a partition in this example in 
57b6a43

examples/pytorch/graphsage/train_sampling_multi_gpu.py
```diff
@@ -73,32 +91,48 @@ def run(proc_id, n_gpus, args, devices, data):
...
+            sampler,
+            device=dev_id,
+            comm=nccl_comm,
+            node_feat=train_nfeat,
+            node_label=train_labels,
+            use_ddp=n_gpus > 1,
+            batch_size=args.batch_size,
+            shuffle=True,
+            drop_last=False,
+            num_workers=args.num_workers)
```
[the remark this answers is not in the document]
nv-dlasalle: Yes. It fully wraps the NodeDataLoader, and only operates on its output each iteration.

examples/pytorch/graphsage/train_sampling_multi_gpu.py
```diff
@@ -118,13 +152,17 @@ def run(proc_id, n_gpus, args, devices, data):
...
         # blocks.
-        for step, (input_nodes, seeds, blocks) in enumerate(dataloader):
+        for step, data in enumerate(dataloader):
             if proc_id == 0:
                 tic_step = time.time()
 
-            # Load the input features as well as output labels
-            batch_inputs, batch_labels = load_subtensor(train_nfeat, train_labels,
-                                                        seeds, input_nodes, dev_id)
+            if len(data) == 3:
```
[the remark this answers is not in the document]
nv-dlasalle: MultiGPUNodeDataLoader is a child of NodeDataLoader won't work. I could do the inverse:
```
if not isinstance(dataloader, MultiGPUNodeDataLoader):
```
But my thoughts were since we're unpacking `data`, we should depend on `data` to make the decision rather than on `dataloader` which isn't directly used in this statement. If `node_feat` and `node_label` weren't specified to `MultiGPUNodeDataLoader`, then it would also return only 3 arguments per-iteration and would work in that code path.

python/dgl/backend/mxnet/tensor.py
```diff
@@ -144,6 +144,10 @@ def asnumpy(input):
 def copy_to(input, ctx, **kwargs):
     return input.as_in_context(ctx)
 
+def is_pinned(input):
+    # Utilizing pinned memory from tensorflow isn't yet supported by DGL.
```
Kh4L, line 148:
```suggestion
    # Utilizing pinned memory from MXNet isn't yet supported by DGL.
```

jermainewang:
@nv-dlasalle I changed this PR to draft as it perhaps needs some major refactoring. Also, this feature could be a perfect example to showcase the design of `GraphStorage` and `FeatureStorage` but may need to come after a formal specification of the components (see feature request #4295 ). If you think a new PR will probably be created in the future, you could also close this. Thanks for the great effort!

PR #3024, opened 2021-06-15
[CPU][Kernel] Single socket spmm

sanchit-misra:
## Description
This PR contains the single socket optimizations of SpMM for Xeon as mentioned in our DistGNN paper: https://arxiv.org/abs/2104.06700

## Checklist
- [ ] To the my best knowledge, examples get faster or equal performance and accuracy is not affected.

## Changes
Provided Xeon optimized implementations of SpMMSumCsr() and SpMMCmpCsr(). We have observed up to 4.4x speedup on the SpMM kernel without change in accuracy.


src/array/cpu/spmm.h
```diff
@@ -2,6 +2,10 @@
  *  Copyright (c) 2020 by Contributors
  * \file array/cpu/spmm.h
  * \brief SPMM CPU kernel function header.
+ * \author Sanchit Misra <sanchit.misra@intel.com>,
+ *         Ramanarayan Mohanty <ramanarayan.mohanty@intel.com>,
+ *         Vasimuddin Md <vasimuddin.md@intel.com>,
+ *         Sasikanth Avancha <sasikanth.avancha@intel.com>
```
zheng-da, line 8: it's better to add the original authors as well.

zheng-da: actually, i think it's better to move the block spmm code to another header file. the code will also look cleaner.

src/array/cpu/spmm.h
```diff
@@ -12,15 +16,568 @@
...
+#if SPMM_LOG_INFO
+  endTick = __rdtsc();
+  printf("spmmcmpcsr nthreads = %d, LLC_SIZE = %d\n", nthreads, LLC_SIZE);
+  printf("M = %d, K = %d, N = %d, use_lhs = %d, use_rhs = %d\n", M, K, N, Op::use_lhs, Op::use_rhs);
+  printf("total_nnz = %d, avgDegree = %f\n", total_nnz, avgDegree);
+  printf("has_idx = %d\n", has_idx);
+  printf("nnz_prob = %lf\n", nnz_prob);
+  printf("K_BLOCK_SIZE = %d, M_BLOCK_SIZE = %d\n", K_BLOCK_SIZE, M_BLOCK_SIZE);
+  printf("num_K_blocks = %d, num_M_blocks = %d\n", num_K_blocks, num_M_blocks);
+  printf("stage0: %ld ticks\n", endTick - startTick);
```
zheng-da, line 533: here it prints too much information.

zheng-da:
could you also provide some performance results to help people understand the benefit of this PR?

zheng-da:
i see the code constructs block sparse matrix in every sparse matrix multiplication. do you know how much is the overhead of block sparse matrix construction? can we get further speedup if we can cache the block sparse matrix?

src/array/cpu/spmm.h
sanchit-misra: Absolutely. I didn't put their names only because I didn't know the name of the original authors of this file. But I like your suggestion of moving block spmm code to another header file. I will do that. Should that header file be in src/array/cpu/ ? How about src/array/cpu/spmm_blocking.h?

src/array/cpu/spmm.h
sanchit-misra: All of this is disabled by a macro (SPMM_LOG_INFO) and will be enabled only for debugging.

sanchit-misra:
> could you also provide some performance results to help people understand the benefit of this PR?

Sure!

sanchit-misra:
> i see the code constructs block sparse matrix in every sparse matrix multiplication. do you know how much is the overhead of block sparse matrix construction? can we get further speedup if we can cache the block sparse matrix?

Block sparse matrix construction is only invoked in case of denser matrices with decently large dimensions. For example, for Reddit with dim = 602, it is invoked and consumes 10-12% of the time. But given that DGL now only performs SpMM on Reddit hidden layer with dim = 16, blocking is not used. So, in case of denser matrices with large enough dimensions for full batch training, it will help by about 10% if we cache the blocks. 

It is also not used for minibatch training due to sparse matrices and we can't cache them anyway.

src/array/cpu/spmm.h
zheng-da: `src/array/cpu/spmm_blocking.h` sounds good to me.

zheng-da:
i'm just curious. it's better to not cache it for now, which makes the implementation more compatible with DGL's current API.

sanchit-misra:
Hi @zheng-da I have moved the optimized code to a new header file. I have named it "src/array/cpu/spmm_blocking_libxsmm.h" because I thought "spmm_blocking.h" did not capture all its improvements. 
I am running tests again and will share the speedup numbers as soon as I have them.

src/array/cpu/spmm.h
jermainewang: Could you use DGL's log macro for code consistency?

sanchit-misra: Sure! What is the preferred way to add debugging logs? Should I use DLOG(INFO) or use:
#ifdef DEBUG
LOG(INFO) << ...;
#endif

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,576 @@
...
+  }
+  else if(std::is_same<Op, op::CopyLhs<DType>>::value)
+  {
+    opredop_flags = LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_OP_COPY;
+  }
+  else if(std::is_same<Op, op::CopyRhs<DType>>::value)
+  {
+    opredop_flags = LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_OP_COPY;
+  }
+  if(std::is_same<Op, op::CopyLhs<DType>>::value)
```
tpatejko, line 207: Why is this condition done twice, here and in line 199, and `opredop_flags` or'ed?

sanchit-misra:
> could you also provide some performance results to help people understand the benefit of this PR?

Please find the graphs for multiple workloads below. For baseline, we have enabled the xbyak optimization.

Full batch Training:
image
image
image
image


Minibatch Training:
image
image
image
image

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,576 @@
...
+
+  SpMMFreeBlocks(block_csr_array, num_M_blocks, num_K_blocks, Op::use_lhs, Op::use_rhs);
+#if SPMM_LOG_INFO
+  endTick = __rdtsc();
+  printf("stage4: %ld ticks\n", endTick - startTick);
+#endif
+}
+ 
+template <typename IdType, typename DType, typename Op, typename Cmp>
+void SpMMCmpCsrOpt(
```
tpatejko, line 471: The function is very similar to `SpMMSumCsrOpt`. Is there a way to refactor both functions? They could be turned into a single unified and configurable function, or shared code could be moved to separate functions, or both.

sanchit-misra: Sure, I can turn it into a single unified and configurable function.

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,577 @@
...
+#include <x86intrin.h>
+#endif
+#include <omp.h>
+
+namespace dgl {
+namespace aten {
+namespace cpu {
+
+template <typename IdType, typename DType>
+  struct CSRM{
```
jermainewang, line 33: No extra indentation before struct definition.

jermainewang: Also suggest having a more expressive name than `CSRM` to indicate its purpose.

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,577 @@
...
+    IdType num_rows;
+    IdType num_cols;
+    IdType *indptr;
+    IdType *indices;
+    DType *data;
+  };
+
+int32_t get_llc_size() {
```
jermainewang, line 41: Function names follow camel style.
```suggestion
int32_t GetLLCSize() {
```

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,577 @@
...
+}
+
+
+template <typename IdType, typename DType>
+inline void SpMMCreateBlocks(
+    const CSRMatrix& csr,
+    CSRM<IdType, IdType> *block_csr_array,
+    IdType num_M_blocks,
+    IdType num_K_blocks,
+    IdType M_BLOCK_SIZE,
```
jermainewang, line 54: Variable names generally follow snake style.
```suggestion
    IdType m_block_size,
```

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,577 @@
...
+    IdType K_BLOCK_SIZE,
+    DType *C, IdType N,
+    bool use_lhs, bool use_rhs) {
+
+  const IdType M = csr.num_rows;
+  const IdType K = csr.num_cols;
+  IdType* IndPtr = csr.indptr.Ptr<IdType>();
```
jermainewang, line 61:
```suggestion
  IdType* indptr = csr.indptr.Ptr<IdType>();
```

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,577 @@
...
+  if(num_K_blocks > 1)
+  {
+    IdType *indptr = (IdType *)aligned_alloc(64, (M_BLOCK_SIZE + 1) * num_M_blocks * num_K_blocks * sizeof(IdType));
+
+#pragma omp parallel
+    {
+      IdType *my_cur_col_id = (IdType *)aligned_alloc(64, 2 * M_BLOCK_SIZE * sizeof(IdType));
+
+#pragma omp for
+      for(IdType m = 0; m < num_M_blocks; m++)
```
jermainewang, line 79: Add a space between `for` and parenthesis:
```suggestion
      for (IdType m = 0; m < num_M_blocks; m++)
```

src/array/cpu/spmm_blocking_libxsmm.h
jermainewang: This `if` is sneaky, esp. when there are a series of if-else above. Please add some comments between to explain it.
```c++
// Firstly, set flags for ....
if(std::is_same<Op, op::Add<DType>>::value) {
  ...
} else if (...) {
  ...
}

// Secondly, set flags for ...
if (std::is_same<Op, op::CopyLhs<DType>>::value) {
  ...
}
...
```

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,577 @@
...
+      opredop_flags = (libxsmm_meltw_opreduce_vecs_flags)(opredop_flags | LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_RECORD_ARGOP_OFF_VEC_0);
+    }
+    if(Op::use_rhs)
+    {
+      opredop_flags = (libxsmm_meltw_opreduce_vecs_flags)(opredop_flags | LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_RECORD_ARGOP_OFF_VEC_1);
+    }
+  }
+  libxsmm_meltwfunction_opreduce_vecs_idx kernel=nullptr;
+  if(std::is_same<DType, uint16_t>::value)
+    kernel = libxsmm_dispatch_meltw_opreduce_vecs_idx(N, &_ld, &_ld, LIBXSMM_DATATYPE_BF16, LIBXSMM_DATATYPE_BF16, (sizeof(IdType) == 8) ? LIBXSMM_DATATYPE_I64 : LIBXSMM_DATATYPE_I32, opredop_flags);
```
jermainewang, line 245: This line might be too long. Our line limit is 100 characters.

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,577 @@
...
+  else if(std::is_same<DType, DType>::value)
+    kernel = libxsmm_dispatch_meltw_opreduce_vecs_idx(N, &_ld, &_ld, LIBXSMM_DATATYPE_F32, LIBXSMM_DATATYPE_F32, (sizeof(IdType) == 8) ? LIBXSMM_DATATYPE_I64 : LIBXSMM_DATATYPE_I32, opredop_flags);
+  if(kernel == nullptr)
+  {
+    LOG(FATAL) << "libxsmm Op-redop kernel is nullptr! Bailing...\n";
```
jermainewang, line 250: No need to have the `\n` at the end. The `LOG` macro automatically prints new line.

jermainewang (changes requested):
First batch of suggestions on coding style to help pass the link check.

jermainewang:
Hi @sanchit-misra , thank you very much for the contribution. The overall change makes sense but I may need more time to understand the details. I saw the link check has failed so I put down some suggestions on the coding style to help you pass that. Generally, we follow the [Google Coding Style](https://google.github.io/styleguide/cppguide.html). Our line limit is 100 characters.
You can click the CI details button to see the lint errors. Once that is done, you also need to pass the building phase and testing phase. In the meanwhile I will study your code and give more comments.

BTW, I saw you added a new submodule libxsmm. Is it the snapshot of the latest stable version? I saw the [latest release](https://github.com/hfp/libxsmm/releases/tag/1.16.1) is 1.16.1. If all you need are available in it, I will suggest check out that version.

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,577 @@
...
+    {
+      opredop_flags = (libxsmm_meltw_opreduce_vecs_flags)(opredop_flags | LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_RECORD_ARGOP_OFF_VEC_0);
+    }
+    if(Op::use_rhs)
+    {
+      opredop_flags = (libxsmm_meltw_opreduce_vecs_flags)(opredop_flags | LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_RECORD_ARGOP_OFF_VEC_1);
+    }
+  }
+  libxsmm_meltwfunction_opreduce_vecs_idx kernel=nullptr;
+  if(std::is_same<DType, uint16_t>::value)
```
tpatejko, line 244: This if-else statement is a bit hard to read. It's long and contains repeated code. I think it could be turned into a function/lambda and `kernel` variable could be defined at the declaration point without initializing it with `nullptr`

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Unified the two functions into one.

src/array/cpu/spmm.h
```diff
@@ -42,52 +45,75 @@ void SpMMSumCsr(const BcastOff& bcast, const CSRMatrix& csr, NDArray ufeat,
   const DType* W = efeat.Ptr<DType>();
   int64_t dim = bcast.out_len, lhs_dim = bcast.lhs_len, rhs_dim = bcast.rhs_len;
   DType* O = out.Ptr<DType>();
+  assert(indptr != nullptr);
```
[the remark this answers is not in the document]
sanchit-misra: Done

src/array/cpu/spmm.h
```diff
@@ -42,52 +45,75 @@ void SpMMSumCsr(const BcastOff& bcast, const CSRMatrix& csr, NDArray ufeat,
...
+  assert(O != nullptr);
+  if(Op::use_lhs)
+  {
+      assert(indices != nullptr);
+      assert(X != nullptr);
```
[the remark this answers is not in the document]
sanchit-misra: Done

src/array/cpu/spmm.h
```diff
@@ -42,52 +45,75 @@ void SpMMSumCsr(const BcastOff& bcast, const CSRMatrix& csr, NDArray ufeat,
   const DType* W = efeat.Ptr<DType>();
   int64_t dim = bcast.out_len, lhs_dim = bcast.lhs_len, rhs_dim = bcast.rhs_len;
   DType* O = out.Ptr<DType>();
+  assert(indptr != nullptr);
+  assert(O != nullptr);
+  if(Op::use_lhs)
```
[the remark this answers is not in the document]
sanchit-misra: Done

src/array/cpu/spmm.h
```diff
@@ -42,52 +45,75 @@ void SpMMSumCsr(const BcastOff& bcast, const CSRMatrix& csr, NDArray ufeat,
...
-                            ? asm_kernel_ptr.get()
-                            : nullptr;
-  if (cpu_spec && dim > 16 && !bcast.use_bcast) {
+#ifdef USE_LIBXSMM  
+  bool special_condition = bcast.use_bcast || (Op::use_lhs && (dim != lhs_dim)) || (Op::use_rhs && (dim != rhs_dim));
+  if(!special_condition)
+  {
+      SpMMSumCsrOpt<IdType, DType, Op>(bcast, csr, ufeat, efeat, out);
+  } else {
+#endif
```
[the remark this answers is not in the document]
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,577 @@
...
+
+#if !defined(_WIN32)
+#ifdef USE_AVX
+#ifdef USE_LIBXSMM
+#include <unistd.h>
+#include <libxsmm.h>
+#ifdef DEBUG
+#include <x86intrin.h>
+#endif
+#include <omp.h>
```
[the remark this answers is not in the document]
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Modified to M_block_size to emphasize that it is the block size corresponding to M.

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: There is already a variable named indptr

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Done

sanchit-misra:
> Hi @sanchit-misra , thank you very much for the contribution. The overall change makes sense but I may need more time to understand the details. I saw the link check has failed so I put down some suggestions on the coding style to help you pass that. Generally, we follow the [Google Coding Style](https://google.github.io/styleguide/cppguide.html). Our line limit is 100 characters.
> You can click the CI details button to see the lint errors. Once that is done, you also need to pass the building phase and testing phase. In the meanwhile I will study your code and give more comments.
> 
> BTW, I saw you added a new submodule libxsmm. Is it the snapshot of the latest stable version? I saw the [latest release](https://github.com/hfp/libxsmm/releases/tag/1.16.1) is 1.16.1. If all you need are available in it, I will suggest check out that version.

Hi @jermainewang Thanks a lot for your comments. I have made some of the coding style changes suggested by you. Will try to make all the changes required to pass the link test today. The libxsmm release 1.16.1 is very old and does not have any of the things I am using it for. So, I am using a recent commit.

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,517 @@
...
+    opredop_flags = LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_OP_COPY;
+  }
+  else if (std::is_same<Op, op::CopyRhs<DType>>::value) {
+    opredop_flags = LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_OP_COPY;
+  }
+  if (std::is_same<Op, op::CopyLhs<DType>>::value) {
+    opredop_flags = (libxsmm_meltw_opreduce_vecs_flags)(opredop_flags |
+                     LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_OPORDER_VECIDX_VECIN);
+  }
+  else if (std::is_same<Op, op::CopyRhs<DType>>::value) {
```
IzabelaMazur, line 198: Is there any reason to duplicate the if statement from line 191?

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,517 @@
...
+  else if (std::is_same<Op, op::Div<DType>>::value) {
+    opredop_flags = LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_OP_DIV;
+  }
+  else if (std::is_same<Op, op::CopyLhs<DType>>::value) {
+    opredop_flags = LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_OP_COPY;
+  }
+  else if (std::is_same<Op, op::CopyRhs<DType>>::value) {
+    opredop_flags = LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_OP_COPY;
+  }
+  if (std::is_same<Op, op::CopyLhs<DType>>::value) {
```
IzabelaMazur, line 194: Is there any reason to duplicate the if statement from line 188?

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: I have added comments as suggested by @jermainewang . Hopefully, it is easier to read now. The kernel variable is initialized to nullptr so that we can test on line 237 whether it got initialized or not.

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Added the comments and split the longer lines into multiple lines.

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Yes, I have added comments for the same in the code.

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Yes, I have added comments for the same in the code.

sanchit-misra: We are setting different parts of the opredop_flag in these blocks of if statements.

sanchit-misra:
Hi @jermainewang @IzabelaMazur @tpatejko @zheng-da,
I have cleared the lint tests and I believe I have addressed all your comments. Please let me know if I need to do anything else.

src/array/cpu/spmm_blocking_libxsmm.h
BarclayII: Then I suggest renaming either one to a more informative name (e.g. `IndPtr -> indptr_unaligned`).  The difference between `IndPtr` and `indptr` is not immediately clear.

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,500 @@
...
+  if (kernel == nullptr) {
+    LOG(FATAL) << "libxsmm Op-redop kernel is nullptr!";
+  }
+  return kernel;
+}
+
+template <typename IdType, typename DType>
+inline void SpMMBlockwiseOpSum(
+    CSRMatrixInternal<IdType, IdType> *block_csr_array,
+    DType *B, DType *E, DType *C, bool has_idx, IdType N,
```
BarclayII, line 246: I assume that B and E should be const pointers?

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,500 @@
...
+    IdType num_M_blocks, IdType num_K_blocks, IdType M_block_size,
+    libxsmm_meltwfunction_opreduce_vecs_idx kernel) {
+
+  DType (*in_matrix1)[N] = (DType (*)[N])B;
```
BarclayII, line 250: Looks like this is casting the DType pointer B (which contains the source features, as an array of DType data) to an array of N DType pointers?  If so, would you mind explaining why it works?

src/array/cpu/spmm.h
```diff
@@ -42,52 +45,73 @@ void SpMMSumCsr(const BcastOff& bcast, const CSRMatrix& csr, NDArray ufeat,
...
+        DType* out_off = O + rid * dim;
+        for (IdType j = row_start; j < row_end; ++j) {
+          const IdType cid = indices[j];
+          const IdType eid = has_idx ? edges[j] : j;
+          cpu_spec->run(out_off, X + cid * lhs_dim, W + eid * rhs_dim, dim);
+        }
       }
-    }
-  } else {
+    } else {
```
BarclayII, line 87: The nesting is getting deep.  I'm thinking of functionizing it into 
```C++
    if (cpu_spec && dim > 16 && !bcast.use_bcast) {
      SpMMSumCsrAVX(...);
    } else {
#endif  // USE_AVX
#endif  // _WIN32
      SpMMSumCsrNaive(...);
```

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,500 @@
...
+  IdType* IndPtr = csr.indptr.Ptr<IdType>();
+  CHECK_NOTNULL(IndPtr);
+  int total_nnz = IndPtr[M];
+  if (M <= 0 || K <= 0 || N <= 0 || total_nnz <= 0) return;
+
+  float avgDegree = total_nnz * 1.0 / M;
+  float nnz_prob = avgDegree / K;
+
+  IdType K_block_size = llc_size / (N * sizeof(DType) * nnz_prob * 500);
+  IdType M_block_size = M / (nthreads * 20);
```
BarclayII, line 394: Why 500 and 20?

jermainewang:
CI failed in building stage. It cannot find libxsmm.h. Could you please fix it?

sanchit-misra:
> CI failed in building stage. It cannot find libxsmm.h. Could you please fix it?

@jermainewang : Sure, let me check that.

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: This is casting B to an array of pointers pointing to arrays of size N. This is equivalent to casting B into a 2D array representation where the row size is N. This is done because libxsmm expects input arrays to have that 2D representation.

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Yes. Done.

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Named as macros

src/array/cpu/spmm_blocking_libxsmm.h
BarclayII: Thanks.  Now I learned a new syntax...

src/array/cpu/spmm.h
jermainewang: +1

jermainewang:
I am reading the code and find that the core kernel is generated by `libxsmm_dispatch_meltw_opreduce_vecs_idx` , but I cannot find any docstring describing this API from https://libxsmm.readthedocs.io/en/latest/. Could you please tell me where can I find it?

jermainewang:
Another question is about the performance breakdown. In the [main operator logic](https://github.com/dmlc/dgl/blob/3eaa153c515c6f3d5df83aefd490cdf00af1b5e6/src/array/cpu/spmm_blocking_libxsmm.h#L359), you have four stages (thanks for documenting every stage!):
* Tile the CSR matrix to roughly make sure that the column tiles fit into LLC and the row tiles are assigned to OMP threads.
* Create kernel using libxsmm.
* Launch kernel on the tiled CSR matrix.
* Free the tiled CSR matrix data.

_Have you benchmarked how much time each part takes?_ This question will be related to whether we could cache the generated kernel and tiled CSR matrix in certain application scenarios.

sanchit-misra:
@jermainewang It seems that the build went past "not finding libxsmm.h" but is stuck at the following error. Could you please help me understand what the following error is about? It seems that it is coming from Visual Studio. Would the correct approach be to disable our blocking and libxsmm based optimizations for Visual Studio (MSVC)? 


14>C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(231,5): error MSB6006: "cmd.exe" exited with code 9009. [C:\Jenkins\workspace\dgl_PR-3024\build\libxsmm.vcxproj]

    14>Done Building Project "C:\Jenkins\workspace\dgl_PR-3024\build\libxsmm.vcxproj" (default targets) -- FAILED.

     9>Done Building Project "C:\Jenkins\workspace\dgl_PR-3024\build\libxsmm.vcxproj.metaproj" (default targets) -- FAILED.

sanchit-misra:
> I am reading the code and find that the core kernel is generated by `libxsmm_dispatch_meltw_opreduce_vecs_idx` , but I cannot find any docstring describing this API from https://libxsmm.readthedocs.io/en/latest/. Could you please tell me where can I find it?

Let me get back to you on that.

CMakeLists.txt
```diff
@@ -210,6 +218,15 @@ if(NOT MSVC)
   list(APPEND DGL_LINKER_LIBS metis)
 endif(NOT MSVC)
 
+# Compile LIBXSMM
+if(USE_LIBXSMM)
+  add_custom_target(libxsmm COMMAND make realclean COMMAND make -j BLAS=0
```
VoVAllen, line 223: The error is saying there's no `make` on the windows. I think you can just skip build LIBXSMM if it's on windows?

VoVAllen:
```
    14>CustomBuild:

         'make' is not recognized as an internal or external command,

         operable program or batch file.
```

CMakeLists.txt
sanchit-misra: What is the best way to check if it's on windows? Will "NOT MSVC" suffice?

VoVAllen: Yes. That's how we did now

sanchit-misra: Ok great. Will do the same.

sanchit-misra: It is building now, but it seems all the tests that are using "tests/compute/test_sparse.py" are failing. If there was some bug in the new SpMM code, then some of the tests would fail, not all. Since all the tests are failing, I am wondering if it is some setup issue. In the sense that the tests are not being built correctly. How do I run these tests? I tried ./runUnitTests at my end and all the tests passed.

BarclayII: `runUnitTests` runs the C++ unit tests, and indeed they are all correct according to CI since you don't have new C++ unit tests (in `tests/cpp`).  The unit tests are failing on Python side, which needs `pytest` to run:
```
# Assuming that your current directory is the Git root directory
export PYTHONPATH=$PWD/tests
pytest tests/compute/test_sparse.py
# Or you can run pytest with Python builtin debugger, though I'm not sure if that will help for C++ debugging
pytest --pdb tests/compute/test_sparse.py
```

jermainewang:
> > I am reading the code and find that the core kernel is generated by `libxsmm_dispatch_meltw_opreduce_vecs_idx` , but I cannot find any docstring describing this API from https://libxsmm.readthedocs.io/en/latest/. Could you please tell me where can I find it?
> 
> Let me get back to you on that.

Any update? I saw you've cleared the unit test errors. Would like to proceed and merge it to 0.7.

sanchit-misra:
> > > I am reading the code and find that the core kernel is generated by `libxsmm_dispatch_meltw_opreduce_vecs_idx` , but I cannot find any docstring describing this API from https://libxsmm.readthedocs.io/en/latest/. Could you please tell me where can I find it?
> > 
> > 
> > Let me get back to you on that.
> 
> Any update? I saw you've cleared the unit test errors. Would like to proceed and merge it to 0.7.

Since the API is under continuous development, there is no documentation of libxsmm yet. But I can give you a high level summary of what it does. 
libxsmm_dispatch_meltw_opreduce_vecs_idx creates a JIT'ed kernel. Given a node u, the kernel performs an elementwise "op" on the embeddings of the neighbors and/or the edges incident on u. Subsequently, it performs an elementwise "redop" on all such embeddings created and stores into the embedding of node u. It uses a SIMD and a cache efficient design and also provides support to enable software prefetching if needed. For IdType, it supports INT32 and INT64. For DType, it supports BF16 and FP32. It supports all the "ops" and "redops" supported by DGL. Once a kernel is generated by libxsmm_dispatch_meltw_opreduce_vecs_idx, it is cached for the entire duration of the execution of a program so that subsequently if the kernel is needed again, it just returns the cached copy.

jermainewang:
> > > > I am reading the code and find that the core kernel is generated by `libxsmm_dispatch_meltw_opreduce_vecs_idx` , but I cannot find any docstring describing this API from https://libxsmm.readthedocs.io/en/latest/. Could you please tell me where can I find it?
> > > 
> > > 
> > > Let me get back to you on that.
> > 
> > 
> > Any update? I saw you've cleared the unit test errors. Would like to proceed and merge it to 0.7.
> 
> Since the API is under continuous development, there is no documentation of libxsmm yet. But I can give you a high level summary of what it does.
> libxsmm_dispatch_meltw_opreduce_vecs_idx creates a JIT'ed kernel. Given a node u, the kernel performs an elementwise "op" on the embeddings of the neighbors and/or the edges incident on u. Subsequently, it performs an elementwise "redop" on all such embeddings created and stores into the embedding of node u. It uses a SIMD and a cache efficient design and also provides support to enable software prefetching if needed. For IdType, it supports INT32 and INT64. For DType, it supports BF16 and FP32. It supports all the "ops" and "redops" supported by DGL. Once a kernel is generated by libxsmm_dispatch_meltw_opreduce_vecs_idx, it is cached for the entire duration of the execution of a program so that subsequently if the kernel is needed again, it just returns the cached copy.

@sanchit-misra Thanks for the explanation. It will be great if you could include that in the code docstring.

sanchit-misra:
> > > > > I am reading the code and find that the core kernel is generated by `libxsmm_dispatch_meltw_opreduce_vecs_idx` , but I cannot find any docstring describing this API from https://libxsmm.readthedocs.io/en/latest/. Could you please tell me where can I find it?
> > > > 
> > > > 
> > > > Let me get back to you on that.
> > > 
> > > 
> > > Any update? I saw you've cleared the unit test errors. Would like to proceed and merge it to 0.7.
> > 
> > 
> > Since the API is under continuous development, there is no documentation of libxsmm yet. But I can give you a high level summary of what it does.
> > libxsmm_dispatch_meltw_opreduce_vecs_idx creates a JIT'ed kernel. Given a node u, the kernel performs an elementwise "op" on the embeddings of the neighbors and/or the edges incident on u. Subsequently, it performs an elementwise "redop" on all such embeddings created and stores into the embedding of node u. It uses a SIMD and a cache efficient design and also provides support to enable software prefetching if needed. For IdType, it supports INT32 and INT64. For DType, it supports BF16 and FP32. It supports all the "ops" and "redops" supported by DGL. Once a kernel is generated by libxsmm_dispatch_meltw_opreduce_vecs_idx, it is cached for the entire duration of the execution of a program so that subsequently if the kernel is needed again, it just returns the cached copy.
> 
> @sanchit-misra Thanks for the explanation. It will be great if you could include that in the code docstring.

Done!

src/array/cpu/spmm.h
sanchit-misra: Done.

sanchit-misra:
> Another question is about the performance breakdown. In the [main operator logic](https://github.com/dmlc/dgl/blob/3eaa153c515c6f3d5df83aefd490cdf00af1b5e6/src/array/cpu/spmm_blocking_libxsmm.h#L359), you have four stages (thanks for documenting every stage!):
> 
> * Tile the CSR matrix to roughly make sure that the column tiles fit into LLC and the row tiles are assigned to OMP threads.
> * Create kernel using libxsmm.
> * Launch kernel on the tiled CSR matrix.
> * Free the tiled CSR matrix data.
> 
> _Have you benchmarked how much time each part takes?_ This question will be related to whether we could cache the generated kernel and tiled CSR matrix in certain application scenarios.

Yes, we have. 
* We have seen tiling take up to 6% of the total time of SpMM, but that worst case scenario gets activated very rarely. In other cases, it takes negligible amount of time. 
* Creation of libxsmm kernel may take about 1 milliseconds in the worst case in the first invocation but the kernel already gets cached. Therefore, any further invocations for the same type of kernel take < 10 microseconds.
* Performing SpMM on the tiled CSR matrix consumes most of the time.
* Free is also typically negligible taking not more than a few microseconds. But in very rare scenarios it could take longer thus consuming up to 2-3% of the total time.

jermainewang:
@dgl-bot run c5.9xlarge kernel dmlc/master

sanchit-misra:
Hi @jermainewang,

The regression tests seem to be showing that while our PR is faster by up to 6.7x in many cases, in some cases it is also slower than the master branch.
Nearly 10 of the tests failing are from gsddmm even though we did not change anything in gsddmm. Could you please help me understand why these are failing?
There are another 10 tests that are failing in gspmm. They are all u_mul_e_sum type. Could you please let me know how I can run these individual tests so as to figure out what went wrong?
Some of the tests don't have any output. Is that a problem?

VoVAllen:
Reddit dataset in regression test has some problems. Not related to this pull request.

VoVAllen:
Regression test is at https://github.com/dmlc/dgl/blob/master/benchmarks/benchmarks/kernel/bench_gspmm_copy_u.py. And you can call the function directly by comment out those decorators. 

The utils module is at https://github.com/dmlc/dgl/blob/master/benchmarks/benchmarks/utils.py. You can append the path to the `os.path` and import utils directly

sanchit-misra:
> Regression test is at https://github.com/dmlc/dgl/blob/master/benchmarks/benchmarks/kernel/bench_gspmm_copy_u.py. And you can call the function directly by comment out those decorators.
> 
> The utils module is at https://github.com/dmlc/dgl/blob/master/benchmarks/benchmarks/utils.py. You can append the path to the `os.path` and import utils directly

Thanks @VoVAllen. It seems from the output of regression tests, that tests in bench_gspmm_copy_u.py is fine.
But there are issues in https://github.com/dmlc/dgl/blob/master/benchmarks/benchmarks/kernel/bench_gspmm_u_mul_e_sum.py
Is that the correct understanding?

VoVAllen:
The result of ogbn-proteins in bench_gspmm_copy_u seems much worse than before. And ogbn-proteins is a very dense graph, which is a bit different from other dataset. Dataset stats can be got at https://ogb.stanford.edu/docs/nodeprop/#ogbn-proteins

ksadowski13:
Did you run regression tests with flag / env variable DGL_CPU_INTEL_KERNEL_ENABLED=1? If not, is it possible to rerun tests with this flag and compare results? It will give good historical overview regarding kernel optimizations.

sanchit-misra:
> The result of ogbn-proteins in bench_gspmm_copy_u seems much worse than before. And ogbn-proteins is a very dense graph, which is a bit different from other dataset. Dataset stats can be got at https://ogb.stanford.edu/docs/nodeprop/#ogbn-proteins

I am a little confused. The csv file reports GFLOPS numbers for this PR and master branch. I am guessing higher is better, right? So, for gspmm_copy_u with proteins datasets, the GFLOPS for this PR are 3.3-6.7x higher than master. So, this PR is much better than master. Or am I reading it wrong?

sanchit-misra:
> Did you run regression tests with flag / env variable DGL_CPU_INTEL_KERNEL_ENABLED=1? If not, is it possible to rerun tests with this flag and compare results? It will give good historical overview regarding kernel optimizations.

The regression tests were run by dgl bot.

ksadowski13:
> > Did you run regression tests with flag / env variable DGL_CPU_INTEL_KERNEL_ENABLED=1? If not, is it possible to rerun tests with this flag and compare results? It will give good historical overview regarding kernel optimizations.
> 
> The regression tests were run by dgl bot.

@jermainewang @VoVAllen Could you check if dgl bot is running benchmarks with flag / env variable DGL_CPU_INTEL_KERNEL_ENABLED=1? It will be great to have comparison regarding xbyak optimization which will give great insights for future  kernel optimizations.

VoVAllen:
@sanchit-misra My bad. You are right

VoVAllen:
@ksadowski13 Sure. I'll manually test that and get the result back later

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+  // For rhs, we also set whether to use implicit indices or provided indices.
+  if (std::is_same<Op, op::CopyLhs<DType>>::value) {
+    opredop_flags = (libxsmm_meltw_opreduce_vecs_flags)(opredop_flags |
+                     LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_OPORDER_VECIDX_VECIN);
+  } else if (std::is_same<Op, op::CopyRhs<DType>>::value) {
+    opredop_flags = (libxsmm_meltw_opreduce_vecs_flags)(opredop_flags |
+                     LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_OPORDER_VECIN_VECIDX);
+    if (!has_idx) {
+      opredop_flags = (libxsmm_meltw_opreduce_vecs_flags)(opredop_flags |
+                       LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_IMPLICIT_INDEXED_VECIDX);
```
jermainewang, line 233: Why no need to set `LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_INDEXED_VEC` flag as in L238-241?

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+    }
+    if (Op::use_rhs) {
+      opredop_flags = (libxsmm_meltw_opreduce_vecs_flags)(opredop_flags |
+                       LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_RECORD_ARGOP_OFF_VEC_1);
+    }
+  }
+  libxsmm_meltwfunction_opreduce_vecs_idx kernel = nullptr;
+  if (std::is_same<DType, uint16_t>::value) {
+    kernel = libxsmm_dispatch_meltw_opreduce_vecs_idx(
+               N, &_ld, &_ld, LIBXSMM_DATATYPE_BF16, LIBXSMM_DATATYPE_BF16,
```
jermainewang, line 262: Is this for half precision? If so, we haven't officially supported it yet for CPU (see https://github.com/dmlc/dgl/blob/195f99362d883f8b6d131b70a7868a537e55b786/src/array/cpu/spmm_binary_ops.h#L149). You could safely delete this branch. However, we do allow double as DType and I don't see the branch to handle that.

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+               (sizeof(IdType) == 8) ? LIBXSMM_DATATYPE_I64 : LIBXSMM_DATATYPE_I32, opredop_flags);
+  } else if (std::is_same<DType, float>::value) {
+    kernel = libxsmm_dispatch_meltw_opreduce_vecs_idx(
+               N, &_ld, &_ld, LIBXSMM_DATATYPE_F32, LIBXSMM_DATATYPE_F32,
+               (sizeof(IdType) == 8) ? LIBXSMM_DATATYPE_I64 : LIBXSMM_DATATYPE_I32, opredop_flags);
+  }
+  if (kernel == nullptr) {
+    LOG(FATAL) << "libxsmm Op-redop kernel is nullptr!";
```
jermainewang, line 270: `LOG(FATAL) << "Failed to generate libxsmm kernel for the SpMM operation!";`

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+  {
+    for (IdType k = 0; k < num_K_blocks; k++) {
+#pragma omp for schedule(dynamic)
+      for (IdType m = 0; m < num_M_blocks; m++) {
+        CSRMatrixInternal<IdType, IdType> cur_csr = block_csr_array[m * num_K_blocks + k];
+
+        int32_t cur_M = cur_csr.num_rows;
+
+        IdType M_start = m * M_block_size;
+        for (IdType i = 0; i < cur_M; i++) {
```
jermainewang, line 308: `for (IdType i = 0; i < cur_csr.num_rows; i++) {` and delete `cur_M`.

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+#pragma omp parallel
+  {
+    for (IdType k = 0; k < num_K_blocks; k++) {
+#pragma omp for schedule(dynamic)
+      for (IdType m = 0; m < num_M_blocks; m++) {
+        CSRMatrixInternal<IdType, IdType> cur_csr = block_csr_array[m * num_K_blocks + k];
+
+        int32_t cur_M = cur_csr.num_rows;
+
+        IdType M_start = m * M_block_size;
```
jermainewang, line 307: const

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+        for (IdType i = 0; i < cur_M; i++) {
```
jermainewang, line 371: similar comments here

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+    libxsmm_meltwfunction_opreduce_vecs_idx kernel) {
+
+  DType (*in_matrix1)[N] = (DType (*)[N])B;
+  DType (*in_matrix2)[N] = (DType (*)[N])E;
+  DType (*output)[N] = (DType (*)[N])C;
+#pragma omp parallel
+  {
+    for (IdType k = 0; k < num_K_blocks; k++) {
+#pragma omp for schedule(dynamic)
+      for (IdType m = 0; m < num_M_blocks; m++) {
```
jermainewang, line 302: I wonder why not
```c++
#pragma omp parallel for
    for (IdType m = 0; m < num_M_blocks; m++) {
      for (IdType k = 0; k < num_K_blocks; k++) {
```

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+    argE = arge.Ptr<IdType>();
+  }
+
+  int nthreads = omp_get_max_threads();
+  const IdType M = csr.num_rows;
+  const IdType N = bcast.out_len;
+  const IdType K = csr.num_cols;
+  const IdType* indptr = csr.indptr.Ptr<IdType>();
+  CHECK_NOTNULL(indptr);
+  int total_nnz = indptr[M];
```
jermainewang, line 466: const

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+  DType* C = out.Ptr<DType>();
+  const DType* B = ufeat.Ptr<DType>();
+  const DType* E = efeat.Ptr<DType>();
+  IdType *argB, *argE;
+  if (std::is_same<Redop, op::Max<DType>>::value || std::is_same<Redop, op::Min<DType>>::value) {
+    argB = argu.Ptr<IdType>();
+    argE = arge.Ptr<IdType>();
+  }
+
+  int nthreads = omp_get_max_threads();
```
jermainewang, line 460: const

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+  const IdType M = csr.num_rows;
+  const IdType N = bcast.out_len;
+  const IdType K = csr.num_cols;
+  const IdType* indptr = csr.indptr.Ptr<IdType>();
+  CHECK_NOTNULL(indptr);
+  int total_nnz = indptr[M];
+  if (M <= 0 || K <= 0 || N <= 0 || total_nnz <= 0) return;
+
+  float avgDegree = total_nnz * 1.0 / M;
```
jermainewang, line 469: const float avg_degree

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+  float nnz_prob = avgDegree / K;
```
jermainewang, line 470: const

jermainewang:
@sanchit-misra I've uploaded my comments. The overall code logic is quite clean (Thanks for the comments! They indeed help a lot!). I feel it is very close to be merged unless there are other performance regressions. Thanks for the great effort!

sanchit-misra:
> @sanchit-misra I've uploaded my comments. The overall code logic is quite clean (Thanks for the comments! They indeed help a lot!). I feel it is very close to be merged unless there are other performance regressions. Thanks for the great effort!

Thanks. Will work on them.

sanchit-misra:
> @sanchit-misra My bad. You are right

Great! I wanted to mention the following points wrt results.csv at the following link showing worse performance of this PR for some cases. 
https://dgl-asv-data.s3-us-west-2.amazonaws.com/2b58e634ad578b0b48a4c8144c2b01c76038d1d1_c59xlarge/results/result.csv

1. This PR has not modified the sddmm code at all. So, all the cases where sddmm code is running slower is not because of this PR.
2. For gspmm for which result.csv shows worse performance of PR #3024, the naive SpMM code is being run for both master branch and PR #3024. This is because that path is not supported by the optimized code for this PR. So, the comparison is just between naive code run for master branch and naive code running for this PR. Also, performance oscillates quite a bit from run to run and could be the reason for the observed slowdown.
3. I also did a run at my end comparing this PR with xbyak code and naive code. These experiments were performed on a single socket of Intel Xeon Platinum 8280 consisting of 28 cores (https://ark.intel.com/content/www/us/en/ark/products/192478/intel-xeon-platinum-8280-processor-38-5m-cache-2-70-ghz.html). I used 28 threads with scatter KMP_AFFINITY. Wherever applicable, xbyak code runs faster than naive code. The code of this PR that is based on libxsmm is applicable in more cases than xbyak and whenever applicable, it is faster than naive and xbyak code.

Please find attached detailed results here:
[result.xlsx](https://github.com/dmlc/dgl/files/6771531/result.xlsx)

CMakeLists.txt
```diff
@@ -108,9 +109,15 @@ if(USE_OPENMP)
 endif(USE_OPENMP)
 
 if(USE_AVX)
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DUSE_AVX")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DUSE_AVX")
-  message(STATUS "Build with AVX optimization.")
+  if(USE_LIBXSMM)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DUSE_AVX -DUSE_LIBXSMM -DDGL_CPU_LLC_SIZE=40000000")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DUSE_AVX -DUSE_LIBXSMM -DDGL_CPU_LLC_SIZE=40000000")
```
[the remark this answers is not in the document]
sanchit-misra: This is the size of the LLC of the CPU the code is run on. This macro here is just failsafe. The code automatically gets the LLC size using sysconf() in the function getLLCSize(). Only if that fails, it uses this number. The number I have used as default here is quite safe for most server class CPUs. Only if sysconf fails and this number is also bigger than the user's LLC, then they will have to set the LLC size here.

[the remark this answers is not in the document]
jermainewang: > Only if sysconf fails and this number is also bigger than the user's LLC, then they will have to set the LLC size here.

In this case, will the program crash or just run with under-optimal config?

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: That is how libxsmm API works. There is no need to set `LIBXSMM_MELTW_FLAG_OPREDUCE_VECS_INDEXED_VEC` flag as in L238-241?

src/array/cpu/spmm.h
```diff
@@ -42,52 +45,74 @@ void SpMMSumCsr(const BcastOff& bcast, const CSRMatrix& csr, NDArray ufeat,
...
-  /* Prepare an assembler kernel */
-  static std::unique_ptr<ElemWiseUpd> asm_kernel_ptr(
-    (dgl::IntelKernel<>::IsEnabled()) ? new ElemWiseUpd() : nullptr);
-  /* Distribute the kernel among OMP threads */
-  ElemWiseUpd* cpu_spec = (asm_kernel_ptr && asm_kernel_ptr->applicable())
-                            ? asm_kernel_ptr.get()
-                            : nullptr;
-  if (cpu_spec && dim > 16 && !bcast.use_bcast) {
+#ifdef USE_LIBXSMM
+  bool special_condition =
```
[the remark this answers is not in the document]
sanchit-misra: Done

src/array/cpu/spmm.h
```diff
@@ -42,52 +45,74 @@ void SpMMSumCsr(const BcastOff& bcast, const CSRMatrix& csr, NDArray ufeat,
...
+       bcast.use_bcast || (Op::use_lhs && (dim != lhs_dim)) || (Op::use_rhs && (dim != rhs_dim))
```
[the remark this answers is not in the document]
sanchit-misra: Right. Removed the redundant parts.

src/array/cpu/spmm.h
```diff
@@ -172,6 +197,33 @@ void SpMMCmpCsr(const BcastOff& bcast, const CSRMatrix& csr, NDArray ufeat,
...
+    if (has_idx)
+      CHECK_NOTNULL(edges);
+    CHECK_NOTNULL(W);
+    CHECK_NOTNULL(argW);
+  }
+#if !defined(_WIN32)
+#ifdef USE_AVX
+#ifdef USE_LIBXSMM
+
+  bool special_condition =
```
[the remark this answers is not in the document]
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+                                                             num_K_blocks * sizeof(IdType)));
+
+#pragma omp parallel
+    {
+      IdType *my_cur_col_id = reinterpret_cast<IdType *>(aligned_alloc(64, 2 * M_block_size *
+                                                                          sizeof(IdType)));
+
+#pragma omp for
+      for (IdType m = 0; m < num_M_blocks; m++) {
+        IdType M_start = m * M_block_size;
```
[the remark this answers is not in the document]
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+        IdType M_end = (m + 1) * M_block_size;
```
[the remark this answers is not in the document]
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+        if (M_end > M) M_end = M;
+        IdType nnz = indptr[M_end] - indptr[M_start];
```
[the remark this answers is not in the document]
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+          indices_block_buf = reinterpret_cast<IdType *>(aligned_alloc(64, nnz * sizeof(IdType)));
+        if (use_rhs)
+          edges_block_buf = reinterpret_cast<IdType *>(aligned_alloc(64, nnz * sizeof(IdType)));
+
+        for (IdType i = M_start; i < M_end; i++) {
+          my_cur_col_id[(i - M_start) * 2] = indptr[i];
+          my_cur_col_id[(i - M_start) * 2 + 1] = indptr[i + 1];
+        }
+        for (IdType k = 0; k < num_K_blocks; k++) {
+          IdType K_start = k * K_block_size;
```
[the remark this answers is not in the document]
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+          IdType K_end = (k + 1) * K_block_size;
```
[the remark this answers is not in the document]
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Yes, this is to support BF16. Libxsmm supports BF16 and FP32 so far. It does not support "double" yet. So, I have excluded that path in no_libxsmm condition here:
 https://github.com/dmlc/dgl/pull/3024/files#diff-7f69c0c94acb191f7e1fa6482228a6cf7cf4c71dd69358e142e4d72d5e30ed70R139
We are working on adding support for double in libxsmm but it will not be ready for another two weeks.

sanchit-misra: I have removed support for BF16 for now.

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Done

CMakeLists.txt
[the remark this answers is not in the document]
sanchit-misra: It won't crash. Will just run with under-optimal config.

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+template <typename IdType, typename DType>
+struct CSRMatrixInternal {
+  IdType num_rows;
+  IdType num_cols;
+  IdType *indptr;
+  IdType *indices;
+  DType *data;
+};
+
+int32_t getLLCSize() {
```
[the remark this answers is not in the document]
sanchit-misra: Done

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+
+#pragma omp for
+      for (IdType m = 0; m < num_M_blocks; m++) {
+        IdType M_start = m * M_block_size;
+        IdType M_end = (m + 1) * M_block_size;
+        if (M_end > M) M_end = M;
+        IdType nnz = indptr[M_end] - indptr[M_start];
+
+        IdType cur_indices_id = 0;
+        IdType *indices_block_buf, *edges_block_buf;
```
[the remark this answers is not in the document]
sanchit-misra: Yes, that can be done. The current version is an artifact of an older version of the code. I am changing it according to what you suggested and also see some speedup :)

src/array/cpu/spmm_blocking_libxsmm.h
sanchit-misra: Please see this picture for tiling for copylhs-reduce kind of SpMM. Here, A is the adjacency matrix, B is the input feature containing features of all the neighbors and C is the output feature array. This is sort of equivalent of doing AxB = C. If we use the k-loop as the outer loop, that ensures that all the threads are working on the same block of B and that block stays in cache. On the other hand, if we use m-loop as the outer loop, then different threads might be working on different blocks of B. That may make it harder to keep blocks of B in cache. While it may work for some workloads, keeping k-loop as outer loop provides better guarantees.

image

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+int32_t getLLCSize() {
+  int32_t cache_size = sysconf(_SC_LEVEL3_CACHE_SIZE);
+  if (cache_size < 0) cache_size = DGL_CPU_LLC_SIZE;
+  return cache_size;
+}
+
+/*!
+ * \brief Tile the CSR matrix to roughly make sure that the column tiles and
+ *        corresponding neighbor features fit into LLC and the row tiles
+ *        are assigned to OMP threads.
```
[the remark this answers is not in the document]
sanchit-misra: Provided an example to your loop reordering question below.

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -0,0 +1,598 @@
...
+              }
+              CHECK_LT(cur_indices_id + cur_nnz, nnz);
+              if (use_lhs)
+                cur_csr_indices[cur_nnz] = src;
+              if (use_rhs)
+                cur_csr_edges[cur_nnz] = edge;
+              cur_nnz++;
+            }
+            my_cur_col_id[(i - M_start) * 2] = eid;
+          }
```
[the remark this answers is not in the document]
sanchit-misra: Provided the response to your loop reordering question below.

sanchit-misra:
Hi @jermainewang, I believe I have made all the changes suggested by you and answered all your questions. Please let me know if you need anything else.

src/array/cpu/spmm_blocking_libxsmm.h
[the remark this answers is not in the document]
sanchit-misra: I have modified it according to your suggestion and committed the code.

VoVAllen:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/zhoujinj/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/zhoujinj/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">



  | test_name | params | unit | number_3024 | machine | DGL_CPU_INTEL_KERNEL_ENABLED=0 | number_DGL_CPU_INTEL_KERNEL_ENABLED=1
-- | -- | -- | -- | -- | -- | -- | --
27 | kernel.bench_gspmm_copy_u.track_flops | 'ogbn-arxiv', 4, 'sum' | GFLOPS | 6.05 | c5.9xlarge-cpu | 3.96 | 5.23
28 | kernel.bench_gspmm_copy_u.track_flops | 'ogbn-arxiv', 4, 'max' | GFLOPS | 2.85 | c5.9xlarge-cpu | 1.96 | 2.28
29 | kernel.bench_gspmm_copy_u.track_flops | 'ogbn-arxiv', 32, 'sum' | GFLOPS | 9.03 | c5.9xlarge-cpu | 4.4 | 7.66
30 | kernel.bench_gspmm_copy_u.track_flops | 'ogbn-arxiv', 32, 'max' | GFLOPS | 2.58 | c5.9xlarge-cpu | 1.81 | 1.93
31 | kernel.bench_gspmm_copy_u.track_flops | 'ogbn-arxiv', 256, 'sum' | GFLOPS | 5.98 | c5.9xlarge-cpu | 4.29 | 4.71
32 | kernel.bench_gspmm_copy_u.track_flops | 'ogbn-arxiv', 256, 'max' | GFLOPS | 1.93 | c5.9xlarge-cpu | 1.6 | 1.66
39 | kernel.bench_gspmm_copy_u.track_flops | 'ogbn-proteins', 4, 'sum' | GFLOPS | 29.24 | c5.9xlarge-cpu | 8.92 | 9.22
40 | kernel.bench_gspmm_copy_u.track_flops | 'ogbn-proteins', 4, 'max' | GFLOPS | 23.54 | c5.9xlarge-cpu | 5.83 | 5.91
41 | kernel.bench_gspmm_copy_u.track_flops | 'ogbn-proteins', 32, 'sum' | GFLOPS | 43.61 | c5.9xlarge-cpu | 9.32 | 40.1
42 | kernel.bench_gspmm_copy_u.track_flops | 'ogbn-proteins', 32, 'max' | GFLOPS | 35.58 | c5.9xlarge-cpu | 7.67 | 7.85
43 | kernel.bench_gspmm_copy_u.track_flops | 'ogbn-proteins', 256, 'sum' | GFLOPS | 58.5 | c5.9xlarge-cpu | 9.19 | 15.96
44 | kernel.bench_gspmm_copy_u.track_flops | 'ogbn-proteins', 256, 'max' | GFLOPS | 47 | c5.9xlarge-cpu | 6.98 | 7.1
45 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-arxiv', 4, 0 | GFLOPS | 4.83 | c5.9xlarge-cpu | 3.54 | 4.69
46 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-arxiv', 4, 1 | GFLOPS | 3.66 | c5.9xlarge-cpu | 3.89 | 5.12
47 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-arxiv', 4, 4 | GFLOPS | 4.84 | c5.9xlarge-cpu | 3.46 | 4.73
48 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-arxiv', 32, 0 | GFLOPS | 9.94 | c5.9xlarge-cpu | 4.61 | 9.27
49 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-arxiv', 32, 1 | GFLOPS | 5.82 | c5.9xlarge-cpu | 6.7 | 6.74
50 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-arxiv', 32, 4 | GFLOPS | 4.86 | c5.9xlarge-cpu | 5.71 | 5.62
51 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-arxiv', 256, 0 | GFLOPS | 8.25 | c5.9xlarge-cpu | 6.47 | 6.12
52 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-arxiv', 256, 1 | GFLOPS | 6.01 | c5.9xlarge-cpu | 6.67 | 7.04
53 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-arxiv', 256, 4 | GFLOPS | 5.88 | c5.9xlarge-cpu | 6.67 | 6.87
63 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-proteins', 4, 0 | GFLOPS | 7.87 | c5.9xlarge-cpu | 3.28 | 3.41
64 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-proteins', 4, 1 | GFLOPS | 3.01 | c5.9xlarge-cpu | 3.03 | 3.13
65 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-proteins', 4, 4 | GFLOPS | 7.87 | c5.9xlarge-cpu | 3.27 | 3.38
66 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-proteins', 32, 0 | GFLOPS | 21.12 | c5.9xlarge-cpu | 4.95 | 12.71
67 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-proteins', 32, 1 | GFLOPS | 6.29 | c5.9xlarge-cpu | 6.76 | 7.03
68 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-proteins', 32, 4 | GFLOPS | 5.55 | c5.9xlarge-cpu | 5.95 | 6.21
69 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-proteins', 256, 0 | GFLOPS |   | c5.9xlarge-cpu |   |  
70 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-proteins', 256, 1 | GFLOPS | 8.8 | c5.9xlarge-cpu | 10.1 | 10.4
71 | kernel.bench_gspmm_u_mul_e_sum.track_flops | 'ogbn-proteins', 256, 4 | GFLOPS | 8.5 | c5.9xlarge-cpu | 9.77 | 10.04



</body>

</html>

VoVAllen:
[intel_kernel.xlsx](https://github.com/dmlc/dgl/files/6781023/intel_kernel.xlsx)
Here's the result comparing to the previous kernel with DGL_CPU_INTEL_KERNEL_ENABLED=1

sanchit-misra:
> [intel_kernel.xlsx](https://github.com/dmlc/dgl/files/6781023/intel_kernel.xlsx)
> Here's the result comparing to the previous kernel with DGL_CPU_INTEL_KERNEL_ENABLED=1

Thanks VoVAllen. As can be seen in the results, this PR gets speedups up to 6.7x compared to xbyak run by using DGL_CPU_INTEL_KERNEL_ENABLED=1. There are a few cases in your excel sheet for which this PR gets lower performance. I went through all of those cases and verified that in all those cases, naive SpMM code is running whether we use this PR or master branch. That is because neither this PR nor xbyak provide optimized implementations for when use_bcast=1. So, the same naive code is performing differently between the two runs (probably because of run to run variation) and the slow down has nothing to do with this PR.

ksadowski13:
@VoVAllen Thanks for the results. Do kernel benchmarks cover only full-graph scenario?

jermainewang (approved):
LGTM. I think the PR is ready to go.

sanchit-misra:
Great to know that this has been merged. Thanks everyone for all the help!

PR #3025, opened 2021-06-15
[Model] RGCN with new heterograph API

isratnisa:
## Description
The new heterogeneous graph APIs, .e.g, `update_all`, `apply_edges` invoke one python call to handle all the relation types of a heterogeneous graph and performs one-level message aggregation. Previously, either separate python kernels were invoked for each relation type, or converts the graph to a homogeneous one which have one relation type and calls one python kernel. This PR implements RGCN-hetero with the new APIs and achieves `6x` speedup on `aifb` and `mutag` datasets compared to the old RGCN-hetero code. The corresponding NN modules should be changed to allow the new API. For debugging purpose we modified only the example codes. 

## Checklist
- [ ] Code is well-documented


examples/pytorch/rgcn-hetero/model.py
```diff
@@ -120,6 +121,125 @@ def _apply(ntype, h):
...
+        self.out_feat = out_feat
+        self.rel_names = rel_names
+        self.num_bases = num_bases
+        self.bias = bias
+        self.activation = activation
+        self.self_loop = self_loop
+
+        self.use_weight = weight
+        self.use_basis = num_bases < len(self.rel_names) and weight
+        if self.use_weight:
```
zheng-da, line 170: i'm just curious, why do we have this option? if not using weight, is it a graphsage layer?

zheng-da: it's not even graphsage. there will be no parameters in the model.

examples/pytorch/rgcn-hetero/model.py
```diff
@@ -120,6 +121,125 @@ def _apply(ntype, h):
...
+            weight = self.basis() if self.use_basis else self.weight
+            wdict = {self.rel_names[i] : {'weight' : w.squeeze(0)}
+                     for i, w in enumerate(th.split(weight, 1, dim=0))}
+        else:
+            wdict = {}
+
+        if g.is_block:
+            inputs_src = inputs
+            inputs_dst = {k: v[:g.number_of_dst_nodes(k)] for k, v in inputs.items()}
```
zheng-da, line 215: is this part of code tested? i suppose you only have full-graph training?

examples/pytorch/rgcn-hetero/model.py
```diff
@@ -253,3 +373,98 @@ def inference(self, g, batch_size, device, num_workers, x=None):
...
+
+        ***NOTE***
+        For node classification, the model is trained to predict on only one node type's
+        label.  Therefore, only that type's final representation is meaningful.
+        """
+
+        if x is None:
+            x = self.embed_layer()
+
+        for l, layer in enumerate(self.layers):
```
zheng-da, line 443: it seems you train the model with full graph but perform inference with mini-batch. we normally don't do it.

zheng-da:
will you implement a mini-batch version as well?

examples/pytorch/rgcn-hetero/model.py
isratnisa: Removed for now. Will add back to support minibatch version.

examples/pytorch/rgcn-hetero/model.py
isratnisa: I see. Apparently, the original RGCN-hetero example with `heteroConv` also using the mini-batch based inferencing. Both cases should be changed?

zheng-da: this implementation is very weird.

zheng-da (approved):

PR #3034, opened 2021-06-17
[Bug fix] Use shared memory for grad sync when NCCL is not avaliable as PyTorch distributed backend.

classicsong:
## Description
Use shared memory for grad sync when NCCL is not available as PyTorch distributed backend.

Fix small bugs and update unit tests.

## Checklist
- [x] Code is well-documented


nv-dlasalle (approved):

python/dgl/optim/pytorch/sparse_optim.py
```diff
@@ -284,9 +287,11 @@ def _shared_step(self):
                                 # The overall buffer cost will be smaller than three times
                                 # the maximum memory requirement for sharing gradients.
                                 buffer_size = 128 if idx_i.shape[0] < 128 else idx_i.shape[0] * 2
-                                idx_shmem = create_shared_mem_array(idx_shmem_name, \
+                                idx_shmem = create_shared_mem_array(
+                                    '{}_{}'.format(idx_shmem_name, buffer_size), \
                                     (buffer_size,), idx_dtype)
-                                grad_shmem = create_shared_mem_array(grad_shmem_name, \
+                                grad_shmem = create_shared_mem_array(
+                                    '{}_{}'.format(grad_shmem_name, buffer_size), \
```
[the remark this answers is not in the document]
classicsong: When we want to extend the shared memory size, if we use the same name, the old one will be released and will cause some bugs (if the shared memory is freed by the host first, any call to free the same memory in other mirror process will crash the program..

src/runtime/cuda/nccl_api.cu
```diff
@@ -178,7 +178,7 @@ std::pair<IdArray, NDArray> SparsePush(
   // permute the indices and values
   if (num_in > 0) {
     const dim3 block(256);
-    const dim3 grid((num_in+block.x-1)/block.x);
+    const dim3 grid((num_in+block.x-1)/block.x == 0 ? 1 : (num_in+block.x-1)/block.x);
```
nv-dlasalle, line 181: Now that the `if (num_in > 0) {` was added above, this change is no longer needed.

nv-dlasalle (approved):
Looks good to me--just the now redundant check for an empty kernel grid, after the other PR was merged.

src/runtime/cuda/nccl_api.cu
classicsong: I c, will remove it.

PR #3040, opened 2021-06-19
[Example] Fix error in gcn train due to missing argument

### LINKED ISSUE #3039: Error due to missing argument to specify dataset name in gcn example
## 🐛 Bug

In `gcn/train.py` it raises error due to `Unknown dataset`.

```ValueError: Unknown dataset: None```

## To Reproduce

1. Run `gcn/train.py` without arguments

## Environment

 - DGL Version (e.g., 1.0): 0.7
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): 1.9.0
 - OS (e.g., Linux): Windows 10
 - How you installed DGL (`conda`, `pip`, source): source
 - Build command you used (if compiling from source): same as the installation guide
 - Python version: 3.9.5

## Additional context

It is because this example didn't pass `dataset` to `parser.add_argument` but, in `main()`, `dataset` is required to specify which dataset is used.

I didn't run `gcn/train.py`  in MXNet, but I guess it raises the same error when using MXNet.

europeanplaice:
## Description
Fixes #3039

I finished coding on this PR.

## Checklist

## Changes
* Add `parser.add_argument` to receive a dataset name. (default `cora`)
* When name `dataset` was used as an argument, error `argparse.ArgumentError: argument --dataset: conflicting option string: --dataset` was raised. Therefore I changed `dataset` to `dataset_name`.
* Remove unused imports

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,6 +331,10 @@ We are currently in Beta stage.  More features and improvements are coming.
 
 1. [**Improving Graph Representation Learning by Contrastive Regularization**](https://arxiv.org/pdf/2101.11525.pdf), *Kaili Ma, Haochen Yang, Han Yang, Tatiana Jin, Pengfei Chen, Yongqiang Chen, Barakeel Fanseu Kamhoua, James Cheng*
 
+1. [**VIKING: Adversarial Attack on Network Embeddings via Supervised Network Poisoning**](https://arxiv.org/pdf/2102.07164.pdf), PAKDD'21, *Viresh Gupta, Tanmoy Chakraborty*
+
+1. [**Knowledge Graph Embedding using Graph Convolutional Networks with Relation-Aware Attention**](https://arxiv.org/pdf/2102.07200.pdf), *Nasrullah Sheikh, Xiao Qin, Berthold Reinwald, Christoph Miksovic, Thomas Gschwind, Paolo Scotton*
+
 </details>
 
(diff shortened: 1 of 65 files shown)


jermainewang:
Hi @europeanplaice , thanks for the fix! The issue you raised makes a lot of sense. We shouldn't let the script fail with an awkward (and confusing) `None` dataset. Here are two of my suggestions:

* We are in the process of deprecating the usage of `register_data_args(parser)`. Since you are modifying these examples anyway, could you help delete the corresponding lines? This is also the source of the duplicate argument error you've encountered, so you should be able to use the `--dataset` argument once again after this change.
* In the argument description, it is better to also give the allowed choices. For example `"Dataset name ('cora', 'citeseer', 'pubmed')."`.

europeanplaice:
Thank you for the review!

I had missed `register_data_args` and now have found how this function works.
I removed this function and reverted the argument `dataset_name` to `dataset`.
I also modified the argument description as you suggested.

I have chosen `cora` as the default dataset, however, I'm not sure if this is preferable.
Otherwise, should I remove `default` and set `required`?

jermainewang (approved):

jermainewang:
> Otherwise, should I remove `default` and set `required`?

I think either is ok. Thanks again for the fix!

PR #3046, opened 2021-06-21
[Doc] Update NodeDataLoader and EdgeDataLoader for GPU-based neighbor sampling

BarclayII:
With a series of PRs we now support uniform neighbor sampling on GPUs.  This is an update of the following documentations:

* `NodeDataLoader`
* `EdgeDataLoader`
* `sample_neighbors`
* `to_block`

Also comes with a compatibility fix where `num_workers > 0` and `device = 'cuda'` throws an error if a CPU graph is given as input: now it reverts to CPU-based neighbor sampling and `to_block` instead.


python/dgl/dataloading/pytorch/dataloader.py
```diff
@@ -0,0 +1,912 @@
...
+    
+    * Feed in the graph on GPU.
+    * Set :attr:`device` to GPU.
+    * Set :attr:`num_workers` to 0.
+
+    If your graph is too big, we recommend you
+
+    * Feed in the graph on CPU.
+    * Set :attr:`device` to GPU.
+    * (You can set :attr:`num_workers` to anything you like.)
```
nv-dlasalle, line 464: Should it be noted here that setting `num_workers > 0` will force `to_block()` to run on the CPU instead of the GPU? Maybe something like:
```
* If :attr:`num_workers` is set to 0 in this case, the sampling will happen on the CPU, and then the subgraphs will be constructed directly on the GPU. This is the recommend setting in multi-GPU configurations.
* Otherwise, if :attr:`num_workers` is greater than 0, both the sampling and subgraph construction will take place on the CPU. This is the recommended setting when using a single-GPU and the whole graph does not fit in GPU memory.
```

python/dgl/dataloading/pytorch/dataloader.py
```diff
@@ -0,0 +1,912 @@
...
+
+    If you still want to enjoy GPU-based neighbor sampling, you will need to pop the features
+    outside the graph.
```
jermainewang, line 467: Why is that?

python/dgl/dataloading/pytorch/dataloader.py
jermainewang: Let's not use you or we in the sentences. How about:
```
**Tips for selecting the proper device**

* If the input graph :attr:`g` is on GPU, the output device :attr:`device` must be the same GPU
  and :attr:`num_workers` must be zero. In this case, the sampling and subgraph construction
  will take place on the GPU. This is the recommended setting when using a single-GPU and
  the whole graph fits in GPU memory.
* If the input graph :attr:`g` is on CPU while the output device :attr:`device` is GPU, then depending
  on the value of :attr:`num_workers`:
  - If :attr:`num_workers` is set to 0, the sampling will happen on the CPU, and then the subgraphs
    will be constructed directly on the GPU. This is the recommend setting in multi-GPU configurations.
  - Otherwise, if :attr:`num_workers` is greater than 0, both the sampling and subgraph construction
    will take place on the CPU. This is the recommended setting when using a single-GPU and the whole
    graph does not fit in GPU memory.
```

python/dgl/dataloading/pytorch/dataloader.py
BarclayII: Because you cannot put features onto GPU?

python/dgl/dataloading/pytorch/dataloader.py
BarclayII: Yeah this is much better.

python/dgl/dataloading/pytorch/dataloader.py
```diff
@@ -0,0 +1,915 @@
...
+
+    Parameters
+    ----------
+    collate_fn : Function, default is None
+        The customized collate function. Will use the default collate
+        function if not given.
+    use_ddp : boolean, optional
+        If True, tells the DataLoader to split the training set for each
+        participating process appropriately using
+        :class:`torch.utils.data.distributed.DistributedSampler`.
```
nv-dlasalle, line 832: We should note that if `use_ddp` is true, `set_epoch()` must be called at every epoch or the dataset won't be shuffled.

nv-dlasalle (approved):

PR #3051, opened 2021-06-22
EdgeWeightNorm Fix doc

europeanplaice:
## Description
in https://docs.dgl.ai/en/0.6.x/api/python/nn.pytorch.html#edgeweightnorm
we can't see two math formulas that would be displayed to us.

## Changes
`math:` to `math::`


VoVAllen (approved):
Thanks!

PR #3086, opened 2021-07-02
[Feature][Performance][GPU] Introducing UnifiedTensor for efficient zero-copy host memory access from GPU

davidmin7:
## Description
Unlike some of the conventional NN training, GNN training often involves fetching input features stored in non-contiguous and fine-grained manners. Such characteristics may not matter when the whole input feature can be fit into the GPU memory, but it can be a quite delicate issue with much larger input datasets.

As of now, to resolve this issue, the practice which is the most commonly done is using CPU to slice the original node feature tensor and then sending the resulting tensor to the GPU memory. However, the volume of data which needs to be moved during this "reshaping" process can be non-negligible because the CPU need to first read from the original tensor (1st memory access), write to the temporary buffer (2nd memory access), and finally the content of the temporary buffer is DMAed to GPU (3rd memory access). This means, to send X-byte of data to GPU, we end up accessing memory by 3X-byte.

To overcome such issue, this PR introduces the zero-copy access capability of NVIDIA GPUs. Modern NVIDIA GPUs are capable of accessing the CPU memory directly. If the GPUs can directly access the CPU memory, what we only need to send is the list of node feature indices which is relatively much smaller than the node feature itself. Once the indices are sent to the GPUs, GPUs can freely access the node features without bothering CPU.

In case of training GraphSAGE with Reddit dataset, this method can reduce the training epoch time from 7.4s to 4.2s when the node features are always stored in the CPU memory.

This PR is mainly composed of two parts: First, adding the capability to pin and map CPU memory into the GPU memory. Second, adding the cross-device operators/functions. From the user perspective, both parts are simply packed into the new `UnifiedTensor` class. 

The opt-in to this feature is completely optional. Unless the users actively declare the `UnifiedTensor` class in their codes, there should not be any side-effect from this PR.

## Checklist
- [ ] Code is well-documented

## Changes
- [ ] Feature1: Add the in-place memory pinning/unpinning mechanisms to the DGL ndarray using cudaHostRegister()/cudaHostUnregister()
- [ ] Feature2: Introducing "UVM" kernel category which can operate in cross-device manner. For example, GPU directly reading node features stored in CPU memory.


src/array/uvm_array_op.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <dgl/array.h>
+#include <utility>
+
+namespace dgl {
+namespace aten {
+namespace impl {
+
+template <typename DType, typename IdType>
+NDArray IndexSelectCPUFromGPU(NDArray array, IdArray index);
```
VoVAllen, line 17: Is this `IndexSelectCPUToGPU` or `IndexSelectGPUFromCPU`?

davidmin7: So the rule I've made here is that if I say FuncXfromY, then X is the memory location and Y is the operator (or processor). I think you mean from the perspective of the actual data movement direction. If you think `IndexSelectCPUToGPU` is easier to understand, we can definitely modify to it.

examples/pytorch/ogb_lsc/MAG240M/experimental/train_unified_tensor.py
```diff
@@ -0,0 +1,233 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+import ogb
```
classicsong, line 4: ogb is not used.

examples/pytorch/ogb_lsc/MAG240M/experimental/train_unified_tensor.py
```diff
@@ -0,0 +1,233 @@
...
+import time
+import tqdm
+import dgl.function as fn
+import numpy as np
+import dgl.nn as dglnn
+import torch.nn as nn
+import torch.nn.functional as F
+import argparse
+
+class RGAT(nn.Module):
```
classicsong, line 18: Can we just use from .train_multi_gpus import RGAT?

examples/pytorch/ogb_lsc/MAG240M/experimental/train_unified_tensor.py
```diff
@@ -0,0 +1,233 @@
...
+        batch_size=1024,
+        shuffle=True,
+        drop_last=False,
+        collate_fn=train_collator.collate,
+        num_workers=4
+    )
+    valid_dataloader = torch.utils.data.DataLoader(
+        valid_collator.dataset,
+        batch_size=1024,
+        shuffle=True,
```
classicsong, line 104:
```suggestion
        shuffle=False,
```
We do not need to shuffle the data for validation.

examples/pytorch/ogb_lsc/MAG240M/experimental/train_unified_tensor.py
```diff
@@ -0,0 +1,233 @@
...
+    sched = torch.optim.lr_scheduler.StepLR(opt, step_size=25, gamma=0.25)
+
+    best_acc = 0
+
+    for _ in range(args.epochs):
+        model.train()
+        with tqdm.tqdm(train_dataloader) as tq:
+            for i, (input_nodes, output_nodes, mfgs) in enumerate(tq):
+                mfgs = [g.to('cuda') for g in mfgs]
+                x = feats.gather_row(input_nodes.cuda()).float()
```
classicsong, line 122: Can we use feats[input_nodes.cuda()] ?
Furthermore, the dtype of feats can be inferred when we initialize the UnifiedTensor.

examples/pytorch/ogb_lsc/MAG240M/experimental/train_unified_tensor.py
```diff
@@ -0,0 +1,233 @@
...
+                opt.step()
+                acc = (y_hat.argmax(1) == y).float().mean()
+                tq.set_postfix({'loss': '%.4f' % loss.item(), 'acc': '%.4f' % acc.item()}, refresh=False)
+
+        model.eval()
+        correct = total = 0
+        for i, (input_nodes, output_nodes, mfgs) in enumerate(tqdm.tqdm(valid_dataloader)):
+            with torch.no_grad():
+                mfgs = [g.to('cuda') for g in mfgs]
+                x = feats.gather_row(input_nodes.cuda()).float()
```
classicsong, line 137: Same here.

python/dgl/contrib/unified_tensor.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __setitem__(self, key, val):
+        self._input[key] = val
+
+    def __del__(self):
+        self._array.unpin_memory_(utils.to_dgl_context(self._device))
+        self._array = None
+        self._input = None
+
+    def gather_row(self, index):
```
classicsong, line 46: Can we merge gather_row with \_\_getitem\_\_ ?

src/array/cuda/uvm/array_index_select_uvm.cu
```diff
@@ -0,0 +1,65 @@
...
+}
+
+template NDArray IndexSelectCPUFromGPU<int8_t, int32_t>(NDArray, IdArray);
+template NDArray IndexSelectCPUFromGPU<int8_t, int64_t>(NDArray, IdArray);
+template NDArray IndexSelectCPUFromGPU<int16_t, int32_t>(NDArray, IdArray);
+template NDArray IndexSelectCPUFromGPU<int16_t, int64_t>(NDArray, IdArray);
+template NDArray IndexSelectCPUFromGPU<int32_t, int32_t>(NDArray, IdArray);
+template NDArray IndexSelectCPUFromGPU<int32_t, int64_t>(NDArray, IdArray);
+template NDArray IndexSelectCPUFromGPU<int64_t, int32_t>(NDArray, IdArray);
+template NDArray IndexSelectCPUFromGPU<int64_t, int64_t>(NDArray, IdArray);
```
classicsong, line 61: Can we also support  `template NDArray IndexSelectCPUFromGPU<float32_t, int64_t>(NDArray, IdArray);` ?

python/dgl/contrib/unified_tensor.py
```diff
@@ -0,0 +1,75 @@
...
+    device : device
+        Device to create the mapping of the unified tensor.
+    '''
+
+    def __init__(self, input, device):
+        if F.device_type(device) != 'cuda':
+            raise ValueError("Target device must be a cuda device")
+
+        self._input = input
+        self._array = F.zerocopy_to_dgl_ndarray(self._input)
```
classicsong, line 24: Do we need to make sure the `input` is pinned in CPU memory? If so we need to highlight it in the doc string.

classicsong:
In general, can you add some unittest for Unified tensor? 
You can add them here:  https://github.com/dmlc/dgl/tree/master/tests/pytorch

examples/pytorch/ogb_lsc/MAG240M/experimental/train_unified_tensor.py
davidmin7: In fact this training example is an exact copy of [https://github.com/dmlc/dgl/blob/master/examples/pytorch/ogb_lsc/MAG240M/train.py](https://github.com/dmlc/dgl/blob/master/examples/pytorch/ogb_lsc/MAG240M/train.py). Based on your comments, it seems like several improvements can be made in the baseline DGL code. However, in terms of consistency, it is kind of weird to make the updates only for this pull request. So I simply created a new example based on commonly used GraphSAGE code ([https://github.com/dmlc/dgl/blob/master/examples/pytorch/graphsage/train_sampling.py](https://github.com/dmlc/dgl/blob/master/examples/pytorch/graphsage/train_sampling.py)). Someone else can update the baseline MAG240M DGL code later.

examples/pytorch/ogb_lsc/MAG240M/experimental/train_unified_tensor.py
davidmin7: Now the example code is changed to the GraphSAGE.

examples/pytorch/ogb_lsc/MAG240M/experimental/train_unified_tensor.py
davidmin7: This should be also addressed by migrating to GraphSAGE example.

examples/pytorch/ogb_lsc/MAG240M/experimental/train_unified_tensor.py
davidmin7: No problem. Both requests are reflected in the new GraphSAGE example.

examples/pytorch/ogb_lsc/MAG240M/experimental/train_unified_tensor.py
davidmin7: Reflected in the newer example.

python/dgl/contrib/unified_tensor.py
davidmin7: No problem. It should be merged now.

src/array/cuda/uvm/array_index_select_uvm.cu
davidmin7: Applied!

python/dgl/contrib/unified_tensor.py
davidmin7: No problem. Added more explanation in the doc string. __init__ process includes pinning the memory so the users do not need to worry about it.

davidmin7:
> In general, can you add some unittest for Unified tensor?
> You can add them here: https://github.com/dmlc/dgl/tree/master/tests/pytorch

I added a `test_unified_tensor.py` test in the directory you suggested.

examples/pytorch/ogb_lsc/MAG240M/experimental/train_unified_tensor.py
classicsong: Yeah. In this PR we only focus on this example.
We can update the baseline coder later in another PR.

src/array/uvm_array_op.h
classicsong: We only need to document it clearly. 
I think IndexSelectCPUFromGPU means selecting data in CPU according to the Index from GPU and return the GPU NDArray.
Can you Add a doc string and also check the context of array and index using assertions?

tests/pytorch/test_unified_tensor.py
```diff
@@ -0,0 +1,25 @@
...
+    rand_test_size = 8192
+
+    input = th.rand((test_row_size, test_col_size))
+    input_unified = dgl.contrib.UnifiedTensor(input, device=th.device('cuda'))
+
+    seq_idx = th.arange(0, test_row_size)
+    assert th.all(th.eq(input[seq_idx], input_unified[seq_idx]))
+
+    rand_idx = th.randint(0, test_row_size, (rand_test_size,))
+    assert th.all(th.eq(input[rand_idx], input_unified[rand_idx]))
```
classicsong, line 22: We also add tests for idx in GPU.
```suggestion
    rand_idx = th.randint(0, test_row_size, (rand_test_size,))
    assert th.all(th.eq(input[rand_idx], input_unified[rand_idx]))
    
    rand_idx = rand_idx.to(th.device('cuda'))
    assert th.all(th.eq(input[rand_idx].to(th.device('cuda')), input_unified[rand_idx]))
```

classicsong (approved):
Overall LGTM
Let some comments.

davidmin7:
> 
> 
> Overall LGTM
> Let some comments.

I added a comment and also the asserts.

maqy1995:
Hi @davidmin7 , thanks for your great work!
I built the DGL from the repo `https://github.com/davidmin7/dgl.git` and run the graphsage examples. It is found that unifiedTensor has indeed achieved a good acceleration effect:

> (work) maqy@ubuntu:~/work/gnn/dglUnifiedTensor/davidfork/dgl/examples/pytorch/graphsage$ python train_sampling.py --num-epochs 10 --data-cpu
Epoch 00000 | Step 00000 | Loss 6.1428 | Train Acc 0.0300 | Speed (samples/sec) nan | GPU 265.5 
MB
Epoch 00000 | Step 00020 | Loss 2.8067 | Train Acc 0.3210 | Speed (samples/sec) 15635.5011 | GPU 512.2 MB
...
Eval Acc 0.9465
Test Acc: 0.9468
...
Epoch 00009 | Step 00140 | Loss 0.3134 | Train Acc 0.9180 | Speed (samples/sec) 15294.2769 | GPU 514.7 MB
Epoch Time(s): 10.4865
Avg epoch time: 10.435481643676757
----------------------------------------------- split line ---------------------------------------------------
(work) maqy@ubuntu:~/work/gnn/dglUnifiedTensor/davidfork/dgl/examples/pytorch/graphsage$ python train_sampling_unified_tensor.py --num-epochs 10 --data-cpu
Epoch 00000 | Step 00000 | Loss 6.0242 | Train Acc 0.0280 | Speed (samples/sec) nan | GPU 11.1 MB
Epoch 00000 | Step 00020 | Loss 3.0486 | Train Acc 0.2370 | Speed (samples/sec) 29305.7057 | GPU 11.6 MB
...
Eval Acc 0.9467
Test Acc: 0.9458
...
Epoch 00010 | Step 00140 | Loss 0.3327 | Train Acc 0.9270 | Speed (samples/sec) 29333.4603 | GPU 351.5 MB
Epoch Time(s): 5.5860
Avg epoch time: 5.597336053848267

But here I'm a little confused as to why the GPU memory usage is less when using UnifiedTensor than the original way(11MB vs 512MB before inference/ 351MB vs 514MB after the first inference)?

Looking forward to your reply.

davidmin7:
Hi, @maqy1995

Thank you for your interest in using `UnifiedTensor`. The example you used, has several aggressive optimizations I added previously so it may not be the best application to get some stable memory allocation status.

Fortunately, @yaox12 rewrote the example here: https://github.com/dmlc/dgl/blob/master/examples/pytorch/graphsage/train_sampling.py. You can use the `UnifiedTensor` by using `uva` for the `--graph-device` (for the graph structure like CSR) and `--data-device` (for the node feature tensor) arguments. Hope this example gives you a better view of memory usage. This commit should be merged in the nightly version of DGL, I think.

yaox12:
@maqy1995 The GPU memory usage reported in the training script is not accurate. It only counts in the GPU memory allocated by PyTorch. Use `nvidia-smi` command in the shell to get an accurate number.

maqy1995:
Thanks for your quick reply @davidmin7 @yaox12 .

> Use nvidia-smi command in the shell to get an accurate number.

I also used the `nvidia-smi` command to see the GPU memory usage yesterday, it seems that it will still be a little less than the original GPU memory usage when using UnifiedTensor (about a few hundred MBs).

The possible reason:
> The example you used, has several aggressive optimizations I added previously so it may not be the best application to get some stable memory allocation status.

Anyway, I'll try to use the latest example. Thanks again for your replies.

davidmin7:
Hi @maqy1995, `UnifiedTensor` will prevent the node feature tensor being loaded to GPU memory. If you can check the size of your node feature tensor, maybe the few hundred MBs of gap can be explained.

davidmin7:
Hi @maqy1995, maybe my guess is that PyTorch is reserving some extra space in the non-UnifiedTensor method. Unfortunately, I don't know which would be the trigger exactly.

PR #3087, opened 2021-07-02
[Model] Official implementation for HiLANDER model. 

hetong007:
## Description
This PR adds the official implementation for the paper *Learning Hierarchical Graph Neural Networks for Image Clustering*.

Will update this PR or in a separate PR to add the link to arxiv.

## Checklist
- [X] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).


examples/pytorch/hilander/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+```bash
+bash scripts/test_deepglint_hannah.sh
+
+bash scripts/test_deepglint_imdb.sh
+
+bash scripts/test_inat_train_on_resampled_1_in_6_per_class.sh
+```
+
+**Clustering with Seen Test Data Distribution Performance**
```
Jim61C, line 85: Typo, Seen -> Unseen

examples/pytorch/hilander/models/focal_loss.py
```diff
@@ -0,0 +1,66 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
```
Jim61C, line 3: Do we still need focal_loss.py if it is not used now?

examples/pytorch/hilander/utils/knn.py
```diff
@@ -0,0 +1,173 @@
...
+        self.verbose = verbose
+        with Timer('[faiss] build index', verbose):
+            feats = feats.astype('float32')
+            size, dim = feats.shape
+            index = faiss.IndexFlatIP(dim)
+            index.add(feats)
+        with Timer('[faiss] query topk {}'.format(k), verbose):
+            sims, nbrs = index.search(feats, k=k)
+            self.knns = [(np.array(nbr, dtype=np.int32),
+                          1 - np.array(sim, dtype=np.float32))
```
Jim61C, line 149: One caveat here is that we would need to add a clamping of the cosine similarity to avoid the assertion error on line 55. No clamping should be ok most of the times but there are occasionally some nodes with neighboring large distances will cause this issue.

`                self.knns = [(np.array(nbr, dtype=np.int32),
                    1.0 - np.clip(np.array(sim, dtype=np.float32), 0.0, 1.0))
                    for nbr, sim in zip(nbrs, sims)]`

Jim61C (changes requested):
Comments on the latest codebase:
- REAME typos
- Necessity of focal_loss.py
- In faiss (CPU) version, request change to clamp cosine similarity to focus on [0, 1], same as in faiss GPU version (row 71 of faiss_search.py).

examples/pytorch/hilander/models/focal_loss.py
sneakerkg: I suggest we keep them since focal still perform better on some dataset like music. As long as this won't affect reproducing number, and considering we've already added them, we don't necessarily have to delete. Same thing for full graph training and inference, now all the numbers in the paper are get from sampled_training, but no harm to put the full graph code there.

examples/pytorch/hilander/utils/knn.py
sneakerkg: Discussed offline, it's OK to release without them since we don't need that to reproduce the numbers in the paper. If we verify no number change when adding this line or we need to add some numbers for dataset requires this line, let's add that.

sneakerkg (approved):

PR #3095, opened 2021-07-04
[Model] add activation in the last layer of graphSAGE model

Wang-Yu-Qing:
## Description
According to the SAGE paper, the forward and inference function should add activation in the last layer. However, current example of graphSAGE model has no activation in the last layer.

## Checklist
- [x] Code is well-documented

## Changes
- [x] Add activation in the last layer of forward and inference function in SAGE model.


BarclayII:
Even though the paper has a non-linear activation in the last layer, in practice we do not encourage doing so because it usually hinders model performance.

PR #3113, opened 2021-07-07
[Feature] enable edge reorder in dgl.reorder_graph()

Rhett-Ying:
## Description

1. change API name from dgl.reorder() to dgl.reorder_graph().
2. enable edge reorder: src/dst supported only.
3. add more examples into doc string.
4. add more tests accordingly.
## Checklist
- [x] Code is well-documented


python/dgl/transform.py
```diff
@@ -2902,27 +2902,36 @@ def sort_csc_by_tag(g, tag, tag_offset_name='_TAG_OFFSET'):
     return new_g
 
 
-def reorder(g, permute_algo='rcmk', store_ids=True, permute_config=None):
-    r"""Return a new graph which re-order and re-label the nodes
+def reorder_graph(g, node_permute_algo='rcmk', edge_permute_algo='src',
+                  store_ids=True, permute_config=None):
+    r"""Return a new graph with nodes and edges are re-ordered/re-labeled
```
jermainewang, line 2907: Return a new graph with nodes and edges ~~are~~ re-ordered/re-labeled

python/dgl/transform.py
```diff
@@ -2902,27 +2902,36 @@ def sort_csc_by_tag(g, tag, tag_offset_name='_TAG_OFFSET'):
...
     according to the specified permute algorithm.
 
     Support homogeneous graph only for the moment.
 
-    This API is basically implemented by leveraging :func:`~dgl.node_subgraph`,
-    so the function signature is similar and raw IDs could be stored
-    in ``dgl.NID`` and ``dgl.EID``.
+    This API is basically implemented by leveraging :func:`~dgl.node_subgraph`
+    and :func:`~dgl.edge_subgraph`, so the function signature is similar and
+    raw IDs could be stored in ``dgl.NID`` and ``dgl.EID`` via :attr:`store_ids`.
```
jermainewang, line 2914: Move this paragraph to the end since it is about implementation details. I actually prefer to delete it.

python/dgl/transform.py
```diff
@@ -2902,27 +2902,36 @@ def sort_csc_by_tag(g, tag, tag_offset_name='_TAG_OFFSET'):
...
 
-    Please note that edges are re-ordered/re-labeled according to re-ordered
-    ``'src'`` nodes. This behavior is realized in :func:`dgl.node_subgraph`.
-    What's more, if user wants to re-order/re-label according to ``'dst'`` nodes
-    or any other algorithms, please use :func:`dgl.edge_subgraph` with new edge
-    permutation.
+    Graph is re-ordered in 2 steps: re-order nodes first and edges then.
```
jermainewang, line 2916: The re-ordering has two 2 steps: first re-order nodes and then re-order endges.

python/dgl/transform.py
```diff
@@ -2936,6 +2941,11 @@ def reorder(g, permute_algo='rcmk', store_ids=True, permute_config=None):
           User should pass in ``'nodes_perm'`` via another argument :attr:`permute_config` with
           ``'custom'`` is specified here. By this way, can the graph be reordered according to
           passed in nodes permutation.
+    edge_permute_algo: str, optional
+        can be ``src`` or ``dst`` only. ``src`` is the default value.
+
+        * ``src``: generate edges permutation according to the src nodes of edges.
+        * ``dst``: generate edges permutation according to the dst nodes of edges.
```
jermainewang, line 2948:
```
    node_permute_algo: str, optional
        The permutation algorithm to re-order nodes. Options are ``'rcmk'`` or ``'metis'`` or ``'custom'``. ``'rcmk'`` is the default value.

        * ``'rcmk'``: Use the `Reverse Cuthill–McKee <https://docs.scipy.org/doc/scipy/reference/
          generated/scipy.sparse.csgraph.reverse_cuthill_mckee.html#
          scipy-sparse-csgraph-reverse-cuthill-mckee>`__ from ``'scipy'`` to generate nodes
          permutation.
        * ``'metis'``: Use the :func:`~dgl.partition.metis_partition_assignment` function
          to partition the input graph, which gives a cluster assignment of each node.
          DGL then sorts the assignment array so the new node order will put nodes of
          the same cluster together.
        * ``'custom'``: Reorder the graph according to the user-provided node
           permutation array (provided in :attr:`permute_config`)
    edge_permute_algo: str, optional
        The permutation algorithm to reorder edges. Options are ``src`` or ``dst``.
        ``src`` is the default value.

        * ``src``: Edges are arranged according to their source nodes.
        * ``dst``: Edges are arranged according to their source nodes.
    store_ids: bool, optional
        If True, DGL will store the original node and edge IDs in the ndata
        and edata of the resulting graph under name ``'dgl.NID'``
        and ``'dgl.EID'``, respectively.
    permute_config: dict, optional
        Additional key-value config data for the specified permutation algorithm.

        * For ``'rcmk'``, this argument is not required.
        * For ``'metis'``, users should specify the number of partitions ``'k'`` (e.g., 
           `permute_config={'k':10}` to partition the graph to 10 clusters).
        * For ``'custom'``, users should provide a node permutation array 
           `'nodes_perm'``. The array must be an integer list or a tensor with the
           same device of the input graph.
```

jermainewang (approved):
Comments on docstring. LGTM otherwise.

Issue #3107, opened 2021-07-06
Reduce memory consumption when create node/edge split for distributed training


JingchengYu94:
## 🚀 Feature
Make node_split and edge_split methods in python/dgl/distributed/dist_graph.py take less memory on each worker when number of workers increasing.

## Motivation
DGL provides node_split() and edge_split() to split the training, validation and test set at runtime for distributed training. But _split_even_to_part method called by node_split and edge_split in python/dgl/distributed/dist_graph.py creates tensor with length of total number of nodes/edges. 
In my case, I am training a large graph with 13 billion edges (training set) on a cluster of 200 workers. The edge_split method takes over 100GB memory. 

To be more specific, [this part](https://github.com/dmlc/dgl/blob/master/python/dgl/distributed/dist_graph.py#L1078-L1083) will create a boolean tensor and a int64 tensor whose length are both 13 billion. And the total memory consumption is 13 billion * (1 byte + 8 byte) = 108GB.

This is a bottleneck of scalability because adding workers cannot decrease the memory consumption.

## Pitch
1. A more scalable implementation of _split_even_to_part. We can compute the partition interval beforehand, and then create the nonzero_1d tensor based on the interval. There might be some extra math work, but it is worthy if 100GB memory can be saved.
2. [Optional] A compression method for creating node/edge split. For example, if the generated split consists of consecutive ids (1, 2, 3, .. 1 million), we can simply use an interval to represent it, and save more memory.

JingchengYu94:
@zheng-da Do you think this feature request reasonable? I am working on this to unblock my experiments. Maybe I can contribute my solution if you are interested.

zheng-da:
If you can contribute, it'll be great. our solution isn't very good. we definitely need to find a better way to split nodes/edges.

JingchengYu94:
PR merged. Close this issue.

PR #3132, opened 2021-07-12
[Feature] Optimize dist_graph/_split_even_to_part memory usage

JingchengYu94:
## Description
According to [this issue](https://github.com/dmlc/dgl/issues/3107), this PR will make the memory consumption of edge_split and node_split decrease when number of worker increases. In my case (graph with 13 billion edges), the memory consumption of edge_split decreases from ~108GB to ~15GB.

@zheng-da Could you please take a look at this PR?

## Checklist
- [x] Code is well-documented

## Changes
* Add API count_nonzero in dgl backend, to count the number of nonzero element in a tensor
* Add API add in dgl backend, to addup a tensor with a scalar (element-wise)
* Change the process of split_even
  * Old: compute nonzero tensor over whole elements, then compute the offset and take the slice
  * New: count the number of nonzero elements, compute offset, iterate through whole elements by blocks, compute nonzero tensor of each block, and concatenate to final answer. Block size is set to be #elements / #partitions, so it will decrease with worker increases.
* Note: If elements is a dist_tensor, whole elements will still be pulled as before. We need to implement count_nonzero for kvstore before applying this change to the case of dist_tensor. We can discuss about this further.


BarclayII:
We support PyTorch as old as 1.5.0 so I would recommend you use numpy's `count_nonzero` for PyTorch as well.

python/dgl/backend/backend.py
```diff
@@ -1125,6 +1125,38 @@ def replace_inf_with_zero(x):
...
+        The tensor to be counted
+
+    Returns
+    -------
+    Integer
+        The result 
+    """
+    pass
+
+def add(input, other):
```
BarclayII, line 1143: I think you can directly write `+` instead of defining this new function.

JingchengYu94: Good suggestion. Updated.

JingchengYu94:
@zheng-da Could you please take a look at this PR?

BarclayII (approved):

python/dgl/backend/pytorch/tensor.py
```diff
@@ -290,6 +291,10 @@ def clamp(data, min_val, max_val):
 def replace_inf_with_zero(x):
     return th.masked_fill(x, th.isinf(x), 0)
 
+def count_nonzero(input):
+    # TODO: fallback to numpy for backward compatibility
+    return np.count_nonzero(input)
```
zheng-da, line 296: do we not need to convert it into numpy array first?

python/dgl/distributed/dist_graph.py
```diff
@@ -1075,26 +1075,53 @@ def _even_offset(n, k):
...
-
-    # compute the offset of each split and ensure that the difference of each partition size
-    # is 1.
-    offsets = _even_offset(len(eles), partition_book.num_partitions())
-    assert offsets[-1] == len(eles)
-
-    # Get the elements that belong to the partition.
-    partid = partition_book.partid
-    part_eles = eles[offsets[partid] : offsets[partid + 1]]
+        elements = F.tensor(elements)
```
zheng-da, line 1095: here `elements` is a vector stored in a local machine?
if we can store the entire `elements` array in the local machine, why do we still have memory issue?

python/dgl/backend/pytorch/tensor.py
JingchengYu94: It will convert implicitly.

BarclayII: You don't have to.  numpy functions can work directly on PyTorch tensors.

python/dgl/distributed/dist_graph.py
JingchengYu94: elements is a boolean mask to split train, validate and test sets (I learnt how to use it from this [example](https://github.com/dmlc/dgl/blob/master/examples/pytorch/graphsage/experimental/train_dist_unsupervised.py#L429-L430)). I agree this is not the best design. 
But in this case, elements is boolean tensor, so it cost 1/8 memory of the nonzero_1d tensor, so at least my machines have enough memory for it.
Do you have any plan about redesign the data split part to further reduce the memory cost?

BarclayII: `elements` is a boolean vector so it consumes 8x less memory than an int64 tensor.  `nonzero` will give you an int64 tensor whose length is roughly the same order of magnitude.

See also https://discuss.dgl.ai/t/edge-split-consumes-too-much-memory/2130.

zheng-da: i see. then my general comment is that we can unify the code for DistTensor and local tensor.

JingchengYu94: Yes. But we need to implement a distributed count_nonzero method first. I can do it next week, what do you think?

PR #3150, opened 2021-07-17
Remove DGL_LOADALL=true

lululxvi:
## Description
DGL_LOADALL=true is not used in the code anymore.

## Changes
Remove DGL_LOADALL=true


VoVAllen (approved):
Thanks!

PR #3158, opened 2021-07-19
[Regression] Speed up random walks

BarclayII:
We have noticed that random walks have a speed regression of around 30%.  The reason was the repetitive call of `CSRHasData`.

With this fix, random walks on livejournal with 1000 nodes and length 10 reduces from 2213μs to 334μs per call (averaged over 5000 calls) on g3.16x.

Previously @VoVAllen warned us about this regression but somehow I failed to reproduce it last time so I dismissed it...


VoVAllen:
@dgl-bot run r5.16xlarge random_walk dmlc/master

BarclayII:
@dgl-bot run r5.16xlarge random_walk dmlc/0.6.x

VoVAllen (approved):

PR #3176, opened 2021-07-23
[Bugfix] fix potential starving in socket receiver

JingchengYu94:
## Description
Current implementation of [SocketReceiver::Recv](https://github.com/dmlc/dgl/blob/master/src/rpc/network/socket_communicator.cc#L222-L236) may cause starving of sender. In extreme cases, sender_0 keeps sending messages, each time receiver enters SocketReceiver::Recv function, it will grab a message from sender 0 and return. This will make other senders never be visited.

This PR changes Recv to continue iterating from the next sender of last Recv call.

In my experiment with 200 workers, the time of sample neighbor will be 5~10x longer for worker 199 comparing to worker 0. After fix, we have a roughly 15% end-to-end throughput boost.

## Checklist
- [x] Code is well-documented


VoVAllen (approved):
LGTM. We haven't tested on such large scale yet. And the fairness scheduling is not formally addressed before. Thanks for your contribution!

JingchengYu94:
Unittest fails  with 'src/rpc/network/msg_queue.cc:27: Message is larger than the queue.'. 
I don't think my change will affect message queue size? @VoVAllen do you have any idea about this?

JingchengYu94:
Fixed unittest.

PR #3190, opened 2021-07-26
[Model] add model example EvolveGCN.

maqy1995:
## Description
EvolveGCN with DGL,  only the Elliptic dataset is used now.   
related discussion: https://discuss.dgl.ai/t/temporal-graph-neural-networks/1485  
@BarclayII 

## Checklist
- [ ] Code is well-documented
- [x] To my best knowledge, examples are either not affected by this change,


examples/pytorch/evolveGCN/model.py
```diff
@@ -0,0 +1,201 @@
...
+            W = self.gcn_weights_list[i]
+            for j, g in enumerate(g_list):
+                # Attention: I try to use the below code to set gcn.weight(similar to pyG_temporal),
+                # but it doesn't work. It seems that the gradient function lost in this situation,
+                # more discussion see here: https://github.com/benedekrozemberczki/pytorch_geometric_temporal/issues/80
+                # ====================================================
+                # W = self.gnn_convs[i].weight[None, :, :]
+                # W, _ = self.recurrent_layers[i](W)
+                # self.gnn_convs[i].weight = nn.Parameter(W.squeeze())
+                # ====================================================
```
BarclayII, line 196: It won't work because the weight parameter is replaced as a whole.

Usually the preferred way to implement NN modules with intermediate results as weights is to express the NN module itself as a function (e.g. `torch.nn.Conv2d` is implemented by `torch.nn.functional.conv2d`, and similar for `torch.nn.functional.PReLU`).  For our GCN in particular though you can supply the GCN weight as `weight` keyword argument in `GraphConv.forward`.  This does not work for other modules.

@jermainewang what's your opinion on this case?  Should we put functionizing the NN modules into backlog?

examples/pytorch/evolveGCN/model.py
```diff
@@ -0,0 +1,201 @@
...
+
+        self.U = Parameter(torch.Tensor(rows, rows))
+        self.reset_param(self.U)
+
+        self.bias = Parameter(torch.zeros(rows, cols))
+
+    def reset_param(self, t):
+        # Initialize based on the number of columns
+        stdv = 1. / math.sqrt(t.size(1))
+        t.data.uniform_(-stdv, stdv)
```
BarclayII, line 63: Is this just `torch.nn.init.xavier_uniform_`?

examples/pytorch/evolveGCN/model.py
```diff
@@ -0,0 +1,201 @@
...
+        self.k = k
+
+    def reset_parameters(self, t):
+        stdv = 1. / math.sqrt(t.size(0))
+        t.data.uniform_(-stdv, stdv)
+
+    def forward(self, node_embs):
+        scores = node_embs.matmul(self.scorer) / self.scorer.norm()
+        vals, topk_indices = scores.view(-1).topk(self.k)
+        tanh = torch.nn.Tanh()
```
BarclayII, line 94: You can just use `torch.tanh`

examples/pytorch/evolveGCN/model.py
```diff
@@ -0,0 +1,201 @@
...
+        self.reset_parameters(self.scorer)
+
+        self.k = k
+
+    def reset_parameters(self, t):
+        stdv = 1. / math.sqrt(t.size(0))
+        t.data.uniform_(-stdv, stdv)
+
+    def forward(self, node_embs):
+        scores = node_embs.matmul(self.scorer) / self.scorer.norm()
```
BarclayII, line 92: Would any value in `self.scorer` go to 0 or super small so the result explodes?  This seems dangerous.

maqy1995: It's reasonable, is there a way to deal with this situation?

examples/pytorch/evolveGCN/model.py
maqy1995: done.

examples/pytorch/evolveGCN/model.py
maqy1995: done.

examples/pytorch/evolveGCN/model.py
BarclayII: You can probably do it by clipping to or adding an epsilon:
```
x / (self.scorer.norm() + 1e-6)
# or
x / self.scorer.norm().clamp(min=1e-6)
```

maqy1995: done.

BarclayII (approved):

PR #3200, opened 2021-07-29
[Feature] Implement one thread multiple socket

JingchengYu94:
## Description

Implement one thread multiple socket with epoll on linux platform. 
Current SocketSender and SocketReceiver will create one thread for each socket. When number of nodes, number of servers/trainers/samplers in each node increase, there will be several thousands of threads created for socket communicating. That is a waste of resources because creating threads requires some memory, and context switch between threads takes time.

### Changes
* Add an argument of max_thread_count for socket communicator's sender and receiver. They will create at most max_thread_count threads instead of one thread for each socket. If max_thread_count is set to zero (by default), it will fallback to one thread per socket.
* Add a class SocketPool to maintain a group of sockets. The sockets in SocketPool will be set to nonblocking, and SocketPool provides a blocking interface GetActiveSocket to get a socket has something to read/write.
* SocketSender now uses thread-level shared message queue. So one thread can handle multiple sending sockets easily.
* SocketReceiver now uses SocketPool to maintain its sockets. Each time this thread will take an active socket out of the pool, and receive message from it.
* Add a semphore in SocketReceiver to prevent busy wait in SocketReceiver::Recv.
* Add a CMake flag 'USE_EPOLL' with default to OFF to control this feature. 
* Currently only supports linux platform because epoll is a linux kernel system call. There are equivalents in other platforms like kqueue for FreeBSD, /dev/poll for Solaris, etc. And there are some cross-platform event-based libraries like [libevent](https://libevent.org/). We can do this in the future.
### Results 
We tested the change in a cluster with 50 containers, each has 4 cores,  2 servers and 2 clients (just like the DGL training model, with 2 graph server, 1 trainer and 1 sampler), and they do all-to-all communications. Every iteration, all clients send a empty rpc messages to all servers, then wait for their responses. 
In current implement, there will be 800 threads each container. With nonblocking socket with epoll, the average iteration time is 8% shorter than current implementation. And with adjusting max_thread_count argument, we can have the same boost with 400 threads each container.

Update: With this change: Add a semphore in SocketReceiver to prevent busy wait in SocketReceiver::Recv, there is another 30% speedup. 

## Checklist
- [x] Code is well-documented

### DIFF
--- a/src/graph/network.cc
+++ b/src/graph/network.cc
@@ -206,7 +206,7 @@ DGL_REGISTER_GLOBAL("network._CAPI_DGLSenderCreate")
     int64_t msg_queue_size = args[1];
     network::Sender* sender = nullptr;
     if (type == "socket") {
-      sender = new network::SocketSender(msg_queue_size);
+      sender = new network::SocketSender(msg_queue_size, 0);
     } else {
       LOG(FATAL) << "Unknown communicator type: " << type;
     }
@@ -220,7 +220,7 @@ DGL_REGISTER_GLOBAL("network._CAPI_DGLReceiverCreate")
     int64_t msg_queue_size = args[1];
     network::Receiver* receiver = nullptr;
     if (type == "socket") {
-      receiver = new network::SocketReceiver(msg_queue_size);
+      receiver = new network::SocketReceiver(msg_queue_size, 0);
     } else {
       LOG(FATAL) << "Unknown communicator type: " << type;
     }
--- a/src/rpc/network/communicator.h
+++ b/src/rpc/network/communicator.h
@@ -28,11 +28,14 @@ class Sender {
   /*!
    * \brief Sender constructor
    * \param queue_size size (bytes) of message queue. 
+   * \param max_thread_count size of thread pool. 0 for no limit
    * Note that, the queue_size parameter is optional.
    */
-  explicit Sender(int64_t queue_size = 0) {
+  explicit Sender(int64_t queue_size = 0, int max_thread_count = 0) {
     CHECK_GE(queue_size, 0);
+    CHECK_GE(max_thread_count, 0);
     queue_size_ = queue_size;
+    max_thread_count_ = max_thread_count;
   }
 
   virtual ~Sender() {}
@@ -86,6 +89,10 @@ class Sender {
    * \brief Size of message queue
    */
   int64_t queue_size_;
+  /*!
+   * \brief Size of thread pool. 0 for no limit
+   */
+  int max_thread_count_;
 };
 
 /*!
@@ -101,13 +108,16 @@ class Receiver {
   /*!
    * \brief Receiver constructor
    * \param queue_size size of message queue.
+   * \param max_thread_count size of thread pool. 0 for no limit
    * Note that, the queue_size parameter is optional.
    */
-  explicit Receiver(int64_t queue_size = 0) {
+  explicit Receiver(int64_t queue_size = 0, int max_thread_count = 0) {
     if (queue_size < 0) {
       LOG(FATAL) << "queue_size cannot be a negative number.";
     }
+    CHECK_GE(max_thread_count, 0);
     queue_size_ = queue_size;
+    max_thread_count_ = max_thread_count;
   }
 
   virtual ~Receiver() {}
@@ -165,6 +175,10 @@ class Receiver {
    * \brief Size of message queue
    */
   int64_t queue_size_;
+  /*!
+   * \brief Size of thread pool. 0 for no limit
+   */
+  int max_thread_count_;
 };
 
 }  // namespace network
--- a/src/rpc/network/socket_communicator.cc
+++ b/src/rpc/network/socket_communicator.cc
@@ -12,6 +12,7 @@
 
 #include "socket_communicator.h"
 #include "../../c_api_common.h"
+#include "socket_pool.h"
 
 #ifdef _WIN32
 #include <windows.h>
@@ -51,15 +52,20 @@ void SocketSender::AddReceiver(const char* addr, int recv_id) {
   address.ip = ip_and_port[0];
   address.port = std::stoi(ip_and_port[1]);
   receiver_addrs_[recv_id] = address;
-  msg_queue_[recv_id] =  std::make_shared<MessageQueue>(queue_size_);
 }
 
 bool SocketSender::Connect() {
   // Create N sockets for Receiver
+  int receiver_count = static_cast<int>(receiver_addrs_.size());
+  if (max_thread_count_ == 0 || max_thread_count_ > receiver_count) {
+    max_thread_count_ = receiver_count;
+  }
+  sockets_.resize(max_thread_count_);
   for (const auto& r : receiver_addrs_) {
-    int ID = r.first;
-    sockets_[ID] = std::make_shared<TCPSocket>();
-    TCPSocket* client_socket = sockets_[ID].get();
+    int receiver_id = r.first;
+    int thread_id = receiver_id % max_thread_count_;
+    sockets_[thread_id][receiver_id] = std::make_shared<TCPSocket>();
+    TCPSocket* client_socket = sockets_[thread_id][receiver_id].get();
     bool bo = false;
     int try_count = 0;
     const char* ip = r.second.ip.c_str();
@@ -83,12 +89,17 @@ bool SocketSender::Connect() {
     if (bo == false) {
       return bo;
     }
-    // Create a new thread for this socket connection
-    threads_[ID] = std::make_shared<std::thread>(
-      SendLoop,
-      client_socket,
-      msg_queue_[ID].get());
   }
+
+  for (int thread_id = 0; thread_id < max_thread_count_; ++thread_id) {
+    msg_queue_.push_back(std::make_shared<MessageQueue>(queue_size_));
+    // Create a new thread for this socket connection
+    threads_.push_back(std::make_shared<std::thread>(
+      SendLoop,
+      sockets_[thread_id],
+      msg_queue_[thread_id]));
+  }
+
   return true;
 }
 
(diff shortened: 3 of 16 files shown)


VoVAllen:
Thanks for your contribution. 
I checked the error, the `event-config.h` is generated, so you need to add `target_include_directories(dgl PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/third_party/libevent/include/")` to the CMakeLists.txt. And you can also add `EVENT__DISABLE_SAMPLES` to disable compling the samples

Do you have any number to share about this improvement? Overall looks good to me.

JingchengYu94:
> Thanks for your contribution.
> I checked the error, the `event-config.h` is generated, so you need to add `target_include_directories(dgl PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/third_party/libevent/include/")` to the CMakeLists.txt. And you can also add `EVENT__DISABLE_SAMPLES` to disable compling the samples
> 
> Do you have any number to share about this improvement? Overall looks good to me.

Thanks for your advice. I will run some experiments and get some number.

JingchengYu94:
libevent may stuck sometimes, I don't have time to debug it recently. So I decided to just implement epoll version first.

VoVAllen:
It seems sem_init is not available on windows. I would suggest using std::mutex or other primitive from std library

VoVAllen:
Also I think epoll is fine, and it can also save the cpu resources, which can lead to better end2end performance

JingchengYu94:
> It seems sem_init is not available on windows. I would suggest using std::mutex or other primitive from std library

semaphore may have a better performance in this case. Let me try to add a semaphore_wrapper for cross platform.

src/rpc/network/socket_communicator.h
```diff
@@ -196,28 +209,39 @@ class SocketReceiver : public Receiver {
...
+   * \brief Recv-loop for each thread
+   * \param sockets client sockets of current thread
+   * \param queue message queues of current thread
    *
    * Note that, the RecvLoop will finish its loop-job and exit thread
    * when the main thread invokes Signal() API on the message queue.
    */ 
-  static void RecvLoop(TCPSocket* socket, MessageQueue* queue);
+  static void RecvLoop(
+    std::unordered_map<int, std::shared_ptr<TCPSocket>> sockets,
```
VoVAllen, line 242: Add comments on what's int (Receiver ID?)

src/rpc/network/socket_communicator.h
```diff
@@ -196,28 +209,39 @@ class SocketReceiver : public Receiver {
...
+    std::unordered_map<int, std::shared_ptr<MessageQueue>> queues,
```
VoVAllen, line 243: Add comments on what's int (thread ID?)

src/rpc/network/socket_pool.h
```diff
@@ -0,0 +1,142 @@
...
+  /*!
+   * \brief array of epoll event
+   */
+  epoll_event events_[MAX_EVENTS];
+};
+
+}  // namespace network
+}  // namespace dgl
+
+#else
```
VoVAllen, line 104: It's hard to track the API if you break it in the header with different conditions. Could you try make a unified interface here and let the different implementation happens in the cc file?

src/rpc/network/communicator.h
```diff
@@ -28,11 +28,14 @@ class Sender {
   /*!
    * \brief Sender constructor
    * \param queue_size size (bytes) of message queue. 
+   * \param max_thread_count size of thread pool. 0 for no limit
    * Note that, the queue_size parameter is optional.
    */
-  explicit Sender(int64_t queue_size = 0) {
+  explicit Sender(int64_t queue_size = 0, int max_thread_count = 0) {
     CHECK_GE(queue_size, 0);
```
VoVAllen, line 35: Is there any check needed on `max_thread_count` when built without epoll support?

VoVAllen (commented):
Overall looks good to me. Please help add comment on all places related to ID to specify whether it's thread ID or sender/receiver ID, since we have multiple levels of abstraction here.

src/rpc/network/socket_pool.h
VoVAllen: You can consider using template or something like`if (EPOLL_DEFINER) {...} else {...}`

JingchengYu94: Sure

src/rpc/network/communicator.h
JingchengYu94: no, Sender's thread limit just makes several senders to share one msg queue, not related to epoll. And receiver's thread limit will be handled in SocketReceiver::Wait

JingchengYu94:
Update to resolve the comments

src/rpc/network/socket_communicator.cc
```diff
@@ -255,65 +288,107 @@ void SocketReceiver::Finalize() {
...
+  runtime::Semaphore *queue_sem) {
+  std::unordered_map<int, std::unique_ptr<RecvContext>> recv_contexts;
+  SocketPool socket_pool;
+  for (auto& socket : sockets) {
+    auto &sender_id = socket.first;
+    socket_pool.AddSocket(socket.second, sender_id);
+    recv_contexts[sender_id] = std::unique_ptr<RecvContext>(new RecvContext());
+  }
+
+  for (;;) {
```
zheng-da, line 352: i don't quite understand the code here. maybe have some comments to explain it?

src/rpc/network/socket_pool.cc
```diff
@@ -0,0 +1,102 @@
...
+
+std::shared_ptr<TCPSocket> SocketPool::GetActiveSocket(int* socket_id) {
+  while (pending_fds_.empty()) {
+    Wait();
+  }
+  int fd = pending_fds_.front();
+  pending_fds_.pop();
+  if (socket_ids_.find(fd) == socket_ids_.end()) {
+    // This socket has been removed
+    return GetActiveSocket(socket_id);
```
zheng-da, line 82: can you use for or while loop instead of using recursive function call?

src/rpc/network/tcp_socket.cc
```diff
@@ -119,7 +119,7 @@ bool TCPSocket::Accept(TCPSocket * socket, std::string * ip, int * port) {
 }
 
 #ifdef _WIN32
-bool TCPSocket::SetBlocking(bool flag) {
+bool TCPSocket::SetNonBlocking(bool flag) {
```
zheng-da, line 122: you didn't change the code in the function, but changed the meaning of the function?

src/runtime/semaphore_wrapper.cc
```diff
@@ -0,0 +1,47 @@
+/*!
+ *  Copyright (c) 2021 by Contributors
+ * \file semaphore_wrapper.cc
+ * \brief A simple corss platform semaphore wrapper
+ */
+#include "semaphore_wrapper.h"
```
zheng-da, line 6: stl has semaphore. should we use it?

src/runtime/semaphore_wrapper.cc
```diff
@@ -0,0 +1,47 @@
...
+
+#include <dmlc/logging.h>
+
+namespace dgl {
+namespace runtime {
+
+#ifdef _WIN32
+
+Semaphore::Semaphore() {
+  sem_ = CreateSemaphore(nullptr, 0, INT_MAX, nullptr);
```
zheng-da, line 16: what library provides this semaphore?

src/graph/network.cc
```diff
@@ -205,7 +205,7 @@ DGL_REGISTER_GLOBAL("network._CAPI_DGLSenderCreate")
     int64_t msg_queue_size = args[1];
     network::Sender* sender = nullptr;
     if (type == "socket") {
-      sender = new network::SocketSender(msg_queue_size);
+      sender = new network::SocketSender(msg_queue_size, 0);
```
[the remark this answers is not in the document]
JingchengYu94: I am a little confused about this network.cc. It is used in network.py, which is only used in python/dgl/contrib/dis_kvstore.py and python/dgl/contrib/dis_sampler.py. What's the difference between contrib/dis_kvstore.py and distributed/kvstore.py?

src/rpc/network/socket_communicator.cc
```diff
@@ -187,44 +207,56 @@ bool SocketReceiver::Wait(const char* addr, int num_sender) {
...
+      msg_queue_,
+      &queue_sem_));
   }
 
   return true;
 }
 
 STATUS SocketReceiver::Recv(Message* msg, int* send_id) {
-  // loop until get a message
+  queue_sem_.Wait();
```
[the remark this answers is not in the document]
JingchengYu94: To prevent busy query of msg_queue_, this will bring some performance gain, and save some cpu resources.

src/rpc/network/tcp_socket.cc
JingchengYu94: Because when flag is true, this function will make the socket nonblocking. So I think it is more consistent to change the function name.

JingchengYu94: This function is not used before, so it is safe to change the function name

src/runtime/semaphore_wrapper.cc
JingchengYu94: stl's <semaphore> requires c++20, and my environment only supports c++11...

src/runtime/semaphore_wrapper.cc
JingchengYu94: <windows.h>

src/graph/network.cc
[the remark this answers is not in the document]
zheng-da: the kvstore of contrib is no longer maintained. we should remove it.

src/rpc/network/socket_pool.cc
JingchengYu94: done

src/rpc/network/socket_communicator.cc
JingchengYu94: done

src/rpc/network/socket_communicator.cc
```diff
@@ -187,44 +207,56 @@ bool SocketReceiver::Wait(const char* addr, int num_sender) {
...
   // Accept all sender sockets
   std::string accept_ip;
   int accept_port;
+  sockets_.resize(max_thread_count_);
   for (int i = 0; i < num_sender_; ++i) {
-    sockets_[i] = std::make_shared<TCPSocket>();
-    if (server_socket_->Accept(sockets_[i].get(), &accept_ip, &accept_port) == false) {
+    int thread_id = i % max_thread_count_;
+    auto socket = std::make_shared<TCPSocket>();
+    sockets_[thread_id][i] = socket;
```
zheng-da, line 235: is the index `i` here correct?

src/rpc/network/socket_communicator.h
```diff
@@ -196,28 +210,41 @@ class SocketReceiver : public Receiver {
...
+  std::vector<std::shared_ptr<std::thread>> threads_;
 
   /*!
-   * \brief Recv-loop for each socket in per-thread
-   * \param socket client socket
-   * \param queue message queue
+   * \brief queue_sem_ semphore to indicate number of messages in the message
+   * queue to prevent busy wait of Recv
+   */
+  runtime::Semaphore queue_sem_;
```
zheng-da, line 232: should we keep this in message queue?

src/rpc/network/socket_communicator.cc
```diff
@@ -83,82 +89,96 @@ bool SocketSender::Connect() {
...
+    while (mq->Empty() == false) {
 #ifdef _WIN32
         // just loop
 #else   // !_WIN32
         usleep(1000);
 #endif  // _WIN32
     }
-    int ID = mq.first;
-    mq.second->SignalFinished(ID);
+    mq->SignalFinished(i);
```
zheng-da, line 128: why do we need `i` as input argument?

src/rpc/network/socket_communicator.cc
```diff
@@ -187,44 +207,56 @@ bool SocketReceiver::Wait(const char* addr, int num_sender) {
...
   std::string accept_ip;
   int accept_port;
+  sockets_.resize(max_thread_count_);
   for (int i = 0; i < num_sender_; ++i) {
-    sockets_[i] = std::make_shared<TCPSocket>();
-    if (server_socket_->Accept(sockets_[i].get(), &accept_ip, &accept_port) == false) {
+    int thread_id = i % max_thread_count_;
+    auto socket = std::make_shared<TCPSocket>();
+    sockets_[thread_id][i] = socket;
+    msg_queue_[i] = std::make_shared<MessageQueue>(queue_size_);
```
zheng-da, line 236: the message queue is per thread or per sender?

src/rpc/network/socket_communicator.h
```diff
@@ -196,28 +210,41 @@ class SocketReceiver : public Receiver {
...
-  std::unordered_map<int /* Sender (virutal) ID */, std::shared_ptr<TCPSocket>> sockets_;
+  std::vector<std::unordered_map<int /* Sender (virutal) ID */,
+    std::shared_ptr<TCPSocket>>> sockets_;
 
   /*!
    * \brief Message queue for each socket connection
    */ 
-  std::unordered_map<int /* Sender (virtual) ID */, std::shared_ptr<MessageQueue>> msg_queue_;
+  std::unordered_map<int /* Sender (virtual) ID */,
+    std::shared_ptr<MessageQueue>> msg_queue_;
```
zheng-da, line 220: it's a bit strange. you made msg_queue here is per socket, but the message queue in the sender is per thread.

src/rpc/network/socket_communicator.cc
```diff
@@ -255,65 +288,122 @@ void SocketReceiver::Finalize() {
...
+      data_size = RecvDataSize(socket.get());
+      if (data_size > 0) {
+        try {
+          buffer = new char[data_size];
+        } catch(const std::bad_alloc&) {
+          LOG(FATAL) << "Cannot allocate enough memory for message, "
+                     << "(message size: " << data_size << ")";
+        }
+        received_bytes = 0;
+      } else if (data_size == 0) {
```
zheng-da, line 385: this means the connection is closed?

src/rpc/network/socket_communicator.cc
```diff
@@ -255,65 +288,122 @@ void SocketReceiver::Finalize() {
...
+    // Nonblocking socket might be interrupted at any point. So we need to
+    // store the partially received data
+    std::unique_ptr<RecvContext> &ctx = recv_contexts[sender_id];
+    int64_t &data_size = ctx->data_size;
+    int64_t &received_bytes = ctx->received_bytes;
+    char*& buffer = ctx->buffer;
+
+    if (data_size == -1) {
+      // This is a new message, so receive the data size first
+      data_size = RecvDataSize(socket.get());
```
zheng-da, line 376: can we always receive the complete data size?

src/rpc/network/socket_communicator.cc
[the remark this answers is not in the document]
zheng-da: i see. `queue_sem_` is a bit confusing. this semaphore indicates that there is a message in one of the queues.

src/rpc/network/socket_communicator.h
```diff
@@ -196,28 +210,41 @@ class SocketReceiver : public Receiver {
...
    */ 
-  std::unordered_map<int /* Sender (virtual) ID */, std::shared_ptr<std::thread>> threads_;
+  std::vector<std::shared_ptr<std::thread>> threads_;
 
   /*!
-   * \brief Recv-loop for each socket in per-thread
-   * \param socket client socket
-   * \param queue message queue
+   * \brief queue_sem_ semphore to indicate number of messages in the message
+   * queue to prevent busy wait of Recv
```
zheng-da, line 230: the semaphore indicates the number of messages in multiple message queues.

src/rpc/network/socket_communicator.cc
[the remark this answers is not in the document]
zheng-da: maybe here require some comments. It took me a while to understand what is going on here. Initially, I was thinking why we cannot use condition if we just want to wake up the thread for receiving a message.

We need a mechanism to indicate the total number of messages stored in multiple queues. Each time the method is invoked, it returns only one message from one of the queues and the counter in the semaphore is decreased by one.

zheng-da:
overall, it looks good to me. just some small comments.

zheng-da (approved):

src/rpc/network/socket_communicator.cc
JingchengYu94: This argument should be zero.

src/rpc/network/socket_communicator.cc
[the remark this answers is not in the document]
JingchengYu94: I will add some comments here

src/rpc/network/socket_communicator.cc
JingchengYu94: yes, this RecvDataSize method is blocking and will always receive full 8 bytes.

src/rpc/network/socket_communicator.cc
JingchengYu94: yes

src/rpc/network/socket_communicator.h
JingchengYu94: Good question. Let me try to unify them. Maybe we can get rid of the semaphore.

JingchengYu94: Just took a closer look. Seems like I did this to keep RecvFrom(msg, send_id) easy. If multiple senders share one msg_queue, this RecvFrom will be hard to write. So I think we can keep current design.

src/rpc/network/socket_communicator.cc
```diff
@@ -83,82 +89,96 @@ bool SocketSender::Connect() {
...
+    while (mq->Empty() == false) {
 #ifdef _WIN32
         // just loop
 #else   // !_WIN32
         usleep(1000);
 #endif  // _WIN32
     }
-    int ID = mq.first;
-    mq.second->SignalFinished(ID);
+    mq->SignalFinished(0);
```
zheng-da, line 128: why is this 0?

PR #3203, opened 2021-07-30
[feature] add count_nonzero function for DistTensor

freeliuzc:
## Description
According to #3107 and #3132, we should develop a new feature for DistTensor to count nonzero value in a optimized way. By registering a new service in kvstore, I develop this function without the need for pulling other machine's data.

@zheng-da Could you please take a look at this PR?

## Checklist
- [x] Code is well-documented

## Changes
- [x] add API count_nonzero in class DistTensor
- [x] add new request/response COUNT_LOCAL_NONZERO in kvstore.py
- [x] optimize _split_even_to_part code in dist_graph.py using this feature


freeliuzc:
Hello @VoVAllen , could you please review this PR

python/dgl/distributed/dist_graph.py
```diff
@@ -1229,48 +1229,38 @@ def _split_even_to_part(partition_book, elements):
...
+    # Get the elements that belong to the partition.
+    partid = partition_book.partid
+    left, right = offsets[partid], offsets[partid + 1]
+
+    x = y = 0
+    num_elements = len(elements)
+    block_size = num_elements // partition_book.num_partitions()
+    part_eles = None
+    # compute the nonzero tensor of each partition instead of whole tensor to save memory
+    for idx in range(0, num_elements, block_size):
```
VoVAllen, line 1250: I'm not familiar with this part. @zheng-da Could you take a look on this?

VoVAllen (commented):
Thanks for your contribution! I reviewed the kvstore count_non_zero part and it looks good to me. 

Could you add a unit test in `tests/distributed/kv_store/test_new_kvstore.py`?

python/dgl/distributed/dist_graph.py
VoVAllen: Also How did this part save the memory?

freeliuzc: Thanks for your review. 
I didn't change this part here, it's pulled request in #3132. 
But after developing count_nonzero function, I optimized it a little without any function changed. 
Count_nonzero funciton is like "pull data function" in kvstore, which should be tested in many machines.
Unfortunately, the unit tests only can be tested in local machine with some servers. 
So I have used four machines with different ip to test it in distributed mode . The result is correct.

python/dgl/distributed/dist_graph.py
```diff
@@ -1229,48 +1229,38 @@ def _split_even_to_part(partition_book, elements):
...
+        x = y
+        y += len(nonzero_block)
+        if y > left and x < right:
+            start = max(x, left) - x
+            end = min(y, right) - x
+            tmp = nonzero_block[start:end] + idx
+            if part_eles is None:
+                part_eles = tmp
+            else:
+                part_eles = F.cat((part_eles, tmp), 0)
```
zheng-da, line 1261: to avoid too many concatenations, can you collect all segments from different blocks first and use one concatenation in the end?

python/dgl/distributed/kvstore.py
```diff
@@ -1283,6 +1343,38 @@ def _take_id(self, elem):
...
+        int
+            the number of nonzero in this data.
+        """
+        total = 0
+        pull_count = 0
+        for machine_id in range(self._machine_count):
+            if machine_id == self._machine_id:
+                local_id = F.tensor(np.arange(self._part_policy[name].get_part_size(),
+                                              dtype=np.int64))
+                local_data = self._pull_handlers[name](self._data_store, name, local_id)
```
zheng-da, line 1365: do you need to have a data copy here? since this is local machine, you can access the local tensor directly.

python/dgl/distributed/kvstore.py
```diff
@@ -1283,6 +1343,38 @@ def _take_id(self, elem):
...
+                server_id = self._main_server_id
+                local_response = PullResponse(server_id, local_data)
```
zheng-da, line 1367: why do you need this?

python/dgl/distributed/kvstore.py
freeliuzc: Yes, I could have used the simpler method you said. 
But in the ”pull data function“, it uses local_data in such way. I guess may be there some other reasons for this?
If possible, I would prefer to use local_data directly.

python/dgl/distributed/dist_graph.py
freeliuzc: Thanks for your suggestions, I will make a new commit for it later

freeliuzc: Hi, we tested on 200 machines with 10 billion edges, and the whole split process was completed in few seconds. 
If save all blocks first, and then cat, the memory usage will double in the worst case. 
So may be the current method is better?

zheng-da: i see. sure, the current solution also works.

python/dgl/distributed/kvstore.py
zheng-da: thanks

zheng-da (approved):

freeliuzc:
@zheng-da @VoVAllen Hi, can you help me merge this pull request？

VoVAllen:
@zheng-da

HuangLED:
I believe this PR caused a bug, when I ran train_dist.py script got following error: 

`Traceback (most recent call last):
  File "train_dist.py", line 319, in <module>
    main(args)
  File "train_dist.py", line 271, in main
    train_nid = dgl.distributed.node_split(g.ndata['train_mask'], pb, force_even=True)
  File "/home/tiger/miniconda3/envs/ruoyun/lib/python3.8/site-packages/dgl-0.8-py3.8-linux-x86_64.egg/dgl/distributed/dist_graph.py", line 1460, in node_split
    part_nid = _split_even_to_part(partition_book, nodes)
  File "/home/tiger/miniconda3/envs/ruoyun/lib/python3.8/site-packages/dgl-0.8-py3.8-linux-x86_64.egg/dgl/distributed/dist_graph.py", line 1307, in _split_even_to_part
    nonzero_count = elements.count_nonzero()
  File "/home/tiger/miniconda3/envs/ruoyun/lib/python3.8/site-packages/dgl-0.8-py3.8-linux-x86_64.egg/dgl/distributed/dist_tensor.py", line 244, in count_nonzero
    return self.kvstore.count_nonzero(name=self.name)
AttributeError: 'KVClient' object has no attribute 'count_nonzero'`

Revert to without this PR, everything works. 

@zheng-da  @freeliuzc

Issue #2760, opened 2021-03-18
RGCN generates nan in PyTorch 1.8 but not in PyTorch 1.7.x


nv-dlasalle:
## 🐛 Bug

Running `/examples/pytorch/rgcn/entity_classify_mp.py` results in the training accuracy staying near zero and the loss becomes `nan`.

## To Reproduce

1. Build from source, and run the example according to the readme:
```
$ OMP_NUM_THREADS=1 python3 entity_classify_mp.py -d ogbn-mag --testing --fanout='30,30' --batch-size 1024 --n-hidden 128 --lr 0.01 --num-worker 1 --eval-batch-size 8 --low-mem --gpu 0 --dropout 0.7 --use-self-loop --n-bases 2 --n-epochs 3 --dgl-sparse --sparse-lr 0.08
Using backend: pytorch
Namespace(batch_size=1024, dataset='ogbn-mag', dgl_sparse=True, dropout=0.7, eval_batch_size=8, fanout='30,30', gpu='0', l2norm=0, layer_norm=False, low_mem=True, lr=0.01, n_bases=2, n_epochs=3, n_hidden=128, n_layers=2, node_feats=False, num_workers=1, sparse_lr=0.08, use_self_loop=True, validation=False)
num_tasks = 1
task_type = multiclass classification
eval_metric = acc
num_classes = 349
is_hetero = True
Opening 'dataset/ogbn_mag_dgl/processed/dgl_data_processed'.
Number of relations: 8
Number of class: 349
Number of train: 629571
Number of valid: 64879
Number of test: 41939
0:author
1:field_of_study
2:institution
3:paper
start training...
Train Accuracy: 0.0117 | Train Loss: 5.9985
Train Accuracy: 0.0107 | Train Loss: nan
Train Accuracy: 0.0029 | Train Loss: nan
Train Accuracy: 0.0029 | Train Loss: nan
Train Accuracy: 0.0010 | Train Loss: nan
Train Accuracy: 0.0020 | Train Loss: nan
Train Accuracy: 0.0029 | Train Loss: nan
Train Accuracy: 0.0029 | Train Loss: nan
Train Accuracy: 0.0039 | Train Loss: nan
Train Accuracy: 0.0039 | Train Loss: nan
Train Accuracy: 0.0020 | Train Loss: nan
Train Accuracy: 0.0010 | Train Loss: nan
Train Accuracy: 0.0029 | Train Loss: nan
[...]

PR #3207, opened 2021-07-30
[Performance] Enable using pinned memory for transfers in SparseAdam optimizer.

nv-dlasalle:
## Description
In python 1.8, when `non_blocking=True` in a GPU->CPU copy, the output tensor is allocated in pinned memory (the cause of #2760). However, this now results in a performance regression between 0.6 and 0.7 when using torch>=1.8.

This PR re-enables using pinned memory for those transfers, while synchronizing afterwards to ensure correctness. This cuts time in the optimizer dramatically (on my system nearly 2x--but it will vary from system to system):
opt_time

## Checklist
- [x] Code is well-documented

## Changes

Used cuda events to allow specifying `non_blocking=True` when transferring from the gradient computation device (GPU) to the state storage device (CPU).

This also re-arranges some operations in order to better in order to reduce the amount of time the GPU needs to wait on the CPU to finish slicing.

Remove variables used for setting `non_blocking=False`, as this is the default.


classicsong (approved):
LGTM

PR #3219, opened 2021-08-04
[bugfix] Fix check for num_workers for using _ScalarDataBatcher

nv-dlasalle:
## Description
It looks like when we refactored in #3046, we changed the check for `num_workers == 0` to `num_workers != 0`. This PR restores the original functionality introduced in #2716, such that when the output device is on the GPU and the number of workers is 0, a pytorch tensor will be used to store the batch, rather than a python list.

This fix results in a ~2x reduction in the runtime of the dataloader, depending on the size of the fanouts.

## Checklist
- [x] Code is well-documented

## Changes
Changed the check for number of workers.

### DIFF
--- a/python/dgl/dataloading/pytorch/dataloader.py
+++ b/python/dgl/dataloading/pytorch/dataloader.py
@@ -353,7 +353,7 @@ def _init_dataloader(collator, device, dataloader_kwargs, use_ddp, ddp_seed):
     use_scalar_batcher = False
     scalar_batcher = None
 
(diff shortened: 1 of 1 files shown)


VoVAllen (approved):

VoVAllen (commented):
Sorry I don't think I fully understand the logic here. Please ignore my approval
@BarclayII should be the one review thie

BarclayII (approved):
That was silly.  Thanks for the fix.

PR #3226, opened 2021-08-06
[Performance][Feature] Implement edge excluding in EdgeDataLoader on GPU

nv-dlasalle:
## Description
When the output device for the dataloader is a GPU, this enable making use of the cuda hash table implementation to find the set of edges to remove.

In my testing using `/examples/pytorch/graphsage/train_sampling_unsupervised.py`, this reduced the time to find the edges from 148ms using numpy's `isin()` method, to 2.0ms to copy the edge ids to the GPU and check them against the excluded edges (74x speedup). The majority of this is performing the CPU to GPU copy. In terms of just checking the edges, it takes just 200us.

Although I believe outside of the scope of this PR, the CPU side of things would benefit from a similar strategy.

## Checklist
- [x] Code is well-documented

## Changes

Add Filter class to store the hashmap of edges to be excluded. This speeds up the process when there are multiple layers, as opposed to generating the same hashmap for each layer.

This also adds a unit test.


BarclayII:
@zheng-da How do you like it combined with #2971?  Ideally I would like to see the entry point of CPU and GPU implementations unified

python/dgl/utils/filter.py
```diff
@@ -0,0 +1,69 @@
+
+"""Utilities for finding overlap or missing items in arrays."""
+
+from .._ffi.function import _init_api
+from .. import backend as F
+
+
+class Filter(object):
+    """Class used to either find either find the subset of ids that are in this
```
BarclayII, line 9: "either find either find".  Also we usually capitalize "ids" as "IDs".

python/dgl/utils/filter.py
```diff
@@ -0,0 +1,69 @@
...
+    filter, or th subset of ids that are not in this filter,
```
BarclayII, line 10: "the"

BarclayII (approved):
After some inspection I found the PR largely orthogonal to #2971 since it is just moving the edge ID exclusion postprocessing to C++/CUDA.  Although if the latter was merged then we probably want to skip the edge ID exclusion postprocessing.

python/dgl/dataloading/dataloader.py
```diff
@@ -46,6 +46,81 @@ def _locate_eids_to_exclude(frontier_parent_eids, exclude_eids):
...
+            for _, v in exclude_eids.items():
+                if device is None:
+                    device = F.context(v)
+                    break
+        else:
+            device = F.context(exclude_eids)
+        self._exclude_eids = None
+        self._filter = None
+
+        if device == F.cpu():
```
BarclayII, line 62: Maybe add a TODO for CPU?  If the filters are implemented for CPU then we won't use `_locate_eids_to_exclude` right?

python/dgl/utils/filter.py
nv-dlasalle: Fixed in 
fe0bc66.

python/dgl/utils/filter.py
nv-dlasalle: Fixed in 
fe0bc66.

python/dgl/dataloading/dataloader.py
nv-dlasalle: Added in 
aa21891.

zheng-da (approved):

PR #3248, opened 2021-08-12
[Bugfix] set omp thread number for graph server

JingchengYu94:
## Description

The rpc call is actually computed in server process, so I think it is necessary to set openmp thread number for server as well.

If not set, openmp will choose an optimal number of threads, which is usually number of cpu cores. 

In my experiments with 50 machines, 16 cores each machine, with graphsage model and batchsize=10000, [neighbor sampling](https://github.com/dmlc/dgl/blob/master/src/array/cpu/rowwise_pick.h#L112) will usually have some thousands of rows to process.  With default omp thread number, the sampling will take about 30ms, and take up all cpu resources (~1300% cpu usage with 16 cores). When set omp thread number to 1, the sampling will take about 4ms, and use only 1 core (~100% cpu usage). 

I am not very clear about why fewer threads leads to faster sampling, maybe it is caused by omp thread creating overhead or something. But at least my experiment shows the necessity of letting user to choose the number of threads.

## Checklist
- [x] Code is well-documented

## Changes
set omp thread number for server in dist_context.py

### DIFF
--- a/python/dgl/distributed/dist_context.py
+++ b/python/dgl/distributed/dist_context.py
@@ -204,7 +204,7 @@ def initialize(ip_config, num_servers=1, num_workers=0,
 
         Default: ``'socket'``
     num_worker_threads: int
-        The number of threads in a worker process.
+        The number of omp threads in server and worker processes.
 
     Note
     ----
@@ -226,6 +226,7 @@ def initialize(ip_config, num_servers=1, num_workers=0,
             'Please define DGL_CONF_PATH to run DistGraph server'
         formats = os.environ.get('DGL_GRAPH_FORMAT', 'csc').split(',')
         formats = [f.strip() for f in formats]
+        utils.set_num_threads(num_worker_threads)
         serv = DistGraphServer(int(os.environ.get('DGL_SERVER_ID')),
                                os.environ.get('DGL_IP_CONFIG'),
                                int(os.environ.get('DGL_NUM_SERVER')),


python/dgl/distributed/dist_context.py
```diff
@@ -226,6 +226,7 @@ def initialize(ip_config, num_servers=1, num_workers=0,
             'Please define DGL_CONF_PATH to run DistGraph server'
         formats = os.environ.get('DGL_GRAPH_FORMAT', 'csc').split(',')
         formats = [f.strip() for f in formats]
+        utils.set_num_threads(num_worker_threads)
```
classicsong, line 229: Can you also update the docstring?

classicsong (commented):
How can we specify this number of threads for server?

VoVAllen (approved):

classicsong:
I double checked the initialization code of servers and trainers in the distributed training.
With this change, the users are forced to write their code to control the OMP threads. 
But why not use the OMP_NUM_THREADS env?

JingchengYu94:
> OMP_NUM_THREADS



> I double checked the initialization code of servers and trainers in the distributed training.
> With this change, the users are forced to write their code to control the OMP threads.
> But why not use the OMP_NUM_THREADS env?

I did a global search, seems like currently OMP_NUM_THREADS is only used by [MaxConcurrency](https://github.com/dmlc/dgl/blob/195f99362d883f8b6d131b70a7868a537e55b786/src/runtime/threading_backend.cc#L202), and MaxConcurrency is used to control [thread pool size](https://github.com/dmlc/dgl/blob/195f99362d883f8b6d131b70a7868a537e55b786/src/runtime/thread_pool.cc#L246), rather than omp thread size.

So seems like we cannot control server omp thread number with OMP_THREAD_NUM env currently. Are you suggesting we calling ``utils.set_num_threads(os.getenv['OMMP_THREAD_NUM'])``in dist_context.py?

classicsong:
> > OMP_NUM_THREADS
> 
> > I double checked the initialization code of servers and trainers in the distributed training.
> > With this change, the users are forced to write their code to control the OMP threads.
> > But why not use the OMP_NUM_THREADS env?
> 
> I did a global search, seems like currently OMP_NUM_THREADS is only used by [MaxConcurrency](https://github.com/dmlc/dgl/blob/195f99362d883f8b6d131b70a7868a537e55b786/src/runtime/threading_backend.cc#L202), and MaxConcurrency is used to control [thread pool size](https://github.com/dmlc/dgl/blob/195f99362d883f8b6d131b70a7868a537e55b786/src/runtime/thread_pool.cc#L246), rather than omp thread size.
> 
> So seems like we cannot control server omp thread number with OMP_THREAD_NUM env currently. Are you suggesting we calling `utils.set_num_threads(os.getenv['OMMP_THREAD_NUM'])`in dist_context.py?

You can take a look at this: https://www.ibm.com/docs/en/xcafbg/9.0.0?topic=SS3KZ4_9.0.0/com.ibm.xlf111.bg.doc/xlfopg/omp_routines.htm#ompsetnumthreads

Issue #3247, opened 2021-08-11
load_partition() fails for heterographs whose edge relations are not unique


jwyyy:
## 🐛 Bug

load_partition() fails to partition a heterogeneous graph whose edges have the same relation but are different. 

## To Reproduce

1. Create a heterograph (**Creating a Heterogeneous Graph** section): https://docs.dgl.ai/en/0.6.x/guide/graph-heterogeneous.html
```python
import dgl
import torch as th
# Create a heterograph with 3 node types and 3 edges types.
graph_data = {
   ('drug', 'interacts', 'drug'): (th.tensor([0, 1]), th.tensor([1, 2])),
   ('drug', 'interacts', 'gene'): (th.tensor([0, 1]), th.tensor([2, 3])),
   ('drug', 'treats', 'disease'): (th.tensor([1]), th.tensor([2]))
}
g = dgl.heterograph(graph_data)
```
2. Run ``` dgl.distributed.partition_graph(g, "example", out_path="output/", num_parts=2) ``` and get the following error:
```python
>>> dgl.distributed.partition_graph(g, "example", out_path="output/", num_parts=2)
Convert a graph into a bidirected graph: 0.000 seconds
Construct multi-constraint weights: 0.000 seconds
[18:35:34] /home/tiger/usr_name/dgl/src/graph/transform/metis_partition_hetero.cc:78: Partition a graph with 10 nodes and 10 edges into 2 parts and get 1 edge cuts
Metis partitioning: 0.000 seconds
Reshuffle nodes and edges: 0.001 seconds
Split the graph: 0.000 seconds
Construct subgraphs: 0.004 seconds
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/tiger/anaconda3/envs/usr_name/lib/python3.8/site-packages/dgl-0.8-py3.8-linux-x86_64.egg/dgl/distributed/partition.py", line 631, in partition_graph
    inner_etype_mask = inner_etype == g.get_etype_id(etype)
  File "/home/tiger/anaconda3/envs/usr_name/lib/python3.8/site-packages/dgl-0.8-py3.8-linux-x86_64.egg/dgl/heterograph.py", line 1261, in get_etype_id
    etid = self._etypes_invmap.get(self.to_canonical_etype(etype), None)
[...]

PR #3249, opened 2021-08-12
Using canonical etype instead of etype in distributed partition

VoVAllen:
## Description
Fix https://github.com/dmlc/dgl/issues/3247

## Checklist
- [ ] Code is well-documented


python/dgl/distributed/partition.py
```diff
@@ -42,6 +44,49 @@ def _get_part_ranges(id_ranges):
...
+    copy_d = copy.deepcopy(part_d)
+    # Iterate over d, modify on copy_d
+
+    def _recursive_convert_str_to_tuple_key(part_d, key_tuple=tuple()):
+        if isinstance(part_d, dict):
+            for k, v in part_d.items():
+                sub_d = copy_d
+                for key in key_tuple:
+                    sub_d = sub_d[key]
+                if k.startswith("tuple"):
```
BarclayII, line 58: What if someone names their edge type "tuple" (though unlikely)?

Is it possible to always encode the canonical edge types in JSON?

BarclayII: NVM the edge type keys are always replaced with canonical edge types so this is not a problem.

python/dgl/distributed/partition.py
```diff
@@ -42,6 +44,49 @@ def _get_part_ranges(id_ranges):
...
+                        sub_d.pop(k)
+                    except ValueError:
+                        print("Error parse: {}".format(k))
+                _recursive_convert_str_to_tuple_key(v, key_tuple+(k,))
+        # return copy_d
+    _recursive_convert_str_to_tuple_key(part_d)
+    return copy_d
+
+
+def _convert_tuple_key_to_str(part_d):
```
BarclayII, line 71: Please add a comment (or docstring?) that the edge type keys should always be in canonical form.

BarclayII (approved):

VoVAllen:
Found this is much more complex than I thought since handling canonical etype/etype is not designed at the beginning. Decided to postpone this feature request

sbocconi:
The problem applies also to `HeteroGraphConv` which has as input:
```
mods : dict[str, nn.Module]
        Modules associated with every edge types. 
```
How to use it in case the same relation is used between different types of nodes?

I get the same error trying to train this kind of graphs:
```
metagraph=[('tweet', 'url', 'links'), ('tweet', 'user', 'mentions'), ('tweet', 'hashtag', 'tags'), ('user', 'url', 'links'), ('user', 'user', 'mentions'), ('user', 'hashtag', 'tags'), ('user', 'tweet', 'tweets')])
```
```
dgl._ffi.base.DGLError: Edge type "links" is ambiguous. Please use canonical edge type in the form of (srctype, etype, dsttype)
```

PR #3250, opened 2021-08-13
[Feature] Update edge features (apply_edge) for heterogeneous graph 

isratnisa:
## Description
Heterogeneous graphs can have few to many relation types. To update the edge features of the different relation types, DGL invokes separate python kernels - `func: apply_edge(fn, etype=rel) ` , which implicitly calls C/CUDA kernels. The overhead of many python kernel launches causes performance degradation. We can minimize the python overhead by modifying `apply_edge` to support multiple relation type in the same python call. Example code (snippets from RGCN-hetero):
```
Old code:       
     def forward(): 
        ...
        for rel in g.canonical_etypes:
            _, etype, _ = rel
            g.apply_edges(fn.copy_u('h', 'm'), etype = rel)
            w = wdict[etype]['weight']
            g.edges[rel].data['h*w_r'] = th.matmul(g.edges[rel].data['m'], w)

Proposed code:
     def forward(): 
        ...
        g.apply_edges(fn.copy_u('h', 'm')) // moved outside of etype loop
        for rel in g.canonical_etypes:
            _, etype, _ = rel
            w = wdict[etype]['weight']
            g.edges[rel].data['h*w_r'] = th.matmul(g.edges[rel].data['m'], w)
```
This PR supports the builtin unary functions, e.g., `copy_u` and `copy_e`. Support for binary functions and UDF will be added in the future.

## Checklist
- [x] Code is well-documented


python/dgl/heterograph.py
```diff
@@ -4423,7 +4431,19 @@ def apply_edges(self, func, edges=ALL, etype=None, inplace=False):
...
-        self._set_e_repr(etid, eid, edata)
+
+        if self._graph.number_of_etypes() == 1 or etype is not None:
+            self._set_e_repr(etid, eid, edata)
+        else:
+            edata_tensor = {}
+            key = list(edata.keys())[0]
+            out_tensor_tuples = edata[key]
+            for rel in g.canonical_etypes:
+                etid = g.get_etype_id(rel)
```
BarclayII, line 4442: You can do
```python
for etid, rel in enumerate(g.canonical_etypes)
```

BarclayII (approved):

python/dgl/heterograph.py
```diff
@@ -4423,7 +4431,19 @@ def apply_edges(self, func, edges=ALL, etype=None, inplace=False):
             edata = core.invoke_gsddmm(g, func)
         else:
             edata = core.invoke_edge_udf(g, eid, etype, func)
-        self._set_e_repr(etid, eid, edata)
+
+        if self._graph.number_of_etypes() == 1 or etype is not None:
+            self._set_e_repr(etid, eid, edata)
+        else:
```
jermainewang, line 4437: I believe `_set_e_repr` supports more than one edata tensors stored in a dictionary. Why making the change?

jermainewang (changes requested):

python/dgl/heterograph.py
BarclayII: It doesn't.  The use case of feeding a dictionary in `edata` is handled by `HeteroEdgeDataView`.

python/dgl/heterograph.py
isratnisa: Done

jermainewang (approved):

PR #3252, opened 2021-08-14
Bug fix: ndim for backend tensorflow

lululxvi:
## Description
Bug fix: ndim for backend tensorflow


VoVAllen (approved):

VoVAllen:
@lululxvi Hi, I found `tensor.ndim` works on my tensorflow. May I ask what's your tf version?

lululxvi:
@VoVAllen `tensor.ndim` only works in eager mode, but not graph mode. But `len(tensor.shape)` works in both modes. For example
```python
@tf.function
def ndim(x):
    return x.ndim

x = tf.zeros((3, 4))
ndim(x)
```
We will have the error: AttributeError: 'Tensor' object has no attribute 'ndim'. But the following works:
```python
@tf.function
def ndim(x):
    return len(x.shape)

x = tf.zeros((3, 4))
ndim(x)
```

BarclayII:
Hmm as far as we know DGL is only supposed to work in eager mode.  Did you get it working in graph mode?

lululxvi:
@BarclayII No. But this is a bug if we try to make it work in graph mode.

BarclayII:
I could merge this PR per se.  However making DGL fully work in graph mode will require a lot of non-trivial changes.  For instance, currently the NN modules and autodiff sparse operators (e.g. GSpMM and GSDDMM) are all written in the eager mode way.  Do you have a plan on this?

lululxvi:
I haven't had a good plan on this yet, and still in the early exploration stage.

jermainewang:
@lululxvi Since the change is not affecting the current users of TF 2.0 and is very light, I will close this PR first. If you have a good idea on how to port DGL to graph mode, feel free to open an issue for discussion.

PR #3254, opened 2021-08-15
[Performance] Cacheline-aligned access for UnifiedTensor

davidmin7:
## Description
When performing cross-device access with UnifiedTensor, the data transfer can be more efficient by enforcing cacheline-aligned accesses. Without the alignment, memory requests become fragmented and this results in generating more number of small PCIe requests. Since each PCIe request requires a fixed size of header, generating more number of PCIe requests for the same volume of data transfer will result in dropping the efficiency of PCIe interface.

The proposed solution is quite simple. Instead of letting GPU threads to start access whatever given node feature tensor, we first check if the byte-address of the node feature is aligned with the GPU cacheline size (128B). If it's not, we slightly shift the starting index to the closest left 128B aligned position. In this case, the subsequent accesses to the node feature will be 128B-aligned.

Since now we have forcefully shifted the starting index, GPU threads can accidentally access elements of other node features. To prevent this, we add additional boundary check. Any attempts to access outside the valid boundary is ignored.

Empirically, this solution can provide 5-10% of PCIe efficiency improvement with a varying size of node feature dimension. 

## Checklist
- [ ] Code is well-documented

## Changes
- [x] Add alignment optimization for UnifiedTensor's indexSelect function.
- [x] Modify the structure of [array_index_select_uvm.cuh](https://github.com/dmlc/dgl/blob/master/src/array/cuda/uvm/array_index_select_uvm.cuh) file so that it only includes the uvm specific (or optimized) functions.


src/array/cuda/uvm/array_index_select_uvm.cuh
```diff
@@ -7,23 +7,14 @@
 #ifndef DGL_ARRAY_CUDA_ARRAY_INDEX_SELECT_UVM_CUH_
 #define DGL_ARRAY_CUDA_ARRAY_INDEX_SELECT_UVM_CUH_
 
+#define CACHE_LINE_SIZE 128
```
classicsong, line 10: Does this vary from GPU to GPU?

src/array/cuda/uvm/array_index_select_uvm.cu
```diff
@@ -48,8 +49,13 @@ NDArray IndexSelectCPUFromGPU(NDArray array, IdArray index) {
           block.y *= 2;
       }
       const dim3 grid((len+block.y-1)/block.y);
-      CUDA_KERNEL_CALL(IndexSelectMultiKernel, grid, block, 0, thr_entry->stream,
-          array_data, num_feat, idx_data, len, ret_data);
+      if (num_feat * sizeof(DType) < 2 * CACHE_LINE_SIZE) {
```
classicsong, line 52: If the data is not aligned we will use the dgl's IndexSelectMultiKernel, otherwise we will use IndexSelectMultiKernelAligned which is specialized for UnifiedTensor?

src/array/cuda/uvm/array_index_select_uvm.cuh
davidmin7: There is no difference in the cacheline size between different GPU models. Not even between the Tesla class and the gaming class (RTX).

src/array/cuda/uvm/array_index_select_uvm.cu
davidmin7: We use `IndexSelectMultiKernelAligned` for the unaligned case. The if else condition there is for the case where the node feature size is too small. If the node feature size is too small (e.g. only 1-2 GPU cacheline size), we don't have enough number of memory requests generated per feature access in the first place to attempt to coalesce the PCIe requests.

In summary:
`IndexSelectMultiKernel` -> The existing indexing kernel in DGL. This does not have the alignment fix.
`IndexSelectMultiKernelAligned` -> An UnifiedTensor specific indexing kernel which applies the alignment fix. Has a little value for the small node features and thus in such case simply fall back to `IndexSelectMultiKernel`.

src/array/cuda/uvm/array_index_select_uvm.cuh
classicsong: glad to know.

classicsong (approved):
LGTM

PR #3258, opened 2021-08-16
[perf] Remove activation cache if not required.

yzh119:
## Description
This is a simple optimization which remove activation cache if not required.
For example, in $Y = AX$, if A doesn't require gradient, we can avoid caching X, thus saving some GPU memory footprint.

## Checklist
- [x] Code is well-documented


yzh119:
Not a significant improvement, for [this script](https://github.com/dglai/dgl-0.5-benchmark/blob/master/end_to_end/full_graph/node_classification/main_dgl_arxiv_sage.py), the GPU memory usage dropped from 3912mb to 3748 mb.

BarclayII (approved):
Fair enough.

Does that mean the sub and div binary operators in the kernels are no longer necessary?

yzh119:
I think `edge_softmax` still uses those C++ kernels.

PR #3265, opened 2021-08-18
[Bugfix] Fix CUDA 11.1 crashing when number of edges is larger than number of node pairs

BarclayII:
## Description
With CUDA 11.1 the following code will crash:
```
In [8]: g = dgl.graph(([0, 0, 0, 0, 0], [1, 1, 1, 1, 1]))

In [9]: g = g.to('cuda')

In [10]: g.ndata['x'] = torch.ones(2, 5).cuda()

In [11]: g.update_all(dgl.function.copy_u('x', 'm'), dgl.function.sum('m', 'y'))
 ** On entry to cusparseCreateCsr(): dimension mismatch, nnz (5) > matrix size (4)
```

The reason is that apparently `cusparseCreateCsr` does not allow nnz to be larger than matrix size, even if duplicate entries are allowed.

I worked around it by adding a check on NNZ versus the matrix size, and falling back to general kernel implementation if NNZ is larger than matrix size.

### DIFF
--- a/src/array/cuda/spmm.cu
+++ b/src/array/cuda/spmm.cu
@@ -413,7 +413,7 @@ void CusparseCsrmm2Hetero(
  * \brief Determine whether cusparse SpMM function is applicable.
  */
 template <int bits, typename IdType>
-inline bool cusparse_available() {
+inline bool cusparse_available(bool more_nnz_than_matrix_size) {
 #if CUDART_VERSION < 11000
   if (std::is_same<IdType, int>::value)
     if (bits > 16)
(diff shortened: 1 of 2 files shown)


yzh119 (approved):
LGTM

xcwanAndy:
Hi @BarclayII , I have the `dimension mismatch, nnz (5) > matrix size (4)` issue now. 

I'm now using dgl-cu111 v0.6.1 installed with pip3. I wonder what is the version of this dgl and I want to install it.

PR #3273, opened 2021-08-19
[Feature] Support builtin binary message function for heterogenenous graph

isratnisa:
## Description
To update the node features of the different relation types of the heterogeneous graph , DGL invokes separate python kernels - `func: update_all (fn, etype=rel) `, which implicitly calls C/CUDA kernels. The overhead of many python kernel launches causes performance degradation. We minimize the python overhead by modifying `update_all` to support multiple relation type in the same python call. This PR extends PR: [https://github.com/dmlc/dgl/pull/2989] to add binary functions on top of unary functions.

## Checklist
- [x] Code is well-documented


jermainewang:
@isratnisa Please remove the WIP flag once you've finished the change.

python/dgl/core.py
```diff
@@ -194,29 +194,48 @@ def data_dict_to_tuple(graph, data_dict, op, lhs_list=None, rhs_list=None):
...
-        ``copy_lhs``, ``copy_rhs``.
-    lhs_list : list[tensor] or list[None]
-        The feature on source nodes, could be list of None if op is ``copy_rhs``.
-    rhs_list : list[tensor] or list[None]
-        The feature on edges, could be list of None if op is ``copy_lhs``.
+    func : dgl.function.BaseMessageFunction
+        Built-in message function.
+    lhs_target : 'u', 'v' or 'e'
+        Could be list of None if op is ``copy_rhs``.
+    rhs_targer : 'u', 'v' or 'e'
```
BarclayII, line 201: `lhs_target` and `rhs_target` are no longer there.  Need to change that.

python/dgl/core.py
```diff
@@ -184,7 +184,7 @@ def bucketor(data):
         return bkts
     return unique_val, bucketor
 
-def data_dict_to_tuple(graph, data_dict, op, lhs_list=None, rhs_list=None):
+def data_dict_to_list(graph, data_dict, func, target):
```
BarclayII, line 187: Any reason changing it to a list?

python/dgl/ops/sddmm.py
```diff
@@ -74,22 +74,10 @@ def gsddmm(g, op, lhs_data, rhs_data, lhs_target='u', rhs_target='v'):
...
-            lhs_data = lhs_data_dict[srctype]
-            rhs_data = rhs_data_dict[dsttype]
-            if op not in ['copy_lhs', 'copy_rhs']:
-                lhs_data, rhs_data = reshape_lhs_rhs(lhs_data, rhs_data)
-            lhs_list[src_id] = lhs_data
-            rhs_list[dst_id] = rhs_data
-        lhs_and_rhs_tuple = tuple(lhs_list + rhs_list)
-        # With max and min reducers infinity will be returned for zero degree nodes
-        return gsddmm_internal_hetero(g, op, lhs_target, rhs_target, *lhs_and_rhs_tuple)
+        # TODO (Israt): Call reshape func
```
BarclayII, line 77: What does this TODO mean?

python/dgl/sparse.py
```diff
@@ -179,15 +190,13 @@ def _gspmm(gidx, op, reduce_op, u, e):
     return v, (arg_u, arg_e)
 
 
-def _gspmm_hetero(g, op, reduce_op, u_and_e_tuple):
+def _gspmm_hetero(g, op, reduce_op, u_len, u_and_e_tuple):
     r""" Generalized Sparse Matrix Multiplication interface.
```
BarclayII, line 194: This is also getting more complicated so docstrings are needed.

python/dgl/sparse.py
```diff
@@ -337,33 +346,33 @@ def _gsddmm(gidx, op, lhs, rhs, lhs_target='u', rhs_target='v'):
     return out
 
 
-def _gsddmm_hetero(g, op, lhs_target='u', rhs_target='v', lhs_and_rhs_tuple=None):
+def _gsddmm_hetero(g, op, lhs_len, lhs_target='u', rhs_target='v', lhs_and_rhs_tuple=None):
     r""" Generalized Sampled-Dense-Dense Matrix Multiplication interface.
```
BarclayII, line 350: Same.

src/array/cuda/spmm.cu
```diff
@@ -519,7 +519,7 @@ void SpMMCsrHetero(const std::string& op, const std::string& reduce,
              const std::vector<NDArray>& out_aux,
              const std::vector<dgl_type_t>& ufeat_ntids,  // ufeat node type id
              const std::vector<dgl_type_t>& out_ntids) {  // output node type id
-  bool is_scalar_efeat = vec_efeat.size() != 0;
+  bool is_scalar_efeat = vec_efeat[0].NumElements() == vec_csr[0].indices->shape[0];
```
BarclayII, line 522: Can I use `vec_efeat[0].ndim == 1`?

src/array/kernel.cc
```diff
@@ -167,6 +166,15 @@ void SDDMMHetero(const std::string& op,
   });
 }
 
+int get_hs_id(HeteroGraphPtr graph, int target, dgl_type_t etype) {
```
BarclayII, line 169: Docstring?  The name should also be more informative as well.

Also I would recommend wrapping it in an anonymous namespace.

src/array/kernel_decl.h
```diff
@@ -96,6 +96,13 @@ void SDDMMCoo(const std::string& op,
               int lhs_target,
               int rhs_target);
 
+/*!
+ * \brief Find the src/dst/etype id based on the target.
+ */
+int get_hs_id(HeteroGraphPtr graph,
```
BarclayII, line 102: No need to include it in `kernel_decl.h`, right?  This should be only used in `kernel.cc`?

BarclayII (changes requested):

python/dgl/sparse.py
```diff
@@ -64,6 +64,17 @@ def to_dgl_nd_for_write(x):
     return nd.NULL['int64'] if x is None else F.zerocopy_to_dgl_ndarray_for_write(x)
 
 
+def get_hs_id(g, rel, target):
```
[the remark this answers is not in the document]
isratnisa: Changed it to `get_hs_tid`. Couldn't come up with a better name. Other options can be `get_lhs_rhs_type_id` or `get_target_id`.

python/dgl/backend/pytorch/sparse.py
```diff
@@ -192,45 +192,79 @@ def backward(ctx, dZ):
 class GSpMM_hetero(th.autograd.Function):
     @staticmethod
     @custom_fwd(cast_inputs=th.float16)
-    def forward(ctx, g, op, reduce_op, *feats): # feats = lhs_data + rhs_data
-        out, (argX, argY) = _gspmm_hetero(g, op, reduce_op, feats)
-        ctx.backward_cache = g, op, reduce_op
+    def forward(ctx, g, op, reduce_op, X_len, *feats): # feats = lhs_data + rhs_data
```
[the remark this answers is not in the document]
isratnisa: Added details in backend.py

python/dgl/backend/pytorch/sparse.py
```diff
@@ -192,45 +192,79 @@ def backward(ctx, dZ):
...
+        req_grad_X = tuple([X[i].requires_grad if X[i] is not None else False
+            for i in range(X_len)])
+        req_grad_Y = tuple([Y[i].requires_grad if Y[i] is not None else False
+            for i in range(len(Y))])
+
+        # checking the first relation to decide for all the relations
+        if not spmm_cache_X(op, reduce_op, req_grad_X[src_id], req_grad_Y[dst_id]):
+            X = tuple([None] * X_len) # None
+        if not spmm_cache_Y(op, reduce_op, req_grad_X[src_id], req_grad_Y[dst_id]):
+            Y = tuple([None] * len(Y))
```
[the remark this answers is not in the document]
isratnisa: Needed this for reduce func max/min. As we are not supporting those funcs now, I removed it.

python/dgl/core.py
isratnisa: Made it easier to handle unary and binary functions. The tupling is now handled in ops/spmm.py and ops/sddmm.py

python/dgl/ops/sddmm.py
isratnisa: When number of etype == 1, the reshape function is called to match the dimension of the lhs and rhs data to avoid broadcasting issue. If in practice, nodes in the same heterogeneous graph usually have the same dimensions, its not an issue. But if different nodes have different dimensions, than different etypes may need different broadcasting dims for the same node. Now sure how to handle that case.

python/dgl/sparse.py
isratnisa: Added in backend.py

src/array/cuda/spmm.cu
isratnisa: Whats the logic here?

src/array/kernel_decl.h
isratnisa: Otherwise get a compile error.

python/dgl/backend/pytorch/sparse.py
```diff
@@ -315,18 +349,86 @@ def backward(ctx, dZ):
 class GSDDMM_hetero(th.autograd.Function):
     @staticmethod
     @custom_fwd(cast_inputs=th.float16)
-    def forward(ctx, g, op, lhs_target, rhs_target, *feats): # feats = X+Y
-        out = _gsddmm_hetero(g, op, lhs_target, rhs_target, feats)
-        ctx.backward_cache = g, op, lhs_target, rhs_target
+    def forward(ctx, g, op, X_len, lhs_target, rhs_target, *feats): # feats = X+Y
```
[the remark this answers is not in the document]
isratnisa: Added in backend.py

src/array/cuda/spmm.cu
BarclayII: Never mind.  `vec_efeat[0].NumElements()` equals `vec_csr[0].indices->shape[0]` means that the number of elements in `vec_efeat[0]` is equal to the number of edges.  This will only happen if the rest of the dimensions are all 1 which doesn't necessarily mean `vec_efeat[0]` is 1-dimensional.

python/dgl/sparse.py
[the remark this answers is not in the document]
BarclayII: `get_typeid_by_target`?

src/array/kernel_decl.h
isratnisa: True! Also move the docstring in kernel.cc?

BarclayII: Sure.

python/dgl/ops/sddmm.py
BarclayII: I see.  Could you put the elaboration as comments just in case we (any of us) forget about what we should do?

BarclayII (approved):
LGTM.  Please proceed after adding the TODO comment.  Thanks a lot!

python/dgl/sparse.py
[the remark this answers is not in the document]
isratnisa: Done!

python/dgl/ops/sddmm.py
isratnisa: Done.

PR #3293, opened 2021-08-24
[Bugfix] Add UVM specialized IndexSelect kernels which perform boundary checks

davidmin7:
## Description
Add UVM specialized `IndexSelect()` kernels which does boundary checks. Normally `IndexSelect()` kernels are not directly called from the user code so this was not an issue, but the UnifiedTensor allows it so add UVM specialized `IndexSelect()` kernels which do the boundary checks.

## Checklist
- [ ] Code is well-documented

## Changes
- [] Add boundary check for IndexSelect cuda functions


jermainewang:
I feel the index select kernels are quite similar to the one in [`array_index_select.cuh`](https://github.com/dmlc/dgl/blob/d3f2ab59be2b030928d7597317835c006cff9397/src/array/cuda/array_index_select.cuh). Can we extend the old array_index select kernels to include stride and feat len? Also, I don't know how costly are boundary checks.

davidmin7:
I've been thinking about this for myself as well quite long, but I think it is better keep them separate, especially for the concerns like you mentioned. The domains where the two versions of kernels in `array_index_select.cuh` and  `array_index_select_uvm.cuh` are used are simply too different.

For the UVM version, since it is mainly targeted for external memory accesses over slow interconnects and thus the addition of more conditional statements has no visible impact in performance, but this might not be true for the internal memory accesses. If we simply merge them into a single kernel, the requirement caused by UnifiedTensor can affect the original kernel's operations. So if the original kernel is happy with its current implementation, I think it is better to keep them separate.

davidmin7:
Btw here is the reference to the PyTorch's CUDA IndexKernel boundary checking: https://github.com/pytorch/pytorch/blob/479fc4e41250a1710e34cc1dcaa6272c89832fb4/aten/src/ATen/native/cuda/IndexKernel.cu#L85-L101

classicsong:
> I've been thinking about this for myself as well quite long, but I think it is better keep them separate, especially for the concerns like you mentioned. The domains where the two versions of kernels in `array_index_select.cuh` and `array_index_select_uvm.cuh` are used are simply too different.
> 
> For the UVM version, since it is mainly targeted for external memory accesses over slow interconnects and thus the addition of more conditional statements has no visible impact in performance, but this might not be true for the internal memory accesses. If we simply merge them into a single kernel, the requirement caused by UnifiedTensor can affect the original kernel's operations. So if the original kernel is happy with its current implementation, I think it is better to keep them separate.

I think this is a valid concern. We can keep them separately.

classicsong (approved):
Generally, the PR LGTM.
Can you add some comments for IndexSelectSingleKernel and IndexSelectMultiKernelAligned in src/array/cuda/uvm/array_index_select_uvm.cuh to clarify in which case you can use IndexSelectMultiKernelAligned? 

@jermainewang Do we need to further discuss on this design choice?

PR #3279, opened 2021-08-20
[Feature] DistDataLoaderWrapper copies features for DistDataLoader in Node/EdgeDataLoader

jwyyy:
## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage
- [x] Code is well-documented
- [x] To the best of my knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [x] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

## Changes
This PR is related to https://github.com/dmlc/dgl/issues/3278.

Added a wrapper to `DistDataLoader` in `NodeDataLoader` and `EdgeDataLoader` to copy features for sampled blocks.

PR #3323, opened 2021-09-04
[Feature] Add DistTensor reference to blocks returned by DataLoader

jwyyy:
## Description
This is a new implementation of https://github.com/dmlc/dgl/pull/3279.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/dataloading/pytorch/dataloader.py
+++ b/python/dgl/dataloading/pytorch/dataloader.py
@@ -539,12 +539,13 @@ class NodeDataLoader:
                 device = 'cpu'
             assert device == 'cpu', 'Only cpu is supported in the case of a DistGraph.'
             # Distributed DataLoader currently does not support heterogeneous graphs
-            # and does not copy features.  Fallback to normal solution
+            # Add a wrapper to Distributed DataLoader to copy features
             self.collator = NodeCollator(g, nids, block_sampler, **collator_kwargs)
             _remove_kwargs_dist(dataloader_kwargs)
-            self.dataloader = DistDataLoader(self.collator.dataset,
-                                             collate_fn=self.collator.collate,
-                                             **dataloader_kwargs)
+            self.dataloader = _DistDataLoaderWrapper(g,
+                                                     DistDataLoader(self.collator.dataset,
+                                                                    collate_fn=self.collator.collate,
+                                                                    **dataloader_kwargs))
             self.is_distributed = True
         else:
             if device is None:
@@ -573,7 +574,8 @@ class NodeDataLoader:
     def __iter__(self):
         """Return the iterator of the data loader."""
         if self.is_distributed:
-            # Directly use the iterator of DistDataLoader, which doesn't copy features anyway.
+            # Directly use the iterator of DistDataLoader
+            # Wrapped DistDataLoader works like DistDataLoader and also copy features
             return iter(self.dataloader)
         else:
             return _NodeDataLoaderIter(self)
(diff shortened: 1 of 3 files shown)


python/dgl/dataloading/pytorch/dataloader.py
```diff
@@ -325,6 +325,51 @@ def __next__(self):
...
+                block_nodes.data[ndata_name] = (g_nodes.data[ndata_name], nids)
+        # copy edge feature DistTensor reference
+        for etype in block.etypes:
+            eids = block.edges[etype].data[EID]
+            block_edges = block.edges[etype]
+            g_edges = self.g.edges[etype]
+            for name in self.g._get_edata_names(etype):
+                edata_name = name.get_name()
+                block_edges.data[edata_name] = (g_edges.data[edata_name], eids)
+        return block
```
jermainewang, line 371: You may find some internal utility functions like [`extract_node_subframes_for_block`](https://github.com/dmlc/dgl/blob/master/python/dgl/utils/internal.py#L793) designed for similar purposes. Ideally, you don't need to change the `__init__` function of a `Column`. The `subframe` method will save the index array for you.

python/dgl/frame.py
```diff
@@ -120,9 +121,14 @@ class Column(object):
...
-        self.scheme = scheme if scheme else infer_scheme(storage)
-        self.index = index
+        if isinstance(storage, tuple):
+            # handle DistTensor and node/edge ids in subgraphs
+            self.storage = storage[0]
+            self.index = storage[1]
+        else:
+            self.storage = storage
+            self.index = index
+        self.scheme = scheme if scheme else infer_scheme(self.storage)
```
jermainewang, line 131: Try to avoid changing the `__init__` function as I suggested earlier.

python/dgl/frame.py
```diff
@@ -163,8 +184,18 @@ def data(self):
...
-        self.storage = val
+        if isinstance(val, tuple):
+            # This case is used to handle DistTensor returned by DistGraph.
+            # index is used to identify rows in DistTensor
+            # that are present in the current graph/block
+            data, index = val
+            assert isinstance(data, distributed.dist_tensor.DistTensor)\
+                , "Input data must be DistTensor when index is specified."
+            self.index = index
+            self.storage = data
```
jermainewang, line 195: This part should also be gone. The `val` should just be a `DistTensor` instead of a pair.

python/dgl/frame.py
```diff
@@ -181,6 +212,9 @@ def to(self, device, **kwargs): # pylint: disable=invalid-name
         Column
             A new column
         """
+        if isinstance(self.storage, distributed.dist_tensor.DistTensor):
+            # copy actual tensor from DistTensor, if self.storage is a DistTensor
+            self._copy_dist_tensor()
```
jermainewang, line 217: Is this needed? `Column.data` will call `_copy_dist_tensor` anyway.

jermainewang (changes requested):

python/dgl/dataloading/pytorch/dataloader.py
jwyyy: I revised `_copy_features()` but didn't use `extract_node_subframes_for_block()` directly, since `DistGraph` doesn't have [self._node_frames](https://github.com/dmlc/dgl/blob/master/python/dgl/utils/internal.py#L820). ( For `DistGraph`, it is `self._ndata_store`.) So `DistTensor` is copied to blocks first and then block is used as `DGLHeteroGraph`. I am not sure if it is a good idea to write a new function like `extract_node_subframes_for_block()` but for DistGraph:
```
node_frames = []
for i, ind_nodes in enumerate(srcnodes):
    data = {}
    ntype = dist_g.ntypes[i]
    for name in dist_g._get_ndata_names(ntype):
        data[name.get_name()] = dist_g.nodes[ntype].data[name.get_name()]
    nframe = Frame(data, ind_nodes)
    subf = nframe.subframe(ind_nodes)
    node_frames.append(subf)
```
With this, `_copy_features()` can be simplified. Which approach seems more appropriate to you? Do you have any better ideas?

python/dgl/frame.py
jwyyy: Removed this part and rolled back to the original implementation.

python/dgl/frame.py
jwyyy: Removed this part and rolled back to the original implementation.

python/dgl/frame.py
jwyyy: I considered a case where users can call `to()` without accessing attributes. [Column.clone()](https://github.com/dmlc/dgl/blob/master/python/dgl/frame.py#L257) directly copies `self.storage` to a new `Column` but not `self.data`.

PR #3333, opened 2021-09-08
[Bugfix]  Fix #3291

### LINKED ISSUE #3291: NCCL based SparseEmbedding fails for pytorch < 1.7, due to missing `delete_key` method in TCPStore
## 🐛 Bug

NCCL based SparseEmbedding fails for pytorch < 1.7, due to missing `delete_key` method in TCPStore:
```
AttributeError: ‘torch.distributed.TCPStore’ object has not attribute ‘delete_key’
```

## To Reproduce

1. Install DGL 0.7 and pytorch 1.6.
1. Run the `/examples/pytorch/rgcn/entity_classify_mp.py` example using a GPU and the `--dgl-sparse` flag.
1. This will result in:
```
AttributeError: ‘torch.distributed.TCPStore’ object has not attribute ‘delete_key’
```

## Environment

 - DGL Version (e.g., 1.0): 0.7-cu111 from pip
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): pytorch 1.6
 - OS (e.g., Linux): ubuntu
 - Python version: 3.6

classicsong:
## Description
FIx #3291

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/optim/pytorch/sparse_optim.py
+++ b/python/dgl/optim/pytorch/sparse_optim.py
@@ -51,7 +51,7 @@ class SparseGradOptimizer(abc.ABC):
                     'MultiGPU world_size for each embedding should be same.'
         assert not self._rank is None
         assert not self._world_size is None
-        self._nccl_root_id = 'nccl_root_id''
+        self._nccl_root_id = 'SparseGradOptimizer.nccl_root_id''
 
     def step(self):
         ''' The step function.
--- a/python/dgl/optim/pytorch/sparse_optim.py
+++ b/python/dgl/optim/pytorch/sparse_optim.py
@@ -51,7 +51,7 @@ class SparseGradOptimizer(abc.ABC):
                     'MultiGPU world_size for each embedding should be same.'
         assert not self._rank is None
         assert not self._world_size is None
-        self._nccl_root_id = 'SparseGradOptimizer.nccl_root_id''
+        self._nccl_root_id = 'SparseGradOptimizer.nccl_root_id'
 
     def step(self):
         ''' The step function.
--- a/python/dgl/optim/pytorch/sparse_optim.py
+++ b/python/dgl/optim/pytorch/sparse_optim.py
@@ -35,6 +35,8 @@ class SparseGradOptimizer(abc.ABC):
         # otherwise it will crash the training
         self.shmem_buffer_holder = []
 
+        assert len(params) > 0, 'Empty parameters'
+        name = "opt"
         # if we are using shared memory for communication
         for emb in params:
             assert isinstance(emb, NodeEmbedding), \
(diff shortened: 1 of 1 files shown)


python/dgl/optim/pytorch/sparse_optim.py
```diff
@@ -48,6 +50,9 @@ def __init__(self, params, lr):
                     'MultiGPU rank for each embedding should be same.'
                 assert self._world_size == emb.world_size, \
                     'MultiGPU world_size for each embedding should be same.'
+            name = "{}_{}".format(name, emb.name)
+        self._name = name
+        self._short_name = name[:len(name) if len(name) < 32 else 32]
```
nv-dlasalle, line 55: Do we gaurantee/require that embeddings have unique names? If not, would a solution be to store a global communicator, and only build a new one if one has not yet been established, thus ensuring we only insert the key once? Essentially how it's done here: https://github.com/dmlc/dgl/blob/0.7.1/python/dgl/nn/pytorch/sparse_emb.py#L117
(Not sure why I didn't do that here previously)

classicsong: Yeah. For DistTensor upon which the distributed.NodeEmbedding is built, the name should be unique.

If a user creates multiple sparse optimizer and we use a global communicator here, is there any racing problems?

python/dgl/optim/pytorch/sparse_optim.py
```diff
@@ -119,9 +120,6 @@ def _comm_setup(self):
                 self._comm = nccl.Communicator(self._world_size,
                                                self._rank,
                                                nccl_id)
-                if self._rank == 0:
-                    # clear the store entry for future communicators
-                    store.delete_key('nccl_root_id')
```
nv-dlasalle, line 124: If we won't delete the key, I think we need to be something less generic than `nccl_root_id`. Maybe `SparseGradOptimizer.nccl_root_id`

classicsong: You mean, initialize it at the beginning? Make sense.

python/dgl/optim/pytorch/sparse_optim.py
```diff
@@ -50,6 +51,7 @@ def __init__(self, params, lr):
                     'MultiGPU world_size for each embedding should be same.'
         assert not self._rank is None
         assert not self._world_size is None
+        self._nccl_root_id = 'nccl_root_id''
```
nv-dlasalle, line 54: I meant this string should be more unique:
```
self._nccl_root_id = 'SparseGradOptimizer.nccl_root_id'
```

python/dgl/optim/pytorch/sparse_optim.py
classicsong: `SparseGradOptimizer.nccl_root_id` sounds good

nv-dlasalle (approved):

PR #3334, opened 2021-09-08
[Feature] enable create/set/free cuda stream for internal use

Rhett-Ying:
## Description

In order to transfer data(tensor, graph) from cpu to gpu asynchronously, we need to trigger such transfer on non-default stream. As DGL has stream-related calls already, I'd like to expose them into python level and wrap them for easier use. For now, stream/context is not fully supported/designed in DGL, so I'd like to use the wrappers internally for the moment. we could improve it in near future if needed.
## Checklist
- [x] Code is well-documented


src/runtime/c_runtime_api.cc
```diff
@@ -356,6 +356,15 @@ int DGLSetStream(int device_type, int device_id, DGLStreamHandle stream) {
   API_END();
 }
 
+int DGLGetStream(int device_type, int device_id, DGLStreamHandle* stream) {
+  API_BEGIN();
+  DGLContext ctx;
+  ctx.device_type = static_cast<DLDeviceType>(device_type);
+  ctx.device_id = device_id;
+  *stream = DeviceAPIManager::Get(ctx)->GetStream();
```
BarclayII, line 364: There's no return code.

Also would GetStream() always succeed?  Better to have a sanity check here and return -1 if it fails.

tests/pytorch/test_stream.py
```diff
@@ -0,0 +1,39 @@
...
+    s.synchronize()
+    OPS.copy_u_sum(gg, xx)
+
+    # launch on new stream created via torch.cuda
+    s = torch.cuda.Stream(device=F.ctx())
+    with torch.cuda.stream(s):
+        xx = x.to(device=F.ctx(), non_blocking=True)
+    with FS.stream(s):
+        gg = g.to(device=F.ctx())
+    s.synchronize()
```
BarclayII, line 29: Is it possible to check if the stream is indeed non-default in unit tests?

BarclayII (approved):
LGTM.  Some minor comments.

src/runtime/c_runtime_api.cc
Rhett-Ying: `API_END()` includes exception catch and the return code.
`GetStream()` just return a data member in a thread_local which is nullptr in default. so I think it always succeeds.

tests/pytorch/test_stream.py
Rhett-Ying: If you mean the created stream `s`, s.cuda_stream is nullptr if default, otherwise non-nullptr.
If you mean g.to() is really running on the stream we set, the most reliable way is printing something when doing GPUCopy which is not easy or elegant to implement. the following OPS.copy_u_sum() I called is an indirect way to check g.to() running on the stream we set, though it's not very reliable I think.

which one do u mean here?

PR #3335, opened 2021-09-08
[RPC] Use tensorpipe for rpc communication

VoVAllen:
## Description
Refactor the RPC communication layer by using tensorpipe
Core changes:
- Communicator no longer works at the Message (iovec with buffer and size) but on the RPCMessage level
- Now the incoming RPCMessage from different target will be put in one queue, which should resolve the imbalance issue mentioned and fixed before
- By default we will use 3 sockets concurrently to send/receive the message (the mpt channel in the code). Tensorpipe also supports infiniband(ibv) and in future supports efa. However I didn't add it here since I cannot test ibv. 

This is only the first refactor of the rpc system. Later I'll try making all the rpc call return a future directly.

## Checklist
- [ ] Code is well-documented


VoVAllen:
Running distributed training on `dgl/examples/pytorch/graphsage/experimental/train_dist.py` on m5n.24xlarge, with 2 trainer, 2 sampler, 2 server per machine, with OMP_NUM_THREADS 24. 
For per epoch training time, got 2.80 (this branch with tensorpipe) vs 2.87 (current master). Didn't see significant changes

VoVAllen:
With 3 multiplex socket, per epoch time becomes 2.77

VoVAllen:
2.63s per epoch with EFA

src/rpc/rpc.cc
```diff
@@ -3,323 +3,294 @@
...
-    if (ptr.size == 0) {
-      LOG(FATAL) << "Cannot send a empty NDArray.";
-    }
-    ndarray_data_msg.size = ptr.size;
-    NDArray tensor = ptr.tensor;
-    ndarray_data_msg.deallocator = [tensor](network::Message*) {};
-    CHECK_EQ(RPCContext::ThreadLocal()->sender->Send(
-      ndarray_data_msg, target_id), ADD_SUCCESS);
-  }
+  RPCContext::ThreadLocal()->sender->Send(msg, target_id);
```
zheng-da, line 33: with this, can we send messages in multiple threads?

src/rpc/rpc.cc
```diff
@@ -358,128 +329,128 @@ DGL_REGISTER_GLOBAL("distributed.server_state._CAPI_DGLRPCGetServerState")
...
-  std::vector<int64_t> local_data_shape;
-  std::vector<std::vector<dgl_id_t> > remote_ids(machine_count);
-  std::vector<std::vector<dgl_id_t> > remote_ids_original(machine_count);
-  // Get row size (in bytes)
-  int row_size = 1;
-  for (int i = 0; i < local_data->ndim; ++i) {
-    local_data_shape.push_back(local_data->shape[i]);
-    if (i != 0) {
-      row_size *= local_data->shape[i];
+  .set_body([](DGLArgs args, DGLRetValue* rv) {
```
zheng-da, line 350: could you please not change the indent? it's very difficult for me to see what you changed.

src/rpc/rpc.h
```diff
@@ -12,11 +12,13 @@
...
+#include <deque>
 #include <vector>
 #include <string>
-#include "./network/communicator.h"
-#include "./network/socket_communicator.h"
-#include "./network/msg_queue.h"
+#include "./tensorpipe/tp_communicator.h"
+// #include "./network/communicator.h"
+// #include "./network/socket_communicator.h"
+// #include "./network/msg_queue.h"
```
zheng-da, line 21: remove

src/rpc/tensorpipe/tp_communicator.h
```diff
@@ -0,0 +1,186 @@
...
+   * \brief pipe for each client connections
+   */
+  std::unordered_map<int /* Sender (virutal) ID */,
+                     std::shared_ptr<tensorpipe::Pipe>>
+    pipes_;
+
+  /*!
+   * \brief RPCMessage queue
+   */
+  std::shared_ptr<RPCMessageQueue> queue_;
```
zheng-da, line 180: i try to understand how send and receive work after using TensorPipe. Do we still need to have a thread dedicated to waiting for messages? I see receiver has a method called Wait.

src/rpc/rpc.cc
VoVAllen: fixed. It's a mistake

VoVAllen:
Explanation on the current implementation:
There is a dedicated thread managed by libuv using epoll for the socket communication (for both send and recv). Tensorpipe registered a callback when there's incoming/outcoming messages. The thread is not busy waiting, and will be only waken up when there's event happening, by calling the registered callback. 
In our rpc communication, we turn Tensorpipe message into our own RPCMessage and push it into the threadsafe queue in the read callback function. 

For multithread communication, I can see larger scope refactoring is needed. Including message id incremental, RPCContext and other things. My current thinking is refactoring by the following step:
- Provide new API to make rpc call using futures and abandon the message queue. Each message will contain an id for future and set the result to future with the message returned from server
- Refactor all API using this interface
- Make everything multithread-able

Do you have any suggestions? @zheng-da

src/rpc/rpc.cc
VoVAllen: You're right. Cannot do this now since the context is `ThreadLocal`.

src/rpc/tensorpipe/tp_communicator.cc
```diff
@@ -0,0 +1,154 @@
...
+using namespace tensorpipe;
+
+void TPSender::AddReceiver(const std::string& addr, int recv_id) {
+  receiver_addrs_[recv_id] = addr;
+}
+
+bool TPSender::Connect() {
+  for (auto kv : receiver_addrs_) {
+    std::shared_ptr<Pipe> pipe;
+    for (;;) {
```
Rhett-Ying, line 30: endless retrying if no connection could be built? Is it on-purpose? or could we skip unavailable ones after retrying several times?

src/rpc/tensorpipe/tp_communicator.cc
```diff
@@ -0,0 +1,154 @@
...
+namespace rpc {
+
+using namespace tensorpipe;
+
+void TPSender::AddReceiver(const std::string& addr, int recv_id) {
+  receiver_addrs_[recv_id] = addr;
+}
+
+bool TPSender::Connect() {
+  for (auto kv : receiver_addrs_) {
```
Rhett-Ying, line 28: const auto& ?

src/rpc/tensorpipe/tp_communicator.cc
```diff
@@ -0,0 +1,154 @@
...
+        sleep(5);
+        LOG(INFO) << "Cannot connect to remove server. Wait to retry";
+      }
+    }
+    pipes_[kv.first] = pipe;
+  }
+  return true;
+}
+
+void TPSender::Send(RPCMessage msg, int recv_id) {
```
Rhett-Ying, line 54: const RPCMessage&  or &&?

src/rpc/tensorpipe/tp_communicator.h
```diff
@@ -0,0 +1,186 @@
+/*!
```
Rhett-Ying, line 1: do we have any unit tests for these files? If have some, I think it may help new fresh like me to understand how these classes are used.

Rhett-Ying (commented):
I'm just curious and trying to learn/follow this change. pls feel free to ignore incorrect/trivial comments I dropped.

src/rpc/tensorpipe/queue.h
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
```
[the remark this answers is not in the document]
VoVAllen: I borrowed it from `pytorch/tensorpipe`, which used BSD license. According BSD license, I should keep the license here. However, I believe the code is simple enough that it doesn't really matter

src/rpc/tensorpipe/queue.h
```diff
@@ -0,0 +1,53 @@
...
+namespace dgl {
+namespace rpc {
+
+template <typename T>
+class Queue {
+ public:
+  // Capacity isn't used actually
+  explicit Queue(int capacity = 1) : capacity_(capacity) {}
+
+  void push(T t) {
```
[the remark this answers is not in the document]
VoVAllen: Is it necessary to declare T& or T&& if I push the element with `std::move`? I'm not so familiar at this part.

src/rpc/tensorpipe/tp_communicator.cc
VoVAllen: Current distributed design requires all clients connecting to server at the beginning. Therefore if some are unavailable, the whole program just steps into the failure stage

src/rpc/rpc.h
```diff
@@ -74,12 +76,17 @@ struct RPCContext {
...
+  std::shared_ptr<TPSender> sender;
 
   /*!
    * \brief Receiver communicator.
    */
-  std::shared_ptr<network::Receiver> receiver;
+  std::shared_ptr<TPReceiver> receiver;
+
+  std::shared_ptr<tensorpipe::Context> ctx;
+  // std::shared_ptr<te>
```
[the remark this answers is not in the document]
VoVAllen: fixed

src/rpc/rpc.h
VoVAllen: fixed

src/rpc/tensorpipe/tp_communicator.cc
VoVAllen: fixed

src/rpc/tensorpipe/tp_communicator.h
VoVAllen: Yes. Wait is at the connection stage, which is corresponding to sender's `Connect` function. It's not for waiting a RPC Message

src/rpc/rpc.cc
VoVAllen: Updated. Now it supports multithread with unit test

src/rpc/tensorpipe/queue.h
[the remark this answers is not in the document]
Rhett-Ying: I think it's not necessary, but be preferable. If const T& and T&& interfaces are provided, we could support any kind calls of push() and remove any unnecessary constructors of T. Though push(T t) is mighty as well, one copy constructor is required always. I think pass by value is required only when we want a copy and modify it in function body.

src/rpc/tensorpipe/tp_communicator.cc
Rhett-Ying: so some logic exist outside to detect and step into failure stage if some connections are unavailable? timeout?

VoVAllen: Currently the answer is no.

src/rpc/rpc.h
```diff
@@ -92,74 +101,35 @@ struct RPCContext {
    */
   std::shared_ptr<ServerState> server_state;
 
-  /*! \brief Get the thread-local RPC context structure */
-  static RPCContext *ThreadLocal() {
-    return dmlc::ThreadLocalStore<RPCContext>::Get();
+  /*! \brief Get the RPC context singleton */
+  static RPCContext* getInstance() {
```
Rhett-Ying, line 105: do we need really singleton? if yes, we could use below one which is thread-safe since C++11.
static RPCContext& getInstance() {static RPCContext ctx; return ctx; }

VoVAllen: Fixed

VoVAllen:
Update: Tensors are transferred through Channels and payloads are transferred through Transport. Now all NDArray are set as tensor to use channel to transfer

src/rpc/rpc.cc
```diff
@@ -3,257 +3,266 @@
...
-    buffer_list[i] = ndarray_data_msg.data;
-  }
-  StreamWithBuffer zc_read_strm(rpc_meta_msg.data, rpc_meta_msg.size-sizeof(int32_t), buffer_list);
-  zc_read_strm.Read(msg);
-  rpc_meta_msg.deallocator(&rpc_meta_msg);
+  RPCContext::getInstance()->receiver->Recv(msg);
   return kRPCSuccess;
 }
 
+void InitGlobalTpContext() {
```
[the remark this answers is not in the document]
VoVAllen: It should be initialized once. However based on current DGL logic, it should be only called in the main thread. Therefore we won't meet concurrency problem here

src/rpc/rpc.cc
```diff
@@ -3,257 +3,266 @@
...
+    std::vector<std::shared_ptr<tensorpipe::transport::Listener>> listeners;
+    for (int i = 0; i < kNumUvThreads; i++) {
+      auto context = tensorpipe::transport::uv::create();
+      std::string address = guessAddress();
+      contexts.push_back(std::move(context));
+      listeners.push_back(contexts.back()->listen(address));
+    }
+    auto mptChannel = tensorpipe::channel::mpt::create(std::move(contexts),
+                                                       std::move(listeners));
+    context->registerChannel(10, "mpt", mptChannel);
```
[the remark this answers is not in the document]
VoVAllen: added comment here. Just any number larger than the basic channel

src/rpc/tensorpipe/tp_communicator.cc
```diff
@@ -0,0 +1,162 @@
...
+    cpu_buffer.ptr = ptr.data;
+    tp_msg.tensors[i].buffer = cpu_buffer;
+    tp_msg.tensors[i].length = ptr.size;
+    if (ptr.size == 0) {
+      LOG(FATAL) << "Cannot send a empty NDArray.";
+    }
+  }
+  pipe->write(tp_msg, [ndarray_holder](const tensorpipe::Error& error) {
+    if (error) {
+      LOG(FATAL) << error.what();
```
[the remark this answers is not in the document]
VoVAllen: Added receiver id for debugging

src/rpc/tensorpipe/tp_communicator.cc
```diff
@@ -0,0 +1,162 @@
...
+    }
+  });
+}
+
+void TPSender::Finalize() {}
+void TPReceiver::Finalize() {}
+
+bool TPReceiver::Wait(const std::string& addr, int num_sender) {
+  std::string addr_str(addr);
```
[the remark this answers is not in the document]
VoVAllen: Removed

src/rpc/tensorpipe/tp_communicator.cc
```diff
@@ -0,0 +1,162 @@
...
+    CHECK(checkConnect.get_future().get()) << "Invalid connect message.";
+    pipes_[i] = pipe;
+    ReceiveFromPipe(pipe, queue_);
+  }
+  return true;
+}
+
+void TPReceiver::ReceiveFromPipe(std::shared_ptr<Pipe> pipe,
+                                 std::shared_ptr<RPCMessageQueue> queue) {
+  pipe->readDescriptor([pipe, queue = std::move(queue)](const Error& error,
```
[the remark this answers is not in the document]
VoVAllen: I added C++14 support in cmake. CUDA 11 requires C++14 and we already turned it on for cuda compilation.

src/rpc/tensorpipe/tp_communicator.cc
```diff
@@ -0,0 +1,162 @@
...
+    }
+    Allocation allocation;
+    CHECK_EQ(descriptor.payloads.size(), 0) << "Invalid DGL RPC Message";
+
+    int tensorsize = descriptor.tensors.size();
+    if (tensorsize > 0) {
+      allocation.tensors.resize(tensorsize);
+      for (int i = 0; i < descriptor.tensors.size(); i++) {
+        tensorpipe::CpuBuffer cpu_buffer;
+        cpu_buffer.ptr = new char[descriptor.tensors[i].length];
```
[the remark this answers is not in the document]
VoVAllen: This will be released at https://github.com/dmlc/dgl/blob/c8b18b793c5e8c1d3cd81060b6f6c1540a52c127/src/graph/serialize/zerocopy_serializer.cc#L24. Basically the CpuBuffer here will be used to construct a NDArray, which owns the underlying pointer. Therefore it will free the memory once destructed.

Rhett-Ying (approved):

src/rpc/tensorpipe/tp_communicator.cc
```diff
@@ -0,0 +1,162 @@
...
+  auto& buffer_list = zc_write_strm.buffer_list();
+  for (int i = 0; i < buffer_list.size(); i++) {
+    auto& ptr = buffer_list[i];
+    (*ndarray_holder.get())[i] = ptr.tensor;
+    tensorpipe::CpuBuffer cpu_buffer;
+    cpu_buffer.ptr = ptr.data;
+    tp_msg.tensors[i].buffer = cpu_buffer;
+    tp_msg.tensors[i].length = ptr.size;
+    if (ptr.size == 0) {
+      LOG(FATAL) << "Cannot send a empty NDArray.";
```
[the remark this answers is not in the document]
VoVAllen: This should not happen since the empty NDArray should be skipped in the previous steps (when serialize to the RPCMessage)

VoVAllen:
Update: I tested on the ogb-paper100M dataset, and found the end2end time is significantly slower. However the sample+data copy part is faster about 10%. But the backward part is much slower, which I guess is due to kind of imbalance issues. That some workers are slower than others, thus they all stopped at the synchronous point

zheng-da:
> Update: I tested on the ogb-paper100M dataset, and found the end2end time is significantly slower. However the sample+data copy part is faster about 10%. But the backward part is much slower, which I guess is due to kind of imbalance issues. That some workers are slower than others, thus they all stopped at the synchronous point

do you think if we can co-exist the current communication backend and tensorpipe? so that we won't have dramatic changes.

VoVAllen:
@zheng-da The problem is now fixed by using single socket. It's viable but needs extra effort.

Issue #3318, opened 2021-09-03
`farthest_point_sampler` return wrong idx when gpu is set > 0


sangyx:
## 🐛 Bug

`farthest_point_sampler` is a sampling algorithm, which returns the index of the nodes that are selected. However, if the data is in the gpu > 0, it will not run as expected. In fact, it will return an all zero tensor.

## To Reproduce
Run the code:
```python
import torch
from dgl.geometry import farthest_point_sampler

x = torch.randn((2, 10, 3))
point_idx1 = farthest_point_sampler(x, 3)
print(point_idx1)

point_idx2 = farthest_point_sampler(x.to("cuda:0"), 3)
print(point_idx2)

point_idx3 = farthest_point_sampler(x.to("cuda:1"), 3)
print(point_idx3)
```

The output is:
```
Using backend: pytorch
tensor([[0, 1, 5],
        [5, 3, 6]])
tensor([[5, 8, 1],
        [1, 5, 8]], device='cuda:0')
tensor([[0, 0, 0],
        [0, 0, 0]], device='cuda:1')
```

## Expected behavior
A correct return should be like `point_idx1` and `point_idx2`.

## Environment

 - DGL Version (e.g., 1.0):0.7.1
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):PyTorch 1.9.0
 - OS (e.g., Linux): Ubuntu
 - How you installed DGL (`conda`, `pip`, source): conda
 - Python version: 3.9
 - GPU models and configuration (e.g. V100): Tesla T4

PR #3341, opened 2021-09-10
[Bugfix] Fix SetDevice issue for NeighborMatching

lygztq:
## Description
Fix issues similar to #3318

## Checklist
- [ ] Code is well-documented


mufeili:
@hetong007 Are you going to review this PR?

hetong007:
@BarclayII Do you think this fix is fine, or suggest we wait for a unified solution for https://github.com/dmlc/dgl/issues/3342?

lygztq:
Also, I think `CUDAThreadEntry` is bound with threads currently and multiple devices will share the same entry. This may cause error if multiple devices use the same handle in the `CUDAThreadEntry`. I will have another fix for this problem here, providing we have a multi-device version of `CUDAThreadEntry`.

BarclayII:
Somehow I missed this PR.  This is still a temporary fix I guess?  I would probably implement a device guard for all kernels that involve CUDA input tensors and graphs.  This will require a sweep of existing CUDA code though.

Also @jermainewang we should not miss fixing #3342 for 1.0.

lygztq:
Another problem is that some handles in `CUDAThreadEntry`, such as `curandGenerator_t `, have device-side resources. I think we may need separate `CUDAThreadEntry` for each device. Current shared `CUDAThreadEntry` will cause error.

BarclayII:
> Another problem is that some handles in `CUDAThreadEntry`, such as `curandGenerator_t `, have device-side resources. I think we may need separate `CUDAThreadEntry` for each device. Current shared `CUDAThreadEntry` will cause error.

So the device guard should switch not only the device but also use the corresponding curand generator?  I assume that also applies to other handles (like cusparse handles)?

@nv-dlasalle what's your opinion on the device guard idea?  I would imagine that we will need to call `GetDevice` and `SetDevice` for all kernels that receive GPU tensors from outside.  Would it degrade performance?

nv-dlasalle:
I think the current model of having GPU resources per CPU thread is probably not correct. For the most part, DGL's threading model is multiple threads per training process, which means all threads would share the same GPU. We should probably move all of these GPU resources to the DLContext object (or a subclass specifically for CUDA contexts).

Do we know how the device gets set to the wrong device in this case? I'm thinking there may be some larger bug that we have one process/thread associated with multiple devices. In general I think we need a good holistic approach to managing GPU resources, as if we assume processes will access multiple devices, memory allocations, streams, events, handles, etc., all need careful management.

As far as the `CUDADeviceGuard` guard, it should have a pretty minimal impact on performance except for the smallest kernels. What's the use case where one process manages multiple cuda devices?

lygztq:
> So the device guard should switch not only the device but also use the corresponding curand generator? I assume that also applies to other handles (like cusparse handles)?

I think it is true for curand, but I'm not sure for other libraries.

> Do we know how the device gets set to the wrong device in this case?

Here is the script that will cause an error like `dgl/src/array/cuda/array_op_impl.cu:219: Check failed: e == cudaSuccess || e == cudaErrorCudartUnloading: CUDA kernel launch error: an illegal memory access was encountered`

```
import dgl
from dgl.geometry import neighbor_matching

g1 = dgl.graph(([0, 1, 1, 2], [1, 0, 2, 1]), device="cuda:0")
res1 = neighbor_matching(g1)
print(res1)

g2 = g1.to("cuda:1")
res2 = neighbor_matching(g2)
print(res2)
```

I think the `curandGenerator_t` in `runtime::CUDAThreadEntry::ThreadLocal()` is first bound to `cuda:0` in `res1 = neighbor_matching(g1)`, and in `res2 = neighbor_matching(g2)` dgl directly uses the same handle to generate random numbers on `cuda:1`, which may cause an error.

Now I change to another implementation using curand device-api to avoid this problem as a temporary fix.

Issue #3312, opened 2021-09-01
[BUG] Kernel crashes using new hg.update_all


fmellomascarenhas:
## 🐛 Bug

I am using the latest version of DGL (PIP nightly), where update_all doesnt require the etype for heterogeneous graphs.
I have a batched graph with N=263 subgraphs. When I do hg.update_all(efunc, rfunc), my kernel restarts if N is big enough, without displaying any messages. I also dont see any spike or major change in memory (but I believe that this is probably the issue).

## To Reproduce

I tried to create a dummy code, but I couldn't reproduce the error, so I made my code available below and my batched graph available here: https://drive.google.com/file/d/1QIS1X3lfnkZynhaMJqDHzcky3dg3pxE_/view?usp=sharing

To narrow down the problem, I iterated over the edgetypes doing `subgraph_G = G.edge_type_subgraph([edge_to_test]) `, and found that edges types [2] and [27] where causing the issue.

Then I iterated over all b_i subgraphs `dgl.slice_batch(G, b_i) `, followed by .update_all, and didnt have any issue, so it isnt a subgraph problem.

Then I increased the number of subgraphs per batch by slicing N at a time: `sliced_G = dgl.batch([dgl.slice_batch(G, b_i) for b_i in range(init_idx, end_idx)])`, and found that if I sample enough subgraphs, then the crash will happen.

This makes me think that it is not a graph/node/edge/feature issue, however no errors are thrown, the kernel just restarts, and I dont see any memory spikes or increase while monitoring it.

```
import torch
import dgl
import dgl.function as fn

#load batched_graph
G = torch.load('subgraph')

#add features
for ntype in G.ntypes:
    v = torch.rand(G.num_nodes(ntype),4,8, device = 'cpu')
    node_att = torch.rand(G.num_nodes(ntype),4, device = 'cpu')
    G.nodes[ntype].data['v'] = v
    G.nodes[ntype].data['node_att'] = node_att

etypes = G.canonical_etypes

#create a smaller version of the batched_graph to prove that the problem is the size of sliced_G
#For ex: subgraphs 73-260 breaks for edge #2. subgraphs 74-261 works for edge #2.
init_idx = 0
end_idx = 262
sliced_G = dgl.batch([dgl.slice_batch(G, b_i) for b_i in range(init_idx, end_idx)])

########### BREAKS KERNEL ##############
#reduce the subgraph to a specific edgetype that is causing issues. I have found that edge#2 and edge#27 causes it. 
[...]

PR #3345, opened 2021-09-10
Fixes bug #3312

sanchit-misra:
## Description
This PR fixes the bug reported in issue #3312

## Checklist

### DIFF
--- a/src/array/cpu/spmm_blocking_libxsmm.h
+++ b/src/array/cpu/spmm_blocking_libxsmm.h
@@ -460,7 +460,8 @@ void SpMMRedopCsrOpt(
   const double avg_degree = total_nnz * 1.0 / M;
   const double nnz_prob = avg_degree / K;
 
-  IdType K_block_size = std::min((int64_t)K, (int64_t)(llc_size / (N * sizeof(DType) * nnz_prob * BLOCKING_HEURISTIC_PARAM)));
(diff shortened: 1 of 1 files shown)


VoVAllen (approved):

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -457,10 +457,11 @@ void SpMMRedopCsrOpt(
   const int total_nnz = indptr[M];
   if (M <= 0 || K <= 0 || N <= 0 || total_nnz <= 0) return;
 
-  const float avg_degree = total_nnz * 1.0 / M;
-  const float nnz_prob = avg_degree / K;
+  const double avg_degree = total_nnz * 1.0 / M;
+  const double nnz_prob = avg_degree / K;
```
BarclayII, line 461: Why does using double fix it?

sanchit-misra: The bug was that on line 463, llc_size/(N * ... ) was coming out to be more than 2^31 for the specific dataset used in #3312. As a result, it was treated as a negative number. Since we are doing a min there, K_block_size was set to a negative number. 
Using double increases the precision of that computation and then typecasting the result to int64 ensures that we are still using high precision.

PR #3356, opened 2021-09-16
[Performance]: NodeDataLoader CPU Affinity optimization part 1/2

tpatejko:
This PR turns parallel implementation of COOToCSR algorithm for unsorted and dense data into sequential one.
The PR is one of the PRs targeting optimizations for dataloading for CPUs. The second one is #3355.


VoVAllen:

pawelpiotrowicz:
@VoVAllen  You are right that worker has only one thread but point is that we want to isolate the thread from the rest of compute and assign it to specific core , it means that single worker is assigned to single core, thus avoiding cache-pollution.

nv-dlasalle:
This is a pretty big regression for `num_workers==0` and `num_workers==1`, and it also affects other parts of the code outside of dataloading (e.g., when loading the whole graph into shared memory). Can we just call `omp_set_num_threads()` inside of each dataloader worker if there are multiple workers?

pawelpiotrowicz:
@nv-dlasalle I don’t see “pretty big regression” from csv. Would you mind share your results with num-workers=0 or 1?. 
Does this solution affect somehow GPU stuffs, data loading? I’m not sure what you are talking about.

nv-dlasalle:
@pawelpiotrowicz Sorry, I was mistaken here. Now that I've measured it, for most cases in sampling, I do see better performance with fewer than the maximum number of threads even with 1 or no workers. However, when generating graph formats of the whole, it is much faster with all of the threads.

Rather than fully remove the parallel region, can we add an `if` clause to the omp parallel so that for small mini-batch graphs, it runs serially, but for larger graphs we use parallelism?
```
#pragma omp parallel if (N > X)
```
We'd need to tune when to spawn threads `X` (maybe ~100K), but that way we would still get the benefits of parallelism for large graphs.

Rhett-Ying:
@dgl-bot run m6i.16xlarge bench_format_conv dmlc/master

Rhett-Ying:
@nv-dlasalle how to quantify **large enough**?

Rhett-Ying:
> Finished the Regression test. Result table is at https://dgl-asv-data.s3-us-west-2.amazonaws.com/f3137c7d604ccf07492d33af57ced7b391e42c20_m6i16xlarge/results/result.csv. Jenkins job link is https://ci.dgl.ai/job/dgl/job/PR-3356/4/display/redirect.

get much worse in several cases when converting from coo to csr.

nv-dlasalle:
> 
> 
> @nv-dlasalle how to quantify **large enough**?

On my system with 12 cores, ~100K edges seems to be the break even point, where the serial "dense" version is the same speed as the parallel "sparse" version. But I'd expect this number to vary widely across systems and datasets.

My suggestion would be to change this line: https://github.com/dmlc/dgl/blob/master/src/array/cpu/spmat_op_impl_coo.cc#L628 to something like:
```
if (num_edges < 200000) {
```
With `200,000` being a guess as to the proper threshold--we probably need a good set of experiments to find where this should reasonably be.

Rhett-Ying:
> > @nv-dlasalle how to quantify **large enough**?
> 
> On my system with 12 cores, ~100K edges seems to be the break even point, where the serial "dense" version is the same speed as the parallel "sparse" version. But I'd expect this number to vary widely across systems and datasets.
> 
> My suggestion would be to change this line: https://github.com/dmlc/dgl/blob/master/src/array/cpu/spmat_op_impl_coo.cc#L628 to something like:
> 
> ```
> if (num_edges > 200000) {
> ```
> 
> With `200,000` being a guess as to the proper threshold--we probably need a good set of experiments to find where this should reasonably be.

Yes, it's quite subtle. I'm wondering if we could provide an flexible way to let the user choose appropriate algo to obtain best perf.

BarclayII:
My suggestion is (1) merge the two PRs into one PR and rerun the regression since the current regression result is unlikely to reflect the actual benefit, and (2) add a benchmark that demonstrates the benefit of fixing affinity (which can precede this PR).

pawelpiotrowicz:
Due to regression we decided to suspend both PRs.

Issue #3402, opened 2021-10-08
nvlink NCCL bug when Using NodeEmbedding on 8GPU-V100 


lixusign:
## 🐛 Bug

I'm sorry there is problem when I test my model ,  the maybe a bug  and  mark this in https://discuss.dgl.ai/t/dgl0-7-with-nccl-init-error-when-using-nvlink-v100/2389   

when use embedding and gpus cross numa node  this will happen

thx 

## Environment

 - DGL Version (e.g., 1.0):  dgl-cu110
 - Backend Library & Version (e.g., PyTorch 1.7, MXNet/Gluon 1.3):  PyTorch 1.7
 - OS (e.g., Linux): linux
 - How you installed DGL (`conda`, `pip`, source): pip
 - Python version:3.6.8
 - CUDA/cuDNN version (if applicable):  11.0
 - nccl:2.8.4
 - GPU models and configuration (e.g. V100):  v100-nvlink * 8

lixusign:
this merge(https://github.com/dmlc/dgl/pull/3404) repair this issue thx

PR #3404, opened 2021-10-08
[Bugfix] Split nccl sparse push into two groups

nv-dlasalle:
## Description
It appears that in some cases with some versions of NCCL, the sparse all to all push operations fails write the proper data to the outputs, which in turn causes strange deadlocking behavior. This may be the cause of #3402.

The cause looks to be grouping two sends and two receives together, so this PR splits those into two separate communication groups. In testing with the /examples/pytorch/rgcn/entity_classify_mp.py example, this appears to fix the issue with PyTorch 1.9.1.

## Checklist
- [x] Code is well-documented


VoVAllen:
@classicsong Have you tested it?

Issue #3216, opened 2021-08-04
Edge subgraph with relabeling does not support GPU.


senllh:
When I use "cpu" to execute the code there is no problem
But when I use "cuda" running code

for input_nodes, edge_subgraph, blocks in train_dataloader:
    blocks = [b.to(torch.device(device)) for b in blocks]

there is a bug: Check failed: hg->Context().device_type != kDLGPU: Edge subgraph with relabeling does not support GPU.

VoVAllen:
Can you post the full traceback? Thanks

senllh:
File "G:/../main.py", line 245, in <module>
    run_train()
  File "G:/../main.py", line 188, in run_train
    for input_nodes, edge_subgraph, blocks in train_dataloader:
  File "D:\anaconda\lib\site-packages\dgl\dataloading\pytorch\dataloader.py", line 339, in __next__
    result_ = next(self.iter_)
  File "D:\anaconda\lib\site-packages\torch\utils\data\dataloader.py", line 435, in __next__
    data = self._next_data()
  File "D:\anaconda\lib\site-packages\torch\utils\data\dataloader.py", line 475, in _next_data
    data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
  File "D:\anaconda\lib\site-packages\torch\utils\data\_utils\fetch.py", line 47, in fetch
    return self.collate_fn(data)
  File "D:\anaconda\lib\site-packages\dgl\dataloading\pytorch\dataloader.py", line 288, in collate
    result = super().collate(items)
  File "D:\anaconda\lib\site-packages\dgl\dataloading\dataloader.py", line 778, in collate
    return self._collate(items)
  File "D:\anaconda\lib\site-packages\dgl\dataloading\dataloader.py", line 682, in _collate
    pair_graph = self.g.edge_subgraph(items)
  File "D:\anaconda\lib\site-packages\dgl\utils\internal.py", line 904, in _fn
    return func(*args, **kwargs)
  File "D:\anaconda\lib\site-packages\dgl\subgraph.py", line 299, in edge_subgraph
    sgi = graph._graph.edge_subgraph(induced_edges, not relabel_nodes)
  File "D:\anaconda\lib\site-packages\dgl\heterograph_index.py", line 885, in edge_subgraph
    return _CAPI_DGLHeteroEdgeSubgraph(self, eids, preserve_nodes)
[...]

PR #3445, opened 2021-10-27
[Feature] aten::Relabel_() for the GPU

yaox12:
## Description

This implements `aten:Relabel_()` for the GPU, which thus enables `g.edge_subgraph()/in_subgraph()/out_subgraph()` with `relabel_node=True`, as well as `EdgeDataLoader(negative_sampler=None)`, run on the GPU.
The ordered hashtable in CUDA is used for node mapping and relabeling.

Fix this issue https://github.com/dmlc/dgl/issues/3216.

## Checklist

- [x] Code is well-documented

## Changes
- Modify `src/array/array.cc`: enable switch for the GPU version
- Modify `src/array/cuda/array_op_impl.cu`: CUDA implementation of `Relabel_()`
- Modify `src/graph/heterograph.cc`: enable `g.edge_subgraph()` with `preserve_nodes=False` for graphs on the GPU
- Modify `tests/cpp/test_aten.cc`: enable the unit test for `aten::Relabel_()` on the GPU
- Modify `tests/compute/test_subgraph.py`: enable unit tests for `g.edge_subgraph()/in_subgraph()/out_subgraph()` on the GPU

### DIFF
--- a/src/array/cuda/array_op_impl.cu
+++ b/src/array/cuda/array_op_impl.cu
@@ -277,12 +277,16 @@ __global__ void _RelabelKernel(
 
 template <DLDeviceType XPU, typename IdType>
 IdArray Relabel_(const std::vector<IdArray>& arrays) {
+  IdArray all_nodes = Concat(arrays);
+  const int64_t total_length = all_nodes->shape[0];
+
+  if (total_length == 0) {
+    return all_nodes;
+  }
+
   const auto& ctx = arrays[0]->ctx;
   auto device = runtime::DeviceAPI::Get(ctx);
   auto* thr_entry = runtime::CUDAThreadEntry::ThreadLocal();
-
-  IdArray all_nodes = Concat(arrays);
-  const int64_t total_length = all_nodes->shape[0];
 
   // build node maps and get the induced nodes
   OrderedHashTable<IdType> node_map(total_length, ctx, thr_entry->stream);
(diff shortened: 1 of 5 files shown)


src/array/cuda/array_op_impl.cu
```diff
@@ -258,6 +260,84 @@ IdArray Range(IdType low, IdType high, DLContext ctx) {
...
+  device->StreamSync(ctx, thr_entry->stream);
+  device->FreeWorkspace(ctx, num_induced_device);
+
+  // resize the induced nodes
+  induced_nodes->shape[0] = num_induced;
+
+  // relabel
+  for (IdArray arr : arrays) {
+    const int64_t length = arr->shape[0];
+    int nt = cuda::FindNumThreads(length);
```
nv-dlasalle, line 328: I think we should stay away from `cuda::FindNumThreads()`, as it tries to use the maximum number of threads per block, which can lead insufficient use of available SMs, as well as performance anomalies as the number of threads changes with the input size. I'd advice using a fixed number of threads (128 or 256).

nv-dlasalle (approved):
Overall looks good to me.

src/array/cuda/array_op_impl.cu
yaox12: Fixed f8d4bb8.

nv-dlasalle:
@BarclayII Are you satisfied with the enabled unit tests, or do you want any other changes before merging?

BarclayII (approved):
Thanks a lot!  Appreciate it.

PR #3450, opened 2021-10-28
[Optimization] Overhead optimization for heterogeneous graph

isratnisa:
## Description
Optimization 1:  Instead of calling `F.zeros()`  (cuda malloc + initialize to zero) for each etype separately, compute the total memory requirement and call F.zeros once. Reshaping/splitting the merged tensor for separated etypes is required before calling the CAPI. It provides better GPU utilization and provides 25x speedup (from 1.4 ms to .05ms) on AM dataset.  

Optimization 2:  Avoid ctx check for each etype: Instead of fetching the src and dst node type for each relation, just parsing the first etype. I am “assuming” the context, datatype, ndims, and dimension of the features will be the same for all node types. In the C or CUDA API, we are already assuming all the etypes have the same numbers of ndims, dims, etc.

## Checklist
- [x] Code is well-documented


isratnisa:
PR too old. Need to re-evaluate if the overhead still exists in the latest releases.

PR #3455, opened 2021-10-28
Add edge_weight parameters to nn modules

VoVAllen:
## Checklist
- [ ] Code is well-documented


mufeili:
Add corresponding unit tests

python/dgl/nn/pytorch/conv/appnpconv.py
```diff
@@ -80,6 +80,9 @@ def forward(self, graph, feat):
         feat : torch.Tensor
             The input feature of shape :math:`(N, *)`. :math:`N` is the
             number of nodes, and :math:`*` could be of any shape.
+        edge_weight: Optional[torch.Tensor]
```
mufeili, line 83: I think previously we use "torch.Tensor, optional"

python/dgl/nn/pytorch/conv/appnpconv.py
```diff
@@ -80,6 +80,9 @@ def forward(self, graph, feat):
...
+            edge_weight used in the message passing process. Default to th.ones
+            if not specified
```
mufeili, line 85: How will this affect the equation above? Replace the binary adjacency matrix $A$ with a weighted one?

python/dgl/nn/pytorch/conv/appnpconv.py
```diff
@@ -99,8 +102,10 @@ def forward(self, graph, feat):
                 # normalization by src node
```
mufeili, line 102: Should edge weight also participate in symmetric adjacency normalization as in [this implementation](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.conv.APPNP)?

python/dgl/nn/pytorch/conv/gcn2conv.py
```diff
@@ -182,11 +182,13 @@ def forward(self, graph, feat, feat_0):
...
             The graph.
         feat : torch.Tensor
-             The input feature of shape
+            The input feature of shape
             :math:`(N, D_{in})`
             where :math:`D_{in}` is the size of input feature and :math:`N` is the number of nodes.
         feat_0 : torch.Tensor
-                The initial feature of shape :math:`(N, D_{in})`
+            The initial feature of shape :math:`(N, D_{in})`
+        edge_weight: Optional[torch.Tensor]
```
mufeili, line 190: Ditto

python/dgl/nn/pytorch/conv/gcn2conv.py
```diff
@@ -170,7 +170,7 @@ def set_allow_zero_in_degree(self, set_value):
         """
         self._allow_zero_in_degree = set_value
 
-    def forward(self, graph, feat, feat_0):
+    def forward(self, graph, feat, feat_0, edge_weight=None):
```
mufeili, line 173: In ":math: `\beta` ensures that the decay of", the rendering of beta is not correct for the built documentation. Can you also fix that?

python/dgl/nn/pytorch/conv/gcn2conv.py
```diff
@@ -182,11 +182,13 @@ def forward(self, graph, feat, feat_0):
...
         feat : torch.Tensor
-             The input feature of shape
+            The input feature of shape
             :math:`(N, D_{in})`
             where :math:`D_{in}` is the size of input feature and :math:`N` is the number of nodes.
         feat_0 : torch.Tensor
-                The initial feature of shape :math:`(N, D_{in})`
+            The initial feature of shape :math:`(N, D_{in})`
+        edge_weight: Optional[torch.Tensor]
+            edge_weight used in the message passing process.
```
mufeili, line 191: How does edge weight participate in the equations above?

python/dgl/nn/pytorch/conv/gcn2conv.py
```diff
@@ -230,7 +232,11 @@ def forward(self, graph, feat, feat_0):
```
mufeili, line 232: Should edge weight also participate in symmetric adjacency normalization?

mufeili:
Basically I have the same questions for all NN Modules.

python/dgl/nn/pytorch/conv/gcn2conv.py
VoVAllen: Fixed

python/dgl/nn/pytorch/conv/gcn2conv.py
VoVAllen: https://github.com/pyg-team/pytorch_geometric/blob/74245f3a680c1f6fd1944623e47d9e677b43e827/torch_geometric/nn/conv/gcn_conv.py#L30-L65

I checked pyg's implementation. It uses the weight in the normalization part. Do you think it's reasonable?

mufeili: Personally I find this reasonable. Previously Tong has added `edge_weight` support to `GraphConv`, which does not perform normalization for weighted adjacency matrices. @hetong007 Do you find this reasonable? If so, what do you think of the backward compatibility for `edge_weight` in `GraphConv`?

hetong007: My PR didn't normalize the input because it accepts multi-dimensional weight vectors and there can be multiple ways to normalize it. Instead, I created a separate module that normalize a positive scalar edge weight as a standard implementation, and also let users to use customized normalization before feeding the edge-weight to the module.

mufeili: @hetong007 I see. Thanks for the clarification.
@VoVAllen I think [`EdgeWeightNorm`](https://docs.dgl.ai/api/python/nn.pytorch.html#dgl.nn.pytorch.conv.EdgeWeightNorm) already covers the cases we want. Please have a double check. If this is the case, then I guess we don't need to do anything :) We will however need to add a pointer to `EdgeWeightNorm` in the doc of these NN modules.

python/dgl/nn/pytorch/conv/appnpconv.py
```diff
@@ -80,6 +83,10 @@ def forward(self, graph, feat):
         feat : torch.Tensor
             The input feature of shape :math:`(N, *)`. :math:`N` is the
             number of nodes, and :math:`*` could be of any shape.
+        edge_weight: torch.Tensor, optional
+            edge_weight used in the message passing process.This is equivalent to use real-weighted
```
mufeili, line 87: 1. `used` -> `to use`
2. Add an empty space after "process."

python/dgl/nn/pytorch/conv/appnpconv.py
```diff
@@ -80,6 +83,10 @@ def forward(self, graph, feat):
...
+            adjacency in the equation above, and the degree normalization uses edge weight also,
```
mufeili, line 88: "adjacency" -> "adjacency matrix"

python/dgl/nn/pytorch/conv/gcn2conv.py
```diff
@@ -182,11 +184,16 @@ def forward(self, graph, feat, feat_0):
...
+            The input feature of shape
             :math:`(N, D_{in})`
             where :math:`D_{in}` is the size of input feature and :math:`N` is the number of nodes.
         feat_0 : torch.Tensor
-                The initial feature of shape :math:`(N, D_{in})`
+            The initial feature of shape :math:`(N, D_{in})`
+        edge_weight: torch.Tensor, optional
+            edge_weight used in the message passing process.This is equivalent to use real-weighted
+            adjacency in the equation above, and the degree normalization uses edge weight also,
+            based on :ref:`EdgeWeightNorm`.
```
mufeili, line 195: Ditto as in APPNP

python/dgl/nn/pytorch/conv/sgconv.py
```diff
@@ -144,6 +144,8 @@ def forward(self, graph, feat):
         feat : torch.Tensor
             The input feature of shape :math:`(N, D_{in})` where :math:`D_{in}`
             is size of input feature, :math:`N` is the number of nodes.
+        edge_weight: torch.Tensor, optional
+            edge_weight used in the message passing process.
```
mufeili, line 148: "used" -> "to use"

mufeili: You may want to add a pointer to `EdgeNorm`

python/dgl/nn/pytorch/conv/tagconv.py
```diff
@@ -103,6 +104,8 @@ def forward(self, graph, feat):
         feat : torch.Tensor
             The input feature of shape :math:`(N, D_{in})` where :math:`D_{in}`
             is size of input feature, :math:`N` is the number of nodes.
+        edge_weight: torch.Tensor, optional
+            edge_weight used in the message passing process.
```
mufeili, line 108: Ditto

mufeili:
It seems `SGConv` and `TAGConv` also involve symmetric adjacency normalization. Are you going to update them and add corresponding unit tests?

mufeili (approved):
LGTM

PR #3487, opened 2021-11-08
[Sampling] Cluster-GCN and ShaDow-GNN DataLoader

BarclayII:
## Description
This is an implementation of two subgraph samplers: Cluster-GCN and ShaDow-GNN, which are representative for two kinds of subgraph samplers:
* One that iterates over the subgraphs without explicit notion of seed nodes/edges (e.g. Cluster-GCN, GraphSAINT).
* Otherwise (e.g. ShaDow-GNN, which expands the seed nodes' neighborhood much like neighbor sampler but returns an induced subgraph instead).

Basically, the experience for both samplers looks like follows:

```python
# Cluster-GCN
dataset = dgl.data.CoraFullDataset()
g = dataset[0]
sgiter = dgl.dataloading.ClusterGCNSubgraphIterator(g, 100, '.', refresh=True)
dataloader = dgl.dataloading.GraphDataLoader(sgiter, batch_size=4, num_workers=num_workers)
for sg in dataloader:
    train_on(sg)
```

```python
# ShaDow-GNN
dataset = dgl.data.CoraFullDataset()
g = dataset[0]
sampler = dgl.dataloading.ShaDowKHopSampler([3, 3])
dataloader = dgl.dataloading.NodeDataLoader(g, train_nids, sampler, batch_size=4, num_workers=num_workers)
for input_nodes, output_nodes, (subgraph,) in dataloader:
    train_on(sg)
```

## Checklist
- [x] Code is well-documented


VoVAllen:
Potentially we might want support loading subgraph from disk for ClusterGCNTraining. And also is it possible to use NodeDataloader inside the GraphDataloader loop?

BarclayII:
> In future we might want to support loading subgraph from disk for ClusterGCNTraining.

This should be doable with a switch of the cache format.  Currently it's saving the partition assignment and I can switch it to saving the actual partitions.  I'll leave it to a later PR though.

> is it possible to use NodeDataloader inside the GraphDataloader loop?

This is tricky.  ClusterGCN does not have a notion of seed nodes so I think `NodeDataLoader` cannot be used here.  It might be possible if we extend `NodeDataLoader` to allow customized minibatch iteration but I'm not sure what is the best practice (could be possible using PyTorch's custom minibatch sampler).

python/dgl/dataloading/cluster_gcn.py
```diff
@@ -0,0 +1,80 @@
...
+        The path to the cache directory for storing the partition result.
+    refresh : bool
+        If True, recompute the partition.
+
+    Examples
+    --------
+    Assuming that you have a graph ``g``:
+
+    >>> sgiter = dgl.dataloading.ClusterGCNSubgraphIterator(g, 100, '.', refresh=True)
+    >>> dataloader = dgl.dataloading.GraphDataLoader(sgiter, batch_size=4, num_workers=num_workers)
```
mufeili, line 38: Give a specific value for `num_workers` for consistency.

mufeili:
Update the `.rst` files for documentation.

python/dgl/dataloading/shadow.py
```diff
@@ -0,0 +1,82 @@
+"""ShaDow-GNN subgraph samplers."""
+from ..utils import prepare_tensor_or_dict
+from ..base import NID
+from ..sampling import sample_neighbors
+from .neighbor import NeighborSamplingMixin
+from .dataloader import exclude_edges, Sampler
+
+class ShaDowKHopSampler(NeighborSamplingMixin, Sampler):
+    """K-hop subgraph sampler used by
+    `ShaDow-GNN <https://openreview.net/forum?id=GIeGTl8EYx>`__.
```
mufeili, line 10: It might be better to give the latest ArXiv link: https://arxiv.org/abs/2012.01380

python/dgl/dataloading/shadow.py
```diff
@@ -0,0 +1,82 @@
...
+        List of neighbors to sample per edge type for each GNN layer, starting from the
+        first layer.
+
+        If the graph is homogeneous, only an integer is needed for each layer.
+
+        If None is provided for one layer, all neighbors will be included regardless of
+        edge types.
+
+        If -1 is provided for one edge type on one layer, then all inbound edges
+        of that edge type will be included.
```
mufeili, line 30: I feel the description for `fanouts` needs to be improved. 
1. Why not stick to `-1` or `None` for both cases?
2. Prepare two paragraphs, one for homogeneous graphs and one for heterogeneous graphs.
3. If `fanouts` is a `list[int]`, is `fanouts[0]` for the first GNN layer or the last GNN layer? Even if users have read the user guide, they may not clearly remember that. So it's better to spell it out.
4. Provide concrete examples below for the different cases mentioned here.

python/dgl/dataloading/shadow.py
```diff
@@ -0,0 +1,82 @@
...
+    replace : bool, default True
+        Whether to sample with replacement
+    prob : str, optional
```
mufeili, line 33: I guess `prob` should be a tensor instead?

python/dgl/dataloading/shadow.py
```diff
@@ -0,0 +1,82 @@
...
+    To train a 3-layer GNN for node classification on a set of nodes ``train_nid`` on
+    a homogeneous graph where each node takes messages from 5, 10, 15 neighbors for
+    the first, second, and third layer respectively (assuming the backend is PyTorch):
+
+    >>> sampler = dgl.dataloading.ShaDowKHopSampler([5, 10, 15])
+    >>> dataloader = dgl.dataloading.NodeDataLoader(
+    ...     g, train_nid, sampler,
+    ...     batch_size=1024, shuffle=True, drop_last=False, num_workers=4)
+    >>> for input_nodes, output_nodes, (subgraph,) in dataloader:
+    ...     train_on(subgraph)
```
mufeili, line 48: Print the values of `input_nodes`, `output_nodes` and `subgraph`.

python/dgl/dataloading/shadow.py
```diff
@@ -0,0 +1,82 @@
...
+
+    If training on a heterogeneous graph and you want different number of neighbors for each
+    edge type, one should instead provide a list of dicts.  Each dict would specify the
```
mufeili, line 51: Extra empty space before "Each"

python/dgl/dataloading/dataloader.py
```diff
@@ -195,15 +204,67 @@ def _find_exclude_eids(g, exclude_mode, eids, **kwargs):
     else:
         raise ValueError('unsupported mode {}'.format(exclude_mode))
 
+class Sampler(object):
+    """An abstract class that takes in a graph and a set of seed nodes and returns a
+    structure representing a smaller portion of the graph for computation.  It can
```
mufeili, line 209: Extra empty space before "It"

python/dgl/dataloading/dataloader.py
```diff
@@ -195,15 +204,67 @@ def _find_exclude_eids(g, exclude_mode, eids, **kwargs):
...
+    be either a list of bipartite graphs (i.e. :class:`BlockSampler`), or a single
+    subgraph (i.e. :class:`SubgraphSampler`).
```
mufeili, line 211: `SubgraphSampler` cannot be linked.

python/dgl/dataloading/dataloader.py
```diff
@@ -195,15 +204,67 @@ def _find_exclude_eids(g, exclude_mode, eids, **kwargs):
...
+        raise NotImplementedError
+
+    def set_output_context(self, ctx):
+        """Set the device the generated block or subgraph will be output to.
+        This should only be set to a cuda device, when multi-processing is not
+        used in the dataloader (e.g., num_workers is 0).
+
+        Parameters
+        ----------
+        output_ctx : DGLContext, default None
```
mufeili, line 248: `output_ctx` -> `ctx`

python/dgl/dataloading/neighbor.py
```diff
@@ -5,7 +5,55 @@
...
+    - :attr:`prob`: The edge feature name that stores the (unnormalized) probability.
+    - :attr:`fanouts`: The list of fanouts (either an integer or a dictionary of edge
+      types and integers).
+
+    The mixin will generate the following attributes:
+
+    - :attr:`prob_arrays`: List of DGL NDArrays containing the unnormalized probabilities
+      for every edge type.
+    - :attr:`fanout_arrays`: List of DGL NDArrays containing the fanouts for every edge
+      type at every later.
```
mufeili, line 23: `later` -> `layer`.

python/dgl/dataloading/neighbor.py
```diff
@@ -5,7 +5,55 @@
...
+        elif self.prob_arrays is None:
+            # build prob_arrays only once
+            self.prob_arrays = [nd.array([], ctx=nd.cpu())] * len(g.etypes)
+
+    def _build_fanout(self, block_id, g):
+        assert not self.fanouts is None, \
+            "_build_fanout() should only be called when fanouts is not None"
+        # build fanout_arrays only once for each layer
+        while block_id >= len(self.fanout_arrays):
+            for i in range(len(self.fanouts)):
```
mufeili, line 42: L41 - L42 is a bit weird. Is it equivalent to use the minimum of `block_id` and `len(self.fanouts)`?

python/dgl/transform.py
```diff
@@ -1303,8 +1304,8 @@ def laplacian_lambda_max(g):
         adj = g_i.adj(transpose=True, scipy_fmt=g_i.formats()['created'][0]).astype(float)
         norm = sparse.diags(F.asnumpy(g_i.in_degrees()).clip(1) ** -0.5, dtype=float)
         laplacian = sparse.eye(n) - norm * adj * norm
-        rst.append(sparse.linalg.eigs(laplacian, 1, which='LM',
-                                      return_eigenvectors=False)[0].real)
+        rst.append(scipy.sparse.linalg.eigs(
+            laplacian, 1, which='LM', return_eigenvectors=False)[0].real)
```
mufeili, line 1308: What's the purpose of the modifications made in this file?

python/dgl/dataloading/neighbor.py
```diff
@@ -30,6 +78,9 @@ class MultiLayerNeighborSampler(BlockSampler):
     return_eids : bool, default False
         Whether to return the edge IDs involved in message passing in the MFG.
         If True, the edge IDs will be stored as an edge feature named ``dgl.EID``.
+    prob : str, optional
+        If given, the probability of each neighbor being sampled is proportional
+        to the edge feature with the given name.
```
mufeili, line 83: Ditto

python/dgl/utils/internal.py
```diff
@@ -691,6 +691,11 @@ def __getitem__(self, idx):
         g = self._groups[k]
         return k, g[j]
 
+def maybe_flatten_dict(data):
+    """Return a FlattenedDict if the input is a Mapping, or the data itself otherwise.
+    """
+    return FlattenedDict(data) if isinstance(data, Mapping) else data
```
mufeili, line 697: Is this used anywhere?

python/dgl/dataloading/neighbor.py
BarclayII: I think it can be rewritten to `for i in range(len(self.fanouts))`.  @nv-dlasalle any idea why it's written this way?

python/dgl/transform.py
BarclayII: To pass lint tests.  Otherwise lint will complain that the import `scipy.sparse.linalg` is never used.

python/dgl/dataloading/shadow.py
BarclayII: I simplified the interface so that there's only -1.  None is no longer accepted.  A minor breaking change though as I don't think people uses None explicitly instead of using `MultiLayerFullNeighborSampler`.

python/dgl/dataloading/shadow.py
BarclayII: This is how I put it now:
```
If given, the probability of each neighbor being sampled is proportional
to the edge feature value with the given name.  The feature must be
a scalar on each edge.
```

mufeili:
Are you going to open follow-up PRs for adding/updating examples of SEAL, GraphSAINT and ShaDow-GNN?

docs/source/api/python/dgl.dataloading.rst
```diff
@@ -39,6 +39,28 @@ the ``sample_blocks`` methods.
...
     :show-inheritance:
 
+Subgraph Iterators
+------------------
+Subgraph iterators iterate over the original graph in subgraphs. One should use subgraph
+iterators with ``GraphDataLoader`` like follows:
+
+.. code:: python
+
+   sgiter = dgl.dataloading.ClusterGCNSubgraphIterator(g, 100, '.', refresh=True)
```
mufeili, line 49: `g, 100, '.', refresh=True` -> `g, num_partitions=100, cache_directory='.', refresh=True`

docs/source/api/python/dgl.dataloading.rst
```diff
@@ -39,6 +39,28 @@ the ``sample_blocks`` methods.
...
+   dataloader = dgl.dataloading.GraphDataLoader(sgiter, batch_size=4, num_workers=0)
+   for subgraph_batch in dataloader:
+       train_on(subgraph_batch)
+
+.. autoclass:: dgl.dataloading.dataloader.SubgraphIterator
```
mufeili, line 54: Will `.. autoclass::` display private methods like `_cache_file_path`, `_load_parts`, etc?

python/dgl/dataloading/shadow.py
mufeili: Add `in g.edata` after `the given name`. Also, are both `(E)` and `(E, 1)` acceptable? It might be good to spell it out in the doc.

mufeili (approved):
LGTM in general

python/dgl/dataloading/neighbor.py
nv-dlasalle: I don't remember exactly it was written this way (it also doesn't appear entirely correct). Since we never partially build the fanout arrays, it works as is, but in theory if we built it partially, this wouldn't function properly. Especially now that it's part of a separate class, it probably should be something like to only build the  missing layers.
```
if len(self.fanouts) > len(self.fanout_arrays):
  for i in range(len(self.fanout_arrays), len(self.fanouts)):
```

BarclayII:
They will be in separate PRs.

docs/source/api/python/dgl.dataloading.rst
BarclayII: No.

python/dgl/dataloading/neighbor.py
BarclayII: Will fix in the next PR.

PR #3490, opened 2021-11-09
[NN] GNNExplainer

mufeili:
## Description
@BarclayII It turns out that I also need the new IDs of the input nodes to k hop subgraph extraction and I've added that. See if that meets your need.
@zhjwy9343 It will be great if you can take a look. Once we are good with the design, I will update the GNNExplainer example with this built-in module.

## Checklist
- [x] Code is well-documented


conda/dgl/meta.yaml
```diff
@@ -19,6 +19,7 @@ requirements:
     - scipy
     - networkx
     - requests
+    - tqdm
```
VoVAllen, line 22: Is this necessary? Because tqdm may have unexpected bugs based on user environment, such as terminal without certain functions or writing to logs.

python/dgl/nn/pytorch/explain/gnnexplainer.py
```diff
@@ -0,0 +1,412 @@
...
+    critical role in GNN-based node classification and graph classification.
+
+    Parameters
+    ----------
+    model : nn.Module
+        The GNN model to explain.
+
+        * The required arguments of its forward function are graph and feat.
+          The latter one is for input node features.
+        * It should also optionally take an eweight argument for edge weights
```
VoVAllen, line 29: But eweight is also required right? Does it have to be a keyword argument or positional argument as the 3rd?

VoVAllen: You can write a simple code to described to interface

python/dgl/subgraph.py
```diff
@@ -726,8 +747,11 @@ def khop_out_subgraph(graph, nodes, k, *, relabel_nodes=True, store_ids=True):
 
     Returns
     -------
-    G : DGLGraph
+    DGLGraph
         The subgraph.
+    Tensor or dict[str, Tensor], optional
```
VoVAllen, line 752: Is this a breaking change? Is it possible to store the inverse in the ndata so that the returns are the same for both cases?

VoVAllen (approved):
Generally looks good to me

PR #3495, opened 2021-11-12
Fix User Guide EN and CN typo

yang-su2000:
## Changes

- [1.5]  `tensor([0, 1, 2, 0, 1, 2])` should be output instead of code
- [2.1]  `updata_all_example()` should be `update_all_example()`
- [2.2]  `cat_feat` need to concatenate with `dim=1` for the # edge features to match # edges
- [3.1, 3.2] all `max_pool` in the aggregator type of `SAGEConv` should be `pool` instead


zhjwy9343 (approved):
LGTM

PR #3500, opened 2021-11-14
[Randomwalk] Fix off-by-one bug in GenericRandomWalk()

erickim555:
## Description
This PR fixes an off-by-one bug in `GenericRandomWalk()`. The `traces_data` is a tensor with shape=[num_seed_nodes, trace_length] which is supposed to store the visited node ids for each seed node.

The line in question is supposed to something like:
```
dgl_id_t curr;  // current node_id in rwalk
int64_t i;  // current step iter
...
// perform an rwalk step()
const auto &succ = step(&traces_data[seed_id, :], curr, i);
```

However, in the current implementation the indexing into `traces_data` is incorrect: it's off by one, so traces from other seed nodes can clobber other seed node traces.

Anecdotally, when I was implementing a new Step() function where I needed to read into the passed in `traces_data`, I observed a memory corruption bug (leading to crashes), which led me to this bug. Enabling multithreading (running rwalks for multiple seed nodes concurrently) led to frequent segfaults/failures. This PR fixed the memory corruption bugs on my end.

## Checklist
- [x ] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x ] Changes are complete (i.e. I finished coding on this PR)
- [x ] All changes have test coverage
- [x ] Code is well-documented
- [x ] To the best of my knowledge, examples are either not affected by this change,

### DIFF
--- a/src/graph/sampling/randomwalks/randomwalks_cpu.h
+++ b/src/graph/sampling/randomwalks/randomwalks_cpu.h
@@ -59,7 +59,7 @@ std::pair<IdArray, IdArray> GenericRandomWalk(
       CHECK_LT(curr, max_nodes) << "Seed node ID exceeds the maximum number of nodes.";
 
       for (i = 0; i < max_num_steps; ++i) {
-        const auto &succ = step(traces_data + seed_id * max_num_steps, curr, i);
+        const auto &succ = step(traces_data + seed_id * trace_length, curr, i);
         traces_data[seed_id * trace_length + i + 1] = curr = std::get<0>(succ);
         eids_data[seed_id * max_num_steps + i] = std::get<1>(succ);
         if (std::get<2>(succ))


BarclayII (approved):

BarclayII:
Thanks for catching this!  What was the random walk algorithm you were using?  This bug never got noticed because our random walk algorithms never required the entire random walk history.

erickim555:
@BarclayII I'm currently implementing a new random walk algorithm with the following behavior:

Pinsage-like random walk improved performance + visit count tracking. For a given seed node, perform N=10_000 random walk steps with random restart, and gather the top_n_visited_neighbors=50 along with their visit counts. This is essentially the existing `RandomWalkNeighborSampler`, but with the expensive Python overhead (duplicating seed nodes + computing visit counts) shifted into the C++ implementation:
https://github.com/dmlc/dgl/blob/195f99362d883f8b6d131b70a7868a537e55b786/python/dgl/sampling/pinsage.py#L14

Non-Metapath based random walks. Rather than passing in an explicit meta-path to do random walks in, I enable a conceptually simpler random walk approach on heterogeneous graphs: to take a step, first randomly sample an outgoing edge type, then randomly sample a successor node from the chosen edge type.

In my use case, my neighbor sampling throughput is highly bottlenecked by the implementation details of the existing DGL random walk code (eg the kind of expensive Python overhead that `RandomWalkNeighborSampler` is forced to do due to the ABI of the underlying DGL C++ random walk functions). So, I've opted to implement a new random walk variant that is both performant and allows the "non-metapath" behavior. Early testing is promising, I'm seeing several orders of magnitude improved random walk throughput compared to the baseline `RandomWalkNeighborSampler`.

If things go well, I'd like to push this into upstream eventually if it'd be useful/helpful for others.

BarclayII:
@erickim555 Yes please do!

PR #3501, opened 2021-11-15
[Doc] Fix the example code of negative sampler

yaox12:
## Description

Fix the example code of the negative sampler, where the argument `eid` is required to be `Tensor or dict[etype, Tensor]`.


VoVAllen (approved):
Thanks

Issue #3459, opened 2021-10-29
Error on LibXSMM cpu kernels


VoVAllen:
## 🐛 Bug

User met errors raised at https://github.com/dmlc/dgl/blob/983a4fdd1981a6eaa4a3343ec4116739e9f97dfa/src/array/cpu/spmm_blocking_libxsmm.h#L267

```
 "Failed to generate libxsmm kernel for the SpMM operation!"
```

We should not raised the error but fallback to the naive kernel.

User's CPU model: Xeon(R) CPU E5-2695 v2 @ 2.40GHz
This is an old CPU produced in 2013, which might not be supported by LibXSMM now. https://ark.intel.com/content/www/us/en/ark/products/75281/intel-xeon-processor-e52695-v2-30m-cache-2-40-ghz.html

## Possible Solution

Catch error at https://github.com/dmlc/dgl/blob/983a4fdd1981a6eaa4a3343ec4116739e9f97dfa/src/array/cpu/spmm.h#L144, and run the naive kernel if error detected

sanchit-misra:
@BarclayII will take a look.

sanchit-misra:
I was not able to reproduce this but I don't have access to such an old system. :-)
According to the libxsmm developers (who are my colleagues), while libxsmm did not support any architectures without at least AVX2 support, it did not explicitly check whether the underlying architecture is supported. So, it would not throw an error if the architecture was not supported. So, not sure where this error came from.

Having said that, libxsmm *now* explicitly checks whether the architecture is supported. And if it is not supported, it returns a nullptr kernel. I check this and fall back to naive kernel.

sixtyfive:
I'm hitting this on an older Xeon as well:

```
$ cat /proc/cpuinfo
...
vendor_id       : GenuineIntel
cpu family      : 6
model           : 62
model name      : Intel(R) Xeon(R) CPU E5-2470 v2 @ 2.40GHz
stepping        : 4
microcode       : 0x42d
cpu MHz         : 2399.897
cache size      : 25600 KB
...
```

There's a GeForce RTX 2060 present in the system. Running a Python 3.9 venv with stable DGL installed with `pip install dgl dglgo -f https://data.dgl.ai/wheels/repo.html`. Happy to provide more info...

Soothysay:
Hi, I've got the same issue as well:
`File "HGCNN_2_caller.py", line 387, in <module>
    output = model(dynamic_graphs, timestamps)
[...]

PR #3510, opened 2021-11-16
Fixed issue #3459

sanchit-misra:
## Description
This PR fixes the bug reported in issue #3459.

## Checklist


VoVAllen (approved):
Thanks for the fix

src/array/cpu/spmm.h
```diff
@@ -141,7 +141,9 @@ void SpMMSumCsr(const BcastOff& bcast, const CSRMatrix& csr, NDArray ufeat,
   const bool no_libxsmm =
        bcast.use_bcast || std::is_same<DType, double>::value;
   if (!no_libxsmm) {
-    SpMMSumCsrLibxsmm<IdType, DType, Op>(bcast, csr, ufeat, efeat, out);
+    int ret = SpMMSumCsrLibxsmm<IdType, DType, Op>(bcast, csr, ufeat, efeat, out);
+    if (ret != 0)
+      SpMMSumCsrNaive<IdType, DType, Op>(bcast, csr, X, W, O);
```
BarclayII, line 146: So it will try generating a CPU kernel every time even if it failed previously?  Need to benchmark how it impacts the performance (or we still need to provide a runtime switch that turns LibXSMM off).  I'm OK with the PR for now though.

sanchit-misra: Yes, it will be good to have a runtime switch. Should we take it up later?

BarclayII: Do you think it's easy to add it?  If so I would like to have it in this PR as well.

PR #3514, opened 2021-11-17
[Feature] Add Min/max reducer in heterogeneous API for unary message functions

isratnisa:
## Description
Added support for min/max reducer in the heterogeneous graph APIs.  One can use the message_passing API `update_all` using `sum`, `max`, `min` operator directly on heterogeneous graphs. Example code snippet as follows:

`hetero_graph.update_all(fn.copy_u('h', 'm'), fn.max('m', 'y'))`   

The changes include DGL's own kernel implementation to compute the final gradients when max/min reducer is used. For homogeneous graph, gradients are computed using PyTorch's `_scatter_add` (`math: out[idx[i], *] += feat[i, *]`) operator.  In the heterogeneous API, gradients are computed on ALL source node types in one kernel. If multiple destination nodes have the same source node type, it will compute partial derivatives on the source node.

Currently, it supports  unary message functions, e.g., `copy_u`, `copy_e` on CPU.  GPU support will be added soon.

## Checklist
- [x] Code is well-documented


src/array/cpu/segment_reduce.h
```diff
@@ -101,6 +103,58 @@ void ScatterAdd(NDArray feat, NDArray idx, NDArray out) {
...
+      int src_id = dst_src_ntids[dst_id][j];
+      if (updated[src_id]) continue;
+      const DType* feat_data = list_feat[dst_id].Ptr<DType>();
+      const IdType* idx_data = list_idx[dst_id].Ptr<IdType>();
+      const IdType* idx_ntype_data = list_idx_ntypes[dst_id].Ptr<IdType>();
+      DType* out_data = list_out[src_id].Ptr<DType>();
+
+      int dim = 1;
+      for (int i = 1; i < list_out[src_id]->ndim; ++i)
+        dim *= list_out[src_id]->shape[i];
```
BarclayII, line 141: You could use `NumElements` method.

src/array/cpu/segment_reduce.h
```diff
@@ -101,6 +103,58 @@ void ScatterAdd(NDArray feat, NDArray idx, NDArray out) {
...
+void UpdateGradMinMax_hetero(HeteroGraphPtr graph,
+                       std::vector<NDArray> list_feat,
+                       std::vector<NDArray> list_idx,
+                       std::vector<NDArray> list_idx_ntypes,
+                       std::vector<NDArray> list_out) {
+  std::vector<std::vector<dgl_id_t>> dst_src_ntids(graph->NumVertexTypes(),
+    std::vector<dgl_id_t>());
+  for (dgl_type_t etype = 0; etype < graph->NumEdgeTypes(); ++etype) {
+    auto pair = graph->meta_graph()->FindEdge(etype);
+    const dgl_id_t dst_id = pair.first;  // graph is reveresed
```
BarclayII, line 124: typo

src/array/cpu/segment_reduce.h
```diff
@@ -101,6 +103,58 @@ void ScatterAdd(NDArray feat, NDArray idx, NDArray out) {
...
+      for (int i = 1; i < list_out[src_id]->ndim; ++i)
+        dim *= list_out[src_id]->shape[i];
+      int n = list_feat[dst_id]->shape[0];
+  #pragma omp parallel for
+      for (int i = 0; i < n; ++i) {
+        for (int k = 0; k < dim; ++k) {
+          if (src_id == idx_ntype_data[i * dim + k]) {
+            const int write_row = idx_data[i * dim + k];
+  #pragma omp atomic
+            out_data[write_row * dim + k] += feat_data[i * dim + k];  // feat = dZ
```
BarclayII, line 149: An optimization is to check if `feat_data[i * dim + k]` is zero before `omp atomic`, since atomic updates is quite costly.

I'm also wondering if we can get around atomic additions (I don't know though).

src/array/cpu/spmm.cc
```diff
@@ -77,6 +77,28 @@ void SpMMCsrHetero(const std::string& op, const std::string& reduce,
...
+        for (dgl_type_t etype = 0; etype < ufeat_node_tids.size(); ++etype) {
+          DType *out_off = vec_out[out_node_tids[etype]].Ptr<DType>();
+          if (reduce == "max")
+            std::fill(out_off, out_off + vec_csr[etype].num_rows * dim, cpu::op::Max<DType>::zero);
+          else
+            std::fill(out_off, out_off + vec_csr[etype].num_rows * dim, cpu::op::Min<DType>::zero);
+          const dgl_type_t dst_id = out_node_tids[etype];
+          if (!updated[dst_id]) {
+            updated[dst_id] = true;
+            // TODO(Israt): what to fill with?
```
BarclayII, line 91: Remove this comment?  Is there any problem with -1?

src/array/cpu/segment_reduce.h
```diff
@@ -101,6 +103,58 @@ void ScatterAdd(NDArray feat, NDArray idx, NDArray out) {
...
+ * \param idx The indices tensor.
+ * \param out The output tensor.
+ */
+
+template <typename IdType, typename DType>
+void UpdateGradMinMax_hetero(HeteroGraphPtr graph,
+                       std::vector<NDArray> list_feat,
+                       std::vector<NDArray> list_idx,
+                       std::vector<NDArray> list_idx_ntypes,
+                       std::vector<NDArray> list_out) {
```
BarclayII, line 119: So this only handles `copy_u` + `max/min`, right?  Just to make sure.

src/array/cpu/segment_reduce.h
```diff
@@ -101,6 +103,58 @@ void ScatterAdd(NDArray feat, NDArray idx, NDArray out) {
...
+    const dgl_id_t dst_id = pair.first;  // graph is reveresed
+    const dgl_id_t src_id = pair.second;
+    dst_src_ntids[dst_id].push_back(src_id);  // can have duplicates. Use Hashtable to optimize.
+  }
+  std::vector<bool> updated(graph->NumVertexTypes(), false);
+  for (int dst_id = 0; dst_id < dst_src_ntids.size(); ++dst_id) {
+    updated[dst_id] = false;
+    for (int j = 0; j < dst_src_ntids[dst_id].size(); ++j) {
+      int src_id = dst_src_ntids[dst_id][j];
+      if (updated[src_id]) continue;
```
BarclayII, line 133: So if a source type is already updated given the destination node type, the update will skip.  What if there are multiple edge types connecting between the same source/destination node type?  I don't think this is right but correct me if I'm wrong.

src/array/cpu/segment_reduce.h
```diff
@@ -101,6 +103,58 @@ void ScatterAdd(NDArray feat, NDArray idx, NDArray out) {
...
+    std::vector<dgl_id_t>());
+  for (dgl_type_t etype = 0; etype < graph->NumEdgeTypes(); ++etype) {
+    auto pair = graph->meta_graph()->FindEdge(etype);
+    const dgl_id_t dst_id = pair.first;  // graph is reveresed
+    const dgl_id_t src_id = pair.second;
+    dst_src_ntids[dst_id].push_back(src_id);  // can have duplicates. Use Hashtable to optimize.
+  }
+  std::vector<bool> updated(graph->NumVertexTypes(), false);
+  for (int dst_id = 0; dst_id < dst_src_ntids.size(); ++dst_id) {
+    updated[dst_id] = false;
```
BarclayII, line 130: Why only `dst_id`'s flag gets cleared?

src/array/cpu/spmm.cc
```diff
@@ -132,42 +154,42 @@ template void SpMMCsrHetero<kDLCPU, int32_t, 16>(
     const std::string& op, const std::string& reduce,
     const BcastOff& bcast, const std::vector<CSRMatrix>& csr,
     const std::vector<NDArray>& ufeat, const std::vector<NDArray>& efeat,
-    std::vector<NDArray> out, const std::vector<NDArray>& out_aux,
+    std::vector<NDArray> out, const std::vector<std::vector<NDArray>>& out_aux,
```
BarclayII, line 157: Better to use `std::vector<NDArray> *` (`&` is also OK though I prefer the former) for in/out arguments.  Otherwise it will pass the vector by value.

src/array/cpu/spmm.h
```diff
@@ -309,6 +309,93 @@ void SpMMCmpCsr(const BcastOff& bcast, const CSRMatrix& csr, NDArray ufeat,
...
+ * responsible for the computation of different nodes. \note The result will
+ * contain infinity for zero-degree nodes.
+ */
+template <typename IdType, typename DType, typename Op, typename Cmp>
+void SpMMCmpCsrHetero(const BcastOff& bcast, const CSRMatrix& csr, NDArray ufeat,
+                NDArray efeat, NDArray out, NDArray argu, NDArray arge,
+                NDArray argu_ntype, NDArray arge_etype,
+                const int ntype, const int etype) {
+  const bool has_idx = !IsNullArray(csr.data);
+  const IdType* indptr = static_cast<IdType*>(csr.indptr->data);
```
BarclayII, line 335: `csr.indptr.Ptr<IdType>` (similar for other `static_cast` usages).

src/array/cuda/segment_reduce.cu
```diff
@@ -48,6 +50,18 @@ void ScatterAdd(NDArray feat,
...
 
+template <int XPU, typename IdType, int bits>
+void UpdateGradMinMax_hetero(HeteroGraphPtr g,
+                       std::vector<NDArray> feat,
+                       std::vector<NDArray> idx,
+                       std::vector<NDArray> idx_etype,
+                       std::vector<NDArray> out) {
+  SWITCH_BITS(bits, DType, {
+    // cuda::ScatterAdd<IdType, DType>(feat, idx, out);
+  });
```
BarclayII, line 61: Raise an error.

You can also as well not write this function and use `ATEN_XPU_SWITCH` instead of `ATEN_XPU_SWITCH_CUDA` in `ScatterAddDispatchHetero`.

src/array/cuda/spmm.cu
```diff
@@ -612,17 +612,17 @@ void SpMMCsrHetero(const std::string& op, const std::string& reduce,
                 NullArray() : vec_ufeat[src_id];
             NDArray efeat = (vec_efeat.size() == 0) ?
                 NullArray() : vec_efeat[etype];
-            cuda::SpMMCsr<IdType, DType, Op, cuda::reduce::Max<IdType, DType> >(
-                bcast, csr, ufeat, efeat, vec_out[dst_id], out_aux[0], out_aux[1]);
+            // cuda::SpMMCsr<IdType, DType, Op, cuda::reduce::Max<IdType, DType> >(
+            //     bcast, csr, ufeat, efeat, vec_out[dst_id], out_aux[0], out_aux[1]);
```
BarclayII, line 616: I think it's better to remove these two blocks if they are not implemented.

src/array/kernel.cc
```diff
@@ -58,7 +58,7 @@ void SpMMHetero(const std::string& op, const std::string& reduce,
           std::vector<NDArray> ufeat_vec,
           std::vector<NDArray> efeat_vec,
           std::vector<NDArray> out,
-          std::vector<NDArray> out_aux) {
+          std::vector<std::vector<NDArray>> out_aux) {
```
BarclayII, line 61: Same.  Use const references or pointers/references.

src/array/kernel.cc
```diff
@@ -226,6 +226,23 @@ void ScatterAddDispatch(NDArray feat, NDArray idx, NDArray out) {
   });
 }
 
+/*! \brief Scatter Add (on first dimension) dispatch function on heterogeneous graph. */
+void ScatterAddDispatchHetero(HeteroGraphPtr graph,
+                        std::vector<NDArray> feat,
+                        std::vector<NDArray> idx,
+                        std::vector<NDArray> idx_etype,
+                        std::vector<NDArray> out) {
```
BarclayII, line 234: Same.

src/array/kernel.cc
```diff
@@ -226,6 +226,23 @@ void ScatterAddDispatch(NDArray feat, NDArray idx, NDArray out) {
   });
 }
 
+/*! \brief Scatter Add (on first dimension) dispatch function on heterogeneous graph. */
+void ScatterAddDispatchHetero(HeteroGraphPtr graph,
```
BarclayII, line 230: You can just write `UpdateGradMinMax_hetero` to keep it consistent with other places.

src/array/kernel.cc
```diff
@@ -333,35 +350,41 @@ DGL_REGISTER_GLOBAL("sparse._CAPI_DGLKernelSpMMHetero")
...
-    for (Value val : list_V) {
-      V_vec.push_back(val->data);
-    }
-    for (Value val : list_E) {
-      E_vec.push_back(val->data);
-    }
+
+    for (Value val : list_U) U_vec.push_back(val->data);
+    for (Value val : list_V) V_vec.push_back(val->data);
+    for (Value val : list_E) E_vec.push_back(val->data);
```
BarclayII, line 367: There should be a `ListValueToVector` function doing the same thing.

src/array/kernel.cc
```diff
@@ -440,6 +463,32 @@ DGL_REGISTER_GLOBAL("sparse._CAPI_DGLKernelScatterAdd")
...
+
+    vec_feat.reserve(list_feat.size());
+    vec_idx.reserve(list_idx.size());
+    vec_idx_etype.reserve(list_idx_etype.size());
+    vec_out.reserve(list_out.size());
+
+    for (Value val : list_feat) vec_feat.push_back(val->data);
+    for (Value val : list_idx) vec_idx.push_back(val->data);
+    for (Value val : list_idx_etype) vec_idx_etype.push_back(val->data);
+    for (Value val : list_out) vec_out.push_back(val->data);
```
BarclayII, line 486: Same.

src/array/kernel_decl.h
```diff
@@ -130,6 +130,16 @@ void ScatterAdd(NDArray feat,
                 NDArray idx,
                 NDArray out);
 
+/*!
+ * \brief Scatter Add on first dimension.
```
BarclayII, line 134: Better docstring.

tests/compute/test_new_update_all_hetero.py
```diff
@@ -58,52 +57,66 @@ def _test(mfunc, rfunc):
...
             g.multi_update_all(
-                {'plays' : (mfunc('h', 'm'), rfunc('m', 'y')),
-                'follows': (mfunc('h', 'm'), rfunc('m', 'y')),
-                'develops': (mfunc('h', 'm'), rfunc('m', 'y')),
-                'wishes': (mfunc('h', 'm'), rfunc('m', 'y'))},
-                'sum')
+                {'follows': (mfunc('h', 'm'), rfunc('m', 'y')),
+                'plays' : (mfunc('h', 'm'), rfunc('m', 'y')),
+                'wishes': (mfunc('h', 'm'), rfunc('m', 'y')),
+                'develops': (mfunc('h', 'm'), rfunc('m', 'y'))},
```
BarclayII, line 63: Is it possible to generate a more complicated graph with randomized metagraphs?  This is a very important functionality (especially with backward computation which is very hard to debug from Python) and we would like to cover as many cases as possible.

BarclayII (changes requested):

python/dgl/backend/backend.py
```diff
@@ -1675,6 +1675,33 @@ def scatter_add(x, idx, m):
     """
     pass
 
+def scatter_add_hetero(gidx, x, idx, idx_etype, dX):
```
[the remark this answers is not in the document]
isratnisa: Removed.

python/dgl/sparse.py
```diff
@@ -199,11 +199,25 @@ def _gspmm_hetero(gidx, op, reduce_op, u_len, u_and_e_tuple):
```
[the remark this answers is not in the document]
isratnisa: Added.

python/dgl/sparse.py
```diff
@@ -481,6 +503,43 @@ def _scatter_add(x, idx, m):
     return out
 
 
+def _update_grad_minmax_hetero(gidx, list_x, list_idx, list_idx_etype, list_dX):
+    r""" Scatter add operator (on first dimension) implementation.
```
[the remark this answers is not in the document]
isratnisa: Fixed.

src/array/cpu/segment_reduce.h
isratnisa: Both `copy_u` and `copy_e`

src/array/cpu/spmm.cc
isratnisa: Don't think so. Etype or node type cannot be -1.

src/array/cpu/spmm.cc
isratnisa: `&` will pass the vector by value?

src/array/cuda/spmm.cu
isratnisa: Removed for now.

src/array/kernel.cc
isratnisa: Done.

src/array/kernel.cc
isratnisa: Done

src/array/kernel.cc
isratnisa: Yes, found it. Replaced.

src/array/kernel.cc
isratnisa: Done.

src/array/kernel_decl.h
isratnisa: Updated.

src/array/cpu/segment_reduce.cc
```diff
@@ -50,6 +50,18 @@ void ScatterAdd(NDArray feat,
   });
 }
 
+/*! \brief Scatter Add on heterogeneous graph.*/
+template <int XPU, typename IdType, int bits>
+void UpdateGradMinMax_hetero(HeteroGraphPtr g,
+                std::vector<NDArray> feat,
```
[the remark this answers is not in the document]
isratnisa: Added.

src/array/cpu/segment_reduce.cc
```diff
@@ -50,6 +50,18 @@ void ScatterAdd(NDArray feat,
   });
 }
 
+/*! \brief Scatter Add on heterogeneous graph.*/
+template <int XPU, typename IdType, int bits>
+void UpdateGradMinMax_hetero(HeteroGraphPtr g,
```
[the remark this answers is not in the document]
isratnisa: Added.

src/array/cpu/segment_reduce.h
isratnisa: We store the source node type or edge-type of the max/min value for a destination type after comparing all the edge types. So, I think we need to update only once. The current unittest covers this step:
        `('user', 'plays', 'game'): ([0, 1, 2, 1], [0, 0, 1, 1]),
        ('user', 'wishes', 'game'): ([0, 1, 1], [0, 0, 1]),`

src/array/cpu/segment_reduce.h
isratnisa: Does the previous explanation clarifies this question?

src/array/cuda/segment_reduce.cu
isratnisa: Raised an error.

python/dgl/backend/pytorch/sparse.py
```diff
@@ -242,6 +246,18 @@ def backward(ctx, *dZ):
                 elif op == 'copy_lhs':
                     tpl_None = tuple([None] * len(Y))
                     dX = gspmm_hetero(g_rev, 'copy_lhs', 'sum', len(X), *tuple(dZ + tpl_None))
+            else:  # max/min
+                # TODO(Israt): dont just depend on the first etype
```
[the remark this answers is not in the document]
isratnisa: If the feature dimensions of all node types are same then checking only one etype is sufficient. Output shape is determined by `dZ[dst_id].shape[1:]` whre the `dst_id` is fetched from the first etype. If we want to support nodes with different feature length, we should use a for loop over etypes.

python/dgl/backend/pytorch/sparse.py
```diff
@@ -242,6 +246,18 @@ def backward(ctx, *dZ):
...
+                src_id, dst_id = gidx.metagraph.find_edge(0)
+                dX = tuple([th.zeros((X_shape[i][0],) + dZ[dst_id].shape[1:], dtype=dtype, device=device)
+                    if X[i] is not None else None for i in range(len(X))])
+                if op == 'mul':
+                    grad = _expand(Y, dZ.shape[1:]).gather(
+                        0, argY.long()) * dZ
+                    dX.scatter_add_(0, argX.long(), grad)
+                elif op in ['add', 'copy_lhs']:
+                    # TODO(Israt): argX.long()
```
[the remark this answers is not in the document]
isratnisa: PyTorch's version uses `argX.long()` as shown in line 256. Not sure the reason behind it and whether we need to do something similar.

src/array/cpu/segment_reduce.h
isratnisa: I see your point. I have the issue with `copy_e`. When multiple edge types share the same `src` and `dst` node type, only one etype computes the gradient instead of both. Maybe we can check the `op` to  see whether its `copy_u` or `copy_e`.

python/dgl/backend/pytorch/sparse.py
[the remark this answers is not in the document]
BarclayII: Did we check for the size in the forward function?  If so, I think this is probably not a TODO item but something like
```
# Assuming that the features are of the same dimension (enforced by the forward function)
```

python/dgl/backend/pytorch/sparse.py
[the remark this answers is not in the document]
BarclayII: PyTorch's version uses `long` because they only support 64-bit indices but not 32-bit indices.  If your `_update_grad_minmax_hetero` supports 64-bit indices then you don't need to convert it.

python/dgl/sparse.py
```diff
@@ -190,7 +190,49 @@ def _gspmm(gidx, op, reduce_op, u, e):
...
+        The binary op's name, could be ``add``, ``sub``, ``mul``, ``div``, ``copy_lhs``,
+        ``copy_rhs``.
+    reduce_op : str
+        Reduce operator, could be ``sum``, ``max``, ``min``.
+    u_len : int
+        The number of tensors in ``u`` (source node features)
+    u_and_e_tuple : Tuple of tensors
+        Tuple of source nodes' features and edges' features.
+        The feature on source nodes could be None if op is ``copy_rhs``.
+        The feature on edges, could also be None if op is ``copy_lhs``.
```
BarclayII, line 223:
```
The first few elements are source nodes' features, one for each node type.  If
``op`` is ``copy_rhs``, all of them can be None.
The rest of them are edges' features, one for each edge type.  If ``op`` is ``copy_lhs``,
all of them can be None.
```
Also, say if `op` is `copy_rhs`, we set *all* of the source node features to be None, correct?

src/array/cpu/segment_reduce.h
BarclayII: I believe the same issue will occur for `copy_u`.  Say if you have `('user', 'clicks', 'item')` and `('user', 'likes', 'item')`, you have two edge types sharing the same source node type and the same destination node type.  In this case your code will only compute the gradient through `clicks` but not `likes` for both cases.

python/dgl/backend/pytorch/sparse.py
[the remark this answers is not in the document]
isratnisa: Got it!

src/array/cpu/segment_reduce.h
BarclayII: Another question though is regarding to the following:
> If multiple edge types share the same src but different dst node types, the partial derivatives of all the dst type needs to be summed and update src node type. For example, in the case of ('user', 'plays', 'game') and ('user', 'follows, 'user'), node type user will get updates from both game and user. That's why we clear dst_id's flag updated[dst_id] = false; to get update from all dsttypes.

In this case, I still don't understand why you can skip `src_id` when two edge types share the same source type but different destination type.

Say that you have edge type `(user, clicks, item)` and `(user, likes, vendor)`, and both edge types have edges (0,0) and (0,0) (meaning that user 0 clicks item 0 and user 0 likes vendor 0).  Running a `copy_u_max` will give both item 0 and vendor 0 the same value from user 0.  In backward propagation, user 0's gradient will be the sum of item 0's gradient and vendor 0's gradient.  With your implementation though, `updated[user]` will be set to True after `clicks` relation is computed.  When it comes to the edge type `vendor`, the computation will skip because `updated[user]` is True.  So user 0's gradient will be only from item 0.  Correct?  Or am I missing something?

isratnisa: You are correct! Fixed now.

python/dgl/backend/pytorch/sparse.py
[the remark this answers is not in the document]
isratnisa: Added check in the C API (SpMMHetero) in kernel.cc

src/array/cpu/segment_reduce.h
isratnisa: You are correct. Fixed!

tests/compute/test_new_update_all_hetero.py
isratnisa: Added complexity to unittest.

src/array/cpu/segment_reduce.h
isratnisa: Addressed and fixed.

python/dgl/sparse.py
isratnisa: Clarified the comments.

src/array/cpu/segment_reduce.h
```diff
@@ -101,6 +104,86 @@ void ScatterAdd(NDArray feat, NDArray idx, NDArray out) {
...
+        if (updated[src_id]) continue;
+        const DType* feat_data = list_feat[dst_id].Ptr<DType>();
+        const IdType* idx_data = list_idx[dst_id].Ptr<IdType>();
+        const IdType* idx_ntype_data = list_idx_ntypes[dst_id].Ptr<IdType>();
+        DType* out_data = list_out[src_id].Ptr<DType>();
+        int dim = 1;
+        for (int i = 1; i < list_out[src_id]->ndim; ++i)
+          dim *= list_out[src_id]->shape[i];
+        int n = list_feat[dst_id]->shape[0];
+    #pragma omp parallel for
```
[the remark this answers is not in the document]
isratnisa: Done.

BarclayII (approved):

Issue #2278, opened 2020-10-15
Can not save the GINconv model by torch.sae


wayneowen7:
when i use the GINConv model instead of GATConv,i come cross the error when i save the model ,

AttributeError: Can't pickle local object '_gen_reduce_builtin.<locals>.func'

And i tried some other methods to save my model,like pickle or joblib,also point at the function _gen_reduce_builtin

Can't pickle <function _gen_reduce_builtin.<locals>.func at 0x7fb268179158>

i think it's the code in GINconv (        elif aggregator_type == 'mean':
            self._reducer = fn.mean),where the " fn"  actually is reducer in dgl/function,
and i find _gen_reduce_builtin in dgl/function/reducer

i can not figure out the problem ,hope some can help me, thank you so much

BarclayII:
Can you run it again with `faulthandler` enabled and post the traceback?
```
python -X faulthandler <your-script.py> <your-arguments...>
```

VoVAllen:
Hi,

Nested function is not supported by pickle. And I think it's suggested by torch using `torch.save(model.state_dict())`, can you use this instead?

wayneowen7:
I just figure out where the problem is ,than you @VoVAllen  your approch (state_dict) also can't slove my problem(i first used model.state_dict,but when i load the model,it's performance isn't the same as i tested before,and when i use summary(model) to check the parameters loaded by state_dic, i found that the parametes only include the my deeplearning model part which is upstream model before the gragh model ),but torch.save(model.state_dict()) can also be used in other model like(GCN,GAT etc)
here is my solution for GIN:

Since Gin use the aggregator ,which lead to dgl/function/reducer.py
and i suggest change the function: def _gen_reduce_builtin(reducer):
in this function,there is a line which lead the my error "Can't pickle <function _gen_reduce_builtin..func at 0x7fb268179158>
"
the line is 
    func.__name__ = reducer
i suggest just change the reducer to any string or just remove it sicne it's a  __name__ attribute
I think the type error in here lead to my problem,__name__ is a string type, and reducer is a function get by other.

HenryKenlay:
Can the above fix be considered? 

Saving and loading by state dict is of course possible but loading this way can have considerable boilerplate code if your code is flexible in datasets/model because loading by state dict requires the specific class model and all the relevant arguments (which usually depends on the dataset)

lizeyan:
I have the same problem here.

PR #3516, opened 2021-11-18
[Bug Fix] Make modules using reducers like GINConv pickleable

lizeyan:
## Description
#2278
Some modules like GINConv are not pickleable and thus, cannot be saved as a whole model, because their `__qualname__` is still the original ones (i.e., `_gen_reduce_builtin.<locals>.func`).
 

## Checklist

## Changes
- [x] To make the reduce functions generated in `dgl.function` pickleable, I set the` __name__` and `__qualname__` attributes for them.


Rhett-Ying (changes requested):
pls create PR based on master branch only.

PR #3518, opened 2021-11-18
[NN] EdgePredictor

mufeili:
## Description
NN modules for scoring edges/node pairs. After a discussion with @BarclayII , we decided to support two cases -- scoring arbitrary node pairs and scoring edges of a graph.

@VoVAllen @BarclayII I'd like you to take a look and see if you have any suggestions/questions for the interface. I will continue the rest work in the meantime. 

## Checklist
- [x] Code is well-documented


python/dgl/nn/pytorch/link.py
```diff
@@ -0,0 +1,128 @@
...
+        r"""
+
+        Description
+        -----------
+        Reinitialize learnable parameters.
+        """
+        for layer in self.layers:
+            layer.reset_parameters()
+
+    def forward(self, feat, graph=None):
```
VoVAllen, line 88: I don't like this argument list. Since the order is not consistent with other NN modules. 
One possible solution is to use `forward(self, *args, **kwargs)` like solution, and judge whether there's graph passed in. Thus we can keep it consistent with other modules and graph go first and feature go next
Would like to head from @BarclayII

VoVAllen: Also I suspect the solution I proposed above is not friendly to Torch's jit system in the future.

BarclayII: I think this is OK as long as other predictors have the same signature.

python/dgl/nn/pytorch/link.py
```diff
@@ -0,0 +1,128 @@
...
+    >>> predictor((feat_i, feat_j)).shape
+    torch.Size([3, 1])
+    """
+    def __init__(self,
+                 in_feats,
+                 hidden_feats,
+                 out_feats,
+                 num_layers=2,
+                 dropout=0.,
+                 activation=F.relu):
```
BarclayII, line 65: Why not let the users pass in his/her own MLP module instead?  Something like
```
HadamardPredictor(
    nn.Sequential(
        nn.Linear(in_feats, hid_feats),
        nn.Dropout(0.5),
        nn.Linear(hid_feats, out_feats)))
```
This way could give them more flexibility like GIN.

mufeili: In that case, I wonder if it will be better to simply have this function compute the Hadamard product. Since the MLP is applied at the end, we may not need to wrap an MLP this class.

BarclayII: That depends on what a predictor will produce.  If we say a predictor is "something that takes in the incident nodes' representations and output a single scalar for each edge", then wrapping an MLP still makes sense.

mufeili: We can rename `HadamardPredictor` to `HadamardProduct` in that case, just like `DotProduct`, `CosineDistance`, etc. Meanwhile, this is the same case as `EdgeMLP`, which concatenates pairs of node representations and passes the results to an MLP. If we do not include an MLP, is it still beneficial to have a function/class instance that concatenates pairs of node representations?

VoVAllen (commented):
I'm thinking of following dgl-ke's interface that provides two functions, forward and infer.
forward works like other nn module that takes (g, feat) as argument and infer takes (src_feat, dst_feat). 
How do you think about this?

mufeili:
> I'm thinking of following dgl-ke's interface that provides two functions, forward and infer. forward works like other nn module that takes (g, feat) as argument and infer takes (src_feat, dst_feat). How do you think about this?

I think this is less expressive than the original design. DGL-KE is designed to be primarily used as command-line tools and can have internal assumptions like the DGLGraph only has a single edge type. Such assumptions may not hold for all users and users should have the freedom to decide if they want to do one way or the other during both training and inference.

mufeili:
@jermainewang suggested benchmarking link score computation with `h_i, h_j` (w/o graph) against `g, h` (graph), in order to investigate if the former option is good enough. 

We consider a case where a batch contains 64 * 1024  links/node pairs and the feature size is 256, as in OGB baselines. 
- For the case of `h_i, h_j`, the profiling only considers `test` or `test_gpu` below.

```python
import torch

batch_size = 64 * 1024
num_nodes = 1000000
feats = 256
node_feats = torch.randn(num_nodes, feats)
src_ids = torch.randint(0, num_nodes, (batch_size,))
dst_ids = torch.randint(0, num_nodes, (batch_size,))

def test(op, node_feats, src_ids, dst_ids):
    u = node_feats[src_ids]
    v = node_feats[dst_ids]
    if op == 'dot':
        N, D = u.shape
        return torch.bmm(u.view(N, 1, D), v.view(N, D, 1)).squeeze(-1)
    elif op == 'cos':
        return F.cosine_similarity(u, v).unsqueeze(-1)
    else:
        return u * v

def test_gpu(op, node_feats, src_ids, dst_ids):
    u = node_feats[src_ids].to('cuda:0')
    v = node_feats[dst_ids].to('cuda:0')
    if op == 'dot':
        N, D = u.shape
        return torch.bmm(u.view(N, 1, D), v.view(N, D, 1)).squeeze(-1)
    elif op == 'cos':
        return F.cosine_similarity(u, v).unsqueeze(-1)
    else:
        return u * v
``` 

- For the case of `g, h`, the profiling only considers `test` or `test_gpu` below.

```python
import dgl
import dgl.function as fn
import torch

def apply_op(op, g, h):
    with g.local_scope():
        if op == 'dot':
            g.ndata['h'] = h
            g.apply_edges(fn.u_dot_v('h', 'h', 'h'))
            return g.edata['h']
        elif op == 'cos':
            g.ndata['h'] = F.normalize(h, p=2, dim=-1)
            g.apply_edges(fn.u_dot_v('h', 'h', 'h'))
            return g.edata['h']
        else:
            g.ndata['h'] = h
            g.apply_edges(fn.u_mul_v('h', 'h', 'h'))
            return g.edata['h']

num_edges = 64 * 1024
# deg 10
deg = 100
num_nodes = num_edges // deg
h = torch.randn(num_nodes, 256)

def test(op, h, num_nodes, num_edges):
    g = dgl.rand_graph(num_nodes, num_edges)
    apply_op(op, g, h)

def test_gpu(op, h, num_nodes, num_edges):
    g = dgl.rand_graph(num_nodes, num_edges)
    g = g.to('cuda:0')
    h = h.to('cuda:0')
    result = apply_op(op, g, h)
```

The result numbers are as follows.

- Time using CPU

| Op | w/o graph | graph (deg 10) | graph (deg 100) |
|-|-|-|-|
| dot | 65 ms | 24 ms | 13 ms |
| cos | 123 ms | 27 ms | 14 ms |
| ele | 77 ms | 42 ms | 32 ms |

- Time using GPU

| Op | w/o graph | graph (deg 10) | graph (deg 100) |
|-|-|-|-|
| dot | 80 ms | 15 ms | 4 ms |
| cos | 81 ms | 15 ms | 4 ms |
| ele | 80 ms | 15 ms | 4 ms |

- GPU memory usage

| Op | w/o graph | graph (deg 10) | graph (deg 100) |
|-|-|-|-|
| dot | 128 MiB | 8 MiB | 2 MiB |
| cos | 193 MiB | 14 MiB | 3 MiB |
| ele | 192 MiB | 71 MiB | 66 MiB |

Overall, it seems that using graph is still much more efficient. Also cc @VoVAllen and @BarclayII .

python/dgl/nn/pytorch/link.py
```diff
@@ -0,0 +1,170 @@
...
+        The input feature size of :math:`h_i` and :math:`h_j`. It is required
+        only if a linear layer is to be applied.
+    out_feats : int, optional
+        The output feature size. It is reuiqred only if a linear layer is to be applied.
+    bias : bool, optional
+        Whether to use bias for the linear layer if it applies.
+
+    Examples
+    --------
+    >>> import torch as th
```
VoVAllen, line 58: It's better to have an example starting from a graph (how to slice data from edge to node).

VoVAllen (approved):
Overall looks good to me

python/dgl/nn/pytorch/link.py
mufeili: Done.

Issue #3488, opened 2021-11-09
Is there a function to merge graphs？


ZhouMingjie-code:
## ❓ Questions and Help

Before proceeding, please note that we recommend
using our discussion forum (https://discuss.dgl.ai) for
general questions. As a result, this issue will
likely be CLOSED shortly.

VoVAllen:
Does dgl.batch works for you?

ZhouMingjie-code:
Thank you! But dgl.batch will reset index. It is not what i want. 

> Does dgl.batch work for you?

ZhouMingjie-code:
For example, g1 has nodes[0,1,2,3,4] and g2 has nodes[0,2,4,6,8]. I want to get a big graph with nodes[0,1,2,3,4,6,8]

ZhouMingjie-code:
For example, g1 has nodes[0,1,2,3,4] and g2 has nodes[0,2,4,6,8]. I want to get a big graph with nodes[0,1,2,3,4,6,8]

VoVAllen:
I don't think we have such function. But You can achieve this by concating the edge array of each graph.

PR #3522, opened 2021-11-19
[Feature] method for merging graphs into graphs

noncomputable:
## Description

Added method to DGLHeteroGraph to update a graph with another graph, analogous to a dictionary update.

Issue #3488 asked if it was possible to merge a graph into another graph in DGL. @VoVAllen suggested such a method does not exist and @BarclayII assigned the issue a feature request label, so I suggest they review.

## Checklist
- [x] Code is well-documented


BarclayII:
I think a more flexible and efficient API would be something like `dgl.merge()` function that takes in a list of graphs and returns a single union.  With this API, if you want to merge multiple graphs, you will need to call `update()` repeatedly.

What do you think?

noncomputable:
> I think a more flexible and efficient API would be something like `dgl.merge()` function that takes in a list of graphs and returns a single union. With this API, if you want to merge multiple graphs, you will need to call `update()` repeatedly.
> 
> What do you think?

I thought it would make sense for graphs to have a method analogous to `dict.update`, but I agree with you. I'll switch it to a `dgl.merge` function. How do you feel about a `DGLHeteroGraph.update()` method that wraps `dgl.merge()`?

noncomputable:
@BarclayII I've implemented the suggested changes.

python/dgl/merge.py
```diff
@@ -0,0 +1,103 @@
...
+    # Merge edges and edge data.
+    for etype in etypes:
+        merged_us = F.zeros(0, ref.idtype, ref.device)
+        merged_vs = F.zeros(0, ref.idtype, ref.device)
+        edata_frames = []
+        for graph in graphs:
+            etype_id = graph.get_etype_id(etype)
+            us, vs = graph.edges(etype=etype)
+            merged_us = F.cat([merged_us, us], dim=0)
+            merged_vs = F.cat([merged_vs, vs], dim=0)
```
BarclayII, line 76: It's best not to do recursive concatenation like:
```python
for x in xs:
    y = F.cat([y, x])
```
because concatenation is not an in-place operation and the above code will generate a lot of intermediate tensors and a long computation graph.

Instead, try gathering the small tensors in a list and call `F.cat` once.

python/dgl/merge.py
```diff
@@ -0,0 +1,103 @@
...
+from . import backend as F
+from .base import DGLError
+
+__all__ = ['merge']
+
+def merge(graphs):
+    r"""Merge a sequence of graphs together into a single graph.
+
+    Nodes and edges that exist in ``graphs[i+1]`` but not in ``dgl.merge(graphs[0:i+1])``
+    will be added to ``dgl.merge(graphs[0:i+1])`` along with their data.
```
BarclayII, line 13: When will we merge graphs with different number of nodes?  I think restricting them to have the same number of nodes will make things easier, and will also match the semantics of graph union (i.e. two graphs having the same nodes has the edge sets unioned).

BarclayII (approved):
There's still space to optimize but I'll approve this for now.  The interface looks reasonable to me.

python/dgl/merge.py
noncomputable: The feature request that prompted this PR specifically gave an example of merging graphs with different numbers of nodes, so I think it's appropriate for the operation to accommodate that use case.

python/dgl/merge.py
noncomputable: Thank you, good catch! I’ve implemented this change.

PR #3530, opened 2021-11-21
[NN] TransE and TransR

mufeili:
## Description
@VoVAllen This PR adds `TransE` and `TransR` for scoring triples in the scenario of heterogeneous graphs. I'd like you to take a look first and see if we can reach an agreement on the interface design. After that, I can add doc and unit tests.

It will be great if @classicsong can also take a look.

Two points that we may need to discuss:
- Whether we should add the loss function as a class method:
    - Pros: Allow people to reuse the implementation
    - Cons: Inconsistent with other score/distance functions like DotProduct, HadamardProduct, etc.
- Whether we should allow taking `graph` as an optional argument:
    - Pros: Consistent with the case for homogeneous graphs
    - Cons: Relation specific projection can make graph-based computation rather costly  

## Checklist
- [ ] Code is well-documented


jermainewang:
I'm not sure about what you mean in the first issue.

For the second one, I tried to prototype it a little bit. Suppose we have a node representation module `encoder` and wish to use the TransE/TransR for computing the score. Will the code look like this?

```python
class MyModel(nn.Module):
    def __init__(self, ...):
        self.encoder = MyNodeRepreModel(...)
        self.score_func = dgl.nn.TransE(num_rels, embed_size)
    def forward(self, g, nfeat):
        # g is a homogeneous graph with relation type stored in the edata
        h_n = self.encoder(g, nfeat)
        head, tail = g.edges()
        h_head, h_tail, rels = h_n[head], h_n[tail], g.edata['rels']
        return self.score_func(h_head, h_tail, rels)
```

Having `g` as the argument will probably make the code appear like this:

```python
class MyModel(nn.Module):
    def __init__(self, ...):
        self.encoder = MyNodeRepreModel(...)
        self.score_func = dgl.nn.TransE(num_rels, embed_size)
    def forward(self, g, nfeat):
        # g is a homogeneous graph with relation type stored in the edata
        h_n = self.encoder(g, nfeat)
        return self.score_func(g, h_n)
```

The later one has fewer LoC, but it imposes some implicit assumptions on the input graph and how the relation information is stored (e.g., in edata). It perhaps is also quite different than what most ppl would expect.

mufeili:
> I'm not sure about what you mean in the first issue.

The question is whether we should add the loss function as a class method of `TransR`/`TransE`, which should take both positive and negative samples, score them and return the loss value based on the scores. The forward function of `TransR`/`TransE` just scores triples. I had a brief discussion with @VoVAllen about this and our conclusion is to go with the current design first and improve the interface if necessary when we work on DGL-Enter.

> The later one has fewer LoC, but it imposes some implicit assumptions on the input graph and how the relation information is stored (e.g., in edata). It perhaps is also quite different than what most ppl would expect.

Yes, for the PR #3518 on `ElementwisePredictor` (I renamed it from HadamardPredictor to make it more accessible) and `ConcatPredictor`, we had an optional argument `g` so that users can choose to compute over arbitrary node pairs or nodes connected in a graph. The order `(h, g)` is unfortunately inconsistent with our previous default argument order `(g, h)`.

For the score functions in knowledge graph embedding, `g` is less preferred as score functions like `TransR` involve relation-specific projection M_r * h and it can be tedious and costly to work with `g.ndata` rather than `g.srcdata`/`g.dstdata`.

mufeili:
@jermainewang Previously you suggested using a single module to integrate multiple KE score functions in an offline chat. I think this is a good idea for homogeneous score functions like elementwise product, dot product, etc. For KE score functions, they can be much more complex and each score function is associated with a paper. Therefore, I think it's still better to have independent modules for KE score functions.

python/dgl/nn/pytorch/link.py
```diff
@@ -173,3 +173,204 @@ def forward(self, h_src, h_dst):
...
+
+        Returns
+        -------
+        torch.Tensor
+            The triple scores. The tensor is of shape :math:`(E)`.
+        """
+        h_rel = self.rel_emb(rels)
+        proj_rel = self.rel_project(rels).reshape(-1, self.nfeats, self.rfeats)
+        h_head = torch.einsum('ab,abc->ac', h_head, proj_rel)
+        h_tail = torch.einsum('ab,abc->ac', h_tail, proj_rel)
```
jermainewang, line 374: Is this a matmul between `(E, D1) @ (E, D1, D2)`? You could utilize torch's batched matmul.
```python
(h_head.unsqueeze(1) @ proj_rel).squeeze(1)
```
Probably need to benchmark which implementation is faster.

mufeili: Thanks for the advice. The PyTorch team mentioned that currently einsum goes through bmm most of the time. I tested a few cases. BMM brings some gain in µs for small batch size and feature size. The differences become negligible for larger batch size and feature size. I've updated the implementation with bmm.

VoVAllen (approved):
LGTM. Better to break functions into several files which makes it easier to maintain, but not critical.

mufeili:
@VoVAllen Done

VoVAllen (approved):

mufeili:
@VoVAllen It seems that the number of times of retrying connecting to remote server with TensorPipe can be >100. This doesn't look correct.

VoVAllen:
@mufeili Got. The testing of distributed tasks seems unstable

PR #3566, opened 2021-12-04
[Feature] Add CUDA support for `min` and `max` reducer in heterogeneous API for unary message functions

isratnisa:
## Description
To update the node features of a graph, DGL uses the message passing API `update_all`. This PR adds  CUDA support for `min` and `max` reducer in addition to `sum` reducer in the `update_all` API for heterogeneous graphs. This PR extends [PR 3514](https://github.com/dmlc/dgl/pull/3514) by adding GPU counterparts for the required functionalities. Currently, it supports unary message functions, e.g., `copy_u`, `copy_e` on both CPU and GPU platforms. 

This PR implements a  CUDA kernel `UpdateGradMinMax_hetero` to support the backward operator of `update_all` for heterogeneous graphs. The homogeneous counterpart uses PyTorch's `scatter_Add` function for this purpose.  We compare the performance of the backward operators on NVIDIA Tesla T4. The experiments are performed on randomly generated heterogeneous graph with 5 relation types where each relation type has 5K nodes and 25K edges.

```
// uses index of torch.int32
multi_update_all (uses PyTorch's scatter add): 1.57ms
update_all (uses UpdateGradMinMax_hetero): 1.15 ms
```
```
// uses index of torch.int64
multi_update_all (uses PyTorch's scatter add): .89 ms
update_all (uses UpdateGradMinMax_hetero): .99 ms
```

## Checklist
- [x] Code is well-documented
- [ x] To the best of my knowledge, examples are either not affected by this change,


BarclayII:
@nv-dlasalle Could you also help take a look at the CUDA kernels if you have time?  The backward function (i.e. `UpdateGradMinMax_hetero` in `segment_reduce.cuh`) currently makes heavy usage on atomic additions (which is not deterministic) but I'm not sure what is the best way to do those computation.

src/array/cuda/segment_reduce.cuh
```diff
@@ -63,6 +63,27 @@ __global__ void ScatterAddKernel(
...
+template <typename IdType, typename DType>
+__global__ void UpdateGradMinMaxHeteroKernel(
+    const DType *feat, const IdType *idx, const IdType *idx_type, DType *out,
+    int64_t n, int64_t dim, int type) {
+  for (int row = blockIdx.x; row < n; row += gridDim.x) {
+    int col = blockIdx.y * blockDim.x + threadIdx.x;
+    while (col < dim) {
+      if (type == idx_type[row * dim + col]) {
+        const int write_row = idx[row * dim + col];
+        cuda::AtomicAdd(out + write_row * dim + col, feat[row * dim + col]);
```
nv-dlasalle, line 80: This atomic operation is necessary because `idx` may contain duplicates (because the one node can have multiple edges connected)?

src/array/cuda/segment_reduce.cuh
```diff
@@ -63,6 +63,27 @@ __global__ void ScatterAddKernel(
...
+ */
+template <typename IdType, typename DType>
+__global__ void UpdateGradMinMaxHeteroKernel(
+    const DType *feat, const IdType *idx, const IdType *idx_type, DType *out,
+    int64_t n, int64_t dim, int type) {
+  for (int row = blockIdx.x; row < n; row += gridDim.x) {
+    int col = blockIdx.y * blockDim.x + threadIdx.x;
+    while (col < dim) {
+      if (type == idx_type[row * dim + col]) {
+        const int write_row = idx[row * dim + col];
```
nv-dlasalle, line 79: Why do we have a two dimensional index that maps to one dimensional index? Is each row of `idx` identical?

src/array/cuda/segment_reduce.cuh
```diff
@@ -63,6 +63,27 @@ __global__ void ScatterAddKernel(
...
+ *       to a target row in output tensor.
+ */
+template <typename IdType, typename DType>
+__global__ void UpdateGradMinMaxHeteroKernel(
+    const DType *feat, const IdType *idx, const IdType *idx_type, DType *out,
+    int64_t n, int64_t dim, int type) {
+  for (int row = blockIdx.x; row < n; row += gridDim.x) {
+    int col = blockIdx.y * blockDim.x + threadIdx.x;
+    while (col < dim) {
+      if (type == idx_type[row * dim + col]) {
```
nv-dlasalle, line 78: Is it possible for different columns to have different types?

src/array/cuda/segment_reduce.cuh
```diff
@@ -155,6 +176,59 @@ void ScatterAdd(
...
+      const IdType* idx_type_data = list_idx_types[dst_ntype].Ptr<IdType>();
+      int type = (op == "copy_lhs") ? src_ntype : etype;
+      DType* out_data = (*list_out)[type].Ptr<DType>();
+      int dim = 1;
+      for (int i = 1; i < (*list_out)[type]->ndim; ++i)
+        dim *= (*list_out)[type]->shape[i];
+      int n = list_feat[dst_ntype]->shape[0];
+      auto *thr_entry = runtime::CUDAThreadEntry::ThreadLocal();
+      const int nbx = FindNumBlocks<'x'>(n);
+      const int ntx = FindNumThreads(dim);
```
nv-dlasalle, line 219: We may need to handle the case where `dim` < 64 specially, as otherwise we'll get very poor performance.

src/array/cuda/segment_reduce.cuh
isratnisa: Right!

src/array/cuda/segment_reduce.cuh
isratnisa: Again I wanted to write a similar code to existing DGL codes with 2D thread mapping. Pushed a new commit with 1D mapping.

src/array/cuda/segment_reduce.cuh
isratnisa: I think the new commit will address this issue too.

nv-dlasalle (approved):
In general I think this looks good. @BarclayII if the non-determinism of the atomic is a blocker, it can probably be accomplished via sorting. If its acceptable as is, then I'd be in favor of a follow-up PR to see if we achieve better performance via sorting to avoid the atomic.

src/array/cpu/segment_reduce.h
```diff
@@ -117,61 +118,35 @@ void UpdateGradMinMax_hetero(HeteroGraphPtr graph,
...
-      const dgl_id_t dst_id = pair.first;  // graph is reversed
-      const dgl_id_t src_id = pair.second;
-      const DType* feat_data = list_feat[dst_id].Ptr<DType>();
-      const IdType* idx_data = list_idx[dst_id].Ptr<IdType>();
-      const IdType* idx_ntype_data = list_idx_ntypes[dst_id].Ptr<IdType>();
-      DType* out_data = (*list_out)[etid].Ptr<DType>();
+      const dgl_id_t dst_ntype = pair.first;  // graph is reversed
+      const dgl_id_t src_ntype = pair.second;
+      auto same_src_dst_ntype = std::find(begin(src_dst_ntypes[dst_ntype]),
+        end(src_dst_ntypes[dst_ntype]), src_ntype);
```
BarclayII, line 132: Where is `begin` and `end` defined?  Somewhere must `using`'d `std::begin` and `std::end`, which is bad since `begin` and `end` alone can very well be used as variable names.  I prefer explicitly use `std::begin` and `std::end` (and find and remove the `using std::begin` and `using std::end` statements).

BarclayII (approved):
LGTM.  Minor comment.

src/array/cpu/segment_reduce.h
isratnisa: Agreed. Changed in CPU and GPU version.

PR #3567, opened 2021-12-04
[PinSAGESampler] support PinSAGE sampler on GPU

lixiaobai09:
## Description

Support PinSAGE sampler running on GPU, which includes "randomwalk with restart" algorithm and "_select_pinsage_neighbors".

## Checklist
- [x] Code is well-documented


src/graph/sampling/randomwalks/atomic.cuh
```diff
@@ -0,0 +1,201 @@
+/*!
```
BarclayII, line 1: You don't need to duplicate the file; either `#include "../../array/cuda/atomic.cuh"` or move the `array/cuda/atomic.cuh` file into `include` directory if you don't like including with double quotes (although that will require changing some other places).

src/graph/sampling/randomwalks/frequency_hashmap.cu
```diff
@@ -0,0 +1,439 @@
+/*!
+ *  Copyright (c) 2018 by Contributors
+ * \file graph/sampling/frequency_hashmap.cu
+ * \brief frequency hashmap - used to select top-k frequency edges of each node
+ */
```
BarclayII, line 5: We have a `OrderedHashTable` class in `src/runtime/cuda/cuda_hashtable.cuh`.  Could you see if you can utilize that?  My intuition is that you can subclass it and add a new select-top-k interface.

src/graph/sampling/randomwalks/get_node_types_gpu.cu
```diff
@@ -0,0 +1,90 @@
...
+
+template<DLDeviceType XPU, typename IdxType>
+TypeArray GetNodeTypesFromMetapath(
+    const HeteroGraphPtr hg,
+    const TypeArray metapath) {
+
+  uint64_t num_etypes = metapath->shape[0];
+  TypeArray result = TypeArray::Empty(
+      {metapath->shape[0] + 1}, metapath->dtype, metapath->ctx);
+  IdxType *result_data = static_cast<IdxType *>(result->data);
```
BarclayII, line 31: `static_cast<T *>(array->data)` could be replaced with `array.Ptr<T>()` which is more concise.

src/graph/sampling/randomwalks/get_node_types_gpu.cu
```diff
@@ -0,0 +1,90 @@
...
+  auto h_metapath_data = static_cast<IdxType*>(
+      cpu_device->AllocWorkspace(cpu_ctx, sizeof(IdxType) * (num_etypes)));
+  auto h_result_data = static_cast<IdxType*>(
+      cpu_device->AllocWorkspace(cpu_ctx, sizeof(IdxType) * (num_etypes + 1)));
+
+  metapath_device->CopyDataFromTo(static_cast<const IdxType*>(metapath->data), 0,
+                                  h_metapath_data, 0,
+                                  sizeof(IdxType) * (num_etypes),
+                                  metapath_ctx, cpu_ctx,
+                                  metapath->dtype, stream);
```
BarclayII, line 49: There is an `NDArray::CopyTo(DLContext)` method esp. for default stream copies.

Also, it might be easier if you restrict the metapath array on CPU.

src/graph/sampling/randomwalks/get_node_types_gpu.cu
```diff
@@ -0,0 +1,90 @@
...
+      LOG(FATAL) << "source of edge type #" << i <<
+        " does not match destination of edge type #" << i - 1;
+    }
+    curr_type = dsttype;
+    h_result_data[i + 1] = dsttype;
+  }
+  metapath_device->CopyDataFromTo(h_result_data, 0, result_data, 0,
+                                  sizeof(IdxType) * (num_etypes + 1),
+                                  cpu_ctx, metapath_ctx,
+                                  metapath->dtype, stream);
```
BarclayII, line 69: Same.

src/graph/sampling/randomwalks/get_node_types_gpu.cu
```diff
@@ -0,0 +1,90 @@
...
+  // release the data
+  metapath_device->StreamSync(metapath_ctx, stream);
```
BarclayII, line 71: Is this necessary for default streams?

src/graph/sampling/randomwalks/randomwalks.cc
```diff
@@ -93,7 +93,7 @@ std::tuple<IdArray, IdArray, TypeArray> RandomWalkWithStepwiseRestart(
 
   TypeArray vtypes;
   std::pair<IdArray, IdArray> result;
-  ATEN_XPU_SWITCH(hg->Context().device_type, XPU, "RandomWalkWithStepwiseRestart", {
+  ATEN_XPU_SWITCH_CUDA(hg->Context().device_type, XPU, "RandomWalkWithStepwiseRestart", {
```
BarclayII, line 96: Could you also enable CUDA support for `RandomWalk` and `RandomWalkWithRestart`?  The former could be done with skipping the `restart_prob` logic in `RandomWalkWithStepwiseRestart` and the latter could be done with an array with the same value I think.

BarclayII (commented):
Really appreciate it!

src/graph/sampling/randomwalks/get_node_types_gpu.cu
nv-dlasalle: If we are to allow this to happen on a non-default stream, we need a `metapath_device->StreamSync(metapath_ctx, stream);` before accessing the data on the CPU.

nv-dlasalle (approved):
Once @BarclayII's and my comments are addressed, looks good to me.

@lixiaobai09 do you have any performance results?

lixiaobai09:
> Once @BarclayII's and my comments are addressed, looks good to me.
> 
> @lixiaobai09 do you have any performance results?

Test environment:
2 * E5-2650 v4, 256GB of DRAM, V100-PCIE-32GB

I use this script to get the performance results. 
```python
import time
import dgl
import torch

num_nodes = 1000000
src = [(i // 3) for i in range(3 * num_nodes)]
dst = [(i // 3 + 1 + i - i // 3 * 3) for i in range(3 * num_nodes)]
# src = [0, 0, 0, 1, 1, 1, 2, 2, 2, 3, 3, 3, ...]
# dst = [1, 2, 3, 2, 3, 4, 3, 4, 5, 5, 6, 7, ...]

def sampling_perf_test(device, dtype, sampling_seeds):
    # print(f'device: {device}, sampling_seeds: {sampling_seeds}')
    if (dtype == torch.int64):
        G = dgl.graph((src, dst)).long().to(device)
    elif (dtype == torch.int32):
        G = dgl.graph((src, dst)).int().to(device)
    sampler = dgl.sampling.RandomWalkNeighborSampler(G, 3, 0.5, 200, 10)
    seeds = torch.tensor([i for i in range(sampling_seeds)], dtype=dtype).to(device)
    repeat = 10
    # print('graph edges: {}'.format(G.num_edges()))
    t1 = time.time()
    for _ in range(repeat):
        frontier = sampler(seeds)
        # print('frontier edges: {}'.format(frontier.num_edges()))
    time_cost = (time.time() - t1)
    # frontier.all_edges(form='uv')
    # print('time cost 64bit: {:.5f}'.format(time_cost / repeat))
    print('{:.5f}'.format(time_cost / repeat))


if __name__ == '__main__':
    devices = ['cpu', 'cuda:0']
    dtypes = [torch.int32, torch.int64]
    sampling_sizes = [100, 1000, 10000, 100000]
    for device in devices:
        for dtype in dtypes:
            print(f'device:{device}, dtype:{dtype}')
            for sampling_size in sampling_sizes:
                sampling_perf_test(device, dtype, sampling_size)
```

And here is the result:
Snipaste

lixiaobai09:
> Could you also enable CUDA support for `RandomWalk` and `RandomWalkWithRestart`? The former could be done with skipping the `restart_prob` logic in `RandomWalkWithStepwiseRestart` and the latter could be done with an array with the same value I think.

I implement the CUDA-supported 'RandomWalk' and 'RandomWalkWithRestart' only with uniform choice, the non-uniform choice with 'prob' is a little complex. And I think the non-uniform choice of each step is heavy for each GPU thread, maybe it is not suitable for GPU. Do you have any suggestions about CUDA-supported non-uniform choice with 'prob'?

BarclayII:
> > Could you also enable CUDA support for `RandomWalk` and `RandomWalkWithRestart`? The former could be done with skipping the `restart_prob` logic in `RandomWalkWithStepwiseRestart` and the latter could be done with an array with the same value I think.
> 
> I implement the CUDA-supported 'RandomWalk' and 'RandomWalkWithRestart' only with uniform choice, the non-uniform choice with 'prob' is a little complex. And I think the non-uniform choice of each step is heavy for each GPU thread, maybe it is not suitable for GPU. Do you have any suggestions about CUDA-supported non-uniform choice with 'prob'?

Uniform is good enough.  Thanks a lot!

BarclayII:
Thanks again!  We will leave non-default streams to future.

zilinly:
Howablout the packages version？

PR #3572, opened 2021-12-06
The revision of model HAN

songshuhan:
Hello, I read the code for the HAN model，found the method of realizing Semantic-level Attention is different from that of the paper. I have made a simple modification, you can see if it is more consistent with the statement in the paper.

### DIFF
--- a/examples/pytorch/han/model_hetero.py
+++ b/examples/pytorch/han/model_hetero.py
@@ -25,9 +25,13 @@ class SemanticAttention(nn.Module):
         )
 
     def forward(self, z):
-        w = self.project(z).mean(0)                    # (M, 1)
-        beta = torch.softmax(w, dim=0)                 # (M, 1)
-        beta = beta.expand((z.shape[0],) + beta.shape) # (N, M, 1)
+        #w = self.project(z).mean(0)                    # (M, 1)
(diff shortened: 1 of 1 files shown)


examples/pytorch/han/model_hetero.py
```diff
@@ -25,9 +25,13 @@ def __init__(self, in_size, hidden_size=128):
...
     def forward(self, z):
-        w = self.project(z).mean(0)                    # (M, 1)
-        beta = torch.softmax(w, dim=0)                 # (M, 1)
-        beta = beta.expand((z.shape[0],) + beta.shape) # (N, M, 1)
+        #w = self.project(z).mean(0)                    # (M, 1)
+        #beta = torch.softmax(w, dim=0)                 # (M, 1)
+        #beta = beta.expand((z.shape[0],) + beta.shape) # (N, M, 1)
+
+        w = self.project(z)                         # (N,M,1)
+        beta = torch.softmax(w, dim=1)              # (N,M,1)
```
BarclayII, line 33: I think this is not correct.  According to equation 7, the importance of metapath is computed by averaging the projected metapath-specific representations `z_i^{\phi_p}` over all the *nodes*.  That's the reason we have `w` as an Mx1 vector.

jermainewang:
Closed due to lack of activity.

PR #3574, opened 2021-12-07
[Feature][Dist] change TP::Receiver/TP::Sender for multiple connections

Rhett-Ying:
## Description

improve sender/receiver to wait in non-blocking way

This is the 1st part of support long live graph server in dist so that multiple client groups could join/leave freely.
Basically, here's my schedule:

This PR:
- [x] change C++_receiver to wait in non-blocking way and always be able to accept new connections from senders.
- [x] change C++_sender to be able to connect/disconnect to receiver one by one while the current one works in a blocking way.

Future PR:
- [ ] attach group_id/client_id for each request msg in python level so that server could manage clients into groups.
- [ ] change launch script to support long live server and multiple client groups.

## Checklist
- [ ] Code is well-documented


python/dgl/distributed/rpc_client.py
```diff
@@ -161,17 +162,17 @@ def connect_to_server(ip_config, num_servers, max_queue_size=MAX_QUEUE_SIZE, net
...
         server_port = addr[2]
-        rpc.add_receiver_addr(server_ip, server_port, server_id)
-    rpc.sender_connect()
+        while not rpc.connect_receiver(server_ip, server_port, server_id):
+            time.sleep(1)
     # Get local usable IP address and port
     ip_addr = get_local_usable_addr(server_ip)
     client_ip, client_port = ip_addr.split(':')
+    # wait server connect back
+    rpc.receiver_wait(client_ip, client_port, num_servers, blocking=False)
```
VoVAllen, line 171: Will it cause problem if it's not blocking here? Thus the message might be sent out below before the remote sender connect to local receiver?

tests/distributed/utils.py
```diff
@@ -24,4 +25,11 @@ def get_local_usable_addr():
     port = sock.getsockname()[1]
     sock.close()
 
-    return ip_addr + ' ' + str(port)
\ No newline at end of file
+    return ip_addr + ' ' + str(port)
+
+
+def reset_envs():
```
VoVAllen, line 31: What's this for?

VoVAllen (approved):
Overall looks good to me

tests/distributed/utils.py
Rhett-Ying: we set env in some tests and such kind of env is global and will affect following tests such as `os.environ['DGL_NUM_SAMPLER']`. we boot the client in the main process and such kind of env will be read. For example, if I run `test_distributed_sampling.py` then `test_mp_dataloader.py`, it  works fine.  If I change the order, it failed as more clients are booted due to `DGL_NUM_SAMPLER`.

python/dgl/distributed/rpc_client.py
Rhett-Ying: it's ok. because sender is independent of receiver for the same client. it's ok to send message before remote sender connects to local receiver. The remote sender sends messages only after connects to local receiver. And local receiver will be blocked in rpc.recv_response() until remote sender connects to local receiver and send messges.

VoVAllen:
I read the pytorch code about how to continuously accept new clients. They did the new accept in the callback functions instead of creating a new thread. Not sure about which one is better but you can take a look at it. 

https://github.com/pytorch/pytorch/blob/7b376bf844e88a8583d88772c30797ac4215c7a4/torch/csrc/distributed/rpc/tensorpipe_agent.cpp#L485-L505

Rhett-Ying:
> I read the pytorch code about how to continuously accept new clients. They did the new accept in the callback functions instead of creating a new thread. Not sure about which one is better but you can take a look at it.
> 
> https://github.com/pytorch/pytorch/blob/7b376bf844e88a8583d88772c30797ac4215c7a4/torch/csrc/distributed/rpc/tensorpipe_agent.cpp#L475-L509

I followed the way used in `pytorch` which is cleaner and consistent with other calls.

PR #3599, opened 2021-12-19
[Feature] Negative sampling

BarclayII:
## Description
This provides a `dgl.dataloading.negative_sampler.GlobalUniform` which generates true source-destination node pairs: it is guaranteed that the negative examples will not have edges in between, unlike `dgl.dataloading.negative_sampler.Uniform`.  This should be beneficial for negative sampling on dense graphs.

## Checklist
- [x] Code is well-documented


BarclayII:
@zjost Perhaps you might be interested; you can find the interface in `python/dgl/dataloading/negative_sampler.py`.

python/dgl/dataloading/negative_sampler.py
```diff
@@ -65,3 +66,64 @@ def _generate(self, g, eids, canonical_etype):
...
+        before finding the unique pairs.
+
+        Only effective if :attr:`unique` is True.
+
+        Increasing it will increase the likelihood of getting :attr:`k` negative examples,
+        but will also take more time and memory.
+
+    Notes
+    -----
+    This negative sampler may not always return :attr:`k` negative samples.  Consider
```
zjost, line 108: Might be worth a brief explanation as to why this might happen--e.g., a small graph generates many redundant edges when random sampling and/or the graph is so dense that not many unique negative edges exist, but that this shouldn't be the case for medium sized graphs that are relatively sparse.

zjost (commented):
Happy to see this.  Adding a few comments.

python/dgl/dataloading/negative_sampler.py
```diff
@@ -65,3 +66,64 @@ def _generate(self, g, eids, canonical_etype):
...
+        Whether to exclude self-loops from negative examples.  (Default: False)
+    unique : bool, optional
+        Whether to sample unique negative examples.  Setting it to False will make things
+        faster.  (Default: True)
+    num_trials : int, optional
+        The number of rejection sampling trials.
+
+        Increasing it will increase the likelihood of getting :attr:`k` negative examples,
+        but will also take more time if the graph is dense.
+    redundancy : float, optional
```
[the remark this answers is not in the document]
BarclayII: Added.

python/dgl/dataloading/negative_sampler.py
```diff
@@ -65,3 +66,64 @@ def _generate(self, g, eids, canonical_etype):
...
+    all the nodes of type ``srctype`` and ``v'`` is chosen uniformly from all the nodes
+    of type ``dsttype``.  The resulting edges will also have type
+    ``(srctype, etype, dsttype)``.
+
+    Parameters
+    ----------
+    k : int
+        The number of negative examples to generate per minibatch.
+    exclude_self_loops : bool, optional
+        Whether to exclude self-loops from negative examples.  (Default: False)
```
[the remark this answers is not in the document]
BarclayII: I think it should be default to True.

python/dgl/dataloading/negative_sampler.py
BarclayII: Good point.

jermainewang (commented):
The overall question I have is how a user will use the negative samples computed by the function?   The major concern is the number of negative samples may not be equal to the given `num_samples`, which is quite counter-intuitive IMO.

python/dgl/sampling/negative.py
```diff
@@ -0,0 +1,72 @@
...
+    and the latter in the range of ``[0, num_dst)``, where ``num_src`` and ``num_dst``
+    represents the number of nodes with the source and destination node type respectively.
+    It guarantees that no edge will exist between the corresponding pairs of ``src``
+    with the source node type and ``dst`` with the destination node type.
+
+    .. note::
+
+       This function uses rejection sampling, and may not always return the same number
+       of negative examples as the given :attr:`num_samples`.  This is more likely to
+       happen when a graph is so small or dense that not many unique negative examples exist.
```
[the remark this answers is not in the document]
BarclayII: No.  If the graph is very dense, with rejection sampling you may get less than k negative examples even if the number of possible negative examples is larger than k.  Say `|V|=20, |E|=300`, with 3 trials of rejection sampling you still have 42% probability of failure for each negative example generation, irrespective of the value of `k`.

python/dgl/sampling/negative.py
```diff
@@ -0,0 +1,72 @@
...
+    ----------
+    g : DGLGraph
+        The graph.
+    num_samples : int
+        The number of negative examples to generate.
+    exclude_self_loops : bool, optional
+        Whether to exclude self-loops from the negative examples.  Only impacts the
+        edge types whose source and destination node types are the same.
+
+        Default: True.
```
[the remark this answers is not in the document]
BarclayII: Could be.  Say you compute the score with a dot product between the representation of two entities, a self-loop as a negative example will essentially gear the representation's magnitude towards 0, which might be undesirable.  You could argue that this is a regularization technique, but I would rather leave this as an option just in case people would like to turn this off.

@zjost you could also add your thoughts on this.

python/dgl/sampling/negative.py
```diff
@@ -0,0 +1,72 @@
...
+    unique : bool, optional
+        Whether to sample unique negative examples.  Setting it to False will make things
+        faster.  (Default: True)
+    etype : str or tuple of str, optional
+        The edge type.  Can be omitted if the graph only has one edge type.
```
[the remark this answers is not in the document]
BarclayII: I don't think this is a common use case as the task will be a lot easier than negative examples with the same types; most of the time you will only need to know which types every node belongs to.

@zjost If you agree with Minjie then I'll consider adding a homogenized negative sampler later.

python/dgl/sampling/negative.py
```diff
@@ -0,0 +1,72 @@
...
+    num_trials : int, optional
+        The number of rejection sampling trials.
+    redundancy : float, optional
+        Indicates how much more negative examples to actually generate during rejection sampling
+        before finding the unique pairs.
+
+        Increasing it will increase the likelihood of getting :attr:`k` negative examples,
```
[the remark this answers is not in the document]
BarclayII: The reason why this argument exists relates to the implementation detail.  The negative sampling works like this: first generate a bunch of pairs with rejection sampling *in parallel*, then we take a unique across them if required.  `num_trials` in this case only tries to prevent generation of positive examples, but not duplicate examples.  So one might still have less than `k` negative pairs even if you try many times.  PyG also has this concept, except that it's a hardcoded constant.

python/dgl/sampling/negative.py
```diff
@@ -0,0 +1,72 @@
...
+        edge types whose source and destination node types are the same.
+
+        Default: True.
+    unique : bool, optional
+        Whether to sample unique negative examples.  Setting it to False will make things
+        faster.  (Default: True)
+    etype : str or tuple of str, optional
+        The edge type.  Can be omitted if the graph only has one edge type.
+    num_trials : int, optional
+        The number of rejection sampling trials.
```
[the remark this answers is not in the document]
BarclayII: Added.

python/dgl/sampling/negative.py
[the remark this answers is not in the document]
BarclayII: Replaced `k` with `num_samples`.

include/dgl/runtime/parallel_for.h
```diff
@@ -110,6 +111,61 @@ void parallel_for(
...
+
+/*!
+ * \brief OpenMP-based two-stage parallel reduction.
+ *
+ * The first-stage reduction function \a f works in parallel.  Each thread's workload has
+ * at least \a grain_size elements.  The loop body will be a function that takes in
+ * the starting index, the ending index, and the reduction identity.
+ *
+ * The second-stage reduction function \a sf is a binary function working in the main
+ * thread. It aggregates the partially reduced result computed from each thread.
```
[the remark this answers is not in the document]
BarclayII: Added.

python/dgl/dataloading/negative_sampler.py
```diff
@@ -65,3 +66,67 @@ def _generate(self, g, eids, canonical_etype):
...
+    For each edge ``(u, v)`` of type ``(srctype, etype, dsttype)``, DGL generates
+    :attr:`k` pairs of negative edges ``(u', v')``, where ``u'`` is chosen uniformly from
+    all the nodes of type ``srctype`` and ``v'`` is chosen uniformly from all the nodes
+    of type ``dsttype``.  The resulting edges will also have type
+    ``(srctype, etype, dsttype)``.
+
+    Parameters
+    ----------
+    k : int
+        The number of negative examples to generate per minibatch.
```
mufeili, line 86: "examples" -> "samples"

python/dgl/dataloading/negative_sampler.py
```diff
@@ -65,3 +66,67 @@ def _generate(self, g, eids, canonical_etype):
...
+    exclude_self_loops : bool, optional
+        Whether to exclude self-loops from negative examples.  (Default: True)
```
mufeili, line 88: Ditto

python/dgl/dataloading/negative_sampler.py
```diff
@@ -65,3 +66,67 @@ def _generate(self, g, eids, canonical_etype):
...
+    unique : bool, optional
+        Whether to sample unique negative examples.  Setting it to False will make things
+        faster.  (Default: True)
```
mufeili, line 91: Is it same as sampling with or without replacement? If so, why not call it `replacement`?

python/dgl/dataloading/negative_sampler.py
mufeili, line 86: With `num_trials`, does it mean that `k` is the maximal number of negative samples you can get per minibatch? If so, make it clear.

python/dgl/dataloading/negative_sampler.py
```diff
@@ -65,3 +66,67 @@ def _generate(self, g, eids, canonical_etype):
...
+        The number of rejection sampling trials.
+
+        Increasing it will increase the likelihood of getting :attr:`k` negative examples,
+        but will also take more time if the graph is dense.
+    redundancy : float, optional
+        Indicates how much more negative examples to actually generate during rejection sampling
+        before finding the unique pairs.
+
+        Increasing it will increase the likelihood of getting :attr:`k` negative examples,
+        but will also take more time and memory.
```
mufeili, line 102: Is it a standard way to expose both `num_trials` and `redundancy`? They are hard to understand if users are not familiar with the implementation.

python/dgl/dataloading/negative_sampler.py
```diff
@@ -65,3 +66,67 @@ def _generate(self, g, eids, canonical_etype):
...
+
```
mufeili, line 103: Do you allow excluding positive samples from the sampled negative samples?

python/dgl/sampling/negative.py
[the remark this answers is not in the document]
zjost: I know that PyG excludes self-loops from their negative samples, but I haven't invested energy into understanding why.  Intuitively, it seems strange that you would use a self-loop as a negative edge sample, but I haven't conducted or seen any studies of this question.

python/dgl/dataloading/negative_sampler.py
zjost: Perhaps we could calculate "redundancy" based on the sparsity of the graph and hard-coding some probabilistic measure?  For example, the following function calculates the number of node pairs required given the desired "k_hat" number of negative samples:

```python
import numpy as np
from numpy.polynomial import polynomial

def calc_draw(k_hat, num_edges, num_nodes, r=3):
    """
    Args:
    ---------
    k_hat : int
        Number of desired samples
    num_edges : int
        Number of edges in the graph
    num_nodes : int
        Number of nodes in the graph
    r : float
        Number of standard deviations
        
    Calculates the number of samples required based on a lower-bound
    of the expected number of negative samples, based on N draws from 
    a binomial distribution.  Solves the following equation for N:
    
    k_hat = N*p_k - r * np.sqrt(N*p_k*(1-p_k))
        
    where p_k is the probability that a node pairing is a negative edge 
    and r is the number of standard deviations to construct the lower bound
    """
    p_m = num_edges / (num_nodes*(num_nodes-1))
    p_k = 1 - p_m
    
    a = p_k**2
    b = -p_k*(2*k_hat+r**2*p_m)
    c = k_hat**2
    
    poly = polynomial.Polynomial([c, b, a])
    N = poly.roots()[-1]
    redundancy = N / k_hat - 1.
    return redundancy
```

For example, using 3 standard deviations to give a probability >99.9% and the graph stats from DDI (the most dense link prediction graph on OGB by far @ 14% density), I get a redundancy of 21% for a batch size of 1024.  The second most dense graph (biokg @ 0.1% density) gets a redundancy of 0.4%.  

The good news is that this is relatively insensitive to the `r` parameter above, since the variance scales as `sqrt(N*p_k*(1-p_k))`.  For DDI, the redundancies for 1, 3 and 5 standard deviations are 18%, 21%, and 24% respectively.  

Maybe we just set it to something like 5 and issue a warning if something is strange, like a redundancy > 1...etc?  

For the number of trails, perhaps a similar analysis can be done based on the combinatorics.  For example, the number of node combinations in DDI is 9.1M, and if 85% of those are negative edges, that's 7.8M.  For a batch size of 1024, there's around 0.01% chance you would have a duplicate.  In that case, maybe we just get the redundancy right and remove the num_trials option?

python/dgl/sampling/negative.py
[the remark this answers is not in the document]
bioannidis: Self loops should be excluded since if you add self loop as a negative edge the optimum solution for the optimization problem is a vector of all zero. I would remove self-loop.

python/dgl/sampling/negative.py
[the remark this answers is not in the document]
zjost: Is the `etype` needed because we use it to construct a heterogeneous graph for the negative samples?  If that's the case, is the question whether or not we should sample the edge type or keep it static?

A follow-up question:  given `etype`, let's say you generate a negative sample, `(u, v)`.  If `(u, v)` shares an edge in the real graph, but it's of a different type than `etype`, do you still exclude it, or do you only remove positive pairs with `etype` edges?

python/dgl/dataloading/negative_sampler.py
```diff
@@ -65,3 +66,67 @@ def _generate(self, g, eids, canonical_etype):
...
         return src, dst
+
+# Alias
+Uniform = PerSourceUniform
+
+class GlobalUniform(_BaseNegativeSampler):
+    """Negative sampler that randomly chooses negative source-destination pairs according
+    to a uniform distribution.
+
+    For each edge ``(u, v)`` of type ``(srctype, etype, dsttype)``, DGL generates
```
zjost, line 77: I'm confused by this.  This almost sounds like `k=2` means you'll get `2*g.num_edges()` returned, not just 2.  Maybe we can start this doc string at "DGL generates..." and exclude this first phrase?  I'm also confused by how srctype, etype and dsttype come into play.  Which parts are specified by the user, what assumptions are made, and how does it impact the results?

python/dgl/dataloading/negative_sampler.py
zjost: We should perhaps rename this to `k_max`, or something similar to indicate that `k` samples are not guaranteed.

python/dgl/sampling/negative.py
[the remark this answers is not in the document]
BarclayII: > Is the `etype` needed because we use it to construct a heterogeneous graph for the negative samples?

Yes.

> If that's the case, is the question whether or not we should sample the edge type or keep it static?

Yes, or "should we sample the edge type for negative examples as well".  I think those negative examples will be too easy.

> If `(u, v)` shares an edge in the real graph, but it's of a different type than `etype`, do you still exclude it, or do you only remove positive pairs with `etype` edges?

I only remove positive pairs with `etype` edges.

python/dgl/dataloading/negative_sampler.py
BarclayII: Since the positive examples can have different edge types (i.e. the `eids` will be a dictionary), I'm trying to say DGL will generate k negative pairs per edge with the same edge type.

python/dgl/dataloading/negative_sampler.py
BarclayII: They are by default excluded.

python/dgl/dataloading/negative_sampler.py
BarclayII: Yes.  If we can determine the redundancy intelligently we can remove both arguments actually.  I would probably hardcode the number of trials to 3 or something.

python/dgl/dataloading/negative_sampler.py
BarclayII: I'm hesitant to name it `k_max` because most of the time it does generate `k` examples, barring some corner cases where the graph is extremely dense.  Naming it to `k_max` may give people impression that it often does not generate `k` examples.

python/dgl/dataloading/negative_sampler.py
BarclayII: Because the number of edges per type could vary across minibatches, I think it's not a good idea to fix the number of total negative samples to generate per minibatch.  I also changed the implementation to generate `len(eid) * k` negative samples.

python/dgl/dataloading/negative_sampler.py
BarclayII: Added in description.

python/dgl/dataloading/negative_sampler.py
BarclayII: Changed the implementation of `global_uniform_negative_sampling` to include @zjost 's suggestion.  Now it only has a single redundancy argument which defaults to None (where the redundancy is inferred from the density of the graph).  The number of trials is fixed to 3.

python/dgl/dataloading/negative_sampler.py
BarclayII: Replaced.

jermainewang (approved):

mufeili (approved):
I'm good.

python/dgl/dataloading/negative_sampler.py
```diff
@@ -65,3 +66,60 @@ def _generate(self, g, eids, canonical_etype):
...
+    This negative sampler will try to generate as many negative samples as possible, but
+    it may rarely return less than :attr:`k` negative samples per edge.
+    This is more likely to happen if a graph is so small or dense that not many unique
+    negative samples exist.
+
+    Examples
+    --------
+    >>> g = dgl.graph(([0, 1, 2], [1, 2, 3]))
+    >>> neg_sampler = dgl.dataloading.negative_sampler.GlobalUniform(4, True)
+    >>> neg_sampler(g)
```
VoVAllen, line 113: Why this example doesn't have eid as argument, i.e. why it's not `ned_sampler(g, eids)`?

zjost:
Thanks for the great work team!

Issue #3357, opened 2021-09-16
Dataloaders making use of DDP and _ScalarDataBatcher are not memory scalable, currently O(N*P).


nv-dlasalle:
## 🐛 Bug

Currently, the dataloaders when use_ddp=True, use O(N*P) memory and have spend O(N) to shuffle the dataset when using P processes.

This problem is most prominent when using an EdgeDataLoader where the number of input edges is on the same order as the number of edges in the graph, while using multiple GPUs, due to the way the permutation is generated (from https://github.com/dmlc/dgl/blob/master/python/dgl/dataloading/pytorch/dataloader.py#L134):
```
        if self.shuffle:
            # deterministically shuffle based on epoch and seed
            g = th.Generator()
            g.manual_seed(self.seed + self.epoch)
            indices = th.randperm(len(self.dataset), generator=g)
        else:
            indices = th.arange(len(self.dataset))

        if not self.drop_last:
            # add extra samples to make it evenly divisible
            indices = th.cat([indices, indices[:(self.total_size - indices.shape[0])]])
        else:
            # remove tail of data to make it evenly divisible.
            indices = indices[:self.total_size]
        assert indices.shape[0] == self.total_size
```
This allocate an array of length the entire training dataset in each process. When using 8 training processes on a dense graph, this means just the dataloader we can end up using 8 times more memory than for storing the graph itself.

This problem also appears to be present when `_ScalarDataBatcher` is not used, but I believe that is a problem with how pytorch creates a `list` of batch items, and outside of the scope of DGL.

## Expected behavior

From a user perspective, it would be expected that using more GPUs would not significantly increase the amount of CPU memory required.

## Environment

 - DGL Version (e.g., 1.0): 0.7.1
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): pytorch 1.9
 - OS (e.g., Linux): ubuntu 20.04
 - How you installed DGL (`conda`, `pip`, source): pip
 - Build command you used (if compiling from source):
 - Python version: 3.6
 - CUDA/cuDNN version (if applicable): 11.3
 - GPU models and configuration (e.g. V100): A100
 - Any other relevant information:

## Additional context

A memory scalable way to generate the indices is needed. Either by splitting the target edges/nodes statically among processes, and only shuffling within a process, or have processes cooperatively generate a split permutation vector, such that each process only hold N/P indices.

[...]

PR #3614, opened 2021-12-28
[Performance] Shared-memory Indices Permutation in DDP and _ScalarDataBatcher

yaox12:
## Description
This is to tackle issue #3357. We share the indices tensor among all the worker processes. At the beginning of each epoch, each worker process shuffles a continuous range of the shared indices tensor and then gets its own indices by stride slicing. This is not truly random but should be pretty good.

This should use O(N) memory and O(N/p) time for shuffling. However, I fail to capture the memory usage precisely due to noises. 

cc @BarclayII @nv-dlasalle 

## Performance

Run the script
```bash
python3 train_sampling_unsupervised.py --num-layers 1 --fan-out 8 --batch-size 500000 --gpu 0,1,2,3,4,5,6,7 --num-epochs 1 --dataset ogbn-products --data-cpu`
```
I got the time of shuffling indices tensor (average of processes) and enumerating the dataloader for the first time as below (in seconds):

|                    | shuffle | enum dataloader |
|--------------|---------|-----------------|
| dgl nightly   |    6.56 |            9.72 |
| randperm and assign  |    0.95 |            2.39 |
| shuffle a continuous range | 1.26 |        2.90 |

- `randperm and assign`: Implementation in this PR.
- `shuffle a continuous range`: It's something like the following code, we find it a bit slower.
```python
np.random.shuffle(self._shared_indices[self.num_samples*self.rank:self.num_samples*(self.rank + 1)].numpy())
```

## Known Issue

Currently, we use `dist.broadcast_object_list()`, which is introduced in PyTorch 1.7, to share the filename of `_shared_indices` among worker processes. We may need a workaround for lower PyTorch versions.


python/dgl/dataloading/pytorch/dataloader.py
```diff
@@ -101,6 +101,32 @@ def __init__(self, dataset, shuffle=False, batch_size=1,
...
+                indices = indices.share_memory_()
+                _shared_filename = list(indices.storage()._share_filename_())
+            else:
+                _shared_filename = [None, None, None]
+            # Note: dist.broadcast_object_list() is a new feature introduced in PyTorch 1.7.0.
+            #       We may need a fallback for lower PyTorch versions.
+            dist.broadcast_object_list(_shared_filename, src=0)
+            # reconstruct the shared tensor
+            if self.rank != 0:
+                indices = th.LongTensor(th.LongStorage._new_shared_filename(*_shared_filename))
```
VoVAllen, line 125: Interesting usage that I don't know before. Can we also use this to share the graph across multiple process? @BarclayII

VoVAllen (commented):
The solution is quite neat. Thanks for the contribution. I'm not so familiar with this part and will let @BarclayII to review this.

python/dgl/dataloading/pytorch/dataloader.py
BarclayII: Since they are prefixed with `_` I suspect that they are undocumented hence unstable.  Could we use a queue instead since it's the preferred way of sharing tensors across processes?

BarclayII: An alternative might be using `NDArray::EmptyShared` which creates a shared memory tensor directly.  The shared memory name can be generated in rank 0 and broadcasted to other ranks.

yaox12: >  use a queue

We have to create a queue and pass it to the worker processes when spawning/forking them, hard to do it inside the dataloader.

> using `NDArray::EmptyShared`

In fact I have tried it. `empty_shared_mem` uses a name to identify the shared memory. But I realize that there may be multiple dataloaders for training/test/validation, so we cannot use a hard code here.

@BarclayII

BarclayII: > `empty_shared_mem` uses a name to identify the shared memory. But I realize that there may be multiple dataloaders for training/test/validation, so we cannot use a hard code here.

That's because you use the same name `_shared_indices_for_ddp`?  I think you could generate a name in rank 0 and distribute it with `dist.broadcast` like you are doing here.

VoVAllen: I think it's fine to use this internal API since the semantic is clear. We can add some check or fallback solution if needed.
Our EmptyShared doesn't support windows as what I remembered. That will break the user experience on windows.

VoVAllen: Also I believe for error handling and error messages, Pytorch does a better job than us

BarclayII: You could also partition the training indices without shuffling to every rank, and then shuffle each partition individually.

BarclayII: @VoVAllen I can take care of supporting shared memory on Windows since they have a very similar interface.

yaox12: > You could also partition the training indices without shuffling to every rank, and then shuffle each partition individually.

Do you mean using a static split, i.e., the partition doesn't change between epochs? For this case, I think letting user partition the indices and initialize the dataloader with `use_ddp=False` will make things much easier.

If you mean something like:
```python
self._indices = _shared_indices[self.rank:self.total_size:self.num_replicas]
# then in iter_ddp():
np.random.shuffle(self._indices.numpy())
```
This would be even slower, which takes 2x times for shuffling compared to the above `randperm and assign`.

BarclayII: > Do you mean using a static split, i.e., the partition doesn't change between epochs? For this case, I think letting user partition the indices and initialize the dataloader with `use_ddp=False` will make things much easier.

Right, say that we have N nodes and R workers, then each rank `r` is only responsible for shuffling the node ID `[N // R * r, N // R * (r+1)]`.  You could still use `randperm` and assign.  This should be OK because even if the nodes are ordered by label, since the gradients are synchronized across workers and would not be too far off.

yaox12: @BarclayII I have pushed a version that uses a static indices split for each process. It's less random but being the fastest way.

yaox12:
This is covered by the new data loader. So close it.

PR #3615, opened 2021-12-29
[Windows] Support NDArray in shared memory on Windows

BarclayII:
Also fixed a compilation error with Windows SDKs later than 10.0.18362.0.


VoVAllen (commented):
Looks good to me/

My reference: https://stackoverflow.com/questions/29260148/how-to-use-shared-memory-correctly-under-windows

VoVAllen (approved):

PR #3616, opened 2021-12-29
[Feature] Pin dgl.graph to the page-locked memory

yaox12:
## Description

This pins dgl.graph to the page-locked memory for GPU zero-copy access.

Note: This leverages `cudaHostRegister` to pin an existing host memory, which is visible by all the CUDA context and can be shared between processes. It's for zero-copy access, e.g., UVA-based GPU sampling, not for async H2D transfer.

## Checklist
- [x] Code is well-documented

## Changes

C++ side:
- inplace `PinMemory_()`/`UnpinMemory_()` for `NDArray`/`COOMatrix`/`CSRMatrix`/`COO` graph/`CSR` graph
- static `PinData()`/`UnpinData()` for `DLTensor`
- static `PinMemory()`/`UnpinMemory()` for `UnitGraph`/`HeteroGraph`
- `IsPinned()` for above classes.

Python side:
- inplace `pin_memory_()`/`unpin_memory_()` and `is_pinned()` for `heterograph` and `heterograph_index`
- Unit test

Others:
- The `pin_memory_()` method is considered read-only, i.e., new sparse formats materialization is not allowed for pinned graphs. Users should materialize all the needed formats before pinning. We will give proper doc and error information for this.


yaox12:
One thing for discussion:

If a pinned graph creates a new format, do we have to pin the new format?
If not, the UnitGraph would be partly pinned. The result returned by `IsPinned()`, which returns `GetAny()->IsPinned()`, may be not reliable.
Or we just keep the pinning behavior read-only, which is a common practice.

include/dgl/runtime/device_api.h
```diff
@@ -196,6 +194,15 @@ class DeviceAPI {
...
   DGL_DLL static DeviceAPI* Get(DGLContext ctx, bool allow_missing = false);
+
+
+  /*!
+   * \brief Get device API base don context.
+   * \param dev_type The device type
+   * \param allow_missing Whether allow missing
+   * \return The corresponding device API.
+   */
+  DGL_DLL static DeviceAPI* Get(int dev_type, bool allow_missing = false);
```
yaox12, line 205: Add this because when pinning an array on CPU, we have to use methods in `CUDADeviceAPI`. With this, we can get the needed device API by an int `device_type`.

yaox12:
cc @VoVAllen @BarclayII @nv-dlasalle @davidmin7

VoVAllen:
I think `shared_memory_()` should take formats as argument, such as `shared_memory_(['csc', 'csr'])`. Then the graph will materialize those formats and make them pinned.

davidmin7:
Looks like a very neat implementation! I have a couple of questions related to the usage of `kDLCPUPinned`. This also bothered me when I was implementing the UnifiedTensor.

1. As far as I know, [DLPack](https://github.com/dmlc/dlpack/blob/main/include/dlpack/dlpack.h) has changed the `kDLCPUPinned` to `kDLCPUHost`. So if we use `kDLCPUPinned` in DGL, can it cause a compatibility issue later?

2. `kDLCPUHost` specifically targets the memory allocated with `cudaMallocHost()`. But as we know, we want to pin the memory here with `cudaHostRegister()` following some regular CPU memory allocation. Can there be any compatibility issue here by trying to treat both the `cudaHostReigser()` memory and the `cudaMallocHost()` memory identically? I personally think it would be really great if a dedicated `kDL` type added in DLPack for the `cudaHostRegister()` type memory.

Thanks.

VoVAllen:
@davidmin7 Change name won't break the compatibility of enum.

VoVAllen:
Also I think we need to consider both shared memory and pin memory together. Since we want to support multi GPU training with pinned memory

BarclayII:
@ZenoTan This might be relevant to DGL + torch-quiver as well.

yaox12:
@davidmin7 Thanks for reviewing.  I have looked into recent changes to DLPack.

1. Changing name is ABI compatible. Frameworks using different versions of DLPack can safely exchange tensors. If we want to upgrade the DLPack version in DGL in the future, some name replacing work has to be done.
2. The main difference between `cudaHostRegister()` memory and the `cudaMallocHost()` memory are their ways for allocation/freeing. If someone out of DGL borrowed an NDArray pinned by `cudaHostRegister()` in DGL, it would not be a big issue as long as they correctly use our `deleter`. Of course, I'm supportive of adding a new enum for `cudaHostRegister()` memory but it's not easy I think.

yaox12:
@VoVAllen I have a rough version for single-GPU sampling from pinned memory, and the results are pretty good. Will go ahead with the multi-GPU case.

VoVAllen:
@yaox12 For the format problem, I've discussed with @jermainewang and @BarclayII. We think it should take a list of formats  as argument, such as `graph.pin_memory_(['csc', 'csr'])`. And those formats will be materialized and pinned, and other formats won't be allowed. Do you think it's reasonable?

nv-dlasalle:
> @yaox12 For the format problem, I've discussed with @jermainewang and @BarclayII. We think it should take a list of formats as argument, such as `graph.pin_memory_(['csc', 'csr'])`. And those formats will be materialized and pinned, and other formats won't be allowed. Do you think it's reasonable?

Do we expect users to know which formats they need for which operations? Could we just pin all existing formats in a graph, and when a new format is materialized in a `pinned` graph, it would also get pinned? That way the user experience would be more in line with non-pinned graphs, and power users who know exactly which formats that need could instantiate only those formats before pinning the graph?

davidmin7:
@VoVAllen and @yaox12 thank you for the explanation. Yup, I don't have any further immediate concerns (I wouldn't comment on the coding style since I don't know too much) and I think this would be a very nice foundational work.

yaox12:
@VoVAllen @nv-dlasalle Considering multi-GPU sampling on the graph stored in pinned and shared host memory. We have to copy the graph to shared memory before spawning/forking worker processes, and then pin the shared graph in every worker process. At this time, if a new format is materialized, they (materialized in different worker processes) will not be shared.
I suggest we just pin all existing formats in a graph, throw an error if a new format is materialized and tell users which format is needed.

VoVAllen:
Usually it's not ideal to materialize all formats. For example when graph sampling on in-edges, only csc is needed. Materializing all formats will consume unnecessary memory. 
I think by default you can materialize all formats, but can also take the formats as argument at the pin memory step. How do you think about this?

VoVAllen:
@nv-dlasalle I think pin_memory is an advanced function, therefore it's fine for user to control the formats by themselves? My concern is the total pin memory size is limited, that materialize all formats will cause error in some scenarios?

yaox12:
@VoVAllen Sorry, I mean pinning all the formats that have already been materialized. If an implicit format materialization happens, we throw an error and tell users that you have to materialize that format before pinning. This is to avoid the case that extra formats are created in sub-processes when pinning.

VoVAllen:
@yaox12 Sure, I think we're on the same page, that you can use current dgl format mechanism to raise error when new formats happening. My point is that adding format argument to pin_memory function can give user a hint that this migh be something you want to take care of.

yaox12:
@VoVAllen I'm concerning such a case:
```python
# in the main process, uses load a graph g, may be a COO graph
g = load_graph(...) 

# then they perform multi-GPU training
for proc_id in range(n_gpus):
    # fork will put the graph to shared memory
    p = mp.Process(target=run, args=(proc_id, n_gpus, args, devices, g, ...)) 
    p.start()
    procs.append(p)
for p in procs:
    p.join()

# then in worker processes, users pin the csc format for UVA sampling
# however, the csc format is materialized in all sub-processes, which is not expected
g.pin_memory_(['csc'])
```

VoVAllen:
Do you mean like this?
```python
def run(...):  
  # then in worker processes, users pin the csc format for UVA sampling
  # however, the csc format is materialized in all sub-processes, which is not expected
  g.pin_memory_(['csc'])
  # main loop


# in the main process, users load a graph g, may be a COO graph
g = load_graph(...) 

# then they perform multi-GPU training
for proc_id in range(n_gpus):
    # fork will put the graph to shared memory
    p = mp.Process(target=run, args=(proc_id, n_gpus, args, devices, g, ...)) 
    p.start()
    procs.append(p)
for p in procs:
    p.join()

```

I see what you mean. That seems all the graph formats should be materialized before spawning/forking the subprocess?

VoVAllen:
The grammar I think is still compatible here. And seems that it might be different for spawn/fork? Will duplicate call pin_memory on the forked memory cause problem?

VoVAllen:
```
def run(...):  
  # then in worker processes, users pin the csc format for UVA sampling
  # however, the csc format is materialized in all sub-processes, which is not expected
  g.pin_memory_()
  # main loop


# in the main process, users load a graph g, may be a COO graph
g = load_graph(...) 
g.formats(['csc'])

# then they perform multi-GPU training
for proc_id in range(n_gpus):
    # fork will put the graph to shared memory
    p = mp.Process(target=run, args=(proc_id, n_gpus, args, devices, g, ...)) 
    p.start()
    procs.append(p)
for p in procs:
    p.join()
```
Is this what you expect?

yaox12:
> That seems all the graph formats should be materialized before spawning/forking the subprocess?

All graph formats or all the needed formats (if users do know). 

> Will duplicate call pin_memory on the forked memory cause problem?

There might be a problem if the forked processes share the same virtual address of the forked memory, e.g., the graph structure. But it works well in my multi-GPU UVA sampling example, I will look into this.
For spawn, it should be fine.

yaox12:
> ```python
> def run(...):  
>   # then in worker processes, users pin the csc format for UVA sampling
>   # however, the csc format is materialized in all sub-processes, which is not expected
>   g.pin_memory_()
>   # main loop
> 
> 
> # in the main process, users load a graph g, may be a COO graph
> g = load_graph(...) 
> g.formats(['csc'])
> 
> # then they perform multi-GPU training
> for proc_id in range(n_gpus):
>     # fork will put the graph to shared memory
>     p = mp.Process(target=run, args=(proc_id, n_gpus, args, devices, g, ...)) 
>     p.start()
>     procs.append(p)
> for p in procs:
>     p.join()
> ```
> 
> Is this what you expect?

Right. I think adding format arguments in `pin_memory_()` may give some misleading. I prefer to keep them separated.

VoVAllen:
@yaox12 I see what you mean. Please go ahead as you suggested and I think we can finalize the interface when you have the full example code ready.

ZenoTan:
If the current use of pinned memory is for multi-GPU sampling, I think the most common case is for DGL sampler to pin memory and create formats if necessary. Are users happy to deal with these?

BarclayII:
Side question: I noticed that this `pin_memory` is not for async H2D transfer as per the description.  Why is that?  And do we have plans to support async H2D transferring of graphs?  I think supporting that is preferable since it enables us to use `pin_memory` option in PyTorch's DataLoader.

yaox12:
> @yaox12 I see what you mean. Please go ahead as you suggested and I think we can finalize the interface when you have the full example code ready.

@VoVAllen I'm not sure if I should include the UVA sampling related code and examples in this PR, or in a separate PR to make the review/discussion more focused.

yaox12:
@BarclayII `cudaHostRegister()` is as slow or slower than copy, and is mainly used for long-lived memory. Shared memory followed by `cudaHostRegister()` offers a way to access the host pinned memory from multiple GPUs (manipulated by multiple processes), which is especially useful in UVA-based multi-GPU sampling.

For not long-lived memory, e.g., the dataloader case, we should use the pinned memory managed by `cudaMallocHost()` or `cudaHostAlloc()`, as PyTorch does. However, pinned memory allocated in this way cannot be shared among processes.

See this [issue](https://github.com/pytorch/pytorch/issues/32167#issuecomment-574272151) for more discussion. @davidmin7 Correct me if I make any mistakes, thanks.

yaox12:
> If the current use of pinned memory is for multi-GPU sampling, I think the most common case is for DGL sampler to pin memory and create formats if necessary. Are users happy to deal with these?

@ZenoTan Currently, the behaviors of DGL dataloaders change with the input device. For example, if the graph is copied to GPU ahead of time (users are responsible for this), the `output_device`  is set to a GPU device and `num_workers=0`, then all the sampling and subgraph construction will be done on GPU. Here I just follow the routine, if uses are willing to use UVA sampling, they are responsible for pinning it.

But I do think we need a summary doc or table for this, i.e., how the dataloader behaves w.r.t. various input device, `output_device`, `nid` device, `num_workers`, etc.

VoVAllen:
I think we should have the example now to see whether current api design works well. But we can have it in a separate if you like

BarclayII:
> @BarclayII `cudaHostRegister()` is as slow or slower than copy, and is mainly used for long-lived memory. Shared memory followed by `cudaHostRegister()` offers a way to access the host pinned memory from multiple GPUs (manipulated by multiple processes), which is especially useful in UVA-based multi-GPU sampling.
> 
> For not long-lived memory, e.g., the dataloader case, we should use the pinned memory managed by `cudaMallocHost()` or `cudaHostAlloc()`, as PyTorch does. However, pinned memory allocated in this way cannot be shared among processes.

In this case, I think later we will need a separate `pin_memory` call that copies the graph into pinned memory like `Tensor.pin_memory()`.  This is necessary for using PyTorch's DataLoader and enable async H2D transferring of graphs.  Not sharing among processes is OK in this case since the main process is responsible for copying the graph to pinned memory.  Although the cost of copying from pageable memory to pinned memory cannot be avoided, it can at least save the cost of `cudaMallocHost` or `cudaHostRegister` if we use PyTorch's pinned memory allocator (similar to how we use PyTorch's allocator using `tensoradapter`).

davidmin7:
> See this [issue](https://github.com/pytorch/pytorch/issues/32167#issuecomment-574272151) for more discussion. @davidmin7 Correct me if I make any mistakes, thanks.

Yup, your understanding is correct. The point of using `cudaHostRegister()` is to create a zero-copy accessible memory for the data which will be intensively accessed by GPUs during the training. So what you said basically, the long-lived memory (or dataset) is a correct explanation. We don't want to use this for some transient data created during the training.

yaox12:
Since the UVA sampling code and examples contain a lot of modifications and are based on this PR, I put them in a separate PR in my forked DGL repo. If you'd like to have concrete examples to see whether the current API design works well, you can refer to [[Feature] CUDA UVA sampling for MultiLayerNeighborSampler](https://github.com/yaox12/dgl/pull/2). @VoVAllen @BarclayII

include/dgl/runtime/device_api.h
```diff
@@ -196,6 +194,15 @@ class DeviceAPI {
    * \return The corresponding device API.
    */
   DGL_DLL static DeviceAPI* Get(DGLContext ctx, bool allow_missing = false);
+
+
+  /*!
+   * \brief Get device API base don context.
```
BarclayII, line 200: typo.

include/dgl/runtime/device_api.h
BarclayII: Use `DLDeviceType` rather than `int`.

Is this function added for convenience so that you don't have to write something like `Get(DLContext{kDLGPU, 0})`?

BarclayII (approved):
I'm good with this PR overall.  Please also remember adding the corresponding entries in `docs/source/api/python/dgl.DGLGraph.rst`.

include/dgl/runtime/device_api.h
yaox12: Yes. I think it's clearer to omit the `device_id` in this case, though it doesn't actually matter.

yaox12:
CI seems to fail out of my code.

include/dgl/runtime/ndarray.h
```diff
@@ -272,6 +291,27 @@ class NDArray {
...
+  /*!
+   * \brief Function to pin the data.
+   * \param tensor The array to be pinned.
+   * \note Data of the given array will be pinned inplace.
+   *       Behavior depends on the current context,
+   *       kDLCPU: will be pinned;
+   *       kDLCPUPinned: directly return;
+   *       kDLGPU: invalid, will throw an error.
+   */
+  DGL_DLL static void PinData(DLTensor* tensor);
```
VoVAllen, line 303: What's the difference between `pinmemory` and `pindata`?
I prefer to use the same word here. Otherwise it confused me.

VoVAllen (commented):
Generally looks good to me.
Please also add comment to unitgraph and heterograph's pin memory to indicate this is an inplace method

include/dgl/runtime/ndarray.h
yaox12: `PinData` is a static method to pin the underlying DLTensor of an `NDArray`, while `PinMemory_()` is an inplace method to pin the current `NDArray` by calling `PinData`. `PinData` was previously introduced as a device API by David Min for UnifiedTensor, which directly manipulates the DGLArray. So I just follow it.

davidmin7: I don't mind of renaming it! Current naming does not have a special meaning.

yaox12:
> Please also add comment to unitgraph and heterograph's pin memory to indicate this is an inplace method

@VoVAllen, I'd like to make some changes to UnitGraph and HeteroGraph's `PinMemory()` methods, changing them from static methods to in-place methods.

include/dgl/runtime/ndarray.h
yaox12: I prefer to keep `PinData`. `PinMemory` (without underline) could be reserved for future out-of-place pinning for async transfer.

yaox12: Update comments to clarify their difference.

VoVAllen (approved):
Thanks

VoVAllen:
@yaox12 Did you finish the PR? Can I merge it?

yaox12:
Yes, please merge it. Thanks.

PR #3641, opened 2022-01-11
[Feature] Gather mm

isratnisa:
## Description
Dense Matrix Multiplication interface. It multiplies a 2D dense tensor H and a 3D dense tensor W according to their relation types. Two versions of `gather_mm` is proposed in this PR: 1) `se_gather_mm`: H is sorted according to relation types and, 2) `gather_mm`: H is unsorted and the relation type can be fetched from `etypes`. Example code: 
```
*** H is sorted ***
\param seglen_A : Tensor
        An integer tensor of shape (R,). Each element is the length of segments
        of input ``A``. The summation of all elements must be equal to N.

import backend as F
out_sortedH = F.se_gather_mm(H, W, seglen_A)
```
```
*** H is NOT sorted ***
\param idx_a : Tensor, optional
        If specified, must be a 1-D integer tensor of shape (K,)
\param idx_b : Tensor, optional
        If specified, must be a 1-D integer tensor of shape (N,)

out_unsortedH = F.gather_mm(H, W, idx_a, idx_b=etypes)
```

The following table shows the runtime in milliseconds on a NVIDIA Tesla T4 machine using different feature length. The `PyTorch matmul (low-mem)` version loops through the etypes of the input graph and each etype performs a multiplication `out[etype] = H[etype] @ W[etype]`. `PyTorch bmm (high-mem)` copies the weight matrices to edges and performs batched multiplication `out = bmm(H,W)`.  The `_gather_mm` versions are proposed in this PR and support when input A is sorted and unsorted. 

Input is a random graph with #relations=44, #total edges=16612, average #edges per rel=377,  min #edges=10, max #edges=9200 

| feat_len  | PyTorch matmul (low-mem) |  PyTorch bmm (high-mem) |  _gather_mm sortedE | _gather_mm unsortedE |
| ---------| ------------- |--|--|--|
16 | 0.48 | 0.47 | 0.44 | 0.10
32 | 0.54 | 1.34 | 0.54 | 0.15
64 | 0.68 | 4.64 | 0.32 | 0.21
128 | 1.02 | 11.52 | 0.43 | 0.72
256 | 1.93 | 49.95 | 0.81 | 3.95

## Checklist
- [ ] Code is well-documented


src/array/cuda/gather_mm.cu
```diff
@@ -0,0 +1,334 @@
...
+    if (row < num_rows) {
+        unsigned int local_row = row & 3;  // hardcoded for TB size 128 (4 warps)
+        const int sh_h_tile = 64;
+        __shared__ DType sh_H[4 * sh_h_tile];
+        int h_tile = sh_h_tile;
+        if ((in_len - k_start) < h_tile) h_tile = in_len - k_start;
+        /* Load A in shared mem in a coalesced way */
+        for (unsigned int l = laneId; l < h_tile; l += 32)
+            sh_H[local_row * h_tile + l] = A[row * in_len + (k_start + l)];
+        __syncthreads();
```
nv-dlasalle, line 137: Since each warp acts independently, I believe we can replace this with `__sync_warp()`, and should see better performance as a result.

src/array/cuda/gather_mm.cu
```diff
@@ -0,0 +1,334 @@
...
+        }
+    });
+}
+
+/*!
+ * \brief Implementation of Gather_mm operator. The input matrix A is
+ *        expected to be sorted according to relation type.
+ * \param A The input dense matrix of dimension m x k
+ * \param B The input dense matrix of dimension k x n
+ * \param C The output dense matrix od dimension m x n
```
nv-dlasalle, line 279: od -> of

src/array/cuda/gather_mm.cu
```diff
@@ -0,0 +1,334 @@
...
+    unsigned int row = warpId;
+    if (row < num_rows) {
+        unsigned int local_row = row & 3;  // hardcoded for TB size 128 (4 warps)
+        const int sh_h_tile = 64;
+        __shared__ DType sh_H[4 * sh_h_tile];
+        int h_tile = sh_h_tile;
+        if ((in_len - k_start) < h_tile) h_tile = in_len - k_start;
+        /* Load A in shared mem in a coalesced way */
+        for (unsigned int l = laneId; l < h_tile; l += 32)
+            sh_H[local_row * h_tile + l] = A[row * in_len + (k_start + l)];
```
nv-dlasalle, line 136: We should probably keep indexing into `sh_H` fixed, so as to avoid weird performance behavior related to memory bank conflicts as `h_tile` changes:
```
sh_H[local_row * sh_h_tile +l]
```

src/array/cuda/gather_mm.cu
```diff
@@ -0,0 +1,334 @@
...
+            DType out_reg = 0;  // thread private
+            unsigned int k = laneId;
+            if (k < out_len) {
+                /* iterate over elements of a row of A */
+                for (unsigned int i = 0; i < h_tile; i++) {
+                    DType h_val =  sh_H[local_row * h_tile + i];
+                    /* iterate over elements of a row of B in parallel */
+                    out_reg += h_val * B[B_offset + ((i + k_start) * out_len + (k_outloop + k))];
+                }
+                C[row * out_len + (k_outloop + k)] += out_reg;
```
nv-dlasalle, line 150: The CPU implementation cublas implementations directly assign to C (overwriting any previous values), but this one adds to them. I think we should make sure the behavior is consistent across the implementations.

src/array/cuda/gather_mm.cu
```diff
@@ -0,0 +1,334 @@
...
+        for (int i = 0; i < num_rel; ++i)
+            tot_num_rows += A_rel_data[i];
+
+        const int ntx = 128;
+        const int warp_size = 32;
+        const int nbx =  ((tot_num_rows * warp_size + ntx - 1) / ntx);
+        const dim3 nblks(nbx);
+        const dim3 nthrs(ntx);
+        const int warp_per_block = 4;  // ntx / warp_size;
+        for (int k_start = 0; k_start < k; k_start += 64) {
```
nv-dlasalle, line 181: What keeps us from putting this loop inside of the kernel?

src/array/cuda/gather_mm.cu
```diff
@@ -0,0 +1,334 @@
...
+        IdType tot_num_rows = 0;
+        for (int i = 0; i < num_rel; ++i)
+            tot_num_rows += A_rel_data[i];
+
+        const int ntx = 128;
+        const int warp_size = 32;
+        const int nbx =  ((tot_num_rows * warp_size + ntx - 1) / ntx);
+        const dim3 nblks(nbx);
+        const dim3 nthrs(ntx);
+        const int warp_per_block = 4;  // ntx / warp_size;
```
nv-dlasalle, line 180: Is this variable used?

src/array/cuda/gather_mm.cu
```diff
@@ -0,0 +1,334 @@
...
+                if (b_trans)  {
+                    k = tmp;
+                    std::swap(n, k);
+                }
+            }
+            int ldb = n, lda = k, ldc = n;
+            CUBLAS_CALL(cublasGemm<DType>(
+                thr_entry->cublas_handle,
+                CUBLAS_OP_N,
+                CUBLAS_OP_N,
```
nv-dlasalle, line 256: Are you seeing better performance manually transposing the matrices, and then calling cublas, as opposed to just setting the transpose flag in the cublas call?

src/array/cuda/gather_mm.cu
```diff
@@ -0,0 +1,334 @@
...
+          NDArray C,
+          const NDArray A_dim1_per_rel,
+          const NDArray B_dim1_per_rel,
+          const NDArray etype,
+          bool sortedA, bool a_trans, bool b_trans) {
+    if (sortedA) {  // similar to low-mem matmul
+        gatherMM_SortedEtype<XPU, IdType, bits>(A, B, C, A_dim1_per_rel,
+            B_dim1_per_rel, a_trans, b_trans);
+    } else {  // similar to bmm (high-mem) without copying weights to edges
+        // TODO(Israt): Add support for A and B to have different dimension in relation types
```
nv-dlasalle, line 299: Does this need an assert/check?

tests/compute/test_gatherMM.py
```diff
@@ -0,0 +1,121 @@
...
+num_rel_scale = 1
+in_feat = 16
+out_feat = 8
+print("in/out feat", in_feat, out_feat)
+
+@unittest.skipIf(dgl.backend.backend_name != 'pytorch', reason='Only support PyTorch for now')
+@unittest.skipIf(F._default_context_str == 'cpu', reason="Not implemented.")
+
+@parametrize_dtype
+def test_low_mem(idtype):
```
nv-dlasalle, line 21: Should this be named "test_gathermm" since it seems like that's what it tests?

src/array/cuda/gather_mm.cu
isratnisa: The addition is required here as we have a loop over `in_len` (cols of A/rows of B) which contributes to the same output location. The C code doesn't use any outer loop.

src/array/cuda/gather_mm.cu
isratnisa: Yes. Added a check inside the called function.

src/array/cuda/gather_mm.cu
isratnisa: Actually, I didn't compare the performance.  Wanted to keep a generic cublas gemm call  with or without transposed versions. But you might be right. Transposing using flags might be faster.  Btw, we also are using cublas for the transpose. I think that's what you meant by "manual"?

nv-dlasalle: Yes, sorry, by "manually" I really meant explicitly transposing them with cublas. I'd expect better performance simply telling cublas the matrices are to be multiplied transposed in it's gemm call, as it can then do something like transposing only blocks in shared memory.

isratnisa: You are right! By implicitly transposing, I got 10%~45% improvement. I experimented with feature length from 16 to 512 on a random hetero-graph. Will stick to the implicit transposition. Thanks.

jermainewang:
Please clean up the PR as follows:
* Remove all benchmark codes.
* `gather_mm` is a general operator that may/may not be used with DGLGraph. Therefore, I suggest to change the interface to be as plain tensor operator as possible:
    ```python
    def gather_mm(a, b, idx_a=None, idx_b=None):
        """Gather data according to the given indices and perform matrix multiplication.
    
        Let the result tensor be c, the operator conducts the following computation:
    
        If both idx_a and idx_b are not none:

          c[i] = a[idx_a[i]] @ b[idx_b[i]]
          , where len(c) == len(idx_a) == len(idx_b)

        If idx_a is given but not idx_b:
    
          c[i] = a[idx_a[i]] @ b[i]
          , where len(c) == len(idx_a)

        If idx_b is given but not idx_a:

          c[i] = a[i] @ b[idx_b[i]]
          , where len(c) == len(idx_b)

        Parameters
        ----------
        a : Tensor
            A 2-D tensor of shape (N, D1)
        b : Tensor
            A 3-D tensor of shape (M, D1, D2)
        idx_a : Tensor, optional
            If specified, must be a 1-D integer tensor of shape (K,).
        idx_b : Tensor, optional
            If specified, must be a 1-D integer tensor of shape (K,).
    
        Returns
        -------
        c : Tensor
            A 2-D tensor of shape (K, D2).
        """
        pass
    ```
    For sorted version,
    ````python
    def segmm(a, b, seglen_a):
        """Perform matrix multiplication according to segments.
    
        Suppose ``seglen_a == [10, 5, 0, 3]``, the operator will perform four matrix multiplications:
          a[0:10] @ b[0], a[10:15] @ b[1], a[15:15] @ b[2], a[15:18] @ b[3]

        Parameters
        ----------
        a : Tensor
            A 2-D tensor of shape (N, D1).
        b : Tensor
            A 3-D tensor of shape (K, D1, D2).
        seglen_a : Tensor
            An integer tensor of shape (K,). Each element is the length of segments of input ``a``. The summation
            of all elements must be equal to N.
    
        Returns
        -------
        c : Tensor
            A 2-D tensor of shape (N, D2).
        """
        pass
    ````
* Users should not be aware of output data allocation, so please hide that under the hood. Here is the expected user experience:
    ```python
    import dgl
    import torch
    a = torch.randn(10, 32)
    b = torch.randn(5, 32, 64)
    idx_a = torch.tensor([1, 0, 4, 4, 0, 0, 2, 2, 3])  # 9 IDs
    c = dgl.sparse.gather_mm(a, b, idx_a)
    print(c.shape)    # (9, 64)
    ```

src/array/cuda/gather_mm.cu
```diff
@@ -0,0 +1,598 @@
...
+        if (b_trans)
+            device->FreeWorkspace(B->ctx, B_trans_data);
+    });
+}
+
+}  // namespace cuda
+
+/* \brief Implementation of SegmentMM operator. Each segment calls cuBLAS
+ * GEMM operator to multiply segment of A and B. When A or B needs to be
+ * tranposed, cuBLAS GEMM swtiches it's transpose parameter (CUBLAS_OP_T).
```
jermainewang, line 325: switches

src/array/cuda/gather_mm.cu
```diff
@@ -0,0 +1,598 @@
...
+    });
+}
+
+/* \brief Implementation of SegmentMM operator. Each segment calls cuBLAS
+ * GEMM operator to multiply segment of A and B. When A or B needs to be
+ * tranposed, cuBLAS's Xgeam operation is used to tranpose the teh matrix
+ * prior to calling the GEMM operation.
+ */
+template <int XPU, typename IdType, int bits>
+void segment_mm_explicitTranspose(const NDArray A,
```
jermainewang, line 394: Will this function be needed again in the future?

python/dgl/backend/pytorch/sparse.py
```diff
@@ -691,6 +693,62 @@ def backward(ctx, dB_weights):
...
+        B = B.reshape(B.shape[0] * B.shape[1], B.shape[2])
+        C = th.zeros((A.shape[0], B.shape[1]), device=A.device, dtype=A.dtype)
+        C = _segment_mm(A, B, C, seglen_A)
+        ctx.backward_cache = A, B, seglen_A, B_3D_shape
+        return C
+
+    @staticmethod
+    def backward(ctx, dZ):
+        A, B, seglen_A, B_3D_shape = ctx.backward_cache
+        #  Compute A_grad = Out_grad * B^T
```
jermainewang, line 716: Please change to the following structure.
```python
A_grad = B_grad = None
if ctx.needs_input_grad[0]:
    # Compute A_grad
    ...
if ctx.needs_input_grad[1]:
    # Compute B_grad
    ...
return A_grad, B_grad, None
```

python/dgl/backend/pytorch/sparse.py
```diff
@@ -691,6 +693,62 @@ def backward(ctx, dB_weights):
...
+        # Reshaping B form 3D to 2D
+        B_3D_shape = B.shape
+        B = B.reshape(B.shape[0] * B.shape[1], B.shape[2])
+        C = th.zeros((A.shape[0], B.shape[1]), device=A.device, dtype=A.dtype)
+        C = _gather_mm(A, B, C, B_3D_shape[0], idx_a, idx_b)
+        ctx.backward_cache = A, B, idx_a, idx_b, B_3D_shape
+        return C
+
+    @staticmethod
+    def backward(ctx, dZ):
```
jermainewang, line 741: Similar comment as above

src/array/cuda/gather_mm.cu
```diff
@@ -0,0 +1,598 @@
...
+                static_cast<DType*>(A->data),
+                static_cast<DType*>(B->data),
+                static_cast<DType*>(C->data),
+                static_cast<IdType*>(idx_a->data),
+                static_cast<IdType*>(idx_b->data),
+                static_cast<IdType*>(idx_c->data),
+                tot_num_rows,
+                in_len, out_len);
+        } else if (!idx_c_data) {  // use generic gather_mm
+            if (!a_trans) {  // Can't transpose A w/o E per etype info
```
jermainewang, line 302: Merge this if condition with the one above. In general, you shouldn't leave unhandled condition. Rightnow, `(!idx_c_data && a_trans)` is not handed.

src/array/cuda/gather_mm.cu
```diff
@@ -0,0 +1,598 @@
...
+            }
+            std::swap(in_len, out_len);
+        }
+        IdType tot_num_rows = A->shape[0];
+        const int ntx = 128;
+        const int warp_size = 32;
+        const int nbx =  ((tot_num_rows * warp_size + ntx - 1) / ntx);
+        const dim3 nblks(nbx);
+        const dim3 nthrs(ntx);
+        if (a_trans && idx_c_data) {
```
jermainewang, line 288: I think you could ignore `a_trans` for now? since each row of A must be a vector.

python/dgl/backend/pytorch/sparse.py
```diff
@@ -691,6 +693,62 @@ def backward(ctx, dB_weights):
...
+        A, B, idx_a, idx_b, B_3D_shape = ctx.backward_cache
+        #  Compute A_grad = Out_grad * B^T
+        A_grad = th.zeros(A.shape, device=A.device, dtype=A.dtype)
+        A_grad = _gather_mm_scatter(dZ, B, A_grad, B_3D_shape[0], idx_b=idx_b, b_trans=True)
```
jermainewang, line 745: Should it be:
`A_grad = _gather_mm_scatter(dZ, B, A_grad, B_3D_shape[0], idx_b=idx_b, idx_c=idx_a, b_trans=True)`? Although in current cases, `idx_a` is None.

python/dgl/backend/pytorch/sparse.py
```diff
@@ -691,6 +693,62 @@ def backward(ctx, dB_weights):
...
+        #  Compute B_grad = A^T * Out_grad
+        B_grad = th.zeros(B.shape, device=B.device, dtype=B.dtype)
+        B_grad = _gather_mm_scatter(A, dZ, B_grad, B_3D_shape[0], idx_c=idx_b, a_trans=True)
```
jermainewang, line 748: Should it be:
`B_grad = _gather_mm_scatter(A, dZ, B_grad, B_3D_shape[0], idx_a=idx_a, idx_c=idx_b, a_trans=True)`?

jermainewang (changes requested):

jermainewang (approved):

PR #3645, opened 2022-01-12
[Feature] long live server for multiple client groups

Rhett-Ying:
## Description

This change aims to support long live server which supports multiple client groups connect. Different client groups could be launched simultaneously as they are independent from each other. Within the same group, all behavior is same as before.

Main restrictions of multiple client groups:
1. each client group should have same `num_clients` for now. This could be improved in the near future.
2. The name of `DistTensor` should be globally unique for different groups, as `SERVERS` don't distinguish tensors from different groups while it cares about the name only. Current behavior is easy to implement and intuitive as `shared memory` is widely used among servers.
3. Simultaneously reading the `DistTensor` that already exist when `DistGraphServer` is constructed is allowed. These tensors are the ones in original graph, such as `g.ndata['feature']`. These tensors are mapped into each local client when constructing `DistGraph`.

Main sub-features in this change(namely, how to achieve `long live`):
1. `keep_alive` is added into `ServerState` to control whether to boot servers which keeps alive when clients exit.
2. `group_id` is added into `CLIENT` and attached to each request from to server. Server could distinguish the source group of incoming request.

TODO:
1. modify `//tools/launch.py` to support launching multiple groups. The required change for user is to specify unique names for `DistTensor`, `DistEmbedding` in different groups. All servers/clients still could share the same code scripts which is same as single group training.

How to Review:
check the added testcases is good entry I think.

## Checklist
- [ ] Code is well-documented


python/dgl/distributed/kvstore.py
```diff
@@ -362,6 +365,9 @@ def process_request(self, server_state):
         meta = {}
         kv_store = server_state.kv_store
         for name, data in kv_store.data_store.items():
+            if server_state.keep_alive:
+                if name not in kv_store.orig_data:
+                    continue
```
jermainewang, line 370: Not sure why you need to keep a separate `orig_data` names.

jermainewang (approved):
Generally LGTM. Some minor comments.

jermainewang:
One suggestion is to write down the shutdown protocol (probably in the new `shutdown_server` function). It will help the others to understand the code. I'd like to see say if there are two client groups, what kind of RPC messages will be exchanged during the shutdown phase.

python/dgl/distributed/rpc_client.py
```diff
@@ -187,16 +194,28 @@ def connect_to_server(ip_config, num_servers, max_queue_size=MAX_QUEUE_SIZE, net
...
     """
-    from .dist_context import set_initialized
-    set_initialized(False)
-    if rpc.get_rank() == 0:  # Only client_0 issue this command
-        req = rpc.ShutDownRequest(rpc.get_rank())
-        for server_id in range(rpc.get_num_server()):
-            rpc.send_request(server_id, req)
+    rpc.register_service(rpc.SHUT_DOWN_SERVER,
+                         rpc.ShutDownRequest,
+                         None)
```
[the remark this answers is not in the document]
Rhett-Ying: Server is kept alive and waiting for requests.  We have 2 ways to shut servers down. 1: kill server process directly. this is brute. 2. boot a new client group which aims to send shutdown requests to server. This is more elegant.

python/dgl/distributed/rpc_server.py
```diff
@@ -63,39 +67,52 @@ def start_server(server_id, ip_config, num_servers, num_clients, server_state, \
...
+            if len(ips) < rpc.get_num_client():
+                continue
+            else:
+                del recv_clients[group_id]
+            # a new client group is ready
+            ips.sort()
+            client_namebook = {client_id:addr for client_id, addr in enumerate(ips)}
+            for client_id, addr in client_namebook.items():
+                client_ip, client_port = addr.split(':')
+                # TODO[Rhett]: server should not be blocked endlessly.
```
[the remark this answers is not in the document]
Rhett-Ying: No for now, this should be improved in the near future. I plan to cache such `not-connected` clients and re-try periodically. I'd like to do in another change as it's not quite simple/minor.

python/dgl/distributed/kvstore.py
Rhett-Ying: The reason is to map the data that exist when DistGraphServer is constructed only. This check aims to avoid read/write race condition if multiple client groups are booted simultaneously. some new DistTensor may be added by Group_A, but these tensors may be not ready yet when Clients in Group_B is initializing DistGraph. So only the totally ready Tensor is mapped to Group_B.

Rhett-Ying:
> 

Actually the shutdown protocol is same as before for each client/groups. The only difference lies in server side, as an additional field is added to control whether to really shut down server. 
For end users who do not utilize `long live server` feature, they do not need to be aware of shutdown protocol or call explicitly.
For end users who DO utilize `long live server` feature, they just need to call `dgl.distributed.shutdown_servers()` which will boot a special client group which has only one target: send force shutdown request to servers.

python/dgl/distributed/rpc_server.py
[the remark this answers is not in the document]
jermainewang: It'll be great to check out pytorch's RPC to see if such an option is available (probably is).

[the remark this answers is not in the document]
Rhett-Ying: yes. that's desirable...

jermainewang:
Understand that the protocol is usually invisible to users. This docstring is more for us in case someone in the future need to pick up what you've left.

python/dgl/distributed/rpc_client.py
```diff
@@ -187,16 +194,28 @@ def connect_to_server(ip_config, num_servers, max_queue_size=MAX_QUEUE_SIZE, net
     atexit.register(exit_client)
     set_initialized(True)
 
-def shutdown_servers():
+def shutdown_servers(ip_config, num_servers):
```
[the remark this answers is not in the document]
Rhett-Ying: fixed.

python/dgl/distributed/rpc_server.py
```diff
@@ -37,6 +38,9 @@ def start_server(server_id, ip_config, num_servers, num_clients, server_state, \
     assert num_clients >= 0, 'num_client (%d) cannot be a negative number.' % num_clients
     assert max_queue_size > 0, 'queue_size (%d) cannot be a negative number.' % max_queue_size
     assert net_type in ('socket'), 'net_type (%s) can only be \'socket\'' % net_type
+    if server_state.keep_alive:
+        print("As configured, this server will keep alive for multiple"
+              " client groups connect until force shutdown request is received.")
```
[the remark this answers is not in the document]
Rhett-Ying: fixed.

python/dgl/distributed/rpc_server.py
```diff
@@ -63,39 +67,52 @@ def start_server(server_id, ip_config, num_servers, num_clients, server_state, \
...
-    if rpc.get_rank() == 0: # server_0 send all the IDs
-        for client_id, _ in client_namebook.items():
-            register_res = rpc.ClientRegisterResponse(client_id)
-            rpc.send_response(client_id, register_res)
-    # main service loop
+    recv_clients = {}
     while True:
-        req, client_id = rpc.recv_request()
+        # go through if any client group is ready for connection
+        for group_id in list(recv_clients):
```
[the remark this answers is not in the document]
Rhett-Ying: yes. fixed.

python/dgl/distributed/kvstore.py
Rhett-Ying: yes. I am working on it and implement such logic in `launch.py` so that users could pay minimum attention on these constraints.

python/dgl/distributed/rpc_client.py
```diff
@@ -187,16 +194,44 @@ def connect_to_server(ip_config, num_servers, max_queue_size=MAX_QUEUE_SIZE, net
...
     """Issue commands to remote servers to shut them down.
 
+    This function is required to be called manually only when we
+    have booted servers which keep alive even clients exit. In
```
zheng-da, line 201: is it right that this function is called in the process that tries to shut down all servers?

here are my questions:
* if the servers are launched with keepalive=False, how to shut them down?
* if the servers are launched with keepalive=True, does a trainer need to do anything before exits?

python/dgl/distributed/rpc_server.py
```diff
@@ -63,39 +67,52 @@ def start_server(server_id, ip_config, num_servers, num_clients, server_state, \
...
-    if rpc.get_rank() == 0: # server_0 send all the IDs
-        for client_id, _ in client_namebook.items():
-            register_res = rpc.ClientRegisterResponse(client_id)
-            rpc.send_response(client_id, register_res)
-    # main service loop
+    recv_clients = {}
     while True:
-        req, client_id = rpc.recv_request()
+        # go through if any client group is ready for connection
+        for group_id in list(recv_clients.keys()):
```
zheng-da, line 77: why do you put the code of handling client registration code here? it looks weird.

tests/distributed/test_dist_graph_store.py
```diff
@@ -148,7 +150,9 @@ def check_dist_emb(g, num_clients, num_nodes, num_edges):
     from dgl.distributed import DistEmbedding
     # Test sparse emb
     try:
-        emb = DistEmbedding(g.number_of_nodes(), 1, 'emb1', emb_init)
+        group_id = os.environ.get('DGL_GROUP_ID', 0)
+        emb_name = f'emb1_{group_id}'
```
zheng-da, line 154: do we need to change the embedding names or any DistTensor names? who is responsible for adding the group id? the model developer or the framework?

zheng-da: if a user doesn't add the group id in the tensor name, will there be conflicts?

zheng-da:
we should add the group id in the DistTensor name directly to make the name globally unique.

how to stop the servers? is there a command for users to run?

python/dgl/distributed/rpc_client.py
Rhett-Ying: if `keep_alive=True`,  if servers are launched manually, `dgl.distributed.shutdown_servers()` should be called just like what we do in `test_xxx.py`. If servers are launched with `tools/launch.py`, we could terminate the server processes directly or call `dgl.distributed.shutdown_servers()`.

if `keep_alive=False`, everything is exactly same as before. servers will exit once client_0 exit.

python/dgl/distributed/rpc_server.py
Rhett-Ying: as this is where we try to connect peer clients for the same group. It's business logic and I think it's better to put here than in underlying c++ libs. In the future, I prefer to repalce sender/receiver with client/server. Once connection arrives, it's bi-directional, we do not need to create a sender for sending request.

tests/distributed/test_dist_graph_store.py
Rhett-Ying: for now, it's user's responsibility to specify globally unique name or conflict occurs. so does `group_id`. But I will add the `group_id` specify logic into `tools/launch.py` then user don't need to specify manually. it's transparent for users. But user tries to launch clients/servers on their own without `launch.py`, they should be aware of this and takes repsonsibility.

Rhett-Ying:
> we should add the group id in the DistTensor name directly to make the name globally unique.
> 
> how to stop the servers? is there a command for users to run?

1. change is updated which generate globally unique name for `DistTensor` automatically. User DONOT need to care about this. It's transparent to users.
2. call `dgl.distributed.shutdown_servers()` manually or kill/terminate server process forcely.

zheng-da (approved):

PR #3653, opened 2022-01-16
[Bugfix] Fix GPU global negative sampling code

BarclayII:
Noticed two bugs in GPU's global negative sampling code:
(1) Type casting from uint4 to int8 messed up, always returning 0s in CUDA kernels.
(2) The cuRAND RNG I used actually generates 96-bit random numbers instead of 128-bit.

@VoVAllen This should fix your problem.

### DIFF
--- a/src/array/cuda/negative_sampling.cu
+++ b/src/array/cuda/negative_sampling.cu
@@ -43,8 +43,8 @@ __global__ void _GlobalUniformNegativeSamplingKernel(
     for (int i = 0; i < num_trials; ++i) {
       uint4 result = curand4(&rng);
       // Turns out that result.x is always 0 with the above RNG.
-      uint64_t y_hi = y >> 16;
-      uint64_t y_lo = y & 0xFFFF;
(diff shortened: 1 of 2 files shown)


src/array/cuda/negative_sampling.cu
```diff
@@ -42,8 +42,13 @@ __global__ void _GlobalUniformNegativeSamplingKernel(
   while (tx < num_samples) {
     for (int i = 0; i < num_trials; ++i) {
       uint4 result = curand4(&rng);
-      IdType u = ((result.x << 32) | result.y) % num_row;
-      IdType v = ((result.z << 32) | result.w) % num_col;
+      // Turns out that result.x is always 0 with the above RNG.
+      uint64_t y_hi = y >> 16;
```
VoVAllen, line 46: y is undefined as error shows

VoVAllen:
Also I found it doesn't raise error when eid is cpu tensor and graph is on cuda. Is this expected?

VoVAllen (approved):

PR #3658, opened 2022-01-18
[Fix] reduce error msg, refine fetch logic of available ports

Rhett-Ying:
## Description

1. increase re-connect interval in `rpc_client` to reduce error msg printed
2. keeps re-trying to connect to client in `rpc_server` as client is not really listening yet sometimes. This is caused by the asycn call of `context->liseten()` in `tp_communicator.cc::TPReceiver::wait()`.
3. refine the logic of getting available port in tests to avoid `port in use already` issue during test which happens sometimes.
4. re-locate `test_new_kvstore.py` as `reset_envs()` is supposed to solve the test failure issue hit before.
5. un-initialize client before sending shutdown request as `DistTensor::__del__` is called when sending shutdown request sometimes. If `__del__` is called, `DELETE_DATA` request will be sent. But any request is not allowed at this moment, except the shutdown request, as clients which has non-zero ID have already exited...
## Checklist
- [ ] Code is well-documented


VoVAllen (approved):
Looks good to me. Did you check the `DLOG` can be print out normally with `DMLC_LOG_DEBUG`? Also could you add the `DMLC_LOG_DEBUG=1` in the `task_unit_test.sh` in terms of debugging in the future on CI?

python/dgl/distributed/rpc_client.py
```diff
@@ -163,7 +163,7 @@ def connect_to_server(ip_config, num_servers, max_queue_size=MAX_QUEUE_SIZE, net
         server_ip = addr[1]
         server_port = addr[2]
         while not rpc.connect_receiver(server_ip, server_port, server_id):
-            time.sleep(1)
+            time.sleep(3)
```
zheng-da, line 166: does connecting the receiver generate the error messages if the server isn't available? if so, sleeping 3 seconds will still print many error messages.

python/dgl/distributed/rpc_server.py
```diff
@@ -78,7 +79,9 @@ def start_server(server_id, ip_config, num_servers, num_clients, server_state, \
         client_namebook[client_id] = addr
     for client_id, addr in client_namebook.items():
         client_ip, client_port = addr.split(':')
-        assert rpc.connect_receiver(client_ip, client_port, client_id)
+        # TODO[Rhett]: server should not be blocked endlessly.
+        while not rpc.connect_receiver(client_ip, client_port, client_id):
+            time.sleep(1)
```
zheng-da, line 84: why do we need to do this?

tests/distributed/utils.py
```diff
@@ -2,30 +2,38 @@
...
+        ip = '127.0.0.1'
     finally:
         sock.close()
+
+    # scan available PORT
+    ports = []
     sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    sock.bind(("", 0))
-    sock.listen(1)
-    port = sock.getsockname()[1]
```
zheng-da, line 25: this will get a used port?

python/dgl/distributed/rpc_client.py
Rhett-Ying: `DLOG` instead of `LOG` is used in `tp_communicator`, so for normal run, `Failed to connect...` will not be printed out unless `DMLC_LOG_DEBUG=1` is defined.

python/dgl/distributed/rpc_server.py
Rhett-Ying: sometimes `rpc_client::receiver` is not waiting for connection, so we need to re-try...

tests/distributed/utils.py
Rhett-Ying: this will get available/unused port. My change mainly aims to get a few continuous ports in case of `port already in use` when server gets started.

Rhett-Ying:
> Looks good to me. Did you check the `DLOG` can be print out normally with `DMLC_LOG_DEBUG` in current compile options? Also could you add the `DMLC_LOG_DEBUG=1` in the `task_unit_test.sh` in terms of debugging in the future on CI?

`DMLC_LOG_DEBUG=1` is added and works as expected.

PR #3665, opened 2022-01-19
[Sampling] New sampling pipeline plus asynchronous prefetching

BarclayII:
## Description
This is a more complete version of the new sampling experience.  A few things to note:
* The user experience of neighbor sampling for node classification with prefetching is in `examples/pytorch/__temporary__/graphsage/normal.py`.
* That for link prediction is in `examples/pytorch/__temporary__/graphsage/link_pred.py`.
* That for multi-GPU with PyTorch DDP is in `examples/pytorch/__temporary__/graphsage/ddp.py`.  It also includes the shuffling with shared memory array idea from #3614 .
* That for prefetching from feature storages other than PyTorch tensors (as well as a custom graph storage) is in `examples/pytorch/__temporary__/graphsage/disk_storage.py`.
  * The specs for custom graph storage and feature storage is in `examples/pytorch/__temporary__/dglnew/graph/graph.py` and `python/dgl/storages/base.py` respectively.
* That for cluster-GCN training is in `examples/pytorch/__temporary__/cluster_gcn/cluster_gcn.py`.
* New samplers took the place of `python/dgl/dataloading`.  The older ones are moved to `python/dgl/_dataloading` for backup.

Things not done yet are:
[ ] Check whether it works with CPU training
[ ] Check whether it works with graph and data on GPU
[ ] Distributed training
[ ] Heterogeneous graph examples
[ ] PyTorch Lightning compatibility check

Currently this sampling pipeline does not support prefetching of writable embedding (e.g. trainable node/edge embedding) yet.

## Checklist
- [ ] Code is well-documented


python/dgl/dataloading/base.py
```diff
@@ -0,0 +1,221 @@
...
+        self.prefetch_node_feats = prefetch_node_feats or []
+        self.prefetch_labels = prefetch_labels or []
+        self.prefetch_edge_feats = prefetch_edge_feats or []
+        self.output_device = output_device
+
+    def sample_blocks(self, g, seed_nodes, exclude_edges=None):
+        raise NotImplementedError
+
+    def assign_lazy_features(self, result):
+        # A LazyFeature is a placeholder telling the dataloader where and which IDs
```
nv-dlasalle, line 28: I think we need to adjust the naming here as `lazy fetching` and `prefetching` have the opposite meanings.

examples/pytorch/__temporary__/dglnew/graph/other_feature.py
```diff
@@ -0,0 +1,26 @@
...
+
+    @property
+    def edata(self):
+        return self._edata
+
+    def __getattr__(self, key):
+        # I wrote it in this way because I'm too lazy to write "def sample_neighbors"
+        # or stuff like that.
+        if key in ['ntypes', 'etypes', 'canonical_etypes', 'sample_neighbors',
+                   'subgraph', 'edge_subgraph', 'find_edges', 'num_nodes']:
```
nv-dlasalle, line 22: One issue with this design is if I come up with a custom sampler (e.g., TimeSensitiveEdgeDataLoader), which uses the custom method (e.g., `sample_historical_neighbors()`) which I've added to my GraphStorage class (probably a child of DGLGraph, but perhaps not), the sampler would no longer be able to invoke the custom sampling function since it's not in this list.

Could we decouple our GraphStorage and FeatureStorage, such both get passed into a dataloader (for backward compatibility allow passing in `feature_storage=None` if the GraphStorage object is also a FeatureStore-like object)?

python/dgl/storages/pytorch_tensor.py
```diff
@@ -0,0 +1,35 @@
...
+def _fetch_cpu(indices, tensor, feature_shape, device, pin_memory):
+    result = torch.empty(
+        indices.shape[0], *feature_shape, dtype=tensor.dtype,
+        pin_memory=pin_memory)
+    torch.index_select(tensor, 0, indices, out=result)
+    result = result.to(device, non_blocking=True)
+    return result
+
+def _fetch_cuda(indices, tensor, device):
+    return torch.index_select(self.tensor, 0, indices).to(device)
```
nv-dlasalle, line 15: `self.tensor` -> `tensor`

nv-dlasalle:
It seems there's somewhat of a regression concerning storing the graph in the GPU. I see about a 25% slowdown (the prefetching thread needs to be disabled, or it's about 200% slowdown). It looks like the slowdown mostly has to do with the index selects being launched on a different stream and pytorch's allocator consistently calling `cudaMalloc()` plug `cudaFree()`. This is with `use_asyncio=False`, and remove `prefetch_node_feats` and `prefetch_labels` didn't seem to have any effect on this:
extra_mallocs_on_second_stream

examples/pytorch/__temporary__/dglnew/graph/graph.py
```diff
@@ -0,0 +1,181 @@
...
+        and the inner keys are canonical edge types or edge type names:
+
+        .. code::
+
+           self.edata[feature_name][etype]
+
+        Note that to maximize the compatibility with other DGL code it's better to support
+        both canonical edge types and edge type names as inner keys since users might use both.
+        """
+        pass
```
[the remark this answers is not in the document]
BarclayII: Interface-wise I think `ndata` and `edata` should be stayed as-is rather than adding a new interface.  The graph storage's implementation is free to support writing to edata and ndata, etc., and should report error there if it's not supported.

examples/pytorch/__temporary__/dglnew/graph/graph.py
```diff
@@ -0,0 +1,181 @@
...
+
+    # Required in Link Prediction
+    @property
+    def etypes(self):
+        """The list of edge types."""
+        pass
+
+    # Required in Link Prediction
+    @property
+    def canonical_etypes(self):
```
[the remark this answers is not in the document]
BarclayII: I doubt it.  While having a single `etype` is more friendly we did saw people having the same name of `etype` but connecting different source and destination node types.

examples/pytorch/__temporary__/dglnew/graph/graph.py
```diff
@@ -0,0 +1,181 @@
...
+            * iterable[int]: Each element is a node ID.
+            * Bool Tensor: Each :math:`i^{th}` element is a bool flag indicating whether
+              node :math:`i` is in the subgraph.
+
+            If the graph is homogeneous, one can directly pass the above formats.
+            Otherwise, the argument must be a dictionary with keys being node types
+            and values being the node IDs in the above formats.
+        relabel_nodes : bool, optional
+            If True, the extracted subgraph will only have the nodes in the specified node set
+            and it will relabel the nodes in order.
```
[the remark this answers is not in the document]
BarclayII: This is not generating a block.  It's generating a subgraph (e.g. for subgraph samplers).

python/dgl/dataloading/neighbor_sampler.py
```diff
@@ -0,0 +1,27 @@
...
+        blocks = []
+        for fanout in reversed(self.fanouts):
+            frontier = g.sample_neighbors(
+                seed_nodes, fanout, edge_dir=self.edge_dir, prob=self.prob,
+                replace=self.replace, output_device=self.output_device,
+                exclude_edges=exclude_edges)
+            eid = frontier.edata[EID]
+            block = to_block(frontier, seed_nodes)
+            block.edata[EID] = eid
+            seed_nodes = block.srcdata[NID]
```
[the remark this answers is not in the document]
BarclayII: If the graph is on GPU, does passing a GPU indices array to the dataloader solve the problem?

python/dgl/frame.py
```diff
@@ -38,6 +38,19 @@ def flatten(self):
...
         return flat_index
 
+class LazyFeature(object):
+    __slots__ = ['name', 'id_']
+    def __init__(self, name=None, id_=None):
+        self.name = name
+        self.id_ = id_
+
+    def to(self, *args, **kwargs):
+        return self
```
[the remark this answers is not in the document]
BarclayII: `LazyFeature` does not actually contain any feature so it should not involve any copying when a graph is copied to another device.  The actual features will be sliced by calling `FeatureStorage.fetch(LazyFeature.id_, device)` in the DataLoader so there is no inference of the output device.  I put this method here so that `Frame.to()` would not break.

python/dgl/dataloading/base.py
BarclayII: The purpose of this class is indeed for prefetching though...

examples/pytorch/__temporary__/dglnew/graph/other_feature.py
BarclayII: > One issue with this design is if I come up with a custom sampler (e.g., TimeSensitiveEdgeDataLoader), which uses the custom method (e.g., sample_historical_neighbors())
Do you have code online?  `TimeSensitiveEdgeDataLoader` sounds just like what TGAT/TGN needed.  I would try making it work with DGLGraphs if necessary.

> the sampler would no longer be able to invoke the custom sampling function since it's not in this list.
This `__getattr__` implementation is just an alternative to:
```
@property
def ntypes(self):
    return self.g.ntypes

@property
def etypes(self):
    return self.g.etypes
...
```
It's **not** meant to be a spec for all other graph storages like "all graph storages must have and only have this list of methods".  If you have a new sampling primitive for your GraphStorage, the sampler could still call your custom sampling function via duck-typing.  Then it will be DGL's job to catch up with your new primitive.

> Could we decouple our GraphStorage and FeatureStorage, such both get passed into a dataloader (for backward compatibility allow passing in feature_storage=None if the GraphStorage object is also a FeatureStore-like object)?
What is the use case?  I didn't quite understand why a GraphStorage could also be a FeatureStorage.

BarclayII:
> It seems there's somewhat of a regression concerning storing the graph in the GPU. I see about a 25% slowdown (the prefetching thread needs to be disabled, or it's about 200% slowdown). It looks like the slowdown mostly has to do with the index selects being launched on a different stream and pytorch's allocator consistently calling `cudaMalloc()` plug `cudaFree()`. This is with `use_asyncio=False`, and remove `prefetch_node_feats` and `prefetch_labels` didn't seem to have any effect on this:

Many thanks for benchmarking.  I think we should at least add a `use_alternate_stream` option so that we do not launch on different streams.  This option could default to False if everything is on GPU and True if the graph is on CPU.

BarclayII:
@nv-dlasalle Could you verify if the slowdown is indeed because of index slicing on different streams?  I have included a `use_alternate_streams` option which should turn off index select on different streams.

examples/pytorch/__temporary__/dglnew/graph/graph.py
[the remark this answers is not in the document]
nv-dlasalle: My big concern is implementing this correctly and efficiently is quite challenging.

Not only is there the combination of homogeneous graph vs. heterogeneous graph, etype vs. canonical etype, but then should each inner dictionary only contain the etypes and canonical etypes that have that feature, or should they have empty/None tensors associated with them? Also, the DGLHeteroGraph currently returns special view objects rather than dictionaries, which have hooks to allow modifying the underlying graph data, but for storage that return newly created dictionaries, such modifications would silently have no effect.

In terms of performance, creating dictionaries and new objects is quite expensive in python (in the context of mini-batch generation), this interface requires two-layers of objects just to get to the point of slicing a feature tensor.
However if we use a single function, not only will I not need to worry about the differences between homogeneous graphs and heterogeneous graphs, I can immediately drop into C code for efficiency:
```
def get_edge_features(feature_name, eids, output_device, edge_type):
  return _CAPI_MyCustomFunction(feature_name, eids, output_device, edge_type)
```
Which if our dataloader goes through each feature on each edge type, will be significantly faster than instantiating two levels of dictionaries / dictionary-like objects for each access.

examples/pytorch/__temporary__/dglnew/graph/graph.py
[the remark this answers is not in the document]
nv-dlasalle: Sorry, I missed that part. I don't think this is important for 0.8, but I think for 1.0 it would make sense to make it consistent between layer-wise samplers relying on the dataloader to call to_block() and having subgraph samplers rely on the dataloader to relabel the nodes (especially since it is a separate step anyways: https://github.com/dmlc/dgl/blob/master/src/graph/unit_graph.cc#L377, https://github.com/dmlc/dgl/blob/master/src/graph/unit_graph.cc#L797).

python/dgl/dataloading/dataloader.py
```diff
@@ -570,3 +574,190 @@ def __init__(self, graph, indices, graph_sampler, device='cpu', use_ddp=False,
             batch_size=batch_size, drop_last=drop_last, shuffle=shuffle, use_asyncio=use_asyncio,
             use_prefetch_thread=use_prefetch_thread, use_alternate_streams=use_alternate_streams,
             **kwargs)
+
+
```
nv-dlasalle, line 578: Was the below code meant for dgl/dataloading/pytorch/dataloader.py since it is pytorch specific? It's currently missing the import for `LooseVersion` and `th`.

nv-dlasalle:
> @nv-dlasalle Could you verify if the slowdown is indeed because of index slicing on different streams? I have included a `use_alternate_streams` option which should turn off index select on different streams.

Setting that to false makes it pretty close to the previous behavior. Thanks!

Although not taking much extra time, I do see eight extra index selects happening before the forward pass, and then 9 device-to-device copies happening at the end of the backward pass, that aren't there in the previous version of graphsage. I haven't tracked down exactly where they are coming from, so they may not be related to the new dataloader (but instead to the differences between `ddp.py` and `train_sampling_multi_gpu.py`).

examples/pytorch/__temporary__/dglnew/graph/graph.py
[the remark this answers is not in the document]
jermainewang: I think the major suggestion is to restrict the interface a little bit, e.g., disallow `g_store.ndata[feat_name]` which returns a dictionary of feature storage for each node/edge type. I like the idea but will tweak the proposal a little bit. How about:
```python
def get_ndata(name : str, ntype : str):
    # Get the feature storage associated with the given name and node type.
    pass

def get_edata(name : str, etype : Tuple[str, str, str]):
    # Get the feature storage associated with the given name and edge type.
    # Edge type is a string triplet.
    pass
```
Slicing will typically happen subsequently on the returned feature storage. See the `fetch` interface in the RFC.

examples/pytorch/__temporary__/dglnew/graph/graph.py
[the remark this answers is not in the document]
jermainewang: There seem to be some new requirements in addition to the current graph storage interface proposal (e.g., `ntypes`, `etypes`, `canonical_etypes`), when coming to heterograph. We perhaps need a thorough discussion on that.

examples/pytorch/__temporary__/dglnew/graph/other_feature.py
jermainewang: Is `OtherFeatureGraphStorage` your testing code or something that will be checked into the main repo?

BarclayII:
Due to the size and urgency of the PR I'm compromising a bit, and I'll fix them in later PRs if possible:
* The docstrings are either minimal or copied from previous implementations, and therefore may not be correct.
* Did not implement the `use_ddp` check.
* `ntypes` and `canonical_etypes` are still left in GraphStorage's spec.  `etypes` is no longer necessary though and the edge data will be indexed by canonical edge types.
* `use_alternate_streams` flag is left there and I'll see if I can remove it later.  This may need some changes in the DataLoader implementation but it should not affect other components.
* I no longer use `asyncio` for asynchronous operations.  Instead I put up a `ThreadedFuture` object that has a `wait()` method.  It is also a minimal implementation that spawns threads eagerly (i.e. without a thread pool), so there's definitely space for optimization.

python/dgl/dataloading/dataloader.py
```diff
@@ -1,900 +1,602 @@
...
+        type_ = types[tid]
+        default_id = frame.get(id_name, None)
+        for key in frame.keys():
+            column = frame[key]
+            if isinstance(column, LazyFeature):
+                parent_key = column.name or key
+                if column.id_ is None and default_id is None:
+                    raise DGLError(
+                        'Found a LazyFeature with no ID specified, '
+                        'and the graph does not have dgl.NID or dgl.EID columns')
```
jermainewang, line 239: I wonder how to fix it if a user got such an error? Consider reformatting the error message as follows:
```python
Feature "{key}" is a LazyFeature but with no ID specified. Please create a LazyFeature with ....
```

python/dgl/dataloading/dataloader.py
```diff
@@ -1,900 +1,602 @@
...
+                feats[tid, key] = get_storage_func(parent_key, type_).fetch(
+                    column.id_ or default_id, device, pin_memory)
+
+
+# This class exists to avoid recursion into the feature dictionary returned by the
+# prefetcher when calling recursive_apply().
+class _PrefetchedGraphFeatures(object):
```
jermainewang, line 246: Could make it a namedtuple.

python/dgl/dataloading/dataloader.py
```diff
@@ -1,900 +1,602 @@
...
+        edge_feats, subg._edge_frames, subg.canonical_etypes, dataloader.graph.get_edge_storage,
+        EID, dataloader.device, dataloader.pin_memory)
+    return _PrefetchedGraphFeatures(node_feats, edge_feats)
+
+
+def _prefetch_for(item, dataloader):
+    if isinstance(item, DGLHeteroGraph):
+        return _prefetch_for_subgraph(item, dataloader)
+    elif isinstance(item, LazyFeature):
+        return dataloader.other_storages[item.name].fetch(
```
jermainewang, line 268: Does `dataloader` always have an `other_storages` attribute?

python/dgl/dataloading/dataloader.py
```diff
@@ -1,900 +1,602 @@
...
-            Note that the metagraph of this graph will be identical to that of the original
-            graph.
-        blocks : list[DGLGraph]
-            The list of MFGs necessary for computing the representation of the edges.
-        """
-        if self.negative_sampler is None:
-            return self._collate(items)
-        else:
-            return self._collate_with_negative_sampling(items)
+            dgl_warning(
```
jermainewang, line 565: Throw error instead of warning?

python/dgl/dataloading/dataloader.py
```diff
@@ -975,8 +677,89 @@ def collate(self, items):
...
+            self.collate = collate_fn
+
+        self.use_ddp = use_ddp
+        if use_ddp:
+            self.dist_sampler = _create_dist_sampler(dataset, dataloader_kwargs, ddp_seed)
+            dataloader_kwargs['sampler'] = self.dist_sampler
+
+        super().__init__(dataset=dataset, collate_fn=self.collate, **dataloader_kwargs)
+
+    def set_epoch(self, epoch):
```
jermainewang, line 749: Is this deprecated? If so, add warning.

python/dgl/dataloading/base.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def assign_lazy_features(self, result):
+        """Assign lazy features for prefetching."""
+        # A LazyFeature is a placeholder telling the dataloader where and which IDs
+        # to prefetch.  It has the signature LazyFeature(name, id_).  id_ can be None
+        # if the LazyFeature is set into one of the subgraph's ``xdata``, in which case the
+        # dataloader will infer from the subgraph's ``xdata[dgl.NID]`` (or ``xdata[dgl.EID]``
+        # if the LazyFeature is set as edge features).
+        #
+        # If you want to prefetch things other than ndata and edata, you can also
```
jermainewang, line 76: Just a suggestion. Try to avoid second person in the docstring or description. Use "users "or just omit subjects.

jermainewang (approved):

jermainewang:
Approve to unblock development need this. You could merge it and address my comments later.

python/dgl/dataloading/dataloader.py
BarclayII: This is from GraphDataLoader so it's not deprecated.

python/dgl/dataloading/dataloader.py
BarclayII: Not for `GraphDataLoader`.

python/dgl/_dataloading/dataloader.py
```diff
@@ -0,0 +1,982 @@
...
+    elif exclude_mode == 'self':
+        return eids
+    elif exclude_mode == 'reverse_id':
+        return _find_exclude_eids_with_reverse_id(g, eids, kwargs['reverse_eid_map'])
+    elif exclude_mode == 'reverse_types':
+        return _find_exclude_eids_with_reverse_types(g, eids, kwargs['reverse_etype_map'])
+    else:
+        raise ValueError('unsupported mode {}'.format(exclude_mode))
+
+class Sampler(object):
```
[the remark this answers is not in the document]
BarclayII: `_dataloading` is legacy code and will be removed later.

examples/pytorch/__temporary__/rgat/rgat.py
```diff
@@ -0,0 +1,105 @@
...
+dataloader = dgl.dataloading.NodeDataLoader(
+        graph,
+        train_idx,
+        sampler,
+        device='cuda',
+        batch_size=1000,
+        shuffle=True,
+        drop_last=False,
+        pin_memory=True,
+        num_workers=8,
```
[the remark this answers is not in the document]
BarclayII: I don't think so.

python/dgl/dataloading/shadow.py
```diff
@@ -70,29 +68,32 @@ class ShaDowKHopSampler(NeighborSamplingMixin, Sampler):
...
+                seed_nodes, fanout, output_device=self.output_device,
+                replace=self.replace, prob=self.prob, exclude_edges=exclude_edges)
             block = transform.to_block(frontier, seed_nodes)
             seed_nodes = block.srcdata[NID]
 
-        subg = g.subgraph(seed_nodes, relabel_nodes=True)
-        subg = exclude_edges(subg, exclude_eids, self.output_device)
+        subg = g.subgraph(seed_nodes, relabel_nodes=True, output_device=self.output_device)
+        if exclude_edges is not None:
+            subg = EidExcluder(exclude_edges)(subg)
```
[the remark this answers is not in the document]
BarclayII: `exclude_edges` can change for every minibatch.

python/dgl/dataloading/dataloader.py
BarclayII: This only happens when people wants to prefetch features that are neither ndata nor edata.  In that case he/she needs to specify IDs in the LazyFeature object as well.

python/dgl/dataloading/dataloader.py
BarclayII: A namedtuple is subject to `recursive_apply` which will return a list instead.

VoVAllen:
🚀

PR #3674, opened 2022-01-21
[Feature] CUDA UVA sampling for MultiLayerNeighborSampler

yaox12:
## Description

This enables single- and multi-GPU CUDA UVA sampling for `MultiLayerNeighborSampler` combined with both `NodeDataLoader`  and `EdgeDataLoader`. 

## Checklist

- [x] Code is well-documented
- [x] Results on 1, 2, 4, and 8x A100s with GraphSAGE and OGBN-Papers100M

## Changes
- Enable CUDA UVA operator with `ATEN_XPU_SWITCH_CUDA_UVA`.
- GraphSAGE examples using CUDA UVA sampling.

We need to specify a CUDA context to perform GPU sampling on the pinned graph, so we assume `nid` (or `seed_nodes`) is on the GPU, which breaks the previous assumption that `nid` must be on the same device as the graph. To make it, I skip some device check for pinned graphs on the Python side.

For CSR operators, my current strategy is:
- For the ops that have one (or more) `NDArray` as the second argument, e.g., `op(CSRMatrix csr, NDArray row, …)`, we dispatch it using the context of `row` and dtype of `csr.indprt`. In this way, the computing context of a pinned CSR matrix is dominated by that of `row`, which is identical to the behavior of UnifiedTensor.
- For the ops that do not have `NDArray` as arguments, e.g. `op(CSRMatrix csr, int64_t row, …)`, we keep it as it is. These ops will treat the pinned matrix as ordinary objects on host memory.

## Performance

Benchmark GraphSAGE training on a machine with 128 CPU cores and 4x A100 GPU cards.

### Supervised GraphSAGE

- 1x A100

| sampling | epoch time (s) | GPU memory (MiB) |              |
|----------|------------|------------------|--------------|
| CPU      |       3.45 |             2553 | num_worker=6 |
| CPU      |       2.04 |             2609 | num_worker=0 |
| GPU      |       1.35 |             5805 |              |
| UVA      |       1.63 |             3837 |              |

- 4x A100

| sampling | epoch time (s) | GPU memory (MiB) |              |
|----------|------------|------------------|--------------|
| CPU      |       2.32 |             3704 | num_worker=6 |
| CPU      |       0.60 |             3758 | num_worker=0 |
| GPU      |       0.38 |             6812 |              |
| UVA      |       0.45 |             4910 |              |

Note:
1. `OMP_NUM_THREADS=8` is set for all tests.
2. `num_workers=0` will conduct `to_block()` on the GPU, and is therefore much faster.

### Unsupervised GraphSAGE

- 1x A100

| sampling | epoch time (s) | GPU memory (MiB) |              |
|----------|------------|------------------|--------------|
| CPU      |    2723.08 |             5055 | num_worker=6 |
| CPU      |    6257.76 |             6575 | num_worker=0 |
| GPU      |     301.88 |            15833 |              |
| UVA      |     748.51 |            10119 |              |

- 4x A100

| sampling | epoch time (s) | GPU memory (MiB) |              |
|----------|------------|------------------|--------------|
| CPU      |     779.65 |             6184 | num_worker=6 |
| CPU      |    1663.02 |             6676 | num_worker=0 |
| GPU      |      72.69 |            16752 |              |
| UVA      |     169.20 |            11086 |              |

Note:
1. `OMP_NUM_THREADS=8` is set for all tests.
2. For `EdgeDataLoader`, there is an expensive `compact_graph()` running on CPU even if `num_workers=0` is specified.


VoVAllen:
I think we have `compact_graph ` operator on GPU. Can it happen on GPU?

VoVAllen:
Also why UVA sampling will use more GPU memory than CPU sampling?

yaox12:
> I think we have `compact_graph` operator on GPU. Can it happen on GPU?

`compact_graph` is conducted on GPU when using GPU sampling for now. Currently, it will not work on GPU with CPU sampling. 

`to_block` does because it's called inside the block sampler and we have an `output_device` argument to tell it to copy the frontier to GPU and do `to_block` there. 
`compact_graph` is called in the collator, we don't have a way to tell it to do `compatc_graph` on GPU while the input graph is on CPU. 

I don't find an elegant solution for it so I left it here.

yaox12:
> Also why UVA sampling will use more GPU memory than CPU sampling?

Because the sampling kernels are executed on GPU. The intermediate variables, sampled frontiers and block construction are all on GPU. You can also find that it takes more GPU memory when `to_block` is done on GPU even with CPU sampling.

VoVAllen:
I discussed with Quan. Ideally the graph before compact_graph should be on GPU already.

That at here https://github.com/dmlc/dgl/blob/b8ce0f41a3d440aa367e076e2c1556e26bce383d/python/dgl/dataloading/dataloader.py#L822, the edge_subgraph can take an `output_device` argument, to put the output graph on device.
And you can pass the device into EdgeCollator in https://github.com/dmlc/dgl/blob/b8ce0f41a3d440aa367e076e2c1556e26bce383d/python/dgl/dataloading/pytorch/dataloader.py#L823

examples/pytorch/graphsage/train_sampling.py
```diff
@@ -52,13 +52,17 @@ def run(args, device, data):
...
-    if args.sample_gpu:
+    if args.sample_device == 'gpu':
         train_nid = train_nid.to(device)
         # copy only the csc to the GPU
         train_g = train_g.formats(['csc'])
         train_g = train_g.to(device)
-        dataloader_device = device
+        args.num_workers = 0
+    elif args.sample_device == 'uva':
+        train_nid = train_nid.to(device)
```
nv-dlasalle, line 62: Would this impact performance significantly if `train_nid` were a unified tensor rather than being stored on the device?

yaox12: @nv-dlasalle In fact `train_nid` has to be on the GPU because we use the context of it (or `seed_nodes` more precisely) to execute the sampling kernel.

yaox12: Actually, I feel that the device switching for sampling and subgraph construction (`to_block`) is a bit complicated now. I summarize it as a table and plan to put it in the doc.

<table>
<thead>
  <tr>
    <th>graph device</th>
    <th>nids device</th>
    <th>dataloader device</th>
    <th>num_workers</th>
    <th>sampling</th>
    <th>subgraph construction</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td colspan="2">GPU</td>
    <td>GPU</td>
    <td>0</td>
    <td>GPU</td>
    <td>GPU</td>
  </tr>
  <tr>
    <td colspan="2" rowspan="2">CPU</td>
    <td rowspan="2">GPU</td>
    <td>0</td>
    <td>CPU</td>
    <td>GPU</td>
  </tr>
  <tr>
    <td>&gt;0</td>
    <td rowspan="2">CPU</td>
    <td>CPU</td>
  </tr>
  <tr>
    <td colspan="2">CPU</td>
    <td>CPU</td>
    <td>any</td>
    <td>CPU</td>
  </tr>
  <tr>
    <td>Pinned</td>
    <td>GPU</td>
    <td>GPU</td>
    <td>0</td>
    <td>GPU</td>
    <td>GPU</td>
  </tr>
</tbody>
</table>

nv-dlasalle: While we may have to do some tuning to ensure its faster in all cases, I'd be in favor of deprecating the normal sample-on-cpu to-block-on-gpu in favor UVA sampling, to keep things simple.

python/dgl/dataloading/dataloader.py
```diff
@@ -436,7 +436,10 @@ def sample(self, g, seed_nodes, exclude_eids=None):
             graph_device = g.device
 
         for block_id in reversed(range(self.num_layers)):
-            seed_nodes_in = to_device(seed_nodes, graph_device)
+            if not g.is_pinned():
+                seed_nodes_in = to_device(seed_nodes, graph_device)
```
yaox12, line 440: In fact I have a question here. AFAIK, `seed_nodes` here usually has been processed by `utils.prepare_tensor_or_dict()`, for example:
https://github.com/dmlc/dgl/blob/08f464348f5f4855941130356d2db9da7ba45ab9/python/dgl/dataloading/dataloader.py#L584-L586
So when would `seed_nodes` has a different context from `graph_device`? I tried commenting out this line and it threw an error.

examples/pytorch/graphsage/train_sampling.py
yaox12: Should we include it in this PR? Another question is what to do with the `output_device` argument in `BlockSampler`? I think it is introduced to enable the hybrid mode.

nv-dlasalle: I think we could simplify it as:

<html><body>


graph device | nids device | dataloader device | num_workers | sampling | subgraph construction
-- | -- | -- | -- | -- | --
GPU | GPU | GPU | 0 | GPU | GPU
CPU | CPU | CPU | any | CPU | CPU
Pinned | GPU | GPU | 0 | GPU | GPU


</body>
</html>

We could also simplify this further, by having the DataLoader copy the nids to the proper device for each mini-batch, especially since relative to the output of the mini-batch they should be quite small.

python/dgl/dataloading/dataloader.py
yaox12: I see. When we might deprecate the CPU-GPU hybrid sampling in the future, we can also simplify this.

BarclayII: I think we will deprecate CPU-GPU hybrid sampling as soon as this PR is merged.

include/dgl/aten/macro.h
```diff
@@ -232,6 +232,14 @@
     });                                                     \
   });
 
+// Dispatching according to the context of array to enable CUDA UVA
+#define ATEN_CSR_SWITCH_CUDA_UVA(csr, array, XPU, IdType, op, ...) \
```
BarclayII, line 236: Is it possible to check whether the graph is pinned in this macro?

yaox12: Can you explain it a bit more? Do you mean to put the `CHECK_VALID_CONTEXT` macro inside the  `ATEN_CSR_SWITCH_CUDA_UVA` macro to avoid confusion with `CHECK_SAME_CONTEXT` when adding a new op or something else?

BarclayII: > Do you mean to put the `CHECK_VALID_CONTEXT` macro inside the `ATEN_CSR_SWITCH_CUDA_UVA` macro to avoid confusion with `CHECK_SAME_CONTEXT` when adding a new op

Yes.

python/dgl/_dataloading/pytorch/dataloader.py
```diff
@@ -599,9 +600,23 @@ def __init__(self, g, nids, graph_sampler, device=None, use_ddp=False, ddp_seed=
...
                 dataloader_kwargs[k] = v
 
-        if device is None:
-            # default to the same device the graph is on
-            device = th.device(g.device)
+        # default to the same device the graph is on
+        device = th.device(g.device if device is None else device)
+        num_workers = dataloader_kwargs.get('num_workers', 0)
+
+        if F.device_type(g.device) == 'cuda' or g.is_pinned():
```
VoVAllen, line 607: You don't need F.device_type here due to dataloaders are only supported in PyTorch. You can use `g.device.type == 'cuda'` directly.

VoVAllen (approved):
Generally looks very well to me. 

One question is that the uva sampling scenario is not covered in the current unit tests. We need to add specific situations for graph/tensor on UVA and computation on GPU.

VoVAllen:
Also could you find a dataset or large model that the full graph training on GPU in not viable? For example ogbn-product with  GAT, or even larger dataset, that is impossible to train by putting whole graph into GPU.
Then have some number comparing between cpu sampling + copy to GPU, etc, for the later press release? Thanks

yaox12:
I have the numbers of supervised GraphSAGE on ogbn-papers100M with 4x A100 but they are under the old data loaders. 
<table>
<thead>
  <tr>
    <th>sampling device</th>
    <th>data device</th>
    <th>epoch time (s)</th>
    <th>GPU memory (MiB)</th>
    <th></th>
  </tr>
</thead>
<tbody>
  <tr>
    <td rowspan="2">CPU</td>
    <td>CPU</td>
    <td>44.13</td>
    <td>2533</td>
    <td rowspan="2">num_workers=6</td>
  </tr>
  <tr>
    <td>UnifiedTensor</td>
    <td>39.62</td>
    <td>2715</td>
  </tr>
  <tr>
    <td rowspan="2">CPU</td>
    <td>CPU</td>
    <td>50.83</td>
    <td>2633</td>
    <td rowspan="2">num_workers=0</td>
  </tr>
  <tr>
    <td>UnifiedTensor</td>
    <td>29.67</td>
    <td>2741</td>
  </tr>
  <tr>
    <td rowspan="2">UVA</td>
    <td>CPU</td>
    <td>19.23</td>
    <td>4015</td>
    <td></td>
  </tr>
  <tr>
    <td>UnifiedTensor</td>
    <td>4.09</td>
    <td>4123</td>
    <td></td>
  </tr>
</tbody>
</table>

I will update them later.

examples/pytorch/graphsage/README.md
```diff
@@ -71,6 +71,20 @@ Notably,
 
 Micro F1 score reaches 0.9212 on test set.
 
+### Use GPU sampling and CUDA UVA sampling
+
+For training scripts `train_sampling.py`, `train_sampling_multi_gpu.py` and `train_sampling_unsupervised.py`, we provide arguments `--sample-device` and `--data-device`.
+
+For `--sample-device`, we provide the following choices:
```
VoVAllen, line 78: Also is it better to call it graph device instead of sample device?

yaox12: Agreed.

include/dgl/aten/macro.h
yaox12: Resolved.

VoVAllen:
Generally I think this PR is ok to merge. Please add the unit test in https://github.com/dmlc/dgl/blob/90f10b31cb4bbbb6579f929a1ce5abd2dd2a4fe8/tests/pytorch/test_dataloader.py#L436 to cover the uva scenarios. Thanks!

BarclayII (approved):

Issue #3646, opened 2022-01-12
NameError: name '_CAPI_DGLNCCLGetUniqueId' is not defined


mufeili:
## 🐛 Bug

> Traceback (most recent call last):
  File "entity_sample.py", line 196, in <module>
    main(args)
  File "entity_sample.py", line 153, in main
    labels, emb_optimizer, optimizer)
  File "entity_sample.py", line 98, in train
    emb_optimizer.step()
  File "/home/ubuntu/anaconda3/envs/RGCN/lib/python3.7/site-packages/dgl-0.8-py3.7-linux-x86_64.egg/dgl/optim/pytorch/sparse_optim.py", line 80, in step
    self._comm_setup()
  File "/home/ubuntu/anaconda3/envs/RGCN/lib/python3.7/site-packages/dgl-0.8-py3.7-linux-x86_64.egg/dgl/optim/pytorch/sparse_optim.py", line 109, in _comm_setup
    self._comm = nccl.Communicator(1, 0, nccl.UniqueId())
  File "/home/ubuntu/anaconda3/envs/RGCN/lib/python3.7/site-packages/dgl-0.8-py3.7-linux-x86_64.egg/dgl/cuda/nccl.py", line 22, in __init__
    self._handle = _CAPI_DGLNCCLGetUniqueId()
NameError: name '_CAPI_DGLNCCLGetUniqueId' is not defined

## To Reproduce

1. `git clone https://github.com/mufeili/dgl.git -b simplify`
1. Install DGL from source
1. `cd dgl/examples/pytorch/rgcn`
1. `python entity_sample.py -d am --n-bases 40 --gpu 0 --fanout '35,35' --batch-size 64 --n-hidden 16 --use-self-loop --n-epochs=20 --dgl-sparse --sparse-lr 0.02 --dropout 0.7`

## Expected behavior

Running without error

## Environment

 - DGL Version (e.g., 1.0):
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): PyTorch 1.10.0
 - OS (e.g., Linux): Linux
 - How you installed DGL (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version: 3.8
 - CUDA/cuDNN version (if applicable): 10.2
 - GPU models and configuration (e.g. V100):
 - Any other relevant information:

nv-dlasalle:
Can you specify the full command used to build DGL?

mufeili:
[...]

PR #3713, opened 2022-02-05
[bugfix] Allow communicators of size one when NCCL is missing

nv-dlasalle:
## Description
This is the first of two PRs to address #3646. This enables the NCCL communicator python API, but if NCCL is not compiled in, limits to size 1 communicators (where operations are just a memcpy).

A second PR will update the SparseOptimizer to use shared memory if NCCL is not enabled.

## Checklist
- [x] Code is well-documented

## Changes

This allows the Communicators functions to work for communicators of size one if NCCL is missing, and then also adds a python function for checking if NCCL support is enabled.


src/runtime/cuda/nccl_api.cu
```diff
@@ -601,6 +612,24 @@ void NCCLCommunicator::AllToAllV(
     }
   }
   NCCL_CALL(ncclGroupEnd());
+  #else
+  CHECK_EQ(send_prefix[1]-send_prefix[0], recv_prefix[1]-recv_prefix[0]) <<
+      "Send message size must equal receive message size.";
```
classicsong, line 617: Does this due to the limitation of device->CopyDataFromTo?

classicsong (commented):
Is it possible to write some test code?

src/runtime/cuda/nccl_api.cu
nv-dlasalle: Yes, but I'm not sure I'd call it a limitation. Since our communicator is size 1 in the case, there is only one message, so it's size must be identical both from the buffer it's being copied from and the buffer it's being copied to.

jermainewang:
The RGCN example codes have been reworked for simplicity. Currently put this PR on hold. Need to find a better place for all examples with advanced optimizations.

BarclayII:
@nv-dlasalle @classicsong Do you need this for 0.9?

python/dgl/cuda/nccl.py
```diff
@@ -210,4 +210,14 @@ def size(self):
         """
         return self._size
 
+def is_supported():
```
jermainewang, line 213: Is `dgl.cuda.nccl` included in the API doc? We could say the package is in experimental stage.

yaox12: It's not exposed to users directly for now and used in `dgl.nn.NodeEmbedding` and `dgl.optim.SparseAdam` only, where we have a docstring saying it's experimental.

yaox12 (approved):

jermainewang:
I'm fine with the PR too. Strangely I didn't see the newly added `dgl.nccl.is_supported` being used anywhere.

yaox12:
`dgl.nccl.is_supported` is not used yet but for potential checks I think. We will copy things with native device API if NCCL is missing for now.

nv-dlasalle:
> I'm fine with the PR too. Strangely I didn't see the newly added `dgl.nccl.is_supported` being used anywhere.

It is meant to be used in the SparseAdam optimizer, where it would use shared memory on the host if NCCL is not available.

PR #3720, opened 2022-02-08
[Performance] Workaround CSR conversion barrier stalls

daniil-sizov:
This PR works around OpenMP barrier stalls observed on multiple workloads with 2+ dataloader workers.

If merged, OpenMP will be disabled for small sparse matrix (<1024 non-zero elements) CSR conversion
The change has no negative performance impact outside of barrier stalls issue since overhead on thread management exceeds the gain from parallelizing the computation.

### DIFF
--- a/src/array/cpu/spmat_op_impl_coo.cc
+++ b/src/array/cpu/spmat_op_impl_coo.cc
@@ -532,7 +532,7 @@ template <class IdType> CSRMatrix UnSortedDenseCOOToCSR(const COOMatrix &coo) {
   std::vector<std::vector<IdType>> local_ptrs;
   std::vector<int64_t> thread_prefixsum;
 
-#pragma omp parallel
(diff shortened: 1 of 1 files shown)


daniil-sizov:

VoVAllen:
Hi,

Could you open an issue to describe the rationale behind your PRs for us to understand?

Thanks.

daniil-sizov:
@VoVAllen I am observing significant OMP barrier stalls on some workloads (notably examples/pytorch/ogb/ogbn-products/graphsage), which are especially pronounced with dataloader affinity change (https://github.com/dmlc/dgl/pull/3355) enabled

NodeDataLoader CPU Affinity optimization gives very significant performance gains on some workloads, but suffers a lot from such barrier stalls. This PR is supposed to work around that with no negative performance impact since disabling parallel execution is a reasonable thing to do for small computations anyway.

Added description

PR #3723, opened 2022-02-09
[Feature] Dataloader worker affinitization

daniil-sizov:
Adds worker threads affinitization option for dataloader. Improves performance on CPU by ~3.5% on avg measured on

bench_gat_ns.track_time
bench_rgcn_hetero_ns.track_time
bench_rgcn_homogeneous_ns.track_time
bench_sage_ns.track_time


daniil-sizov:

VoVAllen:

VoVAllen:
Sorry the regression test can only be triggered by team member.

VoVAllen:
The benchmark code seems broken now due to API change

python/dgl/dataloading/dataloader.py
```diff
@@ -677,7 +678,8 @@ class DataLoader(torch.utils.data.DataLoader):
     def __init__(self, graph, indices, graph_sampler, device=None, use_ddp=False,
                  ddp_seed=0, batch_size=1, drop_last=False, shuffle=False,
                  use_prefetch_thread=None, use_alternate_streams=None,
-                 pin_prefetcher=None, use_uva=False, **kwargs):
+                 pin_prefetcher=None, use_uva=False,
+                 use_cpu_worker_affinity=False, cpu_worker_affinity_cores=None, **kwargs):
```
BarclayII, line 682: Is it possible to remove these two arguments (i.e. is there any case where we want to set affinity to False, or manually specify the cores?)

python/dgl/dataloading/dataloader.py
```diff
@@ -825,6 +827,26 @@ def __init__(self, graph, indices, graph_sampler, device=None, use_ddp=False,
...
+                raise Exception('ERROR: affinity should be used with --num_workers=X')
+            if len(cpu_worker_affinity_cores) not in [0, nw_work]:
+                raise Exception('ERROR: cpu_affinity incorrect '
+                                'settings for cores={} num_workers={}'
+                                .format(cpu_worker_affinity_cores, nw_work))
+
+            self.cpu_cores = (cpu_worker_affinity_cores
+                                if len(cpu_worker_affinity_cores)
+                                else range(0, nw_work))
+            worker_init_fn = self.worker_init_function
```
BarclayII, line 848: This will overwrite `worker_init_fn` above, so it disables the `WorkerInitWrapper` above that sets the number of OpenMP threads per process, and it will not invoke the custom `worker_init_fn`.  Is this intended?  If not, you could put the logic in `WorkerInitWrapper`.

BarclayII (approved):
Discussed with @daniil-sizov offline and I'll merge this for now.  The two arguments will be removed in future PRs.  LGTM except for one issue.

Issue #3706, opened 2022-02-02
Proper link prediction evaluation in DistDGL


bioannidis:
DistDGL does not allow two graphs for inference. In link prediction this will create problem of either not correct evaluation if testing edges are included in the graph or a lot of custom code to do it properly.  I have some ideas I can share with the team.

**Main concern**
Imagine you train a GNN and the testing edges exists in the graph. Naturally, you will see the performance in the training set improving as the model trains. You will also see the performance in the testing edge improve and at a certain epoch K it will start degrading. The problem is that the overfitting will happen much earlier than epoch K, but since the testing edges exist in the graph we will not be able to see it. This will lead to a huge problem in production where the inference happens usually in unobserved edges. 

The main solution here is to support a testing and a training graph in the DistDGL as we do in DGL.

BarclayII:
I think the easiest way would be adding a `always_exclude` argument to `EdgeDataLoader` that excludes the given edge IDs no matter what.  Then you could build a single graph that contains not only the training edges, but also the validation and test edges (including the positive and negative pairs).  Something like:

```python
train_dataloader = dgl.dataloading.EdgeDataLoader(
    g,
    train_eids,
    sampler,
    exclude='reverse_types',    # exclude the minibatch itself as well as the reverse edges of the reverse types
    reverse_etypes=...,
    always_exclude=torch.cat([val_eids, test_eids]), # always exclude the validation/test edges from neighbor sampling
    ...)
```

```python
val_dataloader = dgl.dataloading.EdgeDataLoader(
    g,
    val_eids,
    sampler,
    exclude='reverse_types',    # exclude the minibatch itself as well as the reverse edges of the reverse types
    reverse_etypes=...,
    always_exclude=test_eids, # always exclude the test edges from neighbor sampling
    ...)
```

Does that make sense?

PR #3727, opened 2022-02-11
[Bugfix] Bug fixes in new dataloader

BarclayII:
## Description
This PR makes GraphSAGE node classification example work on (1) pure CPU training, (2) UVA sampling, (3) CPU-sampling and prefetching to CUDA, (4) pure-GPU sampling.

Also
* Fixes a bug where graphs being passed to the subprocesses spawned with `mp.spawn` do not preserve CSR and CSC representations created in the main process.
* Fixes a bug where EdgeDataLoader does not properly exclude the edges sampled from the minibatch.
* Adds an `always_exclude` argument in `EdgeDataLoader` as well as `EdgeBlockSampler` as mentioned in #3706 - although it's only for single machine.

## Checklist
- [x] Code is well-documented


BarclayII:
Things to note:

* @jermainewang I'm not sure how we should organize the example with different sampling and training devices.  Currently I'm using a `MODE` variable, but of course we can change it or split it to multiple files.
* @yaox12 Could you also take a look and see if the UX makes sense?  You could also benchmark it and see if the performance is close to what you have.
* I also found that PyTorch still sets the DLPack device type as `kDLCPU` even if the tensor is pinned.  I'm working around it by patching the device type in `src/runtime/ndarray.cc::PatchCPUPinnedDeviceType()`, and checking whether the CPU tensor is indeed in pinned memory with `cudaPointerGetAttributes` in `src/runtime/cuda/cuda_device_api.cc.  Need your thoughts on this @jermainewang @yaox12 .

davidmin7:
> I also found that PyTorch still sets the DLPack device type as kDLCPU even if the tensor is pinned. I'm working around it by patching the device type in src/runtime/ndarray.cc::PatchCPUPinnedDeviceType(), and checking whether the CPU tensor is indeed in pinned memory with cudaPointerGetAttributes in src/runtime/cuda/cuda_device_api.cc. Need your thoughts on this @jermainewang @yaox12 .

I need to think a bit more, but seems like this case-study brought up an interesting corner case that I've never thought of. Basically how the DGL's ndarray should deal with the imported tensor which is pinned? As I have raised a question at https://github.com/dmlc/dgl/pull/3616#issuecomment-1002656117, the tensor pinned by external framework is probably created by using `cudaMallocHost()`, not by using the combination of `new` (or `malloc()`) + `cudaHostRegister()`.

Therefore, in this case, `unpin_memory_()` should not be called in this tensor, but it is difficult to figure out from DGL if this tensor is created with `cudaMallocHost()` or `cudaHostRegister()` after finishing `NDArray::FromDLPack()`.

I'll try to think through a little bit more and see if this is actually a valid concern. Please don't mind if this already handled and if I'm missing something obvious.

davidmin7:
And I can confirm that PyTorch does not care to explicitly set the `CPUPinned` flag when transforming to the DLPack tensor.

Where the ctx copying happens in PyTorch: https://github.com/pytorch/pytorch/blob/1776caf361fc2c652f0c00b2fd42714447cefc29/aten/src/ATen/DLConvertor.cpp#L227

The `getDLDevice()` function used in this line only checks the device type (e.g., CPU vs. GPU), not the property (e.g., pinned).

getDLDevice() function: https://github.com/pytorch/pytorch/blob/1776caf361fc2c652f0c00b2fd42714447cefc29/aten/src/ATen/DLConvertor.cpp#L71

yaox12:
@davidmin7 Your concerns are reasonable. I think one solution is that we make sure `unpin_memory_()` should only be called in pairs with `pin_memory_()`, though this relies on manual code review and can be fragile. For example in this data loader case, the PyTorch tensors are previously pinned by `F.to_dgl_nd(tensor).pin_memory_()`, so it is safe to call `unpin_memory_()`. If tensors were pinned outside DGL, `unpin_memory_()` should not be called on it.

yaox12:
> @yaox12 Could you also take a look and see if the UX makes sense?

For the graph index, the underlying NDArray will be unpinned in their deconstructor. So it's not necessary to call `graph.unpin_memory_()` explicitly before training ends. However, I think it's our duty to unpin the pinned `ndata` and `edata` (PyTorch tensors) either at the end of training scripts or in the deconstructor (we don't have one) of `dgl.graph`?

> You could also benchmark it and see if the performance is close to what you have.

No problem.

BarclayII:
@davidmin7 Unfortunately this commit is still not complete as we cannot rely on `__del__` to deconstruct stuff, since `__del__` can be executed in any order (meaning that the C API functions themselves can be deconstructed before the tensors themselves), so there's still possibility that the memory is not properly unregistered.

I think I'll have to do that in C.

davidmin7:
@BarclayII I see, that's an interesting thing to know. Thanks for the notice.

BarclayII:
@davidmin7 Just a quick question though: what will happen if I call `cudaHostRegister` and then forget calling `cudaHostUnregister`?  Would the CUDA runtime unregister them for me when the program terminates?

yaox12:
Epoch times

<table>
<thead>
  <tr>
    <th rowspan="2">mode</th>
    <th colspan="2">1 GPU</th>
    <th colspan="2">2 GPUs</th>
    <th colspan="2">4 GPUs</th>
    <th colspan="2">8 GPUs</th>
  </tr>
  <tr>
    <th>separated</th>
    <th>attached</th>
    <th>separated</th>
    <th>attached</th>
    <th>separated</th>
    <th>attached</th>
    <th>separated</th>
    <th>attached</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>cuda</td>
    <td>8.78</td>
    <td>8.64</td>
    <td>5.51</td>
    <td>4.83</td>
    <td>3.69</td>
    <td>3.13</td>
    <td>2.62</td>
    <td>2.57</td>
  </tr>
  <tr>
    <td>uva</td>
    <td>1.74</td>
    <td>4.13</td>
    <td>1.15</td>
    <td>2.68</td>
    <td>0.71</td>
    <td>1.88</td>
    <td>0.56</td>
    <td>1.61</td>
  </tr>
  <tr>
    <td>allcuda</td>
    <td>1.56</td>
    <td>1.88</td>
    <td>0.83</td>
    <td>0.99</td>
    <td>0.54</td>
    <td>0.67</td>
    <td>0.28</td>
    <td>0.35</td>
  </tr>
</tbody>
</table>

- separated: i.e., `train_sampline_multi_gpu.py`. Features are popped out before sampling and sliced with `load_subtensor()`. Also based on the new dataloader code.
- attached: i.e., `ddp.py`. Features are attached to the graph and fetched.

Seems the new UX suffers a regression in `uva` and `gpu` mode, especially the former. I have done some profiling with the 8x A100s case.

- separated:
image

- attached:
image

Many DtoH Memcpy during dataloading. A more detailed analysis shows they happen in Line 179:
https://github.com/dmlc/dgl/blob/63e7296dcd41cd5ad24eebe270532bb5a81a653a/python/dgl/frame.py#L175-L179
I changed line 175 to 
```python
if F.context(self.storage) != F.context(self.index) and not self.pinned:
```
but didn't help. I'm not familiar with this part so may need your help. @BarclayII

BarclayII:
> Many DtoH Memcpy during dataloading. A more detailed analysis shows they happen in Line 179:

Yeah that's because I accidentally performed the slicing when traversing all the columns.  I think I should have fixed it in the latest commit.

davidmin7:
> @davidmin7 Just a quick question though: what will happen if I call cudaHostRegister and then forget calling cudaHostUnregister? Would the CUDA runtime unregister them for me when the program terminates? If so, then probably I don't need to worry about the whole __del__ shenanigans.

Sorry for the late answer. The host memory will be released on the program termination, but this finding is based on my observation, not based on the exact mechanism that I know about `cudaHostRegister()`. So I know it's a boring answer, but I think we need to make sure `cudaHostUnregister()` is called, because we don't exactly know what that API is wrapping around.

yaox12:
> Would the CUDA runtime unregister them for me when the program terminates?

The answer is yes from an expert in the CUDA team. He reminds us to be careful to two potential risks:
- Say you malloc(), cudaHostRegister, then free() the memory, the memory is still accessible to the device.  Then you call malloc() again, the pointer returned can get reused.
- If host register leak happens in a loop, the pinned sysmem usage can run out of RAM.

yaox12:
Updated benchmark numbers. Supervised GraphSAGE, 3 layers with `fan_out=5,5,5` and `num_hidden=256`, on up to 8x A100 GPU cards.

<table>
<thead>
  <tr>
    <th rowspan="2">mode</th>
    <th colspan="2">1 GPU</th>
    <th colspan="2">2 GPUs</th>
    <th colspan="2">4 GPUs</th>
    <th colspan="2">8 GPUs</th>
  </tr>
  <tr>
    <th>separated</th>
    <th>attached</th>
    <th>separated</th>
    <th>attached</th>
    <th>separated</th>
    <th>attached</th>
    <th>separated</th>
    <th>attached</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>cuda<br>(cpu sampling)</td>
    <td>9.33</td>
    <td>8.21</td>
    <td>5.28</td>
    <td>4.18</td>
    <td>3.08</td>
    <td>2.79</td>
    <td>2.31</td>
    <td>2.05</td>
  </tr>
  <tr>
    <td>uva</td>
    <td>2.84</td>
    <td>2.65</td>
    <td>1.29</td>
    <td>1.31</td>
    <td>0.99</td>
    <td>0.83</td>
    <td>0.53</td>
    <td>0.58</td>
  </tr>
  <tr>
    <td>allcuda</td>
    <td>2.15</td>
    <td>2.48</td>
    <td>1.10</td>
    <td>1.01</td>
    <td>0.73</td>
    <td>0.71</td>
    <td>0.35</td>
    <td>0.35</td>
  </tr>
</tbody>
</table>

Looks pretty good.

BTW, I think we should update current multi-GPU examples with `mp.spawn`. `fork` seems to stop working with this PR.

BarclayII:
> The answer is yes from an expert in the CUDA team

Awwwesome.  Host register leak shouldn't be an issue since we usually only do it once.

> BTW, I think we should update current multi-GPU examples with `mp.spawn`. `fork` seems to stop working with this PR.

The new example already uses `mp.spawn`.

One thing I noticed though is that the OGB baselines uses neighbor size of [15, 10, 5].  So I changed it.  Also tested link prediction example and UVA worked fine.

python/dgl/dataloading/dataloader.py
```diff
@@ -485,23 +487,92 @@ def create_tensorized_dataset(indices, batch_size, drop_last, use_ddp, ddp_seed)
...
+
+        # Check graph and indices device as well as num_workers
+        if use_uva:
+            if self.graph.device.type != 'cpu':
+                raise ValueError('Graph must be on CPU if UVM sampling is enabled.')
+            if num_workers > 0:
+                raise ValueError('num_workers must be 0 if UVM sampling is enabled.')
+
+            self.graph.create_formats_()
+            self.graph.pin_memory_()
```
nv-dlasalle, line 537: How does this work with multi-processing?

python/dgl/dataloading/dataloader.py
```diff
@@ -485,23 +487,92 @@ def create_tensorized_dataset(indices, batch_size, drop_last, use_ddp, ddp_seed)
...
+            indices = recursive_apply(indices, lambda x: x.to(self.device))
+        else:
+            if self.graph.device != indices_device:
```
nv-dlasalle, line 540: Are we still trying to support GraphStorage-like objects? If so, for situations where the graph-structure is a different device than the `Frame`s, what should this value return? The alternate streams and prefetcher arguments are specific to the features, so if I have the graph structure in shared memory on the CPU, and features spread across the GPUs (or vice versa), this logic is problematic.

yaox12:
> The new example already uses `mp.spawn`.

I mean existing examples under like `examples/pytorch/graphsage/`. Should we update them?

examples/pytorch/__temporary__/graphsage/normal.py
```diff
@@ -28,42 +32,63 @@ def forward(self, blocks, x):
...
-        sampler,
-        device='cuda',
-        batch_size=1000,
-        shuffle=True,
-        drop_last=False,
-        pin_memory=True,
-        num_workers=16,
-        persistent_workers=True,
-        use_prefetch_thread=True)       # TBD: could probably remove this argument
+        [15, 10, 5], prefetch_node_feats=['feat'], prefetch_labels=['label'])
```
yaox12, line 77: Should this be `[5, 10, 15]`?

python/dgl/dataloading/dataloader.py
BarclayII: When using multiprocess.spawn PyTorch should already put the tensors in shared memory.  Graphs will be pickled to tensors and go through shared memory using ForkingPickler (I have code in `heterograph_index.py` which does this).

python/dgl/dataloading/dataloader.py
BarclayII: Changes to only perform sanity checks on DGLGraphs.

examples/pytorch/__temporary__/graphsage/normal.py
BarclayII: Looks like the OGB official example samples 15 neighbors on the first layer.

BarclayII:
> > The new example already uses `mp.spawn`.
> 
> I mean existing examples under like `examples/pytorch/graphsage/`. Should we update them?

We should.

examples/pytorch/__temporary__/graphsage/normal.py
yaox12: Yes, but I think DGL samples the neighbors in a reversed order? So the right most number is for the first layer.

src/runtime/ndarray.cc
```diff
@@ -343,6 +347,18 @@ std::shared_ptr<SharedMemory> NDArray::GetSharedMem() const {
   return this->data_->mem;
 }
 
+bool NDArray::IsPinned() const {
```
yaox12, line 350: I'm thinking of removing all the `kDLCPUPinned` device type and using `IsPinned()` instead to avoid many workarounds for the context. I can help you with that by sending a PR to this branch in your repo.

BarclayII: Sure, please do.

yaox12: Submitted the PR here: https://github.com/BarclayII/dgl/pull/2

BarclayII:
Figured out that pinning the features with PyTorch's `pin_memory` (i.e. not using cudaHostRegister) solves the crashing problem.

BarclayII:
OK so the memory leak is fixed but now `normal.py` keeps segfaulting randomly during inference...

python/dgl/heterograph_index.py
```diff
@@ -1365,4 +1366,22 @@ def __setstate__(self, state):
...
+    states = _CAPI_DGLHeteroForkingPickle(graph_index)
+    arrays = [F.from_dgl_nd(arr) for arr in states.arrays]
+    # Similar to what being mentioned in HeteroGraphIndex.__getstate__, we need to save
+    # the tensors as an attribute of the original graph index object.  Otherwise
+    # PyTorch will throw weird errors like bad value(s) in fds_to_keep or unable to
+    # resize file.
+    graph_index._forking_pk_state = (states.meta, arrays)
+    return _forking_rebuild, (graph_index._forking_pk_state,)
+
+ForkingPickler.register(HeteroGraphIndex, _forking_reduce)
```
VoVAllen, line 1385: I thought you used shared memory? If using original pickling method, why bothering register this?

BarclayII:
> Did you see the error before this PR? Or is this PR trying to fix the problem? Do you see it only happens with old PyTorch or it happens also for 1.10?

It happens on 1.10.0+cu102.  And unfortunately it also occurs in the master branch (remove pin_prefetcher=False to get it work on master branch).  This PR wasn't aware of that because in master branch I never wrote any inference code.

BarclayII:
Let's move the discussion of the bug to #3755 .

python/dgl/dataloading/dataloader.py
```diff
@@ -64,7 +64,8 @@ def _next_indices(self):
     def __next__(self):
         batch = self._next_indices()
         if self.mapping_keys is None:
-            return batch
+            # clone() fixes #3755.  Not sure why.  Need to take a look afterwards.
+            return batch.clone()
```
VoVAllen, line 68: Is batch a Torch Tensor created by PyTorch?

BarclayII: It's a view of `self.dataset`, which in turn comes from https://github.com/dmlc/dgl/blob/a05a624d8f4820a8755d75a0e7a81eb04ec19de8/python/dgl/dataloading/dataloader.py#L141-L144.

BarclayII:
The MXNet CI crashes because of this statement:
https://github.com/dmlc/dgl/blob/a05a624d8f4820a8755d75a0e7a81eb04ec19de8/python/dgl/heterograph_index.py#L5

Removing this statement will make the CI pass.  And it only crashes with Python 3.6; Python 3.7+ worked fine.

Since Python 3.6 is EOL last year, should we upgrade our CI's python to 3.7 and release packages from 3.7-3.10 instead?

VoVAllen:
Sure. Also you can even skip all tests for MXNet.

python/dgl/dataloading/dataloader.py
```diff
@@ -485,23 +553,103 @@ def create_tensorized_dataset(indices, batch_size, drop_last, use_ddp, ddp_seed)
...
+                # Create all the formats and pin the features - custom GraphStorages
+                # will need to do that themselves.
+                self.graph.create_formats_()
+                self.graph.pin_memory_()
+                for ntype in self.graph.ntypes:
+                    for key, value in self.graph.nodes[ntype].data.items():
+                        self.graph.nodes[ntype].data[key] = value.pin_memory()
+                for etype in self.graph.canonical_etypes:
+                    for key, value in self.graph.edges[etype].data.items():
+                        self.graph.edges[etype].data[key] = value.pin_memory()
```
yaox12, line 613: If the crash is not due to in-place `pin_memory_`, should we bring it back? `pin_memory()` method in PyTorch will copy tensors and thus duplicate them in multi-GPU training.

python/dgl/dataloading/dataloader.py
```diff
@@ -485,23 +553,103 @@ def create_tensorized_dataset(indices, batch_size, drop_last, use_ddp, ddp_seed)
...
+        self.device = _get_device(device)
+
+        # Sanity check - we only check for DGLGraphs.
+        if isinstance(self.graph, DGLHeteroGraph):
+            # Check graph and indices device as well as num_workers
+            if use_uva:
+                if self.graph.device.type != 'cpu':
+                    raise ValueError('Graph must be on CPU if UVM sampling is enabled.')
+                if num_workers > 0:
+                    raise ValueError('num_workers must be 0 if UVM sampling is enabled.')
```
yaox12, line 602: `UVM` -> `UVA` according to the release note.

python/dgl/dataloading/dataloader.py
```diff
@@ -485,23 +553,103 @@ def create_tensorized_dataset(indices, batch_size, drop_last, use_ddp, ddp_seed)
...
+                for etype in self.graph.canonical_etypes:
+                    for key, value in self.graph.edges[etype].data.items():
+                        self.graph.edges[etype].data[key] = value.pin_memory()
+
+                indices = recursive_apply(indices, lambda x: x.to(self.device))
+            else:
+                if self.graph.device != indices_device:
+                    raise ValueError(
+                        'Expect graph and indices to be on the same device. '
+                        'If you wish to use UVM sampling, please set use_uva=True.')
```
yaox12, line 620: `UVM` -> `UVA` according to the release note.

python/dgl/dataloading/dataloader.py
BarclayII: I figured out that even if I explicitly call `share_memory_` in the main process before `mp.spawn` I still cannot avoid duplicating the features; the memory consumption and `/dev/shm` consumption still linearly scales with the number of GPUs.  How did you write it so that the memory is not duplicated across processes/GPUs?

My PyTorch version is 1.10.0.

BarclayII: Fixed.  For now I noticed that even if the features are not duplicated, the graph itself is not.  I'll leave it for future inspection.  Did you observe the graph itself being duplicated in your own example?  If not, how did you avoid it?

BarclayII: Figured it out that the current "proper" way is to first copy the graph into shared memory, then pass the shared memory *name* to the trainer processes, which will get it with `hetero_from_shared_memory`.

jermainewang (approved):

BarclayII:
There's still one issue left: the graph will be duplicated across (1) DataLoader worker processes when sampling on CPU, (2) trainer processes when under multi-GPU.  I'll merge this commit without the fix since it is already getting very big.

PR #3742, opened 2022-02-16
[NN] Rework RelGraphConv and HGTConv

jermainewang:
## Description

- [x] Add a new `TypedLinear` module which applies linear transformations according to types.
- [x] Rework `RelGraphConv` using the new module. (simplified the code from 203L to 64L)
- [x] Fixed a bunch of bugs in the kernel implementation.
- [x] Add simple CPU gather_mm and segment_mm implementation.
- [x] Updated the RGCN example and change the README.
- [x] Add `HGTConv` using it.

## Checklist
- [x] Code is well-documented


jermainewang:
@zjost It will be great if you could also take a look at the PR. You could focus on the code of the new `RelGraphConv` layer and the newly-added `TypedLinear` layer.

mufeili:
Do we need to perform regression test for this PR?

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+    r"""Linear transformation according to types.
+
+    For each sample of the input batch :math:`x \in X`, apply linear transformation
+    :math:`xW_t`, where :math:`t` is the type of :math:`x`.
+
+    The module supports two regularization methods (basis-decomposition and
+    block-diagonal-decomposition) proposed by "`Modeling Relational Data
+    with Graph Convolutional Networks <https://arxiv.org/abs/1703.06103>`__"
+
+    The basis regularization decomposes :math:`W_t` by:
```
mufeili, line 20: "basis regularization" -> "basis decomposition regularization"

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+
+    .. math::
+
+       W_t^{(l)} = \sum_{b=1}^B a_{tb}^{(l)}V_b^{(l)}
+
+    where :math:`B` is the number of bases, :math:`V_b^{(l)}` are linearly combined
+    with coefficients :math:`a_{tb}^{(l)}`.
+
+    The block-diagonal-decomposition regularization decomposes :math:`W_t` into :math:`B`
+    number of block diagonal matrices. We refer :math:`B` as the number of bases.
```
mufeili, line 30: 1. "number of" is redundant.
2. "block diagonal" -> "block-diagonal"

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+
+    The block regularization decomposes :math:`W_t` by:
```
mufeili, line 32: Either say "block-diagonal-decomposition regularization" or "The regularization"

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+
+    .. math::
+
+       W_t^{(l)} = \oplus_{b=1}^B Q_{tb}^{(l)}
+
+    where :math:`B` is the number of bases, :math:`Q_{tb}^{(l)}` are block
+    bases with shape :math:`R^{(d^{(l+1)}/B)*(d^{l}/B)}`.
```
mufeili, line 39: "*" -> "\times"

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+        Total number of types.
+    regularizer : str, optional
+        Which weight regularizer to use "basis" or "bdd":
+
+         - "basis" is short for basis-decomposition.
+         - "bdd" is short for block-diagonal-decomposition.
+
+        Default applies no regularization.
+    num_bases : int, optional
+        Number of bases. Needed when ``regularizer`` is specified. Default: ``None``.
```
mufeili, line 57: Any recommended value for "basis" and "bdd"?

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+
+    Examples
+    --------
+
+    No regularization.
+    >>> from dgl.nn import TypedLinear
```
mufeili, line 63: Does it render properly with a line of text right above?

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+    >>> x = torch.randn(100, 32)
+    >>> x_type = torch.randint(0, 5, (100,))
+    >>> m = TypedLinear(32, 64, 5)
+    >>> y = m(x, x_type)
+    >>> print(y.shape)
+    torch.Size([100, 64])
+
+    With basis regularization
+    >>> from dgl.nn import TypedLinear
+    >>> import torch
```
mufeili, line 75: Ditto. Also I guess you do not need to import twice.

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+    >>>
+    >>> x = torch.randn(100, 32)
+    >>> x_type = torch.randint(0, 5, (100,))
+    >>> m = TypedLinear(32, 64, 5, regularizer='basis', num_bases=4)
+    >>> y = m(x, x_type)
+    >>> print(y.shape)
+    torch.Size([100, 64])
+    """
+    def __init__(self, in_size, out_size, num_types,
+                 regularizer=None, num_bases=None):
```
mufeili, line 85: Should we also support type-specific bias terms?

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+        if regularizer is None:
+            self.W = nn.Parameter(torch.Tensor(num_types, in_size, out_size))
+        elif regularizer == 'basis':
+            if num_bases is None:
+                raise ValueError('Missing "num_bases" for basis regularization.')
+            self.W = nn.Parameter(torch.Tensor(num_bases, in_size, out_size))
+            self.coeff = nn.Parameter(torch.Tensor(num_types, num_bases))
+            self.num_bases = num_bases
+        elif regularizer == 'bdd':
+            if in_size % num_bases != 0 or out_size % num_bases != 0:
```
mufeili, line 99: Also need to check if `num_bases` is None here.

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+                raise ValueError(
+                    'Input and output sizes must be a multiplier of num_bases.'
```
mufeili, line 101: "a multiplier of" -> "divisible by" as a multiplier can be a float.

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+        else:
+            raise ValueError(
+                f'Supported regularizer options: "basis", "bdd", but got {regularizer}')
+        self.regularizer = regularizer
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        """Reset parameters"""
+        if self.regularizer is None:
+            nn.init.xavier_uniform_(self.W, gain=nn.init.calculate_gain('relu'))
```
mufeili, line 117: This applies to all regularizers.

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+        elif self.regularizer == 'basis':
+            nn.init.xavier_uniform_(self.W, gain=nn.init.calculate_gain('relu'))
+            nn.init.xavier_uniform_(self.coeff, gain=nn.init.calculate_gain('relu'))
+        elif self.regularizer == 'bdd':
+            nn.init.xavier_uniform_(self.W, gain=nn.init.calculate_gain('relu'))
+        else:
+            raise ValueError(
+                f'Supported regularizer options: "basis", "bdd", but got {regularizer}')
```
mufeili, line 125: This is already checked in initialization.

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+        if self.regularizer is None:
+            return self.W
+        elif self.regularizer == 'basis':
+            W = self.W.view(self.num_bases, self.in_size * self.out_size)
+            return (self.coeff @ W).view(self.num_types, self.in_size, self.out_size)
+        elif self.regularizer == 'bdd':
+            return self.W
+        else:
+            raise ValueError(
+                f'Supported regularizer options: "basis", "bdd", but got {regularizer}')
```
mufeili, line 138: Ditto

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def forward(self, x, x_type, sorted_by_type=False):
+        """Forward computation.
+
+        Parameters
+        ----------
+        x : torch.Tensor
+            A 2D input tensor. Shape: (N, D1)
+        x_type : torch.Tensor
+            A 1D integer tensor. Each element is the corresponding type of ``x``. Shape: (N,)
```
mufeili, line 148: "Each element is the corresponding type of ``x``." -> "Type of elements in ``x`` with one-to-one correspondence."

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+            return gather_mm(x, w, idx_b=x_type)
+
+    def __repr__(self):
+        if self.regularizer is None:
+            return (f'TypedLinear(in_size={self.in_size}, out_size={self.out_size}, '
+                    f'num_types={self.num_types})')
+        else:
+            return (f'TypedLinear(in_size={self.in_size}, out_size={self.out_size}, '
+                    f'num_types={self.num_types}, regularizer={self.regularizer}), '
+                    f'num_bases={self.num_bases}')
```
mufeili, line 178: Probably you should move ")" from L177 to L178.

mufeili:
LGTM in general. Have you tried running the RGCN link prediction experiment? Does that yield numbers that look fine?

PS: I have not carefully checked the part related to kernels.

python/dgl/nn/pytorch/linear.py
jermainewang: Probably should in the future. Currently we are lacking the operator to support that.

python/dgl/nn/pytorch/linear.py
jermainewang: I feel this way is easier to read and to be modified in the future (there is no need to share the same variable for different regularization methods).

mufeili (approved):

python/dgl/nn/pytorch/linear.py
```diff
@@ -0,0 +1,178 @@
...
+    >>> import torch
+    >>>
+    >>> x = torch.randn(100, 32)
+    >>> x_type = torch.randint(0, 5, (100,))
+    >>> m = TypedLinear(32, 64, 5, regularizer='basis', num_bases=4)
+    >>> y = m(x, x_type)
+    >>> print(y.shape)
+    torch.Size([100, 64])
+    """
+    def __init__(self, in_size, out_size, num_types,
```
zjost, line 84: Can we make `in_size` optionally a list/dict, which indicates the input sizes for each type enumerated by `num_types`?  I understand that this would then make the regularization methods not work.  Whether or not this makes sense depends on the point of this module.  If it's only to encapsulate this typed linear transform that R-GCN does, then it perhaps doesn't make sense to do as I suggest.  However, if this is R-GCN specific, then it raises the question of why we should separate out this component.

Alternatively, users might want to use this as a general PyTorch convenience module for projecting features of different node types into the same embedding space.  In that case, we do not need to support those regularization methods here, and instead can have those as stand-alone modules that appropriately assume everything is of the same dimensionality.

To summarize, I think it might make more sense to remove these regularization methods from this module and instead make this more general, in which node types with different feature lengths can be handled, and the purpose is to project into a common dimensionality.  Otherwise I don't understand where this module would be used other than in R-GCN, which begs the question as to why it's kept separate.

jermainewang: > Can we make in_size optionally a list/dict,
> Alternatively, users might want to use this as a general PyTorch convenience module for projecting features of different node types into the same embedding space. In that case, we do not need to support those regularization methods here, and instead can have those as stand-alone modules that appropriately assume everything is of the same dimensionality.

We decided to have a separate module for this. See [`HeteroLinear`](https://docs.dgl.ai/en/latest/api/python/nn.pytorch.html#heterolinear). `TypedLinear` and `HeteroLinear` are quite different in many aspects, i.e., `TypedLinear` only allows embeddings of the same length but supports regularization while `HeteroLinear` supports embeddings of different lengths stored in a dict but no weight regularization. Separating the two implementations also makes the code much shorter (saving quite a lot of if-else).

> Otherwise I don't understand where this module would be used other than in R-GCN, which begs the question as to why it's kept separate.

It is also used in `HGTConv`. See the code [here](https://github.com/dmlc/dgl/blob/49563d28d1b1279c9f7f5c2ba3cd9d9f9dbf33eb/python/dgl/nn/pytorch/conv/hgtconv.py).

jermainewang:
That's correct.

PR #3745, opened 2022-02-17
[Performance][GPU] Improving _SegmentMaskColKernel

ayasar70:
## Description

`node_subgraph` function seems to be computationally expensive when sample size is large and GPU execution becomes slower than CPUs. Bottleneck method is _SegmentMaskColKernel. The problem is caused because of this [for loop]( https://github.com/dmlc/dgl/blob/b3d3a2c4b0e04320bd9c0cedeb7ed70813d2d824/src/array/cuda/spmat_op_impl_csr.cu#L428). For each neighbors of a vertex that loop seeks existence of that neighbor in the `cols` array doing a linear search. This approach becomes inefficient when `col_len` is large. Issue can be reproduced using the following code snippet:

`g = dgl.heterograph({('N','E','N'):(torch.arange(E) % (E//20), (torch.arange(E) +1) % (E//20))}, device='cuda')`
`a = torch.arange(int((E//20)*.9), device='cuda')`
`sg = dgl.node_subgraph(g, a)`

After changes, I ran the snippet on a CPU and a GPU for N=10000, 100000, 1000000, and 10000000. I picked E=N*20. Copying the execution times (in seconds): ``GPU'' is the execution time before updates and ``GPU W/ CHANGES''  represents the execution times after the changes on the code submitted to this PR.

| N | CPU	| GPU | GPU W/ CHANGES |
|-----|-----|-----|-----|
|10,000 | 0.01 | 0.02 | 0.002 |
|100,000 |	0.06 | 0.22 | 0.004 |
|1,000,000 | 0.94 | 10.16 | 0.020 |
|10,000,000 | 10.61 | N/A | 0.200 |

## Checklist
- [x] Code is well-documented

## Changes
Change on the kernel: 
- [ ] instead of assigning a thread to rows switching to assigning a thread for each non-zero.
- [ ] instead of doing linear scan, switching to binary search on `cols' array.
- [ ] checking and creating a sorted version of `cols` if required. 
- [ ] adding a couple of binary search methods to `utils.h` and updated a couple of kernels to avoid code repetition.


VoVAllen:
@dgl-bot run g4dn.4xlarge api dmlc/master

src/array/cuda/spmat_op_impl_csr.cu
```diff
@@ -415,25 +405,33 @@ template std::vector<NDArray> CSRGetDataAndIndices<kDLGPU, int64_t>(
...
-          break;
-        }
-      }
+  while (tx < n-1) {
+    bool f = true;
+    if (A[tx] > A[tx+1]) {
+      f = false;
     }
-    count[tx] = cnt;
+    flags[tx] = f;
```
nv-dlasalle, line 434: It should compile to the same, but can we simplify this to `flags[tx] = A[tx] <= A[tx+1]`?

src/array/cuda/utils.h
```diff
@@ -192,6 +192,53 @@ inline DType GetCUDAScalar(
   return result;
 }
 
+/*!
+ * \brief Given a sorted array and a value this function returns the index
+ * of the firs element which compares greater than value.
+ *
+ * @param A: ascending sorted array
+ * @param n: size of the A
+ * @param x: value to search in A
```
nv-dlasalle, line 201: Can we add a `@returns:` field to document the return value?

src/array/cuda/utils.h
```diff
@@ -192,6 +192,53 @@ inline DType GetCUDAScalar(
   return result;
 }
 
+/*!
+ * \brief Given a sorted array and a value this function returns the index
+ * of the firs element which compares greater than value.
```
nv-dlasalle, line 197: `firs` -> `first`

src/array/cuda/utils.h
```diff
@@ -192,6 +192,53 @@ inline DType GetCUDAScalar(
...
+  return l;
+}
+
+/*!
+ * \brief Given a sorted and a value this function returns the index
+ * of the element who is equal to val. If not exist returns n+1
+ *
+ * @param A: ascending sorted array
+ * @param n: size of the A
+ * @param x: value to search in A
```
nv-dlasalle, line 223: Again, lets add the `@return` field (looks like we could just move some of the description to there).

nv-dlasalle: I also think we need to expand the documentation to cover the cases:
`x < A[0]`,
`x > A[0]` and `x < A[n-1]`, but `x` is not in `A`,
A has duplicate values (e.g., do we always return the first index?).

src/array/cuda/spmat_op_impl_csr.cu
```diff
@@ -464,12 +462,44 @@ CSRMatrix CSRSliceMatrix(CSRMatrix csr, runtime::NDArray rows, runtime::NDArray
...
+  int8_t* flags = static_cast<int8_t*>(device->AllocWorkspace(ctx, cols_size-1));
+  int nb = (cols_size + nt - 1) / nt;
+  CUDA_KERNEL_CALL(_CheckSorted,
+      nb, nt, 0, thr_entry->stream,
+      cols.Ptr<IdType>(), cols_size, flags);
+  bool is_sorted = dgl::cuda::AllTrue(flags, cols_size-1, ctx);
+  device->FreeWorkspace(ctx, flags);
+
+  IdType *sorted_cols = cols.Ptr<IdType>();
+  if (is_sorted == false) {
```
nv-dlasalle, line 483: If we are going to have separate code paths for when columns is sorted or not, I think we need add a test for unsorted columns in here https://github.com/dmlc/dgl/blob/master/tests/cpp/test_spmat_csr.cc#L541.

nv-dlasalle (changes requested):

src/array/cuda/spmat_op_impl_csr.cu
ayasar70: Based on our discussions and observing that sorting does not bring more than 10% overhead, I removed that check to simplify the implementation.

src/array/cuda/utils.h
ayasar70: I have updated comments and added return parameter for different cases.

nv-dlasalle (approved):

PR #3769, opened 2022-02-24
[Dataset] add train_idx/val_idx/test_idx as dataset attributes

Rhett-Ying:
## Description

Add for `AsNodePredDataset` only, as `AsLinkPredDataset` has already corresponding attributes.
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/data/adapter.py
+++ b/python/dgl/data/adapter.py
@@ -58,11 +58,11 @@ class AsNodePredDataset(DGLDataset):
     num_classes : int
         Number of classes to predict.
     train_idx : Tensor
-        Indexes for train.
+        An 1-D integer tensor of training node IDs.
     val_idx : Tensor
-        Indexes for validation.
+        An 1-D integer tensor of validation node IDs.
     test_idx : Tensor
-        Indexes for test.
+        An 1-D integer tensor of test node IDs.
 
     Examples
     --------
--- a/python/dgl/data/adapter.py
+++ b/python/dgl/data/adapter.py
@@ -22,11 +22,14 @@ class AsNodePredDataset(DGLDataset):
 
       - Contains only one graph, accessible from ``dataset[0]``.
       - The graph stores:
+
         - Node labels in ``g.ndata['label']``.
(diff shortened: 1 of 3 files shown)


python/dgl/data/adapter.py
```diff
@@ -54,6 +57,12 @@ class AsNodePredDataset(DGLDataset):
     ----------
     num_classes : int
         Number of classes to predict.
+    train_idx : Tensor
+        Indexes for train.
```
jermainewang, line 61:
```suggestion
        An 1-D integer tensor of training node IDs.
```

python/dgl/data/adapter.py
```diff
@@ -54,6 +57,12 @@ class AsNodePredDataset(DGLDataset):
...
+    val_idx : Tensor
+        Indexes for validation.
```
jermainewang, line 63:
```suggestion
        An 1-D integer tensor of validation node IDs.
```

python/dgl/data/adapter.py
```diff
@@ -54,6 +57,12 @@ class AsNodePredDataset(DGLDataset):
...
+    test_idx : Tensor
+        Indexes for test.
```
jermainewang, line 65:
```suggestion
        An 1-D integer tensor of test node IDs.
```

python/dgl/data/adapter.py
```diff
@@ -114,6 +123,8 @@ def process(self):
                 print('Generating train/val/test masks...')
             utils.add_nodepred_split(self, self.split_ratio, self.target_ntype)
```
jermainewang, line 124: Why not putting the logic to the `add_nodepred_split` function?

Rhett-Ying: in order to share among different code path of `split_ratio`. and shared among `process` and `load`. what's more, as this api(`set_split_index`) amis to add attributes into class, I think it's better and clearer to be a method of dataset, instead of a utility function.

jermainewang (approved):

PR #3779, opened 2022-02-28
Fix a bug in sparse optimizer.

zheng-da:
## Description
When a model has sparse embeddings for some node types, it will access the embeddings required by the nodes in a mini-batch. However, it is possible that some mini-batches in a heterogeneous graph don't contain nodes of the type that require the sparse embeddings. In this case, the `grad` is None. We need to handle this case in the sparse optimizer.

## Checklist
- [ ] Code is well-documented


classicsong (approved):
LGTM

PR #3786, opened 2022-02-28
[Bugfix] Fix the bug of RGCN example 

paoxiaode:
## Checklist
- [ ] Code is well-documented

## Changes

This allows graph and train_nid copy to GPU when using GPU sample, the previous code is always CPU sample.


nv-dlasalle:
We would need to add separate arguments to support this (as in some cases we want to store the data on CPU when training on the GPU).

@BarclayII Is it intentional that the new RGCN examples don't support UVA?

PR #3787, opened 2022-02-28
[Doc] Update doc for GPU and UVA sampling

yaox12:
Update doc for GPU and UVA sampling to be compatible with the current interface.


yaox12:
@BarclayII I think this should go into 0.8

BarclayII (approved):

PR #3789, opened 2022-02-28
[Build] Split spmm.cu and sddmm.cu for building on Windows

BarclayII:
VS2019 16.5 apparently has trouble building a large object file :(


yzh119 (approved):

yzh119:
There are some namespace issues.

PR #3796, opened 2022-03-02
[Examples] Add pure gpu example of graphsage

nv-dlasalle:
## Description
Add an example of running graphsage from GPU memory. This was available before new dataloader, so this re-adds an example of that. The majority of this is the `node_classification.py` script, but the modified to store the graph on the GPU, and then several optimizations for inference.

## Checklist
- [x] Code is well-documented


BarclayII (approved):
I'm moving the example to the advanced directory.

PR #3801, opened 2022-03-03
[Bug] Fix ddp dataloader in heterogeneous cases

### LINKED ISSUE #3799: DDP Tensorized Dataset fails when passing a dictionary of indices.
## 🐛 Bug

Specifying a dictionary of indices to load from (mapping node types to tensor arrays) causes DDP datasets to fail

## To Reproduce

1. Create a node dataloader with use_ddp=True and an index set specified via dictionary (mapping node type to tensors). The mapping can be only from a single type to a single tensor array, and this will still fail.
2. Attempt to load data.

## Expected behavior

DDP Dataloading should function as before 0.8, where specifying a dictionary of indices causes no issue.

## Environment

DGL 0.8
Pytorch 
Linux
Source

## Additional context

The functionality to create a dataloader with a dictionary of indices existed prior to 0.8, and worked correctly.

BarclayII:
Fixes #3799 .  This is an embarrassing error.  Will release 0.8.0post1 after this PR.


BarclayII:
@kkranen could you try this PR and see if it fixes your problem?

VoVAllen (approved):

martin503:
DDP doesn't work for me even in version 0.8.1 for heterogenous graphs. Training freezes after first pass on `for _, edge_subgraph, blocks in dataloader:`. Both of my gpus have 100% usage, it looks like there is some deadlock connected with processes.

PR #3803, opened 2022-03-03
[Feature] Compatibility to DLPack 0.6 in tensoradapter

BarclayII:
This is to make tensoradapter compilable against PyTorch 1.11, which upgrades DLPack to 0.6.0 that renamed a bunch of stuff.

### DIFF
--- a/tensoradapter/include/tensoradapter.h
+++ b/tensoradapter/include/tensoradapter.h
@@ -10,32 +10,10 @@
 #ifndef TENSORADAPTER_H_
 #define TENSORADAPTER_H_
 
-#include <dlpack/dlpack.h>
-#include <vector>
-
 #if defined(WIN32) || defined(_WIN32)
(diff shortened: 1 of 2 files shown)


yzh119 (approved):
I noticed that PyTorch has made `dlpack` a part of the package:
https://github.com/pytorch/pytorch/blob/1a917e637c969de7e20c66227e689e007c9b2fc2/aten/src/ATen/dlpack.h

I wonder if it is possible to include dlpack from PyTorch instead of from our 3rd-party libraries (which is usually not synced with PyTorch).

BarclayII:
> I wonder if it is possible to include dlpack from PyTorch instead of from our 3rd-party directory

I think it's not a big deal since the ABI did not change.

TristonC:
@BarclayII When can I expect this PR being merged?

PR #3810, opened 2022-03-05
[BugFix] involve indices_devices for other dataset object.

yinpeiqi:
## Description

DGL 0.8 support custom dataset and we can parse a customize "TensorizedDataset" in the indices field. However here cause an error that indices_device will not be created if a customize dataset is parsed. @BarclayII 

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/dataloading/dataloader.py
+++ b/python/dgl/dataloading/dataloader.py
@@ -697,19 +697,23 @@ class DataLoader(torch.utils.data.DataLoader):
         self.indices = indices      # For PyTorch-Lightning
         num_workers = kwargs.get('num_workers', 0)
 
+        indices_device = None
         try:
             if isinstance(indices, Mapping):
                 indices = {k: (torch.tensor(v) if not torch.is_tensor(v) else v)
                            for k, v in indices.items()}
                 indices_device = next(iter(indices.values())).device
             else:
-                if hasattr(indices, "device"):
-                    indices_device = indices.device
                 indices = torch.tensor(indices) if not torch.is_tensor(indices) else indices
                 indices_device = indices.device
         except:     # pylint: disable=bare-except
             # ignore when it fails to convert to torch Tensors.
             pass
+        if indices_device is None:
+            if not hasattr(indices, 'device'):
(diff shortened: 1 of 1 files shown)


python/dgl/dataloading/dataloader.py
```diff
@@ -703,6 +703,8 @@ def __init__(self, graph, indices, graph_sampler, device='cpu', use_ddp=False,
                            for k, v in indices.items()}
                 indices_device = next(iter(indices.values())).device
             else:
+                if hasattr(indices, "device"):
+                    indices_device = indices.device
```
BarclayII, line 707: Good catch.  But why does this solve the problem where `indices_device` is missing?  If the `indices` object does not have a `device` argument, then the variable `indices_device` will still not appear.  This will still cause problems in sanity checks below.

I guess a better approach will be either (1) setting `indices_device` to None in `except` block, then skip the checks involving `indices_device` if it is None (might be tricky if someone uses custom dataset and found the device not matching the device argument and graph's device), or (2) enforce the indices objects to have a `device` argument in case conversion to a torch tensor fails:

```python
try:
    reraise = False
    if ...:
        ...
    else:
        indices = torch.tensor(indices) if not torch.is_tensor(indices) else indices
        if not hasattr(indices, 'device'):
            reraise = True
            raise AttributeError('Custom indices dataset requires a 'device' attribute indicating where the indices is.')
        indices_device = indices.device
except:
    if reraise:
        raise
```

BarclayII (changes requested):

python/dgl/dataloading/dataloader.py
yinpeiqi: I think in your solution, `torch.tensor(indices)` will failed when indices is a customize dataset, and `indices_device = indices.device` will not be executed. Since we don't know whether indices is a non tensor array-like nids or a customize dataset. Another solution is add a addition try before `indices = torch.tensor(indices)`. What about this solution?

BarclayII (approved):

python/dgl/dataloading/dataloader.py
BarclayII: Looks good.

PR #3826, opened 2022-03-10
[Performance]Fused Edge Softmax GPU Ops

binfoo1993:
## Description
Hi,

I added fused EdgeSoftmax kernel support (forward and backward) in this PR.

The forward kernel and backward kernel both use vertex-centric parallelism.

The task mapping is the same as SpMM kernels: x-axis for feature dimension, y-axis for vertices.

The forward kernel consists of three loops over the edge list for each vertex. The first one finds the max value, the second one calculates the exp sum, the last one calculates the attention of each edge.

The backward kernel consists of two loops over the edge list.

### Cora

| #dim  | #Orginal forward | #New forward |
| ------------- | ------------- | ------- |
| 8  |  0.00051 | 0.000071 |
| 4  | 0.00060 | 0.0000705 |
| 1  | 0.00052 | 0.000082 |

### Pubmed

| #dim  | #Orginal forward | #New forward |
| ------------- | ------------- | ------- |
| 8  |  0.00053 | 0.000099 |
| 4  | 0.00052 |  0.000096 |
| 1  | 0.00053 | 0.000096 |

### Ogbn-arxiv

| #dim  | #Orginal forward | #New forward |
| ------------- | ------------- | ------- |
| 8  |  0.0070 | 0.0125 |
| 4  | 0.0063 |  0.0126 |
| 1  | 0.0040 | 0.0106 |

### Reddit

| #dim  | #Orginal forward | #New forward |
| ------------- | ------------- | ------- |
| 8  |  0.0922 | 0.0979 |
| 4  | 0.0582 |  0.0789 |
| 1  | 0.0586 | 0.1287 |

We can observe significant speedups over the first two small datasets, but slowdowns over larger datasets.

I guess the reason could be the workload imbalance in the large graph and the original implementation uses `sddmm`, which is edge parallelism.

### DIFF
--- a/python/dgl/backend/pytorch/sparse.py
+++ b/python/dgl/backend/pytorch/sparse.py
@@ -473,15 +473,7 @@ class EdgeSoftmax(th.autograd.Function):
             gidx = gidx.edge_subgraph([eids], True).graph
         if norm_by == 'src':
             gidx = gidx.reverse()
-        #Note: Now _edge_softmax_forward op only supports CPU
-        #TODO(Zhejiang): We will support GPU in the future
-        if(score.is_cuda):
-            score_max = _gspmm(gidx, 'copy_rhs', 'max', None, score)[0]
-            score = th.exp(_gsddmm(gidx, 'sub', score, score_max, 'e', 'v'))
-            score_sum = _gspmm(gidx, 'copy_rhs', 'sum', None, score)[0]
-            out = _gsddmm(gidx, 'div', score, score_sum, 'e', 'v')
-        else:
-            out = _edge_softmax_forward(gidx, score, 'copy_rhs')
+        out = _edge_softmax_forward(gidx, score, 'copy_rhs')
         ctx.backward_cache = gidx
         ctx.save_for_backward(out)
         return out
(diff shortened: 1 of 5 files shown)


nv-dlasalle:
For cases when the number of output dimensions is very small as is the case with GAT, we may consider using CUB's DeviceSegmentedReduce https://nvlabs.github.io/cub/structcub_1_1_device_segmented_reduce.html#a366739c1fc3cab723f216110f010c6d0 to perform independent reductions for sum and max.

binfoo1993:
> @binfoo1993 I'm concerned about this parallelization strategy when the output dimension is less than 32, and our accesses become uncoalesced and we start suffering from branch divergence.
> 
> Can you provide details for re-running the above experiments (which scripts were run, what argument if any were provided)?

I agree with you, this implementation will suffer from uncoalesced memory access and branch divergence if feature size is small. But the original implementation uses two SpMMs to calculate  the max and exp_sum, they should have the same problem if the feature size is small. Does the SpMM kernel have optimization for small feature size?

I use the [bench_edgesoftmax.py](https://github.com/dmlc/dgl/blob/master/benchmarks/benchmarks/kernel/bench_edgesoftmax.py) script for the benchmark. 

You can run `asv run -n -e --python=same --verbose --bench kernel.bench_edgesoftmax` in the [benchmark](https://github.com/dmlc/dgl/tree/master/benchmarks) folder.

I will try the reduce function.

Thanks a lot.

binfoo1993:
> As @nv-dlasalle mentioned, it's better considering other parallelization strategies for small feature sizes, you can either:
> 
> 1. Use `cub` as Dominique suggested.
> 2. Check graphiler's implementation: https://github.com/xiezhq-hermann/graphiler/blob/main/src/ops/segment_softmax/segment_softmax_kernel.cu
> 
> I believe transpose the input shape from `(E, feat_size)` to `(feat_size, E)`, and the output shape from `(feat_size, E)` to `(feat_size, E)` should help.

Got it. Will try. Thank you!

TristonC:
@yaox12 Could you help to review this PR draft?

mufeili:
> @yaox12 Could you help to review this PR draft?

@TristonC We did not merge this PR as it did not yield consistent improvement. Is this no longer the case now or is there any other update?

PR #3840, opened 2022-03-14
[Bug Fix] Fix the case when reverse_edge is False for citation graphs

### LINKED ISSUE #3832: CoraGraphDataset `reverse_edge=False` still includes bi-directional edges
## To Reproduce

```
>>> import torch, dgl
>>> g0 = dgl.data.CoraGraphDataset(force_reload=True, reverse_edge=True)[0]
>>> g1 = dgl.data.CoraGraphDataset(force_reload=True, reverse_edge=False)[0]
>>> g0.number_of_edges(), g1.number_of_edges()
(10556, 10556)

```

## Expected behavior
when `reverse_edge=False`, only include half of the edges 

## Environment

 - DGL Version (e.g., 1.0): '0.6.1'
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): `1.9.0`
 - OS (e.g., Linux): Linux
 - Python version:

mufeili:
## Description
Fix #3832 

## Checklist
- [ ] Code is well-documented


python/dgl/data/citation_graph.py
```diff
@@ -118,8 +118,13 @@ def process(self):
 
         if self.reverse_edge:
             graph = nx.DiGraph(nx.from_dict_of_lists(graph))
+            g = from_networkx(graph)
         else:
             graph = nx.Graph(nx.from_dict_of_lists(graph))
+            edges = list(graph.edges())
+            u, v = map(list, zip(*edges))
+            g = dgl_graph((u, v))
+            graph = g.to_networkx()
```
jermainewang, line 127: Even if the fix works, the change is very convoluted. You are creating a networkx graph, converting it to DGLGraph and then converting it back to networkx graph. Why can't we just have networkx graph -> DGLGraph? Please simplify the code.

jermainewang (changes requested):

python/dgl/data/citation_graph.py
mufeili: 1. This class used to have an attribute `._graph` that is a NetworkX graph, do we still want it?
2. The NetworkX graph at L123 is an undirected one and the implementation obtained edges for one direction by taking `graph.edges()`. However, this will not give a directed NetworkX graph with edges for a single direction, hence L127.

mufeili: For 2, the original implementation relies on `NetworkX` to drop half of the edges as the graph loaded by pickle has edges for both directions.

jermainewang: Anyway, the goal is to have simpler implementation. You could start with a networkx graph to perform some pruning or processing and convert to DGLGraph later. Or you could start with DGLGraph directly and use DGL APIs to achieve the same logics. Either way, just do not have networkx->DGLGraph->networkx->DGLGraph multiple times.

mufeili: I've removed the `._graph` attribute as it does not seem to be used anywhere. This avoids reconverting the `DGLGraph` into a NetworkX graph in the end. @jermainewang

python/dgl/data/citation_graph.py
```diff
@@ -122,8 +122,12 @@ def process(self):
 
         if self.reverse_edge:
             graph = nx.DiGraph(nx.from_dict_of_lists(graph))
+            g = from_networkx(graph)
         else:
             graph = nx.Graph(nx.from_dict_of_lists(graph))
+            edges = list(graph.edges())
+            u, v = map(list, zip(*edges))
+            g = dgl_graph((u, v))
```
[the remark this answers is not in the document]
mufeili: You can, but the graph will have edges for both directions, which is not desired.

PR #3841, opened 2022-03-14
[Bugfix] Fix multiple bugs and code refactor

### LINKED ISSUE #3802: I can't run dgl/examples/pytorch/han/train_sampling.py with cuda
## To Reproduce

1. I just run dgl/examples/pytorch/han/train_sampling.py with cuda, it has an error

image

## Environment

 - DGL Version: 0.8.0+cu11.1
 - Backend Library & Version:  PyTorch0.9.1
 - OS: Linux
 - How you installed DGL: pip
 - Python version: 3.7.4
 - CUDA/cuDNN version: cuda11.1
 - GPU models and configuration: A100

BarclayII:
* Removed `SetCOOMatrix` and similar methods since they are not necessary.
* Fix `CopyToSharedMem` assuming that all relation graphs are homogeneous.
* Fix #3802 by maintaining `is_pinned` flag in the GraphIndex rather than checking for the arrays' pinned memory status every time.

### DIFF
--- a/include/dgl/aten/coo.h
+++ b/include/dgl/aten/coo.h
@@ -47,6 +47,8 @@ struct COOMatrix {
   bool row_sorted = false;
   /*! \brief whether the column indices per row are sorted */
   bool col_sorted = false;
+  /*! \brief whether the matrix is in pinned memory */
+  bool is_pinned = false;
   /*! \brief default constructor */
   COOMatrix() = default;
   /*! \brief constructor */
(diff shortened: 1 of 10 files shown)


nv-dlasalle (approved):
Looks good to me.

jermainewang:
Should the flag belong to COOMatrix or CSRMatrix instead?

wsjeon:
Is this PR still in the middle of merging process? I found that this is useful when I use both pytorch_lightning and dgl.

PR #3842, opened 2022-03-14
[NN] Grouped reversible residual connections for GNNs

mufeili:
## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/nn/pytorch/conv/grouprevres.py
+++ b/python/dgl/nn/pytorch/conv/grouprevres.py
@@ -132,8 +132,12 @@ class GroupRevRes(nn.Module):
 
     def _forward(self, x, g, *args):
         xs = torch.chunk(x, self.group, dim=-1)
-        chunked_args = list(map(lambda arg: torch.chunk(arg, self.group, dim=-1), args))
-        args_chunks = list(zip(*chunked_args))
+
+        if len(args) == 0:
+            args_chunks = [()] * self.group
+        else:
+            chunked_args = list(map(lambda arg: torch.chunk(arg, self.group, dim=-1), args))
+            args_chunks = list(zip(*chunked_args))
         y_in = sum(xs[1:])
 
         ys = []
@@ -149,8 +153,12 @@ class GroupRevRes(nn.Module):
 
     def _inverse(self, y, g, *args):
         ys = torch.chunk(y, self.group, dim=-1)
-        chunked_args = list(map(lambda arg: torch.chunk(arg, self.group, dim=-1), args))
-        args_chunks = list(zip(*chunked_args))
+
+        if len(args) == 0:
+            args_chunks = [()] * self.group
+        else:
+            chunked_args = list(map(lambda arg: torch.chunk(arg, self.group, dim=-1), args))
+            args_chunks = list(zip(*chunked_args))
 
         xs = []
         for i in range(self.group-1, -1, -1):
--- a/python/dgl/nn/pytorch/conv/grouprevres.py
+++ b/python/dgl/nn/pytorch/conv/grouprevres.py
@@ -14,7 +14,8 @@ class InvertibleCheckpoint(torch.autograd.Function):
         ctx.weights = inputs_and_weights[num_inputs:]
         inputs = inputs_and_weights[:num_inputs]
 
-        ctx.input_requires_grad = [element.requires_grad for element in inputs]
+        # ctx.input_requires_grad = [element.requires_grad for element in inputs]
+        ctx.input_requires_grad = []
 
         with torch.no_grad():
             # Makes a detached copy which shares the storage
@@ -22,8 +23,10 @@ class InvertibleCheckpoint(torch.autograd.Function):
             for element in inputs:
                 if isinstance(element, torch.Tensor):
                     x.append(element.detach())
+                    ctx.input_requires_grad.append(element.requires_grad)
                 else:
                     x.append(element)
+                    ctx.input_requires_grad.append(None)
             # Detach y in-place (inbetween computations can now be discarded)
             outputs = ctx.fn(*x).detach_()
 
@@ -60,14 +63,13 @@ class InvertibleCheckpoint(torch.autograd.Function):
         # compute gradients
         with torch.set_grad_enabled(True):
             detached_inputs = []
-            for element in inputs:
+            for i, element in enumerate(inputs):
                 if isinstance(element, torch.Tensor):
-                    detached_inputs.append(element.detach())
-                else:
-                    detached_inputs.append(element)
+                    element = element.detach()
+                    element.requires_grad = ctx.input_requires_grad[i]
+                detached_inputs.append(element)
+
             detached_inputs = tuple(detached_inputs)
-            for det_input, requires_grad in zip(detached_inputs, ctx.input_requires_grad):
-                det_input.requires_grad = requires_grad
             temp_output = ctx.fn(*detached_inputs)
 
         filtered_detached_inputs = tuple(filter(lambda x: x.requires_grad, detached_inputs))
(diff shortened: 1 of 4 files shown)


python/dgl/nn/pytorch/conv/grouprevres.py
```diff
@@ -0,0 +1,186 @@
...
+
+            tmp_out = self.gnn_modules[i](g, y_in, *args_chunks[i])
+            x = ys[i] - tmp_out
+            xs.append(x)
+
+        x = torch.cat(xs[::-1], dim=-1)
+
+        return x
+
+    def forward(self, g, x, *args):
```
jermainewang, line 178: Need docstring for forward.

python/dgl/nn/pytorch/conv/grouprevres.py
```diff
@@ -0,0 +1,186 @@
...
+
+        gradients = tuple(input_gradients) + gradients[-len(ctx.weights):]
+
+        return (None, None, None) + gradients
+
+
+class GroupRevRes(nn.Module):
+    r"""Grouped reversible residual connections for GNNs, as introduced in
+    `Training Graph Neural Networks with 1000 Layers <https://arxiv.org/abs/2106.07476>`__
+
```
jermainewang, line 97: Add math description?

jermainewang, line 97: Add math description?

python/dgl/nn/pytorch/conv/grouprevres.py
```diff
@@ -0,0 +1,186 @@
...
+    Parameters
+    ----------
+    gnn_module : nn.Module
+        GNN module for message passing. The input and output node representation size need
+        to be the same.
```
jermainewang, line 102:
```
The GNN module for message passing. `GroupRevRes` will clone the module for `group`
number of times. The module must have exactly two forward arguments:
a `DGLGraph` object and a node feature tensor.
```

I don't quite understand how `*args` works here.

python/dgl/nn/pytorch/conv/grouprevres.py
jermainewang: I will add more description about the general usage of this module. Please refer to torch's doc for group conv2d: https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html
```
Similar to group convolution, `GroupRevRes` applies a group of GNN modules
to the input node features. Given an input value with shape :math:`(V_{src}, D)`, the output
value has shape :math:`(V_{dst}, D)`, where :math:`V_{src}` and :math:`V_{dst}` are the
number of source and destination nodes in the graph, respectively. The feature size
:math:`D` must be divisible by `group`.
```

python/dgl/nn/pytorch/conv/grouprevres.py
```diff
@@ -0,0 +1,186 @@
...
+    group : int, optional
```
jermainewang, line 103: Suggest to use `groups` to comply with PyTorch.

python/dgl/nn/pytorch/conv/grouprevres.py
```diff
@@ -0,0 +1,186 @@
...
+        The input node features will be partitioned into the specified number of groups.
```
jermainewang, line 104:
```suggestion
        The size of GNN module group.
```

python/dgl/nn/pytorch/conv/grouprevres.py
```diff
@@ -0,0 +1,186 @@
...
+        super(GroupRevRes, self).__init__()
+        self.gnn_modules = nn.ModuleList()
+        for i in range(group):
+            if i == 0:
+                self.gnn_modules.append(gnn_module)
+            else:
+                self.gnn_modules.append(deepcopy(gnn_module))
+        self.group = group
+
+    def _forward(self, x, g, *args):
```
jermainewang, line 133: Suggest use `(self, g, x, *args)` instead of `(self, x, g, *args)` for a consistent style. Similar suggestion to `_inverse`

python/dgl/nn/pytorch/conv/grouprevres.py
jermainewang: Also, just curious. Does it work with `DGLBlock`?

python/dgl/nn/pytorch/conv/grouprevres.py
jermainewang: I feel having `*args` here is quite confusing and not obvious why we need it. Consider simplify it away.

python/dgl/nn/pytorch/conv/grouprevres.py
```diff
@@ -0,0 +1,186 @@
...
+            args_chunks = [()] * self.group
+        else:
+            chunked_args = list(map(lambda arg: torch.chunk(arg, self.group, dim=-1), args))
+            args_chunks = list(zip(*chunked_args))
+        y_in = sum(xs[1:])
+
+        ys = []
+        for i in range(self.group):
+            tmp_out = self.gnn_modules[i](g, y_in, *args_chunks[i])
+            y = xs[i] + tmp_out
```
jermainewang, line 146: Consider refactor it to the following to make it close to the math in paper:
```python
ys = [None] * len(xs)
ys[0] = sum(xs[1:])
for i in range(1, self.group):
    ys[i] = self.gnn_modules[i](g, ys[i-1]) + xs[i]
```

python/dgl/nn/pytorch/conv/grouprevres.py
yzh119: Considering `DGLBlock` was used together with Neighbor Sampling where we only consider <=2 layers, and in RevNet we have thousands of layers. There is no need to support `DGLBlock` IMO.

mufeili: Consider the equation X_i' = f_{w_i}(X_{i-1}', A, U) + X_i. 

For a DGLBlock, the shape of f_{w_i}(X_{i-1}', A, U) is (N_{dst}, D), while X_i is (N_{src}, D), yielding a size mismatch issue. Extending it might be possible, but non-trivial.

python/dgl/nn/pytorch/conv/grouprevres.py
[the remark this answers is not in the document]
mufeili: Added.

python/dgl/nn/pytorch/conv/grouprevres.py
mufeili: Updated the description, though different from your proposal.

python/dgl/nn/pytorch/conv/grouprevres.py
mufeili: It allows optionally passing additional arguments like `U` in the original paper and edge features in the original implementations.

python/dgl/nn/pytorch/conv/grouprevres.py
mufeili: Changed.

python/dgl/nn/pytorch/conv/grouprevres.py
mufeili: Updated to "The number of groups." The meaning should be covered in the doc above.

python/dgl/nn/pytorch/conv/grouprevres.py
mufeili: Changed `(self, x, g, *args)` -> `(self, g, x, *args)`.

python/dgl/nn/pytorch/conv/grouprevres.py
mufeili: Strictly speaking, if you really want that, you need to have

```python
ys = [None] * (len(xs) + 1)
ys[0] = sum(xs[1:])
for i in range(self.group):
    ys[i + 1] = self.gnn_modules[i](g, ys[i]) + xs[i]
```

jermainewang:
Question: if the `*args* in the forward function requires autograd, will the module still be reversible?

jermainewang (approved):

Issue #3853, opened 2022-03-17
Make errors for invalid arguments/data assertions instead of exceptions.


nv-dlasalle:
## 🚀 Feature
There are many errors/checks in the python code which cannot be gracefully handled, and essentially require program termination (e.g., invalid device inputs), and updating arguments/parameters. Some of these checks require significant overhead.

If we switch these checks to be `assert`s, then users have to option to disable them and see improved performance (when they know their code behaves correctly) via `python -O`.

An example is in `DGLHeteroGraph.find_edges()`:
```
         if len(eid) > 0:
             min_eid = F.as_scalar(F.min(eid, 0))
             if min_eid < 0:
                 raise DGLError('Invalid edge ID {:d}'.format(min_eid))
             max_eid = F.as_scalar(F.max(eid, 0))
             if max_eid >= self.num_edges(etype):
                 raise DGLError('Invalid edge ID {:d}'.format(max_eid))
```
which requires traversing the edge list twice to find the min_eid and the max_eid, and the user is not intended to catch the `DGLError`. Instead this could be:
```
         assert len(eid) > 0 or F.as_scalar(F.min(eid, 0)) >= 0, 'Invalid edge ID {:d}'.format(F.as_scalar(F.min(eid, 0)))
         assert len(eid) > 0 or F.as_scalar(F.max(eid, 0)) < self.num_edges(etype), 'Invalid edge ID {:d}'.format(F.as_scalar(F.max(eid, 0)))
```

BarclayII:
For the sake of performance, I would rather move those checks into C++ (PyTorch)/Cython (like numpy)/etc.  Those Python checks are very likely duplicating what we already did in C++.

I think the checks were there because our exceptions thrown from C++ looks ugly...

jermainewang:
I wonder is there any recommendation of best practices? I see two things here:

* Try to avoid expensive checks in Python. The one in `DGLGraph.find_edges` needs to scan the given `eid` tensor (may be in GPU) and fetch the result to CPU (cause device synchronization). I think in general it's better to conduct such check **in the kernel level**. For example, in the CPU kernel, when looping over the `eid` tensor, assert each eid is within range; in the GPU kernel, add proper assertion.
* For Python-side checks, we should follow the standard practice. `assert` is fine but the error message doesn't look good sometimes. I saw some projects (e.g., torchvision, numpy) directly raise torch native errors (e.g., TypeError, ValueError, RuntimeError, etc.).

PR #3857, opened 2022-03-18
[Feature] Enable UVA for GPU PinSAGE and RandomWalk

yaox12:
## Description
Since the GPU PinSAGE sampler has been added into DGL, this PR enables UVA for it and adds/unblocks necessary unit tests.

## Checklist
- [x] Code is well-documented

## Changes
- Enable UVA for `dgl.sampling.random_walk()`/`dgl.sampling.PinSAGESampler()`/`dgl.sampling.RandomWalkNeighborSampler`
- Enable GPU test for `test_random_walk()` 
- Enable UVA test for `test_pinsage_sampling()`
- Add a `test_random_walk_uva()`


python/dgl/sampling/randomwalks.py
```diff
@@ -169,8 +169,17 @@ def random_walk(g, nodes, *, metapath=None, length=None, prob=None, restart_prob
         metapath = [g.get_etype_id(etype) for etype in metapath]
 
     gidx = g._graph
-    nodes = F.to_dgl_nd(utils.prepare_tensor(g, nodes, 'nodes'))
-    metapath = F.to_dgl_nd(utils.prepare_tensor(g, metapath, 'metapath'))
+    nodes = utils.prepare_tensor(g, nodes, 'nodes')
+    assert F.max(nodes, dim=0) < g.num_nodes(g.canonical_etypes[metapath[0]][0]), \
+        "Seed node ID exceeds the maximum number of nodes."
```
yaox12, line 174: This fixes that the GPU random walk code doesn't check it, while the CPU version does. I do the check on the python side because when it comes to the c++ code, `seeds` and `metapath` are in the GPU memory and we have to copy a lot of things to complete the check.

yaox12:
BTW, do we have a plan to remove the old random walk code under `src/graph/sampler`, along with `python/dgl/contrib/sampling/randomwalk.py` and corresponding tests?

BarclayII:
@yaox12 Please feel free to remove them.

tests/compute/test_sampling.py
```diff
@@ -96,22 +99,91 @@ def test_random_walk():
         g4, [0, 1, 2, 0, 1, 2], metapath=metapath, return_eids=True)
     check_random_walk(g4, metapath, traces, ntypes, trace_eids=eids)
 
-    metapath = ['follow', 'view', 'viewed-by'] * 2
+    # Non-uniform prob is not supported in GPU.
+    if F._default_context_str != 'gpu':
```
nv-dlasalle, line 103: Could we split this to a separate unit test, such that in the log, we can see that non-uniform prob isn't being run on the GPU?

tests/compute/test_sampling.py
```diff
@@ -159,26 +232,49 @@ def _test_sampler(g, sampler, ntype):
...
+            g = g.to(F.ctx())
     sampler = dgl.sampling.PinSAGESampler(g, 'item', 'user', 4, 0.5, 3, 2)
     _test_sampler(g, sampler, 'item')
     sampler = dgl.sampling.RandomWalkNeighborSampler(g, 4, 0.5, 3, 2, ['bought-by', 'bought'])
     _test_sampler(g, sampler, 'item')
     sampler = dgl.sampling.RandomWalkNeighborSampler(g, 4, 0.5, 3, 2,
         [('item', 'bought-by', 'user'), ('user', 'bought', 'item')])
     _test_sampler(g, sampler, 'item')
+    if g.is_pinned():
+        g.unpin_memory_()
```
nv-dlasalle, line 249: Should we be unpinning the memory in a try/finally block to ensure if a test fails, we still unpin the graph?:
```
try:
  ...
finally:
  if g.is_pinned():
    g.unpin_memory()
```

nv-dlasalle (approved):
Looks good to me. Just minor comments on the unit tests.

tests/compute/test_sampling.py
yaox12: Split the random walk test to `test_non_uniform_random_walk()` (for CPU only) and `test_uniform_random_walk(use_uva)` with parametrized CPU(use_uva=False) and GPU(use_uva=[True, False]).

python/dgl/sampling/randomwalks.py
jermainewang: Can we assert that in the GPU kernel? This is related to the discussion in #3853 .

python/dgl/sampling/randomwalks.py
```diff
@@ -169,8 +169,17 @@ def random_walk(g, nodes, *, metapath=None, length=None, prob=None, restart_prob
...
+    metapath = utils.prepare_tensor(g, metapath, 'metapath')
+    if g.is_pinned():
+        assert prob is None, "Prob must be None when the graph is pinned."
+        assert F.device_type(F.context(nodes)) != 'cpu', \
+            "Seed nodes must be on GPU when the graph is pinned."
+        metapath = F.copy_to(metapath, F.context(nodes))
```
jermainewang, line 180: For L176-L178, can we check them in C++?

python/dgl/sampling/randomwalks.py
yaox12: To complete this check, we have to get the maximum number of the start node. For example, in the CPU version:
```c++
const int64_t begin_ntype = hg->meta_graph()->FindEdge(metapath_data[0]).first;
const int64_t max_nodes = hg->NumVertices(begin_ntype);
```
For GPU sampling, the `metapath_data` is on the GPU and we have to copy it back to CPU to get `max_nodes`.
Here on the Python side, we can do the check before moving `metapath` to GPU to save the copy.

python/dgl/sampling/randomwalks.py
yaox12: Moved `prob` check to C++.

For the device check of `nodes`, I suggest keeping it on the Python side. The random walk kernel dispatching is based on the context of `nodes` now. In theory, passing `nodes` on the CPU to a pinned graph should work as well but may have unexpected behavior, for example, sampling is done by CPU and the results are also on CPU memory, which may be hard to realize for users.

yaox12: Resolved.

jermainewang:
Hi @yaox12 , sorry for the late reply, I reply your comment in the thread. I think it maybe better create an array of number of nodes and pass it to the GPU kernel for boundary check.

yaox12:
> Hi @yaox12 , sorry for the late reply, I reply your comment in the thread. I think it maybe better create an array of number of nodes and pass it to the GPU kernel for boundary check.

Hi @jermainewang, do you mean creating an array of the number of all node types in the graph?

yaox12:
I postpone copying `metapath` array to GPU in CUDA `RandomWalkUniform` and move the sanity check to the CUDA kernel function. Two issues for discussion:

1. The assertion failure from CUDA will break the unit test. It will affect all the following tests.
https://github.com/dmlc/dgl/blob/d4bc6d32eeeb7640bbc726842c2f4fa90341d192/tests/compute/test_sampling.py#L111-L113
2. The error message will be something like

> /workspace/dgl/src/graph/sampling/randomwalks/randomwalk_gpu.cu:58: void dgl::sampling::impl::<unnamed>::_RandomWalkKernel(unsigned long, const IdType *, signed long, const IdType *, unsigned long, const dgl::sampling::impl::<unnamed>::GraphKernelData<IdType> *, const FloatType *, signed long, signed long, IdType *, IdType *) [with IdType = signed long; FloatType = float; int BLOCK_SIZE = 256; int TILE_SIZE = 1024]: block: [0,0,0], thread: [3,0,0] Assertion `curr < max_nodes` failed.

Is it acceptable?

jermainewang (approved):
LGTM

PR #3862, opened 2022-03-18
[Bugfix] Fix UVA sampling with lazy feature slicing

### LINKED ISSUE #3847: GraphSAGE UVA sampling code crashes when not specifying feature prefetching
## 🐛 Bug

See below.

## To Reproduce

1. Delete `prefetch_node_feats` and `prefetch_labels` argument in `examples/pytorch/graphsage/node_classification.py`.
2. Run and it crashes with the following error:
   ```
   dgl._ffi.base.DGLError: Cannot assign node feature "h" on device cpu to a graph on device cuda:0. Call DGLGraph.to() to copy the graph to the same device.
   ```

## Expected behavior

It shouldn't crash

## Environment

 - OS (e.g., Linux):
 - Python version:

## Additional context

The reason is because of incorrect handling of lazy feature slicing when the graph is pinned and the subgraph indices are on GPU:
```
In [1]: import dgl

In [2]: import torch

In [3]: g = dgl.graph(([1,2,3],[2,3,4]))

In [4]: g.create_formats_()

In [5]: g.ndata['x'] = torch.randn(5, 6)

In [6]: g.pin_memory_()
Out[6]:
Graph(num_nodes=5, num_edges=3,
      ndata_schemes={'x': Scheme(shape=(6,), dtype=torch.float32)}
      edata_schemes={})

In [7]: sg = g.sample_neighbors(torch.LongTensor([1]).cuda(), 2)

In [8]: sg.device, sg.ndata['x'].device
Out[8]: (device(type='cuda', index=0), device(type='cpu'))
```

The correct way would involve (1) pinning the node features together with the graph when `pin_memory_()` is called, and (2) when slicing, check if the feature is in pinned memory and call the UVA slicing function if so.

### LINKED ISSUE #3861: Python 3.10 Support
## 🐛 Bug
In Python 3.10, `import dgl` will raise follwing error:

ImportError: cannot import name 'Mapping' from 'collections' (/HOME/miniconda3/envs/deepl/lib/python3.10/collections/__init__.py)

## To Reproduce
`import dgl`

## Expected behavior
`import dgl` works successfully without errors

## Environment

 - DGL Version (e.g., 1.0):
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): PyTorch 1.11
 - OS (e.g., Linux): Ubuntu 18.04
 - Python version: 3.10

BarclayII:
Fixes #3847 .

@jermainewang I might need your opinion on the behavior addition on lazy feature slicing.  Now if a graph is previously in CPU pinned memory and a subgraph (`sample_neighbors`) is taken, the new graph's features will be on GPU.  Previously it was on CPU.  This further complicates the code in `Column.data`.

Need to run a regression test on this.

Also fixes #3861 .

### DIFF
--- a/python/dgl/storages/pytorch_tensor.py
+++ b/python/dgl/storages/pytorch_tensor.py
@@ -5,21 +5,22 @@ from .base import register_storage_wrapper
 from .tensor import BaseTensorStorage
 from ..utils import gather_pinned_tensor_rows
 
-def _fetch_cpu(indices, tensor, feature_shape, device, pin_memory):
+def _fetch_cpu(indices, tensor, feature_shape, device, pin_memory, **kwargs):
     result = torch.empty(
         indices.shape[0], *feature_shape, dtype=tensor.dtype,
         pin_memory=pin_memory)
     torch.index_select(tensor, 0, indices, out=result)
-    result = result.to(device, non_blocking=True)
+    kwargs['non_blocking'] = True
+    result = result.to(device, **kwargs)
     return result
 
-def _fetch_cuda(indices, tensor, device):
-    return torch.index_select(tensor, 0, indices).to(device)
+def _fetch_cuda(indices, tensor, device, **kwargs):
+    return torch.index_select(tensor, 0, indices).to(device, **kwargs)
 
 @register_storage_wrapper(torch.Tensor)
 class PyTorchTensorStorage(BaseTensorStorage):
     """Feature storages for slicing a PyTorch tensor."""
-    def fetch(self, indices, device, pin_memory=False):
+    def fetch(self, indices, device, pin_memory=False, **kwargs):
         device = torch.device(device)
         storage_device_type = self.storage.device.type
         indices_device_type = indices.device.type
@@ -36,7 +37,7 @@ class PyTorchTensorStorage(BaseTensorStorage):
             # CPU to CPU or CUDA - use pin_memory and async transfer if possible
             else:
                 return _fetch_cpu(indices, self.storage, self.storage.shape[1:], device,
-                                  pin_memory)
+                                  pin_memory, **kwargs)
         else:
             # CUDA to CUDA or CPU
-            return _fetch_cuda(indices, self.storage, device)
+            return _fetch_cuda(indices, self.storage, device, **kwargs)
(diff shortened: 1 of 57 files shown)


python/dgl/storages/pytorch_tensor.py
```diff
@@ -5,21 +5,22 @@
...
 from ..utils import gather_pinned_tensor_rows
 
-def _fetch_cpu(indices, tensor, feature_shape, device, pin_memory):
+def _fetch_cpu(indices, tensor, feature_shape, device, pin_memory, **kwargs):
     result = torch.empty(
         indices.shape[0], *feature_shape, dtype=tensor.dtype,
         pin_memory=pin_memory)
     torch.index_select(tensor, 0, indices, out=result)
-    result = result.to(device, non_blocking=True)
+    kwargs['non_blocking'] = True
```
nv-dlasalle, line 13: Should this be `kwargs['non_blocking'] = pin_memory`, since otherwise I believe torch >=1.9 will first copy the tensor to a pinned tensor before beginning the async copy?

nv-dlasalle (approved):

python/dgl/utils/internal.py
```diff
@@ -784,15 +786,17 @@ def extract_node_subframes(graph, nodes, store_ids=True):
     list[Frame]
         Extracted node frames.
     """
-    if nodes is None:
+    if nodes_or_device is None:
         node_frames = [nf.clone() for nf in graph._node_frames]
-    else:
+    elif isinstance(nodes_or_device, Sequence):
```
nv-dlasalle, line 791: We currently represent a tensorflow device as a `str` for which `isinstance('/cuda:0', Sequence)` is `True` (which is why CI is failing). We probably need to have a separate parameter for nodes and a separate parameter for device here:
```
def extract_node_subframes(graph, nodes=None, device=None, store_ids=True):
  assert (nodes is None) ^ (device is None), "Only one of nodes or device must be specified."
  ...
```
And then `subgraph.py` would also need to be adjusted to pass them separately.

BarclayII: Added an `is_listlike` utility function that returns False for strings instead.

PR #3867, opened 2022-03-21
[BugFix] send rpc messages blockingly in case of congestion

Rhett-Ying:
## Description

dist training may hang in DistGraph.__init__() when many trainers are booted. The direct cause seems to be that some rpc messages from clients fail to arrive at servers.

needs more confirmations.
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/src/rpc/tensorpipe/tp_communicator.cc
+++ b/src/rpc/tensorpipe/tp_communicator.cc
@@ -66,13 +66,17 @@ void TPSender::Send(const RPCMessage &msg, int recv_id) {
       LOG(FATAL) << "Cannot send a empty NDArray.";
     }
   }
+  // Let's write blockingly in case of congestion in underlying transports.
+  auto done = std::make_shared<std::promise<void>>();
   pipe->write(tp_msg,
-              [ndarray_holder, recv_id](const tensorpipe::Error &error) {
+              [ndarray_holder, recv_id, done](const tensorpipe::Error &error) {
                 if (error) {
                   LOG(FATAL) << "Failed to send message to " << recv_id
                              << ". Details: " << error.what();
                 }
+                done->set_value();
               });
+  done->get_future().wait();
 }
 
 void TPSender::Finalize() {
--- a/tests/distributed/test_rpc.py
+++ b/tests/distributed/test_rpc.py
@@ -16,7 +16,7 @@ if os.name != 'nt':
 INTEGER = 2
 STR = 'hello world!'
 HELLO_SERVICE_ID = 901231
-TENSOR = F.zeros((10, 10), F.int64, F.cpu())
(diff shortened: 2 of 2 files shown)


VoVAllen (approved):
It's a bit weird. Do you think it's a bug of tensorpipe?

Rhett-Ying:
I don't know. I opened an issue in tensorpipe github to see if this could happen. as write operation is added into a loop and I don't know if there exist some options to control the loop or any underlying components. I may need to dig deeper.

zheng-da (approved):

zheng-da:
can you add some description to describe exactly what the modification does?

src/rpc/tensorpipe/tp_communicator.cc
```diff
@@ -66,13 +66,17 @@ void TPSender::Send(const RPCMessage &msg, int recv_id) {
...
   pipe->write(tp_msg,
-              [ndarray_holder, recv_id](const tensorpipe::Error &error) {
+              [ndarray_holder, recv_id, done](const tensorpipe::Error &error) {
                 if (error) {
                   LOG(FATAL) << "Failed to send message to " << recv_id
                              << ". Details: " << error.what();
                 }
+                done->set_value();
               });
+  done->get_future().wait();
```
classicsong, line 79: So you are making the pipe-write synchronous?

tests/distributed/test_rpc.py
```diff
@@ -188,15 +188,16 @@ def test_multi_client():
     os.environ['DGL_DIST_MODE'] = 'distributed'
     generate_ip_config("rpc_ip_config_mul_client.txt", 1, 1)
     ctx = mp.get_context('spawn')
-    pserver = ctx.Process(target=start_server, args=(10, "rpc_ip_config_mul_client.txt"))
+    num_clients = 20
+    pserver = ctx.Process(target=start_server, args=(num_clients, "rpc_ip_config_mul_client.txt"))
     pclient_list = []
-    for i in range(10):
+    for i in range(num_clients):
```
classicsong, line 194: Will num_client = 20 hang the test if you do not fix the bug?

src/rpc/tensorpipe/tp_communicator.cc
Rhett-Ying: yes.

tests/distributed/test_rpc.py
Rhett-Ying: yes, sometimes hang.

Rhett-Ying: actually, this testcase hangs even `num_clients=10` if `cython-enabled` dgl is installed. In our CI test env, `cython` is always not enabled, so we don't hit this before. I need to figure out a way to enable `cython` in CI, so that we could catch such error. Namely, we'd better test on released DGL.

PR #3871, opened 2022-03-22
[Example] Fix multi-GPU RGCN example

yaox12:
## Description

Previously when we change the multiprocessing start method to `spawn`, we leave out this `mp.Queue()`, which could cause an error.

This also sets CUDA device explicitly in each training process and uses the NCCL backend.

Remove `dgl.multiprocessing` and the recommendation for `fork` in tutorials.

## Checklist

### DIFF
--- a/examples/pytorch/rgcn/entity_sample_multi_gpu.py
+++ b/examples/pytorch/rgcn/entity_sample_multi_gpu.py
@@ -31,12 +31,13 @@ def collect_eval(n_gpus, queue, labels):
 
 def run(proc_id, n_gpus, n_cpus, args, devices, dataset, queue=None):
     dev_id = devices[proc_id]
+    th.cuda.set_device(dev_id)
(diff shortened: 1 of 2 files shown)


TristonC:
@jermainewang @BarclayII Could someone help to review this PR?

examples/pytorch/rgcn/entity_sample_multi_gpu.py
```diff
@@ -101,6 +102,7 @@ def main(args, devices):
     g.create_formats_()
 
     n_gpus = len(devices)
+    mp.set_start_method('spawn')
```
nv-dlasalle, line 105: I think we should add a comment like:
```
# required for mp.Queue() to work with mp.spawn()
```
as otherwise I think it will be easy for folks to assume this line is redundant since `mp.spawn()` uses the `spawn` start method.

nv-dlasalle (approved):

PR #3872, opened 2022-03-22
[Example] layers number fix in GAT example

### LINKED ISSUE #3746: There's some bug on GAT layer example
## 🐛 Bug

On the GAT example of DGL, there are some error on the number of layer. [Link](https://github.com/dmlc/dgl/blob/master/examples/pytorch/gat/gat.py)

For example, if I put 4 layers, it counts 5 layers actually which can cause over-smoothing. 

I think we need to fix it!.

decoherencer:
## Description
Change to GAT example to account for the number of layers in the module as per the input layers argument following the GraphSAGE [example](https://github.com/dmlc/dgl/blob/master/examples/pytorch/graphsage/advanced/model.py)
@VoVAllen 

Fix #3746 

## Checklist
- [ ] Code is well-documented

## Changes
- [x] simple checks and adding layers accordingly.

### DIFF
--- a/examples/pytorch/gat/gat.py
+++ b/examples/pytorch/gat/gat.py
@@ -37,7 +37,7 @@ class GAT(nn.Module):
                 in_dim, num_hidden, heads[0],
                 feat_drop, attn_drop, negative_slope, False, self.activation))
             # hidden layers
-            for l in range(1, num_layers):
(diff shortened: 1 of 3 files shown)


jermainewang (approved):

jermainewang:
Thanks for the fix!

PR #3879, opened 2022-03-23
[Sampler] [Example] SAINTSampler and Simplify GraphSAINT Example

mufeili:
## Checklist
- [x] Code is well-documented


python/dgl/dataloading/graphsaint.py
```diff
@@ -0,0 +1,128 @@
...
+    >>> from dgl.dataloading import SAINTSampler, DataLoader
+    >>> num_iters = 1000
+    >>> sampler = SAINTSampler(mode='node', budget=6000)
+    >>> dataloader = DataLoader(g, torch.arange(num_iters), sampler, num_workers=4)
+    >>> for subg in dataloader:
+    ...     train_on(subg)
+    """
+    def __init__(self, mode, budget, prefetch_ndata=None,
+                 prefetch_edata=None, output_device='cpu'):
+        super().__init__()
```
BarclayII, line 66: Since SAINTSampler only supports PyTorch, it's best to raise an error here if the backend is not PyTorch.

examples/pytorch/graphsaint/modules.py
```diff
@@ -3,16 +3,13 @@
 import torch as th
 import dgl.function as fn
 
-
 class GCNLayer(nn.Module):
     def __init__(self, in_dim, out_dim, order=1, act=None,
-                 dropout=0, batch_norm=False, aggr="concat"):
+                 dropout=0, batch_norm=False):
```
BarclayII, line 8: Do you think it's possible to use our own SAGEConv/GraphConv/etc. with weights?

python/dgl/dataloading/graphsaint.py
mufeili: It seems that all these samplers like `ClusterGCNSampler` and `ShaDowKHopSampler` are supposed to be used with `dgl.dataloading.DataLoader`, which inherits PyTorch's `DataLoader`.

examples/pytorch/graphsaint/modules.py
mufeili: We can replace L48-49 with `GraphConv`, though I'm not sure if it is necessary. This layer also needs to project the input node representations and as a result L25-L36 cannot be fully replaced by `GraphConv`.

BarclayII: But L25-36 can be replaced with nn.Linear, nn.BatchNorm, etc., right?  I think `feat_trans` is just a `nn.Sequential` of `Linear`, `ReLU`, and `BatchNorm`.

Side question: `GCNLayer` has some JKNet-like architecture with multiple hops where every hop's output is concatenated at the end, while `GCNNet` stacks multiple `GCNLayer` again.  This looks quite redundant.  Is this what GraphSAINT does?  Moreover, the default uses `1-1-0` architecture so I guess as an example it doesn't have to be that complicated; just showing people how to use `D_norm` should be fine.

examples/pytorch/graphsaint/train_sampling.py
```diff
@@ -1,128 +1,87 @@
...
-        saint_sampler = SAINTEdgeSampler(args.edge_budget, **kwargs)
-    elif args.sampler == "rw":
-        saint_sampler = SAINTRandomWalkSampler(args.num_roots, args.length, **kwargs)
-    else:
-        raise NotImplementedError
-    loader = DataLoader(saint_sampler, collate_fn=saint_sampler.__collate_fn__, batch_size=1,
-                        shuffle=True, num_workers=args.num_workers, drop_last=False)
+    train_g = dgl.node_subgraph(g, data.train_nid)
+    train_g.ndata['D_norm'] = 1. / train_g.in_degrees().clamp(min=1).unsqueeze(1)
+    g.ndata['D_norm'] = 1. / g.in_degrees().clamp(min=1).unsqueeze(1)
```
BarclayII, line 24: Do we have a function for this?  Can `EdgeWeightNorm` do the job?

examples/pytorch/graphsaint/train_sampling.py
```diff
@@ -1,128 +1,87 @@
...
+            edge_count[sg.edata[dgl.EID]] += 1
+            num_sampled_nodes += sg.num_nodes()
+            pbar.update(sg.num_nodes())
+            num_sgs += 1
+    pbar.close()
+
+    train_g.ndata['n_c'] = node_count.clamp(min=1)
+    train_g.edata['e_c'] = edge_count.clamp(min=1)
+    train_g.apply_edges(fn.v_div_e('n_c', 'e_c', 'w'))
+    train_g.ndata['l_n'] = num_sgs / train_g.ndata['n_c'] / num_train_nodes
```
BarclayII, line 51: I feel all the above should be part of `SAINTSampler`'s preprocessing and should not be exposed to users, like `ClusterGCNSampler`.

python/dgl/dataloading/graphsaint.py
```diff
@@ -0,0 +1,128 @@
...
+        return torch.multinomial(prob, num_samples=self.budget, replacement=True)
+
+    def edge_sampler(self, g):
+        """Node ID sampler for random edge sampler"""
+        src, dst = g.edges()
+        in_deg = g.in_degrees().float().clamp(min=1)
+        out_deg = g.out_degrees().float().clamp(min=1)
+        # We can reduce the sample space by half if graphs are always symmetric.
+        prob = 1. / in_deg[dst.long()] + 1. / out_deg[src.long()]
+        prob /= prob.sum()
```
jermainewang, line 96: Since `SAINTSampler` is likely a quite important sampler, I suggest adding some system efficiency consideration here. How about we adding a caching option (default being `True`) which will cache the probability array the first time. The docstring needs to tell users to turn it off if they want to use the sampler across different graphs.

python/dgl/dataloading/graphsaint.py
```diff
@@ -0,0 +1,128 @@
...
+            The graph to sample from.
+        indices : Tensor
+            Placeholder not used.
+
+        Returns
+        -------
+        DGLGraph
+            The sampled subgraph.
+        """
+        node_ids = self.sampler(g).unique().type(g.idtype)
```
jermainewang, line 124: I will put `unique` and `type` in the `xx_sampler` methods.

jermainewang (changes requested):

examples/pytorch/graphsaint/modules.py
mufeili: @BarclayII Yes the model is used by the original GraphSAINT paper [here](https://github.com/GraphSAINT/GraphSAINT/blob/master/graphsaint/pytorch_version/models.py).

examples/pytorch/graphsaint/README.md
```diff
@@ -29,140 +39,11 @@ All datasets used are provided by Author's [code](https://github.com/GraphSAINT/
...
-python train_sampling.py --task $task $online
-# online sampling: e.g. python train_sampling.py --task ppi_n --online
-# offline sampling: e.g. python train_sampling.py --task flickr_e
+python train_sampling.py --task $task
 ```
 
 - `$task` includes `ppi_n, ppi_e, ppi_rw, flickr_n, flickr_e, flickr_rw, reddit_n, reddit_e, reddit_rw, yelp_n, yelp_e, yelp_rw, amazon_n, amazon_e, amazon_rw`. For example, `ppi_n` represents running experiments on dataset `ppi` with `node sampler`
-- If `$online` is `--online`,  we sample subgraphs on-the-fly in the training phase, while discarding pre-sampled subgraphs. If `$online` is empty, we utilize pre-sampled subgraphs in the training phase.
-
-## Experiments
```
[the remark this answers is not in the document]
mufeili: The original work employs offline sampling while the built-in sampler implemented is fully online. I tried experiments on PPI, which give reasonable performance numbers, but I have not tried experiments on all datasets. If you have concerns for that, I'm fine to remove all changes in the example for this PR.

python/dgl/dataloading/graphsaint.py
mufeili: Done

python/dgl/dataloading/graphsaint.py
mufeili: Done

examples/pytorch/graphsaint/train_sampling.py
mufeili: This is something I'm not sure based on previous discussion with @jermainewang . To speedup the computation of it we might want to use multiple processes, which might make the sampler too complex. Also this depends on whether in most use cases users are only interested in sampling subgraphs without the normalization coefficients.

mufeili:
@BarclayII @jermainewang In general, I'm fine with reverting the changes related to the GraphSAINT example for prioritizing the regular update. I was interested in validating the implementation of the samplers on the original GraphSAINT example and found it too complex.

examples/pytorch/graphsaint/README.md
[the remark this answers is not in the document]
mufeili: The original work considered offline sampling while the built-in sampler is in online mode. I tried experiments on only one dataset, which gave reasonable numbers.

mufeili:
@BarclayII @jermainewang I've reverted the changes in the GraphSAINT example for this PR and only keep the sampler implementation. Regarding the normalization coefficient computation for offline sampling, we can add it in the future if it turns out to be important.

jermainewang (approved):
LGTM

examples/pytorch/graphsaint/train_sampling.py
BarclayII: I was too late.  This could be done by having `SAINTSampler` with two modes, one that computes the norm and one that does not (like you already did).  The norm could be computed in a multiprocessing way with a DataLoader together with `SAINTSampler` not computing the norm.  I think this is how PyG does.

mufeili: Yes, this is what I meant. We will need to construct a `DataLoader` inside `SAINTSampler` to sample a number of subgraphs with a nonzero `num_workers` and use these subgraphs to compute normalization coefficients and optionally cache them. PyG's implementation allows optionally doing that with `num_workers=0`. This was my initial design and Minjie found it too complex with three norm-specific arguments.

PR #3884, opened 2022-03-25
[BUGFIX] Heterograph CopyToSharedMem method bugfix

lixusign:
## Description
when Heterograph copy to shared mem , the size of graph will change  , and later use this shared-graph  in sampling will occur a error as follow and coredump:
屏幕快照 2022-03-25 下午2 46 03
compare the Heterograph graph in shared memory and common use  , the data size exist diff , you will see the next pic 
g1 was common use ,  g2 was g1's copy to share mem, but the size was changed
屏幕快照 2022-03-25 下午2 48 54
the coredump because of  the unitgraph number of types was change. you will see as follow:
image

so fix it


BarclayII:
I should have fixed it in https://github.com/dmlc/dgl/pull/3841.  Could you try the latest nightly build?

lixusign:
@BarclayII  thx  I will close

PR #3885, opened 2022-03-25
[Performance][Optimizer] Enable using UVA and FP16 with SparseAdamOptimizer

nv-dlasalle:
## Description
This PR enables transferring optimizer states via UVA (by default) as well as storing them in FP16 (requires opt-in). While there are many factors, combining both of these optimizations improves performance in `backward` by about 2x and significantly cuts down on memory usage (both from FP16, and from not need allocate buffer tensors to copy to from the GPU).

## Checklist
- [x] Code is well-documented

## Changes
- Adds "Scatter" series of UVA function for scatter data from the GPU into pinned CPU memory.
- Adds parameters `use_uva` and `dtype` to the SparseAdamOptimizer.


src/array/cuda/uvm/array_index_select_uvm.cu
```diff
@@ -75,6 +75,59 @@ template NDArray IndexSelectCPUFromGPU<float, int64_t>(NDArray, IdArray);
...
+
+template void IndexScatterGPUToCPU<int8_t, int32_t>(NDArray, IdArray, NDArray);
+template void IndexScatterGPUToCPU<int8_t, int64_t>(NDArray, IdArray, NDArray);
+template void IndexScatterGPUToCPU<int16_t, int32_t>(NDArray, IdArray, NDArray);
+template void IndexScatterGPUToCPU<int16_t, int64_t>(NDArray, IdArray, NDArray);
+template void IndexScatterGPUToCPU<int32_t, int32_t>(NDArray, IdArray, NDArray);
+template void IndexScatterGPUToCPU<int32_t, int64_t>(NDArray, IdArray, NDArray);
+template void IndexScatterGPUToCPU<int64_t, int32_t>(NDArray, IdArray, NDArray);
+template void IndexScatterGPUToCPU<int64_t, int64_t>(NDArray, IdArray, NDArray);
+template void IndexScatterGPUToCPU<float, int32_t>(NDArray, IdArray, NDArray);
```
classicsong, line 126: are we going to store optimizer status as float16 or bfloat16?

tests/pytorch/test_optim.py
```diff
@@ -45,6 +45,86 @@ def test_sparse_adam():
...
 
+@unittest.skipIf(os.name == 'nt', reason='Do not support windows yet')
+@pytest.mark.parametrize('use_uva', [False, True, None])
+def test_sparse_adam_uva(use_uva):
+    if F.ctx().type == 'cpu' and use_uva == True:
+        # we want to only test values of False and None when not using GPU
+        pytest.skip("UVA cannot be used without GPUs.")
+
+    num_embs = 10
+    emb_dim = 4
```
classicsong, line 56: We also need to test emb_dim = 1

python/dgl/optim/pytorch/sparse_optim.py
```diff
@@ -664,11 +724,14 @@ def update(self, idx, grad, emb):
...
-                # before we can use them
-                update_event.wait()
-            state_mem[state_idx] = update_mem_dst
-            state_power[state_idx] = update_power_dst
+
+            if not use_uva:
+                if state_block:
+                    # wait for our transfers from exec_dev to state_dev to finish
+                    # before we can use them
+                    update_event.wait()
```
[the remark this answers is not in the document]
nv-dlasalle: No, since only the optimizer reads the states, we know we will only read them via another UVA kernel on the same stream.

python/dgl/optim/pytorch/sparse_optim.py
```diff
@@ -519,6 +521,14 @@ class SparseAdam(SparseGradOptimizer):
...
         The term added to the denominator to improve numerical stability
         Default: 1e-8
+    use_uva : bool, Optional
+        Whether to use pinned memory for storing 'mem' and 'power' parameters,
+        when the embedding is stored on the CPU. This will improve training
+        speed, but will require locking a large number of virtual memory pages.
+        For embeddings which are stored in GPU memory, this setting will have
+        no effect.
+        Default: True if the gradients are generated on the GPU, and False
+        if the gradients are on the CPU.
```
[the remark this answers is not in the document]
nv-dlasalle: Added in e421b02

python/dgl/optim/pytorch/sparse_optim.py
```diff
@@ -534,66 +544,84 @@ class SparseAdam(SparseGradOptimizer):
...
+    def __init__(self, params, lr, betas=(0.9, 0.999), eps=1e-08, \
+                 use_uva=None, dtype=th.float32):
         super(SparseAdam, self).__init__(params, lr)
         self._lr = lr
         self._beta1 = betas[0]
         self._beta2 = betas[1]
         self._eps = eps
+        self._use_uva = use_uva
+        self._is_using_uva = {}
+        self._dtype = dtype
```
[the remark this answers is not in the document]
nv-dlasalle: Added an explicit check to make sure its float32 and float16 in e421b02. I'd prefer to wait for a follow up PR to add support for bfloat16 (as it needs more testing).

src/array/cuda/uvm/array_index_select_uvm.cu
nv-dlasalle: Upon closer inspection, the float and double instantiations are unused, as this is invoked through the `ATEN_DTYPE_BITS_ONLY_SWITCH()` macro, which selects int types based on width. The standard ` ATEN_DTYPE_SWITCH()` doesn't support 16-bit, requiring a larger change to use that. I removed the unused templates in cbaeaa02.

tests/pytorch/test_optim.py
nv-dlasalle: I parameterized the emb_dim in a0de13d to [1,4,101,1024].

python/dgl/optim/pytorch/sparse_optim.py
```diff
@@ -613,15 +641,26 @@ def update(self, idx, grad, emb):
             Sparse embedding to update.
         """
         with th.no_grad():
+            state_step, state_mem, state_power = emb.optm_state
+            exec_dtype = grad.dtype
+            exec_dev = grad.device
+            state_dev = state_step.device
+
+            if self._is_using_uva[emb.name] is None \
+                    and state_dev == th.device('cpu') and exec_dev != state_dev:
```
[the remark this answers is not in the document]
nv-dlasalle: I switched it to explicitly check in e421b02, as well as add more thorough comments.

classicsong:
LTGM

classicsong (approved):

python/dgl/optim/pytorch/sparse_optim.py
```diff
@@ -647,12 +707,25 @@ def update(self, idx, grad, emb):
...
+                scatter_pinned_tensor_rows(state_power, \
+                                           grad_indices, \
+                                           update_power.to(dtype=self._dtype))
+            else:
+                update_mem_dst = update_mem.to(state_dev, \
+                                               dtype=self._dtype, \
+                                               non_blocking=True)
+                update_power_dst = update_power.to(state_dev, \
+                                                   dtype=self._dtype, \
+                                                   non_blocking=True)
```
yaox12, line 724: D2H copy with a different dtype is always pageable even with `non_blocking=True`. I filed a bug to PyTorch https://github.com/pytorch/pytorch/issues/79933. A temporary workaround is
```python
update_mem_dst = update_mem.to(dtype=self._dtype).to(state_dev, non_blocking=True)
update_power_dst = update_power.to(dtype=self._dtype).to(state_dev, non_blocking=True)
```

yaox12 (changes requested):

yaox12 (approved):

yaox12:
@nv-dlasalle If you don't have anything to add, I think it's OK to merge this PR.

PR #3890, opened 2022-03-28
[Feature] Added floating-point conversion functions to dgl.transforms.functional

ndickson-nvidia:
## Description
Added to_half(), to_float(), and to_double() functions to dgl.transforms.functional, to shallow copy a DGLHeteroGraph, lazily converting floating-point tensor data to float16, float32, or float64 precision, respectively

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/frame.py
+++ b/python/dgl/frame.py
@@ -287,7 +287,24 @@ class Column(TensorStorage):
         """
         col = self.clone()
         if col.dtype != new_dtype:
-            col.deferred_dtype = new_dtype
+            # If there is already a pending conversion, ensure that the pending
+            # conversion and transfer/sampling are done before this new conversion.
+            if col.deferred_dtype is not None:
+                _ = col.data
+
+            if (col.device is None) and (col.index is None):
+                # Do the conversion immediately if no device transfer or index
+                # sampling is pending.  The assumption is that this is most
+                # likely to be the desired behaviour, such as converting an
+                # entire graph's feature data to float16 (half) before transfer
+                # to device when training, or converting back to float32 (float)
+                # after fetching the data to a device.
+                col.storage = F.astype(col.storage, new_dtype)
+            else:
+                # Defer the conversion if there is a pending transfer or sampling.
+                # This is so that feature data that never gets accessed on the
+                # device never needs to be transferred or sampled or converted.
+                col.deferred_dtype = new_dtype
         return col
 
     def __getitem__(self, rowids):
--- a/python/dgl/frame.py
+++ b/python/dgl/frame.py
@@ -792,3 +792,21 @@ class Frame(MutableMapping):
         if necessary."""
         for column in self._columns.values():
             column.unpin_memory_()
+
+    def __astype_float(self, new_type):
+        assert new_type in [F.float64, F.float32, F.float16], \
+            "'new_type' must be floating-point type: %s" % str(type)
+        for key in self:
+            value = self[key]
+            type = value.dtype
+            if type != new_type and type in [F.float64, F.float32, F.float16]:
+                self[key] = F.astype(value, new_type)
+
+    def half_(self):
+        self.__astype_float(F.float16)
+
+    def float_(self):
+        self.__astype_float(F.float32)
+
+    def double_(self):
+        self.__astype_float(F.float64)
--- a/python/dgl/frame.py
+++ b/python/dgl/frame.py
@@ -845,19 +845,29 @@ class Frame(MutableMapping):
     def __astype_float(self, new_type):
         assert new_type in [F.float64, F.float32, F.float16], \
             "'new_type' must be floating-point type: %s" % str(new_type)
+        newframe = self.clone()
+        new_columns = {}
         for name, column in self._columns.items():
             dtype = column.dtype
             if dtype != new_type and dtype in [F.float64, F.float32, F.float16]:
-                self._columns[name] = column.astype(new_type)
+                new_columns[name] = column.astype(new_type)
+            else:
+                new_columns[name] = column
+        newframe._columns = new_columns
+        return newframe
 
-    def half_(self):
-        """Converts all floating-point columns to half-precision (float16)"""
-        self.__astype_float(F.float16)
+    def half(self):
+        """ Return a new frame with all floating-point columns converted
+        to half-precision (float16) """
+        return self.__astype_float(F.float16)
 
-    def float_(self):
-        """Converts all floating-point columns to single-precision (float32)"""
-        self.__astype_float(F.float32)
+    def float(self):
+        """ Return a new frame with all floating-point columns converted
+        to single-precision (float32) """
+        return self.__astype_float(F.float32)
 
-    def double_(self):
-        """Converts all floating-point columns to double-precision (float64)"""
-        self.__astype_float(F.float64)
+    def double(self):
+        """ Return a new frame with all floating-point columns converted
+        to double-precision (float64) """
+        return self.__astype_float(F.float64)
+
--- a/python/dgl/frame.py
+++ b/python/dgl/frame.py
@@ -870,4 +870,3 @@ class Frame(MutableMapping):
         """ Return a new frame with all floating-point columns converted
         to double-precision (float64) """
         return self.__astype_float(F.float64)
-
--- a/python/dgl/frame.py
+++ b/python/dgl/frame.py
@@ -182,11 +182,12 @@ class Column(TensorStorage):
     index : Tensor
         Index tensor
     """
-    def __init__(self, storage, scheme=None, index=None, device=None):
+    def __init__(self, storage, scheme=None, index=None, device=None, deferred_dtype=None):
         super().__init__(storage)
         self.scheme = scheme if scheme else infer_scheme(storage)
         self.index = index
         self.device = device
+        self.deferred_dtype = deferred_dtype
         self.pinned_by_dgl = False
 
     def __len__(self):
@@ -230,6 +231,11 @@ class Column(TensorStorage):
         if self.device is not None:
             self.storage = F.copy_to(self.storage, self.device[0], **self.device[1])
             self.device = None
+
+        # convert data to the right type
+        if self.deferred_dtype is not None:
+            self.storage = F.astype(self.storage, self.deferred_dtype)
+            self.deferred_dtype = None
         return self.storage
 
     @data.setter
@@ -256,6 +262,30 @@ class Column(TensorStorage):
         """
         col = self.clone()
         col.device = (device, kwargs)
+        return col
+
+    @property
+    def dtype(self):
+        if self.deferred_dtype is not None:
+            return self.deferred_dtype
+        return self.storage.dtype
+
+    def astype(self, new_dtype):
+        """ Return a new column such that when its data is requested, it will be converted to new_dtype.
+
+        Parameters
+        ----------
+        new_dtype : Framework-specific type object
+            The type to convert the data to.
+
+        Returns
+        -------
+        Column
+            A new column
+        """
+        col = self.clone()
+        if col.dtype != new_dtype:
+            col.deferred_dtype = new_dtype
         return col
 
     def __getitem__(self, rowids):
@@ -329,7 +359,7 @@ class Column(TensorStorage):
 
     def clone(self):
         """Return a shallow copy of this column."""
-        return Column(self.storage, self.scheme, self.index, self.device)
+        return Column(self.storage, self.scheme, self.index, self.device, self.deferred_dtype)
 
     def deepclone(self):
         """Return a deepcopy of this column.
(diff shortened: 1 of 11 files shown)


BarclayII:
Could you describe the rationale behind this PR?  Currently we are explicitly requiring users to do such conversions themselves because there might be integer tensors as node/edge data and converting to float tensors is not quite appropriate.

python/dgl/heterograph.py
```diff
@@ -6039,6 +6039,60 @@ def int(self):
         """
         return self.astype(F.int32)
 
+    def half_(self):
```
jermainewang, line 6042: Two suggestions:
* Make them out-placed to comply with PyTorch's convention.
* Make these APIs more distinguished with the APIs for converting index types of graph structure.

jermainewang (changes requested):

ndickson-nvidia:
> Could you describe the rationale behind this PR?
> What's the main motivation for this change?

@BarclayII  @jermainewang  The primary objective is to provide an easy way for a user to choose to convert all floating-point feature data to FP16 before transferring data to a GPU, and then convert back to FP32 for training, (see example usage below).  In many cases with large graphs that don't fully fit in GPU memory, a bottleneck for training is just transferring floating-point feature data, so reducing the size of the feature data by a factor of 2 has a significant performance benefit.  In one simple case using ogbn-products with the GraphSAGE node_classification.py example, with conversion to FP16 before transfer and conversion back to FP32, there was a 1.33x speedup on transfer+sampling, and a 1.25x total training speedup, with no measurable effect on accuracy.  Ideally, conversion back to FP32 wouldn't be necessary, since keeping data in FP16 would also allow for larger training batches and faster computation, where appropriate, but additional changes will likely be required in order to support FP16 throughout the rest of the training pipeline, where appropriate.

> Currently we are explicitly requiring users to do such conversions themselves because there might be integer tensors as node/edge data and converting to float tensors is not quite appropriate.

These functions only convert floating-point tensors to different floating-point types, following PyTorch's convention for `Module`'s and `ModuleList`'s, where `half`, `float`, and `double` only convert floating-point tensors, leaving integer tensors untouched, hence the line: `if dtype != new_type and dtype in [F.float64, F.float32, F.float16]:`

> Make them out-placed to comply with PyTorch's convention.

The current code of this PR follows PyTorch's convention of `half_()`, `float_()`, and `double_()` modifying in-place, whereas `half()`, `float()`, or `double()` would return a modified copy.  Also, on PyTorch's `Module` and `ModuleList` classes, the `half()`, `float()`, and `double()` functions **do** modify the `Module`'s in-place, so PyTorch's convention is inconsistent with itself in that regard.  With the motivating use cases, only in-place modification is needed, and `DGLGraph` is more analogous to `Module` than to `Tensor`, so the recommendation was to only provide the in-place conversion functions, but I'd be fine with providing shallow copy functions, as long as it wouldn't end up writing us into a corner where by design, there would be some cases where we couldn't avoid having both copies in memory at the same time.

Example usage of current PR:
```
dataset = DglNodePropPredDataset('ogbn-products')
graph, labels = dataset[0]
graph.half_()
...
for it, (input_nodes, output_nodes, blocks) in enumerate(train_dataloader):
    for block in blocks:
        block.float_()
    ...
```

Example usage if shallow copy versions were added:
```
dataset = DglNodePropPredDataset('ogbn-products')
graph, labels = dataset[0]
graph = graph.half()
...
for it, (input_nodes, output_nodes, blocks) in enumerate(train_dataloader):
    blocks = [block.float() for block in blocks]
    ...
```

It's not a huge difference in usage, so I don't have a strong opinion one way or the other, apart from ensuring that the caveats above are considered.

> Make these APIs more distinguished with the APIs for converting index types of graph structure.

Could you elaborate?  Do you have any specific recommendation on the naming?  They're currently just named based on the functions of the same names in PyTorch, and people seemed to like the idea of following PyTorch's naming.  It is a bit unfortunate that the functions for converting index types currently have names that seem to conflict with those of functions in PyTorch for converting integer tensors.  I don't have any particularly strong opinions on what names to use, as long as it doesn't end up in a heated argument between others, since I know that naming can be contentious at times.

nv-dlasalle:
To add to what @ndickson-nvidia has said about the motivation, converting each feature per-type is pretty unpleasant for users if there is more than one feature or node/edge type, In other words, users currently have to do something like this:
```
# convert all data before training to save host memory
for f in g.ndata.keys():
  for ntype in g.ndata[f].keys():
    g.ndata[f][ntype] = g.ndata[f][type].half()
for f in g.edata.keys():
  for etype in g.ndata[f].keys():
    g.etype[f][etype] = g.edata[f][etype].half()
...
for input_nodes, seed_nodes, blocks in dataloader:
  # only convert those tensors which are to be used, to avoid slicing and copying Frames/Columns unnecessarily
  blocks[0].srcdata['feat']['merchant'] = blocks[0].srcdata['feat']['merchant'].float()
  blocks[0].srcdata['feat']['card_holder'] = blocks[0].srcdata['feat']['card_holder'].float()
  for block in blocks:
    block.edata['weight']['bought_from'] = block.edata['weight']['bought_from'].float()
  ...
```
Which to my mind requires an "advanced" user. I would really like to see FP16 usage make its way into the examples we don't consider "advanced". The goal of this PR is provide means for regular users of DGL reduce PCIe bandwidth and host memory requirements easily. The above would then become (essentially what @ndickson-nvidia has above):
```
# convert all data before training to save host memory
g = g.half()
...
for input_nodes, seed_nodes, blocks in dataloader:
  # implicitly defer conversion for any columns not prefetched
  blocks = [block.float() for block in blocks]
  ...
```
It also makes it a lot easier to convert a graph's feature data to FP16, and then save it to disk (or load graph saved with fp16). Down the road, I think this is a key feature to make AMP training work well with DGL.

I also think since we allow features to be sliced and transferred via operations on the graph (`g.to()` and `g.X_subgraph()`) it would be consistent to allow converting features via the graph as well in this `g.half()` and `g.float()`.

@jermainewang For you point about distinguishing from `g.int()` and `g.long()`, do you have any suggestions? Something like `g.data_half()` seems kind of clunky to me, and attaching it to `ndata`, `edata`, `srcdata` and `dstdata` make it pretty complicated and verbose for the end user. Since the `torch.nn.module.half()` applies only to "floating point parameters and buffers" (https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.half), this is a nice parallel to that, but I agree it does seem at odds with the current behavior of `g.int()` and `g.long()` which apply exclusively to the graph structure but not integer features (I'm guessing to allow subgraphs to converted to int of graphs with more than 2-billion edges)

nv-dlasalle:
Would an option be to add a parameter to the `g.int()` and `g.long()` functions? Something like this DGLHeteroGraph:
```
def int(self, parts='graph'):
  if parts == 'graph':
    g = self.astype(F.int32)
  elif parts == 'features':
    g = self.clone()
    for i, frame in enumerate(g._edge_frames):
       g._edge_frames[i] = frame.int()
    for i, frame in enumerate(g._node_frames):
       g._node_frames[i] = frame.int()
  elif parts == 'all':
    g = self.astype(F.int32)
    for i, frame in enumerate(g._edge_frames):
       g._edge_frames[i] = frame.int()
    for i, frame in enumerate(g._node_frames):
       g._node_frames[i] = frame.int()
  else:
    raise DGLError('Invalid 'parts' value '{}'.format(parts))
  return g
```
While not a perfect solution, it would preserve the default behavior while making `g.int()` and `g.float()` be a more comparable to each other.

jermainewang:
@ndickson-nvidia @nv-dlasalle Thanks for the detailed explanation!

I like the ideas of adding APIs to ease the job of data type casting. Let me summarize the proposals:

1. [Current PR] Add `float()`, `half()` methods to `DGLGraph` which cast all floating point node or edge features into the corresponding data type. The conversion is done in-place and the modified object is returned. Non-floating-point features remain untouched. The design is motivated by [`torch.nn.Module.half`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.half).
1. Add `data_float()`, `data_half()` or attach them to `ndata`, `edata` etc (e.g., `g.ndata.float()`).
1. Since currently we have `DGLGraph.int` and `DGLGraph.long` for changing the ID type of graph structure, we could extend these APIs to allow changing feature data type. As these APIs return a new copy of the graph, the new APIs (e.g., `DGLGraph.float()`, `DGLGraph.half()`) should follow the design.

I'm voting for the third option for the following reasons. For the first option, I'm a bit hesitated to make analogy between DGLGraph and torch Module. Plus, in-place update is similar to out-place update and assign to the same local variable (i.e., `g = g.float()). For the second option, it is indeed quite complex (e.g., don't know what `g.ndata.float()` will return and how to attach it to the original graph). That leaves the third option. Below is the proposed docstring to describe the new UX:

For `int()` and `long()`:

```python
    def int(self, cast_data=False):
        """Cast the graph to one with idtype int32.

        If the graph already has idtype int32, the function directly returns it. Otherwise,
        it returns a cloned graph of idtype int32 with features copied (shallow copy).

        Parameters
        ------------
        cast_data : bool, optional
            If True, also cast feature data to int32 type.

        Returns
        -------
        DGLGraph
            The graph of idtype int32.
        """
        pass
```

For `float()`, `double` and `half()`:

```python
    def float(self, skip_non_float=True):
        """Return a graph with feature data cast to float32.

        Graph structure remains the same.

        Parameters
        ------------
        skip_non_float: bool, optional
            If True, do not cast non-floating-point feature data.

        Returns
        -------
        DGLGraph
            The graph with feature data in float32.
        """
        pass
```

Let me know your opinions.

BarclayII:
I feel using the name `int` and `long` would make the behavior inconsistent.  `int` and `long` by default converts the *graph structure*, but `float` etc by default converts the *data* instead.  PyTorch users will find this interface counter-intuitive as they will expect `int` and `long` to also convert features instead of graph structure.

If `int` and `long` are not used a lot (need to take a survey on GitHub), I would probably just make `int` and `long` cast features and have other interfaces to cast graph structure (e.g. `as32` and `as64`)

ndickson-nvidia:
Sorry for taking a while to reply to this!  I'll be at the meeting this week to discuss options there.

@BarclayII For that second option, as @jermainewang mentioned, it could be difficult to define the expected behaviour to make it valid to assign to `ndata` and `edata` like `g.ndata = g.ndata.float()` and `g.edata = g.edata.float()`.  Given that `ndata` and `edata` are views of the `_node_frames` and `_edge_frames`, but in feature-major order, rather than the type-major order they're stored in, enabling assigning to it may be difficult to safely expose to users, particularly trying to ensure that multiple node/edge types are still supported correctly.

I also have a small change to the deferring logic in relation to type conversion in here, to ensure that things happen in the commonly-expected order by default, and reduce the risk of running into problems in some possible cases.  @nv-dlasalle and I basically went through all of the orders we could think of people reasonably wanting things to occur in and came up with a fairly simple condition that should handle them, but it should have some more eyes on it.  I'll try to get that in before the meeting, in case people want to discuss that aspect of this PR, too.  Hopefully I can remember enough of the cases we discussed that I can explain it. :)

jermainewang:
Hi @ndickson-nvidia , I think the main contention is because the PR tries to add new APIs to `DGLGraph` which is very central to the entire design of DGL. One idea to unblock this development is to limit the changes to data pipeline. Here is the proposal:
* Add a [data transform module](https://docs.dgl.ai/api/python/transforms.html) `ToHalfPrecision(skip_non_float=True)`. It scans all the ndata and edata, and changes them to half precision if they are of floating point type.
* After getting the node features from dataloader, cast it back to float32 using PyTorch's `float()` API.

Here is the proposed end-to-end UX:

```python
import dgl
import dgl.transforms as T
dataset = dgl.CoraGraphDataset(transforms=T.ToHalfPrecision())
g = dataset[0]  # ndata/edata in the graph is in half precision
dataloader = ...  # create dataloader
for subg in dataloader:
    x = subg.ndata['x'].float()  # explicitly cast it back to float32
    train(subg, x)
```

cc @BarclayII @nv-dlasalle

nv-dlasalle:
That makes sense having the user facing API in `dgl.transforms`. However, two things I think the above proposal doesn't address is when the features and node/edge types are non-trivial. 

1) For example, for RGCN we would need to do something like:
```
for subg in dataloader:
  x = {}
  for ntype in subg.srctypes:
    # this is tricky, as in some cases we want dstdata and/or srcdata.
    x[ntype] = subg.ndata['x'][ntype].float()
    train(subg, x)
```
I think ideally we would want to be able to call the transform on the subgraph/block as well. Something like:
```
import dgl
import dgl.transforms as T
dataset = dgl.CoraGraphDataset(transforms=T.ToHalfPrecision())
g = dataset[0]  # ndata/edata in the graph is in half precision
dataloader = ...  # create dataloader
float_trans = T.ToFullPrecision()
for subg in dataloader:
    subg = float_trans(subg)
    x = subg.ndata['x']
    train(subg, x)
```

2) With this simplified interface, we would still want to allow deferring conversions of columns which have deferred transfers of slices pending, so that users would not need to explicitly specify the features/types they are using (as is currently the case for slicing and transfers). I believe in the current code structure, this would still require a non-user facing API in `Frame` to defer the conversions of Columns.

python/dgl/heterograph.py
nv-dlasalle: This has been removed in favor of `dgl.transforms.to_half()/dgl.transforms.to_float()`

python/dgl/frame.py
```diff
@@ -792,3 +841,32 @@ def unpin_memory_(self):
         if necessary."""
         for column in self._columns.values():
             column.unpin_memory_()
+
+    def __astype_float(self, new_type):
```
BarclayII, line 845: I would prefer having a single underscore.

BarclayII (approved):
I'm good.

python/dgl/frame.py
ndickson-nvidia: I've now changed it to one underscore, so that it's not a fully private function.  Thanks!

PR #3895, opened 2022-03-30
[Performance][GPU] Improving Disjoint Union kernel for Graph Dataloaders

ayasar70:
## Description

This PR aims to optimize DisjointUnionCOO kernel. This kernel is crucial for Graph Dataloaders and called by default during batch operation (batch.py). This kernel, accepts a list of graphs, and concatenates them into a new graph where each graph becomes a disjoint component. Current implementation is architecture agnostic and primarily depends on pytorch's concatenation method. However, this approach causes additional allocations and copies. Therefore, the performance of this operation becomes more and more expensive and becomes the bottleneck when the batch size is increased. On the other hand larger batch sizes are better for forward and backward operations due to better GPU utilization.

This PR tries to overcome these issues by separating CPU and GPU kernels and parallelizing them individually. Preliminary performance changes that this PR achieved are listed in the below table for different batch sizes in terms of processed batches/second using the GIN example. 

- Example: ```examples/pytorch/gin```

- Test: ```python main.py --batch_size <X> --dataset COLLAB --epochs 5 ```

 b/s = ```processed batches / second``` (reported by the script for training). We report the median of 5 epochs.

|Batch Sizes| W/O PR | W/ PR|
|-------|-------|-------|
|64 | 45 b/s | 62 b/s |
|128 | 36 b/s | 50 b/s |
|256 | 26 b/s | 37 b/s|
|512 | 16 b/s | 24 b/s|
## Checklist
- [x] Code is well-documented

## Changes
Change on the kernel: 
- [x] Adding CPU and GPU kernels to avoid additional copies and allocations.
- [x] Better data access for GPU execution.
- [x] Parallelizing CPU kernel.
- [x] Doing necessary header changes.


ayasar70:
@dgl-bot run g4dn.4xlarge api dmlc/master

ayasar70:
@dgl-bot run g4dn.4xlarge api dmlc/master

jermainewang:
Hi @ayasar70, thanks for the contribution! I'd like to understand some high-level contexts. Could you briefly describe what are those extra/unnecessary memory allocations? How did you avoid them in this new implementation?

src/array/cpu/disjoint_union.cc
```diff
@@ -0,0 +1,126 @@
...
+  for (size_t i = 0; i < coos.size(); ++i) {
+    CHECK_SAME_DTYPE(coos[0].row, coos[i].row);
+    CHECK_SAME_CONTEXT(coos[0].row, coos[i].row);
+    has_data |= COOHasData(coos[i]);
+  }
+
+  auto prefixes = _ComputePrefixSums<XPU, IdType>(coos);
+  auto prefix_src = static_cast<IdType*>(std::get<0>(prefixes)->data);
+  auto prefix_dst = static_cast<IdType*>(std::get<1>(prefixes)->data);
+  auto prefix_elm = static_cast<IdType*>(std::get<2>(prefixes)->data);
```
BarclayII, line 70: You can use `std::get<0>(prefixes).Ptr<IdType>()` instead of `static_cast`.  Same for elsewhere.

src/array/cpu/disjoint_union.cc
```diff
@@ -0,0 +1,126 @@
...
+
+  auto result_src = NewIdArray(
+    prefix_elm[coos.size()], coos[0].row->ctx, coos[0].row->dtype.bits);
+  auto result_dst = NewIdArray(
+    prefix_elm[coos.size()], coos[0].col->ctx, coos[0].col->dtype.bits);
+  auto result_dat = NullArray();
+  if (has_data) {
+    result_dat =  NewIdArray(
```
BarclayII, line 78: extra space

src/array/cpu/disjoint_union.cc
```diff
@@ -0,0 +1,126 @@
...
+        res_src_data[prefix_elm[i] + j] = edges_src[j] + prefix_src[i];
+      }
+
+      for (IdType j = 0; j < coo.row->shape[0]; j++) {
+        res_dst_data[prefix_elm[i] + j] = edges_dst[j] + prefix_dst[i];
+      }
+
+      if (has_data) {
+        for (IdType j = 0; j < coo.row->shape[0]; j++) {
+          const auto d = (!COOHasData(coo)) ? IdType(j) : edges_dat[j];
```
BarclayII, line 106: Could just write
```C++
const IdType d = (!COOHasData(coo)) ? j : edges_dat[j];
```

BarclayII (approved):
LGTM with minor comments.

Will let either @nv-dlasalle or @yaox12 to check whether the GPU code has any performance issues.

BarclayII:
@jermainewang I think the point is that the original implementation uses `Range`, `Add` and `Concat`, which essentially creates `N` arrays if batching `N` graphs.  This PR works with a single array so there should be less allocation.

src/array/cpu/disjoint_union.cc
ayasar70: I have replaced this part accordingly.

src/array/cpu/disjoint_union.cc
ayasar70: Probably I am missing; where do I allocate extra space?

src/array/cpu/disjoint_union.cc
ayasar70: Updated accordingly.

ayasar70:
> LGTM with minor comments.
> 
> Will let either @nv-dlasalle or @yaox12 to check whether the GPU code has any performance issues.

Sorry, I have just seen your comments. Thanks for the revision. I tried to address them accordingly beside of the one that I  didn't figure out. I have recently pushed changes.

ayasar70:
> Hi @ayasar70, thanks for the contribution! I'd like to understand some high-level contexts. Could you briefly describe what are those extra/unnecessary memory allocations? How did you avoid them in this new implementation?

Hi @jermainewang, thanks for the review. Sorry, I have just seen all those comments. As @BarclayII said range, add and concat operations are causing this. Best of my knowledge element-wise addition on views (https://github.com/dmlc/dgl/blob/43b9d40e5e2da8b129b1191cf0fd98c0fcb4a796/src/array/union_partition.cc#L36) causes a copy before addition. Then running concatenation (https://github.com/dmlc/dgl/blob/43b9d40e5e2da8b129b1191cf0fd98c0fcb4a796/src/array/union_partition.cc#L59) does the required allocation for the larger tensor first and moves the data allocated before.

src/array/cpu/disjoint_union.cc
yaox12: Maybe he means there is an extra space between "=" and "New" in line 78, a linting problem.

ayasar70: I see :). Fixed.

yaox12:
LGTM. Do we have regression tests to cover this?

BarclayII:
@dgl-bot run g4dn.4xlarge api.bench_batch dmlc/master

Issue #3909, opened 2022-04-05
Pickling/deepcopy does not preserve pinned memory status


BarclayII:
## 🐛 Bug

See below.

## To Reproduce

```python
g = dgl.graph(([0,1,2],[1,2,3]))
g.pin_memory_()
with open('g.pkl', 'wb') as f:
    pickle.dump(g, f)
with open('g.pkl', 'rb') as f:
    h = pickle.load(f)
print(h.is_pinned())   # False

from copy import deepcopy
h = deepcopy(g)
print(h.is_pinned())   # False
```

## Expected behavior

Should be True.

## Environment

 - DGL Version (e.g., 1.0): master
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):
 - OS (e.g., Linux):
 - How you installed DGL (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version (if applicable):
 - GPU models and configuration (e.g. V100):
 - Any other relevant information:

## Additional context

This would impact PyTorch Lightning compatibility when PL modifies the DataLoader via deep copying (e.g. when specifying `overfit_batches` option), as reported in #3863.  From what I understood Python's `deepcopy` relies on `__getstate__` and `__setstate__`, meaning that we probably need to make another version of `HeteroPickleStates` where `is_pinned` status is also included.

PR #3914, opened 2022-04-07
[BugFix] record/restore pin status when pickle/unpickle

Rhett-Ying:
## Description

fix for https://github.com/dmlc/dgl/issues/3909
## Checklist
- [ ] Code is well-documented


VoVAllen:
If a pinned graph pickled to file and unpickled back, the flag is still pinned, but is it really on the pin memory?

Rhett-Ying:
NO, that's why I don't recover the flag in COO/CSR. I call `PinMemory_()` instead.

src/graph/pickle.cc
```diff
@@ -97,6 +99,8 @@ HeteroGraphPtr HeteroUnpickle(const HeteroPickleStates& states) {
   std::vector<HeteroGraphPtr> relgraphs(metagraph->NumEdges());
   std::vector<int64_t> num_nodes_per_type;
   CHECK(strm->Read(&num_nodes_per_type)) << "Invalid num_nodes_per_type";
+  bool is_pinned = false;
+  CHECK(strm->Read(&is_pinned)) << "Invalid flag 'is_pinned'";
```
BarclayII, line 103: I think directly modifying and `HeteroUnpickle` will break unpickling from existing binaries.  You will need to version it by
* Skip reading the `is_pinned` flag (and set it to False) when pickle state version is 1.
* Set the pickle state version to 2 in `_CAPI_DGLCreateHeteroPickleStates`.
You could test compatibility by creating a pickle file in 0.8.0post2, and try unpickling it with your new implementation.

tests/compute/test_pickle.py
```diff
@@ -163,6 +163,28 @@ def test_pickling_subgraph():
     f1.close()
     f2.close()
 
+"""
+@unittest.skipIf(F._default_context_str != 'gpu', reason="Need GPU for pin")
+@unittest.skipIf(dgl.backend.backend_name == "tensorflow", reason="TensorFlow create graph on gpu when unpickle")
+@parametrize_dtype
```
BarclayII, line 169: Why is it disabled?

BarclayII (changes requested):

tests/compute/test_pickle.py
Rhett-Ying: It's weird. when I run `DGLBACKEND=tensorflow DGLTESTDEV=gpu python3 -m pytest compute/test_pickle.py`, `row->ctx.device_type` is GPU within `HeteroUnpickle(const HeteroPickleStates& states)`. 
Is it possible?

BarclayII: Yeah I remember Tensorflow has some problems regarding to pinning (I noticed that in UVA sampling as well).  Disabling Tensorflow's unit test is fine.  I was just asking why you quoted the entire unit test code.

Rhett-Ying: The reason I quoted the whole test is I found many tests(not the one I added) failed with my change, so I just debug to figure out which parts to be blamed.

src/graph/pickle.cc
Rhett-Ying: I tried to set state version in `_CAPI_DGLCreateHeteroPickleStates`, but seems `_CAPI_DGLCreateHeteroPickleStates` is called in `pickle.load()`. Is this expected?

BarclayII: Oh you are right.  `pickle.load()` will first create a `HeteroPickleStates` and then populate it via `__setstate__`.  `version` should be a parameter that is loaded from `HeteroPickleStates.__setstate__` method where the `state` argument is unpacked (the first element of `state` is version number).

I also noticed that the version number is never actually stored correctly in the pickle file since `HeteroPickle` C function never set the number.  Could you check if it's indeed the case?  If so then we will also need to fix that.

Rhett-Ying: Confirmed. `HeteroPickleStates.version` is `0`, not the expected `1`. One question: if `g.pkl` whose version is `1` is loaded in new DGL which has version `2`, when dump with new DGL, the dumped version is expected to be `2` or `1`(the old one).

BarclayII: The output should always be the new one (2).

BarclayII (approved):
I'm good with this PR.  Could you re-enable the `pickling_is_pinned` test and see if it passes?

Rhett-Ying:
I find `g.unpin_memory_()` has to be called before a pinned `g` is destructed, or some other tests fail due to weird error. is this expected? Does it make sense?

PR #3927, opened 2022-04-14
[DGL-Go] Graph Property Prediction Pipeline

mufeili:
## Checklist
- [ ] Code is well-documented


dglgo/dglgo/model/graph_model/gin_ogbg.py
```diff
@@ -0,0 +1,106 @@
+import dgl
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from dgl.nn import GINEConv, AvgPooling, SumPooling
```
jermainewang, line 5: `GINConv` or `GINEConv`?

dglgo/dglgo/model/graph_model/gin_ogbg.py
```diff
@@ -0,0 +1,106 @@
...
+            self.virtual_emb = nn.Embedding(1, embed_size)
+            nn.init.constant_(self.virtual_emb.weight.data, 0)
+            self.mlp_virtual = nn.ModuleList()
+            for _ in range(num_layers - 1):
+                self.mlp_virtual.append(MLP(embed_size))
+            self.virtual_pool = SumPooling()
+
+    def forward(self, graph, node_feat, edge_feat):
+        if self.virtual_node:
+            virtual_emb = self.virtual_emb(torch.zeros(graph.batch_size).long().to(graph.device))
```
jermainewang, line 81: `virtual_emb = self.virtual_emb.expand(graph.batch_size, -1)` should do the job

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
```diff
@@ -0,0 +1,117 @@
...
+from tqdm import tqdm
+{{ data_import_code }}
+
+{{ model_code }}
+
+def train(device, loader, model, criterion, optimizer):
+    model.train()
+
+    for _, (g, labels) in enumerate(tqdm(loader, desc="Iteration")):
+        if g.batch_size == 1:
```
jermainewang, line 16: Why?

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
```diff
@@ -0,0 +1,117 @@
...
+
+        g = g.to(device)
+        labels = labels.to(device)
+        node_feat = g.ndata['feat']
+        edge_feat = g.edata['feat']
+
+        pred = model(g, node_feat, edge_feat)
+        optimizer.zero_grad()
+        # ignore nan targets (unlabeled) when computing training loss
+        is_labeled = labels == labels
```
jermainewang, line 27: In what scenario will there be nan labels?

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
mufeili: It's necessary if the final prediction MLP consists of a batch normalization layer, which requires the batch size to be greater than 1 during training. This is not needed for this model, but might be needed for other models.

dglgo/dglgo/model/graph_model/gin_ogbg.py
mufeili: Replaced with `virtual_emb = self.virtual_emb.weight.expand(graph.batch_size, -1)`

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
mufeili: Removed for now, might add it back if necessary

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
mufeili: Copied from the author's code, which seems not to be necessary for ogbg-molhiv. Removed it for now. This can be common if the dataset is for multi-label property prediction.

mufeili:
@jermainewang Another issue is that the `PNAConv` used in the OGB submission is also different from the original `PNAConv` in several places, e.g., it does not consider edge features. I'm not sure if we should put that as another built-in module or merge it with the existing `PNAConv`.

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
mufeili: It turns out that there are many missing labels in ogbg-molpcba, which is for 128 tasks. I've added it back.

mufeili:
Additional note:
- The PNA submission to the ogbg-molpcba leaderboard employs Laplacian positional encodings and it cannot be reproduced without adding transform functionality to DGL-Go.
- The submission mentioned above employs another PNA variant. As a result, there are three PNA variants in total, one for the original paper, one for ogbg-molhiv and one for ogbg-molpcba.
- In light of the two issues above, I will prefer only including three recipes for this PR, i.e., GIN on molhiv, GIN on molpcba, and PNA (simple variant) on molhiv, leaving PNA (complex variant) on molpcba for future work.

mufeili:
The development of this PR motivates #3949

mufeili:
Reference performance numbers for the three recipes over 10 runs.

| Dataset | Model | leaderboard | Go |
| -------- | ------ | ------------ | --- |
| ogbg-molhiv | GIN (virtual node) | 0.7707 +- 0.0149 | 0.7677 +- 0.0156 |
| ogbg-molhiv | PNA | 0.7905 +- 0.0132 | 0.7880 +- 0.0124 |
| ogbg-molpcba | GIN (virtual node) | 0.2703 +- 0.0023 | 0.2673 +- 0.0021 |

mufeili:
> It seems that there is a ~1% gap between the numbers on leaderboard and DGL-Go. Any clue?

This is almost always the case if we implement an experiment with a different framework using the original hyperparameters. Performing hyperparameter tuning might help, but can be very time consuming.

mufeili:
@jermainewang I find it easier to implement a separate built-in module, say `SimplePNAConv`, for the PNA model used here as it differs from the original `PNAConv` in many different places:
- The original `PNAConv` concatenates the source and destination node features in message passing when edge features are not provided while the simplified version here only uses the source node features in message passing.
- The use of linear layers, activation function, dropout, and normalization layers are different in terms of type and order.

jermainewang:
> @jermainewang I find it easier to implement a separate built-in module, say `SimplePNAConv`, for the PNA model used here as it differs from the original `PNAConv` in many different places:
> 
> * The original `PNAConv` concatenates the source and destination node features in message passing when edge features are not provided while the simplified version here only uses the source node features in message passing.
> * The use of linear layers, activation function, dropout, residual connections and normalization layers are different in terms of type and order.

I feel if that's the case, we need to compare the model performance of the original PNA and the simplified one. Probably after this PR.

mufeili:
> Probably wait for #3959 and add test for the newly added models and pipeline.
> 
> Not relevant to this PR, but I'd like to make some suggestions to the code structure. I think it's time to think of a possible Python interface and use that to drive a cleaner project structure. I think we can start by making some of the internal options more accessible. For example,
> 
> ```python
> >>> import dglgo
> >>> pipeline = dglgo.pipelines['graphpred']  # Get a pipeline object
> >>> pipeline.name
> "graphpred"
> >>> pipeline.description
> "Graph property prediction pipeline"
> >>> opts = pipeline.options    # configuration options
> >>> opts['num_runs']
> ConfigureOption(default_val=1, type=int, comment="Number of experiments to run")
> ```
> 
> cc @dddg617

Sure.

mufeili:
@Rhett-Ying I've slightly modified `csv_dataset.py` and its associated unit tests. Do you have any concerns?

mufeili:
@jermainewang I've added CI for the graph property prediction pipeline.

dglgo/recipes/graphpred_hiv_pna.yaml
```diff
@@ -0,0 +1,36 @@
+version: 0.0.1
+pipeline_name: graphpred
+device: cuda:0                # Torch device name, e.q. cpu or cuda or cuda:0
+data:
+  name: ogbg-molhiv
+  split_ratio:                # Ratio to generate data split, for example set to [0.8, 0.1, 0.1] for 80% train/10% val/10% test. Leave blank to use builtin split in original dataset
+model:
+  name: pna
+  embed_size: 80              # Embedding size.
+  aggregators: mean max min std # Aggregation function names separated by comma, can include mean, max, min, std, sum
```
jermainewang, line 10: This is not comma separated

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
```diff
@@ -0,0 +1,133 @@
...
+{{ data_import_code }}
+
+{{ model_code }}
+
+def train(device, loader, model, criterion, optimizer):
+    model.train()
+
+    for _, (g, labels) in enumerate(tqdm(loader, desc="Iteration")):
+        g = g.to(device)
+        labels = labels.reshape(g.batch_size, -1).to(device)
```
jermainewang, line 18: Why need a `reshape` here?

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
```diff
@@ -0,0 +1,133 @@
...
+                                 shuffle=False, num_workers=pipeline_cfg['num_workers'])
+    test_loader = GraphDataLoader(data[data.test_idx], batch_size=pipeline_cfg['eval_batch_size'],
+                                  shuffle=False, num_workers=pipeline_cfg['num_workers'])
+
+    # create model
+    model_cfg = cfg["model"]
+    # data[0] is a tuple (g, label)
+    cfg["model"]["data_info"] = {
+        "name": cfg["data_name"],
+        "node_feat_size": data[0][0].ndata['feat'].shape[-1],
```
jermainewang, line 86: Consider add `node_feat_size` and `edge_feat_size` as class attributes of `AsGraphPredDataset`.

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
```diff
@@ -0,0 +1,133 @@
...
+        "edge_feat_size": data[0][0].edata['feat'].shape[-1],
+        "out_size": data.num_tasks
```
jermainewang, line 88: What if it is a multi-class classification task?

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
```diff
@@ -0,0 +1,133 @@
...
+    }
+    if cfg["model_name"] == 'pna':
+        in_deg = torch.cat([g.in_degrees() for (g, _) in data[data.train_idx]])
+        cfg["model"]["data_info"]["delta"] = torch.mean(torch.log(in_deg + 1))
```
jermainewang, line 92: L90-L92 seem to be a new requirement. Could you take a note and let us discuss how to support it later?

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
```diff
@@ -0,0 +1,133 @@
...
+        else:
+            lr_scheduler.step()
+
+    model.load_state_dict(torch.load(save_path))
+    test_metric = evaluate(device, test_loader, model)
+    print('Test Metric: {:.4f}'.format(test_metric))
+
+    return test_metric
+
+if __name__ == '__main__':
```
jermainewang, line 123: Do we support early stop?

dglgo/dglgo/model/graph_model/pna.py
```diff
@@ -0,0 +1,230 @@
...
+        else:
+            scalers = [scl.strip() for scl in scalers.split(' ')]
+            assert set(scalers).issubset({'identity', 'amplification', 'attenuation'}), \
+                "Expect scalers to be a subset of ['identity', 'amplification', 'attenuation'], \
+                    got {}".format(scalers)
+        self.aggregators = aggregators
+        self.scalers = scalers
+
+        if data_info['name'] in ['ogbg-molhiv', 'ogbg-molpcba']:
+            self.node_encoder = AtomEncoder(embed_size)
```
jermainewang, line 202: Expose the choice of `node_encoder` and `edge_encoder` as model hyperparameter?

dglgo/setup.py
```diff
@@ -16,7 +16,10 @@
           'numpydoc>=1.1.0',
           "pydantic>=1.9.0",
           "ruamel.yaml>=0.17.20",
-          "PyYAML>=5.1"
+          "PyYAML>=5.1",
+          "ogb>=1.3.3",
+          "rdkit-pypi",
```
jermainewang, line 21: Is it very difficult to install `rdkit`? If so, suggest to remove it and make related import optional.

jermainewang (changes requested):

dglgo/dglgo/model/__init__.py
```diff
@@ -1,2 +1,3 @@
 from .node_encoder import *
-from .edge_encoder import *
\ No newline at end of file
+from .edge_encoder import *
+from .graph_model import *
```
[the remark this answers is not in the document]
mufeili: Done

dddg617:
> > Probably wait for #3959 and add test for the newly added models and pipeline.
> > Not relevant to this PR, but I'd like to make some suggestions to the code structure. I think it's time to think of a possible Python interface and use that to drive a cleaner project structure. I think we can start by making some of the internal options more accessible. For example,
> > ```python
> > >>> import dglgo
> > >>> pipeline = dglgo.pipelines['graphpred']  # Get a pipeline object
> > >>> pipeline.name
> > "graphpred"
> > >>> pipeline.description
> > "Graph property prediction pipeline"
> > >>> opts = pipeline.options    # configuration options
> > >>> opts['num_runs']
> > ConfigureOption(default_val=1, type=int, comment="Number of experiments to run")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > cc @dddg617
> 
> Sure.

I think, we may provide interface to show all supported models and datasets. For example.
```python
>>> import dglgo
>>> pipeline = dglgo.pipelines['graphpred']
>>> all_datasets = pipeline.all_datasets
['cora', 'reddit', ....]
>>> all_models = pipeline.all_models
```
If the pipeline is 'linkpred', we may need to distinguish node models and edge models and provide the interface of 'Negative Sampler'.

What is more, we may support more optimizer, like SGD. And provide the interface.

dglgo/recipes/graphpred_hiv_pna.yaml
mufeili: Updated

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
mufeili: Previously this was for handling labels of a CSV dataset. This is now handled by `CSVDataset`.

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
mufeili: That's not supported with this PR. Generally speaking, that will likely require an intermediate wrapper for loss function and metric.

dglgo/dglgo/pipeline/graphpred/graphpred.jinja-py
mufeili: Ok.

dglgo/dglgo/model/graph_model/pna.py
mufeili: What will be the possible options?

dglgo/setup.py
mufeili: Someone built a pip distribution for `rdkit`, so it's no longer the case.

jermainewang (approved):
LGTM. Two small comments.
- For the `rdkit-pypi` package, could we specify the depending version?
- We could expose `node_encoder` and `atom_encoder` to be string options. See my reply above.

dglgo/dglgo/model/graph_model/pna.py
jermainewang: For example, we can expose `node_encoder` as a string with two options `'atom'` for `AtomEncoder` while `'linear'` for `nn.Linear`.

mufeili:
> * For the `rdkit-pypi` package, could we specify the depending version?

Added

> * We could expose `node_encoder` and `atom_encoder` to be string options. See my reply above.

There are several questions then.

1. Whether we want to have separate arguments in the configuration file for node and edge encoders. I'll prefer biasing for simplicity having a single argument for now.
2. `atom_encoder` and `bond_encoder` only work for molecular graphs and they are the default options for `ogbg-mol*`. For other datasets like `ogbg-code2`, there can be other applicable encoders. The question is then if we want to have explicit argument value like `ogbg_mol` or alternatively something like `default` for datasets with default node/edge encoders. It's possible to allow both `ogbg_mol` and `default` at the same time.

jermainewang:
> There are several questions then.

I see. Let us discuss this more but not let it block this PR. Please go ahead and merge this.

mufeili:
> > There are several questions then.
> 
> I see. Let us discuss this more but not let it block this PR. Please go ahead and merge this.

Sure.

Also `rdkit-pypi` uses tags like `2022.3.2.1` rather than version numbers in releases. When specifying `rdkit-pypi==2022.3.2.1` in `setup.py`, the CI cannot properly install the dependencies when installing `DGL-Go` from source. Therefore I've changed `rdkit-pypi==2022.3.2.1` back to `rdkit-pypi` for now.

PR #3928, opened 2022-04-14
Fix benchmark time measurement

daniil-sizov:
Benchmarks should not include dataloader iterator initialization when measuring time, it is constant and should be ignored

* Combines warmup and benchmarked iterations in one loop
* Reduces bench_rgcn_homogeneous_ns am batch size from 1024 to 64


Rhett-Ying:
@dgl-bot run g4dn.8xlarge all dmlc/0.8.x

Rhett-Ying:
@dgl-bot run r6i.16xlarge all dmlc/0.8.x

daniil-sizov:
Reducing bench_rgcn_homogeneous_ns am batch size from 1024 to 64 (sample size is too small for 1024 batch size, was enough only for 1 iteration)

Rhett-Ying:
@daniil-sizov Thanks for your suggestion. But we'd like to include the dataloader instantiation time as we'd better measure the end2end time.

Rhett-Ying (commented):
Rejected as we'd like to include the dataloader instantiation time.

BarclayII:
> But we'd like to include the dataloader instantiation time as we'd better measure the end2end time.

@Rhett-Ying This is not true since the benchmarks only measure 9 iterations, and for a real end2end time we need to run through the entire epoch.  I also think measuring the per-minibatch time is a good idea, since for large datasets the instantiation time will be smoothed out.

@daniil-sizov Does your modification on DataLoader affinity impact the instantiation time significantly (which I don't think it should)?  Or is the instantiation time eclipsing the optimization you made because there's too few minibatches?

BarclayII (approved):

PR #3946, opened 2022-04-22
[Bug] Fix multi-GPU edge classification crashing with pure GPU sampling

BarclayII:
The situation is described in slack channel (with the same OP of https://discuss.dgl.ai/t/heterogenous-graph-edge-sampling-on-gpu/2892).  Basically when doing edge classification with multi-GPU pure GPU sampling it crashes with a device mismatch error.  This is caused by the indices not being copied to GPU in the collate wrapper.


yaox12:
LGTM since it's a small fix.

yaox12 (approved):

Issue #3944, opened 2022-04-22
optimizer definition on line 55 in dgl/examples/pytorch/gnn_explainer/explain_main.py is wrong


sharique1006:
## 🐛 Bug

In `dgl/examples/pytorch/gnn_explainer/explain_main.py` on line 55, you have defined the optimizer as 
`optim = th.optim.Adam(explainer.parameters(), lr=args.lr, weight_decay=args.wd)`

According to the paper, model entropy H(Y) in the objective function is constant as the model is fixed. We only want to optimize the masks.
Since you are passing explainer.parameters(), the optimizer will also optimize the model parameters as the model parameters are present in explainer.parameters(). This will change the model as well as can lead to exploding prediction logits when training the explainer.

Instead, you should define it as 
`optim = th.optim.Adam([explainer.edge_mask, explainer.node_feat_mask], lr=args.lr, weight_decay=args.wd)`

This is how it has also been defined in the source code of GNNExplainer.
You can check the same here on line 614, 622 
[https://github.com/RexYing/gnn-model-explainer/blob/master/explainer/explain.py](url)

mufeili:
Nice catch! Could you open a PR to fix the issue?

CC @zhjwy9343 for awareness.

PR #3948, opened 2022-04-23
[Bug] Fixed optimizer definition in gnn_explainer (#3944)

sharique1006:
## Description
Fixed bug #3944 
optimizer needs to be defined only on the mask since the model is fixed.

@mufeili @zhjwy9343

### DIFF
--- a/examples/pytorch/gnn_explainer/explain_main.py
+++ b/examples/pytorch/gnn_explainer/explain_main.py
@@ -52,7 +52,7 @@ def main(args):
                                     node_feat_dim=feat_dim)
 
     # define optimizer
-    optim = th.optim.Adam(explainer.parameters(), lr=args.lr, weight_decay=args.wd)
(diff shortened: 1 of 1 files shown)


mufeili (approved):
LGTM. Thanks.

mufeili:
Have you tried running the updated example code on your machine?

sharique1006:
> Have you tried running the updated example code on your machine?

Yes, you can see an example run here. https://colab.research.google.com/drive/1fj0cDs53vDODESuCFW2AfigrvzcIuIxj?usp=sharing. The explanations generated after making the change are better than earlier. This is evident from the color of important edges.

Why is CI jenkins build being aborted ?

mufeili:
> > Have you tried running the updated example code on your machine?
> 
> Yes, you can see an example run here. https://colab.research.google.com/drive/1fj0cDs53vDODESuCFW2AfigrvzcIuIxj?usp=sharing. The explanations generated after making the change are better than earlier. This is evident from the color of important edges.
> 
> Why is CI jenkins build being aborted ?

Thanks. We will handle the CI part.

PR #3013, opened 2021-06-11
[bugfix] Fix sparse_optim when the state is stored on the CPU (fixes #2760)

nv-dlasalle:
## Description
This fixes #2760.

Makes copies in the SparseAdamOptimizer blocking if they are from the GPU to the CPU, to ensure we don't operate on them until it is safe to do. Copies in the other directions are safe, since the work will be launched on the same stream as the copy.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage
- [x] Code is well-documented
- [x] To the my best knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [x] Related issue is referred in this PR

PR #3971, opened 2022-05-04
[Dist][Optim] Fixed race conditions in distributed SparseAdam and SparseAdagrad

ndickson-nvidia:
## Description
* Fixed race condition bug in python/dgl/distributed/optim/pytorch/sparse_optim.py's SparseAdam::update and SparseAdagrad::update when transferring state data from GPU to CPU, corresponding with the bug fixed in the non-distributed version in https://github.com/dmlc/dgl/pull/3013 , though using the newer Event-based approach from that corresponding function in python/dgl/optim/pytorch/sparse_optim.py .  The race condition would often result in NaNs, like the previously fixed bug. https://github.com/dmlc/dgl/issues/2760
* Fixed typo in all copies of a repeatedly-copied comment near bugs fixed, checking all implementations nearby for a corresponding bug.  (All of the remaining bugs appear to have been fixed in this set of changes.)

## Checklist
- [ ] All changes have test coverage (could go in tests/pytorch/test_dist_optim.py , but it would need to be updated to support GPUs, which may or may not require infrastructure changes to run)
- [x] Code is well-documented

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -97,7 +97,7 @@ def tutorial_test_linux(backend) {
 def go_test_linux() {
   init_git()
   unpack_lib('dgl-cpu-linux', dgl_linux_libs)
-  timeout(time: 20, unit: 'MINUTES') {
+  timeout(time: 30, unit: 'MINUTES') {
     sh "bash tests/scripts/task_go_test.sh"
   }
 }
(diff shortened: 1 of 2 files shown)


classicsong (approved):
LGTM

PR #3977, opened 2022-05-06
[Dist] Enable maximum try times for socket backend via DGL_DIST_MAX_T…

Rhett-Ying:
…RY_TIMES

## Description

In some cases, it may take long time in server for graph loading and client will timeout for building connection with server. In this fix, user is able to control the maximum try times via environment variable: `DGL_DIST_MAX_TRY_TIMES`.
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/rpc_client.py
+++ b/python/dgl/distributed/rpc_client.py
@@ -165,12 +165,18 @@ def connect_to_server(ip_config, num_servers, max_queue_size=MAX_QUEUE_SIZE,
     rpc.create_sender(max_queue_size, net_type)
     rpc.create_receiver(max_queue_size, net_type)
     # Get connected with all server nodes
+    max_try_times = int(os.environ.get('DGL_DIST_MAX_TRY_TIMES', 1024))
     for server_id, addr in server_namebook.items():
         server_ip = addr[1]
         server_port = addr[2]
+        try_times = 0
         while not rpc.connect_receiver(server_ip, server_port, server_id):
+            try_times += 1
+            if try_times >= max_try_times:
+                raise rpc.DistConnectError(max_try_times, server_ip, server_port)
             time.sleep(3)
-    rpc.connect_receiver_finalize()
+    if not rpc.connect_receiver_finalize(max_try_times):
+        raise rpc.DistConnectError(max_try_times)
     # Get local usable IP address and port
     ip_addr = get_local_usable_addr(server_ip)
     client_ip, client_port = ip_addr.split(':')
--- a/python/dgl/distributed/rpc_client.py
+++ b/python/dgl/distributed/rpc_client.py
@@ -172,6 +172,9 @@ def connect_to_server(ip_config, num_servers, max_queue_size=MAX_QUEUE_SIZE,
         try_times = 0
         while not rpc.connect_receiver(server_ip, server_port, server_id):
             try_times += 1
+            if try_times % 200 == 0:
+                print("Client is trying to connect server receiver: {}:{}".format(
+                    server_ip, server_port))
             if try_times >= max_try_times:
                 raise rpc.DistConnectError(max_try_times, server_ip, server_port)
             time.sleep(3)
--- a/src/rpc/net_type.h
+++ b/src/rpc/net_type.h
@@ -27,7 +27,7 @@ struct RPCBase {
 };
 
 struct RPCSender : RPCBase {
-  static constexpr int kMaxTryCount = 1024;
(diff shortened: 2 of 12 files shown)


python/dgl/distributed/rpc_client.py
```diff
@@ -165,12 +165,18 @@ def connect_to_server(ip_config, num_servers, max_queue_size=MAX_QUEUE_SIZE,
     rpc.create_sender(max_queue_size, net_type)
     rpc.create_receiver(max_queue_size, net_type)
     # Get connected with all server nodes
+    max_try_times = int(os.environ.get('DGL_DIST_MAX_TRY_TIMES', 1024))
```
jermainewang, line 168: On L177, there is a 3 sec interval between two retries. Meaning the max timeout will be ~1 hour. Users may still be confused by this. Consider setting it to be 5~10 minutes (e.g., 100~200 retries).

src/rpc/net_type.h
```diff
@@ -27,6 +27,8 @@ struct RPCBase {
 };
 
 struct RPCSender : RPCBase {
+  static constexpr int kMaxTryCount = 1024;
```
jermainewang, line 30: Consider remove this. You are passing the max retry times from Python to C++ anyway.

tests/distributed/test_rpc.py
```diff
@@ -278,6 +278,35 @@ def test_multi_client_groups():
...
+
+    # small max try times
+    os.environ['DGL_DIST_MAX_TRY_TIMES'] = '1'
+    expect_except = False
+    try:
+        start_client(ip_config, 0, 1, net_type)
+    except dgl.distributed.DistConnectError as err:
+        print("Expected error: {}".format(err))
+        expect_except = True
+    assert expect_except
```
jermainewang, line 300: Will this test case go through somehow? It may cause random CI failure of future PRs.

Rhett-Ying: I don't think it will cause random failure as exception will always be triggered due to server is not started at all.

python/dgl/distributed/rpc_client.py
Rhett-Ying: Timeout in `rpc.connect_receiver` and `rpc.connect_receiver_finalize` are mutually exclusive. The former one  takes effect for `tensorpipe` only, the latter one takes effect for `socket` backend only. It looks confusing.

src/rpc/net_type.h
Rhett-Ying: I assign a default value for `max_try_times` in `ConnectReceiverFinalize` on purpose. As it's in base class and `max_try_times` may not be passed in some cases such as `//src/graph/network.cc`.

python/dgl/distributed/rpc_client.py
jermainewang: Understand that there are two different retry logics. My concern is that the retry times is too large, meaning there will be long silence till a timeout error. Users will probably think the program is hanging.

src/rpc/net_type.h
jermainewang: My concern is that there are two constants: one coded here and another from environment variable. People may forget to change them together in the future. It's better to have only one version.

tests/distributed/test_rpc.py
jermainewang: Is it guaranteed that for this test case the server will not start? I'm afraid that even setting retry number to 1, it will still connect in some cases.

python/dgl/distributed/rpc_client.py
Rhett-Ying: printing some log continuously when re-trying may be better. `10min` may be too smaller for a large graph. I choose `1024` here because the original value is `1000` in `socket` backend. But I'm ok with changing to `10min` here.

jermainewang: +1 on some log print. What will affect the server launching time? Why larger graph will be an issue?

Rhett-Ying: Loading a large graph may take long time and this happens before rpc involves in.

tests/distributed/test_rpc.py
Rhett-Ying: Guaranteed.

src/rpc/net_type.h
Rhett-Ying: ok. let me remove it.

jermainewang (approved):

PR #3978, opened 2022-05-06
[Peformance] Remove unnecessary induced vertices in EdgeSubgraph

yaox12:
## Description

Remove unnecessary induced vertices in `EdgeSubgraph()`, which accounts for the performance regression of EdgeDataLoader on large graphs.

## Checklist
- [x] Code is well-documented

## Changes
BTW, remove some unused code.


src/graph/unit_graph.cc
```diff
@@ -408,9 +408,9 @@ class UnitGraph::COO : public BaseHeteroGraph {
       IdArray new_src = aten::IndexSelect(adj_.row, eids[0]);
       IdArray new_dst = aten::IndexSelect(adj_.col, eids[0]);
       subg.induced_vertices.emplace_back(
-          aten::Range(0, NumVertices(SrcType()), NumBits(), Context()));
+          aten::NullArray(DLDataType{kDLInt, NumBits(), 1}, Context()));
       subg.induced_vertices.emplace_back(
-          aten::Range(0, NumVertices(DstType()), NumBits(), Context()));
+          aten::NullArray(DLDataType{kDLInt, NumBits(), 1}, Context()));
```
jermainewang, line 413: The change seems to violating the designed behavior of edge_subgraph https://docs.dgl.ai/generated/dgl.edge_subgraph.html

yaox12: According to the comments here, it should be fine to return an empty array if the mapping is identity.
https://github.com/dmlc/dgl/blob/34981bd676bc3429ab2ad4c26a27e7e76ee3e0b3/include/dgl/base_heterograph.h#L522-L528

BarclayII: I think it makes sense not to return the node IDs if relabel_nodes is False.

jermainewang: @yaox12 Could you please help to add a proper test at https://github.com/dmlc/dgl/blob/35e66f426a95da5883d04ccc81b7cfa5fd677d57/tests/compute/test_subgraph.py#L32 for such corner case? Just to make sure there is no user-side behavior change.

yaox12: Added unit tests for homograph, as for heterograph, there are tests in https://github.com/dmlc/dgl/blob/35e66f426a95da5883d04ccc81b7cfa5fd677d57/tests/compute/test_subgraph.py#L247-L252

jermainewang (approved):
LGTM

PR #3980, opened 2022-05-07
[Optimization] Memory consumption optimization on index shuffling in dataloader

BarclayII:
The changes include (1) using `np.random.shuffle` to shuffle indices in-place.  (2) use `torch.arange(out=...)` to create the consecutive integer arrays in-place.  This will avoid creating lots of temporary arrays that have the same size as training indices, which should be helpful in large-graph link prediction training.

Also fixes the usage of `graph.device` check in `CollateWrapper` to make it support custom graph storage types where `device` attribute is not applicable.


python/dgl/dataloading/dataloader.py
```diff
@@ -66,7 +67,7 @@ def __next__(self):
         batch = self._next_indices()
         if self.mapping_keys is None:
             # clone() fixes #3755, probably.  Not sure why.  Need to take a look afterwards.
-            return batch.clone()
+            return batch
```
jermainewang, line 70: Did you find out the root cause of L69?

python/dgl/dataloading/dataloader.py
```diff
@@ -134,15 +135,13 @@ def __init__(self, indices, batch_size, drop_last):
...
-        self._indices[:] = torch.arange(self._id_tensor.shape[0])
+        self._indices = torch.arange(self._id_tensor.shape[0], dtype=torch.int64).share_memory_()
         self.batch_size = batch_size
         self.drop_last = drop_last
 
     def shuffle(self):
         """Shuffle the dataset."""
-        # TODO: may need an in-place shuffle kernel
-        self._indices[:] = self._indices[torch.randperm(self._indices.shape[0])]
+        np.random.shuffle(self._indices.numpy())
```
jermainewang, line 144: What if the `_indices` is in GPU?

python/dgl/dataloading/dataloader.py
```diff
@@ -514,10 +517,12 @@ def __init__(self, sample_func, g, use_uva, device):
         self.device = device
 
     def __call__(self, items):
-        if self.use_uva or (self.g.device != torch.device('cpu')):
-            # Only copy the indices to the given device if in UVA mode or the graph is not on
-            # CPU.
-            items = recursive_apply(items, lambda x: x.to(self.device))
+        if self.use_uva:
+            graph_device = getattr(self.g, 'device', None)
```
jermainewang, line 521: If this because graph storage may not have a `device` attribute?

BarclayII: Yes.

python/dgl/dataloading/dataloader.py
BarclayII: `_indices` is always on CPU.

python/dgl/dataloading/dataloader.py
BarclayII: Reverted.  This still works on some machines while not work on others...

jermainewang (approved):

PR #3981, opened 2022-05-07
[Example] a custom graph storage example

BarclayII:
This is an example of using a custom graph storage where node and edge data are numpy arrays.  This should be a minimal example of how to build a custom graph storage with custom feature storages.

@wangxiaoyunNV @nv-dlasalle @jermainewang Please let me know your thoughts on this example.

### DIFF
--- a/examples/pytorch/graphsage/numpy_example.py
+++ b/examples/pytorch/graphsage/numpy_example.py
@@ -1,0 +1,181 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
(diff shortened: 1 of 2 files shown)


jermainewang:
We probably need a more thorough doc on this. This example could belong to it.

BarclayII:
> We probably need a more thorough doc on this. This example could belong to it.

I understand.  This is a user experience example so I think it's better to show you first and discuss, so that it won't take too long for you to notice whether the UX can be improved.  I'm working on the documentation in parallel.

PR #3983, opened 2022-05-08
[CMake] Proposal to make USE_AVX OFF by default

msharmavikram:
## Description
With the emergence of new ISA (like ARM and RISCV) retaining USE_AVX ON default makes the default build instructions fail. Fundamentally DGL does not require the use of AVX for functional working. AVX is mainly needed when to enable optimization. So the proposal is to default turn it off and then later during build instructions, folks with AVX capabilities can enable it with 
`cmake .. -DUSE_AVX=ON`

## Checklist
 
## Changes
- [x] Update to change the default CMakefile fields.


yzh119:
I have no problem setting it to `OFF`.
What concerns me more is whether to release both `dgl-intel` and `dgl` for CPU version of dgl, @BarclayII WDYT?

msharmavikram:
> 

I am not sure if there is a need for dgl-intel and dgl. Because we could essentially add a segment in the README.md file to state for optimized x86 specific architecture, we recommend using `cmake .. -DUSE_AVX=ON`. (Pretty much follow what other popular repos do to handle these situations.)

yzh119:
@msharmavikram what I mean is the pre-built wheels, the `USE_AVX` flag was turned on by default when we build pip wheels.
When users install `dgl-cpu` via conda or pip, intel optimizations would be installed as well even though they will not be activated (we have runtime check about it so you will not encounter an error on non-intel machines).

If we keep the behavior unchanged, then you have to activate `USE_AVX` in our [building script](https://github.com/dmlc/dgl/blob/edf2d52666e60683c276d6cb8aa813be942e4cbb/tests/scripts/build_dgl.sh) so that the CI and wheel building process would not be influenced.

msharmavikram:
@yzh119 Thanks for the reasoning. I understand and agree. 

We tried this on M1 Mac pro: `pip install dgl dglgo -f https://data.dgl.ai/wheels/repo.html` 
But this did not work as there was no matching binary. This was run on top of `nvcr.io/nvidia/pytorch:22.04-py3` docker image as base. After failing this step, we tried source compilation - disabled `USE_AVX`, compiled the binary, and then did `python setup.py install`. This works fine. (I will be publishing soon a git gist for people who are trying to compile DGL on ARM). 

After understanding the root cause, we realized, it makes sense to make this optimization `OFF` by default. Perhaps updating the CI (build scripts) to have the USE_AVX=ON field is better. This will at least ensure - out of the box source compilation works with minimum requirements.

yzh119 (approved):
@BarclayII would you mind updating CI and build scripts accordingly? thanks.

Rhett-Ying:
I think we'd better add `-DUSE_AVX=ON` in any existing build scripts including CI and pre-built packages which may be affected by this change before merge this PR.
@BarclayII pls be aware of the impact of this change.

PR #3984, opened 2022-05-09
[DistTest] add basic pipeline for dist test across machines

Rhett-Ying:
## Description

In order to run dist tests across machines, we need to follow below steps:
1. set up a dist environment which is the same for dist DGL training. just follow the `README` in https://github.com/dmlc/dgl/tree/master/examples/pytorch/graphsage/dist
2. build dgl and tests. `cmake -DUSE_CUDA=ON -DBUILD_TORCH=ON -DTORCH_PYTHON_INTERPS=python3 -DBUILD_CPP_TEST=ON .. `
3. run tests under //dgl directory: `bash tests/scripts/task_dist_test.sh`
## Checklist
- [ ] Code is well-documented


tests/dist/cpp/rpc_client.cc
```diff
@@ -0,0 +1,79 @@
...
+    for (int i = 0; i < kNumSender; ++i) {
+      threads.push_back(std::thread(&RPCClient::start_client, this));
+    }
+    for (auto &&t : threads) {
+      t.join();
+    }
+  }
+
+private:
+  void parse_ips() {
```
jermainewang, line 27: Make sure you stick with camel naming for C++ code.

tests/dist/cpp/rpc_base.h
```diff
@@ -0,0 +1,29 @@
...
+#include "../../../src/rpc/tensorpipe/tp_communicator.h"
+#include "../../cpp/common.h"
+
+namespace {
+const int kNumSender = 30;
+const int kNumReceiver = 10;
+const int kNumMessage = 1024;
+const int kPort = 50090;
+const int kSizeTensor = 10 * 1024;
+const int kNumTensor = 30;
```
jermainewang, line 17: Can we expose them to program arguments? This will make it easier to test different settings.

jermainewang: DMLC provides utilities to parse key-value args. See https://github.com/dmlc/dmlc-core/blob/main/example/parameter.cc . Also JSON support: https://github.com/dmlc/dmlc-core/blob/main/include/dmlc/json.h

jermainewang (changes requested):

tests/dist/cpp/rpc_base.h
Rhett-Ying: I think more about it. As these configurations should be same both for clients and servers. That's why I configure them in the common head file. If we expose them to program arguments, we cannot control or assure these configurations are same for clients and servers. As these are specific for tests on specific machine, it's acceptable to configure like this instead of exposure?

jermainewang (approved):

jermainewang (approved):

jermainewang (approved):

PR #3990, opened 2022-05-10
change the curandState and launch dimension of CSRRowwiseSample kernel

paoxiaode:
## Checklist
- [x] Code is well-documented

## Changes
change the curandState and launch dimension of CSRRowwiseSample kernel
Original curandState: `curandState rng;` , high curand_init cost
new curandState: `curandStatePhilox4_32_10_t rng;`, low curand_init cost

Original launch dimension: 32x4
new launch dimension: 128x1

## Experiments
test on GraphSage、RGCN、GAT e2e training

**RGCN**：ogbn-mag dataset, two layers, fanout = 25,30, bs=512
|  | SM throughput (%) | MEM throughput (%) | Duration(us)|
| ---- | ---- | ---- | ---- |
| Origin | 1.02 | 0.9 | 4808|
| Optimize | 11.81 | 11.21 | 1204|

**GraphSage**：reddit dataset, two layers, fanout = 10,25, bs=1000
|  | SM throughput (%) | MEM throughput (%) | Duration(us)|
| ---- | ---- | ---- | ---- |
| Origin | 25.32 | 16.45 | 848.45|
| Optimize | 10.31 | 58.49 | 245.63|

**GAT**：ogbn-products dataset, three layers, fanout = 10,10,10, bs=512
|  | SM throughput (%) | MEM throughput (%) | Duration(us)|
| ---- | ---- | ---- | ---- |
| Origin | 6.68 | 7.71 | 791|
| Optimize | 17.65 | 7.3 | 331|


src/array/cuda/rowwise_sampling.cu
```diff
@@ -345,10 +345,10 @@ COOMatrix CSRRowWiseSamplingUniform(CSRMatrix mat,
         out_cols,
         out_idxs);
   } else {
-    constexpr int BLOCK_WARPS = 128/WARP_SIZE;
+    constexpr int BLOCK_WARPS = 128/CTA_SIZE;
```
nv-dlasalle, line 348: I think we need to rename this variable from `BLOCK_WARPS`.

src/array/cuda/rowwise_sampling.cu
```diff
@@ -22,6 +22,7 @@ namespace impl {
 namespace {
 
 constexpr int WARP_SIZE = 32;
```
nv-dlasalle, line 24: is `WARP_SIZE` still used?

nv-dlasalle (commented):
Great performance results.

Can we also update the `_CSRRowWiseSampleReplaceKernel()` kernel to keep it in sync with this one?

Also can you elaborate on your timing experiments? Is this the average kernel duration over an epoch? If so can you also provide the maximum?

paoxiaode:
The kernel time/ throughput is recorded by the nsys profile, it is the sum of several instances in one iteration(3 GNN layer corresponding to 3 CSRsample kernel instances)

src/array/cuda/rowwise_sampling.cu
paoxiaode: yes, maybe ` BLOCK_CTA`?

src/array/cuda/rowwise_sampling.cu
paoxiaode: > How much does the launch dimension change affect the performance when you enable it on top of the random number generator change?

If not consider the change of random number generator, the change of launch dimension can significantly reduce the latency of model which has super nodes, eg: for RGCN, the latency reduces from 4ms to 1.5ms.
Combine these two changes, we can see significant change on three models.

paoxiaode:
> Great performance results.
> 
> Can we also update the `_CSRRowWiseSampleReplaceKernel()` kernel to keep it in sync with this one?
> 
> Also can you elaborate on your timing experiments? Is this the average kernel duration over an epoch? If so can you also provide the maximum?

I think for `_CSRRowWiseSampleReplaceKernel()`, we can keep it in sync with this one.

The kernel time/ throughput is recorded by the nsys profile, it is the sum of several instances in one iteration(3 GNN layer corresponding to 3 CSRsample kernel instances), I test it on 10 iteration, this is the average kernel duration.

Due to the input/output shape is same for different epoch, the kernel duration won't change a lot.

yaox12:
LGTM. After discussion with Jiahui, I find the improved kernel also handles the long-tail distribution and large batch size well. Really appreciate it!

yaox12 (approved):

paoxiaode:
Rgcn_neighbor
Graphsage_neighbor

This is two dataset which I test on, the number of non-zero elements varies greatly, I think this leads to **long-tail effect**. The 128*1 launch dimension is set to reduce the latency of rows where the number of non-zero elements is greatly larger than fan-out.
`for (int idx = num_picks+threadIdx.x; idx < deg; idx+=CTA_SIZE)` 128x1 can improve this FOR loop.


Here are the large workload results I test on two datasets:
| |num_rows| kernel duration [ns] |
|-|-|-|
|origin| 21466 | 579,552 |
|optimized| 21466 |  238,208 |
| | | | |
|origin| 9392 | 5,112,640 |
|optimized| 9392 |  1,909,088 |

nv-dlasalle (approved):
Looks good to me. Thanks for providing the stats.

Issue #3976, opened 2022-05-05
Zerocopying a tensor from pytorch is out of place if non-contiguous (removes sharing and causes issues with in place pinning).


nv-dlasalle:
## 🐛 Bug

In the PyTorch backend we use `.contiguous()` in the zerocopy functions, which is an out of place operation.
https://github.com/dmlc/dgl/blob/5b97a1a26ed5c68fb98d4452ea1672398481841f/python/dgl/backend/pytorch/tensor.py#L343

As a result, previously shared tensors become unshared. I believe this is related to the UVA issues #3939 and #3961.
Because we convert the pytorch tensor to a NDArray, but out of place, and then pin the NDArray in place. The NDArray gets freed without being unpinned. PyTorch then re-uses this memory range from its pool, but it remains pinned in the CUDA driver.
See #3800.

## To Reproduce

This script shows the issue outside of DGL:
```
import torch
from torch import multiprocessing as mp

def val(i, x):
    print("{}: x contiguous? {}".format(i, x.is_contiguous()))
    print("{}: x shared? {}".format(i, x.is_shared()))
    print("{}: x ptr 0x{:x}".format(i, x.data_ptr()))
    y = x.contiguous()
    print("{}: y contiguous? {}".format(i, y.is_contiguous()))
    print("{}: y shared? {}".format(i, y.is_shared()))
    print("{}: y ptr 0x{:x}".format(i, y.data_ptr()))

def test():
    x = torch.empty([2,2,10,5], memory_format=torch.channels_last)
    print("is x contiguous? {}".format(x.is_contiguous()))
    mp.spawn(val, (x,), nprocs=2)

if __name__ == '__main__':
        test()
```
I think we either need to support non-contiguous NDArrays (a huge effort as many kernels in DGL would need to be re-written), or throw an error if we attempt to zerocopy a non-contiguous array, and tell the user to convert to .contiguous().

jermainewang:
Can we throw an error only if a tensor needs to be pinned on our side but it is non-contiguous?

nv-dlasalle:
While pinning a temporary tensor is one of the big effects of this issue, it effects other things. Because its not actually a zerocopy at this point, anytime DGL does modifying operation on a pytorch tensor it is potentially a bug here, since any modifications will be out of place, despite the in-place semantics of zerocopy.

nv-dlasalle:
[...]

Issue #3800, opened 2022-03-03
Pinned memory fails to be unregisterd


nv-dlasalle:
## 🐛 Bug
If you allocate and free pinned memory in a loop in DGL currently, you will get:
```
terminate called after throwing an instance of 'c10::CUDAError'
  what():  CUDA error: resource already mapped
```
When that same memory region gets reallocated.

## To Reproduce

Steps to reproduce the behavior if you modify the current `examples/pytorch/graphsage/multi_gpu_node_classification.py` to allocate pinned memory inside of the inference loop:
```
def shared_tensor(*shape, device, q):
    rank = dist.get_rank()
    world_size = dist.get_world_size()
    # make sure q is only used for one operation at a time
    dist.barrier()
    if rank == 0:
        y = torch.zeros(
            *shape, device=device)
        for i in range(1, world_size):
            q.put(y)
    else:
        y = q.get()
    dist.barrier()
    pin_memory_inplace(y) # <-- error occurs here
    return y


class SAGE(nn.Module):
    def __init__(self, in_feats, n_hidden, n_classes):
        super().__init__()
        self.layers = nn.ModuleList()
        self.layers.append(dglnn.SAGEConv(in_feats, n_hidden, 'mean'))
        self.layers.append(dglnn.SAGEConv(n_hidden, n_hidden, 'mean'))
        self.layers.append(dglnn.SAGEConv(n_hidden, n_classes, 'mean'))
        self.dropout = nn.Dropout(0.5)
        self.n_hidden = n_hidden
        self.n_classes = n_classes

    def forward(self, blocks, x):
        h = x
        for l, (layer, block) in enumerate(zip(self.layers, blocks)):
            h = layer(block, h)
            if l != len(self.layers) - 1:
                h = F.relu(h)
                h = self.dropout(h)
        return h

    def inference(self, g, device, batch_size,
                  num_workers, buffer_device, q):
        sampler = dgl.dataloading.MultiLayerFullNeighborSampler(1)

        # manually partition the IDs
        dataloader = dgl.dataloading.NodeDataLoader(
                g, torch.arange(g.num_nodes(), device=device), sampler, device=device,
                batch_size=batch_size, shuffle=False, drop_last=False, num_workers=num_workers,
                persistent_workers=(num_workers > 0), use_ddp=True, use_uva=True)

        feat = g.ndata['feat']
        for l, layer in enumerate(self.layers):
            y = shared_tensor(
[...]

Issue #3939, opened 2022-04-21
0.8.1 UVA mode got error 'CUDA: part or all of the requested memory range is already mapped' when using large graph


rewonderful:
## 🐛 Bug
I modified my GraphSAGE to suit 0.8.1 fashion, however I got error which indicate the graph.feats has already been pinned, but I didn't  invoke any "pin memory operation" outside the "Datalaoder", the script failed in "Dataloader" construction stage with error message "CUDA: part or all of the requested memory range is already mapped".

Im curious that is UVA mode has any restrict on graph size ? Because another similar script using UVA mode runs normally on a smaller graph which size(with feats) is 21GB, the codes below runs on graph size with 145GB, they both run on GPU V100-16GB.

### traceback
```
Traceback (most recent call last):
  File "pred_v0.8.1.py", line 431, in <module>
    run_predict(args)
  File "pred_v0.8.1.py", line 383, in run_predict
    predict(args,model,g,device)
  File "pred_v0.8.1.py", line 286, in predict
    drop_last=False,)
  File "/opt/conda/lib/python3.7/site-packages/dgl/dataloading/dataloader.py", line 766, in __init__
    self.graph.pin_memory_()
  File "/opt/conda/lib/python3.7/site-packages/dgl/heterograph.py", line 5555, in pin_memory_
    col.pin_memory_()
  File "/opt/conda/lib/python3.7/site-packages/dgl/frame.py", line 398, in pin_memory_
    pin_memory_inplace(self.data)
  File "/opt/conda/lib/python3.7/site-packages/dgl/utils/pin_memory.py", line 8, in pin_memory_inplace
    F.to_dgl_nd(tensor).pin_memory_()
  File "/opt/conda/lib/python3.7/site-packages/dgl/_ffi/ndarray.py", line 322, in pin_memory_
    check_call(_LIB.DGLArrayPinData(self.handle))
  File "/opt/conda/lib/python3.7/site-packages/dgl/_ffi/base.py", line 65, in check_call
    raise DGLError(py_str(_LIB.DGLGetLastError()))
dgl._ffi.base.DGLError: [16:34:15] /opt/dgl/src/runtime/cuda/cuda_device_api.cc:183: Check failed: e == cudaSuccess || e == cudaErrorCudartUnloading: CUDA: part or all of the requested memory range is already mapped
Stack trace:
[...]

PR #3992, opened 2022-05-10
Prevent users from attempting to pin PyTorch non-contiguous tensors or views only encompassing part of tensor.

nv-dlasalle:
## Description
This forces `pin_memory_inplace()` to go through `zerocopy_to_dgl_ndarray_for_write()` and adds checks to make sure it can be done in place, and won't pin only a subset of a tensor storage.

This does not completely address #3976 and #3800 remains an issue.

## Checklist
- [x] Code is well-documented


python/dgl/backend/pytorch/tensor.py
```diff
@@ -349,6 +349,10 @@ def zerocopy_to_dgl_ndarray(data):
         return nd.from_dlpack(dlpack.to_dlpack(data.contiguous()))
 
 def zerocopy_to_dgl_ndarray_for_write(input):
+    assert input.is_contiguous(), "Cannot convert non-contiguous tensors " \
+        "to dgl ndarray for write."
+    assert input.numel() == input.storage().size(), "Cannot convert view " \
```
yaox12, line 354: I doubt the error message here would confuse users. For example in this issue https://github.com/dmlc/dgl/issues/3939, users would question: "I just want to use view tensors as input, not for write. Why will it throw a 'not for write' error?"

nv-dlasalle: I wrapped the calls from dgl.utils with `try`/`except`, such that the user will be told pinning the memory failed in e279be8.

tests/pytorch/test_pin_memory.py
```diff
@@ -0,0 +1,38 @@
...
+@unittest.skipIf(F._default_context_str == 'cpu', reason="Need gpu for this test")
+def test_pin_noncontiguous():
+    t = torch.empty([10, 100]).transpose(0, 1)
+
+    assert not t.is_contiguous()
+    assert not F.is_pinned(t)
+
+    try:
+        dgl.utils.pin_memory_inplace(t) 
+    except dgl.DGLError as e:
```
yaox12, line 15: We can use `with pytest.raises(DGLError):` context here.

jermainewang (approved):

tests/pytorch/test_pin_memory.py
nv-dlasalle: Switched in 6961d66

PR #3997, opened 2022-05-11
[bugfix] Explicitly unpin tensoradapter allocated arrays

nv-dlasalle:
## Description
This attempts to provide a workaround for #3800, where if users allocate ndarrays through DGL, they will be properly unpinned (e.g., Heterograph arrays). It will not affect arrays allocated via PyTorch and passed into DGL via DLPack.

## Checklist
- [x] Code is well-documented

## Changes

This adds tracking to NDArray, such that if a tensor is created with the tensoradapter, we check to see if it should be unpinned by DGL. This is safe since the tensoradapter cannot allocate pinned tensors, and pytorch does not allow pinning tensors in place.

### DIFF
--- a/include/dgl/runtime/ndarray.h
+++ b/include/dgl/runtime/ndarray.h
@@ -415,6 +415,8 @@ struct NDArray::Container {
   std::vector<int64_t> stride_;
   /*! \brief The internal array object */
   std::atomic<int> ref_counter_{0};
+
+  bool from_tensor_dispatcher_{false};
 };
 
 // implementations of inline functions
--- a/src/runtime/ndarray.cc
+++ b/src/runtime/ndarray.cc
@@ -78,6 +78,12 @@ struct NDArray::Internal {
   // This enables us to create NDArray from memory allocated by other
   // frameworks that are DLPack compatible
   static void DLPackDeleter(NDArray::Container* ptr) {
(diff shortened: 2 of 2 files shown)


yaox12:
> It will not affect arrays allocated via PyTorch and passed into DGL via DLPack.
Can we maintain a flag `pinned_by_dgl` and check it in `DefaultDeleter` and `DLPackDeleter` to include this case? Any concerns?

davidmin7:
>The problem is for array allocated outside of DGL, we can only modify the Deleter DGL will call, but not the one PyTorch/MXNET/Tensorflow will call.

Is it possible to do a deleter chaining? For example:

```
// curr_deleter is the deleter which will be called by default
original_deleter = curr_deleter;
curr_deleter = unpin;
```

and inside the `unpin()` function we do something like:
```
void unpin() {
    // Do unpinning jobs here
    ...
    // Chain to the original deleter
    original_deleter();
}
```

In this case, wouldn't the unpinning job be done during the destructor stage even if the tensor was allocated by PyTorch? The deleter function is passed as a pointer in dlpack so I think we can create a linked list of the destructors and insert the unpinning stage before the 'original' deleter is called. 2 cents of my thoughts on this issue.

BarclayII (approved):

PR #4036, opened 2022-05-21
[Bugfix] Cython CAPI holding GIL causes deadlock when Python callback is asynchronous

jermainewang:
## Description

Cython will not release GIL (global interpreter lock) when calling into C++ APIs. A deadlock will happen when a thread A calls a C++ API that is waiting for a callback back to Python in thread B. Because thread B tries to grab GIL in order to finish the callback but is held by thread A, a deadlock happens. The bug has been in DGL since the beginning of the project but is recently triggered in a corner case in distributed training.

## Checklist
- [x] Code is well-documented


BarclayII:
Should we also make other Cython functions `nogil`?

jermainewang:
> Should we also make other Cython functions `nogil`?

No. Only APIs that could be blocking need to release GIL.

VoVAllen:
Where did dgl use a python callback in C++?

VoVAllen:
This is also consistent we've seen that ctypes works well because ctypes function call will release the GIL automatically

jermainewang:
There are some reference counting problem when using NDArray argument in python callback. It is not related to the GIL issue so I disabled the unit test for now. Will fix it in another PR.

jermainewang:
@dgl-bot run g4dn.4xlarge model_speed dmlc/0.8.x

Rhett-Ying (approved):

Issue #4067, opened 2022-05-31
Running errors on one example case: example/pytorch/ogb/cluster-gat


chang-l:
## 🐛 Bug

The python script of cluster_gat example case has bugs and it crashes at multiple places.

## To Reproduce

(single GPU run)
1. yes | python main.py  --num-epochs=6  --eval-every=4

Traceback (most recent call last):
  File "main.py", line 269, in <module>
    test_accs.append(run(args, device, data))
  File "main.py", line 193, in run
    eval_acc, test_acc, pred = evaluate(model, g, nfeat, labels, val_nid, test_nid, args.val_batch_size, device)
  File "main.py", line 122, in evaluate
    return compute_acc(pred[val_nid], labels[val_nid]), compute_acc(pred[test_nid], labels[test_nid]), pred
  File "main.py", line 106, in compute_acc
    return (th.argmax(pred, dim=1) == labels).float().sum() / len(pred)
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!

## Expected behavior

The 10 repetitive runs should be finished without err msg.

## Environment

 - DGL Version (e.g., 1.0): 0.9
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): 1.12
 - OS (e.g., Linux): Ubuntu 20.04
 - How you installed DGL (`conda`, `pip`, source): source
 - Build command you used (if compiling from source):
 - Python version: 3.8.13
 - CUDA/cuDNN version (if applicable): 11.7
 - GPU models and configuration (e.g. V100): A100
 - Any other relevant information:

## Additional context

In the script, as num_of_runs is hard-coded to 10, sometimes it can be very costly. It might be a good idea to include num_of_runs in the argument-list.

mufeili:
> In the script, as num_of_runs is hard-coded to 10, sometimes it can be very costly. It might be a good idea to include num_of_runs in the argument-list.

I agree.

> Also, I found the cmd line argument "--data-cpu" is not active, or maybe has not been implemented. Is it correct?
There are two cases have the option "--data-cpu":
ogb/cluster-gat/main.py
rgcn-hetero/entity_classify_mb.py

Most likely the author copied the code from somewhere else and forgot to implement or remove that. I'm fine with either implementing it or removing it.

chang-l:
Okay, sounds good. I can add num_of_runs into argument list, not only for cluster-gat, but for other cases having the similar issue.

[...]

PR #4068, opened 2022-05-31
[Bugfix] Fix cluster-gat example

chang-l:
## Description
Found some typos in the python script for cluster-gat example. Have them fixed.
Reported issue link: https://github.com/dmlc/dgl/issues/4067 

## Changes

- To fix "Expected all tensors to be on the same device" (line 122), make a cpu-cpy to label
- "g.ndata.pop('feat').to(device)" presents within outermost loop of rerun; move it out of the loop to avoid redundant "pop"
- Similar to (1), move best_test_acc to cpu


mufeili (approved):
LGTM. Thanks for the fix.

PR #4073, opened 2022-06-01
[Data] AsGraphPredDataset

mufeili:
## Description
Separate code related to `AsGraphPredDataset` from #3927 

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/data/adapter.py
+++ b/python/dgl/data/adapter.py
@@ -9,16 +9,17 @@ from ..convert import graph as create_dgl_graph
 from ..sampling.negative import _calc_redundancy
 from .dgl_dataset import DGLDataset
 from . import utils
+from ..base import DGLError
 from .. import backend as F
 
-__all__ = ['AsNodePredDataset', 'AsLinkPredDataset']
+__all__ = ['AsNodePredDataset', 'AsLinkPredDataset', 'AsGraphPredDataset']
 
 
 class AsNodePredDataset(DGLDataset):
     """Repurpose a dataset for a standard semi-supervised transductive
     node prediction task.
 
-    The class converts a given dataset into a new dataset object that:
+    The class converts a given dataset into a new dataset object such that:
 
       - Contains only one graph, accessible from ``dataset[0]``.
       - The graph stores:
@@ -40,7 +41,7 @@ class AsNodePredDataset(DGLDataset):
         So do validation and test masks.
 
     The class will keep only the first graph in the provided dataset and
-    generate train/val/test masks according to the given spplit ratio. The generated
+    generate train/val/test masks according to the given split ratio. The generated
     masks will be cached to disk for fast re-loading. If the provided split ratio
     differs from the cached one, it will re-process the dataset properly.
 
@@ -49,7 +50,7 @@ class AsNodePredDataset(DGLDataset):
     dataset : DGLDataset
         The dataset to be converted.
     split_ratio : (float, float, float), optional
-        Split ratios for training, validation and test sets. Must sum to one.
+        Split ratios for training, validation and test sets. They must sum to one.
     target_ntype : str, optional
         The node type to add split mask for.
 
@@ -193,7 +194,7 @@ class AsLinkPredDataset(DGLDataset):
     """Repurpose a dataset for link prediction task.
 
     The created dataset will include data needed for link prediction.
-    Currently only support homogeneous graph. 
+    Currently it only supports homogeneous graphs.
     It will keep only the first graph in the provided dataset and
     generate train/val/test edges according to the given split ratio,
     and the correspondent negative edges based on the neg_ratio. The generated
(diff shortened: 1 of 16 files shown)


python/dgl/data/adapter.py
```diff
@@ -368,3 +369,155 @@ def __getitem__(self, idx):
...
+    def process(self):
+        is_ogb = hasattr(self.dataset, 'get_idx_split')
+        if self.split_ratio is None:
+            if is_ogb:
+                split = self.dataset.get_idx_split()
+                self.train_idx = split['train']
+                self.val_idx = split['valid']
+                self.test_idx = split['test']
+            else:
+                # Handle FakeNewsDataset
```
jermainewang, line 442: Is this comment still necessary? Also, what's the convention of our dataset for graph prediction? Using mask tensor or index tensor?

python/dgl/data/adapter.py
```diff
@@ -368,3 +369,155 @@ def __getitem__(self, idx):
...
+            num_graphs = len(self.dataset)
+            num_train = int(num_graphs * train_ratio)
+            num_val = int(num_graphs * val_ratio)
+
+            idx = np.random.permutation(num_graphs)
+            self.train_idx = F.tensor(idx[:num_train])
+            self.val_idx = F.tensor(idx[num_train: num_train + num_val])
+            self.test_idx = F.tensor(idx[num_train + num_val:])
+
+        if hasattr(self.dataset, 'num_classes'):
```
jermainewang, line 462: Suggest to add `num_classes` to all related Dataset class too.

python/dgl/data/adapter.py
```diff
@@ -368,3 +369,155 @@ def __getitem__(self, idx):
...
+                self.val_idx = split['valid']
+                self.test_idx = split['test']
+            else:
+                # Handle FakeNewsDataset
+                try:
+                    self.train_idx = F.nonzero_1d(self.dataset.train_mask)
+                    self.val_idx = F.nonzero_1d(self.dataset.val_mask)
+                    self.test_idx = F.nonzero_1d(self.dataset.test_mask)
+                except:
+                    raise DGLError('split_ratio is required to generate the split.')
```
jermainewang, line 448: `'The input dataset does not have default train/val/test split. Please specify split_ratio to generate the split.'`

python/dgl/data/adapter.py
```diff
@@ -368,3 +369,155 @@ def __getitem__(self, idx):
...
+            self.num_classes = self.dataset.num_classes
+        else:
+            # None for multi-label classification
+            self.num_classes = None
+
+        if hasattr(self.dataset, 'num_tasks'):
+            # OGB datasets
+            self.num_tasks = self.dataset.num_tasks
+        elif hasattr(self.dataset, 'num_labels'):
+            # QM7bDataset, QM9Dataset, QM9EdgeDataset
```
jermainewang, line 473: Please update these datasets. Change `num_labels` to `num_classes` and deprecate the former one.

python/dgl/data/adapter.py
```diff
@@ -368,3 +369,155 @@ def __getitem__(self, idx):
...
+                 test_idx=F.zerocopy_to_numpy(self.test_idx))
+
+    def __getitem__(self, idx):
+        return self.dataset[idx]
+
+    def __len__(self):
+        return len(self.dataset)
+
+    @property
+    def node_feat_size(self):
```
jermainewang, line 516: Is this documented?

jermainewang (approved):
LGTM. Some minor suggestions.

jermainewang:
@Rhett-Ying  Could you take a look at the changes to `CSVDataset`?

python/dgl/data/csv_dataset.py
```diff
@@ -141,14 +144,21 @@ def load(self):
...
+        if F.is_tensor(i) and F.ndim(i) == 1:
+            return Subset(self, F.copy_to(i, F.cpu()))
+
         if self._transform is None:
             g = self.graphs[i]
         else:
             g = self._transform(self.graphs[i])
 
-        if len(self.data) > 0:
+        if len(self.data) == 1:
```
Rhett-Ying, line 159: after this change, if we have single key in `self.data`, such key will be returned directly such as `label` or `train_mask`. If we have multiple keys in `self.data` such as `label` and `train_mask`, then `label` will not be returned directly. Right? all do we want to always return `label` if it exits?

python/dgl/data/adapter.py
mufeili: `FakeNewsDataset` is so far the only built-in dataset with a specific dataset split. It uses mask tensors. However, in my opinion it makes more sense to use index tensors as most experiments for graph prediction iterate over index tensors for minibatch training.

python/dgl/data/adapter.py
mufeili: Done

python/dgl/data/adapter.py
mufeili: Done

python/dgl/data/adapter.py
mufeili: These are for regression. I've replaced `num_labels` with `num_tasks`.

python/dgl/data/csv_dataset.py
mufeili: > such key will be returned directly such as label or train_mask

The value associated with the key will be returned directly. Mostly this will be the case for labels.

> If we have multiple keys in self.data such as label and train_mask, then label will not be returned directly. Right? all do we want to always return label if it exits?

My implementation is the former case.

python/dgl/data/adapter.py
mufeili: updated

Issue #4031, opened 2022-05-20
AtomicAdd is not implemented correctly for old architectures


nv-dlasalle:
## 🐛 Bug

For double if arch < 60 and for half < 70, it is just implemented as (https://github.com/dmlc/dgl/blob/master/src/array/cuda/atomic.cuh#L227):
```
  return *addr + val;
```
But we should be using atomicCAS to implement such things when not hardware supported (from the CUDA docs https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomic-functions):
```
#if __CUDA_ARCH__ < 600
__device__ double atomicAdd(double* address, double val)
{
    unsigned long long int* address_as_ull =
                              (unsigned long long int*)address;
    unsigned long long int old = *address_as_ull, assumed;

    do {
        assumed = old;
        old = atomicCAS(address_as_ull, assumed,
                        __double_as_longlong(val +
                               __longlong_as_double(assumed)));

    // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
    } while (assumed != old);

    return __longlong_as_double(old);
}
#endif
```

nv-dlasalle:
It looks like min/max operators are similarly broken: https://github.com/dmlc/dgl/blob/master/src/array/cuda/fp16.cuh#L15

mufeili:
Nice catch! Could you help fix this issue?

ndickson-nvidia:
PR #4079 should hopefully have fixed the cases that are possible to fix.  `AtomicAdd<half>` for architectures before 70 won't be atomic, since there's no 16-bit `atomicCAS` support on the older GPUs, but the other functions should be fixed.

jermainewang:
@ndickson-nvidia Do you think we should resolve this issue?

ndickson-nvidia:
@jermainewang I think it's probably as fixed as it can be.  @nv-dlasalle suggested the possibility of raising an error if the FP16 one gets run on old hardware, though I suppose there may be cases where atomicity isn't strictly needed that happen to hit the codepath and it'd be unfortunate if those error out, but maybe there are no cases like that.  Maybe there's a way to warn on the first time encountering it?  Any thoughts?

nv-dlasalle:
[...]

PR #4029, opened 2022-05-19
[Bug][Feature] Added cublasGemm<__half> specialization (#3988)

ndickson-nvidia:
## Description
Added specialization of cublasGemm function for `__half` type, to try to address https://github.com/dmlc/dgl/issues/3988  It's just a copy of the `float` specialization with `float` changed to `__half`, `cublasSgemm` changed to `cublasHgemm`, and adding a `USE_FP16` guard.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage
- [x] Code is well-documented
- [x] To the best of my knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [x] Related issue is referred in this PR

PR #4079, opened 2022-06-02
[Bug] Added common operations for FP16 on older GPUs

ndickson-nvidia:
## Description
* Fixed use of `__CUDA_ARCH__` around `SWITCH_BITS` macro that was causing an incorrect error message and preventing the changes in https://github.com/dmlc/dgl/pull/4029 from working.
* Added support for common operations on FP16 (half or __half) for older GPU architectures, which was necessary in order to get the code to compile for older GPUs after fixing the use of `__CUDA_ARCH__`.
* Fixed AtomicAdd to be atomic for float and double for old GPU architectures, fixing https://github.com/dmlc/dgl/issues/4031 .  Unfortunately, it seems that atomicCAS for unsigned short seems to be unavailable until architecture 70, so half will have to stay non-atomic on old GPUs.  Also fixed min & max as mentioned in the same issue.

## Checklist
- [ ] All changes have test coverage (Thoroughly testing FP16 will require future work.  This is mostly to get it compiling, first.)
- [x] Code is well-documented


nv-dlasalle (approved):

src/array/cuda/atomic.cuh
```diff
@@ -228,23 +238,33 @@ __device__ __forceinline__ double AtomicAdd<double>(double* addr, double val) {
...
 template <>
 __device__ __forceinline__ half AtomicAdd<half>(half* addr, half val) {
+// half make sure we have half support
 #if __CUDA_ARCH__ >= 700
   return atomicAdd(addr, val);
 #else
-  return *addr + val;
+  half res = half(float(*addr) + float(val));
+  *addr = res;
+  return res;
```
yaox12, line 265: I think atomic functions should return the value stored in `addr`. For example in https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomicadd, 
> reads the 16-bit, 32-bit or 64-bit word `old` located at the address `address` in global or shared memory, computes `(old + val)`, and stores the result back to memory at the same address. These three operations are performed in one atomic transaction. The function returns `old`.

ndickson-nvidia: Good catch!  Thanks.  I've fixed it now.

Issue #4076, opened 2022-06-02
Running error on example case: example/pytorch/ogb/ogbn-proteins


chang-l:
## 🐛 Bug

Outdated dgl dataloader API and dataloader iterator needs to be reset for epoch iteration.

## To Reproduce

1. yes | python gat.py --n-runs=2 --n-epochs=10

## Expected behavior

Traceback (most recent call last):
  File "gat.py", line 369, in <module>
    main()
  File "gat.py", line 354, in main
    val_score, test_score = run(args, graph, labels, train_idx, val_idx, test_idx, evaluator, i + 1)
  File "gat.py", line 182, in run
    DataLoader(
  File "/opt/conda/lib/python3.8/site-packages/dgl-0.9-py3.8-linux-x86_64.egg/dgl/dataloading/dataloader.py", line 855, in __init__
    super().__init__(
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 247, in __init__
    raise ValueError(
ValueError: DataLoader with IterableDataset: expected unspecified batch_sampler option, but got batch_sampler=<utils.BatchSampler object at 0x7ff9b3f33fd0>

## Environment

 - DGL Version (e.g., 1.0): 0.9
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): 1.12
 - OS (e.g., Linux): ubuntu
 - How you installed DGL (`conda`, `pip`, source): source
 - Build command you used (if compiling from source):
 - Python version: 3.8.13
 - CUDA/cuDNN version (if applicable): 11.7
 - GPU models and configuration (e.g. V100): A100
 - Any other relevant information:

## Additional context

Similar crash happens to example/pytorch/ogb/ogbn-products as well

PR #4080, opened 2022-06-02
[Bugfix] Fix example case: examples/pytorch/ogb/ogbn-proteins and examples/pytorch/ogb/ogbn-products

chang-l:
## Description
Fixed a typo, and some DGL dataloader API related issues in the example scripts.
Reported issue link: https://github.com/dmlc/dgl/issues/4076

## Changes
- [1] Make compatible to latest Dataloader API by removing batch_sampler. And remove unused BatchSampler class def.
- [2] Update the wrapper of dataloader (in utils.py) to make iterator refresh/rewind after each outside for-loop iteration (epoch-iteration). Otherwise, it will crash after first epoch iteration.
- [3] Typo-fix in ogbn-products example (format: .2s --> .2f)

## Comments/Concerns
Do we really need to have an extra layer (wrapper) for dataloader? 
If the wrapper is removed, e.g., train_dataloader = DataLoader(...)  instead of train_dataloader = DataLoaderWrapper(...), both cases can also run through, and I believe the results are consistent.


TristonC (commented):
No need to put # to comment out the undesired parts. Just delete them as all the histories are in the git repo. They can be reverted later on if needed.

chang-l:
> No need to put # to comment out the undesired parts. Just delete them as all the histories are in the git repo. They can be reverted later on if needed.

Okay. I will remove those commented regions

chang-l:
> Since the dataloader wrapper is so simple, I think it's OK to remove it.

Let me go ahead remove the wrapper and update this branch

yaox12 (approved):

PR #4108, opened 2022-06-12
[Dist] master port should be fixed for all trainers

Rhett-Ying:
## Checklist
- [ ] Code is well-documented


jermainewang:
What is the bug with the original code?

Rhett-Ying:
trainers may have different value of `master_port` then those trainers will fail to connect the master and fail to finish the init of process group in `torch.distributed`.

jermainewang:
But I don't see why does the change fix the problem. Variable `hosts` should be immutable inside the loop so why lifting `master_port` and `master_addr` outside of the loop fixes the problem?

Rhett-Ying:
if inited outside the loop, they will be same for all the trainers inside the loop. the trainers are launched inside the loop.

jermainewang:
Let's add some tests for `launch.py` to cover some basic usage.

jermainewang (approved):

PR #4114, opened 2022-06-13
[Feature] Add an optional argument to always exclude given edge IDs in as_edge_prediction_sampler

### LINKED ISSUE #4072: Dataloader sampling blocks on a different graph
## 🚀 Feature

For link prediction tasks's dataloading it would be nice to have back the possibility of passing a sampling graph on which  the sampling of blocks is performed (i.e. DGL's 0.7 `g_sample` inside `EdgeDataLoader`), separately from the graph on which edge subgraph is constructed.

## Motivation

In DGL 0.7 it was straightforward to implement a validation/test procedure via EdgeDataLoader, where one would use a val/test graph to get edge subgraphs batches, and a train graph to build the blocks structure. It would prevent information leakage issues via edges by construction (i.e. the val/test edges were guaranteed not to be present in blocks).

From DGL 0.8 and onwards, if we want to implement such test/val dataloader for link prediction tasks, we need to:
1. Have a graph that will contain train, val and test edges from which the edge subgraphs will be sampled from.
2. We can limit the edge subgraphs' edges to only val/test by passing appropriate ids to the dataloader class, via `indices` argument.
3. But, we can't easily sample blocks from a different graph that does not contain val/test edges.
To sample blocks from a different graph, we need to either write a custom sampling class (where we would override the neighbor sampling function to sample from a second graph), or to write a custom `exclude` function for the `dgl.dataloading.as_edge_prediction_sampler` (where we would exclude all ids that are relative to test/val graph, AS WELL AS all reverse ids in case of heterographs), which makes it unnecessarily complicated to solve this very common problem.

So what we want to achieve from this blocks sampling proposal is the:
1. Exclusion of edges  that were in test/val graph
2. Sampling of edges only from train graph
3. Exclusion of reverse edges, given a reverse edge mapping

## Alternatives

Right now, to ensure that there are no test/val edges inside blocks we can define one of the following:
1. A custom exclude function for `dgl.dataloading.as_edge_prediction_sampler`.
2. A custom subclass of `dgl.dataloading.NeighborSampler` that will force the sampling to be performed on a different graph.

Both of those do not seem very intuitive, as well as being error-prone, compared to how it was done before in DGL 0.7.

BarclayII:
Fixes #4072 by adding an `exclude_eids` argument in `as_edge_prediction_sampler` (and `EdgeDataLoader`).

@alexpod1000 I think this should address your concern.  Please let me know your thoughts.  Thanks!

### DIFF
--- a/python/dgl/dataloading/base.py
+++ b/python/dgl/dataloading/base.py
@@ -245,7 +245,7 @@ class BlockSampler(Sampler):
         return self.assign_lazy_features(result)
 
 
-def _find_exclude_eids_with_reverse_id(g, eids, reverse_eid_map):
+def _find_exclude_eids_with_reverse_id(eids, reverse_eid_map):
     if isinstance(eids, Mapping):
         assert_canonical_etype_keys(eids)
         exclude_eids = {
@@ -276,7 +276,7 @@ def _find_exclude_eids(g, exclude_mode, eids, **kwargs):
     elif exclude_mode == 'self':
         return eids
     elif exclude_mode == 'reverse_id':
-        return _find_exclude_eids_with_reverse_id(g, eids, kwargs['reverse_eid_map'])
+        return _find_exclude_eids_with_reverse_id(eids, kwargs['reverse_eid_map'])
     elif exclude_mode == 'reverse_types':
         return _find_exclude_eids_with_reverse_types(g, eids, kwargs['reverse_etype_map'])
     else:
--- a/python/dgl/dataloading/base.py
+++ b/python/dgl/dataloading/base.py
@@ -486,7 +486,9 @@ def as_edge_prediction_sampler(
         The edge IDs to always exclude during neighborhood sampling.
 
         Note that if :attr:`exclude` argument is ``reverse_id`` or ``reverse_types``,
-        the reverse edges of this argument will also be excluded.
+        the reverse edges of this argument will also be excluded.  If :attr:`exclude`
+        is a callable, then its argument will be the union of the seed edges in the
+        minibatch and :attr:`exclude_eids`.
     reverse_eids : Tensor or dict[etype, Tensor], optional
         A tensor of reverse edge ID mapping.  The i-th element indicates the ID of
         the i-th edge's reverse edge.
--- a/python/dgl/dataloading/base.py
+++ b/python/dgl/dataloading/base.py
@@ -280,15 +280,17 @@ def _find_exclude_eids(g, exclude_mode, eids, **kwargs):
     else:
         raise ValueError('unsupported mode {}'.format(exclude_mode))
 
-def find_exclude_eids(g, seed_edges, exclude, reverse_eids=None, reverse_etypes=None,
-                      output_device=None):
+def find_exclude_eids(g, seed_edges, exclude, exclude_eids=None, reverse_eids=None,
+                      reverse_etypes=None, output_device=None):
     """Find all edge IDs to exclude according to :attr:`exclude_mode`.
 
     Parameters
     ----------
     g : DGLGraph
         The graph.
-    exclude_mode : str, optional
+    seed_edges : Tensor or dict[etype, Tensor]
+        The edge IDs.
+    exclude :
         Can be either of the following,
 
         None (default)
@@ -320,8 +322,8 @@ def find_exclude_eids(g, seed_edges, exclude, reverse_eids=None, reverse_etypes=
         callable
             Any function that takes in a single argument :attr:`seed_edges` and returns
             a tensor or dict of tensors.
-    eids : Tensor or dict[etype, Tensor]
-        The edge IDs.
+    exclude_eids : Tensor or dict[etype, Tensor]
+        The edge IDs to always exclude from subgraphs regardless of :attr:`seed_edges`.
     reverse_eids : Tensor or dict[etype, Tensor]
         The mapping from edge ID to its reverse edge ID.
     reverse_etypes : dict[etype, etype]
@@ -329,6 +331,8 @@ def find_exclude_eids(g, seed_edges, exclude, reverse_eids=None, reverse_etypes=
     output_device : device
         The device of the output edge IDs.
     """
+    if exclude_eids is not None:
+        seed_edges = utils.union(seed_edges, exclude_eids)
     exclude_eids = _find_exclude_eids(
         g,
         exclude,
@@ -347,7 +351,7 @@ class EdgePredictionSampler(Sampler):
     --------
     as_edge_prediction_sampler
     """
-    def __init__(self, sampler, exclude=None, reverse_eids=None,
+    def __init__(self, sampler, exclude=None, exclude_eids=None, reverse_eids=None,
                  reverse_etypes=None, negative_sampler=None, prefetch_labels=None):
         super().__init__()
         # Check if the sampler's sample method has an optional third argument.
@@ -359,6 +363,7 @@ class EdgePredictionSampler(Sampler):
         self.reverse_eids = reverse_eids
         self.reverse_etypes = reverse_etypes
         self.exclude = exclude
+        self.exclude_eids = exclude_eids
         self.sampler = sampler
         self.negative_sampler = negative_sampler
         self.prefetch_labels = prefetch_labels or []
@@ -414,7 +419,7 @@ class EdgePredictionSampler(Sampler):
         seed_nodes = pair_graph.ndata[NID]
 
         exclude_eids = find_exclude_eids(
-            g, seed_edges, exclude, self.reverse_eids, self.reverse_etypes,
+            g, seed_edges, exclude, self.exclude_eids, self.reverse_eids, self.reverse_etypes,
             self.output_device)
 
         input_nodes, _, blocks = self.sampler.sample(g, seed_nodes, exclude_eids)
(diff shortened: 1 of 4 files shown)


alexpod1000:
@BarclayII 
Yes, this addresses my concern and I think it's even better than using `g_sampling` in terms of compromise between time and space overheads.
It also solves the following use-case I had encountered, where we need to simultaneously:
1. Remove reverse etypes (i.e. in hetero graphs to avoid sampling same edges for reverse relations)
2. Remove edges indicated by `exclude_eids`

Thanks a lot!

BarclayII:
@mufeili Also fixed a bunch of documentation errors.

python/dgl/dataloading/base.py
```diff
@@ -475,6 +482,11 @@ def as_edge_prediction_sampler(
 
         * User-defined exclusion rule. It is a callable with edges in the current
           minibatch as a single argument and should return the edges to be excluded.
+    exclude_eids : Tensor or dict[etype, Tensor], optional
+        The edge IDs to always exclude from during neighborhood sampling.
```
mufeili, line 486: "from" seems to be redundant

mufeili:
I'm fine with the doc changes in this PR. `NodeDataLoader` and `EdgeDataLoader` are also used in some other places like user guide, tutorials, examples and the node prediction neighbor sampling pipeline of DGL-Go. We'll also need to fix these, though not necessarily in this PR.

jermainewang:
I'm worried about the efficiency of this new strategy for excluding edges. The original `g_sampling` is fast because the sampled subgraph naturally will not contain the edges to be excluded while the new strategy needs to filter out edges from the sampled graph. As far as I remember, the underlying computation leverages [`np.isin`](https://github.com/dmlc/dgl/blob/master/python/dgl/sampling/utils.py). Considering the number of validation/test edges could be large, this may considerable slowed down sampling.

python/dgl/dataloading/base.py
```diff
@@ -475,6 +482,11 @@ def as_edge_prediction_sampler(
 
         * User-defined exclusion rule. It is a callable with edges in the current
           minibatch as a single argument and should return the edges to be excluded.
+    exclude_eids : Tensor or dict[etype, Tensor], optional
```
jermainewang, line 485: The docstring is a bit confusing. In the docstring of the `exclude` argument, it says setting it to `None` means excluding no edge, but this is not true anymore when the user provides `exclude_eids`.

BarclayII:
> I'm worried about the efficiency of this new strategy for excluding edges. The original `g_sampling` is fast because the sampled subgraph naturally will not contain the edges to be excluded while the new strategy needs to filter out edges from the sampled graph. As far as I remember, the underlying computation leverages [`np.isin`](https://github.com/dmlc/dgl/blob/master/python/dgl/sampling/utils.py). Since the number of validation/test edges could be large, this may considerably slowdown sampling.

This depends on whether the sampling operator is efficient in excluding the edge IDs.  From what I observe in the GraphSAGE link prediction example on ogbl-citation2 excluding 1M edges out of 31M, the exclusion operation is very slow on UVA but not on CPU (because `sample_neighbors` already supports excluding edge IDs):
| Mode | Without exclusion | With exclusion |
|:-:|:-:|:-:|
| UVA | 110.471 | 612.195 |
| CPU (12 workers) | 151.654 | 142.406 |

Profiling reveals that the bottleneck is in `_CAPI_DGLFilterFindIncludedIndices`.  We are not using `np.isin` in both cases.

BarclayII:
> From what I observe in the GraphSAGE link prediction example on ogbl-citation2 excluding 1M edges out of 31M, the exclusion operation is very slow on UVA but not on CPU (because `sample_neighbors` already supports excluding edge IDs):

@nv-dlasalle Is `Filter` slow in general when the number of elements to exclude is large?

nv-dlasalle:
@BarclayII I haven't had time to run this myself, but filter is implemented with a hashtable, so I expect what is happening is that we are seeing the hashtable be stored on a different device than the sampled edge IDs, and as result we have many finegrain accesses across PCIe.

Given that the CPU implementation does a binary search per edge (https://github.com/dmlc/dgl/blob/d41d07d0f6cbed17993644b58057e280a9e8f011/src/graph/sampling/neighbor/neighbor.cc#L48), it seems counter intuitive that excluding edges should make it faster than not excluding edges (151s vs 142s). Is it possible we have a bug? Or are we excluding so many edges that to_block becomes faster?

In terms of runtime and scaling, I think `g_sampling` is a better solution, as then the work to exclude edges is done only once, and we still will have `fanout` edges after sampling (rather than removing them after sampling). In terms of memory

nv-dlasalle:
> > In terms of runtime and scaling, I think `g_sampling` is a better solution, as it extend to the case where we have billions of edges better (both in the graph and to exclude).
> 
> But doesn't that requires you to have two graphs in the memory at the same time?

True, but my understanding is that `g` is only used for querying endpoints of the training set and performing negative sampling, making it largely disjoint from `g_sampling`. For scaling, I think its important that `g_sampling` allows us to use the GraphStorage-like abstraction (and thus it need not be in memory).

BarclayII:
> For scaling, I think its important that `g_sampling` allows us to use the GraphStorage-like abstraction (and thus it need not be in memory).

How would the UX look like then?  Also, if `g_sampling` is no longer an in-memory graph, then I don't see any difference to the current solution of passing in the exclusion IDs, and the problem with dynamically filtering (hence the performance issue) will still exist.

BarclayII:
> it seems counter intuitive that excluding edges should make it faster than not excluding edges (151s vs 142s). Is it possible we have a bug? Or are we excluding so many edges that to_block becomes faster?

I checked the sampled graphs and the number of edges in the first block is ~950k with exclusion versus 1M without exclusion, so not that much.

nv-dlasalle:
@BarclayII for not in-memory, I'm referring to the graph existing distributed across multiple machines/GPUs or in an external cluster (e.g., graph as a service).

That makes sense about the filter time. Thanks!

BarclayII:
> I'm referring to the graph existing distributed across multiple machines/GPUs or in an external cluster (e.g., graph as a service).

Even with multi-machines or GaaS, with `g_sampling` you will still need twice as much storage cost as the original graph, right?  The maximum size of the graph we can hold will be essentially halved given the same hardware.

jermainewang:
The problem looks like a trade-off between memory consumption and sampling speed.

1. For `g_sampling`, users need to create two graphs (1) one graph of only the training edges and (2) one graph of training + validation edges. For link prediction / edge classification, validation set is usually small compared with training set, e.g., ogbl-citation2 only has 1M (3%) validation edges. If we save both graphs, the total amount of memory for storing graph structure will almost double. This could be quite costly especially for graphs requiring distributed training. The up side is that the sampling speed is fast because the sampled mini-batch naturally does not contain any validation edges.
2. For the new `exclude_eids` proposal, it is the opposite. The memory consumption is one graph of only the training edges + a list of validation edge IDs, so the total memory usage for graph structure is halved compared with `g_sampling`. The down side is that sampling needs to exclude them from every mini-batch. From @BarclayII 's benchmark, the impact to the sampling speed is moderate for CPU but huge for UVA.

We need to fix the UVA regression otherwise we perhaps should table this for v0.9. Note that option#2 can also cover option#1 by creating a graph from the provided `exclude_eids` internally (an one-time cost), which could be a future optimization for large validation set.

@nv-dlasalle what do you think?

nv-dlasalle:
@jermainewang @BarclayII I'm fine this `exclude_eids` in 0.9, especially since it is missing functionality currently.

----

For long term, and particularly for 1.0, my two concerns are for large scale runs where `exclude_eids` is closer to a billion edges, and this is also harder for users to extend to add new functionality.

I ran the following script to quantify the cost of excluding edges:
```
import torch
import time
import dgl
# OGB must follow DGL if both DGL and PyG are installed. Otherwise DataLoader will hang.
# (This is a long-standing issue)
from ogb.linkproppred import DglLinkPropPredDataset

dataset = DglLinkPropPredDataset('ogbl-citation')
graph = dataset[0]
fanout = 20
batch_size = 1024
second_bs = batch_size*fanout

train_nid = torch.arange(graph.num_nodes())
for exc in [0, 1000000, 2000000, 4000000]:
    exclude_eids = torch.randperm(graph.num_edges())[:exc]
    dataloader = torch.utils.data.DataLoader(train_nid, batch_size=second_bs,
        shuffle=True, drop_last=False, num_workers=0)
    print("Excluding {} edges".format(exc))
    for epoch in range(3):
        dur = 0
        for it, seed_nodes in enumerate(dataloader):
            t0 = time.time()
            graph.sample_neighbors(seed_nodes, 20, exclude_edges=exclude_eids)
            t1 = time.time()
            dur += t1 - t0
        print("epoch {}: {:.03f}s".format(epoch, dur))
```
```
Excluding 0 edges
epoch 0: 1.055s
epoch 1: 0.346s
epoch 2: 0.352s
Excluding 1000000 edges
epoch 0: 4.733s
epoch 1: 4.679s
epoch 2: 4.678s
Excluding 2000000 edges
epoch 0: 7.537s
epoch 1: 7.420s
epoch 2: 7.422s
Excluding 4000000 edges
epoch 0: 12.653s
epoch 1: 12.422s
epoch 2: 12.380s

```
With the sampling and edge exclusion time isolated, the cost of excluding edges is relative to sampling is shown. CPU sampling is of complexity of the number edges in the sampled subgraph s, O(M_s), but filtering edges via binary search is of the complexity the number of edges in the sampled subgraph times the logarithm of the number of edges to be filtered (which is generally a constant fraction of the number of edges in the original graph g) O(M_s log M_g). In actuality our current implementation sorts the reverse edges + excluded edges before doing the binary search, so it is O(M_s log M_g + M_g log M_g), but we could improve this to the above complexity.

Having `g` and `exclude_eids` isn't very extendible for users, since `exclude_eids` must be a Tensor-like object. But for `g` and `g_sampling`, they are both GraphStorage-like objects, and a user could implement their own GraphStorage-like object for `g_sampling` that samples `g` and then excludes a custom set of edges, and this exclusion could then happen external to the DGL cluster (e.g., Graph as a Service).

nv-dlasalle (approved):

BarclayII:
This will be handled by the data loading pipeline refactor.

PR #4122, opened 2022-06-14
[Dist][Test] Add tests for multi-node DistTensor and DistEmbedding

Kh4L:
## Description
Add tests for `DistTensor` and `DistEmbedding` in a multi-node environment


tests/dist/python/run_dist_objects.py
```diff
@@ -0,0 +1,88 @@
+import dgl
+import torch
+import os
+import argparse
+import numpy as np
+import dgl.backend as F
+from dgl.distributed import partition_graph, load_partition, load_partition_book
+import time
+
+args_parser = argparse.ArgumentParser()
```
Rhett-Ying, line 10: Environment variables prefixed with `DIST_DGL_TEST_` are used for passing in `rpc_basics.py`. Args is used here. I think both are ok. But could we use a same/unified passing logic? some of the args are required in many unit tests such as `client/server`, `workspace`, `net_type`. The reason why I chose env for passing is mainly env is highly used within DistDGL.

tests/dist/test_dist_objects.py
```diff
@@ -0,0 +1,65 @@
+import os
```
Rhett-Ying, line 1: This script mainly for launching servers and clients so it's ok to test `DistTensor`, `DistEmbedding` , `DistOptmizer` via it, even with `DistGraph`, `DistDataLoader`. But for now, how about focusing on `DistTensor` only? It's much clearer and we could extend it in the near future.

Rhett-Ying (commented):
This draft looks good to me. Let's test `DistTensor` thoroughly first: homogeneous/heterogeneous graph, named/unnamed tensor, more clients/servers, different net_type and so on.

tests/dist/python/run_dist_objects.py
```diff
@@ -0,0 +1,77 @@
...
+net_type = os.environ.get('DIST_DGL_TEST_NET_TYPE')
+
+os.environ['DGL_DIST_MODE'] = 'distributed'
+
+def zeros_init(shape, dtype):
+    return F.zeros(shape, dtype=dtype, ctx=F.cpu())
+
+
+def run_server(graph_name, server_id, server_count, num_clients, shared_mem, keep_alive=False):
+    g = dgl.distributed.DistGraphServer(server_id, shared_workspace + "ip_config.txt",
```
Rhett-Ying, line 24: `ip_config` path could be retrieved from `DIST_DGL_TEST_IP_CONFIG` env?

tests/dist/test_dist_objects.py
```diff
@@ -0,0 +1,66 @@
...
+import unittest
+import pytest
+import multiprocessing as mp
+import subprocess
+import utils
+
+@unittest.skipIf(os.name == 'nt', reason='Do not support windows yet')
+@pytest.mark.parametrize("target", ['DistTensor'])
+@pytest.mark.parametrize("net_type", ['socket', 'tensorpipe'])
+@pytest.mark.parametrize("num_clients", [1, 4])
```
Rhett-Ying, line 11: add `@pytest.mark.parametrize("num_servers", [1, 4])` ?

tests/dist/test_dist_objects.py
```diff
@@ -0,0 +1,66 @@
...
+def test_dist_objects(target, net_type, num_clients):
+    ip_config = os.environ.get('DIST_DGL_TEST_IP_CONFIG', 'ip_config.txt')
+    workspace = os.environ.get('DIST_DGL_TEST_WORKSPACE', '/shared_workspace/dgl_dist_tensor_test/')
+
+    num_part = 2
```
Rhett-Ying, line 16: should be parsed from `ip_config.txt`.

tests/dist/test_dist_objects.py
```diff
@@ -0,0 +1,66 @@
...
+    num_servers = 1
```
Rhett-Ying, line 17: as mentioned before, make it as parameter.

tests/dist/python/create_part_graph.py
```diff
@@ -0,0 +1,18 @@
+import dgl
```
Rhett-Ying, line 1: why do we need an explicit script instead of a utility function for graph partition?

tests/dist/python/run_dist_objects.py
```diff
@@ -0,0 +1,77 @@
...
+def test_dist_tensor(g, rank):
+    dist_tensor_test_sanity(g, rank)
+    dist_tensor_test_sanity(g, rank, name="SomeDistTensor")
+
+
+
+if mode == "server":
+    run_server(graph_name, server_id=part_id, server_count=1, num_clients=num_part*num_client_per_machine, shared_mem=False, keep_alive=False)
+elif mode == "client":
+    dgl.distributed.initialize(shared_workspace + "ip_config.txt")
```
Rhett-Ying, line 56: use `DIST_DGL_TEST_IP_CONFIG` env directly

tests/dist/python/run_dist_objects.py
```diff
@@ -0,0 +1,77 @@
...
+        assert F.allclose(dist_ten[original_rank*3:(original_rank+1)*3], F.ones((3, 2), dtype=F.int32, ctx=F.cpu()) * (original_rank+1))
+
+def test_dist_tensor(g, rank):
+    dist_tensor_test_sanity(g, rank)
+    dist_tensor_test_sanity(g, rank, name="SomeDistTensor")
+
+
+
+if mode == "server":
+    run_server(graph_name, server_id=part_id, server_count=1, num_clients=num_part*num_client_per_machine, shared_mem=False, keep_alive=False)
```
Rhett-Ying, line 54: make `shared_memory` as parameter and test `True/False` ?

tests/dist/python/run_dist_objects.py
```diff
@@ -0,0 +1,77 @@
...
+elif mode == "client":
+    dgl.distributed.initialize(shared_workspace + "ip_config.txt")
+    global_rank = dgl.distributed.get_rank()
+    pid = os.getpid()
+    file_name = f"{pid}_rank_file_{global_rank}"
```
Rhett-Ying, line 59: why do we need to create these files? and these temp files should be removed when tests finish.

tests/dist/python/run_dist_objects.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def dist_tensor_test_sanity(g, rank, name=None):
+    dist_ten = dgl.distributed.DistTensor((g.number_of_nodes(), 2),
+                                           F.int32,
+                                           init_func=zeros_init,
+                                           name=name)
+    if part_id == 0:
+        dist_ten[rank*3:(rank+1)*3] = F.ones((3, 2), dtype=F.int32, ctx=F.cpu()) * (rank+1)
+        dgl.distributed.client_barrier()
```
Rhett-Ying, line 41: do we really need this barrier? it's called automatically when client process exits.

tests/dist/python/run_dist_objects.py
```diff
@@ -0,0 +1,77 @@
...
+    else:
+        dgl.distributed.client_barrier()
+        original_rank = rank // 2
+        assert F.allclose(dist_ten[original_rank*3:(original_rank+1)*3], F.ones((3, 2), dtype=F.int32, ctx=F.cpu()) * (original_rank+1))
+
+def test_dist_tensor(g, rank):
```
Rhett-Ying, line 47: more tests are required. could we import all tests from existing `tests/distributed/test_dist_graph_store.py`?

tests/dist/python/run_dist_objects.py
```diff
@@ -0,0 +1,77 @@
...
+    dist_ten = dgl.distributed.DistTensor((g.number_of_nodes(), 2),
+                                           F.int32,
+                                           init_func=zeros_init,
+                                           name=name)
+    if part_id == 0:
+        dist_ten[rank*3:(rank+1)*3] = F.ones((3, 2), dtype=F.int32, ctx=F.cpu()) * (rank+1)
+        dgl.distributed.client_barrier()
+    else:
+        dgl.distributed.client_barrier()
+        original_rank = rank // 2
```
Rhett-Ying, line 44: why do we need this? How about calling `g.rank()` in current process?

tests/dist/python/run_dist_objects.py
```diff
@@ -0,0 +1,77 @@
...
+    g.start()
+
+
+def dist_tensor_test_sanity(g, rank, name=None):
+    dist_ten = dgl.distributed.DistTensor((g.number_of_nodes(), 2),
+                                           F.int32,
+                                           init_func=zeros_init,
+                                           name=name)
+    if part_id == 0:
+        dist_ten[rank*3:(rank+1)*3] = F.ones((3, 2), dtype=F.int32, ctx=F.cpu()) * (rank+1)
```
Rhett-Ying, line 40: what is `3` for?

Rhett-Ying (changes requested):

tests/dist/python/run_dist_objects.py
Rhett-Ying: could you cover all tests in https://github.com/dmlc/dgl/blob/0f0e7c7fad6ec1399a67881ce9a7e60d14cfe1a0/tests/distributed/test_dist_graph_store.py#L205 ?

tests/dist/python/run_dist_objects.py
Kh4L: We need it as there are different tests and might be more changes in the tests

Rhett-Ying (approved):

PR #4133, opened 2022-06-16
[Feature] Graph level attributes (gdata)

krzysztof-daniell:
## Description
Graph level attributes that can be accessed by `gdata`. Can be useful for all molecule prediction tasks to store target in graph object instead of storing it separately. It will make easier for users to transfer from PyTorch Geometric to DGL if needed, eg. for [Open Catalyst](https://opencatalystproject.org/) dataset. It works the same for homogeneous and heterogenous graphs.

## Checklist
- [x] Code is well-documented

## Changes
- Added graph frame to `DGLHeteroGraph`
- Added `HeteroGraphDataView`
- Added `graph_attr_schemes` method and updated `DGLHeteroGraph` with it
- Added `set/get/pop_g_repr` methods

### DIFF
--- a/python/dgl/heterograph.py
+++ b/python/dgl/heterograph.py
@@ -18,7 +18,8 @@ from . import heterograph_index
 from . import utils
 from . import backend as F
 from .frame import Frame
-from .view import HeteroGraphDataView, HeteroNodeView, HeteroNodeDataView, HeteroEdgeView, HeteroEdgeDataView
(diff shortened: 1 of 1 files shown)


BarclayII:
Could we chat on this PR over slack?  I sent you a message there.

Overall, I wasn't quite sure on why storing the graph-level data into the DGLGraph object is necessary.  I was also curious on whether setting the attributes of DGLGraphs directly works or not.

Issue #4134, opened 2022-06-17
[Bug] Calling `IsPinned()` in a worker process results in RuntimeError: CUDA error: unspecified launch failure


nv-dlasalle:
## 🐛 Bug

This is caused by `IsPinned()` relying on invoking the cuda context, which is not valid in forked processes, and is the reason #3997 was generating errors.

This is tricky to address because both of the following scenarios are valid:

1. The main process initializes a cuda context, but the worker processes in the dataloader never interact with the cuda context.
1. The main process does not initialize a cuda context before spawning workers, but instead the worker processes each initialize a cuda context.

## To Reproduce

1. Initialize a cuda context (e.g., call .cuda() on a model).
1. Use a dataloader with number of workers > 0.
1. Cause `IsPinned()` to be called from one of the workers.

```
dominique@dominique-X299-A:~/src/dgl/examples/pytorch/graphsage$ python3 node_classification.py
num_tasks = 1
task_type = multiclass classification
eval_metric = acc
num_classes = 47
is_hetero = False
Traceback (most recent call last):
  File "node_classification.py", line 96, in <module>
    for it, (input_nodes, output_nodes, blocks) in enumerate(train_dataloader):
  File "/home/dominique/.local/lib/python3.6/site-packages/dgl-0.9-py3.6-linux-x86_64.egg/dgl/dataloading/dataloader.py", line 501, in __next__
    self._next_non_threaded() if not self.use_thread else self._next_threaded()
  File "/home/dominique/.local/lib/python3.6/site-packages/dgl-0.9-py3.6-linux-x86_64.egg/dgl/dataloading/dataloader.py", line 496, in _next_threaded
    exception.reraise()
  File "/home/dominique/.local/lib/python3.6/site-packages/dgl-0.9-py3.6-linux-x86_64.egg/dgl/utils/exception.py", line 57, in reraise
    raise exception
RuntimeError: Caught RuntimeError in prefetcher.
Original Traceback (most recent call last):
  File "/home/dominique/.local/lib/python3.6/site-packages/dgl-0.9-py3.6-linux-x86_64.egg/dgl/dataloading/dataloader.py", line 355, in _prefetcher_entry
    feats = _prefetch(batch, dataloader, stream)
  File "/home/dominique/.local/lib/python3.6/site-packages/dgl-0.9-py3.6-linux-x86_64.egg/dgl/dataloading/dataloader.py", line 306, in _prefetch
[...]

Issue #4107, opened 2022-06-12
AttributeError: 'Column' object has no attribute `deferred_dtype`


mufeili:
## 🐛 Bug

@ndickson-nvidia @BarclayII 

The introduction of `deferred_dtype` in PR #3890 seems to be causing issues when loading a graph file saved with an older version of DGL. Do you have any ideas on fixing the compatibility? We might also want to add a unit test for loading a graph data file saved with an older version of DGL.

![WechatIMG8012](https://user-images.githubusercontent.com/19576924/173212115-6cfe418c-6718-4cce-adbc-d06e193bb1c0.jpeg)

## Environment

 - DGL Version (e.g., 1.0): 0.8.2
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):
 - OS (e.g., Linux):
 - How you installed DGL (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version (if applicable):
 - GPU models and configuration (e.g. V100):
 - Any other relevant information:

ndickson-nvidia:
Are there any plans to add a more robust system for saving/loading graph data?  Relying on just serializing Python objects, like what `dgl.data.utils.save_info` and `dgl.data.utils.load_info` do, seems like a risky approach that may not be able to support any sort of compatibility, especially since there's no version information.  I'd expect that a few data members of `Column`, for example, never need to be saved or loaded, since `device`, `index`, and `deferred_dtype` should be resolved and cleared before saving to disk, and there are probably other similar examples in other classes.

Wxz666666:
博主，您这个问题怎么解决的

Wxz666666:
Blogger, how did you solve this problem

mufeili:
> Blogger, how did you solve this problem

For now, downgrading to `0.8.1` should address the issue.

natalie-0430:
Hello, how can you downgrade to 0.8.1 from 0.8.2? I have a similar problem.

nv-dlasalle:
@natalie-0430 That will depend on how you installed it. If you used pip for example:
```
pip uninstall dgl-cu113
pip install dgl-cu113==0.8.1 dglgo -f https://data.dgl.ai/wheels/repo.html
```

[...]

PR #4135, opened 2022-06-17
[Bugfix][Rework] Automatically unpin tensors pinned by DGL (rework #3997)

yaox12:
## Description

Introduce a new flag `pinned_by_dgl` to `NDArray::Container` to avoid `IsPinned()` being called by most common tensors, which should fix the CUDA context issue in forked processes, e.g., dataloaders, in #4061 #4134. 

With this PR, the pinning status is held by a `dgl.ndarray` handle. Once the handle is deleted/freed, the corresponding tensor will be automatically unpinned. So users are required to store this handle until the pinned tensors are no longer needed.

```python
tensor = ...  # a PyTorch tensor

nd = pin_memory_inplace(tensor)  # nd is the handle that holds the pinning status
assert tensor.is_pinned()

del nd  # once nd is deleted/freed, the corresponding tensor will be automatically unpinned.
        # You can also manually unpin it by nd.unpin_memory_()
assert not tensor.is_pinned()

pin_memory_inplace(tensor)  # nd is not saved, so the tensor will be immediately unpinned after being pinned.
assert not tensor.is_pinned()
```

## Checklist
- [x] Code is well-documented

## Changes

- Add a `pinned_by_dgl` flag to `NDArray::Container`.
- Change pinning/unpinning functions to accept `NDArray::Container` instead of `DLTensor*`.


nv-dlasalle (approved):
Looks good to me.

One thing is we may want to add a unit test for #4134 that uses cuda and a dataloader that has multiple workers.

include/dgl/runtime/ndarray.h
```diff
@@ -415,6 +415,9 @@ struct NDArray::Container {
   std::vector<int64_t> stride_;
   /*! \brief The internal array object */
   std::atomic<int> ref_counter_{0};
+
+  bool pinned_by_dgl_{false};
```
jermainewang, line 419: suggest to just call it `is_pinned_`

include/dgl/runtime/ndarray.h
```diff
@@ -415,6 +415,9 @@ struct NDArray::Container {
...
+  bool from_tensor_dispatcher_{false};
```
jermainewang, line 420: Why need this flag?

src/runtime/ndarray.cc
```diff
@@ -78,6 +77,10 @@ struct NDArray::Internal {
   // This enables us to create NDArray from memory allocated by other
   // frameworks that are DLPack compatible
   static void DLPackDeleter(NDArray::Container* ptr) {
+    // if the array is pinned by dgl, unpin it before freeing
+    if (ptr->from_tensor_dispatcher_ && ptr->pinned_by_dgl_)
```
jermainewang, line 81: `if (... && ptr->IsContainerPinned())`

src/runtime/ndarray.cc
```diff
@@ -327,7 +336,10 @@ std::shared_ptr<SharedMemory> NDArray::GetSharedMem() const {
   return this->data_->mem;
 }
 
-bool NDArray::IsDataPinned(DLTensor* tensor) {
+bool NDArray::IsContainerPinned(NDArray::Container* ptr) {
+  if (ptr->pinned_by_dgl_)
+    return true;
+  auto* tensor = &(ptr->dl_tensor);
```
jermainewang, line 342: Since we have added a flag, when do we need to check the pointer again?

jermainewang:
Another question, in
https://github.com/dmlc/dgl/blob/94806f328382cd0ee00bf205216ea75281626b5d/src/runtime/ndarray.cc#L424-L443

The `MoveAsDLTensor` assigns a `DLTensor` pointer instead of an `NDArray::Container` pointer to the `DGLArrayHandle` variable.    Shall we change that too?

include/dgl/runtime/ndarray.h
yaox12: If we borrowed a tensor pinned by PyTorch, we don't want to unpin it automatically.

src/runtime/ndarray.cc
yaox12: If we borrowed a tensor pinned by PyTorch, we don't want to unpin it automatically.

src/runtime/ndarray.cc
yaox12: If we borrowed a tensor pinned by PyTorch, we don't want to unpin it automatically. But we still need to know whether it's pinned.

include/dgl/runtime/ndarray.h
yaox12: This is to avoid unpinning temporary NDArrays in such a case:
```python
def pin_memory_inplace(tensor):
    F.zerocopy_to_dgl_ndarray_for_write(tensor).pin_memory_()
```

1. However, with it, PyTorch tensors will not be automatically unpinned, which means
```python
tensor = torch.ones(10, 10)
nd = F.zerocopy_to_dgl_ndarray_for_write(t)
nd.pin_memory_()

del tensor
del nd  # will not be unpinned
```
we have to use
```python
nd = ndarray.empty(10, 10)
tensor = F.zerocopy_from_dgl_ndarray(nd)
nd.pin_memory_()

del nd  # will be unpinned
del tensor
```

2. If I remove it, we have to change `pin_memory_inplace()` to
```python
def pin_memory_inplace(tensor):
    nd = F.zerocopy_to_dgl_ndarray_for_write(tensor)
    nd.pin_memory_()
    return nd
```
and require people to save `nd` in somewhere.

I prefer the latter now. How do you think about it? @BarclayII

yaox12:
I think the `MoveAsDLTensor` is fine since we can always reinterpret_cast `DGLArrayHandle` back to an `NDArray::Container` pointer.

include/dgl/runtime/ndarray.h
BarclayII: Yeah it seems that 2 is better since users are unlikely to call `pin_memory_inplace` by themselves.

yaox12: Updated.

src/runtime/ndarray.cc
```diff
@@ -256,16 +257,26 @@ void NDArray::CopyFromTo(DLTensor* from,
...
 
-void NDArray::UnpinData(DLTensor* tensor) {
-  if (!IsDataPinned(tensor)) return;
-  DeviceAPI::Get(kDLGPU)->UnpinData(tensor->data);
+void NDArray::UnpinContainer(NDArray::Container* ptr) {
+  auto container_is_pinned = IsContainerPinned(ptr);
+  // The tensor may be pinned outside of DGL via a different CUDA API,
+  // so we cannot unpin it with cudaHostUnregister.
+  CHECK(ptr->pinned_by_dgl_ || !container_is_pinned)
+    << "Cannot unpin a tensor that is pinned outside of DGL.";
```
yaox12, line 274: Just realized this case and I'd like to give it a clearer error message here. There is also a unit test for it:
https://github.com/dmlc/dgl/blob/d53adc16f0ac43cfd503cddb4bb1c7bff14dbb38/tests/compute/test_pin_memory.py#L22-L25.

However, this would break the previous `unpin_memory_inplace`:
```python
F.zerocopy_to_dgl_ndarray_for_write(tensor).unpin_memory_()
```
because the temporary ndarray created here doesn't have the `pinned_by_dgl_` flag. 

I suggest remove `unpin_memory_inplace()` and unpin the array via `nd.unpin_memory_()` where `nd` is returned from `pin_memory_inplace()`.

include/dgl/runtime/ndarray.h
BarclayII: Could you fix the DataLoader's use_uva option to store the NDArray handle as well?

examples/pytorch/graphsage/multi_gpu_node_classification.py
```diff
@@ -69,13 +69,14 @@ def inference(self, g, device, batch_size):
                 batch_size=batch_size, shuffle=False, drop_last=False,
                 num_workers=0, use_ddp=True, use_uva=True)
 
+        last_nd = None
```
jermainewang, line 72: @BarclayII could we hide this in the `dataloader` object above so there is no explicit `nd` in user's code?

BarclayII: It's hard because `y` is outside the DataLoader (L79 and L89).

yaox12: This is really an advanced usage of pinned tensors in dataloader. It updates the feature and reuses a pre-defined dataloader for UVA feature fetching. 
In common cases, things will be much easier because we seldom modify the pinned tensors in the dataloader.

include/dgl/runtime/ndarray.h
BarclayII: I see.  Makes sense.  Note that adding new attributes in Columns without will cause #4107 however unless we override `__getstate__` and `__setstate__`.

BarclayII: I see.  Makes sense.  Note that adding new attributes in Columns without will cause #4107 however unless we override `__getstate__` and `__setstate__`.

examples/pytorch/graphsage/multi_gpu_node_classification.py
BarclayII: Could we re-initialize the DataLoader for every layer?

python/dgl/frame.py
```diff
@@ -444,7 +444,7 @@ def pin_memory_(self):
         Does nothing if the storage is already pinned.
         """
         if not self.pinned_by_dgl and not F.is_pinned(self.data):
-            pin_memory_inplace(self.data)
+            self.data_nd = pin_memory_inplace(self.data)
```
BarclayII, line 447: What if I do
```
g.ndata['x'] = torch.randn(...)
g.pin_memory_()
g.ndata['x'] = torch.randn(...)
```
`g.ndata['x']` will still be marked as `pinned_by_dgl = True`, but the actual content will not be pinned?

yaox12: This should be handled by the setter:
https://github.com/dmlc/dgl/blob/39721cedafa3caf69eb9e8b6bad9cc84a7dacf7f/python/dgl/frame.py#L241-L246
But you remind me I should delete `self.data_nd` in it.

examples/pytorch/graphsage/multi_gpu_node_classification.py
yaox12: Hided `nd` by re-initializing the dataloader for every layer.

python/dgl/frame.py
yaox12: If I understand correctly, the snippets you gave will create a new Column to replace the old one instead of updating it.

yaox12: Added a unit test for it.

jermainewang (approved):

PR #4137, opened 2022-06-17
[Feature] Limited Edge Dataloader (Part of auto inference).

yinpeiqi:
## Description

This is the limited edge dataloader. (For automatic inference helper).
For the normal node dataloader, the memory consumption of each batch is unbalance. We only fixed the number of target nodes in a batch, while the number of edges are different among different batches. Since we may encounter the cases that many high degree nodes in the same batch, and it become the bottleneck of training/inference. We develop a limited edge dataloader, which user first define two configures: `max_node` and `max_edge`. These two parameters define the maximum number of nodes and the maximum number of edges in the same batch. This dataloader mainly consider the workload of GNN inference.

The limited edge dataloader also support modify `max_node` and `max_edge` during training/inference. (The support for automatic inference).

Here we display an example of using the LimitedEdgeDataloader:
```python
import torch
import dgl
from dgl.dataloading.limited_edge import LimitedEdgeDataloader

if __name__ == '__main__':
    data = RedditDataset(self_loop=True)
    g = data[0].to("cuda")
    nids = torch.arange(g.number_of_nodes()).to(g.device)
    sampler = dgl.dataloading.MultiLayerFullNeighborSampler(1)
    dataloader = LimitedEdgeDataloader(
        g, nids, sampler, max_node=5000, max_edge=500000,
        shuffle=False, drop_last=False, device="cuda", num_workers=0)
    for input_nodes, output_nodes, blocks in dataloader:
        print(blocks)
```
Here I limited maximum number of nodes=5000, maximum number of edges=500000, and part of the result likes:
```
[Block(num_src_nodes=24686, num_dst_nodes=5000, num_edges=85203)]
[Block(num_src_nodes=29197, num_dst_nodes=5000, num_edges=329521)]
[Block(num_src_nodes=29056, num_dst_nodes=4775, num_edges=499702)]
[Block(num_src_nodes=43231, num_dst_nodes=5000, num_edges=477514)]
[Block(num_src_nodes=49092, num_dst_nodes=4407, num_edges=499802)]
[Block(num_src_nodes=63065, num_dst_nodes=5000, num_edges=473265)]
[Block(num_src_nodes=37184, num_dst_nodes=2961, num_edges=499686)]
[Block(num_src_nodes=72631, num_dst_nodes=4771, num_edges=498843)]
[Block(num_src_nodes=48633, num_dst_nodes=2741, num_edges=499758)]
[Block(num_src_nodes=68416, num_dst_nodes=3529, num_edges=499993)]
[Block(num_src_nodes=80186, num_dst_nodes=4128, num_edges=499989)]
[Block(num_src_nodes=78550, num_dst_nodes=3458, num_edges=499995)]
[Block(num_src_nodes=87168, num_dst_nodes=3383, num_edges=499872)]
...
```

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/dataloading/limited_edge.py
+++ b/python/dgl/dataloading/limited_edge.py
@@ -1,6 +1,7 @@
 from typing import Generic
 import functools
 
+import numpy as np
 import torch
 import dgl
 from dgl.dataloading.dataloader import _TensorizedDatasetIter
@@ -60,19 +61,18 @@ class LimitedEdgeDataset(dgl.dataloading.TensorizedDataset):
         self.prefix_sum_in_degrees = prefix_sum_in_degrees
         # Compute the prefix sum in degrees for the graph.
         if self.prefix_sum_in_degrees is None:
-            in_degrees = g.in_degrees(train_nids.to(g.device))
+            in_degrees = g.in_degrees(train_nids.to(g.device)).cpu()
+            prefix_sum_in_degrees = np.cumsum(in_degrees)
             self.prefix_sum_in_degrees = [0]
-            self.prefix_sum_in_degrees.extend(in_degrees.tolist())
-            for i in range(1, len(self.in_degrees)):
-                self.prefix_sum_in_degrees[i] += self.prefix_sum_in_degrees[i - 1]
+            self.prefix_sum_in_degrees.extend(prefix_sum_in_degrees.tolist())
             self.prefix_sum_in_degrees.append(2e18)
 
-        self.curr_iter = CustomDatasetIter(
+        self.curr_iter = LimitedEdgeDatasetIter(
             id_tensor, self.max_node, self.max_edge, self.prefix_sum_in_degrees, self.drop_last, self._mapping_keys)
 
     def __getattr__(self, attribute_name):
-        if attribute_name in CustomDataset.functions:
-            function = functools.partial(CustomDataset.functions[attribute_name], self)
+        if attribute_name in LimitedEdgeDataset.functions:
+            function = functools.partial(LimitedEdgeDataset.functions[attribute_name], self)
             return function
         else:
             return super(Generic, self).__getattr__(attribute_name)
@@ -96,13 +96,13 @@ class LimitedEdgeDatasetIter(_TensorizedDatasetIter):
         if self.prefix_sum_in_degrees[binary_end] - self.prefix_sum_in_degrees[self.index] < self.max_edge:
             return binary_end
         binary_middle = 0
-        while binary_end - binary_start > 5:
+        while binary_end - binary_start > 1:
             binary_middle = (binary_start + binary_end) // 2
             if self.prefix_sum_in_degrees[binary_middle] - self.prefix_sum_in_degrees[self.index] < self.max_edge:
                 binary_start = binary_middle
             else:
                 binary_end = binary_middle - 1
-        return binary_middle
+        return binary_start
 
     def _next_indices(self):
         if self.index >= self.num_item:
--- a/python/dgl/dataloading/limited_edge.py
+++ b/python/dgl/dataloading/limited_edge.py
@@ -1,0 +1,113 @@
+from typing import Generic
+import functools
+
+import torch
+import dgl
+from dgl.dataloading.dataloader import _TensorizedDatasetIter
+
+
+class LimitedEdgeDataloader(dgl.dataloading.NodeDataLoader):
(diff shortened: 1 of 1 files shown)


jermainewang:
Please add some unit tests and fix the lint errors.

python/dgl/dataloading/limited_edge.py
```diff
@@ -0,0 +1,125 @@
...
+                         dataset,
+                         sampler,
+                         device=device,
+                         use_uva=use_uva,
+                         shuffle=shuffle,
+                         drop_last=drop_last,
+                         use_prefetch_thread=False,
+                         num_workers=num_workers)
+
+    def modify_max_edge(self, max_edge):
```
jermainewang, line 35: Is this function used anywhere? Same for `modify_max_node` and `reset_batch_node`.

python/dgl/dataloading/limited_edge.py
```diff
@@ -0,0 +1,125 @@
...
+        self.dataset.max_node = max_node
+        if self.dataset.curr_iter is not None:
+            self.dataset.curr_iter.max_node = max_node
+
+    def reset_batch_node(self, node_count):
+        """Reset batch node."""
+        if self.dataset.curr_iter is not None:
+            self.dataset.curr_iter.index -= node_count
+
+    def __setattr__(self, __name, __value):
```
jermainewang, line 52: Why need to override `__setattr__`?

jermainewang (changes requested):

jermainewang:
I don't fully understand the internal `Dataset` and `Iter` part. @BarclayII could you take a look?

python/dgl/dataloading/limited_edge.py
yinpeiqi: These three functions will be used in the auto tuner (modify the `max_node` and `max_edge` after each batch according to the memory usage).

python/dgl/dataloading/limited_edge.py
```diff
@@ -0,0 +1,113 @@
...
+class LimitedEdgeDataloader(dgl.dataloading.NodeDataLoader):
+    """Limited Edge Dataloader class.
+
+    For the normal dataloader, each batch have a same batch size, which may cause the workload unbanlance
+    problem when some batches contains high degree nodes. The limited edge dataloader can support different
+    batch sizes among different batches. Users can parse two configures: max_node and max_edge, which means
+    the maximum number of nodes in a batch and the maximum number of edges in a batch. The limited edge 
+    dataloader can provide the largest batch which follows the rule.
+    """
+    def __init__(self, g, nids, sampler, max_node, max_edge, prefix_sum_in_degrees=None, \
```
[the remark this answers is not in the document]
jermainewang: 1. You could make `max_edge` (still think `max_degree` is better) a mandatory argument while making `max_node` an optional one. You could give it a default value (e.g., 0) meaning that the dataloader will not control minibatches by num nodes.
2. I see. Then it's ok to leave it here. Just need to update the doc.
3. I see.

python/dgl/dataloading/limited_edge.py
jermainewang: Is autotuner included in this PR?

python/dgl/dataloading/limited_edge.py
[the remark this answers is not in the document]
yinpeiqi: 1. I think set `max_node` as `None` to represent is a possible solution. But in my opinion, `max_edge` is to control the maximum number of edges inside a Graph (also known as block), It is strange to describe this item as "degree"
, since we do not say the degree of a block.
2. Okay, I will add it again. I think we can not allow `prefix_sum_in_degrees` if the `shuffle` is True.

python/dgl/dataloading/limited_edge.py
yinpeiqi: No, here only expose the functions that the autotuner use. I will try to add it in this PR.

python/dgl/dataloading/limited_edge.py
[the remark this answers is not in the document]
jermainewang: 1. It depends on the sampler user provides. For full neighbor sampler, you are right. But users may provide subgraph sampler, etc.
2. Then remember to have a validity check.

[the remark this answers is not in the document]
yinpeiqi: Okay I changed it to `max_degree`.

Issue #3988, opened 2022-05-09
Add FP16 support for GatherMM kernel


pedrocolon93:
## 🐛 Bug

I am trying to compile the library with FP16 support as per the documentation in 
https://docs.dgl.ai/en/0.6.x/guide/mixed_precision.html
However, when cuda starts to compile, I get missing include issues.

## To Reproduce

git clone --recurse-submodules https://github.com/dmlc/dgl.git
cd dgl
mkdir build
cd build
cmake -DUSE_CUDA=ON -DUSE_FP16=ON ..
make -j

That by itself gives an error of an issue in an if at the beginning to detect if its above cuda11 on file:
make/modules/CUDA.cmake
I circunnavegated that by removing the if (I am sure that my gpus suport that arch), but that needs to be fixed

Then later on in compilation when doing make -j
I get a missing include of dgl/array.h etc.. 

I fixed this by adding in:

```
macro(dgl_config_cuda out_variable)
  if(NOT CUDA_FOUND)
    message(FATAL_ERROR "Cannot find CUDA.")
  endif()
  \# always set the includedir when cuda is available
  \# avoid global retrigger of cmake
        include_directories(${CUDA_INCLUDE_DIRS})

        **include_directories("include")
        include_directories("third_party/dlpack/include")
        include_directories("third_party/dmlc-core/include")
        include_directories("third_party/phmap/")
        include_directories("third_party/xbyak/")
        include_directories("third_party/METIS/include/")
        include_directories("tensoradapter/include")
        include_directories("third_party/nanoflann/include")
        include_directories("third_party/libxsmm/include")**

```
## Expected behavior

All in all, with the compile using fp16, it should just work.. 

## Environment

 - DGL Version 0.8:
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): Pytorch 1.11
 - OS (e.g., Linux): Ubuntu 16.04
 - How you installed DGL (`conda`, `pip`, source): source
 - Build command you used (if compiling from source): cmake -DUSE_CUDA=ON -DUSE_FP16=ON .. && make -j
 - Python version: 3.9
 - CUDA/cuDNN version (if applicable): 11.3
 - GPU models and configuration (e.g. V100): A6000
 - Any other relevant information:

I think something happened in the cmake file that the include directories are never passed on correctly into the cuda compilation.

yzh119:
Could you please elaborate on the first issue you met?
[...]

PR #4140, opened 2022-06-17
[Bug][Feature] Added more missing FP16 specializations

ndickson-nvidia:
## Description
* Continuation of adding more missing FP16 specializations after PR #4029 , which addressed the specific case from issue #3988
* Added missing specializations for `__half` of `DLDataTypeTraits`, `IndexSelect`, `Full`, `Scatter_`, `CSRGetData`, `CSRMM`, `CSRSum`
* Fixed casting issue in `_LinearSearchKernel` that was preventing it from supporting `__half`
* Added more specific error messages for unimplemented FP16 specializations of `Xgeam`, `CSRGEMM`, and `CSRGEAM`, which would require functions that aren't provided by cublas

## Checklist
- [x] Code is well-documented

### DIFF
--- a/src/array/cuda/array_index_select.cu
+++ b/src/array/cuda/array_index_select.cu
@@ -71,7 +71,8 @@ DType IndexSelect(NDArray array, int64_t index) {
   // The initialization constructor for __half is apparently a device-
   // only function in some setups, but the current function isn't run
   // on the device.
-  using SafeDType = typename std::conditional<std::is_same<DType,__half>::value, uint16_t, DType>::type;
+  using SafeDType = typename std::conditional<
+      std::is_same<DType, __half>::value, uint16_t, DType>::type;
   SafeDType ret = 0;
 #else
   DType ret = 0;
--- a/src/array/cuda/array_index_select.cu
+++ b/src/array/cuda/array_index_select.cu
@@ -67,12 +67,20 @@ template NDArray IndexSelect<kDLGPU, double, int64_t>(NDArray, IdArray);
 template <DLDeviceType XPU, typename DType>
 DType IndexSelect(NDArray array, int64_t index) {
   auto device = runtime::DeviceAPI::Get(array->ctx);
+#ifdef USE_FP16
+  // The initialization constructor for __half is apparently a device-
+  // only function in some setups, but the current function isn't run
+  // on the device.
+  using SafeDType = typename std::conditional<std::is_same<DType,__half>::value, uint16_t, DType>::type;
+  SafeDType ret = 0;
+#else
   DType ret = 0;
+#endif
   device->CopyDataFromTo(
-      static_cast<DType*>(array->data) + index, 0, &ret, 0,
+      static_cast<DType*>(array->data) + index, 0, reinterpret_cast<DType*>(&ret), 0,
       sizeof(DType), array->ctx, DLContext{kDLCPU, 0},
       array->dtype, nullptr);
-  return ret;
+  return reinterpret_cast<DType&>(ret);
 }
 
 template int32_t IndexSelect<kDLGPU, int32_t>(NDArray array, int64_t index);
--- a/src/array/cuda/array_index_select.cu
+++ b/src/array/cuda/array_index_select.cu
@@ -69,8 +69,8 @@ DType IndexSelect(NDArray array, int64_t index) {
   auto device = runtime::DeviceAPI::Get(array->ctx);
 #ifdef USE_FP16
   // The initialization constructor for __half is apparently a device-
-  // only function in some setups, but the current function isn't run
-  // on the device.
+  // only function in some setups, but the current function, IndexSelect,
+  // isn't run on the device, so it doesn't have access to that constructor.
   using SafeDType = typename std::conditional<
       std::is_same<DType, __half>::value, uint16_t, DType>::type;
   SafeDType ret = 0;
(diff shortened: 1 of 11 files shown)


src/array/cuda/array_index_select.cu
```diff
@@ -63,18 +67,30 @@ template NDArray IndexSelect<kDLGPU, double, int64_t>(NDArray, IdArray);
 template <DLDeviceType XPU, typename DType>
 DType IndexSelect(NDArray array, int64_t index) {
   auto device = runtime::DeviceAPI::Get(array->ctx);
+#ifdef USE_FP16
+  // The initialization constructor for __half is apparently a device-
+  // only function in some setups, but the current function isn't run
+  // on the device.
```
isratnisa, line 73: I am not clear which 'current' function we are talking here.

ndickson-nvidia: "the current function" here refers to the function containing this comment, the function that is currently being run (on the host) as this comment is passed.  If I referred to it by name instead, i.e. `IndexSelect`, it might sound like the comment is referring to the other function named `IndexSelect`, above.  Maybe it would be clear if I included both, though. 🤔  I'll try something when I add the "TODO"s.

src/array/cuda/spmm.cuh
```diff
@@ -32,6 +32,20 @@ cublasStatus_t Xgeam(cublasHandle_t handle, cublasOperation_t transa,
...
 }
 
+#ifdef USE_FP16
+template <>
+cublasStatus_t Xgeam<__half>(cublasHandle_t handle, cublasOperation_t transa,
+    cublasOperation_t transb, int m, int n,
+    const __half* alpha, const __half* A, int lda,
+    const __half* beta, const __half* B, int ldb,
+    __half* C, int ldc) {
+  // There is no cublasHgeam, so a different implementation
```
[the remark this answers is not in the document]
ndickson-nvidia: Added.  Thanks!

src/array/cuda/cusparse_dispatcher.cuh
```diff
@@ -34,6 +34,32 @@ struct CSRGEMM {
   }
 };
 
+#ifdef USE_FP16
+template <>
+struct CSRGEMM<__half> {
+  template <typename... Args>
+  static inline cusparseStatus_t bufferSizeExt(Args... args) {
+    // There is no cusparseHcsrgemm2_bufferSizeExt, so a different implementation
+    // would be required.
```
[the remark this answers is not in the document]
ndickson-nvidia: Added as above.  Thanks!

src/array/cuda/array_index_select.cu
ndickson-nvidia: I updated the comment.  Hopefully it's a bit clearer now.  Thanks!

src/array/cuda/cusparse_dispatcher.cuh
```diff
@@ -34,6 +34,32 @@ struct CSRGEMM {
...
 };
 
+#ifdef USE_FP16
+template <>
+struct CSRGEMM<__half> {
+  template <typename... Args>
+  static inline cusparseStatus_t bufferSizeExt(Args... args) {
+    // TODO(ndickson): There is no cusparseHcsrgemm2_bufferSizeExt, so a different
+    // implementation would be required.
+    BUG_IF_FAIL(false) << "CSRGEMM::bufferSizeExt does not support dtype half (FP16).";
```
nv-dlasalle, line 44: While functionally it doesn't matter, this will instruct users to file a bug report. Since this is expected behavior, we should use `LOG(FATAL)` instead:
```
LOG(FATAL) << "CSRGEMM::bufferSizeExt does not support dtype half (FP16).";
```

src/array/cuda/cusparse_dispatcher.cuh
```diff
@@ -91,6 +117,32 @@ struct CSRGEAM {
...
 };
 
+#ifdef USE_FP16
+template <>
+struct CSRGEAM<__half> {
+  template <typename... Args>
+  static inline cusparseStatus_t bufferSizeExt(Args... args) {
+    // TODO(ndickson): There is no cusparseHcsrgeam2_bufferSizeExt, so a different
+    // implementation would be required.
+    BUG_IF_FAIL(false) << "CSRGEAM::bufferSizeExt does not support dtype half (FP16).";
```
nv-dlasalle, line 127: Same here, use `LOG(FATAL)` instead.

src/array/cuda/spmm.cuh
```diff
@@ -32,6 +32,20 @@ cublasStatus_t Xgeam(cublasHandle_t handle, cublasOperation_t transa,
...
+template <>
+cublasStatus_t Xgeam<__half>(cublasHandle_t handle, cublasOperation_t transa,
+    cublasOperation_t transb, int m, int n,
+    const __half* alpha, const __half* A, int lda,
+    const __half* beta, const __half* B, int ldb,
+    __half* C, int ldc) {
+  // TODO(ndickson): There is no cublasHgeam, so a different
+  // implementation would be required.
+  LOG(INFO) << "Xgeam does not support dtype half (FP16)";
+  return CUBLAS_STATUS_EXECUTION_FAILED;
```
nv-dlasalle, line 45: We should do `LOG(FATAL)` here so the return is not reached.

As is, users may ignore the "INFO" message, and think an error occured inside of cublas (which was not even called).

nv-dlasalle (approved):
Functionally it looks good--just the way some of the errors are reported needs updating.

PR #4141, opened 2022-06-18
[Bug] Change include_dst_in_src to False in NeighborSampler

mfbalin:
Including dst vertices in src causes the recursive sampling procedure to do unnecessary sampling for dst vertices who are not originally src vertices. This causes the sampled neighborhood to grow faster than it should resulting into wasted work. With this simple change, the number of vertices sampled goes from [176k, 84.3k 14.1k, 1k] to [167k, 77.1k, 13.1k, 1k] on reddit for batch size 1000, fanout 10,10,15 with 3 layers.

### DIFF
--- a/python/dgl/dataloading/neighbor_sampler.py
+++ b/python/dgl/dataloading/neighbor_sampler.py
@@ -112,7 +112,7 @@ class NeighborSampler(BlockSampler):
                 replace=self.replace, output_device=self.output_device,
                 exclude_edges=exclude_eids)
             eid = frontier.edata[EID]
-            block = to_block(frontier, seed_nodes)
+            block = to_block(frontier, seed_nodes, False)
             block.edata[EID] = eid
             seed_nodes = block.srcdata[NID]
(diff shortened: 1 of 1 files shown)


BarclayII:
`include_dst_in_src` defaults to True because in most cases (e.g. GraphSAGE, RGCN) you will need to compute the seed node's representation in the previous layer even if it is not a sampled neighbor, and the overhead is small (~10% as you showed).  See the second figure of https://docs.dgl.ai/tutorials/large/L0_neighbor_sampling_overview.html#sphx-glr-tutorials-large-l0-neighbor-sampling-overview-py for an example.

mfbalin:
I see, so since SAGEConv has skip connections enabled by default with the exception of the gcn aggregation type, it is left enabled by default. I have a sampling algorithm for which this option makes 2x difference so I wanted to clarify this. Also when I looked at the code for SAGEConv, dst vertices when they are in src vertices, all come before all the other src vertices, which is good to know.

Issue #4125, opened 2022-06-15
Using Dataloader on a HeteroGraph and setting shuffle to False, the output_nodes are out of order.


maqy1995:
DGL version: 0.8.2

code reproduce:
```python
import dgl
from ogb.nodeproppred import DglNodePropPredDataset

dataset = DglNodePropPredDataset(name="ogbn-mag", root='/mnt/disk1/dataset/gnn_dataset/dgl')
hetero_g, labels = dataset[0]
splitted_idx = dataset.get_idx_split()
train_nid = splitted_idx["train"]['paper']
val_nid = splitted_idx["valid"]['paper']
test_nid = splitted_idx["test"]['paper']

train_fanouts = []
for layer in range(2):
    train_fanouts.append({etype: 10 + layer for etype in hetero_g.canonical_etypes})

train_neighbor_sampler = dgl.dataloading.NeighborSampler(fanouts=train_fanouts)

train_dataloader = dgl.dataloading.DataLoader(
    graph=hetero_g,
    indices={'paper': train_nid},
    graph_sampler=train_neighbor_sampler,
    device="cpu",
    batch_size=100,
    shuffle=False,
    drop_last=False,
    num_workers=0)

train_iter = iter(train_dataloader)
_, first_seeds, _ = next(train_iter)

# should be 0,1,2...98,99 ?
print(first_seeds)

output:
{'paper': tensor([62, 74, 73, 72, 71, 70, 69, 68, 67, 66, 65, 64, 63, 75, 61, 60, 59, 58,
        57, 56, 55, 54, 53, 52, 51, 87, 99, 98, 97, 96, 95, 94, 93, 92, 91, 90,
        89, 88, 50, 86, 85, 84, 83, 82, 81, 80, 79, 78, 77, 76, 12, 24, 23, 22,
        21, 20, 19, 18, 17, 16, 15, 14, 13, 25, 11, 10,  9,  8,  7,  6,  5,  4,
         3,  2,  1, 37, 49, 48, 47, 46, 45, 44, 43, 42, 41, 40, 39, 38,  0, 36,
        35, 34, 33, 32, 31, 30, 29, 28, 27, 26])}
```

As we can see, the `first_seeds` is out of order, rather than a range from 0 to 100.

The reason for that seems to be `torch.argsort()` is not stable.
```python
class _TensorizedDatasetIter(object):

    def __next__(self):
        # convert the type-ID pairs to dictionary
        type_ids = batch[:, 0]
[...]

PR #4147, opened 2022-06-20
[BugFix] fix unstable sort when using dataloader with HeteroGraph

maqy1995:
## Description
When using Dataloader on a HeteroGraph and setting shuffle to False, the output_nodes are out of order because of the unstable sort. related issue:https://github.com/dmlc/dgl/issues/4125

## Checklist
- [ ] Code is well-documented

## relevant information
related pytorch stable sort issue: https://github.com/pytorch/pytorch/issues/38681
pytorch `torch.sort` doc: https://pytorch.org/docs/stable/generated/torch.argsort.html?highlight=argsort#torch.argsort
pytorch `torch.argsort` doc: https://pytorch.org/docs/stable/generated/torch.argsort.html?highlight=argsort#torch.argsort

## may need attention
According the pytorch doc, the cpu version of `stable torch.sort` was introduced in version 1.9, and the cuda version of `stable torch.sort` was introduced in 1.10.

### DIFF
--- a/python/dgl/dataloading/dataloader.py
+++ b/python/dgl/dataloading/dataloader.py
@@ -72,7 +72,7 @@ class _TensorizedDatasetIter(object):
         # convert the type-ID pairs to dictionary
         type_ids = batch[:, 0]
         indices = batch[:, 1]
-        type_ids_sortidx = torch.argsort(type_ids)
+        _, type_ids_sortidx = torch.sort(type_ids, stable=True)
         type_ids = type_ids[type_ids_sortidx]
         indices = indices[type_ids_sortidx]
         type_id_uniq, type_id_count = torch.unique_consecutive(type_ids, return_counts=True)


BarclayII (approved):

python/dgl/dataloading/dataloader.py
```diff
@@ -72,7 +72,7 @@ def __next__(self):
         # convert the type-ID pairs to dictionary
         type_ids = batch[:, 0]
         indices = batch[:, 1]
-        type_ids_sortidx = torch.argsort(type_ids)
+        _, type_ids_sortidx = torch.sort(type_ids, stable=True)
```
BarclayII, line 75: I guess it's better to raise a warning for older versions of PyTorch when stable sorting is not available and resort to non-stable version (or does PyTorch do that already?)

maqy1995: I agree.
It seems older PyTorch will not resort to non-stable version, it just raises an invalid arguments error.
I add the torch version check and raise a warning for older versions of Pytorch. One problem is the warning will also be raised when `shuffle` is `True`, the disorder of the output_nodes should not matter in this scenario.

BarclayII: > One problem is the warning will also be raised when `shuffle` is `True`, the disorder of the output_nodes should not matter in this scenario.

Then you could choose to only raise the warning if `shuffle` is False.

> by the way, dgl_warning() seems will repeat the first line?

You mean the DGLWarning will be repeated for each process?  I guess there's not a good way to handle that, but the purpose of warnings is to let users know that there is a problem when `shuffle` is False.
Another solution would be using `np.argsort` (that has a stable option) rather than `torch.sort` if `type_ids` is on CPU, and just raise an error if `type_ids` is on GPU, PyTorch version is too old, and `shuffle=False`.

maqy1995: > Then you could choose to only raise the warning if shuffle is False.

I thought about this solution, but there is no shuffle parameter passed in _TensorizedDatasetIter(object).

> You mean the DGLWarning will be repeated for each process?

what I mean here is the first line of the message in dgl_warning will be printed again. For example, below is the full output, the line `The current output_nodes are out of order even if set shuffle to False in Dataloader ` will be printed again at the end of the message.

>DGLWarning: The current output_nodes are out of order even if set shuffle to False in Dataloader, the reason is that the current version of torch dose not support stable sort. Please update torch to 1.10.0 or higher to fix it.
' The current output_nodes are out of order even if set shuffle to False in Dataloader, '
set `line=''` in `_default_formatwarning(message, category, filename, lineno, line=None)`  may avoid this.
```
def dgl_warning_format(message, category, filename, lineno, line=None):
    """Format DGL warnings."""
    if isinstance(category, DGLWarning):
        return "DGL Warning: {}\n".format(message)
    else:
        return _default_formatwarning(message, category, filename, lineno, line=None)
```

> Another solution would be using np.argsort (that has a stable option) rather than torch.sort if type_ids is on CPU, and just raise an error if type_ids is on GPU, PyTorch version is too old, and shuffle=False.

What I was thinking about before was trying to get the same behavior for the cpu or gpu,  and avoid adding too much code.

BarclayII: Decided to pass in a shuffle argument to the iterator.

I'm also thinking of the options to just deprecating PyTorch <1.10.  @jermainewang you mentioned previously that we only plan to support the last three stable versions.  Did we formally announce this policy?
(Not an actual review)

jermainewang:
@zjost Could you post your reply to #4224 or create a new thread? It will be hard for people to follow conversations in merged PRs.

PR #4159, opened 2022-06-22
[Example][Refactor] Initial effort on refactoring graphsage examples

chang-l:
Updates on 06/24:

## Major changes since last revision
- Reduced the number of cmd line options, and re-positioned all parameters to make users easy and convenient to change
- Removed load_data() and register_data_arg() functions, due to the limitation of available datasets and lack of documentation.
- Deprecated manually pin_memory_inplace()/unpin_memory_inplace functions, thanks to @yaox12 's latest PR. And, these two functions are not documented as well...
- Slight change on output format, i.e., now report timing w.r.t per epoch and report total train time in the end (What is ETputs? is it a well-accepted quantity?)
```
Epoch xxx | Time(s) 0.0058 | Loss 0.0875 | Accuracy 0.6480 | ETputs(KTEPS) 1579.65
Training time(s) 2.5515

Test Accuracy 0.6760
```
- Fanout is corrected from [15,10,5]-->[5,10,15]
- Support --graph-device cpu/gpu/uva for multi-gpu mini-batch training

## Issues:
- File naming convention. Current file naming is not consistent and confusing, e.g., train_full vs. node_classification.py. Any suggestions?
- File consolidation. Do we need to keep multi-gpu and single-gpu implementations as two sperate files?

---
Graphsage seems one of the most up-to-date example case, and also one of most comprehensive example including single/multi-GPU/multi-node implementations and many other advanced usages. Apparently, these implementations/script files have been developed by different contributors or within different time periods. It might be a good idea to start our example refactoring on this case because of its (1) comprehensiveness, so we can easily extend to other examples; and (2) (relative) up-to-date scripts, i.e., less effort to refactor.

In general, the goals are:
1. Consistency: configurations (definition of input arguments), output format, general code structure, etc.
2. Succinct: avoid redundant code and main script should be succinct. 
3. Maintainability: Modularize the code to ease reusing and future extension

Currently, I have only refactored the basic examples (not touching all `adv.` folders) with full/mini-batch training using CPU/GPU/multi-GPU. Affected files are: `train_full.py`, `node_classification.py` and `multi_gpu_node_classification.py`.

## Checklist
- [x] Code is well-documented

## Detailed Changes
- Move all graphsage model definitions to a new file, `model.py`.
       Within `model.py`, inheritance/dependency list is:   `nn.Module` -> `GraphSage` --> `GraphSageBatch` --> `GraphSageBatchMultiGPU`, thus every implementation can import a corresp. SAGE module to avoid redundant code.
- Align all input/output format.
      Originaly, minibatch runs (`node_classificaiton.py` and `multi_gpu_node_classification.py`) have everything hard-coded and no "control knobs". For example, `multi_gpu_node_classification.py` can only take all available GPUs, and no `--gpu` to test the speed-up of different # GPUs. Also, minibatch runs' output are messy and lack of information.  
       Now, minibatch runs share the same set of inputs, e.g., `--n-epochs`, `--batch-size`, `--gpu`, etc. And they also share a similar set of output, where per-epoch stats is the finest reporting scale (disable per batch printing). Sample out:
```
	---Data Stats ---
	...	
        Epoch 00000 | Avg. Time(s) 6.6292 | Loss 1.3383 | Accuracy 0.8547 | GPU Mem(MB) 523.98
        ...
	Test Accuracy 0.8030
```
- Align code structures in scripts, and replace all `cuda()` --> `to(device)`
- Eliminate warning msgs in train_full.py by updating to the latest dgl API (deprecated `dataset.graph`)

## Remaining Issues
1. `link_pred.py`, `lighting/`, `advanced/`, `dist/`, and `distgnn/` are not updated at this point. Probably will keep tracking these when this PR is open.
2. `multi_gpu_node_classification.py` seems can cover `node_classification.py` (i.e., specifying `--gpu 0`). Do we really need to keep both files?
3. In `node_classification.py`, maybe break input option `--pure-gpu` into `--graph-device` and `--data-device`, similar to `advanced/train_sampling_upsupervised.py`?

## Tests:

- `python3 train_full.py --dataset cora --gpu 0`
```
...
Epoch 00002 | Avg. Time(s) 0.2767 | Loss 1.9332 | Accuracy 0.2120 | ETputs(KTEPS) 38.15
...
Test Accuracy 0.8090
```
- `python3 node_classification.py --n-epochs 5 --gpu 0 --pure-gpu`
```
...
Epoch 00002 | Avg. Time(s) 2.2549 | Loss 0.5173 | Accuracy 0.8833 | GPU Mem(MB) 6933.29
...
Test Accuracy 0.7401
```
- `python3 multi_gpu_node_classification.py --n-epochs 5 --gpu 0,1`
```
...
Epoch 00001 | Avg. Time(s) 2.5231 | Loss 0.8258 | Accuracy 0.8519 | GPU Mem(MB) 521.92
...
Test Accuracy on Master 0.6931
```

### DIFF
--- a/examples/pytorch/graphsage/model.py
+++ b/examples/pytorch/graphsage/model.py
@@ -1,35 +1,36 @@
 import torch
 import torch.nn as nn
 import dgl
-from dgl import DGLGraph
 from dgl.nn.pytorch.conv import SAGEConv
+import torch.nn.functional as F
 import tqdm
-
+# below import for distributed/multi-gpu training examples
 from dgl.multiprocessing import shared_tensor
-from dgl.utils import pin_memory_inplace, unpin_memory_inplace
 import torch.distributed as dist
+
+# parameters can be tuned as below, or as keyword arguments when create the model
+activation = F.relu
+dropout = 0.5
 
 class GraphSAGE(nn.Module):
     def __init__(self,
                  in_feats,
-                 n_hidden,
                  n_classes,
+                 n_hidden,
                  n_layers,
-                 activation,
-                 dropout,
                  aggregator_type):
         super().__init__()
         self.layers = nn.ModuleList()
         self.dropout = nn.Dropout(dropout)
         self.activation = activation
-
+        
         # input layer
         self.layers.append(SAGEConv(in_feats, n_hidden, aggregator_type))
         # hidden layers
         for i in range(n_layers):
             self.layers.append(SAGEConv(n_hidden, n_hidden, aggregator_type))
             # output layer
-        self.layers.append(SAGEConv(n_hidden, n_classes, aggregator_type)) # activation None
+        self.layers.append(SAGEConv(n_hidden, n_classes, aggregator_type))
 
     def forward(self, graph, inputs):
         h = self.dropout(inputs)
@@ -43,13 +44,11 @@ class GraphSAGE(nn.Module):
 class GraphSAGEBatch(GraphSAGE):
     def __init__(self,
                  in_feats,
-                 n_hidden,
                  n_classes,
+                 n_hidden,
                  n_layers,
-                 activation,
-                 dropout,
                  aggregator_type):
-        super().__init__(in_feats, n_hidden, n_classes, n_layers, activation, dropout, aggregator_type)
+        super().__init__(in_feats, n_classes, n_hidden, n_layers, aggregator_type)
         self.n_hidden = n_hidden
         self.n_classes = n_classes
         
@@ -62,33 +61,6 @@ class GraphSAGEBatch(GraphSAGE):
                 h = self.dropout(h)
         return h
 
-    def inference(self, g, device, batch_size):
-        feat = g.ndata['feat']
-        sampler = dgl.dataloading.MultiLayerFullNeighborSampler(1, prefetch_node_feats=['feat'])
-        dataloader = dgl.dataloading.DataLoader(
-                g, torch.arange(g.num_nodes()).to(g.device), sampler, device=device,
-                batch_size=batch_size, shuffle=False, drop_last=False,
-                num_workers=0)
-
-        for l, layer in enumerate(self.layers):
-            y = torch.empty(
-                g.num_nodes(), self.n_hidden if l != len(self.layers) - 1 else self.n_classes,
-                device='cpu', pin_memory=True)
-            feat = feat.to(device)
-            for input_nodes, output_nodes, blocks in tqdm.tqdm(dataloader):
-                # use an explicitly contiguous slice
-                x = feat[input_nodes]
-                h = layer(blocks[0], x)
-                if l != len(self.layers) - 1:
-                    h = self.activation(h)
-                    h = self.dropout(h)
-                # be design, our output nodes are contiguous so we can take
-                # advantage of that here
-                y[output_nodes[0]:output_nodes[-1]+1] = h.to('cpu')
-            feat = y
-        return y
-
-class GraphSAGEBatchMultiGPU(GraphSAGEBatch):
     def inference(self, g, device, batch_size):
         """
         Perform inference in layer-major order rather than batch-major order.
@@ -113,20 +85,53 @@ class GraphSAGEBatchMultiGPU(GraphSAGEBatch):
             tensor
                 The predictions for all nodes in the graph.
         """
-        g.ndata['h'] = g.ndata['feat']
-        sampler = dgl.dataloading.MultiLayerFullNeighborSampler(1, prefetch_node_feats=['h'])
+        feat = g.ndata['feat']
+        sampler = dgl.dataloading.MultiLayerFullNeighborSampler(1, prefetch_node_feats=['feat'])
         dataloader = dgl.dataloading.DataLoader(
-                g, torch.arange(g.num_nodes(), device=device), sampler, device=device,
+                g, torch.arange(g.num_nodes()).to(g.device), sampler, device=device,
                 batch_size=batch_size, shuffle=False, drop_last=False,
-                num_workers=0, use_ddp=True, use_uva=True)
+                num_workers=0)
 
         for l, layer in enumerate(self.layers):
+            y = torch.empty(
+                g.num_nodes(), self.n_hidden if l != len(self.layers) - 1 else self.n_classes,
+                device='cpu', pin_memory=True)
+            feat = feat.to(device)
+            for input_nodes, output_nodes, blocks in tqdm.tqdm(dataloader):
+                # use an explicitly contiguous slice
+                x = feat[input_nodes]
+                h = layer(blocks[0], x)
+                if l != len(self.layers) - 1:
+                    h = self.activation(h)
+                    h = self.dropout(h)
+                # by design, our output nodes are contiguous so we can take
+                # advantage of that here
+                y[output_nodes[0]:output_nodes[-1]+1] = h.to('cpu')
+            feat = y
+        return y
+
+class GraphSAGEBatchMultiGPU(GraphSAGEBatch):
+     """
+        Perform inference in layer-major order rather than batch-major order within the context of multi-gpu training.
+     """
+     def inference(self, g, device, batch_size):
+        g.ndata['h'] = g.ndata['feat']
+        sampler = dgl.dataloading.MultiLayerFullNeighborSampler(1, prefetch_node_feats=['h'])
+        use_uva = (g.device != device)
+
+        for l, layer in enumerate(self.layers):
+            """
+               Have to re-initialize dataloader for every layer to allow tensor pinned by DGL being automatically unpinned
+            """
+            dataloader = dgl.dataloading.DataLoader(
+                g, torch.arange(g.num_nodes(), device=device), sampler, device=device,
+                batch_size=batch_size, shuffle=False, drop_last=False,
+                num_workers=0, use_ddp=True, use_uva=use_uva)
+             
             # in order to prevent running out of GPU memory, we allocate a
-            # shared output tensor 'y' in host memory, pin it to allow UVA
-            # access from each GPU during forward propagation.
+            # shared output tensor 'y' in host memory
             y = shared_tensor(
-                    (g.num_nodes(), self.n_hidden if l != len(self.layers) - 1 else self.n_classes))
-            pin_memory_inplace(y)
+                (g.num_nodes(), self.n_hidden if l != len(self.layers) - 1 else self.n_classes))
 
             for input_nodes, output_nodes, blocks in tqdm.tqdm(dataloader) \
                     if dist.get_rank() == 0 else dataloader:
@@ -138,12 +143,13 @@ class GraphSAGEBatchMultiGPU(GraphSAGEBatch):
                 y[output_nodes] = h.to(y.device)
             # make sure all GPUs are done writing to 'y'
             dist.barrier()
-            if l > 0:
-                unpin_memory_inplace(g.ndata['h'])
             if l + 1 < len(self.layers):
                 # assign the output features of this layer as the new input
                 # features for the next layer
-                g.ndata['h'] = y
+                if use_uva:
+                    g.ndata['h'] = y
+                else:
+                    g.ndata['h'] = y.to(g.device)
             else:
                 # remove the intermediate data from the graph
                 g.ndata.pop('h')
--- a/examples/pytorch/graphsage/model.py
+++ b/examples/pytorch/graphsage/model.py
@@ -1,0 +1,150 @@
+import torch
+import torch.nn as nn
+import dgl
+from dgl import DGLGraph
+from dgl.nn.pytorch.conv import SAGEConv
+import tqdm
+
+from dgl.multiprocessing import shared_tensor
+from dgl.utils import pin_memory_inplace, unpin_memory_inplace
+import torch.distributed as dist
+
+class GraphSAGE(nn.Module):
+    def __init__(self,
+                 in_feats,
+                 n_hidden,
+                 n_classes,
+                 n_layers,
+                 activation,
+                 dropout,
+                 aggregator_type):
+        super().__init__()
+        self.layers = nn.ModuleList()
+        self.dropout = nn.Dropout(dropout)
+        self.activation = activation
+
+        # input layer
+        self.layers.append(SAGEConv(in_feats, n_hidden, aggregator_type))
+        # hidden layers
+        for i in range(n_layers):
+            self.layers.append(SAGEConv(n_hidden, n_hidden, aggregator_type))
+            # output layer
+        self.layers.append(SAGEConv(n_hidden, n_classes, aggregator_type)) # activation None
+
+    def forward(self, graph, inputs):
+        h = self.dropout(inputs)
+        for l, layer in enumerate(self.layers):
+            h = layer(graph, h)
+            if l != len(self.layers) - 1:
+                h = self.activation(h)
+                h = self.dropout(h)
+        return h
+
+class GraphSAGEBatch(GraphSAGE):
+    def __init__(self,
+                 in_feats,
+                 n_hidden,
+                 n_classes,
+                 n_layers,
+                 activation,
+                 dropout,
+                 aggregator_type):
+        super().__init__(in_feats, n_hidden, n_classes, n_layers, activation, dropout, aggregator_type)
+        self.n_hidden = n_hidden
+        self.n_classes = n_classes
+        
+    def forward(self, blocks, x):
+        h = x
+        for l, (layer, block) in enumerate(zip(self.layers, blocks)):
+            h = layer(block, h)
+            if l != len(self.layers) - 1:
+                h = self.activation(h)
+                h = self.dropout(h)
+        return h
+
+    def inference(self, g, device, batch_size):
+        feat = g.ndata['feat']
+        sampler = dgl.dataloading.MultiLayerFullNeighborSampler(1, prefetch_node_feats=['feat'])
+        dataloader = dgl.dataloading.DataLoader(
+                g, torch.arange(g.num_nodes()).to(g.device), sampler, device=device,
+                batch_size=batch_size, shuffle=False, drop_last=False,
+                num_workers=0)
+
+        for l, layer in enumerate(self.layers):
+            y = torch.empty(
+                g.num_nodes(), self.n_hidden if l != len(self.layers) - 1 else self.n_classes,
+                device='cpu', pin_memory=True)
+            feat = feat.to(device)
+            for input_nodes, output_nodes, blocks in tqdm.tqdm(dataloader):
+                # use an explicitly contiguous slice
+                x = feat[input_nodes]
+                h = layer(blocks[0], x)
+                if l != len(self.layers) - 1:
+                    h = self.activation(h)
+                    h = self.dropout(h)
+                # be design, our output nodes are contiguous so we can take
+                # advantage of that here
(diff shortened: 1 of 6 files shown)


mufeili:
One possibility is that we may want to separate advanced usage into separate folders so that there is no interdependency between preliminary examples and advanced examples as their users might belong to different groups. cc @jermainewang for potential discussion.

examples/pytorch/graphsage/node_classification.py
```diff
@@ -3,130 +3,131 @@
 import torch.nn.functional as F
```
mufeili, line 3: Is this script simply sampling-based version of `train_full.py`? If so, the naming can also be confusing.

examples/pytorch/graphsage/train_full.py
```diff
@@ -167,5 +83,42 @@ def main(args):
                         help="Aggregator type: mean/gcn/pool/lstm")
     args = parser.parse_args()
```
mufeili, line 84: I'm not sure if we should have a large number of configurable arguments. This depends on if the README file includes experiment results with different configurations. Otherwise, users should be able to directly modify the script if necessary.

examples/pytorch/graphsage/train_full.py
```diff
@@ -167,5 +83,42 @@ def main(args):
...
     print(args)
+    
+    # load and preprocess dataset
+    data = load_data(args)
```
mufeili, line 88: I prefer directly import the dataset classes from `dgl.data`. `load_data` is just an additional layer of abstraction and it is not applicable for datasets other than Cora, Citeseer, Pubmed. If we believe that they should be loaded with a single interface, then we should use a single class instead.

examples/pytorch/graphsage/train_full.py
```diff
@@ -167,5 +83,42 @@ def main(args):
...
+    print("""----Data statistics------'
+      #Edges %d
+      #Classes %d
+      #Train samples %d
+      #Val samples %d
+      #Test samples %d""" %
+          (n_edges, n_classes,
+           train_mask.int().sum().item(),
+           val_mask.int().sum().item(),
+           test_mask.int().sum().item()))
```
mufeili, line 107: I'm not sure if it is necessary to log dataset statistics.

examples/pytorch/graphsage/train_full.py
```diff
@@ -167,5 +83,42 @@ def main(args):
...
 
-    main(args)
+    if args.gpu < 0:
+        device = torch.device('cpu')
+    else:
+        device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+        features = features.to(device)
+        labels = labels.to(device)
+        g = g.to(device)
```
mufeili, line 115: It will be easier to first do `g = g.to(device)` and then

```
features = g.ndata['feat']
labels = g.ndata['label']
```

examples/pytorch/graphsage/train_full.py
```diff
@@ -167,5 +83,42 @@ def main(args):
...
+        
+    train_nid = train_mask.nonzero().squeeze().to(device)
+    val_nid = val_mask.nonzero().squeeze().to(device)
+    test_nid = test_mask.nonzero().squeeze().to(device)
```
mufeili, line 119: I'm curious if `squeeze` is always used when `train_mask`/`val_mask`/`test_mask` are used.

examples/pytorch/graphsage/train_full.py
```diff
@@ -58,51 +25,11 @@ def evaluate(model, graph, features, labels, nid):
...
-    train_nid = train_mask.nonzero().squeeze()
-    val_nid = val_mask.nonzero().squeeze()
-    test_nid = test_mask.nonzero().squeeze()
-
+    
+def train(args, g, device, data):
+    features, labels, train_nid, val_nid = data
     # graph preprocess and calculate normalization factor
-    g = dgl.remove_self_loop(g)
     n_edges = g.number_of_edges()
```
mufeili, line 32: suggest changing to `g.num_edges()`

examples/pytorch/graphsage/train_full.py
```diff
@@ -58,51 +25,11 @@ def evaluate(model, graph, features, labels, nid):
...
-        print("use cuda:", args.gpu)
-
-    train_nid = train_mask.nonzero().squeeze()
-    val_nid = val_mask.nonzero().squeeze()
-    test_nid = test_mask.nonzero().squeeze()
-
+    
+def train(args, g, device, data):
+    features, labels, train_nid, val_nid = data
     # graph preprocess and calculate normalization factor
```
mufeili, line 31: L31 can be removed

examples/pytorch/graphsage/train_full.py
```diff
@@ -111,40 +38,29 @@ def main(args):
...
     # use optimizer
     optimizer = torch.optim.Adam(model.parameters(), lr=args.lr, weight_decay=args.weight_decay)
 
     # initialize graph
     dur = []
     for epoch in range(args.n_epochs):
         model.train()
-        if epoch >= 3:
-            t0 = time.time()
+        t0 = time.time()
```
mufeili, line 49: The speed becomes more stable after first running for a few epochs, though I'm not sure if it is necessary to include timing logic in the example code. @jermainewang

examples/pytorch/graphsage/train_full.py
```diff
@@ -111,40 +38,29 @@ def main(args):
...
 
         optimizer.zero_grad()
         loss.backward()
         optimizer.step()
-
-        if epoch >= 3:
-            dur.append(time.time() - t0)
-
+        
+        dur.append(time.time() - t0)
```
mufeili, line 58: Ditto

examples/pytorch/graphsage/train_full.py
```diff
@@ -14,40 +14,7 @@
 import dgl
 from dgl import DGLGraph
 from dgl.data import register_data_args, load_data
```
mufeili, line 16: `nx`, `nn`, `DGLGraph`, and `register_data_args` not used

examples/pytorch/graphsage/model.py
```diff
@@ -0,0 +1,150 @@
+import torch
+import torch.nn as nn
+import dgl
+from dgl import DGLGraph
```
mufeili, line 4: Not used

examples/pytorch/graphsage/model.py
```diff
@@ -0,0 +1,150 @@
...
+from dgl.nn.pytorch.conv import SAGEConv
+import tqdm
+
+from dgl.multiprocessing import shared_tensor
+from dgl.utils import pin_memory_inplace, unpin_memory_inplace
+import torch.distributed as dist
```
mufeili, line 10: Logic related to these can be a bit confusing if a user is only interested in simplest usage.

examples/pytorch/graphsage/model.py
```diff
@@ -0,0 +1,150 @@
...
+        self.dropout = nn.Dropout(dropout)
+        self.activation = activation
+
+        # input layer
+        self.layers.append(SAGEConv(in_feats, n_hidden, aggregator_type))
+        # hidden layers
+        for i in range(n_layers):
+            self.layers.append(SAGEConv(n_hidden, n_hidden, aggregator_type))
+            # output layer
+        self.layers.append(SAGEConv(n_hidden, n_classes, aggregator_type)) # activation None
```
mufeili, line 32: The comment `# activation None` might be unnecessary.

examples/pytorch/graphsage/model.py
```diff
@@ -0,0 +1,150 @@
...
+                device='cpu', pin_memory=True)
+            feat = feat.to(device)
+            for input_nodes, output_nodes, blocks in tqdm.tqdm(dataloader):
+                # use an explicitly contiguous slice
+                x = feat[input_nodes]
+                h = layer(blocks[0], x)
+                if l != len(self.layers) - 1:
+                    h = self.activation(h)
+                    h = self.dropout(h)
+                # be design, our output nodes are contiguous so we can take
```
mufeili, line 85: I guess "be design" should be "by design"?

examples/pytorch/graphsage/model.py
```diff
@@ -0,0 +1,150 @@
...
+            device : context
+                The device this process should use for inference
+            batch_size : int
+                The number of items to collect in a batch.
+
+        Returns
+        -------
+            tensor
+                The predictions for all nodes in the graph.
+        """
```
mufeili, line 115: To be consistent with `GraphSAGEBatch`, either we do not add doc for it, or we have a brief doc for both.

examples/pytorch/graphsage/model.py
```diff
@@ -0,0 +1,150 @@
...
+            dist.barrier()
+            if l > 0:
+                unpin_memory_inplace(g.ndata['h'])
+            if l + 1 < len(self.layers):
+                # assign the output features of this layer as the new input
+                # features for the next layer
+                g.ndata['h'] = y
+            else:
+                # remove the intermediate data from the graph
+                g.ndata.pop('h')
```
mufeili, line 149: @BarclayII , does this mean `with g.local_scope` cannot be used here?

yaox12:
I can answer some questions here.
> 2. `multi_gpu_node_classification.py` seems can cover `node_classification.py` (i.e., specifying `--gpu 0`). Do we really need to keep both files?

The last time we refactored graphsage examples,  Quan wanted to keep them as simple as possible for GNN beginners so most options are removed. For example, the single-gpu example has a much simpler inference method than the multi-gpu version and doesn't need to set up the torch.distributed env.

> 3. In `node_classification.py`, maybe break input option `--pure-gpu` into `--graph-device` and `--data-device`, similar to `advanced/train_sampling_upsupervised.py`?

Generally, in the latest DGL, we don't want users to split graph structure and node/edge features. Users should let the dataloader handle them automatically. But I'm supportive of changing `--pure-gpu` into `--graph-device` which allows choices from (`gpu`, `uva`, `cpu`).

examples/pytorch/graphsage/node_classification.py
```diff
@@ -3,130 +3,131 @@
...
+def train(args, g, device, data):
+    in_feats, n_classes, train_idx, valid_idx = data
+    train_idx = train_idx.to(device)
+    valid_idx = valid_idx.to(device)
 
-print(np.mean(durations[4:]), np.std(durations[4:]))
+    model = GraphSAGEBatch(in_feats, args.n_hidden, n_classes, args.n_layers, F.relu, args.dropout, args.aggregator_type).to(device)
+    opt = torch.optim.Adam(model.parameters(), lr=args.lr, weight_decay=args.weight_decay)
+
+    sampler = dgl.dataloading.NeighborSampler([15, 10, 5],
```
yaox12, line 41: Fanout should be `[5, 10, 15]`. This has been a mistake for a long time. PyG uses `[15, 10, 5]` because the first fanout is for seed nodes in PyG. https://github.com/pyg-team/pytorch_geometric/blob/6c6e2cc1c350d97eeeab20f698fb1c506aed441e/examples/ogbn_products_sage.py#L21
However, DGL samples fanouts in a reversed order, that is, the last fanout is for seed nodes.

examples/pytorch/graphsage/model.py
```diff
@@ -0,0 +1,150 @@
...
+                The predictions for all nodes in the graph.
+        """
+        g.ndata['h'] = g.ndata['feat']
+        sampler = dgl.dataloading.MultiLayerFullNeighborSampler(1, prefetch_node_feats=['h'])
+        dataloader = dgl.dataloading.DataLoader(
+                g, torch.arange(g.num_nodes(), device=device), sampler, device=device,
+                batch_size=batch_size, shuffle=False, drop_last=False,
+                num_workers=0, use_ddp=True, use_uva=True)
+
+        for l, layer in enumerate(self.layers):
```
yaox12, line 123: Multi-GPU inference should be updated based on TOT.

examples/pytorch/graphsage/model.py
chang-l: Thanks. I will remove it

examples/pytorch/graphsage/model.py
chang-l: Okay. I will remove the comment

examples/pytorch/graphsage/model.py
chang-l: Thanks for catching it. Fixed.

examples/pytorch/graphsage/train_full.py
chang-l: Thanks for catching these. In fact, register_data_args is used. I removed nx,nn and DGLGraph

chang-l: Also, I cannot find any documentation for functions like: dgl.data.load_data  and dgl.data.register_data_args. Can you please point where I can find the descriptions for these functions in dgl library?

If they have not been documented, I am more inclined to say avoiding use them as much as possible. I will remove&test within my next update.

mufeili: I don't know doc for them as well. I also think it's a bit strange and inconsistent to have them only for `Cora`, `Citeseer`, and `Pubmed`.

examples/pytorch/graphsage/model.py
chang-l: Thank you for pointing this out. I will update it accordingly.

examples/pytorch/graphsage/train_full.py
chang-l: Got it. yeah... like you mentioned, using dgl.data instead of load_data can circumvent it. We should remove all load_data/register_data_args in the rest of the examples as there is no supporting doc. Ideally, it would be nice to completely deprecate load_data/register_data_args, since it seems can be fully covered by dgl.data (right?)

examples/pytorch/graphsage/train_full.py
chang-l: Sure, I will remove it.

examples/pytorch/graphsage/train_full.py
chang-l: Thanks. They are aliasing each other...

examples/pytorch/graphsage/model.py
chang-l: Well, I think it might be better to have such detailed comment in the code, especially for something that is not straightforward. (performing inference layer-major instead of batch-major is typical/standardized or well-known?) 

Also, I did notice another similar comment, like the `inference` function within `advanced/model.py` file. We should avoid repeating ourselves. I will keep in mind when I refactor advanced/ scripts.

I agree with you on consistency here. How about I move this detailed comment to GraphSAGEBatch:inference function, and only paste its first sentence to GraphSAGEBatchMultiGPU:inference?

examples/pytorch/graphsage/train_full.py
mufeili: Yes, but these are long-lasting APIs. So I'll just remove their usage in the examples for now in case deprecating them breaks too many codebases.

examples/pytorch/graphsage/model.py
mufeili: The proposal sounds good to me. However, I guess we can wait till we decide if they should be separate examples.

examples/pytorch/graphsage/model.py
chang-l: Let me add comment to clarify the last several lines are for dist/multi-gpu training

examples/pytorch/graphsage/node_classification.py
chang-l: Yes, it is confusing and quite annoying... we definitely need to fix.
Alternative 1: train_full.py  train_sampling.py, but in this way, link_pred.py is an outlier which is named after type of task..
Alternative 2: node_classification_full.py, node_classification.py link_pred.py.   these sound less confusing ; but quite annoying to adopt such convention to other examples (i.e., always name after type of task) ......

I wanna defer this until I finished updating link_pred.py unless you have better proposal(s) :)

examples/pytorch/graphsage/node_classification.py
chang-l: Thank you Xin. I will fix and note it, and propagate to other cases.

examples/pytorch/graphsage/train_full.py
chang-l: I think it might be better to let users know/be more clear on the scale of the case they are running ahead of the time.
 
But, yeah,  since dgl intrinsic dataset like Cora do auto-logging. I will remove this manual effort for train_full.py (which uses intrinsic dataset unlike node_classification.py)

examples/pytorch/graphsage/train_full.py
chang-l: Okay. See the above comment.  I will replace all load_data to dgl.data

examples/pytorch/graphsage/node_classification.py
mufeili: Sure. We don't necessarily need to have the same file names across all examples. As you said, we can decide later.

examples/pytorch/graphsage/train_full.py
chang-l: In terms of README file, I don't think we need to cover test results with different configurations. We can drop a pointer in README and it should be easy for them to find out what can be/how to configure in the code.

In fact, we should also simplify reported results in README for all cases. Because: (1) results/performance changing rapidly depends on hardware and self-evolving dgl/other libraries; (2) users don't really care/trust what we reported in README. The impression is more from their on-screen observations and own applications.

To sum up, I think, in README, we only need (1) model name, brief description of references; (2) required pkgs (3) sample command line(s) to run examples (4) just approx. train accuracy results for user verification.

examples/pytorch/graphsage/train_full.py
chang-l: Okay. I see. Let me update accordingly.

examples/pytorch/graphsage/train_full.py
chang-l: Not for all examples.  But I think yes in graphsage. 
We should use "squeeze" instead of full train_mask as much as possible for performance, especially when graph size scales up

examples/pytorch/graphsage/train_full.py
chang-l: From my point of view, it doesn't hurt reporting elapse time per epoch. What I don't quite understand is, originally (1) why we report average time for each epoch (that seems not quite intuitive and informative to me....) (2) what is ETputs? edge throughput? Is it a well-acknowledged quantity to quantify speed as (# of edge/ms)? If not, I think we should remove it cuz it's confusing.  Also, I think it makes more sense to use the actual elapse epoch time to quantify "speed" though....

I slightly update the logging format to have per epoch elapse time and total training time as :
```
Epoch xxx | Time(s) 0.0058 | Loss 0.0875 | Accuracy 0.6480 | ETputs(KTEPS) 1579.65
Training time(s) 2.5515

Test Accuracy 0.6760
```

In this way, the first few epoch slow-down is very clear in log.

examples/pytorch/graphsage/train_full.py
```diff
@@ -111,40 +38,29 @@ def main(args):
                       args.n_layers,
                       F.relu,
```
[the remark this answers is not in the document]
chang-l: Updated and move into model.py as a tunable parameter.

examples/pytorch/graphsage/train_full.py
jermainewang: (1) I don't think reporting training time report is very necessary. Users can always add that themselves.
(2) You could remove the reporting of ETputs (edge training throughput).

PR #4169, opened 2022-06-26
[Bug] Fix that pin_prefetcher is not actually enabled

yaox12:
## Description

Fix that `pin_prefetcher` is not actually enabled.

## Changes

1. Set `pin_memory=self.pin_prefetcher` to enable pin memory for returned indices. Note: subgraph is still not pinned.
2. Correctly enable pin memory for column fetching.

### DIFF
--- a/python/dgl/dataloading/dataloader.py
+++ b/python/dgl/dataloading/dataloader.py
@@ -16,7 +16,7 @@ import torch
 import torch.distributed as dist
 from torch.utils.data.distributed import DistributedSampler
 
-from ..base import NID, EID, dgl_warning
+from ..base import NID, EID, dgl_warning, DGLError
 from ..batch import batch as batch_graphs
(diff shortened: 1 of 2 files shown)


yaox12:
I think `pin_prefetcher` now has the same semantic as Pytorch dataloader's `pin_memory` except that sampled subgraphs are not pinned yet. `pin_prefetcher` is valid only when sampling is on CPU and destination device is GPU and not `use_uva`, so pinning CUDA tensors won't occur.
https://github.com/dmlc/dgl/blob/ec7401516c4affb42d7d1f904d1f56e5256fd023/python/dgl/dataloading/dataloader.py#L735-L743

And since the previous `pin_prefetcher` didn't take effect correctly (as fixed in this PR), I think it should be fine to merge `pin_prefetcher` into PyTorch's native `pin_memory`.

BarclayII (approved):

PR #4187, opened 2022-06-28
code changes for bug fixes identified during mag_lsc dataset

kylasa:
These are the bug fixes identified during testing with 'mag_lsc' dataset. Since this is a larger dataset compared to "mag" we were running into peculiar issues as following:

When creating dgl partition objects, in the convert_partition.py, for some partitions unique node-ids in the edge list were not completely representing all the nodes in that partition. To resolve this we need to change the dgl function call appropriately to pass "num_nodes" argument in the argument list.

When creating a Tensor, for a very large integer list, torch.Tensor() function call used earlier. This was working for small integer list, but when it exceeds certain size we were noticing data corruption in the tensor generate from this function call. This was a very tricky bug to solve. Changed this function call with torch.from_numpy() to create a tensor from numpy array. This creates the target tensor without any data corruption.

## Checklist
- [X] Code is well-documented

### DIFF
--- a/tools/distpartitioning/convert_partition.py
+++ b/tools/distpartitioning/convert_partition.py
@@ -228,7 +228,7 @@ def create_metadata_json(graph_name, num_nodes, num_edges, num_parts, node_map_v
                      'ntypes': ntypes_map,
                      'etypes': etypes_map}
 
-    for part_id in range(num_parts): 
+    for part_id in range(num_parts):
         part_dir = 'part' + str(part_id)
         node_feat_file = os.path.join(part_dir, "node_feat.dgl")
         edge_feat_file = os.path.join(part_dir, "edge_feat.dgl")
--- a/tools/distpartitioning/convert_partition.py
+++ b/tools/distpartitioning/convert_partition.py
@@ -9,6 +9,7 @@ import pyarrow
 import pandas as pd
 import constants
 from pyarrow import csv
+from utils import read_json
 
 def create_dgl_object(graph_name, num_parts, \
                         schema, part_id, node_data, \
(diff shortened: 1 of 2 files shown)


tools/distpartitioning/convert_partition.py
```diff
@@ -183,7 +185,7 @@ def create_dgl_object(graph_name, num_parts, \
 
     return compact_g2, node_map_val, edge_map_val, ntypes_map, etypes_map
 
-def create_metadata_json(graph_name, num_nodes, num_edges, num_parts, node_map_val, \
+def create_metadata_json(graph_name, num_nodes, num_edges, part_id, num_parts, node_map_val, \
```
zheng-da, line 188: i don't understand. you changed the interface of `create_metadata_json` but doesn't change the place where it is called. how does it work?
also, why is this change related to the bug.

tools/distpartitioning/globalids.py
```diff
@@ -50,7 +50,7 @@ def get_shuffle_global_nids(rank, world_size, global_nids_ranks, node_data):
     #form the outgoing message
     send_nodes = []
     for i in range(world_size):
-        send_nodes.append(torch.Tensor(global_nids_ranks[i]).type(dtype=torch.int64))
+        send_nodes.append(torch.from_numpy(global_nids_ranks[i]).type(dtype=torch.int64))
```
zheng-da, line 53: is it related to the bug?

kylasa: Yes. This is the bug on which I apparently spent a week debugging with the mag_lsc dataset. Originally the tensor was created, with global_nids, using "torch.Tensor()" function call. Please note that the argument here global_nids_ranks[i] is a numpy array. This works for smaller datasets where this list is on the smaller side and I have not noticed any mem. corruption issues. But when testing with mag_lsc dataset, right after this function call when we compare the actual tensor created vs. the data passed to this function call ... they are not identical. There was some wierd kind of mem. corruption hapenning when the input numpy array is very large and torch.Tensor() was used in creating the Tensor. 

This issue seems to be resolved, when torch.from_numpy() function call was used to create the Tensor. In this case for larger messages the data in the newly created tensor and the actual data passed to this function call are always correct.

tools/distpartitioning/convert_partition.py
```diff
@@ -226,12 +228,12 @@ def create_metadata_json(graph_name, num_nodes, num_edges, num_parts, node_map_v
                      'ntypes': ntypes_map,
                      'etypes': etypes_map}
 
-    for part_id in range(num_parts):
+    for part_id in range(num_parts):
```
zheng-da, line 231: don't add space in the end.

kylasa: removed the space as suggested.

PR #4188, opened 2022-06-28
Added code to support multiple-file-support feature and removed singl…

kylasa:
…e-file-support code

1. Added code to read dataset in multiple-file-format
2. Removed code for single-file format

## Checklist
- [X] Code is well-documented


tools/distpartitioning/dataset_utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+    #read (split) xxx_edges.txt file
+    edge_datadict = {}
+    edge_file = input_dir+'/'+graph_name+'_edges'+'_{:02d}.txt'.format(rank)
+    edge_data = np.loadtxt(edge_file, delimiter=' ', dtype='int64')
+    edge_datadict[constants.GLOBAL_SRC_ID] = edge_data[:,0]
+    edge_datadict[constants.GLOBAL_DST_ID] = edge_data[:,1]
+    edge_datadict[constants.GLOBAL_TYPE_EID] = edge_data[:,2]
+    edge_datadict[constants.ETYPE_ID] = edge_data[:,3]
+    print('[Rank: ', rank, '] Done reading edge_file: ', len(edge_datadict), edge_datadict[constants.GLOBAL_SRC_ID].shape)
```
zheng-da, line 76: these will be changed after using the new input format, right?

tools/distpartitioning/gloo_wrapper.py
```diff
@@ -59,6 +59,40 @@ def alltoall_cpu(rank, world_size, output_tensor_list, input_tensor_list):
...
+
+    Returns
+    -------
+    list: list of objects are received from other processes
+       This is the list of objects which are sent to the current process by
+       other processes as part of this exchange
+    """
+    rcv_list = []
+    output_list = [None] * world_size
+    for i in range(world_size):
```
zheng-da, line 84: is it right how the for loop works as follow:
* in iteration i, the input data is scattered to all processes from process i;
* in other iterations, process i receives one object in each iteration.
pytorch distributed doesn't support an API like that. this is basically all processes scatter its data to all processes.

tools/distpartitioning/initialize.py
```diff
@@ -490,10 +67,12 @@ def multi_dev_init(params):
...
     #init the gloo process group here. 
-    dist.init_prcess_group("gloo", rank=params.rank, world_size=params.world_size)
-    print('[Rank: ', params.rank, '] Done with process group initialization...')
+    dist.init_process_group("gloo", rank=rank, world_size=params.world_size, timeout=timedelta(seconds=5*60))
+    print('[Rank: ', rank, '] Done with process group initialization...')
 
     #invoke the main function here.
-    proc_exec(params.rank, params.world_size, params)
-    print('[Rank: ', params.rank, '] Done with Distributed data processing pipeline processing.')
+    gen_dist_partitions(rank, params.world_size, params)
```
zheng-da, line 77: what does `gen_dist_partitions` do? the naming is also very confusing.

tools/distpartitioning/multi_dev_init.py
```diff
@@ -0,0 +1,499 @@
...
+    for s in recv_sizes: 
+        output_list.append(torch.zeros(s.tolist(), dtype=torch.int64))
+    
+    dist.barrier()
+    alltoallv_cpu(rank, world_size, output_list, input_list)
+    end = timer()
+    print('[Rank: ', rank, '] Time to exchange node data : ', timedelta(seconds=end - start))
+
+    #stitch together the received data to form a consolidated data-structure
+    rcvd_node_data = torch.cat(output_list).numpy()
```
zheng-da, line 74: why do you concatenate the data and then assign each column to the dict of node data below?

tools/distpartitioning/gloo_wrapper.py
kylasa: Correct. Basically this code is just an adaptation of the sample code present on the pytorch website at the following location (https://pytorch.org/docs/stable/distributed.html). This is adapted from the all_to_all sample implementation as mentioned here.

tools/distpartitioning/multi_dev_init.py
```diff
@@ -0,0 +1,499 @@
...
+
+    Parameters:
+    -----------
+    rank : int
+        rank of the process
+    world_size : int
+        total no. of processes
+    edge_data : dictionary
+        edge information, as a dicitonary which stores column names as keys and values
+        as column data. This information is read from the edges.txt file.
```
zheng-da, line 98: edge_data is used for both input and output. this needs to be documented.
however, this is not a right method. you should return the results instead of modifying the input data.

tools/distpartitioning/dataset_utils.py
kylasa: Yup. I will define them in the constants file and use Strings to indicate these numbers... as suggested by Xiang. I will do this in the next iteration of code commits (newly defined file format implementation that I am working now).

tools/distpartitioning/initialize.py
```diff
@@ -5,457 +5,34 @@
...
-    --------
-    numpy array
-        result of the 'recv' gloo primitive
-    """
-    return recv_data(rank, shape, dtype)
-
-def recv_edge_data(rank, shape, dtype): 
-    """
-    Function to receive edge_data, used by non-rank0 processes. 
+def single_dev_proc_init(params):
```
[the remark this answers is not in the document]
kylasa: The idea was to name the starting function with a suffix "_init", indicating that this function will be called as initialization function on each rank. Please suggest a new naming convention which we can adapt. May be we can target in the next code drop as well.

tools/distpartitioning/initialize.py
kylasa: The idea is to generate distributed partitions, "gen_dist_partitions". This is the function which performs all the steps by our algorithm (data loading, data shuffling, and creating dgl objects). Lets target function renaming in the next iteration of code drop.

tools/distpartitioning/multi_dev_init.py
kylasa: This is after performing alltoallv. the output of this operation is a list of numpy arrays. First I concatenate these individual tensors and convert to numpy arrays to store them in the dictionary, which in this case is node_data. Concatenation is needed because each smaller tensor is received by one of the other ranks in the world.

tools/distpartitioning/multi_dev_init.py
kylasa: Agree. Will change the code as suggested. This means that all the following functions will change: "exchange_node_data, exchange_edge_data and exchange_node_features". I will make these code changes and followup with the code drop in a couple of hours.

tools/distpartitioning/initialize.py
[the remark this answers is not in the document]
zheng-da: can you use `_entry` instead? the suffix _init implies that this function tries to initialize the computation. however, this is not the case. this is just an entry point of the entire computation.

similarly, the file name is also confusing. if this python script contains the entry point, main.py will be better than initialize.py.

tools/distpartitioning/multi_dev_init.py
zheng-da: yes, all functions should be changed to return new dicts.

tools/distpartitioning/multi_dev_init.py
```diff
@@ -0,0 +1,499 @@
...
+    node_part_ids : numpy array
+        numpy array which store the partition-ids and indexed by global_nids
+    """
+
+    #determine Global_type_nid for the residing features 
+    start = timer()
+    node_features_rank_lst = []
+    global_nid_rank_lst = []
+    for part_id in np.arange(world_size):
+
```
zheng-da, line 187: remove the empty line.

tools/distpartitioning/multi_dev_init.py
```diff
@@ -0,0 +1,499 @@
...
+        #form outgoing features to each process
+        send_node_features = {}
+        send_global_nids = {}
+        for ntype_name, ntype_id in ntypes_map.items(): 
+
```
zheng-da, line 192: remove empty line.

tools/distpartitioning/multi_dev_init.py
```diff
@@ -0,0 +1,499 @@
...
+    rcvd_edge_data = torch.cat(output_list).numpy()
+    edge_data[constants.GLOBAL_SRC_ID] = rcvd_edge_data[:,0]
+    edge_data[constants.GLOBAL_DST_ID] = rcvd_edge_data[:,1]
+    edge_data[constants.GLOBAL_TYPE_EID] = rcvd_edge_data[:,2]
+    edge_data[constants.ETYPE_ID] = rcvd_edge_data[:,3]
+    edge_data[constants.GLOBAL_EID] = rcvd_edge_data[:,4]
+    edge_data.pop(constants.OWNER_PROCESS)
+
+
+def exchange_node_features(rank, world_size, node_data, node_features, ntypes_map, \
```
zheng-da, line 143: why can i not find `exchange_edge_features`?

tools/distpartitioning/multi_dev_init.py
```diff
@@ -0,0 +1,499 @@
...
+                #now map the global_ntype_id to global_nid 
+                global_nid_offset = ntypes_nid_map[ntype_name][0]
+                global_nid_start = global_type_nid_start + global_nid_offset
+                global_nid_end = global_type_nid_end + global_nid_offset
+
+                #assert (global_nid_end - global_nid_start) == feature_count
+                global_nids = np.arange(global_nid_start, global_nid_end, dtype=np.int64)
+
+                #determine node feature ownership 
+                #TODO: a Bug here.
```
zheng-da, line 218: what do you mean by this? if there is a bug, the bug needs to be fixed.

tools/distpartitioning/multi_dev_init.py
```diff
@@ -0,0 +1,499 @@
...
+                #TODO: a Bug here. 
+                '''
+                part_ids = node_part_ids[global_nids] 
+                idx = (part_ids == part_id)
+                out_global_nid = global_nids[idx == 1]
+                out_type_nid = type_nid[idx == 1]
+                out_features = node_features[ntype_name+'/feat'][out_type_nid]
+                send_node_features[ntype_name+'/feat'] = out_features
+                send_global_nids[ntype_name+'/feat'] = out_global_nid
+                '''
```
zheng-da, line 227: remove the commented code.

tools/distpartitioning/multi_dev_init.py
```diff
@@ -0,0 +1,499 @@
...
+                    torch.cat((rcvd_node_features[ntype_name+'/feat'], output_list[idx][ntype_name+'/feat']))
+                rcvd_global_nids[ntype_name+'/feat'] = \
+                    np.concatenate((rcvd_global_nids[ntype_name+'/feat'], output_nid_list[idx][ntype_name+'/feat']))
+    end = timer()
+    print('[Rank: ', rank, '] Total time for node feature exchange: ', timedelta(seconds = end - start))
+
+    return rcvd_node_features, rcvd_global_nids
+
+def exchange_graph_data(rank, world_size, node_data, node_features, edge_data,
+        node_part_ids, ntypes_map, ntypes_nid_map, ntype_id_count):
```
zheng-da, line 265: you need to shuffle edge data as well.

tools/distpartitioning/multi_dev_init.py
```diff
@@ -0,0 +1,499 @@
...
+
+    exchange_node_data(rank, world_size, node_data)
+    exchange_edge_data(rank, world_size, edge_data)
+    return rcvd_node_features, rcvd_global_nids
+
+def read_dataset(rank, world_size, node_part_ids, params):
+    """
+    After reading input graph files, add additional information(columns) are added
+    to these data structures, as discussed in detail in the initialize.py file for the
+    case of single-file-format dataset use-case.
```
zheng-da, line 305: this function is used for single file format?

tools/distpartitioning/multi_dev_init.py
```diff
@@ -0,0 +1,499 @@
...
+        for ntype_name, ntype_id in ntypes_map.items(): 
+
+            #check if features exist for this node_type
+            if (ntype_name+'/feat' in node_features) and (node_features[ntype_name+'/feat'].shape[0] > 0):
+                feature_count = node_features[ntype_name+'/feat'].shape[0]
+                global_feature_count = ntype_id_count[str(ntype_id)]
+
+                #determine the starting global_nid for this node_type_id
+                feat_per_proc = math.ceil(global_feature_count / world_size)
+                global_type_nid_start = feat_per_proc * rank
```
zheng-da, line 200: i don't quite understand what we try to do here. `global_type_nid_start` and `global_type_nid_end` indicates the per-type node id range for the node features read from the disk? if so, why can we calculate this? do you assume how the node features are partitioned?

tools/distpartitioning/multi_dev_init.py
kylasa: Agreed. This will be done.

tools/distpartitioning/multi_dev_init.py
kylasa: This code is not instrumented yet. Per one of our discussions, we first implement only exchanging node features because there was no dataset which has edge features and this was never used before. So there is no way to test it on a real dataset.

tools/distpartitioning/multi_dev_init.py
kylasa: Done

tools/distpartitioning/multi_dev_init.py
kylasa: Done.

tools/distpartitioning/multi_dev_init.py
kylasa: Yes. But this is not implemented yet.

tools/distpartitioning/multi_dev_init.py
kylasa: For single file format, there is a separate function, which was present in the initialize.py. But per our discussion, we decided that from now on we only support multiple file format. So that code was deleted.

tools/distpartitioning/multi_dev_init.py
kylasa: Here the process is simple. node-features are mapped to appopriate owner processes in the following steps: 
1. node features are read. 
2. global_type_nid_start and global_type_nid_end are calculated 
3. global_nid for the node features read from file are calculated 
4. with the help of global_nids, we use the node_id to partition_id map to get the owner process
5. We do the above steps for all the ranks in the world.
6. when building node_features_rank_list and global_nid_rank_list, we perform alltoallv and send/receive the node features with corresponding global_nids.

tools/distpartitioning/utils.py
```diff
@@ -323,7 +376,8 @@ def write_dgl_objects(graph_obj, node_features, edge_features, output_dir, part_
 
     part_dir = output_dir + '/part' + str(part_id)
     os.makedirs(part_dir, exist_ok=True)
-    write_graph_dgl(os.path.join(part_dir ,'part'+str(part_id)), graph_obj)
+    #write_graph_dgl(os.path.join(part_dir ,'part'+str(part_id)), graph_obj)
```
zheng-da, line 379: remove the commented code.

kylasa: removed the commented out line as per suggestion.

PR #4192, opened 2022-06-29
[Distributed] Distributed partitioning UX

BarclayII:
## Description
I figured out that converting OGB datasets without any configuration would introduce too many assumptions on the input file hierarchy, so I decided to provide "builtin configurations" for OGB-MAG240M (and later on WikiKG90M as well if necessary - since they have completely different storage formats).

Right now the UX looks like this:
```
python tools/partition_algorithms/random.py <input-metadata> <output-directory> <num-partitions>
```

The graph chunking function is in `tools/chunk_graph.py` which chunks an in-memory DGLGraph as well as on-disk node/edge data into chunked graph format as described in [here](http://13.231.216.217/guide/distributed-preprocessing.html#chunked-graph-format) (internal documentation).

(Reopening since the commit history is messed up last time)

## Checklist
- [ ] Code is well-documented


jermainewang:
Could you merge `array_readwriter` to `utils`?

jermainewang:
I didn't see scripts for processing and chunking `mag240m`. Have you uploaded them?

tools/chunk_graph.py
```diff
@@ -0,0 +1,169 @@
...
+        edge data key.
+    num_chunks : int
+        The number of chunks
+    output_path : pathlike
+        The output directory saving the chunked graph.
+    """
+    with setdir(output_path):
+        _chunk_graph(g, name, ndata_paths, edata_paths, num_chunks, output_path)
+
+if __name__ == '__main__':
```
jermainewang, line 153: Could you create a README in this folder and show the code for processing MAG240M data using this `chunk_graph` function?

tools/partition_algorithms/random.py
```diff
@@ -0,0 +1,39 @@
...
+    mapping files named "<node-type>.txt" (e.g. "author.txt", "paper.txt" and
+    "institution.txt" for OGB-MAG240M).  Each file contains one line per node representing
+    the partition ID the node belongs to.
+    """
+    with setdir(output_path):
+        _random_partition(metadata, num_parts)
+
+# Run with PYTHONPATH=${GIT_ROOT_DIR}/tools
+# where ${GIT_ROOT_DIR} is the directory to the DGL git repository.
+if __name__ == '__main__':
```
jermainewang, line 33: Add a `argparser` and make it runnable like:
```bash
python /my/repo/dgl/tools/partition_algo/random.py
    --in-dir=/mydata/MAG240M-LSC_chunked/
    --out-dir=/mydata/MAG240M-LSC_2parts/
    --num-parts=2
```

tools/sample-config/mag240m-metadata.json
```diff
@@ -0,0 +1,206 @@
+{
```
jermainewang, line 1: Put this in README

tools/sample-config/mag240m-metadata.json
```diff
@@ -0,0 +1,206 @@
...
+        ]
+    ],
+    "edges": {
+        "author:affiliated_with:institution": {
+            "format": {
+                "name": "csv",
+                "delimiter": " "
+            },
+            "data": [
+                "/mnt/efs/gq/random-parts/edge_index/author:affiliated_with:institution0.txt",
```
jermainewang, line 73: make up some fake path

tools/chunk_graph.py
BarclayII: Done.

tools/partition_algorithms/random.py
BarclayII: Done.  Although the first argument would be the metadata file path instead of the input directory.

tools/utils/array_readwriter/csv.py
```diff
@@ -0,0 +1,28 @@
...
+        parse_options = pyarrow.csv.ParseOptions(delimiter=self.delimiter)
+        arr = pyarrow.csv.read_csv(path, read_options=read_options, parse_options=parse_options)
+        logging.info('Done reading from %s' % path)
+        return arr.to_pandas().to_numpy()
+
+    def write(self, path, arr):
+        logging.info('Writing to %s using CSV format with configuration %s' % (
+            path, self.__dict__))
+        write_options = pyarrow.csv.WriteOptions(include_header=False, delimiter=self.delimiter)
+        arr = pyarrow.Table.from_pandas(pd.DataFrame(arr))
```
zheng-da, line 26: pyarrow doesn't work with numpy array?

tools/utils/array_readwriter/numpy_array.py
```diff
@@ -0,0 +1,23 @@
+import logging
+import numpy as np
+from numpy.lib.format import open_memmap
+from .registry import register
+
+@register("numpy")
+class Numpy(object):
```
zheng-da, line 7: i wonder why we need this wrapper? where is it used?
what is the benefit of doing this? the code should be intuitive. it's hard for me to see how this code works. i'm not sure if this is easy for others to understand.

tools/utils/files.py
```diff
@@ -0,0 +1,16 @@
+import os
+from contextlib import contextmanager
+import logging
+
+@contextmanager
+def setdir(path):
```
zheng-da, line 6: is there a better benefit of doing this? is this a python conversion?

tools/utils/array_readwriter/numpy_array.py
jermainewang: I guess it's for reflection.
E.g.,
```
"format" : {"name": "csv", "delimiter": " "},
```
should use the `CSV` parser class, while for
```
"format" : {"name": "numpy"},
```
should use the `Numpy` parser class.

However, the code is not very readable, my suggestions:
- Use name like `NumpyArrayParser` instead of `Numpy`.
- Use `register_array_parser` instead of `register`.
- For getting the parser class from the name,
```python
from .array_readwriter import get_parser
parser = get_parser(name='numpy', **other_kwargs)
parser.read(...)
parser.write(...)
```
Currently, it is hided in the `array_readwriter` module. It is usually difficult for ppl to realize that the reflection is done implicit.

tools/utils/files.py
jermainewang: This is pretty common. See
* https://github.com/jaraco/path
* https://dev.to/teckert/changing-directory-with-a-python-context-manager-2bj8

tools/utils/array_readwriter/csv.py
BarclayII: I don't think pyarrow Tables can directly convert to numpy arrays.  https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table

tools/utils/array_readwriter/numpy_array.py
BarclayII: @zheng-da The benefit is not only what @jermainwang mentioned, but also that it will be straightforward to add in new formats later (either by us or by other people who are modifying our code).  One just needs to write a new class in a new Python file and register it instead of modifying an if-else block or a dictionary.  I'm open to other patterns so let me know if you have other suggestions.

tools/utils/files.py
BarclayII: I could `chdir` in at the beginning and `chdir` out at the end, but I don't feel it pythonic enough.

tools/utils/array_readwriter/numpy_array.py
zheng-da: i see what you are doing here. i think i understand the code now.

zheng-da:
i think all of these are good things if we set up these as a convention for the entire DGL project. We are setting up something here for a very small component, which requires people to take some extra effort to understand it.

other than that, the PR looks good to me.

zheng-da (approved):

jermainewang (approved):

BarclayII:
Closing because I based off a wrong branch.

Issue #3957, opened 2022-04-26
[Memory][PyTorch] Use the TensorAdapter for workspace allocations


nv-dlasalle:
## 🚀 Feature
Currently, DGL ends up using two memory pools if TensorAdapter is used, PyTorch's and then the workspace allocator in DGL. This has two issues, 1) having to concurrent memory pools means we cannot re-use allocations between and error messages from either allocator can be pretty confusing to users seeing each on use only half of the memory the and 2) the DGL workspace memory pool can often take about 2x the memory that is required.

This is relevant to #3933.

If tensoradapter is available, could we redirect workspace allocations to that?

yaox12:
We discussed this issue in last week's core meeting and merging the memory allocator is already in our roadmap.

nv-dlasalle:
Thanks, that great to hear!

Issue #3933, opened 2022-04-19
Device memory consumption issue in GPU sampling


yaox12:
## 🐛 Bug

@BarclayII @nv-dlasalle @yinpeiqi 

When applying GPU sampling, either in UVA or pure-GPU mode, we observe a weird device memory consumption.
1. As shown in the picture, the used memory increases from a small number for the first layer(epoch), but starts from a rather big number for the later layers (epochs).
2. The peak memory usage is the same between epochs.

We want to know:
1. Why does memory usage increase within an epoch?
2. Why does it behaves differently between epochs?

![image](https://user-images.githubusercontent.com/3831900/163915465-687cf909-e4a0-4b66-983a-f66b41a9c6a0.png)

## To Reproduce

```python
import argparse
import pynvml
import torch
import dgl
from dgl.data import RedditDataset


def load_reddit():
    data = RedditDataset(self_loop=True)
    g = data[0]
    g.ndata['features'] = g.ndata['feat']
    return g, data.num_classes

if __name__ == '__main__':
    argparser = argparse.ArgumentParser()
    argparser.add_argument('--use-uva', action="store_true")
    argparser.add_argument('--use-gpu', action="store_true")
    args = argparser.parse_args()

    dataset = load_reddit()
    g : dgl.DGLHeteroGraph = dataset[0]
    feat = g.ndata['feat']

    n_keys = list(g.ndata.keys())
    for k in n_keys:
        g.ndata.pop(k)
    e_keys = list(g.edata.keys())
    for k in e_keys:
        g.edata.pop(k)

    device = "cuda"
    if args.use_gpu:
        g = g.to(device)
        feat = feat.to(device)
    nids = torch.arange(g.number_of_nodes()).to(g.device)
    sampler = dgl.dataloading.MultiLayerFullNeighborSampler(1)
    dataloader = dgl.dataloading.NodeDataLoader(
        g, nids, sampler,
        batch_size=50000,
        shuffle=False,
        drop_last=False,
        use_uva=args.use_uva,
        device=device,
        num_workers=0)
    for l in range(3):
        for input_nodes, output_nodes, blocks in dataloader:
            torch.cuda.empty_cache()
            pynvml.nvmlInit()
            handle = pynvml.nvmlDeviceGetHandleByIndex(0)
            info = pynvml.nvmlDeviceGetMemoryInfo(handle)
            print(f"{blocks[0]}, layer {l}, GPU memory used: {info.used // 1024 ** 2} MB")
[...]

PR #4199, opened 2022-07-01
[Performance] Redirect `AllocWorkspace` to PyTorch's allocator if available

yaox12:
## Description

Related issues: #3933 #3957.

Redirect `AllocWorkspace/FreeWorkspace` to PyTorch's allocator via `raw_alloc` and `raw_delete`.

I run `examples/pytorch/graphsage/node_classification.py` and get the GPU memory footprints as:

|                          | nvidia-smi | max_allocated | allocated | max_reserved | reserved |
|--------------------------|------------|---------------|-----------|--------------|----------|
| new allocator + pure_gpu |      10629 |          9944 |      5542 |         9971 |     9971 |
| old allocator + pure_gpu |      10645 |          7930 |      5243 |         7958 |     7958 |
| new allocator + uva      |       2531 |           550 |       266 |         1480 |     1480 |
| old allocator + uva      |       2591 |           550 |       265 |         1241 |     1241 |

*The four columns on the right are reported by `torch.cuda.max_allocated/allocated/max_reserved/reserved`.

The total GPU memory footprints are close. Advantages are:
1. Users can release the reserved GPU memory via PyTorch's APIs if they'd like to.
2. When PyTorch reports OOM, users won't see a big discrepancy between the memory used by PyTorch and the GPUs capacity.

## Checklist
- [x] Code is well-documented


yaox12:
More backgrounds on PyTorch's CUDA allocator:
1. Call stack of relevant objects: [`CudaCachingAllocator device_allocator`](https://github.com/pytorch/pytorch/blob/c93ceef6585e140c1a33744374f76b2341bfe465/c10/cuda/CUDACachingAllocator.cpp#L1736) → [`THCCachingAllocator caching_allocator`](https://github.com/pytorch/pytorch/blob/c93ceef6585e140c1a33744374f76b2341bfe465/c10/cuda/CUDACachingAllocator.cpp#L1687) → `std::vector<std::unique_ptr<DeviceCachingAllocator>> device_allocator`
2. There’re mainly two ways in PyTorch to allocate CUDA memory:
  
    1. `c10::cuda::CUDACachingAllocator::get()->allocate()` [[code](https://github.com/pytorch/pytorch/blob/c93ceef6585e140c1a33744374f76b2341bfe465/c10/cuda/CUDACachingAllocator.cpp#L1706)]. It uses `CudaCachingAllocator device_allocator` and has a life cycle management (will be freed automatically). Inside, it calls `THCCachingAllocator caching_allocator` for the actual allocation. [[code](https://github.com/pytorch/pytorch/blob/c93ceef6585e140c1a33744374f76b2341bfe465/c10/cuda/CUDACachingAllocator.cpp#L1722)]
    2. `c10::cuda::CUDACachingAllocator::raw_alloc()/raw_delete()` [[code](https://github.com/pytorch/pytorch/blob/c93ceef6585e140c1a33744374f76b2341bfe465/c10/cuda/CUDACachingAllocator.cpp#L1874)]. It uses `THCCachingAllocator caching_allocator` directly and requires being freed via `raw_delete` manually.

cc @nv-dlasalle

nv-dlasalle (approved):

tensoradapter/pytorch/torch.cpp
```diff
@@ -47,6 +50,16 @@ TA_EXPORTS DLManagedTensor* TAempty(
   return at::toDLPack(tensor);
 }
 
+#ifdef DGL_USE_CUDA
+TA_EXPORTS void* RawAlloc(size_t nbytes) {
+  return c10::cuda::CUDACachingAllocator::raw_alloc(nbytes);
```
BarclayII, line 55: Can we just use `torch::empty`?

yaox12: Then we have to change the behavior of `AllocWorkspace` because we have to save the tensor to reuse its memory. We either have to:
1. Change all codes using `AllocWorkspace`.
2. Save the allocated tensors somewhere and return `tensor.data_prt` as raw pointers. But when deleting, we have to search for the tensor that matches the pointer to be freed.

BarclayII: You are right.

BarclayII (approved):

shintarok111:
I encountered the same issue as in #3933. Has the problem been resolved?

PR #4213, opened 2022-07-05
[Unittest][Fix] Several unit tests fixes for Ampere+ and PyTorch 1.12+

yaox12:
## Description

Three fixes:

### 1. Fix `test_csrmm` for Ampere+
Starting in PyTorch 1.7, there is a new flag called `allow_tf32`. This flag defaults to `True` in PyTorch 1.7 to PyTorch 1.11, and `False` in PyTorch 1.12 and later. This flag controls whether PyTorch is allowed to use the TensorFloat32 (TF32) tensor cores, available on new NVIDIA GPUs since Ampere, internally to compute matmul (matrix multiplies and batched matrix multiplies) and convolutions.

This flag will break `test_csrmm` on Ampere due to precision issues, so I create this PR to set `allow_tf32 = False`. Besides, for PyTorch > 1.12.0 (e.g., current TOT), we have to ensure `TORCH_ALLOW_TF32_CUBLAS_OVERRIDE` is not set.

### 2. Fix `test_unified_tensor` for PyTorch > 1.12.0
Recently, `cpu_tensor[cuda_indices]` is no longer allowed in PyTorch, see https://github.com/pytorch/pytorch/pull/69607#issuecomment-1035443522. This will break `test_unified_tensor.py`. I just fix it.

### 3. Fix `test_sparse` for PyTorch 1.12+
fp16 support for CPU linalg functions has been removed. See https://github.com/pytorch/pytorch/pull/75647. Skip it for the CPU test.

## Checklist


yaox12:
Still fails on the A100 CI runner but works well in my offline test on A100. Looking into it.

tests/compute/test_csrmm.py
```diff
@@ -5,6 +5,10 @@
 from test_utils import parametrize_idtype
 import backend as F
 
+if F.backend_name == 'pytorch':
+    import torch
+    torch.backends.cuda.matmul.allow_tf32 = False
```
ptrblck, line 10: You may want to use the newly added:
```
torch.set_float32_matmul_precision(“highest”)
```
setting in `1.12.0`.
See: https://github.com/pytorch/pytorch/pull/76509 and https://pytorch.org/blog/pytorch-1.12-released/#changes-to-float32-matrix-multiplication-precision-on-ampere-and-later-cuda-hardware

yaox12: Yeah. But DGL supports PyTorch 1.9+, so I use `torch.backends.cuda.matmul.allow_tf32` for a better compatibility. I think they are equivalent as per https://pytorch.org/docs/stable/generated/torch.set_float32_matmul_precision.html.
Another concern is that, in Nvidia's PyTorch container 22.05, the PyTorch version is recognized as 1.12.0, but `torch.set_float32_matmul_precision` is not available...

ptrblck: OK, makes sense due to BC-breaking behavior. `22.05` uses [1.12.0a0+8a1a93a](https://github.com/pytorch/pytorch/commit/8a1a93a), which is not the stable `1.12.0` release, but a previous commit (note the alpha in the commit),

BarclayII (approved):

PR #4217, opened 2022-07-06
Support new format for multi-file support in distributed partitioning.

kylasa:
Code changes added for this PR includes the following items below:

1. Updated multiple-file-format support. As part of changes introduced in the metadata json we now read nodes/edges/node-features as determined by the metadata json contents. As part of the optimizations, we now generate node information on the fly and they are directly generated on the process which owns the nodes. 
2. For reading large text files, we are now using pyarrow instead of numpy. 

## Checklist
- [X] Code is well-documented


zheng-da:
can you use a better name for the PR? and give better description for the PR?

tools/distpartitioning/convert_partition.py
```diff
@@ -54,14 +54,21 @@ def create_dgl_object(graph_name, num_parts, \
...
-    global_eid_ranges = schema['eid']
-    global_nid_ranges = {key: np.array(global_nid_ranges[key]).reshape(
-        1, 2) for key in global_nid_ranges}
-    global_eid_ranges = {key: np.array(global_eid_ranges[key]).reshape(
-        1, 2) for key in global_eid_ranges}
+    node_info = schema["nid"]
+    offset = 0
+    global_nid_ranges = {}
+    for k, v in node_info.items():
+        global_nid_ranges[k] = np.array([offset + int(v["data"][0][1]), offset + int(v["data"][-1][2])]).reshape(1,2)
```
zheng-da, line 62: i don't understand this. `schema` is changed? if so, what is `schema` now?

tools/distpartitioning/data_proc_pipeline.py
```diff
@@ -44,38 +34,15 @@ def log_params(params):
                      help='The schema of the graph')
     parser.add_argument('--num-parts', required=True, type=int,
                      help='The number of partitions')
-    parser.add_argument('--num-node-weights', required=True, type=int,
-                     help='The number of node weights used by METIS.')
-    parser.add_argument('--workspace', type=str, default='/tmp',
-                    help='The directory to store the intermediate results')
```
zheng-da, line 50: you don't need this any more?

tools/distpartitioning/data_proc_pipeline.py
```diff
@@ -44,38 +34,15 @@ def log_params(params):
...
     parser.add_argument('--world-size', help='no. of processes to spawn',
                     default=1, type=int, required=True)
-    parser.add_argument('--nodes-file', help='filename of the nodes metadata', 
-                    default=None, type=str, required=True)
-    parser.add_argument('--edges-file', help='filename of the nodes metadata', 
-                    default=None, type=str, required=True)
-    parser.add_argument('--node-feats-file', help='filename of the nodes features', 
-                    default=None, type=str, required=True)
-    parser.add_argument('--edge-feats-file', help='filename of the nodes metadata', 
-                    default=None, type=str )
```
zheng-da, line 72: why are `--edges-file`, `--node-feats-file` and `--edge-feats-file` no longer needed?

tools/distpartitioning/utils.py
```diff
@@ -376,6 +375,7 @@ def write_dgl_objects(graph_obj, node_features, edge_features, output_dir, part_
 
     part_dir = output_dir + '/part' + str(part_id)
     os.makedirs(part_dir, exist_ok=True)
+    #write_graph_dgl(os.path.join(part_dir ,'part'+str(part_id)), graph_obj)
```
zheng-da, line 378: remove this.

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,88 @@
...
+    node_part_ids : 
+        numpy array, whose length is same as no. of nodes in the graph. Index in this array is the global_nid
+        and value is the partition-id which owns the node
+    ntid_cnt_map : 
+        a dictionary where keys are node_type ids and values are total nodes which are of that node_type
+    ntid_ntype_map : 
+        a dictionary where keys are node_type ids and values are node_type names
+    nid_schema_map: 
+        a dictionary, which is extracted from the input graph metadata json file for node information, 
+        key is `nid`
```
zheng-da, line 44: key is nid? what is the value?

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,88 @@
...
+    '''
+    ntypes_nid_map -> ntype_name -> str(ntype_id)
+    ntypes_id_count -> str(ntype_id) -> total node count
+    '''
+
+    local_node_data = {}
+    gnid_start = 0
+    gnid_end = 0
+    for ntypeid in range(len(ntid_cnt_map)):
+        ntype_name = ntid_ntype_map[str(ntypeid)]
```
zheng-da, line 62: why is node type id a string?

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,88 @@
...
+    '''
+    '''
+    ntypes_nid_map -> ntype_name -> str(ntype_id)
+    ntypes_id_count -> str(ntype_id) -> total node count
+    '''
+
+    local_node_data = {}
+    gnid_start = 0
+    gnid_end = 0
+    for ntypeid in range(len(ntid_cnt_map)):
```
zheng-da, line 61: `ntid_cnt_map` is only used to define the number of node types? you can get it from others such as `ntid_ntype_map`

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,88 @@
...
+        ntype_name = ntid_ntype_map[str(ntypeid)]
+        ntype_info = nid_schema_map[ntype_name]
```
zheng-da, line 63: the value needs to be explained.

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,88 @@
...
+        rank of the process
     world_size : int
-        total no. of participating processes
-    node_data : dictionary
-        nodes data dictionary with keys as column names and values as
-        columns from the nodes csv file
+        total no. of processes
+    node_part_ids : 
+        numpy array, whose length is same as no. of nodes in the graph. Index in this array is the global_nid
+        and value is the partition-id which owns the node
```
zheng-da, line 37: if length of the array is he number of nodes in a graph, it can be very large. does each process maintain this array?

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,88 @@
...
+        ntype_info = nid_schema_map[ntype_name]
+        type_start = int(ntype_info["data"][0][1])
+        type_end = int(ntype_info["data"][-1][2])
+
+        gnid_end += type_end
+
+        node_partid_slice = node_part_ids[gnid_start:gnid_end]
+        cond = node_partid_slice == rank
+        own_gnids = np.arange(gnid_start, gnid_end, dtype=np.int64)
+        own_gnids = own_gnids[cond == 1]
```
zheng-da, line 72: you can just use `cond`. if you use `cond == 1`, why not just use `node_partid_slice == rank`

zheng-da, line 72: you can just use `cond`. if you use `cond == 1`, why not just use `node_partid_slice == rank`

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,88 @@
...
+
+        own_tnids = np.arange(type_start, type_end, dtype=np.int64)
+        own_tnids = own_tnids[cond == 1]
+        if constants.NTYPE_ID in local_node_data:
+            local_node_data[constants.NTYPE_ID] = np.concatenate([local_node_data[constants.NTYPE_ID],
+                                                                    np.ones(own_gnids.shape, dtype=np.int64)*ntypeid])
+        else:
+            local_node_data[constants.NTYPE_ID] = np.ones(own_gnids.shape, dtype=np.int64)*ntypeid
```
zheng-da, line 80: why don't you do `local_node_data[constants.NTYPE_ID].append(np.ones(own_gnids.shape, dtype=np.int64)*ntypeid]))` and concatenate everything in the end?

zheng-da: In this way, the code is more concise and also more efficient

tools/distpartitioning/data_shuffle.py
```diff
@@ -155,9 +161,7 @@ def exchange_edge_data(rank, world_size, edge_data):
     edge_data.pop(constants.OWNER_PROCESS)
     return edge_data
 
-
-def exchange_node_features(rank, world_size, node_data, node_features, ntypes_map, \
-        ntypes_nid_map, ntype_id_count, node_part_ids):
+def exchange_node_features(rank, world_size, node_feature_tids, ntype_gnid_map, node_part_ids, node_features):
```
zheng-da, line 164: why this PR requires to change this file? our input format change shouldn't affect this function.

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,88 @@
...
 from dataset_utils import get_dataset
 from utils import read_partitions_file, read_json, get_node_types, \
-                    augment_node_data, augment_edge_data, get_ntypes_map, \
+                    augment_edge_data, get_ntypes_map, \
                     write_dgl_objects, write_metadata_json
 from gloo_wrapper import alltoall_cpu_object_lst, alltoallv_cpu, \
                     alltoall_cpu, allgather_sizes, gather_metadata_json
 from globalids import assign_shuffle_global_nids_nodes, \
                     assign_shuffle_global_nids_edges, get_shuffle_global_nids_edges
 from convert_partition import create_dgl_object, create_metadata_json, validateDGLObjects
```
BarclayII, line 21: `validateDGLObjects` is not in `convert_partition.py` and it's not being used anywhere.

tools/distpartitioning/dataset_utils.py
```diff
@@ -33,46 +33,63 @@ def get_dataset(input_dir, graph_name, rank, num_node_weights):
...
-            num_feats = tokens[2]
-            for idx in range(int(num_feats)):
-                feat_file = s +'/node-feat-'+'{:02d}'.format(idx) +'/'+ str(rank)+'.npy'
-                if (os.path.exists(input_dir+'/'+feat_file)):
-                    features = np.load(input_dir+'/'+feat_file)
-                    node_features[ntype+'/feat'] = torch.tensor(features)
+    #iterate over the "node_data" dictionary in the schema_map
+    #read the node features if exists
+    #also keep track of the type_nids for which the node_features are read.
+    dataset_features = schema_map["node_data"]
```
BarclayII, line 41: Same problem.

tools/distpartitioning/convert_partition.py
```diff
@@ -54,14 +54,21 @@ def create_dgl_object(graph_name, num_parts, \
...
     """
-
     #create auxiliary data structures from the schema object
-    global_nid_ranges = schema['nid']
-    global_eid_ranges = schema['eid']
-    global_nid_ranges = {key: np.array(global_nid_ranges[key]).reshape(
-        1, 2) for key in global_nid_ranges}
-    global_eid_ranges = {key: np.array(global_eid_ranges[key]).reshape(
-        1, 2) for key in global_eid_ranges}
+    node_info = schema["nid"]
```
zheng-da, line 58: now `schema` is different and becomes more complex. it should be explained in the comments.

tools/distpartitioning/data_shuffle.py
zheng-da, line 164: you changed the input arguments. you need to update the doc string accordingly.

tools/distpartitioning/data_shuffle.py
```diff
@@ -199,83 +203,77 @@ def exchange_node_features(rank, world_size, node_data, node_features, ntypes_ma
...
-                out_type_nid = type_nid[idx == 1]
-                out_features = node_features[ntype_name+'/feat'][out_type_nid]
-                send_node_features[ntype_name+'/feat'] = out_features
-                send_global_nids[ntype_name+'/feat'] = out_global_nid
-
-        node_features_rank_lst.append(send_node_features)
-        global_nid_rank_lst.append(send_global_nids)
+    own_node_features = {}
+    own_global_nids = {}
+    for ntype_name, ntype_info in node_feature_tids.items():
```
zheng-da, line 213: again, write comments to explain the type info.

tools/distpartitioning/data_shuffle.py
```diff
@@ -199,83 +203,77 @@ def exchange_node_features(rank, world_size, node_data, node_features, ntypes_ma
...
+            gnid_end = begin_global_nid + type_nid_end
+
+            #type_nids for this feature subset on the current rank
+            gnids_feat = np.arange(gnid_start, gnid_end)
+            tnids_feat = np.arange(type_nid_start, type_nid_end)
+            local_idx = np.arange(0, type_nid_end - type_nid_start)
+
+            #check if node features exist for this ntype_name + feat_name
+            #this check should always pass, because node_feature_tids are built
+            #by reading the input metadata json file for existing node features.
```
zheng-da, line 235: where is the check?

tools/distpartitioning/data_shuffle.py
```diff
@@ -199,83 +203,77 @@ def exchange_node_features(rank, world_size, node_data, node_features, ntypes_ma
...
+            if(feat_key in node_features):
```
zheng-da, line 236: should we use assert instead of using if? when it fails?

tools/distpartitioning/convert_partition.py
kylasa: The structure of the schema is explained in detail in the function header section for this function, as we have discussed over the chime call.

tools/distpartitioning/data_proc_pipeline.py
kylasa: The unused parameters are eliminated. Some of these parameters where used when single-file-format was supported. But today we plan to only support multiple machines using multiple-file-format dataset. This was also discussed in our chime call as well.

tools/distpartitioning/data_proc_pipeline.py
kylasa: These are no longer needed with the multiple-file-format support. In this new format, all the information is present in the metadata json file.

tools/distpartitioning/data_shuffle.py
kylasa: "validateDGLObjects" function is removed. This was added only for debugging purposes and should never be released to the general users of this library. Removed this function name from the imports section of this file.

tools/distpartitioning/data_shuffle.py
kylasa: In this iteration of implmentation, we maintain this array of node-id to partition-id which can be used if there is any need. This will change in the next iterations of implementation, where we intend to distributed the node <--> partition across multiple ranks.

tools/distpartitioning/data_shuffle.py
kylasa: nid is the short-form of node-id. Any prefix can be attached to it indicating type of node it we are referring to in this context. Mentioned an example to indicate the structure of the dictionary which will be the value for the "nid" in the schema object.

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,88 @@
...
-        the input argument, node_data, is updated with the node data received by other processes
-        in the world.
-    """
+        dictionary where keys are column names and values are numpy arrays, these arrays are generated by 
+        using information present in the metadata json file
+
+    '''
+    '''
+    ntypes_nid_map -> ntype_name -> str(ntype_id)
+    ntypes_id_count -> str(ntype_id) -> total node count
```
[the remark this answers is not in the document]
kylasa: Removed these two lines... they were NOT supposed to be in the function header.

tools/distpartitioning/data_shuffle.py
kylasa: Agreed. Removed all instances of "ntid_cnt_map" from this file. Since this map is no longer used, I have removed it from this file and this map will never be constructed from now on.

tools/distpartitioning/data_shuffle.py
kylasa: This is how it is stored in the dictionaries. In the json object, for an object to be serializable it is converted into string form. This is the reason, all the keys which are integers are in string forms when json object is created.

tools/distpartitioning/data_shuffle.py
kylasa: Included text in the header of the function which discusses about the structure of the dictionary in detail.

tools/distpartitioning/data_shuffle.py
kylasa: Agreed. changed the code accordingly. Now we use array[condition] to filter the data needed for further processing.

kylasa: Agreed. changed the code accordingly. Now we use array[condition] to filter the data needed for further processing.

tools/distpartitioning/data_shuffle.py
kylasa: Agreed. Changed the code accordingly. Now we concatenate in a single shot after all the data is collected.

tools/distpartitioning/data_shuffle.py
kylasa: Updated the doc-string to match with the updated inpur arguments to this function as suggested.

tools/distpartitioning/data_shuffle.py
kylasa: Added detailed text to discuss the structure of the key, value pairs which are iterated upon when exchanging node features.

tools/distpartitioning/data_shuffle.py
kylasa: Agreed. changed the if-statement to assert, per suggestion.

tools/distpartitioning/data_shuffle.py
```diff
@@ -359,70 +357,35 @@ def read_dataset(rank, world_size, node_part_ids, params):
     dictionary
         edge features which is also a dictionary, similar to node features dictionary
     """
-
     edge_features = {}
-    node_data, node_features, edge_data = \
-        get_dataset(params.input_dir, params.graph_name, rank, params.num_node_weights)
+    node_tids, node_features, node_feat_tids, edge_data, edge_tids = \
+        get_dataset(params.input_dir, params.graph_name, rank, world_size, schema_map)
```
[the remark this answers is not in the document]
kylasa: Updated the signature of the get_dataset function, in the dataset_utils.py, to accept schema_map as an argument. Also updated the signature of the read_dataset function in this file to accept schema_map function as an argument. While merging this code-change might have been missed out. Thank you so much for pointing this one.

tools/distpartitioning/dataset_utils.py
kylasa: Changed the code appropriately so that the function invocation and function signature match w.r.t to the argument list.

tools/distpartitioning/utils.py
kylasa: Removed the commented out code.

tools/distpartitioning/convert_partition.py
```diff
@@ -18,6 +18,36 @@ def create_dgl_object(graph_name, num_parts, \
     This function creates dgl objects for a given graph partition, as in function
     arguments. 
 
+    The input schema object is structure as follows:
```
zheng-da, line 21: The `schema` argument is a dict, which contains the metadata related to node IDs and edge IDs. It contains two keys: "nid" and "eid", whose value has the following format.

tools/distpartitioning/convert_partition.py
```diff
@@ -18,6 +18,36 @@ def create_dgl_object(graph_name, num_parts, \
...
+            "data" : [
+                    [ <path-to-file>/ntype0-name-0, start_id, end_id], 
+                    [ <path-to-file>/ntype0-name-1 start_id, end_id)
+            ]
+        }
+
+    The "nid" value is a dictionary, which described the type_nid that
+    should be assigned to nodes which are read from the corresponding nodes file. 
+
+    Similarly, compared to "nid" dictionary, "eid" dictionary is structure as well.
```
zheng-da, line 43: can you make the description more precise?

zheng-da: eid dictionary is a structure similar to the nid dict.

tools/distpartitioning/convert_partition.py
```diff
@@ -18,6 +18,36 @@ def create_dgl_object(graph_name, num_parts, \
     This function creates dgl objects for a given graph partition, as in function
     arguments. 
 
+    The input schema object is structure as follows: 
+    1. "nid" is a dictionary, which has the following structure.
```
zheng-da, line 22: The key value pairs in the "nid" dict has the following format.

tools/distpartitioning/convert_partition.py
```diff
@@ -18,6 +18,36 @@ def create_dgl_object(graph_name, num_parts, \
     This function creates dgl objects for a given graph partition, as in function
     arguments. 
 
+    The input schema object is structure as follows: 
+    1. "nid" is a dictionary, which has the following structure. 
+        "ntype-name" : {
+            "format" : "csv", 
+            "data" : [
+                    [ <path-to-file>/ntype0-name-0, start_id, end_id], 
+                    [ <path-to-file>/ntype0-name-1 start_id, end_id)
```
zheng-da, line 27: can you explain what are "ntype0-name-0" and "ntype0-name-1"?

zheng-da: I'm confused. i checked the doc and it says each of the files defines the metadata file for a partition. why are there only two files? if there should be p files, i think you should illustrate it with p partitions.

zheng-da: however, your code below suggests that there are two files for each node type.

tools/distpartitioning/data_shuffle.py
zheng-da: can you add a TODO here? This needs to be changed right after this release.

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,109 @@
...
-        columns from the nodes csv file
+        total no. of processes
+    node_part_ids : 
+        numpy array, whose length is same as no. of nodes in the graph. Index in this array is the global_nid
+        and value is the partition-id which owns the node
+    ntid_ntype_map : 
+        a dictionary where keys are node_type ids and values are node_type names
+    nid_schema_map: 
+        a dictionary, which is extracted from the input graph metadata json file for node information, 
+        key is `nid`, which is a string. And value is a dictionary which has information about all the node types.
```
zheng-da, line 42: the map has only one key-value pair? if so, why not pass the value to the function?

tools/distpartitioning/data_shuffle.py
zheng-da: why do we store node type IDs in the JSON file?

tools/distpartitioning/data_shuffle.py
zheng-da: from the code below, it seems to me that the key are node types instead of "nid". please change the statement here accordingly.

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,109 @@
...
+        own_tnids = own_tnids[cond]
+        if constants.NTYPE_ID in local_node_data:
+            local_node_data[constants.NTYPE_ID].append(np.ones(own_gnids.shape, dtype=np.int64)*ntypeid)
         else:
-            input_list.append(torch.from_numpy(filt_data))
-            send_sizes.append(torch.tensor(filt_data.shape, dtype=torch.int64))
-        recv_sizes.append(torch.zeros((2,), dtype=torch.int64))
-    end = timer()
-    print('[Rank: ', rank, '] Preparing node_data to send out: ', timedelta(seconds=end - start))
+            local_node_data[constants.NTYPE_ID] = [np.ones(own_gnids.shape, dtype=np.int64)*ntypeid]
```
zheng-da, line 101: to make code even simpler, you can create lists for each constant when `local_node_data` is created.
```
local_node_data = {
    constants.NTYPE_ID: [],
    ...
}
```

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,109 @@
...
 
-    output_list = []
-    for s in recv_sizes: 
-        output_list.append(torch.zeros(s.tolist(), dtype=torch.int64))
-    
-    dist.barrier()
-    alltoallv_cpu(rank, world_size, output_list, input_list)
-    end = timer()
-    print('[Rank: ', rank, '] Time to exchange node data : ', timedelta(seconds=end - start))
+    local_node_data[constants.OWNER_PROCESS] = np.ones(local_node_data[constants.GLOBAL_NID].shape)*rank
```
zheng-da, line 115: why do we need to create this?

tools/distpartitioning/data_shuffle.py
```diff
@@ -179,103 +204,106 @@ def exchange_node_features(rank, world_size, node_data, node_features, ntypes_ma
...
+            output_nid_list = alltoall_cpu_object_lst(rank, world_size, global_nid_per_rank)
+            output_nid_list[rank] = global_nid_per_rank[rank]
+
+            #stitch node_features together to form one large feature tensor
+            for idx, x in enumerate(output_feat_list):
+                if feat_key in own_node_features:
+                    t = own_node_features[feat_key]
+                    own_node_features[feat_key] = torch.cat([t, x[feat_key]])
+                    t = own_global_nids[feat_key]
+                    own_global_nids[feat_key] = np.concatenate([t, output_nid_list[idx][feat_key]])
```
zheng-da, line 296: the same comments as above. the code here should be simplified.
we should keep all tensors in the list and call concatenate once, instead of calling it in every iteration.

tools/distpartitioning/data_shuffle.py
```diff
@@ -179,103 +204,106 @@ def exchange_node_features(rank, world_size, node_data, node_features, ntypes_ma
...
     end = timer()
     print('[Rank: ', rank, '] Total time for node feature exchange: ', timedelta(seconds = end - start))
+    return own_node_features, own_global_nids
 
-    return rcvd_node_features, rcvd_global_nids
-
-def exchange_graph_data(rank, world_size, node_data, node_features, edge_data,
-        node_part_ids, ntypes_map, ntypes_nid_map, ntype_id_count):
+def exchange_graph_data(rank, world_size, node_features, node_feat_tids, edge_data,
+        node_part_ids, ntypes_map, ntypes_gnid_range_map, ntid_ntype_map, node_tids, schema_map):
```
zheng-da, line 306: update the doc string.

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,109 @@
...
+            "nid" : { #m : no. of node types
+                "ntype0-name": {
+                "format": "csv",
+                "data" : [ #list of lists
+                            ["<path>/ntype0-name-0.txt", 0, id_end0], # These are type_nids for the nodes
+                            ["<path>/ntype0-name-1.txt", id_start1, id_end1],
+                            ...,
+                            ["<path>/ntype0-name-<p-1>.txt", id_start<p-1>, id_end<p-1>]
+                        ]
+                },
```
zheng-da, line 58: does the distributed partition script still needs to read the node metadata files? i thought we discussed this and the partition script doesn't need these files any more.

tools/distpartitioning/data_shuffle.py
```diff
@@ -359,70 +389,105 @@ def read_dataset(rank, world_size, node_part_ids, params):
...
-    Basically, each individual file is split into "p" files, where "p" is the no. of processes in the
-    world. Directory names are encoded strings which consist of prefix and suffix strings. Suffix strings
-    indicate the no. of items present inside that directory. For instance, "nodes-ntype0-2" directory has 
-    "2" node type sub-directories within it. And each feature file, whether it is node features file or edge
-    feature file, is split into "p" numpy files named as 0.npy, 1.npy, ..., <p-1>.npy. 
+    across multiple file format. 
+
+    Input dataset and its file structure is described in metadata json file which is also part of the 
+    input dataset. On a high-level, this metadata json file contains information about the following items
+    a) Nodes metadata, It is assumed that nodes which belong to each node-type are split into p files
```
zheng-da, line 409: do we still need the node metadata files?

tools/distpartitioning/data_shuffle.py
```diff
@@ -480,14 +547,16 @@ def gen_dist_partitions(rank, world_size, params):
...
-            global_nids = rcvd_global_nids[ntype_name+'/feat']
-
-            common, idx1, idx2 = np.intersect1d(node_data[constants.GLOBAL_NID], global_nids, return_indices=True)
-            shuffle_global_ids = node_data[constants.SHUFFLE_GLOBAL_NID][idx1]
-            feature_idx = shuffle_global_ids.argsort()
-            rcvd_node_features[ntype_name+'/feat'] = rcvd_node_features[ntype_name+'/feat'][feature_idx]
+    for ntype_name in ntypes:
+        featnames = get_ntype_featnames(ntype_name, schema_map)
+        for featname in featnames:
+            if (ntype_name+'/'+featname in rcvd_global_nids):
```
zheng-da, line 553: why a node feature doesn't exist?

zheng-da: if we know it exists for sure, we need to use assert instead of "if"

tools/distpartitioning/dataset_utils.py
```diff
@@ -3,7 +3,7 @@
 import constants
 import torch
 
-def get_dataset(input_dir, graph_name, rank, num_node_weights):
+def get_dataset(input_dir, graph_name, rank, world_size, schema_map):
```
zheng-da, line 6: update the doc string.

tools/distpartitioning/utils.py
```diff
@@ -47,6 +50,34 @@ def read_json(json_file):
...
+        metadata json object as a dictionary, which is read from the input
+        metadata file from the input dataset
+
+    Returns:
+    --------
+    list : 
+        a list of feature names for a given node_type
+    """
+    ntype_dict = schema["node_data"]
+    if (ntype_name in ntype_dict):
```
zheng-da, line 72: if node type should always exist, we should use asset to make sure it doesn't fail silently.

tools/distpartitioning/utils.py
```diff
@@ -64,37 +95,26 @@ def get_node_types(schema):
         dictionary with ntype <-> type_nid mappings
         list of ntype strings
     """
```
zheng-da, line 97: the docstring for the return values is out of date.

tools/distpartitioning/utils.py
```diff
@@ -64,37 +95,26 @@ def get_node_types(schema):
...
-    ntypes_map = {e: i for i, e in enumerate(ntypes)}
-
-    return ntypes_map, ntypes
+    ntype_info = schema["nid"]
+    ntypes = []
+    for k in ntype_info.keys(): 
+        ntypes.append(k)
+    ntype_ntypeid_map = {e: i for i, e in enumerate(ntypes)}
+    ntypeid_ntype_map = {str(i): e for i, e in enumerate(ntypes)}
+    return ntype_ntypeid_map, ntypes, ntypeid_ntype_map
```
zheng-da, line 104: why is the return values of `get_node_types` different from `get_edge_types`?

tools/distpartitioning/utils.py
```diff
@@ -108,24 +128,35 @@ def get_edge_types(schema):
...
+        three items: file-name, starting type_nid and ending type_nid
 
     Returns:
     --------
     dictionary : 
-        map between the node_types and global_id ranges for each node_type
+        a dictionary where keys are node_type names and values are global_nid range, which is a tuple.
     dictionary : 
-        map between the node_type and total node count for that type
+        a dictionary where kesy are node_type names and values are total count of nodes for this
```
zheng-da, line 149: you only return one value now.

tools/distpartitioning/utils.py
```diff
@@ -108,24 +128,35 @@ def get_edge_types(schema):
 
     return etypes_map, etypes
 
-def get_ntypes_map(schema): 
+def get_ntypes_map(node_tids):
```
zheng-da, line 131: this function name is confusing. this is actual to get the ID range for each node type.

tools/distpartitioning/utils.py
```diff
@@ -190,61 +221,28 @@ def augment_edge_data(edge_data, part_ids, id_offset):
         array of part_ids indexed by global_nid
     """
     #add global_nids to the node_data
-    global_eids = np.arange(id_offset, id_offset + len(edge_data[constants.GLOBAL_TYPE_EID]), dtype=np.int64)
+    #global_eids = np.arange(id_offset, id_offset + len(edge_data[constants.GLOBAL_TYPE_EID]), dtype=np.int64)
+    #edge_data[constants.GLOBAL_EID] = global_eids
```
zheng-da, line 225: remove commented code.

tools/distpartitioning/utils.py
```diff
@@ -190,61 +221,28 @@ def augment_edge_data(edge_data, part_ids, id_offset):
...
+    etype_offset = {}
+    offset = 0
+    for etype_name, tid_range in edge_tids.items(): 
+        assert int(tid_range[0][0]) == 0
+        assert len(tid_range) == world_size
+        etype_offset[etype_name] = offset + int(tid_range[0][0])
+        offset += int(tid_range[-1][1])
```
zheng-da, line 232: similarly, this is to convert the per-type edge ID ranges to global edge ID ranges.

zheng-da: why don't you create one function for this like you did for nodes?

tools/distpartitioning/utils.py
```diff
@@ -190,61 +221,28 @@ def augment_edge_data(edge_data, part_ids, id_offset):
...
+
+    global_eids = np.array([], dtype=np.int64)
+    for etype_name, tid_range in edge_tids.items(): 
+        global_eid_start = etype_offset[etype_name]
+        begin = global_eid_start + int(tid_range[rank][0])
+        end = global_eid_start + int(tid_range[rank][1])
+        global_eids = np.concatenate((global_eids, np.arange(begin, end)))
```
zheng-da, line 239: again, you should avoid concatenating in every iteration.

tools/distpartitioning/dataset_utils.py
```diff
@@ -33,46 +33,63 @@ def get_dataset(input_dir, graph_name, rank, num_node_weights):
...
-    #done build node_features locally. 
-    if len(node_features) <= 0: 
-        print('[Rank: ', rank, '] This dataset does not have any node features')
-    else: 
-        for k, v in node_features.items():
-            print('[Rank: ', rank, '] node feature name: ', k, ', feature data shape: ', v.size())
+            v = [feat_name, my_feat_data[1], my_feat_data[2]]
+            l = node_feature_tids[ntype_name]
+            l.append(v)
+            node_feature_tids[ntype_name] = l
```
zheng-da, line 57: the code above is very weird. how is it different from
```
node_feature_tids[ntype_name].append([feat_name, my_feat_data[1], my_feat_data[2]])
```

tools/distpartitioning/dataset_utils.py
```diff
@@ -33,46 +33,63 @@ def get_dataset(input_dir, graph_name, rank, num_node_weights):
...
-    nodes_datadict[constants.GLOBAL_TYPE_NID] = node_data[:,type_idx]
-    print('[Rank: ', rank, '] Done reading node_data: ', len(nodes_datadict), nodes_datadict[constants.NTYPE_ID].shape)
+    #read my nodes for each node type
+    node_tids = {}
+    node_data = schema_map["nid"]
+    for ntype_name, ntype_info in node_data.items():
+        v = []
+        node_file_info = ntype_info["data"]
+        for idx in range(len(node_file_info)):
+            v.append((node_file_info[idx][1], node_file_info[idx][2]))
```
zheng-da, line 66: here it still suggests users need to define node metadata files.

tools/distpartitioning/dataset_utils.py
```diff
@@ -33,46 +33,63 @@ def get_dataset(input_dir, graph_name, rank, num_node_weights):
...
     edge_datadict = {}
-    edge_file = input_dir+'/'+graph_name+'_edges'+'_{:02d}.txt'.format(rank)
-    edge_data = np.loadtxt(edge_file, delimiter=' ', dtype='int64')
-    edge_datadict[constants.GLOBAL_SRC_ID] = edge_data[:,0]
-    edge_datadict[constants.GLOBAL_DST_ID] = edge_data[:,1]
-    edge_datadict[constants.GLOBAL_TYPE_EID] = edge_data[:,2]
-    edge_datadict[constants.ETYPE_ID] = edge_data[:,3]
+    edge_data = schema_map["eid"]
+    for etype_name, etype_info in edge_data.items():
+        if(etype_info["format"] == "csv"):
```
zheng-da, line 74: if we only support csv, we should report errors if the input data isn't in the csv format.

tools/distpartitioning/dataset_utils.py
```diff
@@ -33,46 +33,63 @@ def get_dataset(input_dir, graph_name, rank, num_node_weights):
...
+            edge_info = etype_info["data"]
+            assert len(edge_info) == world_size
+
+            data = np.loadtxt(edge_info[rank][0], delimiter=' ', dtype=np.int64)
```
zheng-da, line 78: i thought you want to use pyarrow to load data now. why here it still uses numpy to load data? numpy is quite slow in loading csv.

tools/distpartitioning/dataset_utils.py
```diff
@@ -33,46 +33,63 @@ def get_dataset(input_dir, graph_name, rank, num_node_weights):
...
+            for idx, k in enumerate([constants.GLOBAL_SRC_ID, constants.GLOBAL_DST_ID, constants.GLOBAL_TYPE_EID, constants.ETYPE_ID]):
+                if k in edge_datadict:
+                    v = edge_datadict[k]
+                    v = np.concatenate((v, data[:,idx]))
```
zheng-da, line 82: again, don't use concatenation in each iteration. it makes code long and inefficient.

tools/distpartitioning/convert_partition.py
kylasa: modified the text accordingly.

tools/distpartitioning/convert_partition.py
kylasa: modified the text accordingly.

tools/distpartitioning/convert_partition.py
kylasa: Done.

tools/distpartitioning/convert_partition.py
kylasa: Done.

tools/distpartitioning/data_shuffle.py
kylasa: Done.

tools/distpartitioning/data_shuffle.py
kylasa: Done.

tools/distpartitioning/data_shuffle.py
kylasa: Yes. Agreed. The file names are not needed. But we still need the start_id and end_ids for each entry in the "data"'s value. These start and end ids are used by processes to generate the node ids on each process. Since this same metadata json file is also used, in an indirect way, for ParMETIS as well. So eliminating these nodes files from the metadata json file can be a little tricky one to begin with and needs to be handled very carefully.

tools/distpartitioning/data_shuffle.py
kylasa: Agreed. Now the code looks much cleaner and simpler to read. Thank you !!!

tools/distpartitioning/data_shuffle.py
kylasa: Changed the code appropriately. Will also search for this kind of pattern in the entire code base and change accordingly.

tools/distpartitioning/data_shuffle.py
kylasa: updated accordingly.

tools/distpartitioning/data_shuffle.py
kylasa: Agreed. Changed the code appropriately. This same logic was elsewhere as well and it was mentioned in one of the earlier code review comments.

tools/distpartitioning/dataset_utils.py
kylasa: Updated accordingly.

tools/distpartitioning/dataset_utils.py
kylasa: Agreed. The suggested code is exactly the same as the piece of code pointed out in this CI comment. Changed the code accordingly.

tools/distpartitioning/dataset_utils.py
kylasa: Yes. As I mentioned already removing "nid" information from the metadata is not that straight forward. To completely remove it we must first define a way to compute the global node-ids for the node features which exist in the dataset. If the dataset does not have node-features in it then "nid" key-value pair can be completely removed from the metadata json file.

tools/distpartitioning/dataset_utils.py
kylasa: OK. Now instead of silently proceeding with the execution we stop with the assertion failure if the format is something other than "csv".

tools/distpartitioning/dataset_utils.py
kylasa: This was lost during the change from old file format to new file format. Now changed code to use pyarrow when reading the edges from files.

tools/distpartitioning/dataset_utils.py
kylasa: With the previous comment, this code has been rendered outdated. This code has been updated in response to previous comment.

tools/distpartitioning/utils.py
kylasa: In this case ntype_name is NOT always guaranteed to exist in the node-feature portion of the metadata. So this code is needed for this particular case. Basically not all node types will have node feature data associated with them.

tools/distpartitioning/utils.py
kylasa: Updated accordingly.

tools/distpartitioning/utils.py
kylasa: Removed get_edge_types. Basically this is an unused function. and in the current iteration we do not support edge features.

tools/distpartitioning/utils.py
kylasa: changed the function name to "get_gnid_range_map" which is appropriate in this case. Here we take the node's tid ranges list and return their corresponding gnid range list.

tools/distpartitioning/utils.py
kylasa: changed accordingly.

tools/distpartitioning/utils.py
kylasa: Done.

tools/distpartitioning/utils.py
kylasa: This function no longer exists, because it was never used. So removed this entire function.

tools/distpartitioning/utils.py
kylasa: This function is removed, as this was an unused function.

tools/distpartitioning/data_shuffle.py
```diff
@@ -12,82 +12,113 @@
...
-    node_data : dictionary
-        nodes data dictionary with keys as column names and values as
-        columns from the nodes csv file
+        total no. of processes
+    node_part_ids : 
+        numpy array, whose length is same as no. of nodes in the graph. Index in this array is the global_nid
+        and value is the partition-id which owns the node
+        TODO: For large graphs, this mapping function can be memory intensive. This needs to be changed to 
+        processes owning a set of global-nids, per partitioning algorithm, and messaging will be used to 
+        identify the ownership instead of mem. lookups.
```
zheng-da, line 41: please don't add TODO in the docstrng like this. add the TODO as a comment inside the function.

tools/distpartitioning/dataset_utils.py
```diff
@@ -15,64 +18,90 @@ def get_dataset(input_dir, graph_name, rank, num_node_weights):
         graph name string
     rank : int
         rank of the current process
+    world_size : int
+        total number of process in the current execution
     num_node_weights : int
         integer indicating the no. of weights each node is attributed with
```
zheng-da, line 24: this argument doesn't exist any more.

tools/distpartitioning/data_shuffle.py
kylasa: Moved the TODO to the gen_dist_partitions function, where we read the nodeid<-> partition-ids for the first time in the pipeline.

tools/distpartitioning/dataset_utils.py
kylasa: Removed the 'num_node_weights' from the docstring. Added 'schema_map' to the docstring as it is used now in the multiple-file-format dataset.

Issue #4225, opened 2022-07-08
Can GPU support weighted graph neighbor sampling


xiaopqr:
## 🚀 Feature
Can GPU support weighted graph neighbor sampling

## Motivation
In OPPO, services are basically weighted graphs. Training GNN model is very slow, GPU utilization is not high, and UVA cannot be used because GPU does not support weighted graph sampling

## Pitch

I hope GPU can support weighted graph neighbor sampling.
I hope the weighted graph can use UVA.

yaox12:
CUDA weighted neighborhood sampling has been merged. 
But it cannot work with UVA for now, because the kernel requires the probability array to be in the GPU memory while DGL doesn't allow assigning a GPU edge feature to a CPU graph.
I will open another PR adding UVA support for the weighted neighbor sampler.

PR #4064, opened 2022-05-30
[Feature] Add CUDA Weighted Neighborhood Sampling

yaox12:
## Description

This PR adds CUDA weighted neighborhood sampling. The main algorithm implementation is contributed by @xiaopqr from OPPO in #4226.

### Without replacement:

Implement [Algorithm A-Res](https://en.wikipedia.org/wiki/Reservoir_sampling#Algorithm_A-Res) via segmented sort.

Performance of sampling on ogbn-products with `fanout=20`:

<table>
<thead>
  <tr>
    <th></th>
    <th colspan="2">V100</th>
    <th colspan="2">A100</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>batch size</td>
    <td>1000</td>
    <td>10000</td>
    <td>1000</td>
    <td>10000</td>
  </tr>
  <tr>
    <td>CPU</td>
    <td>1.474</td>
    <td>7.539</td>
    <td>1.157</td>
    <td>7.896</td>
  </tr>
  <tr>
    <td>CUDA</td>
    <td>0.644</td>
    <td>0.995</td>
    <td>0.424</td>
    <td>0.598</td>
  </tr>
</tbody>
</table>

- The reported numbers are Python API time in milliseconds, which is the average time of calling `g.sample_neighbors()` 10,000 times. 
- `OMP_NUM_THREADS=8` is set for all runs.
- CUDA 10.2 is used for V100 while CUDA 11.7 is used for A100 to test the minimal/maximum CUDA versions.
- This implementation requires `cub::DeviceSegmentedSort` from CUB 1.17.

### With replacement:

Implement inverse transform sampling, which is similar to Algorithm 1 in this [paper](https://arxiv.org/pdf/1611.00532.pdf).

Performance of sampling on ogbn-products with `fanout=20`:

<table>
<thead>
  <tr>
    <th></th>
    <th colspan="2">V100</th>
    <th colspan="2">A100</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>batch size</td>
    <td>1000</td>
    <td>10000</td>
    <td>1000</td>
    <td>10000</td>
  </tr>
  <tr>
    <td>CPU</td>
    <td>1.467</td>
    <td>7.161</td>
    <td>1.192</td>
    <td>7.678</td>
  </tr>
  <tr>
    <td>CUDA</td>
    <td>0.589</td>
    <td>0.598</td>
    <td>0.365</td>
    <td>0.370</td>
  </tr>
</tbody>
</table>

[...]

PR #4226, opened 2022-07-08
[Feature] GPU support weighted graph neighbor sampling #4225 

xiaopqr:
## Description

GPU support weighted graph neighbor sampling #4225
## Checklist
- [x] Code is well-documented

## Changes

[x] GPU support weighted graph neighbor sampling, tests


yaox12:
1. Do you have performance numbers compared to the CPU implementation?
2. Some unit tests need to enable and some docs need to update. In fact, I have a similar PR #4064, so if you don't mind, I'd like to merge them together.
3. Some comments on the with replacement implementation.

src/array/cuda/rowwise_sampling_prob.cu
```diff
@@ -0,0 +1,639 @@
...
+            }
+        }
+        __syncthreads();
+        
+        for (int64_t idx = threadIdx.x; idx < num_picks; idx += CTA_SIZE) {
+            // get random value
+            FloatType sum = cdf[cdf_row_start + deg - 1]; 
+            FloatType rand = static_cast<FloatType>(curand_uniform(&rng) * sum);
+            // get the offset of the first value within cdf array which compares greater than random value. 
+            int64_t item = cub::UpperBound<FloatType* , int64_t, FloatType>(&cdf[cdf_row_start], deg, rand);
```
yaox12, line 351: I think it should be `LowerBound` here.

src/array/cuda/rowwise_sampling_prob.cu
```diff
@@ -0,0 +1,639 @@
...
+            if (idx < deg) {
+              cdf[cdf_row_start + idx] = thread_data;
+            }
+        }
+        __syncthreads();
+        
+        for (int64_t idx = threadIdx.x; idx < num_picks; idx += CTA_SIZE) {
+            // get random value
+            FloatType sum = cdf[cdf_row_start + deg - 1]; 
+            FloatType rand = static_cast<FloatType>(curand_uniform(&rng) * sum);
```
yaox12, line 349: Nice code to avoid normalizing the CDF array.

src/array/cuda/rowwise_sampling_prob.cu
yaox12: I'm also wondering if we can use shared memory to make this faster. To make it, we have to use a dynamic shared memory size at `max(deg) + sizeof(temp_storage)`.

yaox12:
I have ported [rowwise_sampling_prob.cu](https://github.com/dmlc/dgl/pull/4226/files#diff-ff0d0290f0ef1e94b075bfd39b48755220c2f3a464cd72c7fb4f2095a414a065) to PR #4064 with some code style and lint fixes. Following updates and reviews will be in PR #4064.

jermainewang:
Hi @xiaopqr , thanks for the PR. To speed up the process, could you briefly describe what kind of algorithms are you using for GPU weighted sampling? I'd like to understand:
(1) How does it parallelize the sampling?
(2) How does it utilize shared memory?
(3) Is there any extra data structure to be maintained? If so, what's the creation/destruction cost of it?
(4) What are the alternative algorithms? And what's the pros & cons for your choice?

xiaopqr:
1. I use the A-Res algorithm for weighted sampling without replacement and the CDF algorithm for weighted sampling with replacement.
2. I am not using shared memory. According to Yao Xin's test, sampling with replacement is fast without using shared memory, but sampling without replacement is not particularly fast compared to CPU due to sorting. Compared to CPU sampling, its value lies in the ability to use multiple GPUs to improve training  speed. @yaox12 

<html>
<body>


  |CUDA(ms)|CPU (ms)| Speed Up
-- | -- | -- | --
Without Replacement | 2.629 | 3.42| 1.3x
With Replacement | 0.253 | 3.348| 13.233x


</body>
</html>

3.Apply for temporary storage of the total degree size of all nodes that need to be sampled in the video memory. I haven't tested the time consuming to apply and destroy.
4.With replacement sampling, the Sum Tree algorithm can also be used, which is not much different from the time complexity of CDF, but CDF can be easily implemented using the cub library. No replacement sampling except A-Res, I have not found a better algorithm.

src/array/cuda/rowwise_sampling_prob.cu
xiaopqr: Because InclusiveSum is used above, UpperBound is used here.
I'm not sure if shared memory will speed up, can try something.

src/array/cuda/rowwise_sampling_prob.cu
```diff
@@ -0,0 +1,639 @@
...
+      slice_rows,
+      in_ptr,
+      in_cols,
+      data, 
+      out_ptr,
+      out_rows,
+      out_cols,
+      out_idxs, 
+      temp_ptr,
+      sort_temp_idxs
```
uiemUI, line 606: It may be appropriate to use sort_values.Current() instead of sort_temp_idxs,and call   device->FreeWorkspace(ctx, sort_temp) after the kernel.

yaox12: You're right.

xiaopqr: Thank you for pointing out this error, which has been fixed in #4064 pr

yaox12:
Close this PR due to #4064 has been merged. :)

PR #4238, opened 2022-07-10
CI Test for distributed graph partitioning pipeline

kylasa:
## Description

Added CI_MODE for the distributed graph partitioning pipeline. A command-line parameter 'ci_mode' is used  for this purposes. If this mode is specified, which should only be used for CI purposes the steps are executed in conjunction with the existing pipeline functionality
    1. node-id to partition-id mappings are generated randomly. However the seed is set to `0`.
       This ensure that same node-id to partition-id mappings are used by both the ranks.
    2. dgl objects and metadata json objects are created, but are not written into files.
    3. 'verify_graph_partitions'  function, which is executed only on rank-0, takes the metadata and partitioned graph
       object generated on rank-0 and runs following sanity checks on the this graph.
        a. node types in the partitioned graph match with the original graph
        b. edge types in the partitioned graph match with the original graph
        c. node features in the partitioned graph match with the original graph.

## Checklist
- [X] Code is well-documented


tools/distpartitioning/constants.py
```diff
@@ -18,3 +18,70 @@
 OWNER_PROCESS = "owner_proc_id"
 
 PART_LOCAL_NID = "part_local_nid"
+
+#CI-Test related definitions below this line
+CI_GRAPH_NUM_NODES = 40
```
zheng-da, line 23: CI code shouldn't be separated from the partitioning code.

tools/distpartitioning/dataset_utils.py
```diff
@@ -3,7 +3,102 @@
 import constants
 import torch
 
-def get_dataset(input_dir, graph_name, rank, num_node_weights):
+def get_ci_nodefeats(rank, feat_name):
```
zheng-da, line 6: the same comment here.

tools/distpartitioning/dataset_utils.py
```diff
@@ -3,7 +3,102 @@
...
+        this string is used as feature name 
+
+    Returns:
+    --------
+    numpy ndarray : 
+        an array of pre-defined dimensions with predetermined data, based on 
+        the rank of the process
+    '''
+    if (feat_name == "nodes-feat-01"):
+        return np.ones((10,20), dtype=np.float)*(rank+1)
```
zheng-da, line 25: it's better to not use value 1 as node features. if you mess up the node IDs, you cannot tell from the node features. you can use either sequence numbers or random numbers to distinguish nodes and edges.

tools/distpartitioning/utils.py
```diff
@@ -24,12 +27,18 @@ def read_partitions_file(part_file):
     numpy array
         array of part_ids and the idx is the <global_nid>
     """
+    if(cimode):
```
zheng-da, line 30: don't mix the CI code with the actual code.

Issue #4234, opened 2022-07-08
[Bug] Multiple crashes due to a recent pytorch update


chang-l:
## 🐛 Bug

Recently, multiple example cases failed due to similar errors, which can be traced back to a recent update from pytorch  ([commit #7b3a0ff](https://github.com/pytorch/pytorch/commit/7b3a0ff87a91c98574ddfd98532062d78c49b2c7))

In short, now, pytorch does no longer support such scenario where tensor indices are in gpu, but tensor itself resides in cpu, i.e., if accessing `array[idx]`, both `array` and `idx` should be in the same device, or (`array` in gpu, `idx` in cpu) but NOT vise versa.

As a result, some example cases, especially with multi-gpu support, failed, moreover, from the core dump stack, it may also indicate our dataloader `/python/dgl/dataloading/dataloader.py` could be problematic under multi-process(gpu) context (`use_ddp=true`) after this pytorch update.

## To Reproduce
- Example 1:  node_classification.py in example/graphsage
```
python node_classification.py
```
Output
```
Traceback (most recent call last):
  File "node_classification.py", line 144, in <module>
    acc = layerwise_infer(device, g, dataset.test_idx.to(device), model, batch_size=4096)
  File "node_classification.py", line 76, in layerwise_infer
    pred = pred[nid]
RuntimeError: indices should be either on cpu or on the same device as the indexed tensor (cpu)

```
- Example 2: multi_gpu_node_classification.py in example/graphsage
```
python multi_gpu_node_classification.py
```
Output
```
Traceback (most recent call last):
  File "/workspace/examples/graphsage/multi_gpu_node_classification.py", line 183, in <module>
    mp.spawn(train, args=(n_procs, graph, num_classes, split_idx), nprocs=n_procs)
  File "/opt/conda/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 240, in spawn
    return start_processes(fn, args, nprocs, join, daemon, start_method='spawn')
  File "/opt/conda/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 198, in start_processes
    while not context.join():
  File "/opt/conda/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 160, in join
    raise ProcessRaisedException(msg, error_index, failed_process.pid)
torch.multiprocessing.spawn.ProcessRaisedException: 

-- Process 0 terminated with the following error:
Traceback (most recent call last):
[...]

PR #4242, opened 2022-07-11
[Bugfix] Accessing data from the indexes stored in same device

chang-l:
## Description

To address https://github.com/dmlc/dgl/issues/4234, this PR fixes the crashing example cases, rgcn and graphsage, due to recent [pytorch update](https://github.com/pytorch/pytorch/commit/7b3a0ff87a91c98574ddfd98532062d78c49b2c7)

## Checklist
- [x] Code is well-documented

## Changes

Move index to the same device as data.

### DIFF
--- a/examples/pytorch/graphsage/advanced/train_sampling_unsupervised.py
+++ b/examples/pytorch/graphsage/advanced/train_sampling_unsupervised.py
@@ -133,6 +133,7 @@ def run(proc_id, n_gpus, args, devices, data):
         # blocks.
         tic_step = time.time()
         for step, (input_nodes, pos_graph, neg_graph, blocks) in enumerate(dataloader):
+            input_nodes = input_nodes.to(nfeat.device)
             batch_inputs = nfeat[input_nodes].to(device)
             pos_graph = pos_graph.to(device)
             neg_graph = neg_graph.to(device)
--- a/examples/pytorch/graphsage/advanced/train_sampling_unsupervised.py
+++ b/examples/pytorch/graphsage/advanced/train_sampling_unsupervised.py
@@ -135,9 +135,7 @@ def run(proc_id, n_gpus, args, devices, data):
         for step, (input_nodes, pos_graph, neg_graph, blocks) in enumerate(dataloader):
             input_nodes = input_nodes.to(nfeat.device)
             batch_inputs = nfeat[input_nodes].to(device)
-            pos_graph = pos_graph.to(device)
-            neg_graph = neg_graph.to(device)
-            blocks = [block.int().to(device) for block in blocks]
+            blocks = [block.int() for block in blocks]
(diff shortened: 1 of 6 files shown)


examples/pytorch/graphsage/advanced/train_sampling_unsupervised.py
```diff
@@ -133,6 +133,7 @@ def run(proc_id, n_gpus, args, devices, data):
         # blocks.
         tic_step = time.time()
         for step, (input_nodes, pos_graph, neg_graph, blocks) in enumerate(dataloader):
+            input_nodes = input_nodes.to(nfeat.device)
```
yaox12, line 136: When features are on the CPU, `input_nodes` will first be copied to GPU in the dataloader and then copied back to the CPU for indexing. It's not caused by this PR and can be eliminated by unifying `--graph_device` and `--data_device` just like other examples. Perhaps we should take a note and fix it when we refactor this example in the future.

yaox12 (approved):

examples/pytorch/graphsage/advanced/train_sampling_unsupervised.py
chang-l: I generally agree. It does sound odds that copying `input_nodes` back and forth...  But I think it cannot be avoided by specifying graph/data devices. The `input_nodes` always belongs to the same `device` as dataloader and model (`--gpu`), which is not configured in the same way as nfeat (`--data-device`) and g (`--graph-device`).

examples/pytorch/graphsage/advanced/train_sampling_unsupervised.py
```diff
@@ -133,10 +133,9 @@ def run(proc_id, n_gpus, args, devices, data):
...
             batch_inputs = nfeat[input_nodes].to(device)
-            pos_graph = pos_graph.to(device)
-            neg_graph = neg_graph.to(device)
-            blocks = [block.int().to(device) for block in blocks]
+            blocks = [block.int() for block in blocks]
```
chang-l, line 138: Since `pos_graph`, `neg_graph`, `blocks` all reside in the same device as dataloader (L105), moving to `device` seems redundant.

examples/pytorch/graphsage/advanced/train_sampling_unsupervised.py
yaox12: With the new sampling pipeline, we can specify `prefetch_node_features` to the sampler and don't need `input_nodes = input_nodes.to(nfeat.device)` anymore.

chang-l: I see... Thanks Xin for noting this. I will keep it in mind :)

PR #4248, opened 2022-07-13
[Docs] fix ffi docs

Zhanghyi:
## Description
I'm following ffi docs to implement CAPI in dgl. However, without the line I added, it will report DGLArgs and DGLRetValue's definitions are not found.


jermainewang (approved):

jermainewang:
Nice catch. Thanks for the fix!

Issue #4078, opened 2022-06-02
Multiple Backwards Pass Error with Retain_Graph=True


smiret-intel:
## 🐛 Bug

In order to run DGL based GNNs on use-cases like the [OpenCatalyst project](https://github.com/Open-Catalyst-Project/ocp), multiple backwards passes are required, which are currently not supported in the latest DGL versions. OpenCatalyst is a major dataset for science based GNN applications and currently support only PyTorch Geometric, which has support for multiple backwards passes.
We believe that we have tracked this error to `sparse.py` and `tensor.py` in the backend folder for pytorch ([linked here](https://github.com/dmlc/dgl/tree/master/python/dgl/backend/pytorch)) where the context (`ctx`) gets set to `None` at various instances ([example](https://github.com/dmlc/dgl/blob/master/python/dgl/backend/pytorch/sparse.py#L150)). We have attached preliminary patches that appear to circumvent this issue, but cause issues with distributed data parallel training, which are necessary for effective training in OpenCatalyst.

Files for the backend-patch and the minimal test script referenced below are attached.
[dgl_bug_report_double_autograd.zip](https://github.com/dmlc/dgl/files/8827152/dgl_bug_report_double_autograd.zip)

## To Reproduce

1. Run the attached minimal test script (`autograd_tester.py` or `simple-test.py` inside the backend-patch directory):

Multiple backwards passes not possible giving the following error:

`Traceback (most recent call last):
  File "autograd_tester.py", line 83, in <module>
    train(g, model)
  File "autograd_tester.py", line 71, in train
    loss.backward()
  File "/opt/conda/lib/python3.8/site-packages/torch/_tensor.py", line 363, in backward
    torch.autograd.backward(self, gradient, retain_graph, create_graph, inputs=inputs)
  File "/opt/conda/lib/python3.8/site-packages/torch/autograd/__init__.py", line 173, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
  File "/opt/conda/lib/python3.8/site-packages/torch/autograd/function.py", line 253, in apply
    return user_fn(self, *args)
  File "/opt/conda/lib/python3.8/site-packages/torch/cuda/amp/autocast_mode.py", line 135, in decorate_bwd
    return bwd(*args, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/dgl/backend/pytorch/sparse.py", line 149, in backward
[...]

PR #3386, opened 2021-09-28
[BugFix] Avoid Memory Leak Issue in PyTorch Backend

chwan-rice:
## Description

This PR targets the issues #3365 and #3230 by referring #1060. I also confront a similar issue in my private project when the input bipartite graph keeps on changing. However, after dropping `ctx.backward_cache` during the backward propagation, there remains ignorable memory leak. I couldn't find out the source of the remaining leak.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [ ] To the best of my knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

Issue #3365, opened 2021-09-19
Memory leakage in `update_all` using heterogeneous graph


isratnisa:
## 🐛 Bug

Using the message passing API `update_all` on heterogeneous (HT) graph accumulates memory with each epoch and eventually runs out of memory. For example, with AM dataset in the RGCN-hetero model, memory consumption is increased by 1GB per epoch (collected by python's line_profiler). Example code snippet from  [PR: RGCN with new heterograph API](https://github.com/dmlc/dgl/pull/3025) :
```        
  for rel in g.canonical_etypes:  
       _, etype, _ = rel  
       g.edges[rel].data['h*w_r'] = th.matmul(g.edges[rel].data['m'], wdict[etype]['weight']  )  
   g.update_all(fn.copy_e('h*w_r', 'm'), fn.sum('m', 'h'))   # triggers bug
```
Replacing the last line with `multi_update_all` solves the issue which mean the bug is indeed coming from the new `update_all` with HT support.

```
 g.multi_update_all({etype : (fn.copy_e('h*w_r', 'm'), fn.sum('m', 'h')) \
           for etype in g.canonical_etypes}, cross_reducer = 'sum')  # works fine
```

I think the bug is caused by Pytorch's autograd function because: 
1. if only inferencing is used (`with th.no_grad():`), the bug doesn't occur anymore.
2. Turning off the [update](https://github.com/dmlc/dgl/blob/bc5cac44265a2234b230d797d41c8d66f43e1c48/python/dgl/heterograph.py#L4892) of destination nodes by the output of `core.message_passing` function, which means the original node features will always be used, avoids causing the bug too.  Probably, the forward function does not store the tensors or intermediate data any more(!)

The issue might be related to Issue: https://github.com/dmlc/dgl/issues/3348

## To Reproduce

1. The [`update_all`](https://github.com/dmlc/dgl/blob/cafff1640deafc74071b503b3d1fb9c9509b06f6/examples/pytorch/rgcn-hetero/model.py#L230) function call in [PR: RGCN with new heterograph API](https://github.com/dmlc/dgl/pull/3025) triggers this issue. 

## Expected behavior

Each epoch in the RGCN model should use the same amount of memory. The memory consumption shouldn't increase or decrease with epochs.

## Environment

 - DGL Version (e.g., 1.0): 0.7, 0.7.1
[...]

Issue #3230, opened 2021-08-07
RAM keeps increasing for every epoch when increasing number of edges and features


afdezt:
## 🐛 Bug

When increasing the number of edges and features of a heterogenous graph, ram memory (CPU) keeps increasing in each epoch. 

## To Reproduce
For instance, let's take the example from the 5.1 section (Node Classification). When changing the number of _n_follows_ and _n_clicks_ to 500,000 and _n_hetero_features_ to 128, the RAM usage increases after each epoch:
```
import numpy as np
import torch
import dgl
import dgl.nn as dglnn
import torch.nn as nn
import torch.nn.functional as F

n_users = 1000
n_items = 500
n_follows = 500000 # Originally 3000
n_clicks = 500000 # Originally 5000
n_dislikes = 500
n_hetero_features = 128 # Originally 10
n_user_classes = 5
n_max_clicks = 10

follow_src = np.random.randint(0, n_users, n_follows)
follow_dst = np.random.randint(0, n_users, n_follows)
click_src = np.random.randint(0, n_users, n_clicks)
click_dst = np.random.randint(0, n_items, n_clicks)
dislike_src = np.random.randint(0, n_users, n_dislikes)
dislike_dst = np.random.randint(0, n_items, n_dislikes)

hetero_graph = dgl.heterograph({
    ('user', 'follow', 'user'): (follow_src, follow_dst),
    ('user', 'followed-by', 'user'): (follow_dst, follow_src),
    ('user', 'click', 'item'): (click_src, click_dst),
    ('item', 'clicked-by', 'user'): (click_dst, click_src),
    ('user', 'dislike', 'item'): (dislike_src, dislike_dst),
    ('item', 'disliked-by', 'user'): (dislike_dst, dislike_src)})

hetero_graph.nodes['user'].data['feature'] = torch.randn(n_users, n_hetero_features)
hetero_graph.nodes['item'].data['feature'] = torch.randn(n_items, n_hetero_features)
hetero_graph.nodes['user'].data['label'] = torch.randint(0, n_user_classes, (n_users,))
hetero_graph.edges['click'].data['label'] = torch.randint(1, n_max_clicks, (n_clicks,)).float()

# randomly generate training masks on user nodes and click edges
hetero_graph.nodes['user'].data['train_mask'] = torch.zeros(n_users, dtype=torch.bool).bernoulli(0.6)
[...]

PR #4249, opened 2022-07-13
[Bug] Revert clearing backward cache for retain_graph flag

BarclayII:
This change is introduced in #3386 where it aims to resolve memory leaks for certain PyTorch versions (1.8.1 at that time; see #3230 and #3365).  The fix however introduced new problems where `retain_graph=True` during backward is broken (see #4078).  I reverted this PR and found that the memory leak no longer appears in at least PyTorch 1.11.0 and 1.12.0.


BarclayII:
The problem also did not appear in 1.9.1.  I think we are good.

jermainewang (approved):

PR #4253, opened 2022-07-13
[Dist][Optim] Change SparseAdagrad to be numerically closer to PyTorch

Kh4L:
Change op order in the distrbuted optimizer `SparseAdagrad` to be numerically closer to PyTorch's implem:
https://github.com/pytorch/pytorch/blob/347b036350e64fde2e68be20f1bfc2872624a1ec/torch/optim/adagrad.py#L267


mufeili:
Could you take a look and see if it is good to you? @classicsong

mufeili (approved):

PR #4260, opened 2022-07-15
[Example][Bugfix] graphsage node classification example

chang-l:
## Description

node_classification.py still fails due to the same issue https://github.com/dmlc/dgl/issues/4234.
This PR fixes node_classification.py runs under all modes (--mode = `cpu`, `mixed`, `puregpu`)
 
## Checklist
- [ ] Code is well-documented

## Changes

- test_idx always on cpu
- Accuracy calculation `MF.accuracy` always on cpu

### DIFF
--- a/examples/pytorch/graphsage/node_classification.py
+++ b/examples/pytorch/graphsage/node_classification.py
@@ -97,7 +97,7 @@ def train(args, device, g, dataset, model):
 
     opt = torch.optim.Adam(model.parameters(), lr=1e-3, weight_decay=5e-4)
     
-    for epoch in range(3):
+    for epoch in range(10):
         model.train()
         total_loss = 0
         for it, (input_nodes, output_nodes, blocks) in enumerate(train_dataloader):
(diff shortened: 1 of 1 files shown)


yaox12 (approved):

examples/pytorch/graphsage/node_classification.py
```diff
@@ -72,9 +72,9 @@ def evaluate(model, graph, dataloader):
 def layerwise_infer(device, graph, nid, model, batch_size):
     model.eval()
     with torch.no_grad():
-        pred = model.inference(graph, device, batch_size).to(device)
```
nv-dlasalle, line 75: Should we remove the `device` parameter then?

nv-dlasalle (approved):
Overall looks good to me, but want to verify that it works fine with all 3 modes: 'cpu', 'mixed', and 'puregpu'.

chang-l:
> Overall looks good to me, but want to verify that it works fine with all 3 modes: 'cpu', 'mixed', and 'puregpu'.

Yes, it works with all 3 modes.

examples/pytorch/graphsage/node_classification.py
chang-l: `device` is still needed to pass into `model.inference()` function.

PR #4261, opened 2022-07-15
[Transform] Allow add data to self loop created by AddSelfLoop or add_self_loop

RecLusIve-F:
## Checklist
- [x] Code is well-documented


python/dgl/transforms/functional.py
```diff
@@ -1764,13 +1765,23 @@ def remove_nodes(g, nids, ntype=None, store_ids=False):
...
-def add_self_loop(g, etype=None):
+def add_self_loop(g, edge_feat_names=None, fill_data=1., etype=None):
     r"""Add self-loops for each node in the graph and return a new graph.
 
     Parameters
     ----------
     g : DGLGraph
         The graph.
+    edge_feat_names : list[str], optional
+        The names of the self-loops features to add data. If None, all added self-loops features are
```
mufeili, line 1776: "The names of the self-loops features to add data" -> "The names of the self-loop features to apply ``fill_data``."

python/dgl/transforms/functional.py
```diff
@@ -1764,13 +1765,23 @@ def remove_nodes(g, nids, ntype=None, store_ids=False):
...
+        0. Default: None.
+    fill_data : float, dict or str, optional
+        The value to fill the self-loops features. Default: 1.
```
mufeili, line 1779: "self-loops" -> "self-loop"

python/dgl/transforms/functional.py
```diff
@@ -1764,13 +1765,23 @@ def remove_nodes(g, nids, ntype=None, store_ids=False):
...
+
+        * If ``fill_data`` is ``int``, ``float`` or ``dict``, self-loops features will be
+          directly given by ``fill_data``.
```
mufeili, line 1782: It's not clear how it works when ``fill_data`` is ``dict``.

python/dgl/transforms/functional.py
```diff
@@ -1764,13 +1765,23 @@ def remove_nodes(g, nids, ntype=None, store_ids=False):
...
+        * if ``fill_data`` is ``str``, self-loops features will be generated by the reducer.
```
mufeili, line 1783: "the reducer" -> "aggregating the features of the incident edges"?

python/dgl/transforms/functional.py
```diff
@@ -1828,20 +1838,64 @@ def add_self_loop(g, etype=None):
     ...                                   torch.tensor([0, 1])),
     ...     ('user', 'plays', 'game'): (torch.tensor([0, 1]),
     ...                                 torch.tensor([0, 1]))})
-    >>> g = dgl.add_self_loop(g, etype='follows')
+    >>> g.edata['feat'] = {('user', 'follows', 'user'): torch.randn(2, 5),
```
mufeili, line 1841: For simplicity, let's undo the changes in L1841 - 1858.

python/dgl/transforms/functional.py
```diff
@@ -1828,20 +1838,64 @@ def add_self_loop(g, etype=None):
...
+
     if etype[0] != etype[2]:
         raise DGLError(
             'add_self_loop does not support unidirectional bipartite graphs: {}.' \
             'Please make sure the types of head node and tail node are identical.' \
             ''.format(etype))
+
+    for feat_name in edge_feat_names:
+        if isinstance(fill_data, (int, float)):
+            dtype = g.edges[etype[1]].data[feat_name].dtype
```
mufeili, line 1878: L1878 does not work if there are multiple canonical edge types with the same middle edge type name.

python/dgl/transforms/functional.py
```diff
@@ -1828,20 +1838,64 @@ def add_self_loop(g, etype=None):
...
+            data[feat_name] = (F.zeros(
+                (g.num_nodes(etype[0]), g.edges[etype[1]].data[feat_name].shape[1]), dtype,
```
mufeili, line 1880: Ditto

python/dgl/transforms/functional.py
```diff
@@ -1828,20 +1838,64 @@ def add_self_loop(g, etype=None):
...
+                g.device) + fill_data).astype(dtype)
```
mufeili, line 1881: Why do you need `dtype` twice?

python/dgl/transforms/module.py
```diff
@@ -415,6 +415,16 @@ class AddSelfLoop(BaseTransform):
         If False, it will first remove self-loops to prevent duplicate self-loops.
```
mufeili, line 415: Same update according to the comments in `functional.py`

python/dgl/transforms/module.py
```diff
@@ -501,6 +548,13 @@ def __call__(self, g):
                 data_dict[c_etype] = g.edges(etype=c_etype)
 
             g = update_graph_structure(g, data_dict)
+
+            for c_etype in g.canonical_etypes:
+                if c_etype[0] == c_etype[2] and c_etype[1] != 'self':
+                    fill_data = {k: v[-g.num_nodes(c_etype[0]):, :] for k, v in
+                                 g.edges[c_etype].data.items()}
+                    g.edges[(c_etype[0], 'self', c_etype[2])].data.update(fill_data)
+
```
mufeili, line 557: I don't think L552-556 are needed. These are new edge types and they do not have features.

python/dgl/transforms/functional.py
```diff
@@ -1764,13 +1765,23 @@ def remove_nodes(g, nids, ntype=None, store_ids=False):
     g.remove_nodes(nids, ntype=ntype, store_ids=store_ids)
     return g
 
-def add_self_loop(g, etype=None):
+def add_self_loop(g, edge_feat_names=None, fill_data=1., etype=None):
```
mufeili, line 1768: Generally I prefer a simpler design unless otherwise requested by many users.

- Apply `fill_data` for all edge features.
- Not allow `fill_data` to be a dict.

This will make both the implementation and documentation a lot more cleaner.

RecLusIve-F: I am totally agree with you. 

But I would like to discuss with you how many types are allowed in `fill_data`. Right now I have it set to allow `int`, `float`, `str` and `dict`. `float` and `str` are fine in most cases, but `int` is a bit more problematic. If you use aggregation to get edge data, the edge data type is not allowed to be `int`. So I think edge data should always be `float`. As for `dict`, I think it is a shortcut for users to add specific data in a self-loop because `add_node` has a `data` parameter. The `data` parameter is the feature data of the added edge. The key is the feature name and the value is the feature data.

mufeili: > So I think edge data should always be float.

Sometimes edge data are for example binary masks indicating the edges to use for training/validation/test, etc. However, I think this is less an issue as you can always cast the new edge feature to the original data type.

> As for dict, I think it is a shortcut for users to add specific data in a self-loop because add_node has a data parameter. The data parameter is the feature data of the added edge. The key is the feature name and the value is the feature data.

I understand. However this makes things rather complicated when you have multiple edge types and multiple edge features per edge type, particularly when you use `add_self_loop` in `AddSelfLoop`.

RecLusIve-F: > Sometimes edge data are for example binary masks indicating the edges to use for training/validation/test, etc. However, I think this is less an issue as you can always cast the new edge feature to the original data type.

For training/validation/testing edge data, it is not appropriate to aggregate the features of the incoming edges of the corresponding nodes, right? But we would like to assume that `fill_data` is applied to all edge features when `edge_feat_names` is `None`. This is actually like the `RowFeatNorm` case at that time. But the case of training/validation/testing edge data is not as much as the case of training/validation/testing node data, so we can leave `edge_feat_names` to the user to make a choice right?

mufeili: It still might make sense if `fill_data` is `1` or `0`. Again it's almost impossible to cover all the cases as likely different edge features and different edge types require different `fill_data`. Let's do a minimal solution first and expand it in the future if this turns out to be really necessary.

python/dgl/transforms/functional.py
```diff
@@ -1764,13 +1765,24 @@ def remove_nodes(g, nids, ntype=None, store_ids=False):
...
     r"""Add self-loops for each node in the graph and return a new graph.
 
     Parameters
     ----------
     g : DGLGraph
         The graph.
+    edge_feat_names : list[str], optional
+        The names of the self-loop features to apply `fill_data`. If None, it will apply `fill_data`
+        to all self-loop features. Default: None.
+    fill_data : float, dict or str, optional
```
mufeili, line 1778: I think `dict` no longer applies.

python/dgl/transforms/functional.py
```diff
@@ -1831,17 +1842,49 @@ def add_self_loop(g, etype=None):
...
+
+    for feat_name in edge_feat_names:
+        if isinstance(fill_data, (int, float)):
+            dtype = g.edges[etype].data[feat_name].dtype
+            tmp_fill_data = F.copy_to(F.tensor([fill_data], dtype), g.device)
+            try:
+                data[feat_name] = F.zeros(
+                    (g.num_nodes(etype[0]), g.edges[etype].data[feat_name].shape[1]), dtype,
+                    g.device) + tmp_fill_data
+            except IndexError:
```
mufeili, line 1872: Rather than use the try-except logic. You can first check the shape of `g.edges[etype].data[feat_name]`.

python/dgl/transforms/module.py
```diff
@@ -415,6 +415,17 @@ class AddSelfLoop(BaseTransform):
         If False, it will first remove self-loops to prevent duplicate self-loops.
     new_etypes : bool, optional
         If True, it will add an edge type 'self' per node type, which holds self-loops.
+    edge_feat_names : list[str], optional
+        The names of the self-loop features to apply `fill_data`. If None, it will apply `fill_data`
+        to all self-loop features. Default: None.
+    fill_data : float, dict or str, optional
```
mufeili, line 421: `dict` does not apply.

python/dgl/transforms/module.py
```diff
@@ -424,36 +435,72 @@ class AddSelfLoop(BaseTransform):
...
+   >>> g.edata['feat1'] = {('user', 'follows', 'user'): torch.randn(2, 15),
+     ...                   ('user', 'plays', 'game'): torch.randn(2, 15)}
     >>> new_g = transform(g)
     >>> print(new_g.edges(etype='plays'))
-    (tensor([0]), tensor([1]))
+    (tensor([0, 1]), tensor([0, 1]))
     >>> print(new_g.edges(etype='follows'))
-    (tensor([1, 0, 1, 2]), tensor([2, 0, 1, 2]))
+    (tensor([1, 2]), tensor([0, 1]))
+    >>> print(new_g.edata['feat'])
```
mufeili, line 469: No need to print the random values. It suffices to print the shape.

python/dgl/transforms/module.py
```diff
@@ -424,36 +435,72 @@ class AddSelfLoop(BaseTransform):
...
     Case3: Add self-etypes for a heterogeneous graph
 
-    >>> transform = AddSelfLoop(new_etypes=True)
+    >>> transform = AddSelfLoop(new_etypes=True, edge_feat_names=['feat', 'feat1'], fill_data='sum')
     >>> new_g = transform(g)
     >>> print(new_g.edges(etype='follows'))
     (tensor([1, 0, 1, 2]), tensor([2, 0, 1, 2]))
     >>> print(new_g.edges(etype=('game', 'self', 'game')))
     (tensor([0, 1]), tensor([0, 1]))
+    >>> print(new_g.edata['feat'])
```
mufeili, line 486: Ditto

tests/compute/test_transform.py
```diff
@@ -1647,17 +1647,27 @@ def test_add_selfloop(idtype):
         raise_error = True
     assert raise_error
 
-    g = create_test_heterograph5(idtype)
-    g = dgl.add_self_loop(g, etype='follows')
+    # g = create_test_heterograph5(idtype)
```
mufeili, line 1650: Remove this line then

tests/compute/test_transform.py
```diff
@@ -1626,15 +1626,15 @@ def test_remove_nodes(idtype):
 def test_add_selfloop(idtype):
     # homogeneous graph
     g = dgl.graph(([0, 0, 2], [2, 1, 0]), idtype=idtype, device=F.ctx())
-    g.edata['he'] = F.copy_to(F.tensor([1, 2, 3], dtype=idtype), ctx=F.ctx())
+    g.edata['he'] = F.copy_to(F.tensor([1., 2., 3.]), ctx=F.ctx())
     g.ndata['hn'] = F.copy_to(F.tensor([1, 2, 3], dtype=idtype), ctx=F.ctx())
-    g = dgl.add_self_loop(g)
+    g = dgl.add_self_loop(g, edge_feat_names=['he'], fill_data='sum')
```
mufeili, line 1631: Add tests for int and float

tests/compute/test_transform.py
```diff
@@ -1634,7 +1636,31 @@ def test_add_selfloop(idtype):
...
+    g = dgl.graph(([0, 0, 2], [2, 1, 0]), idtype=idtype, device=F.ctx())
+    g.edata['he'] = F.copy_to(F.tensor([1, 2, 3]), ctx=F.ctx())
+    g.ndata['hn'] = F.copy_to(F.tensor([1, 2, 3], dtype=idtype), ctx=F.ctx())
+    g = dgl.add_self_loop(g, fill_data=1)
+    assert g.number_of_nodes() == 3
+    assert g.number_of_edges() == 6
+    u, v = g.edges(form='uv', order='eid')
+    assert F.array_equal(u, F.tensor([0, 0, 2, 0, 1, 2], dtype=idtype))
+    assert F.array_equal(v, F.tensor([2, 1, 0, 0, 1, 2], dtype=idtype))
+    assert F.array_equal(g.edata['he'], F.tensor([1, 2, 3, 1, 1, 1]))
```
mufeili, line 1651: Also add an edge feature of dimensionality 2, e.g., [[0., 1.], [2., 3.], [4., 5.]].

mufeili (approved):
LGTM other than a final minor comment

PR #4262, opened 2022-07-16
[Bug] Fix NN modules crashing with FP16 inputs

BarclayII:
Some of the modules compute normalizers etc. as float32 vectors.  As we are going to support float16 inputs, this will result in math between float32 and float16, causing crashes.

Also I attached my GCN example with float16 (EDIT: for awareness only, won't go into actual example), which always return NaN losses.  Have we tested our examples (not necessarily all of them though) with FP16 inputs?  @chang-l @nv-dlasalle 

EDIT: AMP with autocast worked fine.

### DIFF
--- a/examples/pytorch/gcn/train_fp16.py
+++ b/examples/pytorch/gcn/train_fp16.py
@@ -1,0 +1,99 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import dgl
+import dgl.nn as dglnn
+from dgl.data import CoraGraphDataset, CiteseerGraphDataset, PubmedGraphDataset
+from dgl import AddSelfLoop
+import argparse
+
+class GCN(nn.Module):
+    def __init__(self, in_size, hid_size, out_size):
+        super().__init__()
+        self.layers = nn.ModuleList()
+        # two-layer GCN
+        self.layers.append(dglnn.GraphConv(in_size, hid_size, activation=F.relu))
+        self.layers.append(dglnn.GraphConv(hid_size, out_size))
+        self.dropout = nn.Dropout(0.5)
+
+    def forward(self, g, features):
+        h = features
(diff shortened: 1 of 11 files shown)


chang-l:
No, not that I am aware of... I can check to run some examples with FP16 later.

nv-dlasalle:
@BarclayII When you say a crash, do you mean just an error message about mismatched types, or does it exit with segfault unrelated error message? If it is the former, this is consistent behavior with PyTorch:
```
>>> import torch
>>> a = torch.empty(10, dtype=torch.float16)
>>> l = torch.nn.Linear(10,10)
>>> l(a)
  File "/home/dominique/.local/lib/python3.6/site-packages/torch/nn/modules/module.py", line 1051, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/dominique/.local/lib/python3.6/site-packages/torch/nn/modules/linear.py", line 96, in forward
    return F.linear(input, self.weight, self.bias)
  File "/home/dominique/.local/lib/python3.6/site-packages/torch/nn/functional.py", line 1847, in linear
    return torch._C._nn.linear(input, weight, bias)
RuntimeError: expected scalar type Half but found Float
```

yaox12:
Full-FP16 training can seldom meet the convergence. It's usually used for establishing a speed baseline for your model as stated in   https://nvidia.github.io/apex/amp.html#o3-fp16-training. And you may still need `torch.cuda.amp.autocast(enabled=True)` and `torch.cuda.amp.GradScaler` to work with `model.to(dtype=torch.float16)`.

mufeili:
Do we need to update the unit tests to cover these changes?

BarclayII:
Closing because using FP16 directly usually gives bad results.

mufeili:
@BarclayII bad in terms of what? accuracy?

PR #4265, opened 2022-07-19
[Dist][Test] Improves DistTensor test for num_part id > 2

Kh4L:
## Description
Improves DistTensor sanity test when num_part id > 2.

### DIFF
--- a/tests/dist/python/run_dist_objects.py
+++ b/tests/dist/python/run_dist_objects.py
@@ -33,7 +33,6 @@ def run_server(graph_name, server_id, server_count, num_clients, shared_mem, kee
     print('start server', server_id)
     g.start()
 
-
 def dist_tensor_test_sanity(data_shape, rank, name=None):
     dist_ten = dgl.distributed.DistTensor(data_shape,
                                           F.int32,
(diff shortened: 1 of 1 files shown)


tests/dist/python/run_dist_objects.py
```diff
@@ -33,22 +33,22 @@ def run_server(graph_name, server_id, server_count, num_clients, shared_mem, kee
...
-        dgl.distributed.client_barrier()
-    else:
-        dgl.distributed.client_barrier()
-        original_rank = rank % num_client_per_machine
-        assert F.allclose(dist_ten[original_rank*stride:(original_rank+1)*stride],
-                          F.ones((stride, 2), dtype=F.int32, ctx=F.cpu()) * (original_rank+1))
+    local_rank = rank % num_client_per_machine
+    pos = (part_id // 2) * num_client_per_machine + local_rank
+    if part_id % 2 == 0:
+        dist_ten[pos*stride:(pos+1)*stride] = F.ones((stride, 2), dtype=F.int32, ctx=F.cpu()) * (pos+1)
```
Rhett-Ying, line 46: with these stride/pos values, will it be possible that multiple clients are modifying/accessing same tensor data in `dist_ten`?

Kh4L: For a given client of local rank `r` on a machine with even `part_id` such as `part_id % 2 == 0` machine will assign a slice of values in dist_ten. These will be checked by the client itself `r` and the client `r` of the machine `part_id+1`

Rhett-Ying: why `stride=3`? the updated slice size is related to `data_shape`, `local_rank`, `part_id`, right? how about this? let's assign unique values in each update on different clients. and verify it. If so ,we could detect if there's any slice has been updated twice or more.

Kh4L: `he updated slice size is related to data_shape, local_rank, part_id, right? `
The slice size slice size is of shape `(stride, 2)`
The `pos` is calculated from `local_rank` and `part_id`

Rhett-Ying (approved):

PR #4273, opened 2022-07-20
Code changes to address the updated file format support for massively…

kylasa:
… large graphs.

1. Updated the docstring for the starting function 'gen_dist_partitions" to describe the newly proposed file format for input dataset.
2. Code which was dependent on the structure of the old-metadata json object has been updated to read from the newly proposed metadata file.
3. Fixed some errors when appropriate functions were invoked and the calling function expects return values from the invoked furnction.
4. This modified code has been tested on "mag" dataset using 4-way partitions and verified the results

## Checklist
- [X] Code is well-documented

### DIFF
--- a/tools/distpartitioning/convert_partition.py
+++ b/tools/distpartitioning/convert_partition.py
@@ -95,19 +95,29 @@ def create_dgl_object(graph_name, num_parts, \
         map between edge type(string)  and edge_type_id(int)
     """
     #create auxiliary data structures from the schema object
-    node_info = schema["nid"]
+    ntype_names = schema[constants.STR_NODE_TYPE]
     offset = 0
     global_nid_ranges = {}
-    for k, v in node_info.items():
-        global_nid_ranges[k] = np.array([offset + int(v["data"][0][1]), offset + int(v["data"][-1][2])]).reshape(1,2)
-        offset += int(v["data"][-1][2])
+    for idx, ntype in enumerate(ntype_names):
+        chunks = schema[constants.STR_NUM_NODES_PER_CHUNK][idx]
+        tid_start = np.cumsum([0] + chunks[:-1])
+        tid_end = np.cumsum(chunks)
+        tid_ranges = list(zip(tid_start, tid_end))
+        global_nid_ranges[ntype] = np.array([offset + tid_ranges[0][0], offset + tid_ranges[-1][1]]).reshape(1,2)
+        offset += tid_ranges[-1][1]
 
-    edge_info = schema["eid"]
+    etype_names = schema[constants.STR_EDGE_TYPE]
     offset = 0
     global_eid_ranges = {}
-    for k, v in edge_info.items():
-        global_eid_ranges[k] = np.array([offset + int(v["data"][0][1]), offset + int(v["data"][-1][2])]).reshape(1,2)
-        offset += int(v["data"][-1][2])
+    for idx, etype in enumerate(etype_names):
+        chunks = schema[constants.STR_NUM_EDGES_PER_CHUNK][idx]
+        tid_start = np.cumsum([0] + chunks[:-1])
+        tid_end = np.cumsum(chunks)
+        tid_ranges = list(zip(tid_start, tid_end))
+        tokens = etype.split(":")
+        assert len(tokens) == 3
+        global_eid_ranges[tokens[1]] = np.array([offset + tid_ranges[0][0], offset + tid_ranges[-1][1]]).reshape(1,2)
+        offset += tid_ranges[-1][1]
 
     id_map = dgl.distributed.id_map.IdMap(global_nid_ranges)
 
--- a/tools/distpartitioning/convert_partition.py
+++ b/tools/distpartitioning/convert_partition.py
@@ -9,7 +9,7 @@ import pyarrow
 import pandas as pd
 import constants
 from pyarrow import csv
-from utils import read_json
+from utils import read_json, get_idranges
 
 def create_dgl_object(graph_name, num_parts, \
                         schema, part_id, node_data, \
@@ -95,29 +95,11 @@ def create_dgl_object(graph_name, num_parts, \
         map between edge type(string)  and edge_type_id(int)
     """
     #create auxiliary data structures from the schema object
-    ntype_names = schema[constants.STR_NODE_TYPE]
-    offset = 0
-    global_nid_ranges = {}
-    for idx, ntype in enumerate(ntype_names):
-        chunks = schema[constants.STR_NUM_NODES_PER_CHUNK][idx]
-        tid_start = np.cumsum([0] + chunks[:-1])
-        tid_end = np.cumsum(chunks)
-        tid_ranges = list(zip(tid_start, tid_end))
-        global_nid_ranges[ntype] = np.array([offset + tid_ranges[0][0], offset + tid_ranges[-1][1]]).reshape(1,2)
-        offset += tid_ranges[-1][1]
+    ntid_dict, global_nid_ranges = get_idranges(schema[constants.STR_NODE_TYPE], 
+                                    schema[constants.STR_NUM_NODES_PER_CHUNK])
 
-    etype_names = schema[constants.STR_EDGE_TYPE]
-    offset = 0
-    global_eid_ranges = {}
-    for idx, etype in enumerate(etype_names):
-        chunks = schema[constants.STR_NUM_EDGES_PER_CHUNK][idx]
-        tid_start = np.cumsum([0] + chunks[:-1])
-        tid_end = np.cumsum(chunks)
-        tid_ranges = list(zip(tid_start, tid_end))
-        tokens = etype.split(":")
-        assert len(tokens) == 3
-        global_eid_ranges[tokens[1]] = np.array([offset + tid_ranges[0][0], offset + tid_ranges[-1][1]]).reshape(1,2)
-        offset += tid_ranges[-1][1]
+    etid_dict, global_eid_ranges = get_idranges(schema[constants.STR_EDGE_TYPE], 
+                                    schema[constants.STR_NUM_EDGES_PER_CHUNK])
 
     id_map = dgl.distributed.id_map.IdMap(global_nid_ranges)
 
@@ -130,10 +112,10 @@ def create_dgl_object(graph_name, num_parts, \
     etypes.sort(key=lambda e: e[1])
     etype_offset_np = np.array([e[1] for e in etypes])
     etypes = [e[0] for e in etypes]
-    etypes_map = {e: i for i, e in enumerate(etypes)}
+    etypes_map = {e.split(":")[1]: i for i, e in enumerate(etypes)}
 
     node_map_val = {ntype: [] for ntype in ntypes}
-    edge_map_val = {etype: [] for etype in etypes}
+    edge_map_val = {etype.split(":")[1]: [] for etype in etypes}
 
     shuffle_global_nids, ntype_ids, global_type_nid = node_data[constants.SHUFFLE_GLOBAL_NID], \
             node_data[constants.NTYPE_ID], node_data[constants.GLOBAL_TYPE_NID]
@@ -168,8 +150,10 @@ def create_dgl_object(graph_name, num_parts, \
     # Determine the edge ID range of different edge types.
     edge_id_start = edgeid_offset 
     for etype_name in global_eid_ranges:
-        etype_id = etypes_map[etype_name]
-        edge_map_val[etype_name].append([edge_id_start,
+        tokens = etype_name.split(":")
+        assert len(tokens) == 3
+        etype_id = etypes_map[tokens[1]]
+        edge_map_val[tokens[1]].append([edge_id_start,
                                          edge_id_start + np.sum(etype_ids == etype_id)])
         edge_id_start += np.sum(etype_ids == etype_id)
 
--- a/tools/distpartitioning/data_shuffle.py
+++ b/tools/distpartitioning/data_shuffle.py
@@ -11,9 +11,9 @@ import dgl
 from timeit import default_timer as timer
 from datetime import timedelta
 from dataset_utils import get_dataset
-from utils import read_partitions_file, read_json, get_node_types, \
+from utils import read_ntype_partition_files, read_json, get_node_types, \
                     augment_edge_data, get_gnid_range_map, \
-                    write_dgl_objects, write_metadata_json
+                    write_dgl_objects, write_metadata_json, get_ntype_featnames
 from gloo_wrapper import alltoall_cpu_object_lst, alltoallv_cpu, \
                     alltoall_cpu, allgather_sizes, gather_metadata_json
 from globalids import assign_shuffle_global_nids_nodes, \
@@ -21,7 +21,7 @@ from globalids import assign_shuffle_global_nids_nodes, \
                     get_shuffle_global_nids_edges
 from convert_partition import create_dgl_object, create_metadata_json
 
-def gen_node_data(rank, world_size, node_part_ids, ntid_ntype_map, nid_schema_map):
+def gen_node_data(rank, world_size, node_part_ids, ntid_ntype_map, schema_map):
     '''
     For this data processing pipeline, reading node files is not needed. All the needed information about
     the nodes can be found in the metadata json file. This function generates the nodes owned by a given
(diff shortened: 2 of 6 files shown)


jermainewang:
I only managed to review the script entrypoint. I suppose `data_proc_pipeline.py` will be invoked as the UDF command of `launch.py`? I didn't see a `dispatch_data.py` script though, which should follow the following usage:

```python
python /myrepo/dgl/tools/dispatch_data.py         \
   --in-dir=/mydata/MAG240M-LSC_chunked/          \
   --partition-file=/mydata/MAG240M-LSC_2parts/   \
   --out-dir=/data/MAG_LSC_partitioned            \
   --ip-config=ip_config.txt
```

tools/distpartitioning/data_shuffle.py
```diff
@@ -292,18 +299,18 @@ def exchange_node_features(rank, world_size, node_feature_tids, ntype_gnid_map,
             own_node_features[feat_key] = []
             own_global_nids[feat_key] = []
             for idx, x in enumerate(output_feat_list):
-                own_node_features.append(x[feat_key])
+                own_node_features[feat_key].append(x[feat_key])
                 own_global_nids[feat_key].append(output_nid_list[idx][feat_key])
             for k in own_node_features.keys(): 
-                own_node_features[k] = th.cat(own_node_features[k])
```
zheng-da, line 298: how come `th.cat` work before?

tools/distpartitioning/data_shuffle.py
```diff
@@ -292,18 +299,18 @@ def exchange_node_features(rank, world_size, node_feature_tids, ntype_gnid_map,
             own_node_features[feat_key] = []
             own_global_nids[feat_key] = []
             for idx, x in enumerate(output_feat_list):
-                own_node_features.append(x[feat_key])
```
zheng-da, line 295: is this a bug?

zheng-da: why is this bug not captured in the previous test?

tools/distpartitioning/convert_partition.py
```diff
@@ -95,19 +95,29 @@ def create_dgl_object(graph_name, num_parts, \
...
-        global_eid_ranges[k] = np.array([offset + int(v["data"][0][1]), offset + int(v["data"][-1][2])]).reshape(1,2)
-        offset += int(v["data"][-1][2])
+    for idx, etype in enumerate(etype_names):
+        chunks = schema[constants.STR_NUM_EDGES_PER_CHUNK][idx]
+        tid_start = np.cumsum([0] + chunks[:-1])
+        tid_end = np.cumsum(chunks)
+        tid_ranges = list(zip(tid_start, tid_end))
+        tokens = etype.split(":")
+        assert len(tokens) == 3
+        global_eid_ranges[tokens[1]] = np.array([offset + tid_ranges[0][0], offset + tid_ranges[-1][1]]).reshape(1,2)
```
zheng-da, line 119: here is it a good idea to just use etype as the key? what if two canonical etype has the same etype?

tools/distpartitioning/data_shuffle.py
```diff
@@ -92,12 +92,19 @@ def gen_node_data(rank, world_size, node_part_ids, ntid_ntype_map, nid_schema_ma
...
 
-        gnid_end += type_end
+        tid_info = ntype_info[ntypeid] #array of no.... chunks per partition
+        
+        tid_start = np.cumsum([0] + tid_info[:-1])
+        tid_end = np.cumsum(tid_info)
+        tid_ranges = list(zip(tid_start, tid_end))
+
+        type_start = tid_ranges[0][0]
+        type_end = tid_ranges[-1][1]
```
zheng-da, line 105: i feel you should define a function for converting the number of nodes to an ID range.

tools/distpartitioning/data_shuffle.py
```diff
@@ -11,17 +11,17 @@
...
+                    write_dgl_objects, write_metadata_json, get_ntype_featnames
 from gloo_wrapper import alltoall_cpu_object_lst, alltoallv_cpu, \
                     alltoall_cpu, allgather_sizes, gather_metadata_json
 from globalids import assign_shuffle_global_nids_nodes, \
                     assign_shuffle_global_nids_edges, \
                     get_shuffle_global_nids_edges
 from convert_partition import create_dgl_object, create_metadata_json
 
-def gen_node_data(rank, world_size, node_part_ids, ntid_ntype_map, nid_schema_map):
+def gen_node_data(rank, world_size, node_part_ids, ntid_ntype_map, schema_map):
```
zheng-da, line 24: `schema_map` is changed. you should update the docstring.

tools/distpartitioning/data_shuffle.py
```diff
@@ -399,10 +406,11 @@ def read_dataset(rank, world_size, node_part_ids, params, schema_map):
         edge features which is also a dictionary, similar to node features dictionary
     """
     edge_features = {}
+    #node_tids, node_features, edge_datadict, edge_tids
     node_tids, node_features, node_feat_tids, edge_data, edge_tids = \
         get_dataset(params.input_dir, params.graph_name, rank, world_size, schema_map)
 
-    augment_edge_data(edge_data, node_part_ids, prefix_sum_edges[rank])
```
zheng-da, line 405: why can i not find `prefix_sum_edges` anywhere? is the code tested?

tools/distpartitioning/data_shuffle.py
```diff
@@ -534,12 +543,14 @@ def gen_dist_partitions(rank, world_size, params):
     print('[Rank: ', rank, '] Starting distributed data processing pipeline...')
 
     #init processing
+    schema_map = read_json(params.input_dir+'/'+params.schema)
```
zheng-da, line 546: shouldn't you use `os.path.join()` to get the full path?

tools/distpartitioning/dataset_utils.py
```diff
@@ -48,60 +48,171 @@ def get_dataset(input_dir, graph_name, rank, world_size, schema_map):
...
-        node_tids[ntype_name] = v
+    ntype_gnid_offset = {}
+    gnid_offset = 0
+    ntype_names = schema_map[constants.STR_NODE_TYPE]
+    for idx, counts in enumerate(schema_map[constants.STR_NUM_NODES_PER_CHUNK]):
+        ntype_name = ntype_names[idx]
+        ntype_gnid_offset[ntype_name] = gnid_offset 
+        type_nid_start = np.cumsum([0] + counts[:-1])
+        type_nid_end = np.cumsum(counts)
+        type_nid_ranges = list(zip(type_nid_start, type_nid_end))
```
zheng-da, line 139: this logic appears again. we should have a function to convert it.

tools/distpartitioning/dataset_utils.py
```diff
@@ -48,60 +48,171 @@ def get_dataset(input_dir, graph_name, rank, world_size, schema_map):
...
+    etype_geid_offset = {}
+    geid_offset = 0
+    etype_names = schema_map[constants.STR_EDGE_TYPE]
+    etype_name_idmap = {e : idx for idx, e in enumerate(etype_names)}
+    for idx, counts in enumerate(schema_map[constants.STR_NUM_EDGES_PER_CHUNK]):
+        etype_name = etype_names[idx]
+        etype_geid_offset[etype_name] = geid_offset
+        type_eid_start = np.cumsum([0] + counts[:-1])
+        type_eid_end = np.cumsum(counts)
+        type_eid_ranges = list(zip(type_eid_start, type_eid_end))
```
zheng-da, line 173: the same here.

tools/distpartitioning/utils.py
```diff
@@ -8,29 +8,38 @@
...
+        directory in which the node-id to partition-id mappings files are 
+        located for each of the node types in the input graph
 
     Returns:
     --------
-    numpy array
-        array of part_ids and the idx is the <global_nid>
+    numpy array : 
+        array of integers representing mapped partition-ids for a given node-id
+        the line number is used as the type_node_id in each of the files.
```
zheng-da, line 29: what is the line number?

tools/distpartitioning/utils.py
```diff
@@ -8,29 +8,38 @@
...
+    #iterate over the node types and extract the partition id mappings
+    part_ids = []
+    ntype_names = schema_map[constants.STR_NODE_TYPE]
+    for ntype in ntype_names:
+        df = csv.read_csv(os.path.join(input_dir, '{}.txt'.format(ntype)), \
+                read_options=pyarrow.csv.ReadOptions(autogenerate_column_names=True), \
+                parse_options=pyarrow.csv.ParseOptions(delimiter=' '))
+        ntype_partids = df['f0'].to_numpy()
+        part_ids.append(ntype_partids)
+    return np.concatenate(part_ids)
```
zheng-da, line 42: it seems to me that the returned partition IDs are stored for all nodes in the graph.

tools/distpartitioning/utils.py
```diff
@@ -8,29 +8,38 @@
...
 from pyarrow import csv
 
-def read_partitions_file(part_file):
+def read_ntype_partition_files(schema_map, input_dir):
     """
-    Utility method to read metis partitions, which is the output of 
-    pm_dglpart2
+    Utility method to read the partition id mapping for each node.
+    For each node type, there will be an file, in the input directory argument
+    containing the partition id mapping for a given nodeid.
```
zheng-da, line 15: if you return a single numpy, you basically assume the returned partition id array will be used for the homogenized graph. if so, you need to explain it in the doc string.

tools/distpartitioning/dataset_utils.py
```diff
@@ -48,60 +48,171 @@ def get_dataset(input_dir, graph_name, rank, world_size, schema_map):
         edge-type may have several edge features and associated tensor data.
```
zheng-da, line 48: this function is to read the entire dataset? if so, can you provide more information in the docstring. for example, what data is read?

tools/distpartitioning/dataset_utils.py
```diff
@@ -48,60 +48,171 @@ def get_dataset(input_dir, graph_name, rank, world_size, schema_map):
...
+    #stitch together to create the final data on the local machine
+    for col in [constants.GLOBAL_SRC_ID, constants.GLOBAL_DST_ID, constants.GLOBAL_TYPE_EID, constants.ETYPE_ID]:
+        edge_datadict[col] = np.concatenate(edge_datadict[col])
+
+    assert edge_datadict[constants.GLOBAL_SRC_ID].shape == edge_datadict[constants.GLOBAL_DST_ID].shape
+    assert edge_datadict[constants.GLOBAL_DST_ID].shape == edge_datadict[constants.GLOBAL_TYPE_EID].shape
+    assert edge_datadict[constants.GLOBAL_TYPE_EID].shape == edge_datadict[constants.ETYPE_ID].shape
     print('[Rank: ', rank, '] Done reading edge_file: ', len(edge_datadict), edge_datadict[constants.GLOBAL_SRC_ID].shape)
 
     return node_tids, node_features, node_feature_tids, edge_datadict, edge_tids
```
zheng-da, line 218: it seems to me that `node_tids` and `node_feature_tids` have redundant information. both of them are used to store the id ranges for each node type.

tools/distpartitioning/data_shuffle.py
zheng-da: something is wrong. here. the original input arguments are incorrect. what is going on?

tools/distpartitioning/convert_partition.py
kylasa: Agreed. Changed the code to use canonical etype as the key everywhere in the convert_partition.py file. The side effect of this will be that the output metadata json file will have canonical etypes instead of the just etype's in it. Please make sure that the training is affected by these canonical etypes.

tools/distpartitioning/data_shuffle.py
kylasa: Agreed. Updated the docstring to match with the current updated file format. Also explained how this information can be extracted from the metadata json file.

tools/distpartitioning/data_shuffle.py
```diff
@@ -92,12 +92,19 @@ def gen_node_data(rank, world_size, node_part_ids, ntid_ntype_map, nid_schema_ma
     gnid_start = 0
     gnid_end = 0
     for ntypeid in range(len(ntid_ntype_map)):
-        ntype_name = ntid_ntype_map[str(ntypeid)]
-        ntype_info = nid_schema_map[ntype_name]
-        type_start = int(ntype_info["data"][0][1])
-        type_end = int(ntype_info["data"][-1][2])
+        ntype_name = ntid_ntype_map[ntypeid]
```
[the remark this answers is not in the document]
kylasa: ntid_ntype_map, In this dictionary key-value pairs are ntype_id and ntype_name pairs. This dictionary is not changed. It remains the same. Except that instead of strings we now use integers as the keys for this dictionary. This was as per our last dicussion. Will update the docstring accordingly.

tools/distpartitioning/data_shuffle.py
kylasa: Agreed. This logic is used in multiple places (convert_partition.py, data_shuffle.py and dataset_utils.py). Now added a new function in the utils.py to compte the type-ids and global-ids of nodes and edges and this new function is used everywhere such functionality is needed.

tools/distpartitioning/data_shuffle.py
kylasa: Yes. This was a bug which made its way into the repo when the code changes were merged from the private copy of the repo in which actual coding and testing are done and when completed these newly added changes are merged into the github repo. From now on, I changed my setup so that all the changes are done on the github repo itself and testing is performed on the same repo. So this will not happen again.

tools/distpartitioning/data_shuffle.py
kylasa: The same reason as described in the previous comment. The coding/testing is performed on a private copy and changes are merged into the github repo. This error made into the github repo inadverdantly. Now I am using a different setup for coding/testing/github-merge so this will not repeat again.

tools/distpartitioning/data_shuffle.py
kylasa: Agreed. changed appropriately.

tools/distpartitioning/dataset_utils.py
kylasa: Added very detailed description in the comments about nodes, edges and node-types and their assocaited data etc...

tools/distpartitioning/dataset_utils.py
kylasa: Done.

tools/distpartitioning/dataset_utils.py
kylasa: Yes. with the new file format these would be redundant. 
These were introduced during coding the previous version of the input file format. In this format, we had the flexibility of providing different no. of nodes and completely different no. of node features. These two dictionaries where used for this purpose. We can revisit this in the next iteration and change them to our convenience. Lets not make this code change now and target it for the post-release.

tools/distpartitioning/utils.py
kylasa: Yes. This as per the design we all agreed upon. I will elaborate the docstring to make clear about this issue.

tools/distpartitioning/utils.py
kylasa: Explained clearly that the line number in these files are basically per_type ids.

tools/distpartitioning/utils.py
kylasa: Yes. This is the for the entire graph. Currently this is a bottleneck. Per our discussion, we have to change this so that each node only stores the partition-ids for the nodes which it owns. This is targetted for the next version of the release.

tools/distpartitioning/utils.py
```diff
@@ -182,7 +188,7 @@ def write_metadata_json(metadata_list, output_dir, graph_name):
         graph_metadata["part-{}".format(i)] = metadata_list[i]["part-{}".format(i)]
 
     with open('{}/{}.json'.format(output_dir, graph_name), 'w') as outfile: 
-        json.dump(graph_metadata, outfile, sort_keys=True, indent=4)
+        json.dump(graph_metadata, outfile, sort_keys=False, indent=4)
```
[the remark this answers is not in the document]
kylasa: The output metadata json, because we set the sort_keys to True, all the entries are alphabetically sorted, by following either the string ordering or numerical ordering. By setting the sort to False all the keys are properly put in their places. And it is easier to read off from this output metadata json

tools/distpartitioning/dataset_utils.py
zheng-da: how about write a TODO here?

zheng-da (approved):

tools/distpartitioning/dataset_utils.py
kylasa: Done

PR #4276, opened 2022-07-20
Adding launch script and wrapper script to trigger distributed graph …

kylasa:
…partitioning pipeline as defined in the UX document

1. dispatch_data.py is a wrapper script which builds the command and triggers the distributed partitioning pipeline
2. distgraphlaunch.py is the main python script which triggers the pipeline and to simplify its usage dispatch_data.py is included as a wrapper script around it.

## Checklist
- [X] Code is well-documented

### DIFF
--- a/tools/dispatch_data.py
+++ b/tools/dispatch_data.py
@@ -21,16 +21,16 @@ LARG_PROCS_MACHINE = "num_proc_per_machine"
 LARG_IPCONF = "ip_config"
 LARG_MASTER_PORT = "master_port"
 
-def get_launch_cmd(args) -> str: 
+def get_launch_cmd(args, user_python_exe) -> str: 
 
-    cmd = pythonpath=os.environ.get("PYTHONPATH", "") + get_python_path() + " " + os.path.join(INSTALL_DIR, LAUNCH_SCRIPT)
+    cmd = pythonpath=os.environ.get("PYTHONPATH", "") + get_python_path(user_python_exe) + " " + os.path.join(INSTALL_DIR, LAUNCH_SCRIPT)
     cmd = f"{cmd} --{LARG_PROCS_MACHINE} 1 "
     cmd = f"{cmd} --{LARG_IPCONF} {args.ip_config} "
     cmd = f"{cmd} --{LARG_MASTER_PORT} {args.master_port} "
 
     return cmd
 
-def get_python_path() -> str:
+def get_python_path(user_python_exe) -> str:
 
     # Auto-detect the python binary that kicks off the distributed trainer code.
     # Note: This allowlist order matters, this will match with the FIRST matching entry. Thus, please add names to this
@@ -42,41 +42,54 @@ def get_python_path() -> str:
         "python2.7", "python2",
         "python3.6", "python3.7", "python3.8", "python3.9", "python3",
     )
-    return python_bin_allowlist[-1]
+
+    # If none of the candidate python bins match, then we go with the default `python`
+    python_bin = "python"
+    for candidate_python_bin in python_bin_allowlist:
+        if candidate_python_bin == user_python_exe :
+            python_bin = candidate_python_bin
+            break
+    return python_bin
 
 
-def submit_jobs(args) -> str:
+def submit_jobs(args, user_python_exe) -> str:
     wrapper_command = os.path.join(INSTALL_DIR, LAUNCH_SCRIPT)
 
     UDF_ARGS = [UDF_WORLD_SIZE, UDF_PART_DIR, UDF_INPUT_DIR, UDF_GRAPH_NAME, UDF_SCHEMA, UDF_NUM_PARTS, UDF_OUT_DIR]
+
+    #read the json file and get the remaining argument here. 
+    with open(os.path.join(args.in_dir, "meta.json")) as schema:
+        schema_map = json.load(schema)
+
+    num_parts = len(schema_map["num_nodes_per_chunk"][0])
+    graph_name = schema_map["graph_name"]
+
     argslist = ""
     for cmd_arg in UDF_ARGS: 
         if cmd_arg == "world-size": 
-            argslist += "--world-size 4 "
+            argslist += "--world-size 4 ".format(num_parts)
         elif cmd_arg == "partitions-dir":
             argslist += "--partitions-dir {} ".format(args.partitions_dir)
         elif cmd_arg == "input-dir":
             argslist += "--input-dir {} ".format(args.in_dir)
         elif cmd_arg == "graph-name":
-            argslist += "--graph-name mag "
+            argslist += "--graph-name {} ".format(graph_name)
         elif cmd_arg == "schema":
-            argslist += "--schema mag.json "
+            argslist += "--schema meta.json "
         elif cmd_arg == "num-parts":
-            argslist += "--num-parts 4 "
+            argslist += "--num-parts {} ".format(num_parts)
         elif cmd_arg == "output":
             argslist += "--output {} ".format(args.out_dir)
         
-    pythonpath=os.environ.get("PYTHONPATH", "")+get_python_path()
+    pythonpath=os.environ.get("PYTHONPATH", "")+get_python_path(user_python_exe)
     pipeline_cmd = os.path.join(INSTALL_DIR, PIPELINE_SCRIPT)
     udf_cmd = f"{pythonpath} {pipeline_cmd} {argslist}"
 
-    launch_cmd = get_launch_cmd(args)
+    launch_cmd = get_launch_cmd(args, user_python_exe)
     launch_cmd += '\"'+udf_cmd+'\"'
 
     print(launch_cmd)
     os.system(launch_cmd)
-
-
 
 def main():
     parser = argparse.ArgumentParser(description='Launch a distributed job')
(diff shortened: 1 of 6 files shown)


tools/dispatch_data.py
```diff
@@ -0,0 +1,116 @@
...
+    with open(os.path.join(args.in_dir, "meta.json")) as schema:
+        schema_map = json.load(schema)
+
+    num_parts = len(schema_map["num_nodes_per_chunk"][0])
+    graph_name = schema_map["graph_name"]
+
+    argslist = ""
+    for cmd_arg in UDF_ARGS: 
+        if cmd_arg == "world-size": 
+            argslist += "--world-size 4 ".format(num_parts)
```
jermainewang, line 70: World size could be obtained from the `ip_config` file.

tools/dispatch_data.py
```diff
@@ -0,0 +1,116 @@
...
+        elif cmd_arg == "partitions-dir":
+            argslist += "--partitions-dir {} ".format(args.partitions_dir)
+        elif cmd_arg == "input-dir":
+            argslist += "--input-dir {} ".format(args.in_dir)
+        elif cmd_arg == "graph-name":
+            argslist += "--graph-name {} ".format(graph_name)
+        elif cmd_arg == "schema":
+            argslist += "--schema meta.json "
```
jermainewang, line 78: schema name is always `<data_root_dir>/meta.json`.

tools/dispatch_data.py
```diff
@@ -0,0 +1,116 @@
...
+    os.system(launch_cmd)
+
+def main():
+    parser = argparse.ArgumentParser(description='Launch a distributed job')
+
+    parser.add_argument('--in-dir', type=str, help='Location of the input directory where the dataset is located')
+    parser.add_argument('--partitions-dir', type=str, help='Location of the partition-id mapping files which define node-ids and their respective partition-ids')
+    parser.add_argument('--out-dir', type=str, help='Location of the output directory where the graph partitions will be created by this pipeline')
+    parser.add_argument('--ip-config', type=str, help='File location of IP configuration for server processes')
+    parser.add_argument('--master-port', type=int, help='port used by gloo group to create randezvous point')
```
jermainewang, line 101: Could you give a default value to `master-port`?

tools/dispatch_data.py
```diff
@@ -0,0 +1,116 @@
...
+UDF_OUT_DIR = "output"
+
+LARG_PROCS_MACHINE = "num_proc_per_machine"
+LARG_IPCONF = "ip_config"
+LARG_MASTER_PORT = "master_port"
+
+def get_launch_cmd(args, user_python_exe) -> str: 
+
+    cmd = pythonpath=os.environ.get("PYTHONPATH", "") + get_python_path(user_python_exe) + " " + os.path.join(INSTALL_DIR, LAUNCH_SCRIPT)
+    cmd = f"{cmd} --{LARG_PROCS_MACHINE} 1 "
```
jermainewang, line 27: This could be exposed as the main argument of `dispatch_data.py` with a proper default value. One question. Say we have 4 partitions. Is it true that the world size must be 4 but `num_proc_per_machine` could be arbitrary? In other words, any restriction on this?

tools/distgraphlaunch.py
```diff
@@ -0,0 +1,433 @@
+"""Launching tool for DGL distributed training"""
```
jermainewang, line 1: I suppose this file is mostly copied from `launch.py` so I skip it.

tools/dispatch_data.py
```diff
@@ -0,0 +1,116 @@
...
+    return python_bin
+
+
+def submit_jobs(args, user_python_exe) -> str:
+    wrapper_command = os.path.join(INSTALL_DIR, LAUNCH_SCRIPT)
+
+    UDF_ARGS = [UDF_WORLD_SIZE, UDF_PART_DIR, UDF_INPUT_DIR, UDF_GRAPH_NAME, UDF_SCHEMA, UDF_NUM_PARTS, UDF_OUT_DIR]
+
+    #read the json file and get the remaining argument here. 
+    with open(os.path.join(args.in_dir, "meta.json")) as schema:
```
jermainewang, line 61: `"meta.json"` -> `"metadata.json"`

tools/dispatch_data.py
```diff
@@ -0,0 +1,116 @@
+"""Launching distributed graph partitioning pipeline """
+import os
+import sys
+import argparse
+import logging
+import json
+
+INSTALL_DIR = "/home/ubuntu/workspace/DistDgl-Utils-updatedfileformat"
```
zheng-da, line 8: why do we have hard coded path like this?

tools/dispatch_data.py
zheng-da, line 70: what is this? do you mean `--world-size {}`

tools/dispatch_data.py
```diff
@@ -0,0 +1,116 @@
...
+        if cmd_arg == "world-size": 
+            argslist += "--world-size 4 ".format(num_parts)
+        elif cmd_arg == "partitions-dir":
+            argslist += "--partitions-dir {} ".format(args.partitions_dir)
+        elif cmd_arg == "input-dir":
+            argslist += "--input-dir {} ".format(args.in_dir)
+        elif cmd_arg == "graph-name":
+            argslist += "--graph-name {} ".format(graph_name)
+        elif cmd_arg == "schema":
+            argslist += "--schema metadata.json "
```
zheng-da, line 78: here you use `metadata.json`. did you test your code?

tools/dispatch_data.py
```diff
@@ -0,0 +1,116 @@
...
+
+    #read the json file and get the remaining argument here. 
+    with open(os.path.join(args.in_dir, "meta.json")) as schema:
+        schema_map = json.load(schema)
+
+    num_parts = len(schema_map["num_nodes_per_chunk"][0])
+    graph_name = schema_map["graph_name"]
+
+    argslist = ""
+    for cmd_arg in UDF_ARGS:
```
zheng-da, line 68: i don't know why the for loop makes sense here. how is it different from the code following:
```
argslist += "--world-size {} ".format(num_parts)
argslist += "--partitions-dir {} ".format(args.partitions_dir)
...
```
The code above is much easier.
`UDF_ARGS` is hard-coded. all arguments will appear in the arg list.

tools/distgraphlaunch.py
```diff
@@ -0,0 +1,433 @@
...
+import os
```
zheng-da, line 2: why should we have separate executable scripts to launch partition jobs?
only one script is necessary.

jimjlu99:
Based on the feedback, Ray will refactor the change and initiate a new pull request.

tools/distgraphlaunch.py
BarclayII: `distgraphlaunch.py` is a general script to launch any distributed job including training.  `dispatch_data.py` will run `distgraphlaunch.py` for data dispatching specifically.  So users only need to run `dispatch_data.py`.

tools/dispatch_data.py
BarclayII: I think it must be 1.

BarclayII:
@kylasa I just noticed that the `inner_edge` mask seems to be all one in the `dist_part` branch:
https://github.com/dmlc/dgl/blob/60bc0b7692c6733dd930cd4a502f4171720da38d/tools/distpartitioning/convert_partition.py#L232
I think we should assign True if and only if both incident nodes are the inner nodes (@zheng-da correct me if I'm wrong).
Other than that, I think the PR is ready.

tools/dispatch_data.py
```diff
@@ -0,0 +1,86 @@
...
+
+def main():
+    parser = argparse.ArgumentParser(description='Dispatch edge index and data to partitions', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    parser.add_argument('--in-dir', type=str, help='Location of the input directory where the dataset is located')
+    parser.add_argument('--partitions-dir', type=str, help='Location of the partition-id mapping files which define node-ids and their respective partition-ids, relative to the input directory')
+    parser.add_argument('--out-dir', type=str, help='Location of the output directory where the graph partitions will be created by this pipeline')
+    parser.add_argument('--ip-config', type=str, help='File location of IP configuration for server processes')
+    parser.add_argument('--master-port', type=int, default=12345, help='port used by gloo group to create randezvous point')
+    parser.add_argument('--python-exe', type=str, default=sys.executable, help='Path to the Python executable on all workers')
```
jermainewang, line 71:
```suggestion
    parser.add_argument('--python-path', type=str, default=sys.executable, help='Path to the Python executable on all workers')
```

jermainewang (approved):

jermainewang:
> @kylasa I just noticed that the `inner_edge` mask seems to be all one in the `dist_part` branch:
> 
> https://github.com/dmlc/dgl/blob/60bc0b7692c6733dd930cd4a502f4171720da38d/tools/distpartitioning/convert_partition.py#L232
> 
> 
> I think we should assign True if and only if both incident nodes are the inner nodes (@zheng-da correct me if I'm wrong).
> Other than that, I think the PR is ready.

I think this is correct because currently all the edges assigned to a partition are inner edges. This strategy may resort to changes in the future though.

Issue #3975, opened 2022-05-05
InfoGraph example fails on GPU


melnimr:
## 🐛 Bug

Running the InfoGraph example on GPU fails.

```
   return th.repeat_interleave(input, repeats, dim) # PyTorch 1.1
RuntimeError: repeats must have the same size as input along dim
```

All I did is run:

```
 python infograph/semisupervised.py --gpu 0 --target mu
```

## To Reproduce

1. Go to DGL/examples folder
2. Run semisupervised eample

Traceback (most recent call last):
  File "semisupervised.py", line 217, in <module>
    for sup_data, unsup_data in zip(train_loader, unsup_loader):
  File "/home/neo/wellth-wrk/env/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 530, in __next__
    data = self._next_data()
  File "/home/neo/wellth-wrk/env/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 570, in _next_data
    data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
  File "/home/neo/wellth-wrk/env/lib/python3.8/site-packages/torch/utils/data/_utils/fetch.py", line 52, in fetch
    return self.collate_fn(data)
  File "semisupervised.py", line 116, in collate
    graph_id = dgl.broadcast_nodes(batched_graph, graph_id)
  File "/home/neo/wellth-wrk/env/lib/python3.8/site-packages/dgl/readout.py", line 418, in broadcast_nodes
    return F.repeat(graph_feat, graph.batch_num_nodes(ntype), dim=0)
  File "/home/neo/wellth-wrk/env/lib/python3.8/site-packages/dgl/backend/pytorch/tensor.py", line 189, in repeat
    return th.repeat_interleave(input, repeats, dim) # PyTorch 1.1
RuntimeError: repeats must have the same size as input along dim

## Expected behavior

Code runs and finishes training.

## Environment

 - DGL Version (e.g., 1.0): 0.6.1
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):1.11.0
 - OS (e.g., Linux): Ubuntu
 - How you installed DGL (`conda`, `pip`, source): PIP
 - Build command you used (if compiling from source):
 - Python version: 3.8
 - CUDA/cuDNN version (if applicable):  11.4
 - GPU models and configuration (e.g. V100): Titan RTX
 - Any other relevant information:

Rhett-Ying:
[...]

Issue #3349, opened 2021-09-13
results error when using dgl.broadcast_nodes


ChenChenGith:
## To Reproduce

I test the example in the docs: 
![image](https://user-images.githubusercontent.com/45627234/133062126-6e16dccd-1bfe-42b6-a755-e52a53df7121.png)

Here are my output:
![image](https://user-images.githubusercontent.com/45627234/133061930-d057e52a-a4fa-4d61-867f-6efb33aec692.png)

the output shape should be (2,5), but now it's (1, 10)
## Environment

 - DGL Version (e.g., 1.0): 0.6.1
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): torch 1.9.0
 - OS (e.g., Linux): win
 - How you installed DGL (`conda`, `pip`, source):  pip install
 - Build command you used (if compiling from source):
 - Python version: 3.7
 - CUDA/cuDNN version (if applicable): no cuda
 - GPU models and configuration (e.g. V100): 
 - Any other relevant information:

jwyyy:
It seems the problem is: `feat[0]` is a 1-d tensor, and using `torch.repeat_interleave()` only extend it horizontally. Please see these two references: [dgl.broadcast_nodes](https://github.com/dmlc/dgl/blob/195f99362d883f8b6d131b70a7868a537e55b786/python/dgl/readout.py#L416) and [torch.repeat_interleave](https://pytorch.org/docs/stable/generated/torch.repeat_interleave.html). I have created a PR https://github.com/dmlc/dgl/pull/3351 for this. But for now, you can add a new dimension to your 1-d tensor before using `dgl.broadcast_nodes`.

PR #4298, opened 2022-07-26
[Example][Bugfix] Fix infograph example

chang-l:
## Description

To fix https://github.com/dmlc/dgl/issues/3975, the issue is due to this PR: https://github.com/dmlc/dgl/pull/3351

It crashed because of the checking condition `if len(F.shape(graph_feat)) == 1`, which indicates input feature is a 1-dim vector. 

However, 1-dim vector of `graph_feat` may be attributed to two scenarios  (1) when `batch_size = 1`, and then write the input `feat[0]` to `broadcast_nodes()` in a reduced-dimensionality way, e.g., like in our [doc](https://github.com/dmlc/dgl/blob/master/python/dgl/readout.py#:~:text=%3E%3E%3E%20dgl.broadcast_nodes(g1%2C%20feat%5B0%5D)); (2) each graph feature is a scalar (like graph id) and we are batching `graph_id` to constrcut `graph_feat` like [in this example](https://github.com/dmlc/dgl/blob/master/examples/pytorch/infograph/semisupervised.py#:~:text=graph_id%20%3D%20th.arange(n_graphs)).

**For scenario 1, we need to increase the dimensionality by calling `unsqueeze()`; however, for scenario 2, we should not add more dimension.** 

The proposed fix in here is changing the condition to `if len(F.shape(graph_feat)) == 1 and graph.batch_size == 1`. 

However, I think maybe a better idea is reverting the changes in this PR: https://github.com/dmlc/dgl/pull/3351; and update our Doc to avoid inputing reducd-dimensionality tensor `feat[0]`.

## Checklist
- [x] Code is well-documented

## Changes
`if len(F.shape(graph_feat)) == 1`   --> `if len(F.shape(graph_feat)) == 1 and graph.batch_size == 1`  in readout.py

## Results
`python semisupervised.py --gpu 0 --target mu --epochs 1`
```
======== target = mu ========
Epoch: 0, Sup_Loss: 21.652762, Unsup_loss: 0.0466, Consis_loss: 0.0220
Epoch: 0, LR: 0.001, val_error: 2.0564, best_test_error: 2.0551

```

`python semisupervised.py --gpu 0 --target mu --epochs 1 --batch_size 1`
```
======== target = mu ========
Epoch: 0, Sup_Loss:  nan, Unsup_loss: nan, Consis_loss: nan
Epoch: 0, LR: 0.001, val_error: nan, best_test_error: inf

```


mufeili:
Will reverting #3351 bring #3349 back?

mufeili:
Also we might need to add more unit tests in CI

chang-l:
> Will reverting #3351 bring #3349 back?

Yes, you are right, it will. But we can update Doc instead of updating source code, I think.

chang-l:
> Also we need to add more unit tests in CI

I agree!

chang-l:
When I revisit the function `broadcast_nodes()`, I think I found a better way to resolve this issue. As I also updated the Doc in this PR, let me ++ @jermainewang @mufeili @BarclayII @TristonC @yaox12  for comments. 

The problem is/was: we know the dimension of the output tensor after calling `broadcast_nodes()`, is `(batch_number, node_feat_dimension)`, or in other words, `out_dim = dim(node_feat) + 1`.  But we don't know the dimension(s)/structure of the input tensor `graph_feat`. 

There are 3 possible scenarios: 
- Input `graph_feat` has dimension : `(batch_number, node_feat_dimension)`
   This is a general case, and we don't need to `unsqueeze` to expand the dimension.
- Input `graph_feat` has dimension : `(1, node_feat_dimension)`
   This is when we have a single graph. And we don't need to `unsqueeze` either.
   _Example_: when batch_size =1, [here](https://github.com/dmlc/dgl/blob/4f7972957b02088b13c9d23d71646d2d0bab3e4e/python/dgl/nn/pytorch/glob.py#L584-L585) and [here](https://github.com/dmlc/dgl/blob/4f7972957b02088b13c9d23d71646d2d0bab3e4e/python/dgl/nn/pytorch/conv/chebconv.py#L118-L122) and in [this InfoGraph example (node_feat_dimension=0) ](https://github.com/dmlc/dgl/blob/4f7972957b02088b13c9d23d71646d2d0bab3e4e/examples/pytorch/infograph/semisupervised.py#L115-L116) will generate `(1, node_feat_dimension)` for the input of `broadcast_nodes()`

- Input `graph_feat` has dimension : `(node_feat_dimension)`
   This is when we need to `unsqueeze` the input `graph_feat` to add a new dimension. However, this scenario only shows in the Doc, like [here](https://github.com/dmlc/dgl/blob/4f7972957b02088b13c9d23d71646d2d0bab3e4e/python/dgl/readout.py#L406-L408). And unfortunately, [here ](https://github.com/dmlc/dgl/blob/4f7972957b02088b13c9d23d71646d2d0bab3e4e/python/dgl/readout.py#L368-L369) in the description, it asks users to provide this reduced-dimension tensor: `(node_feat_dimension)`, when single graph (`batch_size =1`), which is unfortunately inconsistent to the most scenarios in our own internal scripts (see above) when/if dealing with single graph. 

As a result, I think one solution is: revert the changes in https://github.com/dmlc/dgl/pull/3351/ and update the Doc accordingly.

Pros: very clean solution, no need to update any internal and example scripts
Cons: it would be user's responsibility to get the input dimension right when single graph scenario, e.g., unsqueeze by themselves. But we can alleviate it by improving/updating our Doc to help them.

mufeili:
> When I revisit the function `broadcast_nodes()`, I think I found a better way to resolve this issue. As I also updated the Doc in this PR, let me ++ @jermainewang @mufeili @BarclayII @TristonC @yaox12 for comments.
> 
> The problem is: we know the dimension of the output tensor after calling `broadcast_nodes()`, is `(batch_number, node_feat_dimension)`, or in other words, `out_dimension = dim(node_feat_dimension) + 1`. But we don't know the dimension(s) and structure of the input tensor `graph_feat`.
> 
> As far as I know, there exists 3 possible structures for the input tensor `graph_feat`:
> 
> * Input `graph_feat` with the dimension: `(batch_number, node_feat_dimension)`
>   This is a general case, and we don't need to `unsqueeze` to add one more dimension.
> * Input `graph_feat` with the dimension : `(1, node_feat_dimension)` if single graph
>   This is when we have a single graph. And we don't need to `unsqueeze` either.
>   _Example_: when batch_size =1, [here](https://github.com/dmlc/dgl/blob/4f7972957b02088b13c9d23d71646d2d0bab3e4e/python/dgl/nn/pytorch/glob.py#L584-L585) and [here](https://github.com/dmlc/dgl/blob/4f7972957b02088b13c9d23d71646d2d0bab3e4e/python/dgl/nn/pytorch/conv/chebconv.py#L118-L122) and in [this InfoGraph example (node_feat_dimension=0) ](https://github.com/dmlc/dgl/blob/4f7972957b02088b13c9d23d71646d2d0bab3e4e/examples/pytorch/infograph/semisupervised.py#L115-L116) will generate `(1, node_feat_dimension)` for the input of `broadcast_nodes()`
> * Input `graph_feat` with the dimension : `(node_feat_dimension)`, and it must be a single-graph batch
>   This is when we need to `unsqueeze` the input `graph_feat` to add a new dimension to prepare for the broadcast. However, this scenario only shows in the Doc, like [here](https://github.com/dmlc/dgl/blob/4f7972957b02088b13c9d23d71646d2d0bab3e4e/python/dgl/readout.py#L406-L408). Also, [here ](https://github.com/dmlc/dgl/blob/4f7972957b02088b13c9d23d71646d2d0bab3e4e/python/dgl/readout.py#L368-L369) in the description, it asks users to provide this reduced-dimension tensor: `(node_feat_dimension)`, when single graph (`batch_size =1`), which is unfortunately inconsistent to the most scenarios in our own internal scripts (see above) when dealing with single graph.
> 
> Overall, I think one solution could be: revert the changes in #3351 and update the Doc accordingly, so that the `broadcast` functions never change the dimensionality of `graph_feat`. To sum up, with this PR:
> 
> * Pros: a clean solution, no need to update any existing places calling `broadcast_nodes()` or change the function interface
> * Cons: it would be user's responsibility to get the input dimension right when single graph scenario, e.g., unsqueeze by themselves. But we can alleviate it by improving/updating our Doc to help them.

Are both structure 2 and structure 3 supported now?

chang-l:
> Are both structure 2 and structure 3 supported now?

Structure 2 is supported now. Structure 3, now should never happen, i.e., users should not input a reduced-dimension tensor when their `batch_size =1`. 
It would be nice to support structure 3 together with structure 2. But I don't see how to distinguish the two structures in the code, unless we know the dimension # of node features in advance... In other words, we know the dimension of input `graph_feat`, but we don't know whether it already included batch_id dimension or just pure node_feature_dimension (when batch_size=1)

Of course we can add another input argument, say `node_feat_dim` to make this function flexible to both structure 2&3 (I am Okay with it BTW). Then, we need to update many internal source scripts when calling `broadcast_*` functions. 

But I think it's not sufficient/solid to put checking conditions like: `len(shape(graph_feat))==1` (# of dim = 1)  and/or `batch_size==1` to satisfy the actual desired condition  `dim(graph_feat) == dim(node_feature_dim)`.

mufeili:
I'm still more inclined to support both structure 2 and structure 3 like below

if g.batch_size == 1 and F.shape(graph_feat)[0] != 1:
        graph_feat = F.unsqueeze(graph_feat, dim=0)

Does this work?

I agree with you that it's better structure 3 has never been supported. However since it's been supported, I don't want to break the backward compatibility and harm the user experience.

jermainewang:
M2C:
* Think about `broadcast_nodes` and `sum_nodes` together. They are mutual-derivative operators, so their treatment of input/output shapes should be aligned.
* The underlying tensor operator of `sum_nodes` is `segment_sum`. Although PyTorch currently does not provide such op, we can take a look at [TensorFlow's `segment_sum`](https://www.tensorflow.org/api_docs/python/tf/math/segment_sum) op for reference.
* It seems that the debate is mostly about a corner case (where there is only one single graph). For this reason, slightly breaking  backward compatibility is fine as long as (1) the docstring and code behavior are consistent and (2) error message is clear about what went wrong and how users should fix it.

mufeili:
> M2C:
> 
> * Think about `broadcast_nodes` and `sum_nodes` together. They are mutual-derivative operators, so their treatment of input/output shapes should be aligned.
> * The underlying tensor operator of `sum_nodes` is `segment_sum`. Although PyTorch currently does not provide such op, we can take a look at [TensorFlow's `segment_sum`](https://www.tensorflow.org/api_docs/python/tf/math/segment_sum) op for reference.
> * It seems that the debate is mostly about a corner case (where there is only one single graph). For this reason, slightly breaking  backward compatibility is fine as long as (1) the docstring and code behavior are consistent and (2) error message is clear about what went wrong and how users should fix it.

Rather than throw an error message, I generally prefer raising a deprecation warning first and then deprecate it after waiting for a few more versions.

chang-l:
@mufeili 
> 
> if g.batch_size == 1 and F.shape(graph_feat)[0] != 1: graph_feat = F.unsqueeze(graph_feat, dim=0)
> 
> Does this work?
>
It will not be a complete fix. For example, when node feature is with shape (1, *) and batch_size = 1, this if condition wouldn't be able to check/distinguish the two single graph scenarios: (1) input shape of `graph_feat` is (1, 1, *) with batching dimension;  or (2) input shape of `graph_feat` is (1, *). And the condition `F.shape(graph_feat)[0] != 1` is always `False`.

@jermainewang 
> M2C:
> 
> * Think about `broadcast_nodes` and `sum_nodes` together. They are mutual-derivative operators, so their treatment of input/output shapes should be aligned.
> * The underlying tensor operator of `sum_nodes` is `segment_sum`. Although PyTorch currently does not provide such op, we can take a look at [TensorFlow's `segment_sum`](https://www.tensorflow.org/api_docs/python/tf/math/segment_sum) op for reference.
> * It seems that the debate is mostly about a corner case (where there is only one single graph). For this reason, slightly breaking  backward compatibility is fine as long as (1) the docstring and code behavior are consistent and (2) error message is clear about what went wrong and how users should fix it.

I agree we should add error/warning messages to make users more comfortable. More important, the docstring should be consistent and clearly specified on the difference w.r.t. the previous version.

> Rather than throw an error message, I generally prefer raising a deprecation warning first and then deprecate it after waiting for a few more versions.

Yeah... Now, it seems we agree on we should inform users keeping the input dimension/shape consistent for single graph. 
Adding a "perfect" error/warning message to cover all outliers is as hard as making the code flexible to all single graph scenarios. Here is my attempt (this version):
(1) Changed docstring to make it more clear that users should use `(1, *)` as the input dimension for single graph
(2) Added dgl warning msg when `F.shape(graph_feat)[0] != graph.batch_size` and `batch_size == 1`; yet we still do the `unsqueeze` for them. But as I mentioned, this warning msg is not guaranteed to always show up when `graph_feat` missing the first batching dimension. For example, when node feature with shape (1,*) and single graph, users have to make sure their input is in a correct shape (1, 1, *) and if not, there will be no warning message printed.

mufeili:
> (2) Added dgl warning msg when F.shape(graph_feat)[0] != graph.batch_size and batch_size == 1; yet we still do the unsqueeze for them. But as I mentioned, this warning msg is not guaranteed to always show up when graph_feat missing the first batching dimension. For example, when node feature with shape (1,*) and single graph, users have to make sure their input is in a correct shape (1, 1, *) and if not, there will be no warning message printed.

That's fine.

python/dgl/readout.py
```diff
@@ -403,17 +404,21 @@ def broadcast_nodes(graph, graph_feat, *, ntype=None):
...
             [0.4325, 0.7710, 0.5541, 0.0544, 0.9368]])
 
     See Also
     --------
     broadcast_edges
     """
-    if len(F.shape(graph_feat)) == 1:
+    if (F.shape(graph_feat)[0] != graph.batch_size and graph.batch_size == 1):
+        dgl_warning('If single graph, please make sure the batching dimension is added'
+                    ' and the tensor shape of graph_feat is (1, *).')
```
mufeili, line 421: Add a deprecation warning that the support will be dropped in the future.

chang-l: Updated.

python/dgl/readout.py
```diff
@@ -417,8 +417,9 @@ def broadcast_nodes(graph, graph_feat, *, ntype=None):
     broadcast_edges
     """
     if (F.shape(graph_feat)[0] != graph.batch_size and graph.batch_size == 1):
-        dgl_warning('If single graph, please make sure the batching dimension is added'
-                    ' and the tensor shape of graph_feat is (1, *).')
+        dgl_warning(':func:`broadcast_nodes` directly taking one node feature tensor as `feat[i]`'
```
mufeili, line 420: No need to have `:func:`. This only helps for rendering the documentation with sphinx.

python/dgl/readout.py
```diff
@@ -488,8 +489,9 @@ def broadcast_edges(graph, graph_feat, *, etype=None):
     broadcast_nodes
     """
     if (F.shape(graph_feat)[0] != graph.batch_size and graph.batch_size == 1):
-        dgl_warning('If single graph, please make sure the batching dimension is added'
-                    ' and the tensor shape of graph_feat is (1, *).')
+        dgl_warning(':func:`broadcast_edges` directly taking one edge feature tensor as `feat[i]`'
```
mufeili, line 492: Ditto

python/dgl/readout.py
```diff
@@ -417,8 +417,9 @@ def broadcast_nodes(graph, graph_feat, *, ntype=None):
     broadcast_edges
     """
     if (F.shape(graph_feat)[0] != graph.batch_size and graph.batch_size == 1):
-        dgl_warning('If single graph, please make sure the batching dimension is added'
-                    ' and the tensor shape of graph_feat is (1, *).')
+        dgl_warning(':func:`broadcast_nodes` directly taking one node feature tensor as `feat[i]`'
+                    ' to broadcast to all nodes in a single graph will be deprecated in future,'
+                    ' please make sure the feature tensor shape is :math:`(1, *).`')
```
mufeili, line 422: Perhaps you can say

"""
For a single graph, use a tensor of shape (1, *) for graph_feat. The support of shape (*) will be deprecated.
"""

chang-l: Thank you for you suggestion. It's more concise.

python/dgl/readout.py
chang-l: Thanks.

python/dgl/readout.py
chang-l: I see.. Thanks.

mufeili (approved):
I'm good. Thank you.

PR #4306, opened 2022-07-27
[Example] Fixed device type in GraphSAGE inference

wey-gu:
## Description

The device here was typed in the string, which ended up pin_memory being false anyway to lead to issues in the CPU device case.

https://github.com/dmlc/dgl/blob/2cf05c53420546443a267d91ec624887d0078278/examples/pytorch/graphsage/node_classification.py#L42-L43

## Checklist

## Changes

Fixed the example code for torch-GraphSAGE


wey-gu:
The UT failure seems irrelevant

yaox12:
I'd prefer to change `buffer_device` to `torch.device('cpu')`.

wey-gu:
> torch.device('cpu')

True, that makes more sense, will do so.

yaox12 (approved):

PR #4311, opened 2022-07-28
[Distributed] use alltoall fix to bypass gloo - alltoallv bug in distributed partitioning

kylasa:
…er testing

1. Replaced alltoallv gloo wrapper call with alltoall message.
2. All the messages are padded to be of same length
3. Receiving side unpads the messages and continues processing.

## Checklist
- [X] Code is well-documented

### DIFF
--- a/tools/distpartitioning/data_shuffle.py
+++ b/tools/distpartitioning/data_shuffle.py
@@ -16,7 +16,8 @@ from utils import read_ntype_partition_files, read_json, get_node_types, \
                     write_dgl_objects, write_metadata_json, get_ntype_featnames, \
                     get_idranges
 from gloo_wrapper import alltoall_cpu_object_lst, alltoallv_cpu, \
-                    alltoall_cpu, allgather_sizes, gather_metadata_json
+                    alltoall_cpu, allgather_sizes, gather_metadata_json,\
+                    alltoallv_cpu_data
 from globalids import assign_shuffle_global_nids_nodes, \
                     assign_shuffle_global_nids_edges, \
                     get_shuffle_global_nids_edges
@@ -140,8 +141,6 @@ def exchange_edge_data(rank, world_size, edge_data):
     """
 
     input_list = []
-    send_sizes = []
-    recv_sizes = []
     start = timer()
     for i in np.arange(world_size):
         send_idx = (edge_data[constants.OWNER_PROCESS] == i)
@@ -152,23 +151,13 @@ def exchange_edge_data(rank, world_size, edge_data):
                                     edge_data[constants.ETYPE_ID][send_idx == 1], \
                                     edge_data[constants.GLOBAL_EID][send_idx == 1]))
         if(filt_data.shape[0] <= 0):
-            input_list.append(torch.empty((0,), dtype=torch.int64))
-            send_sizes.append(torch.empty((0,), dtype=torch.int64))
+            input_list.append(torch.empty((0,5), dtype=torch.int64))
         else:
             input_list.append(torch.from_numpy(filt_data))
-            send_sizes.append(torch.tensor(filt_data.shape, dtype=torch.int64))
-        recv_sizes.append(torch.zeros((2,), dtype=torch.int64))
     end = timer()
     
     dist.barrier ()
-    start = timer()
-    alltoall_cpu(rank, world_size, recv_sizes, send_sizes)
-    output_list = []
-    for s in recv_sizes: 
-        output_list.append(torch.zeros(s.tolist(), dtype=torch.int64))
-
-    dist.barrier ()
-    alltoallv_cpu(rank, world_size, output_list, input_list)
+    output_list = alltoallv_cpu_data(rank, world_size, input_list,torch.int64)
     end = timer()
     print('[Rank: ', rank, '] Time to send/rcv edge data: ', timedelta(seconds=end-start))
 
@@ -246,6 +235,7 @@ def exchange_node_features(rank, world_size, node_feature_tids, ntype_gnid_map, 
             global_nid_per_rank = []
             feat_name = feat_info[0]
             feat_key = ntype_name+'/'+feat_name
+            print('[Rank: ', rank, '] processing node feature: ', feat_key)
 
             #compute the global_nid range for this node features
             type_nid_start = int(feat_info[1])
@@ -273,29 +263,25 @@ def exchange_node_features(rank, world_size, node_feature_tids, ntype_gnid_map, 
                 local_idx_partid = local_idx[cond]
 
                 if (gnids_per_partid.shape[0] == 0):
-                    node_feats_per_rank.append({feat_key : torch.empty((0,), dtype=torch.float)})
-                    global_nid_per_rank.append({feat_key : torch.empty((0,), dtype=torch.int64)})
+                    node_feats_per_rank.append(torch.empty((0,1), dtype=torch.float))
+                    global_nid_per_rank.append(np.empty((0,1), dtype=np.int64))
                 else:
-                    node_feats_per_rank.append({feat_key : node_feats[local_idx_partid]})
-                    global_nid_per_rank.append({feat_key : gnids_per_partid})
+                    if (len(list(node_feats[local_idx_partid].size())) == 1):
+                        node_feats_per_rank.append(node_feats[local_idx_partid])
+                    else:
+                        node_feats_per_rank.append(node_feats[local_idx_partid])
+                    global_nid_per_rank.append(torch.from_numpy(gnids_per_partid).type(torch.int64))
 
             #features (and global nids) per rank to be sent out are ready
             #for transmission, perform alltoallv here.
-            output_feat_list = alltoall_cpu_object_lst(rank, world_size, node_feats_per_rank)
-            output_feat_list[rank] = node_feats_per_rank[rank]
+            output_feat_list = alltoallv_cpu_data(rank, world_size, node_feats_per_rank, node_feats.dtype)
 
-            output_nid_list = alltoall_cpu_object_lst(rank, world_size, global_nid_per_rank)
-            output_nid_list[rank] = global_nid_per_rank[rank]
+            #output_nid_list = alltoall_cpu_object_lst(rank, world_size, global_nid_per_rank)
+            output_nid_list = alltoallv_cpu_data(rank, world_size, global_nid_per_rank, torch.int64)
 
             #stitch node_features together to form one large feature tensor
-            own_node_features[feat_key] = []
-            own_global_nids[feat_key] = []
-            for idx, x in enumerate(output_feat_list):
-                own_node_features[feat_key].append(x[feat_key])
-                own_global_nids[feat_key].append(output_nid_list[idx][feat_key])
-            for k in own_node_features.keys(): 
-                own_node_features[k] = torch.cat(own_node_features[k])
-                own_global_nids[k] = np.concatenate(own_global_nids[k])
+            own_node_features[feat_key] = torch.cat(output_feat_list)
+            own_global_nids[feat_key] = torch.cat(output_nid_list).numpy()
 
     end = timer()
     print('[Rank: ', rank, '] Total time for node feature exchange: ', timedelta(seconds = end - start))
--- a/tools/distpartitioning/data_shuffle.py
+++ b/tools/distpartitioning/data_shuffle.py
@@ -15,7 +15,7 @@ from utils import read_ntype_partition_files, read_json, get_node_types, \
                     augment_edge_data, get_gnid_range_map, \
                     write_dgl_objects, write_metadata_json, get_ntype_featnames, \
                     get_idranges
-from gloo_wrapper import alltoall_cpu, allgather_sizes, gather_metadata_json,\
+from gloo_wrapper import allgather_sizes, gather_metadata_json,\
                     alltoallv_cpu
 from globalids import assign_shuffle_global_nids_nodes, \
                     assign_shuffle_global_nids_edges, \
--- a/tools/distpartitioning/data_shuffle.py
+++ b/tools/distpartitioning/data_shuffle.py
@@ -15,8 +15,7 @@ from utils import read_ntype_partition_files, read_json, get_node_types, \
                     augment_edge_data, get_gnid_range_map, \
                     write_dgl_objects, write_metadata_json, get_ntype_featnames, \
                     get_idranges
-from gloo_wrapper import alltoall_cpu_object_lst, alltoallv_cpu, \
-                    alltoall_cpu, allgather_sizes, gather_metadata_json,\
+from gloo_wrapper import alltoall_cpu, allgather_sizes, gather_metadata_json,\
                     alltoallv_cpu_data
 from globalids import assign_shuffle_global_nids_nodes, \
                     assign_shuffle_global_nids_edges, \
@@ -157,7 +156,7 @@ def exchange_edge_data(rank, world_size, edge_data):
     end = timer()
     
     dist.barrier ()
-    output_list = alltoallv_cpu_data(rank, world_size, input_list,torch.int64)
+    output_list = alltoallv_cpu_data(rank, world_size, input_list)
     end = timer()
     print('[Rank: ', rank, '] Time to send/rcv edge data: ', timedelta(seconds=end-start))
 
@@ -266,18 +265,13 @@ def exchange_node_features(rank, world_size, node_feature_tids, ntype_gnid_map, 
                     node_feats_per_rank.append(torch.empty((0,1), dtype=torch.float))
                     global_nid_per_rank.append(np.empty((0,1), dtype=np.int64))
                 else:
-                    if (len(list(node_feats[local_idx_partid].size())) == 1):
-                        node_feats_per_rank.append(node_feats[local_idx_partid])
-                    else:
-                        node_feats_per_rank.append(node_feats[local_idx_partid])
+                    node_feats_per_rank.append(node_feats[local_idx_partid])
                     global_nid_per_rank.append(torch.from_numpy(gnids_per_partid).type(torch.int64))
 
             #features (and global nids) per rank to be sent out are ready
             #for transmission, perform alltoallv here.
-            output_feat_list = alltoallv_cpu_data(rank, world_size, node_feats_per_rank, node_feats.dtype)
-
-            #output_nid_list = alltoall_cpu_object_lst(rank, world_size, global_nid_per_rank)
-            output_nid_list = alltoallv_cpu_data(rank, world_size, global_nid_per_rank, torch.int64)
+            output_feat_list = alltoallv_cpu_data(rank, world_size, node_feats_per_rank)
+            output_nid_list = alltoallv_cpu_data(rank, world_size, global_nid_per_rank)
 
             #stitch node_features together to form one large feature tensor
             own_node_features[feat_key] = torch.cat(output_feat_list)
--- a/tools/distpartitioning/data_shuffle.py
+++ b/tools/distpartitioning/data_shuffle.py
@@ -185,6 +185,9 @@ def exchange_node_features(rank, world_size, node_feature_tids, ntype_gnid_map, 
         retrieved. 
     d. After receiving the corresponding shuffle_global_nids these ids are added to the 
         node_data and edge_data dictionaries
+    
+    This pipeline assumes all the input data in numpy format, except node/edge features which
+    are maintained as tensors throughout the various stages of the pipeline execution.
 
     Parameters: 
     -----------
--- a/tools/distpartitioning/data_shuffle.py
+++ b/tools/distpartitioning/data_shuffle.py
@@ -16,7 +16,7 @@ from utils import read_ntype_partition_files, read_json, get_node_types, \
                     write_dgl_objects, write_metadata_json, get_ntype_featnames, \
                     get_idranges
 from gloo_wrapper import alltoall_cpu, allgather_sizes, gather_metadata_json,\
-                    alltoallv_cpu_data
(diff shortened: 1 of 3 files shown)


tools/distpartitioning/data_shuffle.py
```diff
@@ -273,29 +263,25 @@ def exchange_node_features(rank, world_size, node_feature_tids, ntype_gnid_map,
...
 
                 if (gnids_per_partid.shape[0] == 0):
-                    node_feats_per_rank.append({feat_key : torch.empty((0,), dtype=torch.float)})
-                    global_nid_per_rank.append({feat_key : torch.empty((0,), dtype=torch.int64)})
+                    node_feats_per_rank.append(torch.empty((0,1), dtype=torch.float))
+                    global_nid_per_rank.append(np.empty((0,1), dtype=np.int64))
                 else:
-                    node_feats_per_rank.append({feat_key : node_feats[local_idx_partid]})
-                    global_nid_per_rank.append({feat_key : gnids_per_partid})
+                    if (len(list(node_feats[local_idx_partid].size())) == 1):
```
zheng-da, line 269: what is this?
why not write something like this:
```
if ...:
  ...
elif ...:
  ...
else:
  ...
```

tools/distpartitioning/data_shuffle.py
```diff
@@ -273,29 +263,25 @@ def exchange_node_features(rank, world_size, node_feature_tids, ntype_gnid_map,
...
+                        node_feats_per_rank.append(node_feats[local_idx_partid])
+                    else:
+                        node_feats_per_rank.append(node_feats[local_idx_partid])
```
zheng-da, line 272: what is the difference in the if branch and else branch?

tools/distpartitioning/data_shuffle.py
```diff
@@ -273,29 +263,25 @@ def exchange_node_features(rank, world_size, node_feature_tids, ntype_gnid_map,
...
 
             #features (and global nids) per rank to be sent out are ready
             #for transmission, perform alltoallv here.
-            output_feat_list = alltoall_cpu_object_lst(rank, world_size, node_feats_per_rank)
-            output_feat_list[rank] = node_feats_per_rank[rank]
+            output_feat_list = alltoallv_cpu_data(rank, world_size, node_feats_per_rank, node_feats.dtype)
 
-            output_nid_list = alltoall_cpu_object_lst(rank, world_size, global_nid_per_rank)
-            output_nid_list[rank] = global_nid_per_rank[rank]
+            #output_nid_list = alltoall_cpu_object_lst(rank, world_size, global_nid_per_rank)
```
zheng-da, line 279: remove.

tools/distpartitioning/data_shuffle.py
```diff
@@ -273,29 +263,25 @@ def exchange_node_features(rank, world_size, node_feature_tids, ntype_gnid_map,
...
-            own_node_features[feat_key] = []
-            own_global_nids[feat_key] = []
-            for idx, x in enumerate(output_feat_list):
-                own_node_features[feat_key].append(x[feat_key])
-                own_global_nids[feat_key].append(output_nid_list[idx][feat_key])
-            for k in own_node_features.keys(): 
-                own_node_features[k] = torch.cat(own_node_features[k])
-                own_global_nids[k] = np.concatenate(own_global_nids[k])
+            own_node_features[feat_key] = torch.cat(output_feat_list)
+            own_global_nids[feat_key] = torch.cat(output_nid_list).numpy()
```
zheng-da, line 284: why one is pytorch tensor and the other is numpy array?

tools/distpartitioning/globalids.py
```diff
@@ -29,41 +29,18 @@ def get_shuffle_global_nids(rank, world_size, global_nids_ranks, node_data):
...
-        send_nodes.append(torch.from_numpy(global_nids_ranks[i]).type(dtype=torch.int64))
-
-    #send-recieve messages
-    alltoallv_cpu(rank, world_size, recv_nodes, send_nodes)
-
-    # allocate buffers to receive global-ids
-    recv_shuffle_global_nids = []
-    for i in sizes: 
-        recv_shuffle_global_nids.append(torch.zeros((i), dtype=torch.int64))
+    sizes = [x.shape for x in global_nids_ranks]
```
zheng-da, line 32: where is `sizes` used?

tools/distpartitioning/globalids.py
```diff
@@ -29,41 +29,18 @@ def get_shuffle_global_nids(rank, world_size, global_nids_ranks, node_data):
...
+    global_nids_ranks = [torch.from_numpy(x) for x in global_nids_ranks]
+    recv_nodes = alltoallv_cpu_data(rank, world_size, global_nids_ranks, torch.int64)
 
     # Use node_data to lookup global id to send over.
     send_nodes = []
     for proc_i_nodes in recv_nodes:
         #list of node-ids to lookup
+        if proc_i_nodes == None: 
+            send_nodes.append(torch.empty((0), dtype=torch.int64))
+            continue
```
zheng-da, line 42: is it possible some process sends None?

tools/distpartitioning/globalids.py
```diff
@@ -73,9 +50,8 @@ def get_shuffle_global_nids(rank, world_size, global_nids_ranks, node_data):
             send_nodes.append(torch.empty((0), dtype=torch.int64))
 
     #send receive global-ids
-    alltoallv_cpu(rank, world_size, recv_shuffle_global_nids, send_nodes)
-
-    shuffle_global_nids = np.concatenate([x.numpy() for x in recv_shuffle_global_nids])
+    recv_shuffle_global_nids = alltoallv_cpu_data(rank, world_size, send_nodes, torch.int64)
+    shuffle_global_nids = np.concatenate([x.numpy() if x != None else [] for x in recv_shuffle_global_nids])
```
zheng-da, line 54: should we use `if x is not None`?

zheng-da: did you test the case? if one of the element is `[]` in the input list, does numpy.concatenate works?

tools/distpartitioning/gloo_wrapper.py
```diff
@@ -59,6 +59,74 @@ def alltoall_cpu(rank, world_size, output_tensor_list, input_tensor_list):
...
+    #extract un-padded message from the output_tensor_list and return it
+    return_vals = []
+    for s, t in zip(recv_counts, output_tensor_list):
+        if s[0] == 0:
+            return_vals.append(None)
+        else:
+            return_vals.append(t[0:s[0]])
+    return return_vals
+
 def alltoall_cpu_object_lst(rank, world_size, input_list):
```
zheng-da, line 130: should you delete this?

tools/distpartitioning/gloo_wrapper.py
```diff
@@ -59,6 +59,74 @@ def alltoall_cpu(rank, world_size, output_tensor_list, input_tensor_list):
     for i in range(world_size):
         dist.scatter(output_tensor_list[i], input_tensor_list if i == rank else [], src=i)
 
+def alltoallv_cpu_data(rank, world_size, input_tensor_list, val_dtype):
```
zheng-da, line 62: do you still need to original alltoallv function?

tools/distpartitioning/gloo_wrapper.py
```diff
@@ -59,6 +59,74 @@ def alltoall_cpu(rank, world_size, output_tensor_list, input_tensor_list):
...
+    Parameters:
+    -----------
+    rank : int
+        The rank of current worker
+    world_size : int
+        The size of the entire
+    input_tensor_list : List of tensor
+        The tensors to exchange
+    val_dtype : torch dtype
+        torch dtype which is used to create buffer for receiving messages
```
zheng-da, line 76: is this input argument necessary? the input tensor list should already has it. you should make sure all input tensors have the same dtype.

tools/distpartitioning/gloo_wrapper.py
```diff
@@ -59,6 +59,74 @@ def alltoall_cpu(rank, world_size, output_tensor_list, input_tensor_list):
...
+        list of tensors received from other processes during alltoall message
+
+    """
+    sizes = [list(x.size()) for x in input_tensor_list]
+    for idx in range(1,len(sizes)):
+        assert len(sizes[idx-1]) == len(sizes[idx])
+
+    #decide how much to pad. 
+    #always use the first-dimension for padding. 
+    ll = [ x[0] for x in sizes ]
```
zheng-da, line 90: if the input tensors are matrices, does the function still work? if the input tensors are 3D tensors, does the function still work?
if you want to make it general, you should flatten the input tensor to 1D vector and pad them. otherwise, you should make sure the input tensors are 1D vectors.

tools/distpartitioning/gloo_wrapper.py
```diff
@@ -59,6 +59,74 @@ def alltoall_cpu(rank, world_size, output_tensor_list, input_tensor_list):
...
+
+    #dims of the outgoing messages
+    out_dims = [ [np.amax(ll)] + l[1:] for idx, l in enumerate(sizes) ]
```
zheng-da, line 93: why do you use `enumerate` here?

tools/distpartitioning/gloo_wrapper.py
zheng-da: or do you assume the input tensor may only differ in the first dimension. if so, you should check it explicitly.

tools/distpartitioning/gloo_wrapper.py
```diff
@@ -59,6 +59,74 @@ def alltoall_cpu(rank, world_size, output_tensor_list, input_tensor_list):
...
+    input_tensor_list = [torch.cat((x, torch.zeros(diff_dims[idx]).type(x.dtype))) for idx, x in enumerate(input_tensor_list)]
+
+    #send useful message sizes to all
+    send_counts = []
+    recv_counts = []
+    for idx in range(world_size):
+        #send a 3 element triplet, [a, b, ....] where 
+        #a = useful message dim, b = amount of padding and remaining elements are
+        #the remaining dimensions of the tensor
+        send_counts.append(torch.from_numpy(np.array([sizes[idx][0]] + out_dims[idx])).type(torch.int64))
```
zheng-da, line 108: why the send count in a rank is `[sizes[idx][0]] + out_dims[idx]`?

zheng-da: also it's a bad habit to cast the type in the end. what happens if the input values are too large and cannot be stored in int32

tools/distpartitioning/gloo_wrapper.py
```diff
@@ -59,6 +59,74 @@ def alltoall_cpu(rank, world_size, output_tensor_list, input_tensor_list):
...
+        recv_counts.append(torch.zeros((1 + len(sizes[idx])), dtype=torch.int64))
+    alltoall_cpu(rank, world_size, recv_counts, send_counts)
+
+    #allocate buffers for receiving message
+    output_tensor_list = []
+    recv_counts = [ tsize.numpy() for tsize in recv_counts]
+    for tsize in recv_counts:
+        output_tensor_list.append(torch.zeros(tuple(tsize[1:])).type(val_dtype))
```
zheng-da, line 116: i really don't understand why you need to append some value at the beginning of a shape and not use it afterwards.

tools/distpartitioning/gloo_wrapper.py
```diff
@@ -59,6 +59,74 @@ def alltoall_cpu(rank, world_size, output_tensor_list, input_tensor_list):
...
+
+    #send actual message itself. 
+    alltoall_cpu(rank, world_size, output_tensor_list, input_tensor_list)
```
zheng-da, line 119: the interface of `alltoall_cpu` is not very good. ideally, it should be something like this:
```
output_tensor_list = alltoall_cpu(rank, world_size, input_tensor_list)
```

tools/distpartitioning/gloo_wrapper.py
zheng-da, line 108: how is b the amount of padding? isn't b the total size of the first dimension after padding?

zheng-da: alltoall should send tensors of the same shape, right? out_dims for each rank should be the same?

zheng-da: if so, you only need to exchange `sizes[idx][0]` of each rank.

tools/distpartitioning/data_shuffle.py
kylasa: Agreed. This was the result of the assumption that treating vector of integers as a two dimensional vector and passing it to alltoallv, which was used prior to this fix. Since it is confirmed that this workaround does not work and the behavior is quite unpredictable this code became irrelevant now. But as a side effect the reshaping of vector to two dimensional array has been removed without checking the difference between the "if" case and "else" case in the above piece of code. 

Now we use alltoall for communication, this is no longer needed. And now we just treat the input data to alltoall in its original format and shape. Removing the "if" statement as suggested. 

Nice catch. Thanks so much.

tools/distpartitioning/data_shuffle.py
kylasa: Removed the commented out code as suggested.

tools/distpartitioning/data_shuffle.py
kylasa: Node features are read from the file as tensors, this is how load_tensors function works -- it always returns tensors. The global-nids is a numpy array so this is used as function argument to alltoall_cpu_data this numpy array is explicity converted to tensor and once the results are received they are again converted back to numpy arrays from tensors.

tools/distpartitioning/globalids.py
kylasa: Correct. This variable is not used anymore. Prior to alltoall_cpu_data, when alltoallv was used sizes was used to create appropriate buffers on the receiving side. This is irrelevant now. 

Removed 'sizes'. 

Please note that the logic of creating buffers for receiving messages is created inside the alltoall_cpu_data function call.

tools/distpartitioning/globalids.py
kylasa: This is a possibility, it case happens in "mag" dataset. When cyclic partitioning is used for this dataset. Only process-3 has 1 global-id which needs to be mapped to shuffle-global-id and the owner process is process-0. Rest of the three processes does not have any global-ids which needs to be mapped to shuffle-global-ids.

tools/distpartitioning/globalids.py
kylasa: Code changed to use 'if x is not None:'

Yes this case is specifically tested. This happens in the case of "mag" dataset with cyclic partitioning. The numpy.concatenate works when one of the list elements is '[]'.

tools/distpartitioning/gloo_wrapper.py
kylasa: Removed this function as suggested.

tools/distpartitioning/gloo_wrapper.py
kylasa: Removed the prior function which is a wrapper for providing alltoallv function.

tools/distpartitioning/gloo_wrapper.py
kylasa: The reason behind such flexibility is that provide this option to the user of this function to receive an output_tensor_list which is of type val_dtype. This could be different from the dtype of the tensors in the input_tensor_list. 

I agree that this case is not present in the use cases so far. And this might lead to some confusion. 

I am removing the val_dtype for now and assume that the output_tensor_list is always of the type which is same as in the input tensor list. The code is appropriately changed.

tools/distpartitioning/gloo_wrapper.py
kylasa: This function assumes that the tensors in the input_tensor_list can be of random dimensions. They ALL can be 1-D, 2-D or n-D. But all the tensors should be of the same size. Added 'assert' to ensure that all the input tensors are of same dimensions.

This function also assumes that only the FIRST dimension of the input tensors can differ and padding is needed. Added assert to check this condition as well.

tools/distpartitioning/gloo_wrapper.py
kylasa: Agreed. enumerate is not needed here since idx is not used in this instance. 
Changed the code appropriately.

tools/distpartitioning/gloo_wrapper.py
kylasa: 'b' is the actual message size along the first dimension. The wording in this comment were incorrect. I changed the wording to as mentioned above. 

alltoall should send tensors of same size. Yes out_dims should all be the same. 

sizes[idx][0] is the size of the input tensor along first dimension which is unpadded. out_dim[idx][0] is the size of the input tensor with padding along first dimension. 

Agree. we only need to exchange sizes[idx][0], which is just one integer as opposed to 3 integers. 

After addressing all these comments, and code modified in response to these comments. Following changes are made. 

1. Removed out_dims, as it is not needed. 
2. exchange only sizes[idx][0] which is the actual size along the first dimension (without padding).

tools/distpartitioning/gloo_wrapper.py
kylasa: All the values in the recv_counts list are used appropriately. 
These will be of the form [a, b, ....] as explained in earlier comment. 

[b, ....] is used to create buffer of appropriate size. 
and tensor[0:a] is used to extract unpadded part of the incoming tensor to process the actual message received.

tools/distpartitioning/gloo_wrapper.py
kylasa: Agreed. now this function interface is exactly like you suggested.

tools/distpartitioning/data_shuffle.py
zheng-da: my comment is really why you want to maintain node features as pytorch tensors and nid as numpy arrays. this causes unnecessary complexity for the code. since you are using numpy arrays almost everywhere, why don't you convert it to numpy array?

zheng-da: is the format conversion zero copy?

tools/distpartitioning/globalids.py
zheng-da: i don't see you make such a change.

tools/distpartitioning/gloo_wrapper.py
zheng-da: i don't see the change.

tools/distpartitioning/gloo_wrapper.py
```diff
@@ -59,71 +59,84 @@ def alltoall_cpu(rank, world_size, output_tensor_list, input_tensor_list):
...
+    input_tensor_list = [torch.cat((x, torch.zeros(diff_dims[idx]).type(x.dtype))) for idx, x in enumerate(input_tensor_list)]
+
+    #send useful message sizes to all
+    send_counts = []
+    recv_counts = []
+    for idx in range(world_size):
+        #send a vector, of atleast 3 elements, [a, b, ....] where 
+        #a = useful message dim, b = actual message outgoing message size along the first dimension
+        #and remaining elements are the remaining dimensions of the tensor
+        send_counts.append(torch.from_numpy(np.array([sizes[idx][0]] + [np.amax(ll)] + sizes[idx][1:] )).type(torch.int64))
```
zheng-da, line 119: i don't think `sizes[idx][1:]` don't need to be sent over. all processes know `sizes[idx][1:]`

kylasa: This is on the sending side, where sizes[idx][:1] is same for all the tensors. But on the receiving side, it receives different size'd tensors from everyone else. So potentially the receiving side will receive tensors of dimensions which WILL NOT be of the same size. And for this reason, we need sizes[idx][:1] to be sent to over from the sending process.

tools/distpartitioning/globalids.py
kylasa: Earlier I misunderstood this comment. Now changed the code appropriately to match the description in the CI comment.

tools/distpartitioning/gloo_wrapper.py
kylasa: Renamed this function per our discussion.

tools/distpartitioning/gloo_wrapper.py
kylasa: Added a comment, in the docstring, to indicate the assumption about this comment.

tools/distpartitioning/data_shuffle.py
kylasa: Added a comment, per our discussion, in the docstring.

tools/distpartitioning/globalids.py
```diff
@@ -29,53 +29,27 @@ def get_shuffle_global_nids(rank, world_size, global_nids_ranks, node_data):
...
     for proc_i_nodes in recv_nodes:
         #list of node-ids to lookup
-        global_nids = proc_i_nodes.numpy()
-        if (len(global_nids) != 0):
-            common, ind1, ind2 = np.intersect1d(node_data[constants.GLOBAL_NID], global_nids, return_indices=True)
-            shuffle_global_nids = node_data[constants.SHUFFLE_GLOBAL_NID][ind1]
-            send_nodes.append(torch.from_numpy(shuffle_global_nids).type(dtype=torch.int64))
+        if proc_i_nodes is not None: 
+            global_nids = proc_i_nodes.numpy()
+            if(len(global_nids) != 0):
```
zheng-da, line 41: this actually doesn't meet Python conversion. In python, we should write something like this:
```
if len(global_nids) != 0:
```

zheng-da (approved):

PR #4338, opened 2022-08-05
[Distributed] reduce memory consumption in distributed graph partitioning.

kylasa:
…fault for very large partitions

1. Removed three graph dgl objects and we create the final dgl object directly by maintaining the following constraints
a) nodes are reordered so that local nodes are placed in the beginning of the nodes list compared to non-local nodes.
b)Edges order are maintained as passed into this function.
c) src/dst end points are mapped to target values based on the reshuffle'd nodes order.

## Checklist
- [X] Code is well-documented

### DIFF
--- a/tools/distpartitioning/convert_partition.py
+++ b/tools/distpartitioning/convert_partition.py
@@ -182,29 +182,15 @@ def create_dgl_object(graph_name, num_parts, \
     '''
     Following procedure is used to map the part_local_src_id, part_local_dst_id to account for
     reshuffling of nodes (to order localy owned nodes prior to non-local nodes in a partition)
-    1. Place side-by-side reshuffle_nodes and corresponding index value.▒
-        Basically (reshuffle_node[idx], idx) will form the key-value pair which we have to use
-        for mapping part_local_src_ids, and part_local_dst_ids.
-        Below we use column_stack to generate this mapping, call the resultant map as `node_map`
-
-    2. Sort the node_map array, so that the reshuffle_nodes column are sorted. This is needed because
-        `uniq_ids` are sorted as a result of np.unique operation.
-
-    3. Now, at this time the map is ready to map pre-node-reordering node-ids.▒
-
-    4. Now, concatenate [part_local_src_id, part_local_dst_id, np.arange(len(uniq_ids))] so that we have
-        the same sized array which is used to generate part_local_src_id and part_local_dst_id▒
-        which is the code above this comment section.
-
-    5. Now perform np.unique operation on the resultant list from step (4). This set of uniq_values
-        should follow the same order as `idx (uniq_ids)` above.▒
-
-    6. Now after this step, uniq_mapped_vals and uniq_ids should be the mapping between shuffle_global_nids
-        and part_local_nids
-
-    7. Now use the `node_map` array to map part_local_nids to their appropriate values to account for
-        node reordering in the `reshuffle_nodes` array.
-
+    1. Form a node_map, in this case a numpy array, which will be used to map old node-ids (pre-reshuffling)
+    to post-reshuffling ids.
+    2. Once the map is created, use this map to map all the node-ids in the part_local_src_id 
+    and part_local_dst_id list to their appropriate `new` node-ids (post-reshuffle order).
+    3. Since only the node's order is changed, we will have to re-order nodes related information when
+    creating dgl object: this includes orig_id, dgl.NTYPE, dgl.NID and inner_node.
+    4. Edge's order is not changed. At this point in the execution path edges are still ordered by their etype-ids.
+    5. Create the dgl object appropriately and return the dgl object.
+    
     Here is a  simple example to understand the above flow better.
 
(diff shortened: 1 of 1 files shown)


tools/distpartitioning/convert_partition.py
```diff
@@ -157,73 +157,121 @@ def create_dgl_object(graph_name, num_parts, \
...
-    # belongs to another partition so that the edge is assigned to another partition.
-    # This happens in a directed graph.
-    # To avoid this kind of nodes being removed from the graph, we add node IDs that
-    # belong to this partition.
+    # get the edge list in some order and then reshuffle.
+    # Here the order of nodes is defined by the `np.unique` function
+    # node order is as listed in the uniq_ids array
     ids = np.concatenate(
-        [shuffle_global_src_id, shuffle_global_dst_id, np.arange(shuffle_global_nid_range[0], shuffle_global_nid_range[1] + 1)])
+        [shuffle_global_src_id, shuffle_global_dst_id,▒
```
zheng-da, line 164: remove the weird character at the end.

tools/distpartitioning/convert_partition.py
```diff
@@ -157,73 +157,121 @@ def create_dgl_object(graph_name, num_parts, \
...
-    return compact_g2, node_map_val, edge_map_val, ntypes_map, etypes_map
+    inner_nodes = th.as_tensor(np.logical_and(
+            uniq_ids >= shuffle_global_nid_range[0],
+            uniq_ids <= shuffle_global_nid_range[1]))
+
+    #get the list of indices, from inner_nodes, which will sort inner_nodes as [True, True, ...., False, False, ...]
+    #essentially local nodes will be placed before non-local nodes.
+    reshuffle_nodes = th.arange(len(uniq_ids))
+    reshuffle_nodes = th.cat([reshuffle_nodes[inner_nodes.bool()],
+                              reshuffle_nodes[inner_nodes == 0]])
```
zheng-da, line 180: `reshuffle_nodes` basically defines the mapping from the new partition-local node IDs to the old partition-local node IDs because the values of `reshuffle_nodes` are the old partition-local node IDs and the index is equivalent to the new partition-local node IDs. To generate the reverse mapping, all you need to do is something like this:
```
old2new = np.zeros((len(reshuffle_nodes,)))
old2new[reshuffle_nodes] = np.arange(len(reshuffle_nodes))
```

kylasa: I agree with these proposed changes. I had a convoluted way of doing this. 
Also, I realized that I am rebuilding the part_local_src_id and part_local_dst_id from scratch which is not needed. We can use these arrays which are already precomputed. Now, this new piece of code looks very clean and straight forward.

tools/distpartitioning/convert_partition.py
```diff
@@ -157,73 +157,108 @@ def create_dgl_object(graph_name, num_parts, \
...
+
+    In this graph above, note that nodes {0, 1, 2, 3} are inner_nodes and {4, 5} are NON-inner-nodes
+
+    Since the edge are re-ordered in any way, there is no reordering required for edge related data
+    during the DGL object creation.
+    '''
+    #create the mappings to generate mapped part_local_src_id and part_local_dst_id
+    #This map will map from unshuffled node-ids to reshuffled-node-ids (which are ordered to prioritize 
+    #locally owned nodes).
+    old2New = np.zeros((len(reshuffle_nodes,)))
```
zheng-da, line 234: don't use capital characters. i think this is not very pythonic.

tools/distpartitioning/convert_partition.py
```diff
@@ -157,73 +157,108 @@ def create_dgl_object(graph_name, num_parts, \
...
+    #get the list of indices, from inner_nodes, which will sort inner_nodes as [True, True, ...., False, False, ...]
+    #essentially local nodes will be placed before non-local nodes.
+    reshuffle_nodes = th.arange(len(uniq_ids))
+    reshuffle_nodes = th.cat([reshuffle_nodes[inner_nodes.bool()],
+                              reshuffle_nodes[inner_nodes == 0]])
+
+    '''
+    Following procedure is used to map the part_local_src_id, part_local_dst_id to account for
+    reshuffling of nodes (to order localy owned nodes prior to non-local nodes in a partition)
+    1. Place side-by-side reshuffle_nodes and corresponding index value.▒
```
zheng-da, line 185: how many of the steps are still valid?
please update it. also do you still need this large comment to document the procedure? it's good to have some comments to explain the code, but the code you have below is pretty simple. you don't need such large comment to explain it.

also, please remove all weird characters at the end of lines.

kylasa: The code comments section is reduced a lot with only 4 steps. I left the example graph as-it-is. I guess it will be helpful to readers who are trying to figure out what is going on in this function.

tools/distpartitioning/convert_partition.py
kylasa: renamed this to 'nid_map'

Issue #3461, opened 2021-10-30
Caching node features in GPU


Jeffery-Song:
## 🚀 Feature

Hi, this is researcher from SJTU. We hope to contribute DGL with the ability to cache node features in GPU when it's impossible to hold all features.

## Motivation

When using GPU for sampling-based mini-batch training, it's likely that a GPU can not hold all node features. In this case, DGL choose to store all features at CPU. Our recent work finds that, part of node features cached in GPU and cache policy properly chosen, the cost of copying feature from GPU to CPU can be significantly reduced, e.g. caching 20% features leads to 80% cache hit rate.

## Alternatives

The simplest method is to cache features with smallest node id. An efficient yet simple method is to cache features of nodes with highest degree, as recent work PaGraph proposed. We step further by running an offline epoch of sample and then caching the mostly visited nodes. 

The degree-based method is easier to contribute, since it only requires knowledge of the graph. Though the sample-based policy achieves greatest performance, it's not easy to design an elegant interface, since it requires knowledge of sample algorithm.

## Pitch

We'd like to know how the community think about the necessity of cache? I've lookup to DGL's PR history and find previous efforts to extend DGL with cache but failed to achieve agreement on the design of semantics #316 . We still hope to take over the job and first contribute the ability of degree-based cache, as it requires less design work but boosts performance significantly. As for the sample-based method, we'd like to borrow the help of community to identify the proper way to design new modules and port our prototype to DGL, if the proposal can be accepted.

jermainewang:
Hi @Jeffery-Song , thanks for the proposal. We will discuss it internally first and back to you asap. cc @VoVAllen

Jeffery-Song:
Looking forward to your reply.

yaox12:
Recently, a distributed GNN framework built upon PyG called [Quiver](https://github.com/quiver-team/torch-quiver) achieves impressive speedup in feature coalescing leveraging hierarchical caching strategy. So I believe this proposal is very helpful, especially in large-scale graphs, which is a common and urgent scenario in the industry.

Details of the caching strategy in Quive can be found [here](https://github.com/quiver-team/torch-quiver/blob/main/docs/Introduction_en.md#3-feature-collection).

jermainewang:
[...]

PR #4341, opened 2022-08-08
[Feature] Gpu cache for node and edge data

mfbalin:
## Description
A gpu_cache that can be used to cache vertex or edge feature storage is implemented. The core implementation comes from the HugeCTR repository and this PR is basically a wrapper around the gpu_cache available in it. The planned use case for it is in the Dataloader to seamlessly wrap the feature storage and speedup access to the node or edge features.

Fixes issue #3461.

Example usage:
`python train_sampling_unsupervised_cache.py --graph-device=gpu --data-device=uva --cache-size=1000000 --dataset=ogbn-products`
This gets 206k samples/sec while the without cache version gets 130k samples/sec. When all of the features are on the GPU without the cache, it gets 350k samples/sec.

## Checklist
- [ ] Code is well-documented


src/runtime/cuda/gpu_cache.cu
```diff
@@ -0,0 +1,166 @@
...
+
+    GpuCache(size_t num_items, size_t num_feats) : num_feats(num_feats),
+        cache((num_items + bucket_size - 1) / bucket_size, num_feats) {}
+
+    std::tuple<NDArray, IdArray, IdArray> Query(IdArray keys) {
+        const auto& ctx = keys->ctx;
+        cudaStream_t stream = dgl::runtime::CUDAThreadEntry::ThreadLocal()->stream;
+        auto device = dgl::runtime::DeviceAPI::Get(ctx);
+        CHECK_EQ(keys->ndim, 1) << "The tensor of requested indices must be of "
+      "dimension one.";
```
yaox12, line 58: Merge it with the above line should be fine.

src/runtime/cuda/gpu_cache.cu
```diff
@@ -0,0 +1,166 @@
...
+        cache.Query(
+            static_cast<const key_t *>(keys->data),
+            keys->shape[0],
+            static_cast<float *>(values->data),
+            static_cast<uint64_t *>(missing_index->data),
+            static_cast<key_t *>(missing_keys->data),
+            missing_len, stream);
+        size_t missing_len_host;
+        device->CopyDataFromTo(missing_len, 0, &missing_len_host, 0,
+                sizeof(missing_len_host),
```
yaox12, line 74: Reduce 4 spaces indentation.

src/runtime/cuda/gpu_cache.cu
```diff
@@ -0,0 +1,166 @@
...
+        device->FreeWorkspace(ctx, missing_len);
+        missing_index = missing_index.CreateView({(int64_t)missing_len_host}, missing_index->dtype);
+        missing_keys = missing_keys.CreateView({(int64_t)missing_len_host}, keys->dtype);
+        return std::make_tuple(values, missing_index, missing_keys);
+    }
+
+    void Replace(IdArray keys, NDArray values) {
+        cudaStream_t stream = dgl::runtime::CUDAThreadEntry::ThreadLocal()->stream;
+        CHECK_EQ(keys->shape[0], values->shape[0]) \
+                << "First dimensions of keys and values must match";
```
yaox12, line 88: Reduce indentation.

src/runtime/cuda/gpu_cache.cu
```diff
@@ -0,0 +1,166 @@
...
+            static_cast<const key_t *>(keys->data),
+            keys->shape[0],
+            static_cast<const float *>(values->data),
+            stream);
+    }
+
+ private:
+    size_t num_feats;
+    gpu_cache::gpu_cache<key_t, uint64_t, std::numeric_limits<key_t>::max(),
+                        set_associativity, WARP_SIZE> cache;
```
yaox12, line 100: Add 1 space indentation.

yaox12 (commented):
@mfbalin You can attach an example demonstrating the usage of GPU cache for reviewing. Performance numbers are also welcome.

src/runtime/cuda/gpu_cache.cu
```diff
@@ -0,0 +1,166 @@
...
+
+template <typename key_t>
+class GpuCache : public runtime::Object {
+constexpr static int set_associativity = 2;
+constexpr static int WARP_SIZE = 32;
+constexpr static int bucket_size = WARP_SIZE * set_associativity;
+
+ public:
+    static constexpr const char* _type_key =
+                    sizeof(key_t) == 4 ? "cuda.GpuCache32" : "cuda.GpuCache64";
```
yaox12, line 47: IIRC, C++/CUDA source code in DGL usually uses the indentation of 2 spaces, and another 4 spaces for line-break in a continuous sentence.

src/runtime/cuda/gpu_cache.cu
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+ private:
+    size_t num_feats;
+    gpu_cache::gpu_cache<key_t, uint64_t, std::numeric_limits<key_t>::max(),
+                        set_associativity, WARP_SIZE> cache;
+};
+
+DGL_DEFINE_OBJECT_REF(GpuCacheRef32, GpuCache<unsigned int>);
+DGL_DEFINE_OBJECT_REF(GpuCacheRef64, GpuCache<long long>); // NOLINT
```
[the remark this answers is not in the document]
mfbalin: Yes, the cu file in HugeCtr gpu cache uses unsigned int and long long. Changing to int64_t results in a mismatch of template arguments.

jermainewang:
Hi, thanks for the contribution. I think having a GPU cache in overall is a good addition to DGL but we need to think though the user experience first before making changes. Here are my major questions/suggestions:
* Is it possible to fold the GPU cache into one of the `FeatureStorage` class? You could check out the existing feature storage classes [here](https://github.com/dmlc/dgl/tree/master/python/dgl/storages) and the base class [here](https://github.com/dmlc/dgl/blob/master/python/dgl/storages/base.py#L55). Perhaps you could create a subclass called `GPUCacheFeatureStorage`.
* How to minimize package dependencies? The PR currently introduces a new third party dependency to HugeCTR, which honestly I don't know much about. Is it possible to limit the dependency to Python side? If HugeCTR provides Python APIs for creating and accessing embedding cache, we could wrap it in a Python class.
* Let's not complicate the `train_unsupervised` script further because its purpose is to educate novice users about unsupervised training. Your setting is more advanced and should be demonstrated with a standalone script.

nv-dlasalle:
@jermainewang HugeCTR doesn't currently expose this via pytorch--however I think this only uses a handful of CPP files from it, so alternatively we could include just the needed files in `third_party` instead of a submodule.

I agree, ideally we should have a way to wrap FeatureStore (or FeatureSource from #4431), so that regardless of where you're pulling features from, you could cache them on the training GPU to reduce traffic.

mfbalin:
I have added GpuCacheFeatureStorage and CachedTensor classes for easier use of the GpuCache. With the addition of CachedTensor, it is very simple to use the GpuCache and the modifications to the existing example are minimal now. However, I can still take those changes out and create a standalone example for the GpuCache.

python/dgl/contrib/gpu_cache.py
```diff
@@ -0,0 +1,113 @@
...
+    def replace(self, keys, values):
+        keys = F.astype(keys, self.idtype)
+        values = F.astype(values, F.float32)
+        _CAPI_DGLGpuCacheReplace(self._cache, F.to_dgl_nd(keys), F.to_dgl_nd(values))
+    
+    @property
+    def hit_rate(self):
+        return 1 - self.total_miss / self.total_queries
+
+class GPUCachedTensorStorage(FeatureStorage):
```
jermainewang, line 51: Let's move it to `python/dgl/storages` and only expose `GPUCachedTensorStorage` for users. Other classes are for internal use.

Please provide an example using `GPUCachedTensorStorage`. See the draft how-to doc here: https://hackmd.io/6PJixJj7Ssy19I-N9CV1Ag

@BarclayII as the doc has not been formally shipped, please chime in if @mfbalin has any questions.

jermainewang:
> @jermainewang HugeCTR doesn't currently expose this via pytorch--however I think this only uses a handful of CPP files from it, so alternatively we could include just the needed files in `third_party` instead of a submodule.

Can you list the files to be included? Alternatively, we could borrow them into the source tree directly if there are not many and the license is compatible. The risk is that future patch in the upstream cannot be easily integrated here, which means we need to  have an owner that knows them.

BarclayII:
We need to think of how to use customized FeatureStorage with `dgl.DGLGraph`, in particular without creating a customized wrapper of the DGLGraph object - this is what we previously do but now I think it's burdensome in retrospect.  Perhaps `get_node_storage` and `get_edge_storage` as methods of the GraphStorage is not a good option.

mfbalin:
> > @jermainewang HugeCTR doesn't currently expose this via pytorch--however I think this only uses a handful of CPP files from it, so alternatively we could include just the needed files in `third_party` instead of a submodule.
> 
> Can you list the files to be included? Alternatively, we could borrow them into the source tree directly if there are not many and the license is compatible. The risk is that future patch in the upstream cannot be easily integrated here, which means we need to have an owner that knows them.

There are only 4 required source and header files needed for the gpu cache, basically the files under this directory: https://github.com/NVIDIA-Merlin/HugeCTR/tree/master/gpu_cache

python/dgl/contrib/gpu_cache.py
mfbalin: I have implemented a custom graph with cached feature storage for a given dictionary of features.

mfbalin:
Can I get a second round of reviews for the recent updates implementing FeatureStorage and a new example using it for GPUCache training?

jermainewang:
> Can I get a second round of reviews for the recent updates implementing FeatureStorage and a new example using it for GPUCache training?

Sorry for the late reply. We are waiting for a review of the [Dataloader changes](https://docs.google.com/document/d/14BlB1gppv7s_RiLRqktQde8uP93qiwWufLRJFuyJGF8/edit#) tomorrow, which will decide how to move forward with this PR.

yaox12:
@dgl-bot

yaox12:
@dgl-bot

yaox12:
Two issues:
1. `gpu_cache` requires SM >= 70 (Volta) while DGL supports SM >= 35 (Kepler).
2. `gpu_cache` in fact requires CUDA >= 11.0, though they declare it supports CUDA >= 9.0, while DGL supports CUDA >= 10.2.

examples/pytorch/graphsage/advanced/train_sampling_unsupervised_cache.py
```diff
@@ -0,0 +1,354 @@
...
+        batch_size=args.batch_size,
+        shuffle=True,
+        drop_last=False,
+        num_workers=args.num_workers,
+        use_uva=args.graph_device == "uva",
+    )
+
+    dataloader.attach_ndata(
+        "features", GPUCachedTensorStorage(nfeat, args.cache_size)
+    )
```
mfbalin, line 140: The current implementation uses the active CUDA stream to fetch features from the GPU cache and also from RAM using UnifiedTensor through UVA access. Is there a way to overlap queries to the GPU cache and to the UnifiedTensor
to hide feature fetch latency? I would assume that only a few SMs can use the whole bandwidth of the PCI-e when fetching from UnifiedTensor so using another stream, it may be possible to hide this latency.

yaox12:
@dgl-bot

examples/pytorch/graphsage/advanced/train_sampling_unsupervised_cache.py
yaox12: Once PR https://github.com/dmlc/dgl/pull/4945 gets merged, the FeatureStorage attached to the dataloader will be fetched in another stream.
https://github.com/dmlc/dgl/blob/e0fc038db111dede9792b3a8158af21dfa4c0748/python/dgl/dataloading/dataloader.py#L333-L340
But for now, it won't overlap with training due to sequential CPU workloads. In an internal project, we make it by using another CPU thread for feature fetching. We will see if we can upstream the modification.

yaox12:
PR https://github.com/dmlc/dgl/pull/4945 is probably to be abandoned in favor of a redesign of the dataloader in the next few months. The review of this PR is also suspended, but I'll keep an eye on it. @mfbalin

mfbalin:
It looks like CUDA 10.2 is deprecated now, so we can depend on the latest version of the GPU cache. Also, is the Graphbolt feature store definition mature enough, maybe we can move this PR forward.

mfbalin:
Tests should pass now, I stripped out the items that don't add the core functionality of the GPU cache so that this PR is smaller. Another PR will add the Graphbolt feature store and another will add an example using the feature store with labor with its kappa feature to show cache miss rates can be reduced.

mfbalin:
I think there is a problem with the CI:
`AttributeError: module 'pydantic_yaml' has no attribute 'YamlModel'`

mfbalin:
We can insert the query call to the GPU cache here for the old dataloader: https://github.com/dmlc/dgl/blob/master/python/dgl/storages/pytorch_tensor.py#L38 in a potential future PR.

frozenbugs:
@yaox12 can you help review the CUDA code?

yaox12:
Overall LGTM. @mfbalin I suggest putting the Python files under `dgl.cuda` directory since we have deprecated all code under `dgl.contrib`.

src/runtime/cuda/gpu_cache.cu
```diff
@@ -0,0 +1,173 @@
...
+        *rv = GpuCacheRef32(
+            std::make_shared<GpuCache<unsigned int>>(num_items, num_feats));
+      else
+        *rv = GpuCacheRef64(std::make_shared<GpuCache<long long>>(  // NOLINT
+            num_items, num_feats));
+    });
+
+DGL_REGISTER_GLOBAL("cuda._CAPI_DGLGpuCacheQuery")
+    .set_body([](DGLArgs args, DGLRetValue *rv) {
+      IdArray keys = args[1];
```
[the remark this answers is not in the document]
mfbalin: Added appropriate checks to ensure the arguments are on the same CUDA device as the GPUCache itself.

mfbalin:
@yaox12 Do you think we should update the HugeCTR version now that DGL doesn't support CUDA 10? I think we can depend on the latest version. But I don't know if there was any change to the gpu_cache code in HugeCTR.

yaox12:
> @yaox12 Do you think we should update the HugeCTR version now that DGL doesn't support CUDA 10? I think we can depend on the latest version. But I don't know if there was any change to the gpu_cache code in HugeCTR.

Of course, I think it's reasonable. Not sure of the changes to gpu_cache either. Maybe you can just update it and check if it works as expected.

mfbalin:
I added only the gpu_cache folder from HugeCTR since recent HugeCTR versions have a lot of dependencies, making `git submodule update --init --recursive` step time-consuming. However, linting needs to be disabled for the `third_party` directory for this to work.

yaox12:
@mfbalin You can add an exclude pattern in https://github.com/dmlc/dgl/blob/7ec78bb6d2ff5c9a54962d0f26befba0716bf71e/.lintrunner.toml#L41. Just like Line 19.

yaox12:
@mfbalin The lint check won't block merge. So just ignore it.

mfbalin:
> @mfbalin The lint check won't block merge. So just ignore it.

Was able to fix it by adding an additional `*` to the exclude_path, like `third_party/**`.

yaox12 (approved):

frozenbugs:
@yaox12 @mfbalin in the future, let's separate the 3rd-party code merging together with the 1st-party code. Since we just import the 3rd party code, we are create a quick PR to make sure it compiles and merge it, and make a separate PR to merge the 1st-party code.

yaox12:
OK, I'll be careful.

caojy1998:
I met this error when I try to build dgl with gcc 11 and g++ 11. Does the change in this PR supports gcc 11 and g++ 11?
@mfbalin  @yaox12 
```
CMake Error at gpu_cache_generated_nv_gpu_cache.cu.o.cmake:276 (message):
  Error generating file
  /dgl/build/CMakeFiles/gpu_cache.dir/third_party/HugeCTR/gpu_cache/src/./gpu_cache_generated_nv_gpu_cache.cu.o


make[2]: *** [CMakeFiles/gpu_cache.dir/build.make:77: CMakeFiles/gpu_cache.dir/third_party/HugeCTR/gpu_cache/src/gpu_cache_generated_nv_gpu_cache.cu.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:182: CMakeFiles/gpu_cache.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

mfbalin:
@caojy1998 seems like g++11 doesn't enable some c++17 features in c++14 mode while g++9 does. Modifying the main CMakeLists.txt file to enable c++17, I was able to compile with g++-12. Is it fine if DGL started using c++17 @frozenbugs @Rhett-Ying? Pytorch switched to requiring c++17 long ago I think.

frozenbugs:
Yes, we should start using c++17.

TristonC:
Can we have a CI for this PR? This failed our container build.

TristonC:
Need PR [6065](https://github.com/dmlc/dgl/pull/6065) for our build

frozenbugs:
@mfbalin consider rollback this PR and the related PR https://github.com/dmlc/dgl/pull/6033 to avoid continuous CI failure for NV container.

Issue #4349, opened 2022-08-09
Replace the original node IDs with shuffled node IDs in the edge list by using tensor operations instead of python dict.


jimjlu99:
## 🔨Work Item

**IMPORTANT:**
* This template is only for dev team to track project progress. For feature request or bug report, please use the corresponding issue templates.
* DO NOT create a new work item if the purpose is to fix an existing issue or feature request. We will directly use the issue in the project tracker.

Project tracker: https://github.com/orgs/dmlc/projects/2

## Description

Replace the original node IDs with shuffled node IDs in the edge list by using tensor operations instead of python dict.

PR #4387, opened 2022-08-11
Distributed Lookup service implementation to retrieve node-level mappings

kylasa:
1. Implemented a class to provide distributed lookup service
2. This class can be used to retrieve global-nids mappings

## Description

#4349 

## Checklist
- [X] Code is well-documented


tools/distpartitioning/data_shuffle.py
```diff
@@ -385,7 +389,7 @@ def read_dataset(rank, world_size, node_part_ids, params, schema_map):
     node_tids, node_features, node_feat_tids, edge_data, edge_tids = \
         get_dataset(params.input_dir, params.graph_name, rank, world_size, schema_map)
 
-    augment_edge_data(edge_data, node_part_ids, edge_tids, rank, world_size)
```
zheng-da, line 388: it seems the original implementation modifies `edge_data` directly.

tools/distpartitioning/data_shuffle.py
```diff
@@ -576,7 +584,7 @@ def gen_dist_partitions(rank, world_size, params):
     print('[Rank: ', rank, '] Done assigning global_ids to edges ...')
 
     #determine global-ids for edge end-points
-    get_shuffle_global_nids_edges(rank, world_size, edge_data, node_part_ids, node_data)
```
zheng-da, line 579: the original implementation change `edge_data`?

tools/distpartitioning/dist_lookup.py
```diff
@@ -0,0 +1,289 @@
...
+import pyarrow
+import torch
+
+from pyarrow import csv
+from gloo_wrapper import alltoallv_cpu
+
+class DistLookupService:
+
+    def __init__(self, input_dir, ntype_names, id_map, rank, world_size):
+        '''
```
zheng-da, line 12: this docstring should go to the class.

tools/distpartitioning/dist_lookup.py
```diff
@@ -0,0 +1,289 @@
...
+        Service constructor used to create an instance of this class.
+
+        Parameters:
+        -----------
+        self : instance of this class
```
zheng-da, line 17: you don't need to document `self`

tools/distpartitioning/dist_lookup.py
```diff
@@ -0,0 +1,289 @@
...
+        ntype_count = []
+
+        #iterate over the node types and extract the partition id mappings
+        for ntype in ntype_names:
+            print('[Rank: ', rank, '] Reading file: ', os.path.join(input_dir, '{}.txt'.format(ntype)))
+            df = csv.read_csv(os.path.join(input_dir, '{}.txt'.format(ntype)), \
+                read_options=pyarrow.csv.ReadOptions(autogenerate_column_names=True), \
+                parse_options=pyarrow.csv.ParseOptions(delimiter=' '))
+            ntype_partids = df['f0'].to_numpy()
+            count = np.int64(len(ntype_partids))
```
zheng-da, line 52: what does this do? you want to convert it to int64? does `int64(len(ntype_partids))` works?

zheng-da: shouldn't `len(ntype_partids)` already return int64?

tools/distpartitioning/dist_lookup.py
```diff
@@ -0,0 +1,289 @@
...
+            ntype_count.append(count)
+
+            #each rank assumes a contiguous set of partition-ids which are equally split
+            #across all the processes.
+            split_size = np.int64(np.ceil(count/np.int64(world_size)))
+            start, end = np.int64(rank * split_size), np.int64((rank+1)*split_size)
```
zheng-da, line 58: are these type casts necessary?

tools/distpartitioning/dist_lookup.py
```diff
@@ -0,0 +1,289 @@
...
+        #Split the input global_nids into a list of lists where each list will be 
+        #sent to the respective rank/process
+        #We also need to store the indices, in the indices_list, so that we can re-order
+        #the final result (partition-ids) in the same order as the global-nids (function argument)
+        send_list = []
+        indices_list = []
+        for idx in range(self.world_size):
+            cond = service_owners == idx
+            idxes = np.where(cond)
+            ll = global_nids[idxes[0]]
```
zheng-da, line 133: why don't you write `global_nids[service_owners == idx]`?

tools/distpartitioning/dist_lookup.py
```diff
@@ -0,0 +1,289 @@
...
+        ntype_ids, type_nids = ntype_ids.numpy(), type_nids.numpy()
+        assert len(ntype_ids) == len(global_nids)
+
+        #For each node-type, the per-type-node-id <-> partition-id mappings are
+        #stored as contiguous chunks by this lookup service. 
+        #The no. of these mappings stored by each process, in the lookup service, are
+        #equally split among all the processes in the lookup service, deterministically
+        service_owners = self.ntype_count[ntype_ids]
+        service_owners = np.ceil(service_owners/self.world_size).astype(np.int64)
+        service_owners = np.floor_divide(type_nids, service_owners).astype(np.int64)
```
zheng-da, line 119: maybe it's better to write them together. the way it writes may not be very intuitive.

tools/distpartitioning/dist_lookup.py
```diff
@@ -0,0 +1,289 @@
...
+                out_list.append(torch.empty((0,), dtype=torch.int64))
+                continue
+            #get the node_type_ids and per_type_nids for the incoming global_nids
+            ntype_ids, type_nids = self.id_map(owner_req_list[idx].numpy())
+            nypte_ids, type_nids = ntype_ids.numpy(), type_nids.numpy()
+
+            #lists to store partition-ids for the incoming global-nids
+            type_id_lookups = []
+            local_order_idx = []
+            for tid in range(len(self.partid_list)):
```
zheng-da, line 160: this loop iterates over all types node types. i think you should write comments like that.
also, the for loop iterates over `self.partid_list`. this is very difficult for people to understand it iterates over node types.

tools/distpartitioning/dist_lookup.py
```diff
@@ -0,0 +1,289 @@
...
+
+                cur_owners = self.partid_list[tid][local_type_nids]
+                type_id_lookups.append(cur_owners)
+
+            if len(type_id_lookups) <= 0:
+                out_list.append(torch.empty((0,), dtype=torch.int64))
+            else:
+                #now reorder results for each request
+                sort_order_idx = np.argsort(np.concatenate(local_order_idx))
+                lookups = np.concatenate(type_id_lookups)[sort_order_idx]
```
zheng-da, line 180: the code is very difficult to understand. is there anyway to simplify the code?

tools/distpartitioning/data_shuffle.py
kylasa: Yes. Originally this function call changed the edge_data dictionary (added SHUFFLE_GLOBAL_SRC_ID and SHUFFLE_GLOBAL_DST_ID keys to it). But somehow this was missed when we made a RULE that if a function changes a variable it should `return` the said variable. 

So now I modified this function call to maintain the consistency.

tools/distpartitioning/dist_lookup.py
kylasa: Done.

tools/distpartitioning/dist_lookup.py
kylasa: Added the docstring to the class appropriately.

tools/distpartitioning/dist_lookup.py
kylasa: Per Python3 documentation len() function returns int64 or int32 depending on the hardware platform. 
https://docs.python.org/3/library/sys.html#sys.maxsize
https://docs.python.org/3/library/functions.html#len

Changed the code accordingly.

tools/distpartitioning/dist_lookup.py
kylasa: Removed some of the type_casts now... where it is NOT necessary.

tools/distpartitioning/dist_lookup.py
kylasa: Agreed. Changed.

tools/distpartitioning/dist_lookup.py
kylasa: I agree. The code is a little bit duplicated between the functions 'get_partition_ids' and 'get_shuffle_nids' -- only difference is the way the mapping is done with each function call. Will look into ways for providing a function pointer as inputs to these functions so that the common part can be put into a function -- sends and receives messages, does the mapping, sends/receives messages. Once this is done, 'get_partition_ids' and 'get_shuffle_ids' can use that common function and passing a function pointer for mapping purposes. 

If you have any other ideas... please share

tools/distpartitioning/dist_lookup.py
kylasa: Added comments to indicate the intention

tools/distpartitioning/dist_lookup.py
zheng-da: here is an example of how python docstring works.
https://github.com/dmlc/dgl/blob/master/python/dgl/graph_index.py#L22

tools/distpartitioning/dist_lookup.py
kylasa: Changed the variable names to appropriately represent the values stored in these variable names. This should remove some of the confusion.

tools/distpartitioning/dist_lookup.py
zheng-da: the docstring of the input argument should be part of the docstring of the class.

tools/distpartitioning/data_shuffle.py
```diff
@@ -520,17 +524,21 @@ def gen_dist_partitions(rank, world_size, params):
...
+                                    schema_map[constants.STR_NODE_TYPE],\
+                                    id_map, rank, world_size)
 
     ntypes_ntypeid_map, ntypes, ntypeid_ntypes_map = get_node_types(schema_map)
     print('[Rank: ', rank, '] Initialized metis partitions and node_types map...')
 
     #read input graph files and augment these datastructures with
     #appropriate information (global_nid and owner process) for node and edge data
-    node_tids, node_features, node_feat_tids, edge_data, edge_features = read_dataset(rank, world_size, node_part_ids, params, schema_map)
+    node_tids, node_features, node_feat_tids, edge_data, edge_features = read_dataset(rank, world_size, id_lookup, params, schema_map)
```
frozenbugs, line 543: Break this super long line? The suggested max length is 79 / 80
https://peps.python.org/pep-0008/#maximum-line-length
https://google.github.io/styleguide/pyguide.html#s3.2-line-length

tools/distpartitioning/dist_lookup.py
```diff
@@ -0,0 +1,309 @@
...
+import torch
+
+from pyarrow import csv
+from gloo_wrapper import alltoallv_cpu
+
+class DistLookupService:
+    '''
+    This is an implementation of a Distributed Lookup Service to provide the following
+    services to its users. Map 1) Global Node-ids to partition-ids, and 2) Global Node-ids
+    to Shuffle Global Node-ids (contiguous, within each node for a give node_type and across
```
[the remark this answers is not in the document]
kylasa: Agreed. Changed the code. Will make sure to use small letters for all the instances of "global_nids" and "shuffle_global_nids" whethere they appear (either in the comments or in the code)

tools/distpartitioning/dist_lookup.py
kylasa: Updated the docstring accordingly.

tools/distpartitioning/dist_lookup.py
```diff
@@ -0,0 +1,309 @@
...
+        assert ntype_names is not None
+        assert len(ntype_names) > 0
+
+        #these lists are indexed by ntype_ids
+        type_nid_begin = []
+        type_nid_end = []
+        partid_list = []
+        ntype_count = []
+
+        #iterate over the node types and extract the partition id mappings
```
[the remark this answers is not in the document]
kylasa: Agreed. Changed the comments accordingly.

tools/distpartitioning/dist_lookup.py
```diff
@@ -0,0 +1,306 @@
...
+        this id_map is used to retrieve ntype-ids, node type ids, and type_nids, per type
+        node ids, for any given global node id
+    rank : integer
+        integer indicating the rank of a given process
+    world_size : integer
+        integer indicating the total no. of processes
+    '''
+
+    def __init__(self, input_dir, ntype_names, id_map, rank, world_size):
+
```
zheng-da, line 46: you shouldn't have an empty line here.

tools/distpartitioning/data_shuffle.py
kylasa: Done. 

Da suggested to run lint and make all the suggested changes by lint in a separate PR -- so that it will be easy to do code review. 

Will run lint and open a separate PR to address these lint issues.

tools/distpartitioning/dist_lookup.py
kylasa: Removed the empty line as suggested.

Issue #4323, opened 2022-08-02
Add option for preserving batch information when removing self loops.


nv-dlasalle:
## 🚀 Feature
It seems unlikely that a users calling `dgl.remove_self_loop()` (https://docs.dgl.ai/generated/dgl.remove_self_loop.html?highlight=remove_self_loop#dgl.remove_self_loop) would desire the side effect of batch information being removed. It can also make code confusing to read:

```
graph = load_batched_graphs()
batch_nodes = graph.batch_num_nodes()
graph = dgl.remove_self_loops(graph)
recalculate_batch_info(graph, batch_nodes)
```

I'm requesting we add a parameter `preserve_batch` to `remove_self_loops`:
```
def remove_self_loop(g, etype=None, preserve_batch=True)
```
Whether it should default to `True` or `False` would depend on what release it makes it in--i.e., are we fine changing the default behavior.

ndickson-nvidia:
Just a note that 2 of the occurrences above are `remove_self_loops`, but the real function is singular for some reason, `remove_self_loop`.  Interestingly, there's also `add_self_loop`, which adds self-loops to all nodes if the graph is homogeneous, so I guess someone wanted self loops at some point.

nv-dlasalle:
This appears to be fixed by https://github.com/dmlc/dgl/pull/3119, but:
1. the docs need to be updated,
2. and we should add a unit test to ensure batch information continues to be preserved when removing self-loops (https://github.com/dmlc/dgl/blob/master/tests/compute/test_transform.py#L1750).

Issue #4316, opened 2022-07-29
Add graph batching support to `dgl.knn_graph()`


amorehead:
## 🚀 Feature
Adding graph batching support to `dgl.knn_graph()`

## Motivation

Being able to provide a batch of coordinates/features to `knn_graph()` would allow users to develop much more efficient graph construction pipelines in their projects.

jermainewang:
Hi, if your scenario is to speedup the data pipeline for preparing batched graph from coordinates/features, then [`GraphDataLoader`](https://docs.dgl.ai/generated/dgl.dataloading.GraphDataLoader.html#dgl.dataloading.GraphDataLoader) is designed for that. The design follows PyTorch's DataLoader. You can define your own `Dataset` with the `__getitem__` function returns one graph constructed from the coordinates. The dataloader will automatically batch the graphs and the pipeline can be further accelerated by multi-processing (using the `num_workers` argument).

amorehead:
Hi, @jermainewang. Thanks for your quick response. This sounds like a good solution for use cases where we are only constructing our input DGLGraphs once, before the forward pass of our model. However, in the use case I have in mind, I actually need to reconstruct my DGLGraph batches multiple times during a single forward pass of my model. That's why I am looking to find a convenient way of constructing `knn_graphs` using a batch of node coordinates (e.g., where node coordinates is a Tensor of size [B, N, 3]). Do you know of any ways to accomplish batched construction of `knn_graphs` in DGL without the use of a GraphDataLoader?

jermainewang:
Is N very large in your case? If not, perhaps you could try to compute a pair-wise distance tensor (of shape [B, N, N]) using tensor operators and then use topk to find the indices of the closest neighbors of each node in the batch. You an then use the indices to construct batched graph.

cc the authors of `knn_graph` @hetong007 @lygztq to see if there are other good ideas.

lygztq:
Hi, if your input is a 3D tensor (e.g. [B, N, 3]), I think you can get a batched graph by `knn_graphs`. You can find an example at the end of the doc [here](https://docs.dgl.ai/generated/dgl.knn_graph.html?highlight=knn_graph#dgl.knn_graph)

If your input is not padded, say you have a set of input ([N1, 3], [N2, 3], ...), you can also use the `dgl.segmented_knn_graph`.

nv-dlasalle:
[...]

PR #3119, opened 2021-07-07
[Bug fix] Fix batch information with remove_nodes/edges applied

mufeili:
## Description
This PR updates batch information in `remove_nodes` and `remove_edges` as reported in #2310 .

## Checklist
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [ ] To the my best knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

PR #4389, opened 2022-08-11
[Feature] Added exclude_self and output_batch to knn graph construction (Issues #4323 #4316)

ndickson-nvidia:
## Description
* Added `exclude_self` and `output_batch` options to `knn_graph` and `segmented_knn_graph`, and their wrappers, `dgl.nn.KNNGraph` and `dgl.nn.SegmentedKNNGraph` for Issue #4316 
* Updated out-of-date comments on `remove_edges` and `remove_self_loop`, since they now preserve batch information as of PR #3119 , for Issue #4323
* Added test cases to `test_remove_selfloop` and `test_knn_cuda`

## Checklist
- [x] Code is well-documented


python/dgl/nn/pytorch/factory.py
```diff
@@ -113,13 +114,21 @@ def forward(self, x, algorithm='bruteforce-blas', dist='euclidean'):
               :math:`\sqrt{\sum_{i} (x_{i} - y_{i})^{2}}`.
             * 'cosine': Use cosine distance.
             (default: 'euclidean')
+        exclude_self : bool, optional
+            If True, the output graph will not contain self loop edges, and each node will not
+            be counted as one of its own k neighbors.  If False, the output graph will contain
+            self loop edges, and a node will be counted as one of its own k neighbors.
+        output_batch : bool, optional
+            If True, the output graph will have batch information set according to the
+            shape of ``x``. If False, the output graph will not have batch information set.
```
BarclayII, line 123: "Batch information set according to the shape of x" sounds too vague.  I would prefer something more concrete, like:
"If True, the output will be a batched graph with xxx graphs.  Each graph within will be a KNN graph constructed from ..." (or something similar).

python/dgl/nn/pytorch/factory.py
```diff
@@ -222,14 +232,22 @@ def forward(self, x, segs, algorithm='bruteforce-blas', dist='euclidean'):
               :math:`\sqrt{\sum_{i} (x_{i} - y_{i})^{2}}`.
             * 'cosine': Use cosine distance.
             (default: 'euclidean')
+        exclude_self : bool, optional
+            If True, the output graph will not contain self loop edges, and each node will not
+            be counted as one of its own k neighbors.  If False, the output graph will contain
+            self loop edges, and a node will be counted as one of its own k neighbors.
+        output_batch : bool, optional
+            If True, the output graph will have batch information set according to ``segs``.
+            If False, the output graph will not have batch information set.
```
BarclayII, line 241: Ditto.

python/dgl/transforms/functional.py
```diff
@@ -164,6 +165,13 @@ def knn_graph(x, k, algorithm='bruteforce-blas', dist='euclidean'):
         * 'euclidean': Use Euclidean distance (L2 norm) :math:`\sqrt{\sum_{i} (x_{i} - y_{i})^{2}}`.
         * 'cosine': Use cosine distance.
         (default: 'euclidean')
+    exclude_self : bool, optional
+        If True, the output graph will not contain self loop edges, and each node will not
+        be counted as one of its own k neighbors.  If False, the output graph will contain
+        self loop edges, and a node will be counted as one of its own k neighbors.
+    output_batch : bool, optional
+        If True, the output graph will have batch information set according to the shape of ``x``.
+        If False, the output graph will not have batch information set.
```
BarclayII, line 174: Ditto.

tests/pytorch/test_geometry.py
```diff
@@ -109,7 +109,9 @@ def check_knn(g, x, start, end, k):
 
 @pytest.mark.parametrize('algorithm', ['bruteforce-blas', 'bruteforce', 'bruteforce-sharemem'])
 @pytest.mark.parametrize('dist', ['euclidean', 'cosine'])
-def test_knn_cuda(algorithm, dist):
+@pytest.mark.parametrize('exclude_self', [False, True])
+@pytest.mark.parametrize('output_batch', [False, True])
+def test_knn_cuda(algorithm, dist, exclude_self, output_batch):
```
BarclayII, line 114: We should also add the same test (`exclude_self` and `output_batch`) for `test_knn_cpu`.  In particular I would like to see whether all the underlying kNN algorithms always return the center nodes themselves.
Also, could we add test for `algorithm == 'nn-descent'` (unless there's some reason why we don't test it, @hetong007 )?

python/dgl/nn/pytorch/factory.py
ndickson-nvidia: Thanks.  I've now updated the comments to be more specific, hopefully without being too verbose, since the multiple KNN graphs part is described elsewhere in the doc comment.  It's described in even more detail in the doc comments in functional.py, so maybe that part of the comment could be brought over here in factory.py, if people would prefer. 
 However, the output graph representing multiple KNN graphs or not is independent of whether the batch info is set or not for those graphs, so people might want that doc change to be separate from this one, or maybe not.

python/dgl/nn/pytorch/factory.py
ndickson-nvidia: Updated.  Further discussion is on the first one.

tests/pytorch/test_geometry.py
ndickson-nvidia: I can add the corresponding tests to `test_knn_cpu`.  I'd assume that the `'nn-descent'` algorithm isn't tested because the tests check the indices, and `'nn-descent'` isn't guaranteed to give exact results, though I've now changed the checking function to make checking indices optional, (for the coincident points test case I just added), so I could just have it skip checking indices if it's the approximate algorithm, and instead just check that each node has the correct number of edges, etc.  Would that make sense?

ndickson-nvidia: I consolidated the `test_knn_cpu` and `test_knn_cuda` implementations, so that they all test the same cases now.

I also tried `'nn-descent'` and I confirmed that it doesn't always get the same indices as the other algorithms.  It also doesn't handle `k` being larger than `F.min(x_segs)` in the same way as the other algorithms, so even without index checking, some of the test cases fail, but I could reorganize the code a little so that it would be consistent in that regard, if it would be useful.  It seems a little out of the scope of the current change, though, so it might be better to change that behaviour in a separate PR, unless you'd like that changed sooner.

python/dgl/transforms/functional.py
```diff
@@ -164,6 +166,14 @@ def knn_graph(x, k, algorithm='bruteforce-blas', dist='euclidean'):
...
         * 'cosine': Use cosine distance.
         (default: 'euclidean')
+    exclude_self : bool, optional
+        If True, the output graph will not contain self loop edges, and each node will not
+        be counted as one of its own k neighbors.  If False, the output graph will contain
+        self loop edges, and a node will be counted as one of its own k neighbors.
+    output_batch : bool, optional
+        If True, the output will be a batched graph representing just 1 KNN graph if :attr:`x`
+        is 2D, or representing :math:`shape(x)[0]` KNN graphs if :attr:`x` is 3D, as described
+        above.  If False, the output graph will be unbatched.
```
rudongyu, line 176: "unbatch" means splitting a given graph into a list of smaller ones. How about "If False, the output graph will be a single graph with multiple components".

python/dgl/transforms/functional.py
```diff
@@ -339,6 +380,13 @@ def segmented_knn_graph(x, k, segs, algorithm='bruteforce-blas', dist='euclidean
         * 'euclidean': Use Euclidean distance (L2 norm) :math:`\sqrt{\sum_{i} (x_{i} - y_{i})^{2}}`.
         * 'cosine': Use cosine distance.
         (default: 'euclidean')
+    exclude_self : bool, optional
+        If True, the output graph will not contain self loop edges, and each node will not
+        be counted as one of its own k neighbors.  If False, the output graph will contain
+        self loop edges, and a node will be counted as one of its own k neighbors.
+    output_batch : bool, optional
+        If True, the output will be a batched graph representing :math:`len(segs)` KNN graphs,
+        as described above.  If False, the output graph will be unbatched.
```
rudongyu, line 389: ditto.

rudongyu (commented):
Looks good to me except the doc for `output_batch`.

python/dgl/transforms/functional.py
ndickson-nvidia: Okay, fair enough.  "single graph with multiple components" also seems confusing, though, since in both cases, the return value is a single graph object with multiple components.  The only difference is whether the batch information is set.  That's why the first version of this description described the difference in terms of batch information being set or not.  I've committed a change that hopefully reconciles the concerns.  It's getting very verbose, though, for such a simple parameter that just controls whether batch information is set or not.

python/dgl/transforms/functional.py
ndickson-nvidia: Updated as described in other thread

jermainewang:
One question for the team. Is it possible to further reduce the number of new arguments? For example, can we omit `output_batch` and always assign batch information to the output when the input `x` is 3D? I'm just pushing everyone to think hard if the current one is a minimal design for user.

tests/pytorch/test_geometry.py
hetong007: Yes, `'nn-descent'` is a randomized algorithm that approximates the results.

> instead just check that each node has the correct number of edges, etc

This sounds good, and probably the only appropriate test for this algorithm.

ndickson-nvidia:
> can we omit `output_batch` and always assign batch information to the output when the input `x` is 3D?

I'd be happy to omit both `output_batch` and `exclude_self` (by having it always exclude self-loop edges), if people would be okay with that.  I added them defaulting to `False` to avoid breaking compatibility.

BarclayII:
After the weekly sync we agreed that the batch information (`output_batch`) can be omitted and we can go straight to set the batch information, since it would not break existing code.  We will keep `exclude_self` though.

ndickson-nvidia:
> the batch information (`output_batch`) can be omitted ... We will keep `exclude_self` though.

Done.  Thanks!

BarclayII (approved):
I'm good with this PR.

python/dgl/transforms/functional.py
```diff
@@ -205,26 +211,58 @@ def knn_graph(x, k, algorithm='bruteforce-blas', dist='euclidean'):
...
             x = F.reshape(x, (x_size[0] * x_size[1], x_size[2]))
-            x_seg = x_size[0] * [x_size[1]]
-        else:
-            x_seg = [F.shape(x)[0]]
         out = knn(k, x, x_seg, algorithm=algorithm, dist=dist)
         row, col = out[1], out[0]
-        return convert.graph((row, col))
+        result = convert.graph((row, col))
+
+    if d == 3:
```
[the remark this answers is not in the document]
ndickson-nvidia: I just went with that because it's what @jermainewang suggested, and because in the case of 2D, it'd be a batch of only 1 graph, so the batch information isn't really needed in that case.  The behaviour of `batch_num_nodes` and `batch_num_edges` would be the same either way, since they'll return a list of length 1 if batch information hasn't been set yet.  If you meant for higher dimensions, `x` was already only supported if 2D or 3D.

Rhett-Ying:
Commit ID: 0b9e9c2d6d20dd6f80bb617faa9d68baa39f8b16

Build ID: 9

Status: ✅ CI test succeeded 

Report path: [link](https://dgl-ci-result.s3.us-west-2.amazonaws.com/dgl/PR-4389/9/9/logs/report.html) 

Full logs path: [link](https://dgl-ci-result.s3.us-west-2.amazonaws.com/dgl/PR-4389/9/9/logs/cireport.log)

PR #4394, opened 2022-08-12
[Dist][CI] Unit test for the new distributed partitioning pipeline

mufeili:
## Description

I've locally successfully tested `test_chunk_graph` and `test_partition`. The test for `test_dispatch` raised an error and I'm not sure if this is due to an inappropriate ip config file.

## Checklist
- [ ] Code is well-documented


tests/tools/test_dist_part.py
```diff
@@ -0,0 +1,124 @@
...
+    parser.add_argument('--partitions-dir', type=str, default='2parts')
+    parser.add_argument('--out-dir', type=str, default='partitioned')
+    parser.add_argument('--ip-config', type=str, default='ip_config.txt')
+    parser.add_argument('--master-port', type=int, default=12345)
+    parser.add_argument('--python-path', type=str, default=sys.executable)
+    args, udf_command = parser.parse_known_args()
+
+    with open(args.ip_config, 'w') as f:
+        f.write('127.0.0.0\n')
+        f.write('127.0.0.1\n')
```
mufeili, line 121: I'm not sure if this is correct.

jermainewang:
> The test for `test_dispatch` raised an error and I'm not sure if this is due to an inappropriate ip config file.

Could you post the error here? And also your test setup.

tests/tools/test_dist_part.py
```diff
@@ -0,0 +1,124 @@
...
+        4,
+        output_dir)
+
+def test_partition():
+    # Step1: graph partition
+
+    with open('chunked-data/metadata.json') as f:
+        metadata = json.load(f)
+    output_path = '2parts'
+    random_partition(metadata, num_parts=2, output_path=output_path)
```
jermainewang, line 105: Please verify the output

tests/tools/test_dist_part.py
```diff
@@ -0,0 +1,124 @@
...
+            }
+        },
+        {
+            'cites': {'count': cite_count_path},
+            'writes': {'year': write_year_path},
+            # you can put the same data file if they indeed share the features.
+            'rev_writes': {'year': write_year_path}
+        },
+        4,
+        output_dir)
```
jermainewang, line 97: Same here. Please verify the output.

jermainewang:
Your way of testing `chunk_graph` is ok, but for partitioning and data dispatching, could you use `os.system` to directly test commandlines?

tests/tools/test_dist_part.py
mufeili: Do you find it sufficient to check the output folder structure?

jermainewang: Yes

mufeili:
> > The test for `test_dispatch` raised an error and I'm not sure if this is due to an inappropriate ip config file.
> 
> Could you post the error here? And also your test setup.

`dispatch_data.py` attempts connecting to a cluster via ssh based on the IP configuration file provided. I did not realize it previously and used `127.0.0.1`, which is incorrect. The question is then how should we test this on CI?

mufeili:
> Your way of testing `chunk_graph` is ok, but for partitioning and data dispatching, could you use `os.system` to directly test commandlines?

Done. I slightly modified `random_partition.py` so that `os.system` can work.

tests/tools/test_dist_part.py
mufeili: Done.

mufeili:
It seems that `load_tensors('.../edge_feat.dgl')` gives an empty dict. Is this an expected behavior? @jermainewang @kylasa

mufeili:
> It seems that `load_tensors('.../edge_feat.dgl')` gives an empty dict at the end of `test_dispatch`. Is this an expected behavior? @jermainewang @kylasa

Confirmed with @jermainewang that this is expected

tests/tools/test_dist_part.py
```diff
@@ -0,0 +1,232 @@
...
+            chunk_f_name = os.path.join(output_edge_sub_dir, chunk_f_name)
+            assert os.path.isfile(chunk_f_name)
+            feat_array = np.load(chunk_f_name)
+            assert feat_array.shape[0] == num_edges[etype] // num_chunks
+
+def test_partition():
+    # Step1: graph partition
+    output_dir = '2parts'
+    os.system('python tools/partition_algorithms/random_partition.py '\
+              '--metadata chunked-data/metadata.json --output_path {} --num_partitions 2'.format(output_dir))
```
jermainewang, line 155: Look like this is not following the command usage in the [doc](https://docs.dgl.ai/en/latest/guide/distributed-preprocessing.html#step-1-graph-partitioning). Could you fix it?

tools/dispatch_data.py
```diff
@@ -33,23 +33,23 @@ def get_launch_cmd(args) -> str:
...
         schema_map = json.load(schema)
 
     num_parts = len(schema_map["num_nodes_per_chunk"][0])
     graph_name = schema_map["graph_name"]
 
     argslist = ""
     argslist += "--world-size {} ".format(num_parts)
-    argslist += "--partitions-dir {} ".format(args.partitions_dir)
-    argslist += "--input-dir {} ".format(args.in_dir)
+    argslist += "--partitions-dir {} ".format(os.path.abspath(args.partitions_dir))
```
jermainewang, line 46: The doc said we "support absolute path or path relative to the location of metadata.json." Could you please confirm? If not, we may want to reword the doc.

tests/tools/test_dist_part.py
```diff
@@ -0,0 +1,232 @@
+import argparse
```
jermainewang, line 1: Can we move these test files to the `tests` folder?

tests/scripts/task_distributed_test.sh
```diff
@@ -30,7 +30,7 @@ export CUDA_VISIBLE_DEVICES=-1
 
 conda activate ${DGLBACKEND}-ci
 
-python3 -m pip install pytest psutil pyyaml pydantic pandas rdflib ogb filelock || fail "pip install"
+python3 -m pip install pytest psutil pyyaml pydantic pandas rdflib ogb filelock pyarrow || fail "pip install"
```
jermainewang, line 33: L41 invokes pytest on `tests/tools/*.py`, which may cause issue if the script contains functions starting with `test_`. Can we put all the test files into `tests` folder?

jermainewang (changes requested):
Changes requested

tests/tools/test_dist_part.py
mufeili: We need to fix the doc instead. I'll open a separate PR for the doc fix.

tests/tools/test_dist_part.py
mufeili: By these, do you mean both `test_dist_part.py` and `test_launch.py`?

tools/dispatch_data.py
mufeili: I checked. Only absolute paths work. I'll update the doc.

tests/tools/test_dist_part.py
```diff
@@ -0,0 +1,232 @@
...
+    for ntype in ['author', 'institution', 'paper']:
+        fname = os.path.join(output_dir, '{}.txt'.format(ntype))
+        with open(fname, 'r') as f:
+            header = f.readline().rstrip()
+            assert isinstance(int(header), int)
+
+def test_dispatch():
+    # Step2: data dispatch
+
+    # generate key for ssh
```
Rhett-Ying, line 165: could be removed after rebase on https://github.com/dmlc/dgl/pull/4432.

tests/tools/test_dist_part.py
Rhett-Ying: which `tests` folder? I think these test files are already under `//tests/`.

tools/dispatch_data.py
mufeili: Fixed in #4433

tests/tools/test_dist_part.py
mufeili: Fixed in #4433

tests/tools/test_dist_part.py
```diff
@@ -0,0 +1,226 @@
...
+        assert tensor_dict.keys() == set(all_tensors)
+        for key in all_tensors:
+            assert isinstance(tensor_dict[key], torch.Tensor)
+
+        # edge_feat.dgl
+        fname = os.path.join(sub_dir, 'edge_feat.dgl')
+        assert os.path.isfile(fname)
+        tensor_dict = load_tensors(fname)
+
+if __name__ == '__main__':
```
Rhett-Ying, line 223: are the generated files/dirs removed/deleted once test finishes?

tests/tools/test_dist_part.py
```diff
@@ -0,0 +1,226 @@
...
+    assert meta_data['graph_name'] == 'mag240m'
+
+    all_ntypes = ['author', 'institution', 'paper']
+    for ntype in all_ntypes:
+        assert len(meta_data['node_map'][ntype]) == num_parts
+    assert meta_data['ntypes'].keys() == set(all_ntypes)
+    assert meta_data['num_edges'] == 4200
+    assert meta_data['num_nodes'] == 720
+    assert meta_data['num_parts'] == num_parts
+    assert meta_data['part_method'] == 'metis'
```
Rhett-Ying, line 191: why it's `metis`, did we partition randomly?

tests/tools/test_dist_part.py
```diff
@@ -0,0 +1,226 @@
...
+        for key in all_tensors:
+            assert isinstance(tensor_dict[key], torch.Tensor)
+
+        # edge_feat.dgl
+        fname = os.path.join(sub_dir, 'edge_feat.dgl')
+        assert os.path.isfile(fname)
+        tensor_dict = load_tensors(fname)
+
+if __name__ == '__main__':
+    test_chunk_graph()
```
Rhett-Ying, line 224: One general question: these tests are required to run sequentially and in strict order due to latter ones depend on outputs from former ones? If so, I think we'd better to put them into a single `test_xxx` function to avoid ambiguity. And according to the code, I think these functions are quite coupled such as `dir`, `magic numbers like num_edges`.

tests/tools/test_dist_part.py
mufeili: Jinjing told me before that all the generated files will be removed after the CI is finished.

Rhett-Ying: this is true as we're running in docker which will be destroyed once finished. But I think it's a better practice to always remove tmp dir/files once specific tests finishes. If we have many tests and they all rely on the delete logic when container destroys, then the total occupied disk may be not trivial. It gets worse when multiple docker are running at the same time.

tests/tools/test_dist_part.py
mufeili: I think it's the default value used [here](https://github.com/dmlc/dgl/blob/dist_part/tools/distpartitioning/convert_partition.py#L249). It probably does not make sense, then I can remove it first.

tests/tools/test_dist_part.py
mufeili: I'm fine with it. What do you think? @jermainewang

tests/tools/test_dist_part.py
mufeili: Ok. I will remove them then.

jermainewang: Perhaps use python https://docs.python.org/3/library/tempfile.html

tests/tools/test_dist_part.py
Rhett-Ying: pls let me add more. Personally, I assume all the tests could be run independently and I usually run a specific test via `python -m pytest tests/distributed/test_rpc.py::test_xxx`. But for tests here, it's not possible except the first one. So putting them into single `test_xxx` function may avoid such confusion.

tests/tools/test_dist_part.py
jermainewang: Pytest also provides default way to request temp file: https://docs.pytest.org/en/7.1.x/getting-started.html#request-a-unique-temporary-directory-for-functional-tests

tests/tools/test_dist_part.py
mufeili: Got it.

jermainewang: You should use pytest fixtures to ensure test orders https://docs.pytest.org/en/7.1.x/how-to/fixtures.html

tests/tools/test_dist_part.py
jermainewang: sorry, I misread.

tests/tools/test_dist_part.py
mufeili: Changed `partition_algorithms/random_partition.py` to `partition_algo/random.py`

tests/tools/test_dist_part.py
mufeili: Removed.

tests/tools/test_dist_part.py
mufeili: @jermainewang `pytest fixtures` work for `python -m pytest tests/tools/test_dist_part.py`. However it prohibits specific tests like `python -m pytest tests/tools/test_dist_part.py::test_chunk_graph`.

tests/tools/test_dist_part.py
jermainewang: Got it. `random_partition` is fine.

tests/tools/test_dist_part.py
jermainewang: Rui's solution is also fine for me.

tests/tools/test_dist_part.py
mufeili: Changed back. I'll update the user guide after we merge this PR.

tests/tools/test_dist_part.py
mufeili: I've used `tempfile.TemporaryDirectory`.

tests/tools/test_dist_part.py
mufeili: I've changed the three test functions to one test function.

jermainewang (approved):

Rhett-Ying (approved):
LGTM

PR #4431, opened 2022-08-18
[DRAFT] Proof-of-concept for 1.0 Dataloader Interface

nv-dlasalle:
## Description
This is meant to illustrate how we could handle more use cases with a modified DataLoader interface. As such, this should not be considered a final or complete implementation, as is it missing several features and optimizations of the current DataLoader.

The major change is that the DataLoader takes in GraphSource and FeatureSource objects, rather than a DGLHeterograph and a BlockSampler.

New example files have been added to show-case the user experience:
* /examples/pytorch/graphsage/new_dataloader_online.py
    This shows doing the normal online sampling, but with the graph structure stored in GPU memory, and the features stored in UVA memory.
* /examples/pytorch/graphsage/new_dataloader_offline.py
    This shows doing offline sampling, where the samples of the graph structure are generated before training, but features are still gathered on the fly via UVA.
* /examples/pytorch/gin/new_dataloader.py
    This example is meant to show how using the same dataloader we can handle graph prediction problems. This is important such that optimization made to node classification and link prediction dataloading also extend to this domain. One thing to note here, is the return batches are a [graph, labels], which is a bit different from the [graph] or [block0, block1, block2] batches from node classification and link prediction. We may be able to make the two more similar by introducing `DGLHeterograph.gdata`, where graph labels/features could be stored.


kkranen:
I just reviewed #4443 as well to compare with this approach.

To compare, the pros of #4431 are:
1. A cleaner UX. #4443's UX takes a much less common design pattern of attaching data sources, whereas the UX here is of dataloaders consuming them (in the init). In my opinion, this UX is far more indicative of what is actually happening (that being a dataloader consuming output from said sources)
2. It enforces that users have defined their feature stores before instantiating dataloaders. Allowing attachment can introduce undefined behavior where new feature sources are introduced after the dataloader has already produced samples.
3. Attachment can have some weird implications for graph sources as well (multiple graph sources?)
4. Unifies the dataloader APIs

The cons of #4431 are:
1. A less modular dataloader design. If we DID want to do really complicated things like change/add feature stores mid-training, we would need to instantiate a new dataloader for it.
2. More implementation effort

Issue #4340, opened 2022-08-08
`CUDA error: unspecified launch failure`, similar to #3802


samvanstroud:
## 🐛 Bug

I am seeing the same issue that was reported as fixed in #3841 in the latest `0.9.0` (and everything down to releases lower than `0.8.0`. See #3802 for more context. As previous reported by @wsjeon, I am seeing this issue using DGL with pytorch lightning, though I haven't tried to see if I can reproduce the problem without using this package.

Tagging @BarclayII and @nv-dlasalle who previously investigated this.

## To Reproduce

1. Setup environment
```
conda config --env --add channels dglteam 
conda config --env --add channels pytorch
conda install dgl-cuda11.3 pytorch-lightning cudatoolkit=11.3 pytorch=1.12.1
```
2. Run this:
```
import torch
import dgl
import pytorch_lightning as pl

class MyModel(pl.LightningModule):
    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(10, 10)

    def training_step(self, batch, batch_nb):
        return torch.tensor(2)

    def configure_optimizers(self):
        return torch.optim.Adam(self.parameters(), lr=0.02)

class MyDataset(torch.utils.data.Dataset):
    def __init__(self):
        super().__init__()
        
    def __len__(self):
        return 10

    def __getitem__(self, idx):
        g = dgl.graph(data=([0,1],[1,0]), num_nodes=2)
        return g, torch.tensor([0])

def collate_graphs(samples):
    graphs = [x[0] for x in samples]
    batched_graph = dgl.batch(graphs)
    targets = torch.cat([x[1] for x in samples])
    return batched_graph, targets

loader = torch.utils.data.DataLoader(dataset=MyDataset(), batch_size=2, num_workers=2, collate_fn=collate_graphs)
model = MyModel()

trainer = pl.Trainer(
    strategy='ddp',
    accelerator='gpu',
    devices=[0],
    fast_dev_run=True,
)

trainer.fit(model, loader)
```

Stack trace:

```
Epoch 0:   0%|                                                                                                                                                       | 0/1 [00:00<?, ?it/s]Traceback (most recent call last):
  File "miniconda3/envs/dgl-test/lib/python3.10/site-packages/pytorch_lightning/trainer/trainer.py", line 721, in _call_and_handle_interrupt
    return self.strategy.launcher.launch(trainer_fn, *args, trainer=self, **kwargs)
[...]

PR #4450, opened 2022-08-23
[Bug] Fix unpinning when tensoradaptor is not available

yaox12:
## Description

Check the `pinned_by_dgl_` flag before unpinning to avoid CUDA calls.

I have tested it offline and it fixes https://github.com/dmlc/dgl/issues/4340. Not sure how to include it in the unit test because the dataloader (where the error occurs) is for the PyTorch backend only, for which we always enable the tensoradaptor.

## Checklist

### DIFF
--- a/src/runtime/ndarray.cc
+++ b/src/runtime/ndarray.cc
@@ -67,7 +67,8 @@ struct NDArray::Internal {
       ptr->mem = nullptr;
     } else if (ptr->dl_tensor.data != nullptr) {
       // if the array is still pinned before freeing, unpin it.
-      UnpinContainer(ptr);
+      if (ptr->pinned_by_dgl_)
+        UnpinContainer(ptr);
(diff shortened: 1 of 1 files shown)


BarclayII (approved):

BarclayII:
We can discuss how to perform unit tests or nightly tests under different CUDA/CMake settings in #4452.

src/runtime/ndarray.cc
```diff
@@ -67,7 +67,8 @@ struct NDArray::Internal {
       ptr->mem = nullptr;
     } else if (ptr->dl_tensor.data != nullptr) {
       // if the array is still pinned before freeing, unpin it.
-      UnpinContainer(ptr);
+      if (ptr->pinned_by_dgl_)
```
Rhett-Ying, line 70: why not do the `if` check within `UnpinContainer()`? One more thing, always wrapping body of `if` statement is a better practice even there's single line...

Issue #4470, opened 2022-08-25
Make TensorAdapter for PyTorch Stream Aware


nv-dlasalle:
## 🚀 Feature
Currently, we use PyTorch's memory allocator but ignore stream information, which may lead improper reallocation of memory segments.

This is one of the issues blocking #4436.

The pytorch TensorAdapter calls `CUDACachingAllocator::raw_alloc()` (https://github.com/pytorch/pytorch/blob/master/c10/cuda/CUDACachingAllocator.h#L135) which uses the current pytorch stream, but ignore the current DGL stream.

We should instead be using `CUDACachingAllocator::raw_alloc_with_stream()` (https://github.com/pytorch/pytorch/blob/master/c10/cuda/CUDACachingAllocator.h#L135), utilizing `thr_entry->stream` to get the current DGL stream.

We also need to call `CUDACachingAllocator::recordstream()` (https://github.com/pytorch/pytorch/blob/master/c10/cuda/CUDACachingAllocator.h#L147) as tensors are used across streams, but it's less clear to me how to accomplish that. It should probably start with our context manager (/python/dgl/_ffi/streams.py).

yaox12:
@nv-dlasalle I think there are three tasks here:
1. Replace `CUDACachingAllocator::raw_alloc()` with `CUDACachingAllocator::raw_alloc_with_stream()`. I have a draft PR for this.
2. Make the allocation from `tensoradapter::TAempty` stream-aware.
3. Port `CUDACachingAllocator::recordstream()` into DGL and I'm working on it.

chang-l:
@yaox12  Thank you for looking into this issue! 

I believe this is the first step to fix https://github.com/dmlc/dgl/issues/4436. I think we also need to provide a similar `set_stream()` method in our context manager, so that users can have a handful tool to temporarily fix this stream issue by themselves.  For example, whenever they set a non-default stream for pytorch, they should be able to manually align DGL stream too, in a similar fashion. Given the amount of work for a complete fix, this walkaround can be meaningful.

@TristonC @jermainewang  for any comments.

yaox12:
For task 2, I planned to set the DGL stream as PyTorch's current stream and then allocated tensors on it. However, allowing external CUDA streams to be set as current is a new feature introduced in PyTorch 1.11. https://github.com/pytorch/pytorch/pull/66324.

I'd like to deprecate `tensoradapter::TAempty` and use `raw_alloc()` only.

chang-l:
[...]

PR #4472, opened 2022-08-26
[Feature] Make TensorAdapter Stream Aware

yaox12:
## Description

- Resolves https://github.com/dmlc/dgl/issues/4470 by
  1. Replace `CUDACachingAllocator::raw_alloc()` with `CUDACachingAllocator::raw_alloc_with_stream()`.
  2. Make the allocation from `tensoradapter::TAempty` stream-aware via `c10::cuda::setCurrentCUDAStream()`.

## Checklist
- [x] Code is well-documented


src/runtime/ndarray.cc
```diff
@@ -214,18 +220,23 @@ NDArray NDArray::EmptyShared(const std::string &name,
...
+      // when td is avaible, use AllocWorkspace to use PyTorch's allocator
+      ret.data_->dl_tensor.data =
+          DeviceAPI::Get(ret->ctx)->AllocWorkspace(
+              ret->ctx, size, ret->dtype);
+    } else {
+      ret.data_->dl_tensor.data =
+          DeviceAPI::Get(ret->ctx)->AllocDataSpace(
+              ret->ctx, size, alignment, ret->dtype);
+    }
+  }
```
yaox12, line 239: Wonder if there's a better way to handle this...

tensoradapter/pytorch/torch.cpp
```diff
@@ -5,57 +5,31 @@
...
-   default:
-    // fallback to CPU
-    return at::Device(torch::kCPU);
-    break;
-  }
-}
-
 extern "C" {
 
-TA_EXPORTS DLManagedTensor* TAempty(
```
chang-l, line 41: Instead of deprecating `TAempty`, can we still use torch current stream for `tensor.empty()` but apply stream sync (with DGL stream) inside this function?

yaox12: Then we have to call `record_stream` to allow the tensor being used in another stream.

Deprecating `TAempty` also helps us get rid of the DLPack conversion.

src/runtime/ndarray.cc
```diff
@@ -69,8 +69,14 @@ struct NDArray::Internal {
...
-      dgl::runtime::DeviceAPI::Get(ptr->dl_tensor.ctx)->FreeDataSpace(
-          ptr->dl_tensor.ctx, ptr->dl_tensor.data);
+      TensorDispatcher* td = TensorDispatcher::Global();
+      if (td->IsAvailable()) {
+        DeviceAPI::Get(ptr->dl_tensor.ctx)->FreeWorkspace(
+            ptr->dl_tensor.ctx, ptr->dl_tensor.data);
+      } else {
+        DeviceAPI::Get(ptr->dl_tensor.ctx)->FreeDataSpace(
+            ptr->dl_tensor.ctx, ptr->dl_tensor.data);
+      }
```
BarclayII, line 79: I don't quite understand what's going on here.  What's the difference between FreeWorkspace and FreeDataSpace, and why we use FreeWorkspace if tensoradapter is available and FreeDataSpace otherwise?

src/runtime/ndarray.cc
```diff
@@ -214,18 +220,23 @@ NDArray NDArray::EmptyShared(const std::string &name,
...
+      ret.data_->dl_tensor.data =
+          DeviceAPI::Get(ret->ctx)->AllocWorkspace(
+              ret->ctx, size, ret->dtype);
+    } else {
+      size_t alignment = GetDataAlignment(ret.data_->dl_tensor);
+      ret.data_->dl_tensor.data =
+          DeviceAPI::Get(ret->ctx)->AllocDataSpace(
+              ret->ctx, size, alignment, ret->dtype);
+    }
+  }
```
BarclayII, line 239: Same question.

yaox12: In the original DGL without a TensorAdaptor, `AllocDataSpace` is used for allocating persistent memories, e.g., for NDArrays, while `AllocWorkspace` is used for allocating temporary memories, e.g., computing buffers, and it uses a thread-local `WorkspacePool`. If we simply replace `AllocDataSpace` with `AllocWorkspace` here, it may raise an error when freeing, saying that `trying to free things that has not been allocated`. I think this is due to the temporary memory being freed in another thread instead of the one it was allocated in. (Please correct me if I'm wrong.)

In this PR, I redirect `AllocWorkspace` to use PyTorch's allocators for both CPU and GPU via the TensorAdaptor. That's why we use `AllocWorkspace` here when TensorAdaptor is available.

An alternative is to redirect `AllocDataSpace` to PyTorch's allocators too when the TensorAdaptor is available. But I think `AllocDataSpace` is more like a “primitive” and wonder if it's reasonable to do so.

Feel free to tell me if there're better ways to unify the usage.

yaox12: Reverted this change and redirect `AllocDataSpace` to PyTorch's allocators.

src/runtime/cuda/cuda_device_api.cc
```diff
@@ -246,21 +255,21 @@ class CUDADeviceAPI final : public DeviceAPI {
...
-      return td->AllocWorkspace(size);
-    else
-      return CUDAThreadEntry::ThreadLocal()->pool.AllocWorkspace(ctx, size);
+      return td->CUDAAllocWorkspace(size, CUDAThreadEntry::ThreadLocal()->stream);
+
+    return CUDAThreadEntry::ThreadLocal()->pool.AllocWorkspace(ctx, size);
   }
 
   void FreeWorkspace(DGLContext ctx, void* data) final {
     TensorDispatcher* td = TensorDispatcher::Global();
```
chang-l, line 268: Do we need `SetDevice(ctx)` here?

chang-l (approved):
Looks good to me now.

include/dgl/runtime/tensordispatch.h
```diff
@@ -68,18 +66,27 @@ class TensorDispatcher {
...
   /*!
-   * \brief Allocate an empty tensor.
-   * Used in NDArray::Empty().
+  * \brief Allocate a piece of CPU memory via
+  * PyTorch's CPUAllocator.
+  * Used in CPUDeviceAPI::AllocWorkspace().
+  *
+  * \param nbytes The size to be allocated.
+  * \return Pointer to the allocated memory.
+  */
```
BarclayII, line 75: These asterisks doesn't seem aligned with the first line.

include/dgl/runtime/tensordispatch.h
```diff
@@ -68,18 +66,27 @@ class TensorDispatcher {
...
-  inline NDArray Empty(std::vector<int64_t> shape, DLDataType dtype, DLContext ctx) const {
-    auto entry = entrypoints_[Op::kEmpty];
-    auto result = FUNCCAST(tensoradapter::TAempty, entry)(shape, dtype, ctx);
-    return NDArray::FromDLPack(result);
+  /*!
+  * \brief Free the CPU memory.
+  * Used in CPUDeviceAPI::FreeWorkspace().
+  *
+  * \param ptr Pointer to the memory to be freed.
+  */
```
BarclayII, line 86: Same.

BarclayII (approved):

Issue #4471, opened 2022-08-26
Unify the cuda stream used in core library


chang-l:
## 🐛 Bug

Although now DGL has a stream control as https://github.com/dmlc/dgl/issues/2974 and https://github.com/dmlc/dgl/pull/3334, it's not very thorough...  For example, cub submodule still always uses the default stream, not respect the setting from DGL stream context manager via `with` statement. 

Here is a quick example:
https://github.com/dmlc/dgl/blob/bd311aa68d07baf4ed71a7b81144bc4ca114f3b0/src/array/cuda/spmat_op_impl_csr.cu#L465-L469

The optional argument `stream` has never got passed ([ref](https://nvlabs.github.io/cub/structcub_1_1_device_radix_sort.html)), and therefore it always uses the default one.

We should either align cub stream to the DGL stream `thr_entry->stream`, or at least sync with  `thr_entry->stream`. This can also be included as a partial fix for https://github.com/dmlc/dgl/issues/4436

Issue #4436, opened 2022-08-18
[Bug] CUDA stream synchronization issue between pytorch and DGL internal functions


chang-l:
## CUDA stream support tracker

- [x] Make TensorAdaptor stream-aware. #4470 #4472
- [x] Clean the improper CUDA stream usage in the core library, including CUB and many hard-coded default stream (0 or nullptr). #4471 #4480
- [x] Add stream-related utilities to DGL, e.g., `.record_stream` for dgl.graph #4467, always use PyTorch's CUDA stream when it's available, etc. #4503

## 🐛 Bug

Update (08/19/2022):

When using a non-default stream for all torch operations (`torch.cuda.set_stream()`), there (DGL) lacks synchronizations in the core library (which always uses the default stream) if there is data-dependency.

-------------------------------

Under certain conditions like (1) using a non-default stream and  (2) with `torch.no_grad()`, a device assertion failure is occurred when calling `SpMM()` function.

After enabling `CUDA_LAUNCH_BLOCKING=1`, the assertion failure is disappeared.

I think this is caused by using different cuda streams for data-transferring and DGL internal functions, while lack of stream synchronizations.  Please see investigation section below.

## To Reproduce
Run the following python script:
```
import torch
import dgl
import torch.nn as nn
from dgl.nn.pytorch import RelGraphConv

class GraphConv(nn.Module):
    def __init__(self, in_channels, out_channels, num_edge_type=3):
        super(GraphConv, self).__init__()
        self.conv = RelGraphConv(in_channels, out_channels, num_edge_type)
        self.in_ch = in_channels
        self.out_ch = out_channels

    def forward(self, dgl_graph, e_type, input):
        res = self.conv(dgl_graph, input, e_type)
        return res

device = 0
# use a non-default stream
s = torch.cuda.Stream()
torch.cuda.set_stream(s)
# input_dim: 10, out_dim: 256,
model = GraphConv(10, 256).to(device)

# torch.no_grad() is necessary to reproduce the issue
with torch.no_grad():
    # graph: 10240 nodes, 409600 edges
    # generate src/dst node ids for 409600 edges
    src = torch.randint(10240, (409600,))
    dst = torch.randint(10240, (409600,))

    # generate dgl graph and find # of nodes
    dgl_graph = dgl.graph(data=(src, dst)).to(device)
    n_node = dgl_graph.num_nodes()

    # edge type id: 0, 1, 2 for all 409600 edges
[...]

PR #4480, opened 2022-08-29
[Feature] Unify the cuda stream used in core library

chang-l:
## Description
To fix https://github.com/dmlc/dgl/issues/4471, I changed all hard-coded default stream to DGL internal stream.

- Enforced `CopyDataFromTo` and its associates to use stream `thr_entry->stream`
- Fix all `CUDA_KERNEL_CALL` to use `thr_entry->stream`
- Fix cub submodule to use `thr_entry->stream`
- Initialized dgl stream to `torch.cuda.current_stream` for each thread, to ensure backward propag. still uses the right stream.

## Checklist
- [x] Code is well-documented

## Tests
- Initial test: 
  With the script from https://github.com/dmlc/dgl/issues/4436, run on a side stream and it will not crash anymore.
- [GCN example](https://github.com/dmlc/dgl/blob/master/examples/pytorch/gcn/train.py)
   Switch to a non-default stream (`torch.cuda.set_stream()`), then align DGL stream using `with dgl._ffi.streams.stream(s)`
   Results are consistent and profiler shows 100% kernels launch on the specified side stream.

- [Graphsage example](https://github.com/dmlc/dgl/blob/master/examples/pytorch/graphsage/node_classification.py)
   Switch to a non-default stream (`torch.cuda.set_stream()`), then align DGL stream using `with dgl._ffi.streams.stream(s)`
   Results are consistent and profiler shows ~70% kernels launch on the specified side stream, other (non-default) streams are all for non-blocking data-transfer in layer-wise inference stage (issued from pytorch). Cross-checked/confirmed with the reference profiling from default stream results.

- [GAT example](https://github.com/dmlc/dgl/blob/master/examples/pytorch/gat/train.py)
   Switch to a non-default stream (`torch.cuda.set_stream()`), then align DGL stream using `with dgl._ffi.streams.stream(s)`
   Results are consistent and profiler shows ~99.5% kernels launch on the specified side stream, the rest (<0.5%) are from `cudaMemcpy` issued on the legacy default stream (auto-sync blocking stream). These `cudaMemcpy` functions are ALL called from the following function:
https://github.com/dmlc/dgl/blob/2dd06b21529d77fa398706444f6db080378e8a79/src/array/cuda/macro.cuh#L16-L17

- [Heter-RGCN example](https://github.com/dmlc/dgl/blob/master/examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py)
   Switch to a non-default stream in pytorch, then align to the same stream for DGL.
   Results are consistent and profiler shows 100% kernels launch on the specified side stream.

- Multi-GPU runs ([graphsage](https://github.com/dmlc/dgl/blob/master/examples/pytorch/multigpu/multi_gpu_node_classification.py), and [rgcn](https://github.com/dmlc/dgl/blob/master/examples/pytorch/rgcn/entity_sample_multi_gpu.py))
   For multi-gpu runs, some un-specified stream launches are from (1) NCCL and (2) pytorch.


include/dgl/runtime/ndarray.h
```diff
@@ -147,39 +147,42 @@ class NDArray {
...
-   */
-  inline void CopyTo(DLTensor *other,
-                     const DGLStreamHandle &stream = nullptr) const;
-  inline void CopyTo(const NDArray &other,
-                     const DGLStreamHandle &stream = nullptr) const;
+   * \param stream The stream to perform the copy context,
+   * \note The copy may happen synchronously if specified stream = dgl internal stream.
+   *       Otherwise, synchronization is necessary.
+   */
+  inline void CopyFromAsync(DLTensor* other,
```
yaox12, line 168: I doubt if we really need `CopyFromAsync`/`CopyToAsync`. Where will they be used?

src/array/cuda/array_index_select.cu
```diff
@@ -80,7 +81,7 @@ DType IndexSelect(NDArray array, int64_t index) {
   device->CopyDataFromTo(
       static_cast<DType*>(array->data) + index, 0, reinterpret_cast<DType*>(&ret), 0,
       sizeof(DType), array->ctx, DLContext{kDLCPU, 0},
-      array->dtype, nullptr);
+      array->dtype, thr_entry->stream);
```
yaox12, line 84: How about omitting stream here?

src/dataloading/async_transferer.cc
```diff
@@ -72,7 +72,7 @@ TransferId AsyncTransferer::StartTransfer(
     // get tensor information
     t.event.reset(new Event);
     CUDA_CALL(cudaEventCreate(&t.event->id));
-    t.dst.CopyFrom(t.src, stream_);
+    t.dst.CopyFromAsync(t.src, stream_);
```
yaox12, line 75: I think this is a legacy file and not used anymore. The corresponding Python code is under `dgl._dataloading`. Maybe we should remove this file in v1.0. cc @nv-dlasalle @mufeili

src/array/cuda/disjoint_union.cu
```diff
@@ -135,17 +135,17 @@ COOMatrix DisjointUnionCoo(const std::vector<COOMatrix>& coos) {
   device->CopyDataFromTo(
       &prefix_elm[coos.size()], 0, &n_elements, 0,
       sizeof(IdType), coos[0].row->ctx, DGLContext{kDLCPU, 0},
-      coos[0].row->dtype, 0);
+      coos[0].row->dtype, stream);
```
yaox12, line 138: And here.

src/array/cuda/disjoint_union.cu
```diff
@@ -84,7 +84,7 @@ void _Merge(IdType** arrs, IdType* prefix, IdType* offset, IdType* out,
 
   device->CopyDataFromTo(
       arrs, 0, arrs_dev, 0, sizeof(IdType*)*n_arrs,
-      DGLContext{kDLCPU, 0}, ctx, dtype, 0);
+      DGLContext{kDLCPU, 0}, ctx, dtype, stream);
```
yaox12, line 87: Same. And probably remove `cudaStream_t stream` from arguments.

src/array/cuda/negative_sampling.cu
```diff
@@ -174,7 +174,7 @@ std::pair<IdArray, IdArray> CSRGlobalUniformNegativeSampling(
   CUDA_CALL(cub::DeviceSelect::If(
         tmp, tmp_size, begin, out_begin, num_out_cuda, num_actual_samples, op,
         thr_entry->stream));
-  num_out = cuda::GetCUDAScalar(device, ctx, num_out_cuda, static_cast<cudaStream_t>(0));
+  num_out = cuda::GetCUDAScalar(device, ctx, num_out_cuda, thr_entry->stream);
```
yaox12, line 177: Consider removing `stream` from `cuda::GetCUDAScalar()`.

include/dgl/runtime/ndarray.h
chang-l: Not really. This is mostly for future I guess. Maybe I can comment/mark it as `Todo`...

src/array/cuda/array_index_select.cu
chang-l: Sure. I will get rid of the extra `thr_entry->stream`

src/array/cuda/disjoint_union.cu
chang-l: Sure.

chang-l:
Once I resolve the above comments, there are still two remaining issues for this PR, according to my profiling:

(1) This PR depends on Xin's PR: https://github.com/dmlc/dgl/pull/4472, to completely unify the dgl internal stream.
(2) In pytorch, backward propag. uses a different thread, which makes the dgl stream `CUDAThreadEntry->ThreadLocal()->stream` out of control within `backward propg` section, i.e., it will always use the default stream for a different thread. 
I am thinking maybe we should avoid declaring cuda stream as a _thread-local_ variable, because cuda stream itself is thread-safe...

Please let me know your thoughts @yaox12 @nv-dlasalle @BarclayII @TristonC

yaox12:
Yesterday I have a brief sync with @jermainewang and he suggests we always get the current stream from PyTorch (via TensorAdaptor maybe) instead of `CUDAThreadEntry->ThreadLocal()->stream`. How do you think about it? @chang-l 

Besides, I think PyTorch also stores the current CUDA stream in a thread-local variable. How does it handle the backward thread?
https://github.com/pytorch/pytorch/blob/6a3ecda5a25025d48bbc5f0215db8c338745ef79/c10/cuda/CUDAStream.cpp#L137-L138

chang-l:
>always get the current stream from PyTorch (via TensorAdaptor maybe) instead of CUDAThreadEntry->ThreadLocal()->stream.

Are we going to always query the `current_stream` from pytorch for **EVERY** kernel calls/mem transfers? I think yes because we have no control on pytorch's stream and users may change torch current stream anywhere in their code, yet the new stream has no way propagate to DGL. This would make code a bit messy and may introduce some overhead. 

However, I agree that packing dgl stream and pytorch stream together might be more convenient and safer, from users perspective. Ideally, for now, users should set their desired stream once (for dgl and its backend), and not worry about or not encourage to 'touch' cuda-stream inside the library. How about integrate `torch.set_stream` in `dgl.set_stream`, so that (1) users only need to set stream once and only through DGL, (2) implying them not invoking `torch.set_stream` when coupled with dgl?

In terms of pytorch's implementation of backward propag., I think they use a global stream pool, defined [here](https://github.com/pytorch/pytorch/blob/ece0002c4beaebaf083dc75b7bf8ceb19edf7a0b/c10/cuda/CUDAStream.cpp#L45-L51). And they sync/record the forward streams at the entry point of autograd(backward) engine through function `InputBuffer::add`. Here are their comment descriptions: https://github.com/pytorch/pytorch/blob/ece0002c4beaebaf083dc75b7bf8ceb19edf7a0b/torch/csrc/autograd/engine.cpp#L164-L168
https://github.com/pytorch/pytorch/blob/6a3ecda5a25025d48bbc5f0215db8c338745ef79/c10/cuda/CUDAStream.h#L13-L53

chang-l:
Maybe I can pin to torch's current stream (sth like stream = torch.current_stream()) here in the constructor of `CUDAThreadEntry`: https://github.com/dmlc/dgl/blob/2b76674002cb1df4de237024838513b6679dfbf2/src/runtime/cuda/cuda_device_api.cc#L300-L302

This way whenever it initiates a new `CUDAThreadEntry` instance at each thread, it can setup the correct stream (through torch of course).

src/dataloading/async_transferer.cc
nv-dlasalle: Confirming that AsyncTransferer should be removed.

src/runtime/cuda/cuda_device_api.cc
```diff
@@ -297,8 +298,13 @@ class CUDADeviceAPI final : public DeviceAPI {
 
 typedef dmlc::ThreadLocalStore<CUDAThreadEntry> CUDAThreadStore;
 
+// TODO(cliu): cuda streams should depend on the current device, therefore we should set device
+//             before setting stream.
```
chang-l, line 302: Could this be error-prone of not setting device here? Now, dgl internal cuda stream is stored as a thread-local variable in `CUDAThreadEntry`. I believe it is not exposed/related to the current device in use (`DGLContext`), right?

Now, I guess it relies on whether the device has been set at beginning by pytorch.

src/runtime/cuda/cuda_device_api.cc
```diff
@@ -297,8 +298,13 @@ class CUDADeviceAPI final : public DeviceAPI {
...
 CUDAThreadEntry::CUDAThreadEntry()
     : pool(kDLGPU, CUDADeviceAPI::Global()) {
+  TensorDispatcher* td = TensorDispatcher::Global();
+  if (td->IsAvailable())
+    stream = td->CUDAGetCurrentStream();
 }
```
chang-l, line 308: Could this be error-prone of not setting device here? Now, dgl internal cuda stream is stored as a thread-local variable in CUDAThreadEntry. I believe it is not exposed/related to the current device id in use (DGLContext), right?

Now, I guess it relies on whether the device has been appropriately set at beginning by pytorch.

chang-l:
I tested with several examples and haven't seen any "stream leakage" (pls check the updated PR description).

src/runtime/cuda/cuda_device_api.cc
```diff
@@ -136,9 +136,10 @@ class CUDADeviceAPI final : public DeviceAPI {
                       size_t size,
                       DGLContext ctx_from,
                       DGLContext ctx_to,
-                      DGLType type_hint,
-                      DGLStreamHandle stream) final {
```
yaox12, line 140: I feel for the device API, we could keep an overload version with the stream as an argument.

One difference from PyTorch is that DGL may have very complicated kernels, e.g., graph sampling, and developers may want to overlap the data transferring and computing via explicit stream control.

src/runtime/cuda/cuda_device_api.cc
```diff
@@ -136,9 +136,10 @@ class CUDADeviceAPI final : public DeviceAPI {
...
+                      DGLType type_hint) final {
+    auto stream = static_cast<DGLStreamHandle>(CUDAThreadEntry::ThreadLocal()->stream);
     cudaStream_t cu_stream = static_cast<cudaStream_t>(stream);
```
yaox12, line 141: Can be unified into one line:
```
cudaStream_t cu_stream = CUDAThreadEntry::ThreadLocal()->stream;
```

src/runtime/cuda/cuda_device_api.cc
yaox12: > the device has been appropriately set at beginning by pytorch.

I think it should be fine. For the single GPU case, PyTorch will use device 0 by default. While for multi-GPU DDP, users are required to set the device for each process.

src/runtime/cuda/cuda_device_api.cc
chang-l: I see. 
I will overload it only in cuda device API then.

src/runtime/cuda/cuda_device_api.cc
chang-l: lol.. Thanks. I was cpy/pasting without much thinking... Resolved as I will be outdated if I overload this function.

chang-l:
Performed tests on hetero-rgcn and multi-gpu cases (updated PR description). No stream leakage found within dgl library calls.

Rhett-Ying:
Commit ID: ef136364d82c0957d75a436e50ebc2d60b28f5e0

Build ID: 12

Status: ✅ CI test succeeded 

Report path: [link](https://dgl-ci-result.s3.us-west-2.amazonaws.com/dgl/PR-4480/12/12/logs/report.html) 

Full logs path: [link](https://dgl-ci-result.s3.us-west-2.amazonaws.com/dgl/PR-4480/12/12/logs/cireport.log)

Rhett-Ying:
Commit ID: e9f0013ff76ac7a9d271bc356c16414a6f26c522

Build ID: 13

Status: ✅ CI test succeeded 

Report path: [link](https://dgl-ci-result.s3.us-west-2.amazonaws.com/dgl/PR-4480/13/13/logs/report.html) 

Full logs path: [link](https://dgl-ci-result.s3.us-west-2.amazonaws.com/dgl/PR-4480/13/13/logs/cireport.log)

yaox12 (approved):
LGTM.

yaox12:
After merging this PR, I realized that to keep up with PyTorch's CUDA stream, users have to call `dgl.cuda.set_stream()` after `torch.cuda.set_stream()` every time, or use `dgl.cuda.stream()` context together with `torch.cuda.stream()`.

To simplify this, I think 
> always query the current_stream from pytorch for EVERY kernel calls/mem transfers? I think yes because we have no control on pytorch's stream and users may change torch current stream anywhere in their code, yet the new stream has no way propagate to DGL.

is acceptable, for example, PyTorch will call `c10::cuda::getCurrentCUDAStream()` everytime when it needs a stream.

To make it, I created a method `runtime::getCurrentCUDAStream()` that would query the current stream from pytorch, as a replacement of `runtime::CUDAThreadEntry::ThreadLocal()->stream` in https://github.com/dmlc/dgl/pull/4503.

chang-l:
>I created a method runtime::getCurrentCUDAStream() that would query the current stream from pytorch, as a replacement of runtime::CUDAThreadEntry::ThreadLocal()->stream in https://github.com/dmlc/dgl/pull/4503.

I see... in this case, I wonder (1) are we going to keep `SetStream`/`CreateStream`/... methods in DGL core library? Or, they are going to in fact set/create pytorch stream instead of cuda stream? (2) if pytorch current stream is changing outside, would that introduces any sync issue to DGL core library? (e.g., when users trying to overlap some operations between dgl and pytorch and they want to call `torch.cuda.set_stream` when dgl library is running? or it should almost never happen?)

PR #4483, opened 2022-08-30
[DataLoader] Allow batch_size=None for GraphDataLoader

### LINKED ISSUE #3358: Support `batch_size = None` for GraphDataLoader
## 🚀 Feature
Supporting `None` batch size for unbatched parallel data loading, as the PyTorch DataLoader supports.

## Motivation
Currently, when passing None as `batch_size` for GraphDataLoader, it will give an error
```
  ...
  File "dgl/dataloading/dataloader.py", line 823, in collate
    elem = items[0]
  File "dgl/heterograph.py", line 2231, in __getitem__
    raise DGLError('Invalid key "{}". Must be one of the edge types.'.format(orig_key))
dgl._ffi.base.DGLError: Invalid key "0". Must be one of the edge types.
```

In some cases, for example as we want to preprocess data in parallel, it is simpler to use None as batch size so we do not need to implement the parallel data pipeline ourselves.

## Alternatives
A batch size of 1 with manual unbatching may be used, but it is less convenient.

BarclayII:
Fixes #3358.


python/dgl/dataloading/dataloader.py
```diff
@@ -1197,17 +1197,18 @@ def __init__(self, dataset, collate_fn=None, use_ddp=False, ddp_seed=0, **kwargs
...
             dataloader_kwargs['sampler'] = self.dist_sampler
 
-        super().__init__(dataset=dataset, collate_fn=self.collate, **dataloader_kwargs)
+        super().__init__(dataset=dataset, collate_fn=collate_fn, **dataloader_kwargs)
+
+        # Overwrite PyTorch's default collator function with ours.
+        # After calling super().__init__() PyTorch should have set the collate_fn attribute.
+        # pylint: disable=access-member-before-definition
+        if self.collate_fn is torch.utils.data._utils.collate.default_collate:
+            self.collate_fn = GraphCollator(**collator_kwargs).collate
```
nv-dlasalle, line 1211: I'm concerned we're relying on implementation details of torch's DataLoader, both in how it chooses to set its `self.collate_fn`, and that it make no other decisions in the constructor based on this. Could we instead explicitly check if batch_size is None, and set our collate_fn input to Torch's DataLoader based on that?
```python3
if batch_size is not None and collate_fn is None:
    collate_fn = GraphCollator(**collator_kwargs).collate
```
Or is there another reason we need to check what collator the parent DataLoader chooses?

BarclayII: Makes sense.

Rhett-Ying:
Commit ID: 741d55332ec87da7f8a39c41f668bd05939c914f

Build ID: 5

Status: ❌ CI test failed in Stage [Lint Check]. 

Report path: [link](https://dgl-ci-result.s3.us-west-2.amazonaws.com/dgl/PR-4483/5/5/logs/report.html) 

Full logs path: [link](https://dgl-ci-result.s3.us-west-2.amazonaws.com/dgl/PR-4483/5/5/logs/cireport.log)

nv-dlasalle (approved):
Overall I think it looks good, but I think it would also be good to document what the expected input/output is if batch_size=None.

Issue #3998, opened 2022-05-12
TensorAdapter silently fails to load in PyTorch, significantly hurting performance


nv-dlasalle:
## 🐛 Bug

As of 0.8, if the `libtensoradapter_pytorch` fails to load, it is not reported to users, and they will see performance drop by up to 2x and in increase in GPU memory (and in some cases get confusing out-of-memory errors from pytorch where it is only using a subset of the memory).

This is a bad user experience, and opens DGL up to inaccurate (and unfavorable) comparisons.

If a user is using DGL on a GPU, they should get some warning that they are running in a degraded state. I know our previous warning lead to some confusion (#3377), but ignoring it for GPU runs defeats many of optimization efforts.

Since there is no warning, users have no reason to run with `DGL_LOG_DEBUG=1` set. Even if the user does set the environment variable, the current message is:
```
>>> import dgl
2022-05-11 16:46:15,425 base.py:142 DEBUG p:MainProcess         t:MainThread: Memory optimization with PyTorch is not enabled.
2022-05-11 16:46:15,426 __init__.py:47 DEBUG p:MainProcess         t:MainThread: Using backend: pytorch
```
Which doesn't tell users the impact of 'Memory optimization' being disabled, nor how they might fix it.

I think it would be good to have the warning on by default including instructions on its impact, how to fix it, and how to disable. We could silence the warning if a user's system doesn't have any cuda capable devices (or at least if they didn't build with CUDA support).

If we consider it normal behavior to run without tensoradapter, I think we need to really need to improve our ndarray allocation and memory pool implementation.

nv-dlasalle:
In the future we may have a workaround with https://github.com/pytorch/pytorch/pull/65365

yzh119:
+1 on this, most users are not aware of this get wrong benchmark results on their papers.

PR #4503, opened 2022-09-01
[Feature] Import PyTorch's CUDA stream management

### LINKED ISSUE #4467: [Feature] Add `.record_stream()` support for dgl.graph
## 🚀 Feature

Add `.record_stream()` support for dgl.graph to enable asynchronous copying between CPU and GPU.

## Motivation

Though we have been able to copy dgl.graph asynchronously on a non-default stream with PR https://github.com/dmlc/dgl/pull/3334, there is a severe issue that prevents us from applying this feature in training.

To reuse a PyTorch tensor on a stream different from the stream of allocation, users have to call `tensor.record_stream()` to prevent its memory from being reused for another tensor, i.e., `record_stream()` lets the allocator know which streams have used the tensor. For example, to copy tensors asynchronously, we have the following code
https://github.com/dmlc/dgl/blob/40a2f3c760e238b4fc8cb4d069d692423cbb302f/python/dgl/dataloading/dataloader.py#L335

Since dgl.graph doesn't support `.record_stream()` for now, if we perform asynchronous copy for dgl.graph, it may cause an error, as in #4247.

## Alternatives

No alternatives.

## Additional context

This is related to the behavior of PyTorch's caching allocator. 
If we are going to import PyTorch's `.record_stream()`, I doubt whether we are porting too many things from PyTorch via TensorAdaptor.

yaox12:
## Description

Resolve #4467.

Add stream-related utilities:
1. Add `RecordStream()` for NDArray and HeteroGraph on the C++ side,
2. Add `record_stream()` for dgl.heterograph on the Python side.
3. Replace all `runtime::CUDAThreadEntry::ThreadLocal()->stream` with `runtime::getCurrentCUDAStream()` to ensuring we're using PyTorch's current stream. With this change, we don't need `dgl.set_stream()` method or `dgl.cuda.stream()` context, and just use torch's stream management instead. This will simplify the user experience.
4. Add unit tests.

## Checklist
- [x] Code is well-documented


python/dgl/heterograph.py
```diff
@@ -5584,6 +5584,21 @@ def is_pinned(self):
...
+        ----------
+        stream : torch.cuda.Stream
+            The stream that is using this graph.
+
+        Returns
+        -------
+        DGLGraph
+            self.
+        """
+        return self._graph.record_stream(stream)
```
nv-dlasalle, line 5600: @BarclayII semantically should this be recording the stream for all node and edge frames too?

src/graph/unit_graph.cc
```diff
@@ -1316,6 +1326,15 @@ void UnitGraph::UnpinMemory_() {
...
 }
 
+void UnitGraph::RecordStream(DGLStreamHandle stream) const {
+  if (this->in_csr_->defined())
+    this->in_csr_->RecordStream(stream);
+  if (this->out_csr_->defined())
+    this->out_csr_->RecordStream(stream);
+  if (this->coo_->defined())
+    this->coo_->RecordStream(stream);
+}
```
nv-dlasalle, line 1336: Do we need anything extra to handle the case where new formats are being created when one format is recorded on a non-default stream?

tests/pytorch/test_stream.py
```diff
@@ -1,10 +1,39 @@
...
+import dgl.ndarray as nd
 from dgl import rand_graph
-import dgl._ffi.streams as FS
 import dgl.ops as OPS
-import unittest
+from dgl.cuda import to_dgl_stream_handle
 import backend as F
-import torch
 
+# borrowed from PyTorch
```
nv-dlasalle, line 12: Can you cite the specific file?

src/graph/unit_graph.cc
yaox12: Can we assume `record_stream` is only used in data loading?

yaox12:
Paste @chang-l 's comments in https://github.com/dmlc/dgl/pull/4480#issuecomment-1238438111.

>I (Xin) created a method `runtime::getCurrentCUDAStream()` that would query the current stream from pytorch, as a replacement of `runtime::CUDAThreadEntry::ThreadLocal()->stream`.

(Chang): I see... in this case, I wonder (1) are we going to keep `SetStream`/`CreateStream`/... methods in DGL core library? Or, they are going to in fact set/create pytorch stream instead of cuda stream? (2) if pytorch current stream is changing outside, would that introduces any sync issue to DGL core library? (e.g., when users trying to overlap some operations between dgl and pytorch and they want to call `torch.cuda.set_stream` when dgl library is running? or it should almost never happen?)

chang-l:
@yaox12 Thanks for pulling my comment here :)
In general, I agree on using the current stream from pytorch to launch every dgl kernels. There are two things I suggest to fix in the meanwhile:
(1) Update `SetStream`/`CreateStream` methods as well, to set/create from pytorch stream pool.
(2) Update doc to remind users that all dgl functions use the same current stream as backend (pytorch), so if they changed to another stream, they should always restore the stream before calling next dgl function, for example better using `with` instead of `set_stream`.

jermainewang:
Reassigned the reviewers. For CUDA related PRs, please reach out to @isratnisa  for review. Changes to TensorAdapter needs @BarclayII 's approval.

tests/pytorch/test_stream.py
```diff
@@ -34,6 +62,64 @@ def test_basics():
...
+    result = nd.empty([4], ctx=nd.gpu(0))
+    stream = torch.cuda.Stream()
+    ptr = [None]
+
+    # Performs the CPU->GPU copy in a background stream
+    def perform_copy():
+        with torch.cuda.stream(stream):
+            tmp = t.copyto(nd.gpu(0))
+            ptr[0] = F.from_dgl_nd(tmp).data_ptr()
+        torch.cuda.current_stream().wait_stream(stream)
```
isratnisa, line 102: What's the difference between `torch.cuda.current_stream()` and `stream` which is `torch.cuda.Stream()` here? Can they be same?

tests/pytorch/test_stream.py
```diff
@@ -34,6 +62,64 @@ def test_basics():
...
+        tmp.record_stream(
+            to_dgl_stream_handle(torch.cuda.current_stream()))
+        torch.cuda._sleep(int(50 * cycles_per_ms))  # delay the copy
+        result.copyfrom(tmp)
+
+    perform_copy()
+    with torch.cuda.stream(stream):
+        tmp2 = nd.array(np.array([1., 2., 3., 4.], dtype=np.float32), ctx=nd.gpu(0))
+        assert F.from_dgl_nd(tmp2).data_ptr() != ptr[0], 'allocation re-used to soon'
```
isratnisa, line 111: typo `too soon`

isratnisa: I do not understand the logic here. `ptr[0]` holds the location of `tmp`. Shouldn't it be always different than `tmp2`?

python/dgl/_ffi/streams.py
```diff
@@ -1,17 +1,17 @@
 # pylint: disable=invalid-name, unused-import
-"""Runtime stream api which is maily for internal use only."""
+"""Runtime stream api which is maily for internal use only.
```
isratnisa, line 2: Typo. `... mainly for ...`

isratnisa (changes requested):

python/dgl/_ffi/streams.py
```diff
@@ -63,3 +58,43 @@ def stream(cuda_stream):
...
+def _dgl_set_stream(ctx, cuda_stream):
+    """ Set the current CUDA stream of the given DGL context.
+    Use of this function is discouraged in favor of torch's stream context.
+
+    Parameters
+    ----------
+    ctx: DGLContext
+    stream: DGLStreamHandle
+    """
+    check_call(_LIB.DGLSetStream(ctx.device_type, ctx.device_id, cuda_stream))
```
chang-l, line 84: DGLSetStream needs to be fixed, as now it still fetches `CUDAThreadEntry::ThreadLocal()->stream`

tests/pytorch/test_stream.py
```diff
@@ -34,6 +62,64 @@ def test_basics():
...
+    with torch.cuda.stream(stream):
+        t = nd.array(np.array([1., 2., 3., 4.]), ctx=nd.gpu(0))
+        g = rand_graph(10, 20, device=F.cuda())
+    # NDArray is an internal object, just use DGLStreamHandle
+    t.record_stream(to_dgl_stream_handle(torch.cuda.current_stream()))
+    g.record_stream(torch.cuda.current_stream())
+
+@unittest.skipIf(F._default_context_str == 'cpu', reason="stream only runs on GPU.")
+# borrowed from PyTorch, test/test_cuda.py: test_record_stream()
+def test_record_stream():
```
chang-l, line 88: do we need to test `record_stream` for dgl graph object other than smoke testing?

tests/pytorch/test_stream.py
```diff
@@ -24,7 +53,6 @@ def test_basics():
     s = torch.cuda.Stream(device=F.ctx())
     with torch.cuda.stream(s):
         xx = x.to(device=F.ctx(), non_blocking=True)
-    with FS.stream(s):
         gg = g.to(device=F.ctx())
     s.synchronize()
     OPS.copy_u_sum(gg, xx)
```
chang-l, line 58: I think we need correctness check here as well, sth like: `assert (OPS.copy_u_sum(gg, xx) == OPS.copy_u_sum(g, x))`

tests/pytorch/test_stream.py
yaox12: `stream = torch.cuda.Stream()` will create a new stream. So code in
```
with torch.cuda.stream(stream):
    ...
```
will run in a different stream while code outside this with context still runs in the `torch.cuda.current_stream()`.

tests/pytorch/test_stream.py
yaox12: Imagining we didn't call `.record_stream()` for `tmp`, line 105-106 have launched kernels for sleeping and copying. Then in line 110, from the host side, the lifetime of `tmp` is over, so the caching allocator will reuse the memory block of `tmp` for `tmp2`.

`record_stream` will tell the caching allocator that `torch.cuda.current_stream()` is using `tmp`, please don't free its memory before events in `torch.cuda.current_stream()` have done. So in line 110, the allocator will allocate a new memory block for `tmp2`, and that's what we expect.

python/dgl/_ffi/streams.py
yaox12: I'm in favor of removing `SetStream`, `GetStream` APIs and DGL's `stream()` context, at least at the Python level, because they have been replaced by PyTorch's APIs.

tests/pytorch/test_stream.py
yaox12: A blocker is that we don't have an equivalent API for dgl.graph as `tensor.data_ptr()` to check if the memory has been reused or not. The underlying sparse matrix or NDArrays of dgl.graph have not been exposed to Python and we have not way to access them.

tests/pytorch/test_stream.py
yaox12: Updated accordingly.

python/dgl/_ffi/streams.py
```diff
@@ -63,3 +58,43 @@ def stream(cuda_stream):
...
+def _dgl_get_stream(ctx):
+    """Get the current CUDA stream of the given DGL context.
+
+    Parameters
+    ----------
+    ctx : DGL context.
+
+    Returns
+    -------
+    DGLStreamHandle.
```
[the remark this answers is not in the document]
yaox12: Updated accordingly.

python/dgl/_ffi/streams.py
```diff
@@ -63,3 +58,43 @@ def stream(cuda_stream):
...
+    return ctypes.c_void_p(cuda_stream.cuda_stream)
+
+def _dgl_set_stream(ctx, cuda_stream):
+    """ Set the current CUDA stream of the given DGL context.
+    Use of this function is discouraged in favor of torch's stream context.
+
+    Parameters
+    ----------
+    ctx: DGLContext
+    stream: DGLStreamHandle
```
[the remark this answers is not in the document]
yaox12: Removed `SetStream()`.

tests/pytorch/test_stream.py
chang-l: I see... Can you run a different test other than directly checking the allocator? For example, checking spmm results like
```
# default stream
g = rand_dgl_graph.to(device) # g allocates on gpu with default stream
copy_stream = torch.cuda.Stream()
with copy_stream:
       g2 = g.to(device)  # copy g on copy_stream
       sleep(100000)   # make copy takes very long time
g.record_stream(copy_stream)  # necessary to ensure copy is finished
g = dgl_unit_graph # access and modify g on default stream again 
res = copy_u_sum(g, x)  # check if result is correct
```

python/dgl/_ffi/streams.py
chang-l: I am Okay with removing `SetStream`, `GetStream` in Python level. 
In fact, now I think I am also fine with removing  `SetStream`/`CreateStream` in C++ level, if only single stream (current pytorch stream) is used for now in DGL. We can add it back when necessary in future.

tests/pytorch/test_stream.py
yaox12: Good idea. I have added both positive and negative tests.

python/dgl/heterograph.py
BarclayII: I would prefer recording them separately unless there is a reason to record all the streams together (e.g. some DGL's APIs require recording the graph and the features at the same time).

python/dgl/_ffi/streams.py
BarclayII: > My main concern is whether we would have imported too many helper functions from the TensorAdaptor if we redirected `SetStream`/`GetStream` to PyTorch.

This is OK.  TensorAdapter is just an interface to call PyTorch C++ functions in DGL, so you can have as many functions as you want.

python/dgl/heterograph.py
yaox12: Updated accordingly.

chang-l (approved):
Looks good to me now. Some minor concerns about testing: (1) have you performed some end-to-end tests, like our internal CI (P0/P1/P2)? and (2) can you please try adding `torch.cuda.set_stream(s)` in some examples, to ensure consistent results?

tests/pytorch/test_stream.py
isratnisa: I see. Thanks.

isratnisa (approved):

yaox12:
> (2) can you please try adding torch.cuda.set_stream(s) in some examples, to ensure consistent results?

@chang-l Which examples?

chang-l:
I meant some existing examples in `example/pytorch` folder like gcn/graphsage/... For example, adding `torch.cuda.synchronize(); torch.cuda.set_stream(s)` to switch to another non-default stream at the beginning, and the results should not change significantly, as a higher-level (end2end) sanity check.

python/dgl/frame.py
```diff
@@ -492,6 +495,17 @@ def unpin_memory_(self):
...
+    def record_stream(self, stream):
+        """Record stream that is using the storage.
+        Does nothing if the backend is not PyTorch.
+
+        Parameters
+        ----------
+        stream : torch.cuda.Stream.
+        """
+        if F.get_preferred_backend() == 'pytorch':
+            self.data.record_stream(stream)
```
nv-dlasalle, line 507: Should we throw an error if the backend is not pytorch?

src/graph/unit_graph.cc
nv-dlasalle: I could envision it being used in forward/backward where we have parallel layers (.e.g, on heterogeneous graphs). If its not allowed outside of dataloading, we would need to add some check that prevent users from doing it, or at least warn them.

nv-dlasalle (changes requested):
I think we also need to delete the `CUDAThreadEntry::stream` member, to ensure it is not used after this change.

How do we handle the fact that PyTorch's streams are not created with the [cudaStreamNonBlocking](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g5dbd11a1723d9f5938a133cedbc525e3) flag, so to get real parallelism, users will need to do their training on the non-default stream, and dataloading on other streams as well.

yaox12:
> I think we also need to delete the `CUDAThreadEntry::stream` member, to ensure it is not used after this change.

What if TensorAdaptor is not available? Should `runtime::getCurrentCUDAStream()` just return the default stream?

> How do we handle the fact that PyTorch's streams are not created with the [cudaStreamNonBlocking](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g5dbd11a1723d9f5938a133cedbc525e3) flag, so to get real parallelism, users will need to do their training on the non-default stream, and dataloading on other streams as well.

PyTorch's streams are created in 
https://github.com/pytorch/pytorch/blob/d00cabae7bc894b951d4b2c9c24c7d95bebd86e1/c10/cuda/CUDAStream.cpp#L165-L168
```c++
    C10_CUDA_CHECK(cudaStreamCreateWithPriority(
        &lowpri_stream, kDefaultFlags, kLowPriority));
    C10_CUDA_CHECK(cudaStreamCreateWithPriority(
        &hipri_stream, kDefaultFlags, kHighPriority));
```
with the flag
https://github.com/pytorch/pytorch/blob/d00cabae7bc894b951d4b2c9c24c7d95bebd86e1/c10/cuda/CUDAStream.cpp#L25
```c++
static constexpr unsigned int kDefaultFlags = cudaStreamNonBlocking;
```

nv-dlasalle:
> > I think we also need to delete the `CUDAThreadEntry::stream` member, to ensure it is not used after this change.
> 
> What if TensorAdaptor is not available? Should `runtime::getCurrentCUDAStream()` just return the default stream?
> 

If tensoradapter is not available, I don't think we can safely pass streams between DGL and a backend, so I think returning the default stream is the correct behavior. I think this also makes #3998 more of a priority, since if tensoradapter fails to load we lose more than just faster memory management.


> > How do we handle the fact that PyTorch's streams are not created with the [cudaStreamNonBlocking](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g5dbd11a1723d9f5938a133cedbc525e3) flag, so to get real parallelism, users will need to do their training on the non-default stream, and dataloading on other streams as well.
> 
> PyTorch's streams are created in https://github.com/pytorch/pytorch/blob/d00cabae7bc894b951d4b2c9c24c7d95bebd86e1/c10/cuda/CUDAStream.cpp#L165-L168
> 
> ```c++
>     C10_CUDA_CHECK(cudaStreamCreateWithPriority(
>         &lowpri_stream, kDefaultFlags, kLowPriority));
>     C10_CUDA_CHECK(cudaStreamCreateWithPriority(
>         &hipri_stream, kDefaultFlags, kHighPriority));
> ```
> 
> with the flag https://github.com/pytorch/pytorch/blob/d00cabae7bc894b951d4b2c9c24c7d95bebd86e1/c10/cuda/CUDAStream.cpp#L25
> 
> ```c++
> static constexpr unsigned int kDefaultFlags = cudaStreamNonBlocking;
> ```

Sorry, you're right, I read the pytorch source wrong.

src/graph/unit_graph.cc
yaox12: Calling `RecordStream` on newly created formats now.

python/dgl/frame.py
yaox12: Updated.

nv-dlasalle (approved):

Issue #4410, opened 2022-08-16
[Example][Refactor] Refactor graphsage and rgcn example for link prediction


chang-l:
The graphsage example for link prediction needs refactor ([graphsage/link_pred.py](https://github.com/dmlc/dgl/blob/master/examples/pytorch/graphsage/link_pred.py)).

The rgcn example for link prediction needs refactor ([rgcn/link.py](https://github.com/dmlc/dgl/blob/master/examples/pytorch/rgcn/link.py)).

PR #4186, opened 2022-06-28
[Example][Refactor] Regolden graphsage example for future guide

chang-l:
## Description
Simplify from https://github.com/dmlc/dgl/pull/4159
Focus on single-GPU, mini-batch node classification task (graphsage/node_classification.py) to ease the iterating process.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage
- [x] Code is well-documented
- [x] To the best of my knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

## Changes
- Arguments/keyword arguments naming consistent with doc
- Data logging and output clean-up
- Remove all cmd arguments, only keeping --gpu for simplicity (proper argument naming + doc should be clear enough)
- Add comments for clarity
- Bugfix: fanout correction from [15, 10, 5] --> [5, 10, 15]
- README update

## Tests
- `python3 node_classification.py --gpu 0`
```
----Data statistics------'
    #Edges 123718280
    #Classes 47
    #Train samples 196615
    #Val samples 39323
    #Test samples 2213091
Epoch 00000 | Loss 1.3636 | Accuracy 0.8556 | GPU Mem(MB) 871.24
...
Epoch 00009 | Loss 0.3997 | Accuracy 0.8968 | GPU Mem(MB) 871.50

Test Accuracy 0.7658
```

PR #4526, opened 2022-09-08
[Example][Refactor] Graphsage link prediction example refactor

chang-l:
## Description
To partly resolve https://github.com/dmlc/dgl/issues/4410, this PR refactors the example in a similar style as https://github.com/dmlc/dgl/pull/4186.
## Checklist
- [ ] Code is well-documented

## Changes
- Simplify output format
- Refactor for similar code style

## Results
| Dataset       | Test acc. (before) | Mean |Test acc. with 50epochs (after) | Mean
| ------------- | -------| -------| -------| -------
| ogbn-citation2| [0.8131 0.7619 0.8422] | 0.8057| [0.7816  0.7789 0.7790] | 0.7798


mufeili:
I wonder if we should move the example to https://github.com/dmlc/dgl/tree/master/examples/pytorch/ogb

examples/pytorch/graphsage/README.md
```diff
@@ -59,3 +59,9 @@ Train w/ mini-batch sampling for link prediction on OGB-Citation2:
 ```bash
 python3 link_pred.py
 ```
+
+Results:
+```
+Test Accuracy: 0.8600
```
mufeili, line 65: I think the metric is `MRR` rather than `accuracy`. Also this seems to be too high if we compare with the results in https://ogb.stanford.edu/docs/leader_linkprop/#ogbl-citation2

mufeili: I'm also curious why we implement a metric function ourselves rather than use the `evaluator` provided by OGB.

chang-l: > I think the metric is MRR rather than accuracy. 

Thanks. I fixed it.

> Also this seems to be too high

Good point. Given only 10 epochs, there is sth wrong I think...

> I'm also curious why we implement a metric function ourselves rather than use the evaluator provided by OGB.

I am not sure why not using ogb evaluator... I definitely agree we should use ogb evaluator for all ogb examples. Not sure if there is any compatibility issue though (like if some users have out-dated pkg, would that be a problem?)

chang-l:
I switched to ogb evaluator and the resulting test MRR drops to ~0.6687 (with 10 epochs). After investigating this discrepancy, I think the original implementation is wrong, and the MRR returned from ogb evaluator is more trustworthy. Here is the my modified `compute_mrr` function for validation:

```
def compute_mrr(model, evaluator, node_emb, src, dst, neg_dst, device, batch_size=500):
    rr = torch.zeros(src.shape[0])
    for start in tqdm.trange(0, src.shape[0], batch_size, desc='Evaluate'):
        end = min(start + batch_size, src.shape[0])
        all_dst = torch.cat([dst[start:end, None], neg_dst[start:end]], 1)
        h_src = node_emb[src[start:end]][:, None, :].to(device)
        h_dst = node_emb[all_dst.view(-1)].view(*all_dst.shape, -1).to(device)
        pred = model.predictor(h_src*h_dst).squeeze(-1)
        # original impl.
        relevance = torch.zeros(*pred.shape, dtype=torch.bool).to(pred.device)
        relevance[:, 0] = True
        rr[start:end] = MF.retrieval_reciprocal_rank(pred, relevance)
        first = MF.retrieval_reciprocal_rank(pred[0,:], relevance[0,:]) # first row (edge) recip. rank
        second = MF.retrieval_reciprocal_rank(pred[1,:], relevance[1,:]) # second row (edge) recip. rank
        third = MF.retrieval_reciprocal_rank(pred[2,:], relevance[2,:]) # third row (edge) recip. rank
        print('actual rr (edg 1-3): ', first, second, third)
        print('orig. impl. rr (edg 1-3)', rr[start:start+3])

        # ogb evaluator impl.
        input_dict = {'y_pred_pos': pred[:,0], 'y_pred_neg': pred[:,1:]}
        rr[start:end] = evaluator.eval(input_dict)['mrr_list']
        print('ogb eval rr (edg 1-3): ', rr[start:start+3])
        exit(0)
    return rr.mean()
```
Please feel free to test/comment.

examples/pytorch/graphsage/README.md
mufeili: @BarclayII Could you explain the reasons for the custom evaluation function?

mufeili:
> I wonder if we should move the example to https://github.com/dmlc/dgl/tree/master/examples/pytorch/ogb

@jermainewang What do you think?

jermainewang:
Why do you think it should belong to OGB?

mufeili:
> Why do you think it should belong to OGB?

Because it is on `ogbl-citation2`.

jermainewang:
> > Why do you think it should belong to OGB?
> 
> Because it is on `ogbl-citation2`.

The node prediction example is based on ogbn-product too. We should think of the purpose of these examples and put them in the corresponding folder:
* T0/T1 examples teach new users how to accomplish certain tasks or implement certain models. Readability is thus the first priority.
* OGB examples are baselines for researchers to reproduce leaderboard results. Reproducibility is thus the first consideration.

examples/pytorch/graphsage/link_pred.py
```diff
@@ -97,78 +80,91 @@ def inference(self, g, device, batch_size, num_workers, buffer_device=None):
...
-                print(tt - t0)
-                durations.append(tt - t0)
-                break
-    if epoch % 10 == 0:
-        model.eval()
-        valid_mrr, test_mrr = evaluate(model, edge_split, device, 0 if args.pure_gpu else 12)
-        print('Validation MRR:', valid_mrr.item(), 'Test MRR:', test_mrr.item())
-print(np.mean(durations[4:]), np.std(durations[4:]))
+        drop_last=False, num_workers=0, use_uva=use_uva)
+    opt = torch.optim.Adam(model.parameters(), lr=0.0005)
```
chang-l, line 119: Previously, it was `lr=0.001, weight_decay=5e-4`; but then it always leads to a very bad testing MRR result (~0.66). Referring to ogb leaderboard [neighbor sampler implementation](https://github.com/snap-stanford/ogb/blob/master/examples/linkproppred/citation2/sampler.py), I adjusted to `lr=5e-4`, and it can result in a very close result to ogb leaderboard after 150 epochs (~0.805). With the default 10 epochs, it's around ~0.74 (see updated README).

examples/pytorch/graphsage/link_pred.py
```diff
@@ -97,78 +80,91 @@ def inference(self, g, device, batch_size, num_workers, buffer_device=None):
             feat = y
         return y
 
-
-def compute_mrr(model, node_emb, src, dst, neg_dst, device, batch_size=500):
+def compute_mrr(model, evaluator, node_emb, src, dst, neg_dst, device, batch_size=500):
```
jermainewang, line 83: Maybe add a comment `"Compute Mean Reciprocal Rank (MRR) in batches."`

examples/pytorch/graphsage/link_pred.py
```diff
@@ -68,27 +54,24 @@ def forward(self, pair_graph, neg_pair_graph, blocks, x):
...
-        h_pos = self.predict(h[pos_src], h[pos_dst])
-        h_neg = self.predict(h[neg_src], h[neg_dst])
+        h_pos = self.predictor(h[pos_src] * h[pos_dst])
+        h_neg = self.predictor(h[neg_src] * h[neg_dst])
         return h_pos, h_neg
 
-    def inference(self, g, device, batch_size, num_workers, buffer_device=None):
-        # The difference between this inference function and the one in the official
-        # example is that the intermediate results can also benefit from prefetching.
+    def inference(self, g, device, batch_size):
```
jermainewang, line 61: Add a comment `"Layer-wise inference algorithm to compute GNN node embeddings."`

jermainewang (approved):

PR #4556, opened 2022-09-14
[Performance][bugfix] Implement `is_unibipartite` in C++ with caching.

### LINKED ISSUE #4521: is_unibipartite check in DGLHeterograph constructor can take longer than message passing kernels
## 🐛 Bug

In DGL Heterograph's constructor, where in the case `ntypes` is a tuple it assumed the graph is unibipartite, and a sanity check inspected the metagraph.
https://github.com/dmlc/dgl/blob/master/python/dgl/heterograph.py#L100

This is accomplished by converting the `src` and `dst` to python `set`s and checking if they have overlap, which is quite slow (100us homogeneous DGLBlocks).

This makes sampling slow, as `to_block()` calls this constructor. It also impacts message passing, as a new DGLHeterograph is constructed per relation type per layer (via `__getitem__(self, etype)`).

### Possible fixes
1. I think the best method is to keep track of unibipartite-ness on the metagraph, and preserve the flag in `to_block()` and `__getitem__(self, etype)`, utilizing the same metagraph, avoiding calling `is_unibipartite` more than just once on the original graph.
1. The simplest way to address this is to change the check to an assertion so users who know their code is correct, can call `python -O` to disable it. I would only expect power users to know to use the `-O` flag, so we would need to emphasize this in the documentation.
1. Alternatively we could still do the check, but in C++. This will be faster, but would still repeat calculation only on the CPU for every mini-batch, so I do not favor this approach.

nv-dlasalle:
## Description
This fixes #4521 by moving the code to C++, and preserving the result when taking a subgraph (e.g., a relation type during message passing). This results in the check taking ~4us when performed and ~2us when a cache result is used.

## Checklist
- [x] Code is well-documented


include/dgl/graph_interface.h
```diff
@@ -149,6 +149,37 @@ class GraphInterface : public runtime::Object {
    */
   virtual bool IsMultigraph() const = 0;
 
+  /*!
+   * \return whetehr the graph is unibipartite
```
BarclayII, line 153: typo

include/dgl/immutable_graph.h
```diff
@@ -1000,6 +1009,9 @@ class ImmutableGraph: public GraphInterface {
   std::string shared_mem_name_;
   // We serialize the metadata of the graph index here for shared memory.
   NDArray serialized_shared_meta_;
+
+  mutable bool is_unibipartite_set_ = false;
+  mutable bool is_unibipartite_ = false;
```
BarclayII, line 1014: Better add a description for both flags like the other private members.

BarclayII (approved):
Minor comments.

I'm also thinking whether the name `self._graph.is_unibipartite` is ambiguous, because a homogeneous graph can also be unidirectional bipartite even if the metagraph is not.  Perhaps change the name to `is_metagraph_unibipartite`?

include/dgl/graph_interface.h
nv-dlasalle: Fixed in 11024ec.

include/dgl/immutable_graph.h
nv-dlasalle: Added descriptions in 11024ec.

nv-dlasalle:
I renamed it to `is_metagraph_unibipartite` and the CAPI to `DGLHeteroIsMetaGraphUniBipartite` in 11024ec.

PR #4620, opened 2022-09-22
[Dist] enable to partition many chunks into less partitions via pipeline

Rhett-Ying:
## Description

`num_chunks` could be larger/equal than `num_parts`. Target for https://github.com/dmlc/dgl/issues/4362
## Checklist
- [ ] Code is well-documented


tools/distpartitioning/convert_partition.py
```diff
@@ -95,11 +89,13 @@ def create_dgl_object(graph_name, num_parts, \
         map between edge type(string)  and edge_type_id(int)
     """
     #create auxiliary data structures from the schema object
-    ntid_dict, global_nid_ranges = get_idranges(schema[constants.STR_NODE_TYPE], 
-                                    schema[constants.STR_NUM_NODES_PER_CHUNK])
+    _, global_nid_ranges = get_idranges(schema[constants.STR_NODE_TYPE],
+                                    schema[constants.STR_NUM_NODES_PER_CHUNK],
+                                    expected_num_chunks=num_parts)
```
jermainewang, line 94: Have you formatted the code with autopep8?

tests/tools/test_dist_part.py
```diff
@@ -152,11 +152,12 @@ def rand_edges(num_src, num_dst, num_edges):
                 assert feat_array.shape[0] == num_edges[etype] // num_chunks
 
         # Step1: graph partition
+        num_parts = 2
```
jermainewang, line 155: Let's add more tests instead of having one test case. Can you add some corner cases too? e.g., num_parts=1, or num_chunks=1, or large num_chunks but small num_parts, or num_chunks is not divisible by num_parts, etc.

tools/distpartitioning/data_shuffle.py
```diff
@@ -550,7 +552,7 @@ def gen_dist_partitions(rank, world_size, params):
     ntypes_gnid_range_map = get_gnid_range_map(node_tids)
     node_data, rcvd_node_features, rcvd_global_nids, edge_data  = \
                     exchange_graph_data(rank, world_size, node_features, node_feat_tids, \
-                                        edge_data, id_lookup, ntypes_ntypeid_map, ntypes_gnid_range_map, \
+                                        edge_data, id_lookup, ntypes_gnid_range_map, \
```
jermainewang, line 555: `\` is not needed in multiline parenthesis.

tools/distpartitioning/data_shuffle.py
```diff
@@ -288,7 +289,7 @@ def exchange_node_features(rank, world_size, node_feature_tids, ntype_gnid_map,
     return own_node_features, own_global_nids
 
 def exchange_graph_data(rank, world_size, node_features, node_feat_tids, edge_data,
-        id_lookup, ntypes_ntypeid_map, ntypes_gnid_range_map, ntid_ntype_map, schema_map):
+        id_lookup, ntypes_gnid_range_map, ntid_ntype_map, schema_map):
```
jermainewang, line 292: Any new argument needs proper docstring.

tools/distpartitioning/dataset_utils.py
```diff
@@ -224,26 +226,38 @@ def get_dataset(input_dir, graph_name, rank, world_size, schema_map):
...
+        for idx in read_list[rank]:
+            edge_file = edge_info[idx]
+            if not os.path.isabs(edge_file):
+                edge_file = os.path.join(input_dir, edge_file)
+            logging.info(f'Loading edges of etype[{etype_name}] from {edge_file}')
+            data_df = csv.read_csv(edge_file,
+                        read_options=pyarrow.csv.ReadOptions(autogenerate_column_names=True), 
+                        parse_options=pyarrow.csv.ParseOptions(delimiter=' '))
+            df_0.append(data_df['f0'].to_numpy())
+            df_1.append(data_df['f1'].to_numpy())
```
jermainewang, line 250: Can we have better names than `'f0'` and `'f1'`?

jermainewang: Same for `df_0` and `df_1`.

tools/distpartitioning/utils.py
```diff
@@ -387,6 +381,8 @@ def get_idranges(names, counts):
         list of node/edge types as strings
     counts : list of lists
         each list contains no. of nodes/edges in a given chunk
+    expected_num_chunks : expected number of returned chunks
+        chunks are collapsed according to expected number of chunks
```
jermainewang, line 385: - What goes after the colon symbol `:` should be the type of the argument.
- The description does not make sense to me. Please reword.

tools/distpartitioning/utils.py
```diff
@@ -402,21 +398,31 @@ def get_idranges(names, counts):
...
         gnid_end += tid_ranges[-1][1]
 
         tid_dict[typename] = tid_ranges
         gid_dict[typename] = np.array([gnid_start, gnid_end]).reshape([1,2])
 
         gnid_start = gnid_end
+        orig_num_chunks = len(tid_start)
+
+    assert expected_num_chunks <= orig_num_chunks, \
+        'expected num_chunks should be less/euqual than original one.'
```
jermainewang, line 417: What is "original one" here? The error message will not make any sense to the users.

jermainewang (changes requested):

tools/dispatch_data.py
```diff
@@ -75,6 +78,8 @@ def main():
     parser.add_argument('--ssh-port', type=int, default=22, help='SSH Port.') 
     parser.add_argument('--process-group-timeout', type=int, default=1800,
                         help='timeout[seconds] for operations executed against the process group')
+    parser.add_argument('--num-parts', type=int, default=0,
+                        help='number of target partitions. If not specified, number of chunks are used.')
```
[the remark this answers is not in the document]
Rhett-Ying: yes. a json file with single line is enough.

tools/dispatch_data.py
```diff
@@ -40,7 +40,10 @@ def submit_jobs(args) -> str:
     with open(os.path.join(args.in_dir, schema_path)) as schema:
         schema_map = json.load(schema)
 
-    num_parts = len(schema_map["num_nodes_per_chunk"][0])
+    num_chunks = len(schema_map["num_nodes_per_chunk"][0])
+    if args.num_parts > num_chunks:
+        raise Exception('Number of partitions should be less/equal than number of chunks.')
+    num_parts = num_chunks if args.num_parts <= 0 else args.num_parts
```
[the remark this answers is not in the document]
Rhett-Ying: I don't think so

tools/distpartitioning/data_shuffle.py
Rhett-Ying: here, I just removed unused argument.

tools/dispatch_data.py
```diff
@@ -33,16 +33,29 @@ def get_launch_cmd(args) -> str:
...
+        with open(partition_path) as part_meta:
+            num_parts = json.load(part_meta)['num_parts']
+    if num_parts > num_chunks:
+        raise Exception('Number of partitions should be less/equal than number of chunks.')
+
+    # verify ip_config
+    with open(args.ip_config, 'r') as f:
+        num_ips = len(f.readlines())
+        assert num_ips == num_parts, \
+            f'Number of lines in {args.ip_config} should equal num_parts[{num_parts}].'
```
jermainewang, line 57: The number of lines ... should be equal to ...

tools/partition_algo/random_partition.py
```diff
@@ -17,6 +17,15 @@ def _random_partition(metadata, num_parts):
         parts = np.random.randint(0, num_parts, (n,))
         array_readwriter.get_array_parser(name='csv').write(ntype + '.txt', parts)
 
+def _dump_metadata(num_parts):
+    """
+    Dump metadata into file. For now, below metadata are dumped:
+    ``version``, ``num_parts``.
+    """
+    metadata = {'version': '0.0.1', 'num_parts': num_parts}
```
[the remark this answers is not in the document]
Rhett-Ying: added. pls take a look

kylasa:
On a general level, This implementation ignores this case when merging different chunks to form a total number of chunks equal to no. of partitions required. Since chunks are treated first come first serve, consider the case where we have num_nodes_per_chunk = [10 ,10, 10, 10, 1] and no. of required partitions is set to `3`. If this is the case, this implementation will result in updated num_nodes_per_chunk = [20, 20, 1]. The same goes for edges as well. Now the load on each process is highly lop-sided. We can assume several such cases where the later processes will have smaller amount of load compared to earlier processes. 

For these above cases, if the nodes/edges per chunk is set to millions or billions then this will result in overly lop-sided processes when doing disk I/O. 

For now, we can treat this as a corner case, but at some point we should address this use-case. I think this is very important  unit test case.

Rhett-Ying:
@kylasa Yes. You're right. Current implementation will hit load imbalance issue. I think one quick/simple fix for this issue is to split the array to `[ [10,10], [10, 1], [10] ]`. This will mitigate the imbalance though not the optimized one: in average. Average workload may introduces much logic, I think.

kylasa (approved):
Overall I agree with the changes... Please take appropriate actions...

tools/distpartitioning/utils.py
```diff
@@ -85,15 +85,9 @@ def get_ntype_featnames(ntype_name, schema_map):
...
-        featnames = []
-        ntype_info = ntype_featdict[ntype_name]
-        for k, v in ntype_info.items(): 
-            featnames.append(k)
-        return featnames
-    else: 
-        return []
+    node_data = schema_map[constants.STR_NODE_DATA]
+    feats = node_data.get(ntype_name, {})
+    return [feat for feat in feats]
```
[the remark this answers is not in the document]
Rhett-Ying: I think it behaves the same as previous one. if `ntype_name` not in `node_data`, empty array will be returned.

tools/partition_algo/base.py
```diff
@@ -0,0 +1,52 @@
+from typing import Optional
+import pydantic as dt
+import json
+from dgl import DGLError
+
+class PartitionMeta(dt.BaseModel):
+    """ Describe fields of partition metadata.
```
jermainewang, line 7:
```
"""Metadata that describes the partition assignment results.

Regardless of the choice of partitioning algorithm, a metadata JSON file will be created
in the output directory which includes the meta information of the partition algorithm.

To generate a metadata JSON:

<give an example>

To read a metadata JSON:

<give an example>
"""
```

tools/partition_algo/base.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    version: Optional[str] = '1.0.0'
+    num_parts: int
```
jermainewang, line 10: Add a field "algo_name"

jermainewang: Also add comment for each field.

tools/partition_algo/base.py
```diff
@@ -0,0 +1,52 @@
...
+    -------
+    PartitionMeta
+        The partition metadata.
+    """
+    with open(meta_file) as f:
+        try:
+            part_meta = PartitionMeta(**(json.load(f)))
+        except dt.ValidationError as e:
+            print(
+                "Details of pydantic.ValidationError:\n{}".format(e.json()))
```
jermainewang, line 43: "Invalid partition metadata JSON. Error details:\{}"

tools/partition_algo/base.py
```diff
@@ -0,0 +1,52 @@
...
+            raise DGLError(
+                "Validation Error for YAML fields. Details are shown above.")
```
jermainewang, line 45: - Why not putting the error message here?
- JSON not YAML

tools/dispatch_data.py
```diff
@@ -33,16 +34,29 @@ def get_launch_cmd(args) -> str:
...
     with open(os.path.join(args.in_dir, schema_path)) as schema:
         schema_map = json.load(schema)
 
-    num_parts = len(schema_map["num_nodes_per_chunk"][0])
     graph_name = schema_map["graph_name"]
 
+    # retrieve num_parts
+    num_chunks = len(schema_map["num_nodes_per_chunk"][0])
+    num_parts = num_chunks
+    partition_path = os.path.join(args.partitions_dir, "partition.json")
```
jermainewang, line 47: Rename it to `metadata.json`?

jermainewang (changes requested):

tools/dispatch_data.py
Rhett-Ying: I think `metadata.json` is used abuse, though we could distinguish according to the base dir.

Rhett-Ying: `partition_meta.json` ?

jermainewang (approved):

PR #4621, opened 2022-09-22
[Sampling] Sampling with edge masks

BarclayII:
## Description
This PR extends the `prob` argument of `sample_neighbors` to accept a boolean tensor to represent which edges are eligible for sampling.  It also adds a `mask` argument to `dgl.dataloading.NeighborSampler` for the same purpose.

```python
g.edata['mask'] = torch.BoolTensor(...)
sampler = NeighborSampler([5, 10, 15], mask='mask')
```

I refrained from adding a `mask` argument in `sample_neighbors` function itself because that means it could change the interface of GraphStorage specification, and I'm not sure what is the workflow about it.

Also included a fix for the issue that if -1 is specified as fanout and some of the edges have probability 0, they can still be selected in the output.

Closes #4354 .

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage
- [x] Code is well-documented
- [x] To the best of my knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [x] Related issue is referred in this PR

PR #4633, opened 2022-09-24
[Sampling] Sampling with edge masks

### LINKED ISSUE #4354: Implement sampling with edge masks for NeighborSampler and test on single-machine/distributed setting
## 🔨Work Item

**IMPORTANT:**

## Description

- [x] #4364 
- [x] #4614
- [x] ~~Add edge mask to `sample_neighbors` function~~ (handled in #4364)
- [x] ~~Add edge mask to `sample_etype_neighbors` function~~ (handled in #4364)
- [ ] #4619

### LINKED ISSUE #4441: [RFC] [Sampling] Sampling with edge masks
## 🚀 Feature
Support sampling with a boolean edge mask indicating which edges can be included in message passing.

## Motivation
During validation, only the edges in the training set should be considered for message passing.  During testing, the edges in both the training set and the validation set should be considered for message passing.  If we have sampling methods that support edge masks, the user can just create a single graph with both the training and validation edges, and provide different masks to the sampler so that different subsets can be used during sampling for minibatch message passing.

## Alternatives
Construct two graphs, one with all the edges and the other with only the masked edges.  The problem with this approach is that it will consume twice as much space when the proportion of the masked edges is small, making it unsuitable for large-scale graphs.  Supporting two graphs in distributed training is also difficult since either the user will have to repartition the graph, or a significant refactor is necessary, because currently the distributed training design (load balancing, data partitioning, etc.) is centered around the assumption of having only a single graph.

## Pitch
Have an `include_mask: str` argument for the following methods:
* `subgraph`
* `edge_subgraph`
* `in_subgraph`
* `out_subgraph`
* `sample_neighbors`
* `sample_topk`

And maybe the following interfaces:
* `node_type_subgraph`
* `edge_type_subgraph`
* `random_walk`
* `node2vec`

The `include_mask` argument will be a string indicating the name of the boolean mask on the edges.  Only the edges with a True mask can appear in the results from the methods above.  Or in other words, the edges with a False mask will be skipped during sampling.

Example:
```python
g = dgl.graph(([0,1,2,3],[1,2,3,4]))
g.edata['mask'] = torch.BoolTensor([True, False, True, True])
sg = g.subgraph([0, 1, 2])
sg.edges()    # ([0, 1], [1, 2])
sg = g.subgraphs([0, 1, 2], include_mask='mask')
sg.edges()    # ([0, 1])
```

BarclayII:
## Description
This PR extends the `prob` argument of `sample_neighbors` to accept a boolean tensor to represent which edges are eligible for sampling.  It also adds a `mask` argument to `dgl.dataloading.NeighborSampler` for the same purpose.

```python
g.edata['mask'] = torch.BoolTensor(...)
sampler = NeighborSampler([5, 10, 15], mask='mask')
```

I refrained from adding a `mask` argument in `sample_neighbors` function itself because that means it could change the interface of GraphStorage specification, and I'm not sure what is the workflow about it.

Also included a fix for the issue that if -1 is specified as fanout and some of the edges have probability 0, they can still be selected in the output.

Closes #4354 .  Also closes #4441 .
Replaces #4621.

## Checklist
- [x] Code is well-documented


BarclayII:
I realized that the local partition graph does not contain any node data and edge data of that partition.  Instead, they are loaded into the KVStore instead: see
https://github.com/dmlc/dgl/blob/master/python/dgl/distributed/dist_graph.py#L335-L336
and
https://github.com/dmlc/dgl/blob/master/python/dgl/distributed/dist_graph.py#L375

Does it make sense to put the features inside the local partition DGLGraph (i.e. `self.client_g`)?
@Rhett-Ying

BarclayII:
Turns out that the workload is much larger than I initially anticipated.

The original `sample_etype_neighbors` interface assumed that the probabilities of all edge types are stored in the same single tensor, which is not the case for distributed training.  I reserved a week as a buffer for possible issues and then spent a week changing the interface.

I only tested single machine so far.  Will do more rigorous testing on multiple machine setting in the next cycle.

python/dgl/distributed/graph_services.py
```diff
@@ -138,6 +138,30 @@ def _in_subgraph(local_g, partition_book, seed_nodes):
     global_eids = F.gather_row(local_g.edata[EID], sampled_graph.edata[EID])
     return global_src, global_dst, global_eids
 
+# --- NOTE 1 ---
+# BIG HACK AHEAD.
```
jermainewang, line 142: Could you rewrite it in as a formal docstring. Right now it reads like a casual note for yourself.
- Instead of claiming it as a hack, it is more like a current limitation to me.
- Avoid "I do ...". Use "The implementation does ...".
- "We do ..." is ok but should be used with caution. For example, "If we only use edge data, everything will be fine because the local edge data ..." sounds very casual. Consider change it to "The current implementation only works for sampling algorithms that rely on the data of the seed nodes and neighboring edges because they are locally available. It is not applicable to algorithms that rely on the data of neighboring nodes such as ...".
- "If we don't want to overhaul the design of DistDGL, the best thing I can think of is to cache the halo node features, which (1) incurs a lot more extra memory cost, and (2) needs significant refactor on DistGraph and/or DistTensor." -> This should belong to a design doc or an issue. The general audience will not fully understand this.

Think of you are writing a paper or essay. You will never use some of the tones here. The idea is the same for docstring/comment.

python/dgl/distributed/graph_services.py
```diff
@@ -158,45 +182,52 @@ def __getstate__(self):
     def process_request(self, server_state):
         local_g = server_state.graph
         partition_book = server_state.partition_book
-        global_src, global_dst, global_eids = _sample_neighbors(local_g, partition_book,
-                                                                self.seed_nodes,
-                                                                self.fan_out, self.edge_dir,
-                                                                self.prob, self.replace)
+        kv_store = server_state.kv_store
+        if self.prob is not None:
+            prob = kv_store.local_partition[self.prob]
```
jermainewang, line 187: Use `prob_or_mask` to indicate that it can be a mask.

python/dgl/distributed/graph_services.py
jermainewang: I also think what you described here is the general design philosophy (or restriction) of the entire `graph_services.py` so you could put it to the beginning of file.

python/dgl/distributed/graph_services.py
```diff
@@ -504,12 +538,29 @@ def sample_etype_neighbors(g, nodes, etype_field, fanout, edge_dir='in',
...
+            return _sample_etype_neighbors(
+                    local_g, partition_book, local_nids, etype_field,
+                    eid_field, fanout, edge_dir, None, replace,
+                    etype_sorted=etype_sorted)
+        else:
+            _prob = [g.edges[etype].data[prob].local_partition for etype in g.etypes]
+            return _sample_etype_neighbors(
+                    local_g, partition_book, local_nids, etype_field,
+                    eid_field, fanout, edge_dir, _prob, replace,
+                    etype_sorted=etype_sorted)
```
jermainewang, line 563: L554-L557 and L560-L563 are very similar. Can be merged with an `if` branch to get the right `_prob` array.

python/dgl/distributed/kvstore.py
```diff
@@ -1394,6 +1394,12 @@ def count_nonzero(self, name):
             total += res.num_local_nonzero
         return total
 
+    @property
+    def local_store(self):
+        """Return the local partition of the data storage.
```
jermainewang, line 1399: Could you doc the return type?

python/dgl/distributed/standalone_kvstore.py
```diff
@@ -103,6 +103,12 @@ def count_nonzero(self, name):
         """
         return F.count_nonzero(self._data[name])
 
+    @property
+    def local_store(self):
+        """Return the local partition of the data storage.
```
jermainewang, line 108: same comment here

python/dgl/sampling/neighbor.py
```diff
@@ -14,9 +14,27 @@
     'sample_neighbors_biased',
     'select_topk']
 
-def sample_etype_neighbors(g, nodes, etype_field, fanout, edge_dir='in', prob=None,
-                           replace=False, copy_ndata=True, copy_edata=True, etype_sorted=False,
-                           _dist_training=False, output_device=None):
+def _prepare_edge_arrays(g, arg):
```
jermainewang, line 17: briefly doc what this internal function is for

src/array/array.cc
```diff
@@ -555,7 +555,9 @@ COOMatrix CSRRowWiseSampling(
     // prob is pinned and rows on GPU is valid
     CHECK_VALID_CONTEXT(prob, rows);
     ATEN_CSR_SWITCH_CUDA_UVA(mat, rows, XPU, IdType, "CSRRowWiseSampling", {
-      ATEN_FLOAT_TYPE_SWITCH(prob->dtype, FloatType, "probability", {
+      CHECK(!(prob->dtype.bits == 8 && XPU == kDGLCUDA)) <<
```
jermainewang, line 558: Is this a complete check? Miss type code check?

jermainewang (changes requested):
python code comments

python/dgl/distributed/graph_services.py
```diff
@@ -504,12 +538,29 @@ def sample_etype_neighbors(g, nodes, etype_field, fanout, edge_dir='in',
             homo_nids.append(gpb.map_to_homo_nid(typed_nodes, ntype))
         nodes = F.cat(homo_nids, 0)
     def issue_remote_req(node_ids):
-        return SamplingRequestEtype(node_ids, etype_field, fanout, edge_dir=edge_dir,
-                                    prob=prob, replace=replace, etype_sorted=etype_sorted)
+        if prob is not None:
+            # See NOTE 1
+            # Is there a better way to get the KVStore key given the name and
+            # node/edge type?
```
jermainewang, line 544: remove L543 - L544

jermainewang:
Converted this to draft PR. Not for review.

jermainewang:
Closed as the function has been merged via #4749  and #4748

PR #4635, opened 2022-09-26
[Fix][Unittest] Fix test stream

yaox12:
## Description

Fix the random failure of `test_record_stream_graph_negative`.

Resolves https://github.com/dmlc/dgl/issues/4436.

## Checklist
- [x] Code is well-documented


tests/pytorch/test_stream.py
```diff
@@ -126,14 +127,18 @@ def perform_computing():
     with torch.cuda.stream(stream):
         # since we have called record stream for g2, g3 won't reuse its memory
         g3 = rand_graph(10, 20, device=F.ctx())
+        g3.create_formats_()
     torch.cuda.current_stream().synchronize()
     assert torch.equal(result, results2)
 
 @unittest.skipIf(F._default_context_str == 'cpu', reason="stream only runs on GPU.")
+# this test must follow test_record_stream_graph_positive to ensure
+# enough memory blocks have been reserved.
```
chang-l, line 136: Could you please elaborate more on the comment here? 
What if you release all reserved memory by `empty_cache()` before negative test. Would it fail?

yaox12: Running `test_record_stream_graph_negative()` solely will fail. I don't know why...

jermainewang:
Since I'm not familiar with the effort, I will let @chang-l approve the PR.

tests/pytorch/test_stream.py
chang-l: I profiled it and it seems from the sync attributed to `cusparsecreate()` call (invoking cudaFree() underneath) when we first calling spMM function. https://github.com/dmlc/dgl/blob/d78a3a4baf611b90871a849f58647160c7cd9ab4/src/array/cuda/spmm.cuh#L231 

And because of that sync, if we just call negative test, computation cannot really overlap with data cpy and it will always fail.

yaox12: I have updated both the positive and negative tests to initialize the cusparse handle before computing.

chang-l (approved):
Looks good to me.

PR #4677, opened 2022-10-06
[EXAMPLE]Add device support in hetero_rgcn

xuechendi:
Signed-off-by: Xue, Chendi <chendi.xue@intel.com>

## Description

When testing this example with colab(GPU) and CPU, found that since graph is not binding to the device, which leaded to both GPU and CPU performance were lower than expectation. 
To fix this, add to(device) support in this example.

## Checklist
- [ ] Code is well-documented

## Changes
- [x] add to(device) in prepare_data
- [x] add enable_cpu_affinity in prepare_data and train


xuechendi:
Hi @mufeili , I saw you are the owner for this example, please help to review this PR see if it makes sense to you.
From my evaluation, running this script on colab speeding up from around ~1200 it/s -> ~6000it/s

examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py
```diff
@@ -29,13 +30,16 @@ def prepare_data(args):
 
     # train sampler
     sampler = dgl.dataloading.MultiLayerNeighborSampler([25, 20])
+    num_workers = 0 if device != "cpu" else int(os.cpu_count() / 2)
```
mufeili, line 33: I will suggest adding an argument to `ArgumentParser` and setting the default value to 0.

examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py
```diff
@@ -392,18 +397,36 @@ def main(args):
...
-            labels,
-            logger,
-            device,
-            run,
-        )
+        if device == "cpu":
+            cores = list(range(os.cpu_count()))
+            len_cores = len(cores)
+            dl_cores = cores[:int(len_cores / 2)]
+            with train_loader.enable_cpu_affinity(dl_cores, [i for i in cores if i not in dl_cores]):
```
mufeili, line 404: Are you familiar with this practice? Could you take a look? @BarclayII @chang-l @yaox12

examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py
```diff
@@ -30,7 +32,7 @@ def prepare_data(args, device):
 
     # train sampler
     sampler = dgl.dataloading.MultiLayerNeighborSampler([25, 20])
-    num_workers = 0 if device != "cpu" else int(os.cpu_count() / 2)
+    num_workers = args.num_workers
```
xuechendi, line 35: @mufeili , I changed to use arguments here

examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py
xuechendi: @mufeili , I added a pytorch version check here, since enable_cpu_affinity is enabled in pytorch 1.12. CPU affinity is basically bind partial cpu only for dataloader and others for computing. 
image

chang-l: I have some questions/suggestions listed here: 
> since enable_cpu_affinity is enabled in pytorch 1.12  

(1)  It does not make sense to me... Since `enable_cpu_affinity` is a dgl method, why it has to be restricted to pytorch 1.12+?  Can you pls provide more info here why it fails with pytorch <1.12?

(2) I don't think using `enable_cpu_affinity` as `with train_loader.enable_cpu_affinity(dl_cores, [i for i in cores if i not in dl_cores]):` is a good practice for general users (too much low-level exposure). Would it be enough to use default `with dataloader.enable_cpu_affinity():` as long as `num_workers` is specified properly?

xuechendi: @chang-l , enable_cpu_affinity is actually a feature of pytorch, and it appears to me that pytorch version smaller than 1.12 won't be able to recognize this API, and exception will be raised.
image

And I found the default loader_cores and compute_cores are None, so no argument provided in with dataloader.enable_cpu_affinity() and set num_workers > 0 would also trigger an exception.
image

image

chang-l: For (1), are you sure you were using the same version of dgl for your pytorch 1.10 run? 
For (2), I think the crash is due to your setting of `num_of_workers = os.cpu_count/2`. According to the [doc](https://docs.dgl.ai/en/0.9.x/tutorials/cpu/cpu_best_practises.html), you have to make sure your compute_core + loader_core <= num_core_per_socket (single NUMA node).  I guess you were running on a 2x48 core cpu? If so, please make sure `num_workers<48`; or simply remove the assignment `num_workers = int(os.cpu_count() / 2)`, and use cmd instead.

xuechendi: @chang-l , for (1), not exactly the same, but both are above 0.9.0. For 1.10, dgl is 0.9.0, and for 1.12, dgl is 0.9.1
(2) Yes, That is my way to prove why I used `train_loader.enable_cpu_affinity(dl_cores, [i for i in cores if i not in dl_cores])` instead of `with dataloader.enable_cpu_affinity()`, I found tuning num_workers from 0 to half of cores helped to improve CPU side performance roughly by 20%

chang-l: I think the discrepancy between 0.9.0 and 0.9.1 might explain your first crash. Can you somehow check your installed source code path by doing `pip show dgl` and see if `enable_cpu_affinity` is defined in SourcePath/dgl/dataloading/dataloader.py? I guess your 1.10 pytorch + 0.9.0(dgl) missing the def of such function.

For (2), I see... But you are using 2x48 cores (48 for load, 48 for compute) from both sockets to gain 1.2X, right?
Could you please perform another study on the scaling? I mean, comparing your 2x48 results to the single socket default `with dataloader.enable_cpu_affinity()` with `--num_workers 24`, assuming 48 is your nprocs_per_socket.

BTW, you need to remove your L381, to respect the cmd setting of num_workers

examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py
```diff
@@ -365,13 +371,15 @@ def test(g, model, node_embed, y_true, device, split_idx):
 
     return train_acc, valid_acc, test_acc
 
+def is_support_affinity(v_t):
+    return v_t >= "0.9.1"
```
xuechendi, line 375: @chang-l , you're right, I found it is 0.9.0 DataLoader is not supporting 'enable_cpu_affinity' yet, modified check func here

examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py
```diff
@@ -383,27 +391,49 @@ def main(args):
...
-            embed_layer,
-            optimizer,
-            train_loader,
-            split_idx,
-            labels,
-            logger,
-            device,
-            run,
-        )
+        if args.num_workers != 0 and device == "cpu" and is_support_affinity(v_t):
```
xuechendi, line 407: @chang-l , only when `num_workers `are configured, will do `enable_cpu_affinity `here. Which prevent previous method of using full system cores, and I tried using 24 here, performance is quite similar to using 48, which shows around 20% benefit comparing with not using `enable_cpu_affinity `

examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py
xuechendi: @chang-l , thanks for the comment, I fixed the codes according to your above suggestions. And you're, dgl version is root cause. And num_workers as 24 also bring performance benefit, so I removed my hard code of obtain system num_cores and simply use input num_workers configuration.
image
image

examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py
chang-l: Thanks. I still don't understand why you need loader_cores and compute_cores specified in `enable_cpu_affinity()` at L411. Can you remove L408-L411 and replace with `with train_loader.enable_cpu_affinity()`?
Also, can you pls verify that it is still working when GPU is available (device = 'cuda')?

xuechendi: @chang-l , when replacing with `with train_loader.enable_cpu_affinity()` and set num_workers = 24, it will go crash as the screenshot I pasted last time. That is why `loader_cores` and `compute_cores` are specified.
image

And yes, GPU is working, the PR is both for GPU and CPU performance ceoptimization, for GPU performance improved by 3x withL322 and L382. I added a switch at L407, so GPU won't go `enable_cpu_affinity()`

examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py
```diff
@@ -383,27 +394,49 @@ def main(args):
...
+            expected_max = int(psutil.cpu_count(logical=False)/2)
+            if args.num_workers > expected_max:
+                print(f"[ERROR] You specified num_workers are larger than physical cores, please set any number less or equal to {expected_max}", file=sys.stderr)
+            with train_loader.enable_cpu_affinity():
```
xuechendi, line 414: @chang-l , I cleaned args in `enable_cpu_affinity()` and add some error msg in case num_workers is mis-configured.

examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py
```diff
@@ -383,27 +394,49 @@ def main(args):
...
-            optimizer,
-            train_loader,
-            split_idx,
-            labels,
-            logger,
-            device,
-            run,
-        )
+        if args.num_workers != 0 and device == "cpu" and is_support_affinity(v_t):
+            expected_max = int(psutil.cpu_count(logical=False)/2)
```
chang-l, line 411: I think the physical core count should be `psutil.cpu_count(logical=False)`. Can you double-check?
Also, below if condition (L412) should be `args.num_workers >= expected_max:`, otherwise it will crash.

xuechendi: @chang-l , yes, I checked before and that was also why I thought ld_cores/computer_cores are necessary,
I have nproc as 96, and psutil.cpu_count(logical=False) as 48, but if I set num_workers as 48 while leave `enable_cpu_affinity()` empty with args, it go crashing. And set num_workers as 24, it is running well

chang-l: Thanks @xuechendi . Sorry for the late reply. 
I think it crashed because of this line?https://github.com/dmlc/dgl/blob/b13092172db668c26da487e8f782c7cf1caa92a6/python/dgl/dataloading/dataloader.py#L930-L931
which means your num_workers should be _strictly_ less than physical cores (not equal). Can you please try setting num_workers to 47? That's also why I said: 
>  below if condition (L412) should be args.num_workers >= expected_max:

xuechendi: @chang-l , I see, I fixed the expected num_workers check in codes, and verified with num_workers as 47, appeared to be runnable
image

examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py
```diff
@@ -408,8 +408,8 @@ def main(args):
         optimizer = th.optim.Adam(all_params, lr=0.01)
 
         if args.num_workers != 0 and device == "cpu" and is_support_affinity(v_t):
-            expected_max = int(psutil.cpu_count(logical=False)/2)
-            if args.num_workers > expected_max:
+            expected_max = int(psutil.cpu_count(logical=False))
+            if args.num_workers >= expected_max:
```
xuechendi, line 412: @chang-l , fixed here

examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py
```diff
@@ -383,27 +394,49 @@ def main(args):
...
-            split_idx,
-            labels,
-            logger,
-            device,
-            run,
-        )
+        if args.num_workers != 0 and device == "cpu" and is_support_affinity(v_t):
+            expected_max = int(psutil.cpu_count(logical=False))
+            if args.num_workers >= expected_max:
+                print(f"[ERROR] You specified num_workers are larger than physical cores, please set any number less or equal to {expected_max}", file=sys.stderr)
```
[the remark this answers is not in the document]
xuechendi: @chang-l , committed the fix

chang-l (commented):
I dropped few comments. Otherwise, looks good to me.

xuechendi:
> I dropped few comments. Otherwise, looks good to me.

@chang-l , got it, removed num_workers and dgl version print

chang-l (approved):

chang-l:
@xuechendi Can you please add comment within `is_support_cpu_affinity` function? Especially, explain why 0.9.1 is specified there. (e.g., `return v_t >= "0.9.1"  # (cpu)affinity-aware dataloading only available with dgl 0.9.1+`)

@mufeili  Can you pls take a look before merging? 
@dgl-bot To trigger CI.

examples/pytorch/ogb/ogbn-mag/hetero_rgcn.py
```diff
@@ -365,13 +372,16 @@ def test(g, model, node_embed, y_true, device, split_idx):
 
     return train_acc, valid_acc, test_acc
 
+def is_support_affinity(v_t):
+    # dgl support enable_cpu_affinity since 0.9.1
```
chang-l, line 376:
```suggestion
    # dgl supports enable_cpu_affinity since 0.9.1
```

chang-l:
@dgl-bot

Issue #4510, opened 2022-09-03
Reduce the peak memory in DistDGL


zheng-da:
## 🔨Work Item

**IMPORTANT:**
* This template is only for dev team to track project progress. For feature request or bug report, please use the corresponding issue templates.
* DO NOT create a new work item if the purpose is to fix an existing issue or feature request. We will directly use the issue in the project tracker.

Project tracker: https://github.com/orgs/dmlc/projects/2

## Description

We need to reduce the peak memory in DistDGL. Currently, it uses too much unnecessary memory.
* avoid checking the data correctness when loading the graph structure: https://github.com/dmlc/dgl/blob/master/python/dgl/distributed/partition.py#L100-L119
We can enable the check in the debug mode.
* when init_data is called, it copies data to shared memory. after calling init_data, we should dereference to the tensor and call garbage collection to force python to free memory.
* the DGLGraph object has some unnecessary node/edge data. One example is the node data and edge data of 'orig_id'.
* The DGLGraph object has some fields, such as node type and edge type, which uses too many bytes. For example, we can potentially use 2 bytes to store node type and edge type.

frozenbugs:
Priority: to be confirmed with Da.

frozenbugs:
Hi Da, as we discussed, it is better to itemize the work for us to have better understanding of the priority. For now, I am assign this back to you and lower to Medium.
Feel free to file new issues, and change the priority back to High.

zheng-da:
i have itemized the work. i think we only need to fix the 4 items i listed above.

frozenbugs:
Hi Rui, please re-estimate the workload based on the update.

Rhett-Ying:
re-estimated and PR is sent for review...

PR #4687, opened 2022-10-08
[Dist] Reduce peak memory in DistDGL

Rhett-Ying:
## Description

all the action items listed in https://github.com/dmlc/dgl/issues/4510 have been handled.

## Checklist
- [ ] Code is well-documented


python/dgl/distributed/dist_graph.py
```diff
@@ -381,6 +377,10 @@ def __init__(self, server_id, ip_config, num_servers,
                 self.init_data(name=str(data_name), policy_str=data_name.policy_str,
                                data_tensor=node_feats[name])
                 self.orig_data.add(str(data_name))
+            del node_feats
+            gc.collect()
```
jermainewang, line 381: Look like L370-381 is to get the feature names but it loads the entire feature data instead. Is it possible to read them from some metadata files?

python/dgl/distributed/partition.py
```diff
@@ -96,27 +114,27 @@ def load_partition(part_config, part_id, load_feats=True):
...
-        eids = F.boolean_mask(graph.edata[EID], _get_inner_edge_mask(graph, etype_id))
-        partids1 = gpb.eid2partid(eids)
-        _, per_type_eids = gpb.map_to_per_etype(eids)
-        partids2 = gpb.eid2partid(per_type_eids, etype)
-        assert np.all(F.asnumpy(partids1 == part_id)), 'load a wrong partition'
-        assert np.all(F.asnumpy(partids2 == part_id)), 'load a wrong partition'
-        etypes_list.append(etype)
+    ntypes_list = list(ntypes.keys())
+    etypes_list = list(etypes.keys())
+    if 'DGL_DIST_DEBUG' in os.environ:
```
jermainewang, line 119: We really need a docpage about all the environment variables used in DistDGL. A temporary solution is to add a page to the distributed training user guide.

python/dgl/distributed/partition.py
```diff
@@ -96,27 +114,27 @@ def load_partition(part_config, part_id, load_feats=True):
...
+        for ntype in ntypes:
+            ntype_id = ntypes[ntype]
+            # graph.ndata[NID] are global homogeneous node IDs.
+            nids = F.boolean_mask(graph.ndata[NID], _get_inner_node_mask(graph, ntype_id))
+            partids1 = gpb.nid2partid(nids)
+            _, per_type_nids = gpb.map_to_per_ntype(nids)
+            partids2 = gpb.nid2partid(per_type_nids, ntype)
+            assert np.all(F.asnumpy(partids1 == part_id)), 'load a wrong partition'
+            assert np.all(F.asnumpy(partids2 == part_id)), 'load a wrong partition'
```
jermainewang, line 128: Please change them to more informative error message.

python/dgl/distributed/partition.py
```diff
@@ -96,27 +114,27 @@ def load_partition(part_config, part_id, load_feats=True):
...
+        for etype in etypes:
+            etype_id = etypes[etype]
+            # graph.edata[EID] are global homogeneous edge IDs.
+            eids = F.boolean_mask(graph.edata[EID], _get_inner_edge_mask(graph, etype_id))
+            partids1 = gpb.eid2partid(eids)
+            _, per_type_eids = gpb.map_to_per_etype(eids)
+            partids2 = gpb.eid2partid(per_type_eids, etype)
+            assert np.all(F.asnumpy(partids1 == part_id)), 'load a wrong partition'
+            assert np.all(F.asnumpy(partids2 == part_id)), 'load a wrong partition'
```
jermainewang, line 137: Same here.

python/dgl/distributed/partition.py
```diff
@@ -13,6 +13,24 @@
 from ..partition import metis_partition_assignment, partition_graph_with_halo, get_peak_mem
 from .graph_partition_book import BasicPartitionBook, RangePartitionBook
 
+FIELD_DICT = {'inner_node': F.uint8,    # A flag indicates whether the node is inside a partition.
```
jermainewang, line 16: Let's have a more readable name. How about `RESERVED_FIELD_DTYPE` ?

python/dgl/distributed/partition.py
```diff
@@ -873,10 +901,10 @@ def get_homogeneous(g, balance_ntypes):
...
                     else:
                         edge_feats[etype + '/' + name] = g.edges[etype].data[name]
-        # Some adjustment for heterogeneous graphs.
-        if not g.is_homogeneous:
-            part.ndata['orig_id'] = F.gather_row(sim_g.ndata[NID], part.ndata['orig_id'])
-            part.edata['orig_id'] = F.gather_row(sim_g.edata[EID], part.edata['orig_id'])
+        # delete `orig_id` from ndata/edata
+        if reshuffle:
+            del part.ndata['orig_id']
+            del part.edata['orig_id']
```
jermainewang, line 907: Instead of just patching it at the end, could you take the chance to refactor the code too? For example, could we not generate/save this `'orig_id'` at the beginning?

tools/distpartitioning/utils.py
```diff
@@ -390,7 +390,7 @@ def write_graph_dgl(graph_file, graph_obj):
     graph_file : string
         File name in which graph object is serialized
     """
-    dgl.save_graphs(graph_file, [graph_obj])
+    dgl.distributed.partition._save_graphs(graph_file, [graph_obj])
```
jermainewang, line 393: Is this change needed?

jermainewang (changes requested):

python/dgl/distributed/dist_graph.py
Rhett-Ying: no. feature data is really used in `self.init_data()`. we really need to load the feature data.

python/dgl/distributed/partition.py
Rhett-Ying: how about achieving this in another dedicated PR?

tools/distpartitioning/utils.py
Rhett-Ying: just to make sure the saved graph has expected dtype.

python/dgl/distributed/partition.py
Rhett-Ying: orig_nids/eids are required during update `part_g.ndata[NTYPE]/edata[ETYPE]`. so we have to generate. I think we could avoid attach such data into `part_g.n/edata`.

python/dgl/distributed/partition.py
Rhett-Ying: ok

python/dgl/distributed/partition.py
jermainewang: sure

python/dgl/distributed/partition.py
Rhett-Ying: seems `'orig_id'` is always attached if `reshuffle=True` as shown below.
https://github.com/dmlc/dgl/blob/303b150f8c6b6dbd6ea2b3ca724ded64ae14cb5f/python/dgl/partition.py#L140-L152

Rhett-Ying: 'orig_id' is useful and user may rely on it directly. as a result, I think we'd better leave `orig_id` in `part_g.n/edata`.

jermainewang: ok

jermainewang (approved):
LGTM

python/dgl/distributed/partition.py
chwan1016: @Rhett-Ying Can I revert the change of these lines? 'orig_id' is really useful for me to debug a code. With this key, I can verify whether the node embedding in a subgraph is identical to the result in the original graph.

Rhett-Ying: @chwan1016 I'm afraid not. You need to change your test logic: fetch such info from the saved files.

chwan1016: @Rhett-Ying Could you please be more specific? `orig_id` is deleted in this piece of code and it does not exist in any file. Thanks.

chwan1016: Oh, great! I didn't notice the `return_mapping` param before. Thanks!

chwan1016: @Rhett-Ying Can we replace `return_mapping` with another param like `store_mapping` (similar with `store_id` in `node_subgraph()`)? Storing the `orig_id` with other graph data together would make the file structure cleaner. Thanks!

Rhett-Ying: It's intended to store `orig_id` separately from graph data as we don't always need such info together with graph data when loading a graph for distributed train even we saved such info.

PR #4701, opened 2022-10-11
[Do not merge][Feature] Create an accelerated RGCN model using cugraph-ops

### LINKED ISSUE #4669: [FEA] Create RGCN model using cugraph-ops aggregators
## 🚀 Feature
This integration creates an accelerated R-GCN model that uses the aggregation functions from `pylibcugraphops`, a Python wrapper around cugraph-ops aggregators.

Per our previous discussion, this new model `dgl.contrib.cugraph.nn.RelGraphConv` should mimic the interface of existing model `dgl.nn.RelGraphConv` and should first go to the `contrib` folder first. This accelerated model will only support basis decomposition (in terms of weight regularization) and requires DGL sampled graph as input. It is not intended for full-graph use cases.

## Motivation

Provide a faster RGCN model on GPU to DGL users

tingyu66:
## Description

This PR adds an accelerated R-GCN model that uses the aggregation functions in `pylibcugraphops`, a Python wrapper of `cugraph-ops`.

Closes https://github.com/rapidsai/cugraph-ops/issues/179.
Closes #4669.
Link to internal reviews https://github.com/rapidsai/dgl/pull/15.

To install `pylibcugraphops`:
```
conda install -c nvidia pylibcugraphops=22.10
```

The class `dgl.contrib.cugraph.nn.RelGraphConv` uses a pytorch backend and offers similar functionality and interface as [`dgl.nn.RelGraphConv`](https://docs.dgl.ai/en/latest/generated/dgl.nn.pytorch.conv.RelGraphConv.html), with the only difference being:
- `regularizer` has to be either `None` or `'basis'` (not supporting block-diagonal regularization)
- The `__init__()` function of the class take an extra parameter `fanout` as input. This is because the aggregators in `pylibcugraphops` are designed for mini-batch training and *not* suitable for full-graph use cases. Users should pass in the same `fanout` value used for creating DGL's sampler. Consequently, the `forward` function expect input graphs to be a `DGLBlock` (message flow graph, or MFG).

This PR includes an entity classification example `entity_sample_cugraphops.py`. Compared to the original dgl [example](https://github.com/dmlc/dgl/blob/master/examples/pytorch/rgcn/entity_sample.py), users would only need to change the imports and pass in `fanouts` to the model in order to use this cugraphops-accelerated variant. One caveat is that when the `test_sampler` has a different `fanouts` than the `train_sampler`, we have to update the fanouts in the conv layers using `update_fanouts()`.

This PR also adds tests and an API documentation page. 

## Performance
Hardware: 32-core AMD Ryzen Threadripper Pro 3975WX, NVIDIA RTX A6000
Benchmark: Mini-batch `entity_sample.py` example:
- batch_size: 100, graph idtype is int64.
- 2-layer RGCN, with fanouts: (4,4), hidden embedding dimension: 16, num_bases: 20
- The timings are for 100 epochs and the average across 3 identical runs.

|       | time_dgl (s) | time_cugraph-ops (s) |
|:-----:|:------------:|:--------------------:|
|  aifb |     4.24     |         2.61         |
| mutag |     5.65     |         3.68         |
|  bgs  |     4.19     |         2.70         |
|   am  |     19.11    |         13.71        |

During forward aggregation, most of the time (~70%) is spent on the conversion from DGL graph to cugraph-ops MFG type, i.e., `get_csr`:  grab csr representation of the DGL graph and `create_mfg`: create cugraph-ops MFG. The actual forward aggregation function only takes 1/5 of the total time. 

```
Total Time (ns)  Instances   Avg (ns)     Med (ns)    Min (ns)  Max (ns)   StdDev (ns)   Style       Range    
---------------  ---------  -----------  -----------  --------  ---------  -----------  -------  -------------
    834,193,721        814  1,024,808.0  1,015,957.0   957,171  2,264,271     96,833.9  PushPop  total_agg_fwd
    395,471,358        814    485,837.0    474,488.0   453,493  1,577,778     78,678.5  PushPop  create_mfg   
    182,373,749        814    224,046.4    233,108.5   195,217    398,820     22,571.4  PushPop  get_csr      
    159,149,002        814    195,514.7    193,233.5   185,399    488,559     13,795.3  PushPop  agg_fwd_func
```
This overhead can be avoided if DGL natively supports CSR graphs.

### Accuracy: Entity Classification

test accuracy after 100 epochs, using 20 bases

| Dataset   |  Mini-batch
| ------------ |  ------
| aifb         | ~0.92
| mutag     | ~0.58
| bgs         | ~0.70
| am          | ~0.64

## Checklist
- [x] Code is well-documented


frozenbugs:
Please beautify your code with :https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit

mufeili:
@frozenbugs Perhaps you should consider modifying [this file](https://github.com/dmlc/dgl/blob/master/.github/PULL_REQUEST_TEMPLATE.md).

czkkkkkk:
@dgl-bot

czkkkkkk:
The achieved accuracy is different from DGL. Why is that?

python/dgl/contrib/cugraph/nn/conv/relgraphconv.py
```diff
@@ -0,0 +1,383 @@
+"""Torch Module for Relational graph convolution layer using cugraph-ops"""
+# pylint: disable= no-member, arguments-differ, invalid-name
+import math
+
+import torch as th
+from pylibcugraphops.aggregators.node_level import (
+    agg_hg_basis_post_bwd_int32,
```
czkkkkkk, line 7: Is there any document about these functions?

python/dgl/contrib/cugraph/nn/conv/relgraphconv.py
```diff
@@ -0,0 +1,383 @@
...
+        -------
+        torch.Tensor
+            New node features. Shape: :math:`(|V|, D_{out})`.
+        """
+        _sm_size = (
+            self.fanout * 2 + 3 + self.num_rels * self.num_bases
+            if self.regularizer == "basis"
+            else 0
+        )
+        _sm_size *= 8 if g.idtype == th.int64 else 4
```
czkkkkkk, line 356: Can we have a version when the shared memory is not sufficient? Also is there any other elegant way to get the shared memory size instead of specifying it as a constant?

python/dgl/contrib/cugraph/nn/conv/relgraphconv.py
tingyu66: We are planning the open-sourcing of `pylibcugraphops`, but before it happens, we do not have a public-facing documentation. Users have to refer to the docstrings of these functions for now.

python/dgl/contrib/cugraph/nn/conv/relgraphconv.py
tingyu66: > Can we have a version when the shared memory is not sufficient?

The current aggregation functions in `cugraph-ops` were designed for sampled graph, and 48KB shared memory should be enough for typical `fanout` and `num_rels` values. We are working on a full-graph variant, which can serve as a fallback option in the future.

> Also is there any other elegant way to get the shared memory size instead of specifying it as a constant?

Do you mean whether there is a better way of setting `CUDA_SM_PER_BLOCK = 49152` or a better way of calculating `_sm_size` in the `forward` function?

czkkkkkk:
@dgl-bot

python/dgl/contrib/cugraph/nn/conv/relgraphconv.py
czkkkkkk: Yep. I mean how to set `CUDA_SM_PER_BLOCK` because different GPUs should have different sizes.

tingyu66: You are right. GPUs with different compute capabilities may have a max amount of shared memory per block ranging from 48KB to >200KB, as [Table 15](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#features-and-technical-specifications) shows. 
However, shared memory allocations over 48 KB per block are architecture-specfic and requires extra configurations to allocate. `cugraph-ops` as of now only uses up to 48 KB, so the constant here is valid, but as you said, not elegant. I can discuss with other devs to see how we could get device property and perform the check at the C layer with minimum overhead, so the Python wrapper looks less clunky. This improvement can happen in a separate PR.

python/dgl/contrib/cugraph/nn/conv/relgraphconv.py
```diff
@@ -0,0 +1,386 @@
...
+            agg_fwd_func = agg_hg_basis_post_fwd_int64
+        else:
+            raise TypeError(
+                f"Supported ID type: torch.int32 or torch.int64, "
+                f"but got {g.idtype}"
+            )
+        ctx.graph_idtype = g.idtype
+
+        _in_feat = feat.shape[-1]
+        indptr, indices, edge_ids = g.adj_sparse("csc")
```
chang-l, line 86: I have a quick question here.  
Does cugraph-ops have to rely on homogenous graph adj matrix `g.adj_sparse` to generate mfg?

tingyu66: Yes, cugraph-ops does require a CSR representation of the graph. That is why we need the conversion here.

chang-l: So, does it mean we have to convert to a homogeneous graph first like here, https://github.com/dmlc/dgl/blob/0aa8310a4db072daba51187bbc6d1d725467ba35/examples/pytorch/rgcn/entity_sample.py#L99 to be able to adapt cugraph-ops aggregator? Could it possible to support multiple CSR representations for hetero-graph-RGCN?

chang-l: > You mean 1 CSR adjacency matrix per edge type, right?

Yes. 
> No, our aggregator interface does not support that.

I see. Thanks. Can it support COO format?

tingyu66: No, not natively. Otherwise, we won't need the explicit conversion from COO to CSR in the wrapper.

mufeili:
I believe we don't want to add new stuff to `dgl.contrib`, any suggestions on the directory to use? @jermainewang

mufeili:
I will recommend putting the example in a folder like `examples/pytorch/cugraph`.

tingyu66:
> I believe we don't want to add new stuff to dgl.contrib, any suggestions on the directory to use? @jermainewang

> I will recommend putting the example in a folder like `examples/pytorch/cugraph`.

I put it there because the example highly resembles the current `entity_sample.py` in the same folder. The other PR #4743  from cugraph also follows the same pattern (model/utility goes to `contrib` and examples reside in existing directories). Just want to make sure we are aligned before making any changes.

mufeili:
> > I believe we don't want to add new stuff to dgl.contrib, any suggestions on the directory to use? @jermainewang
> 
> > I will recommend putting the example in a folder like `examples/pytorch/cugraph`.
> 
> I put it there because the example highly resembles the current `entity_sample.py` in the same folder. The other PR #4743 from cugraph also follows the same pattern (model/utility goes to `contrib` and examples reside in existing directories). Just want to make sure we are aligned before making any changes.

Thank you for sharing the information. I've talked to Triston about this and I think we want to discuss the practice during the regular meeting.

mufeili:
As we discussed earlier during the meeting, please update accordingly. I also suggest splitting this PR into two, one for NN module and one for example. It is generally easier to review smaller PRs.

tingyu66:
> The achieved accuracy is different from DGL. Why is that?

Just saw this comment. I think it is because `num_bases` is set to `num_rels` in the current DGL example.
https://github.com/dmlc/dgl/blob/ed8e9c4440c204d19a138bcc552ee476c3a7cbc1/examples/pytorch/rgcn/entity_sample.py#L17

Setting it to 20 should yield similar accuracy.

tingyu66:
> 

Thanks for the reminder. I will break it into two PRs.

tingyu66:
> As we discussed earlier during the meeting, please update accordingly. I also suggest splitting this PR into two, one for NN module and one for example. It is generally easier to review smaller PRs.

Done. This PR is superseded by #4826 and #4827.

Issue #4235, opened 2022-07-09
DistDGL doesn't allow running inference inside the training loop.


zheng-da:
## 🐛 Bug

If we run distributed inference inside the training loop with num_sampler > 0, we will get the following error.
```
Client[6] in group[0] is exiting...
0it [00:01, ?it/s]
Traceback (most recent call last):
  File "m5gnn_lp_huggingface.py", line 26, in <module>
    main(args)
  File "m5gnn_lp_huggingface.py", line 15, in main
    trainer.fit()
  File "/m5-gnn/python/m5gnn/trainer/lp_trainer.py", line 130, in fit
    lp_model.fit(dataloader)
  File "/m5-gnn/python/m5gnn/model/rgnn_lp_base.py", line 211, in fit
    self.save_model_embed(epoch, i, g, bert_emb_cache)
  File "/m5-gnn/python/m5gnn/model/rgnn.py", line 774, in save_model_embed
    embeddings = self.compute_embeddings(g, device, bert_emb_cache, feat_field=self._feat_name)
  File "/m5-gnn/python/m5gnn/model/rgnn.py", line 573, in compute_embeddings
    embeddings = self.compute_gnn_embeddings(g, device, bert_emb_cache, target_nidx, feat_field)
  File "/m5-gnn/python/m5gnn/model/rgnn.py", line 630, in compute_gnn_embeddings
    embeddings = do_fullgraph_infer(g=g,
  File "/m5-gnn/python/m5gnn/model/utils.py", line 858, in do_fullgraph_infer
    embeddings = model.module.dist_inference(g, eval_batch_size, device, 0, node_embed, client=client, mlflow_report_frequency=mlflow_report_frequency)
  File "/m5-gnn/python/m5gnn/model/rgnn_base.py", line 86, in dist_inference
    for iter_l, (input_nodes, output_nodes, blocks) in enumerate(tqdm.tqdm(dataloader)):
  File "/usr/local/lib/python3.8/dist-packages/tqdm/std.py", line 1195, in __iter__
    for obj in iterable:
  File "/usr/local/lib/python3.8/dist-packages/dgl/distributed/dist_dataloader.py", line 110, in __next__
    result = self._get_data_from_result_queue()
  File "/usr/local/lib/python3.8/dist-packages/dgl/distributed/dist_dataloader.py", line 122, in _get_data_from_result_queue
    ret = self.pool.get_result(self.name, timeout=timeout)
[...]

PR #4704, opened 2022-10-12
Multiple dist dl sampler

Rhett-Ying:
## Description

This PR aims to enable the `custom pool` could be shared/used by any number of `DistDataLoaders`. This feature is required when try to inference during train. see more details in https://github.com/dmlc/dgl/issues/4235.

## Checklist
- [ ] Code is well-documented


tests/distributed/test_mp_dataloader.py
```diff
@@ -1,19 +1,24 @@
-import dgl
```
Rhett-Ying, line 1: I refined/formatted this file, pls pay attention to the newly added `test_multiple_dist_dataloaders` which is at the end of file. this should be fine.

python/dgl/distributed/dist_context.py
```diff
@@ -122,8 +123,11 @@ def __init__(self, num_workers, rpc_config):
         """
         ctx = mp.get_context("spawn")
         self.num_workers = num_workers
-        self.queue_size = num_workers * 4
+        # As pool could be used by any number of dataloaders, queues
+        # should be able to take infinite elements to avoid dead lock.
+        self.queue_size = 0
         self.result_queue = ctx.Queue(self.queue_size)
+        self.results = {}
```
jermainewang, line 130: Add some comments about the key and values

python/dgl/distributed/dist_context.py
```diff
@@ -167,16 +172,20 @@ def submit_task_to_all_workers(self, func, args):
 
     def get_result(self, dataloader_name, timeout=1800):
         """Get result from result queue"""
-        result_dataloader_name, result = self.result_queue.get(timeout=timeout)
-        assert result_dataloader_name == dataloader_name
-        return result
+        if dataloader_name not in self.results:
+            raise DGLError(f"{dataloader_name} not found.")
```
jermainewang, line 176:
```suggestion
            raise DGLError(f"Got result from an unknown dataloader {dataloader_name}.")
```

python/dgl/distributed/dist_context.py
```diff
@@ -167,16 +172,20 @@ def submit_task_to_all_workers(self, func, args):
...
+        while len(self.results[dataloader_name]) == 0:
+            dl_name, data = self.result_queue.get(timeout=timeout)
```
jermainewang, line 178: What will happen if `self.result_queue.get` timed out? Are we able to catch that?

jermainewang (approved):
Generally LGTM. Since the original task is to enable model inference during training, is it possible to provide an example?

python/dgl/distributed/dist_context.py
Rhett-Ying: raises the [queue.Empty](https://docs.python.org/3/library/queue.html#queue.Empty) exception if no item was available within that time.
As this should never happen in normal runnings, we don't catch such exception and just let it abort the app. I think this behavior is acceptable. If it happens, something went wrong in the sampler processes.

python/dgl/distributed/dist_context.py
Rhett-Ying: added.

python/dgl/distributed/dist_context.py
Rhett-Ying: accepted.

Rhett-Ying:
> Generally LGTM. Since the original task is to enable model inference during training, is it possible to provide an example?

such example is more about end2end test instead of unit test I added here. For now, we don't have such end2end test framework or environment for the moment. But I believe the unit tests I added simulate the key part of end2end scenario. `iterating another dataloader during iterating on one dataloader` is supposed to mimic `do inference during train`, I think.

jermainewang:
> > Generally LGTM. Since the original task is to enable model inference during training, is it possible to provide an example?
> 
> such example is more about end2end test instead of unit test I added here. For now, we don't have such end2end test framework or environment for the moment. But I believe the unit tests I added simulate the key part of end2end scenario. `iterating another dataloader during iterating on one dataloader` is supposed to mimic `do inference during train`, I think.

Sure. Could you create a task about the end2end test for tracking purposes?

Issue #4681, opened 2022-10-08
[DistPartPipeline] canonical etypes are used in `edge_feat.dgl` instead of etypes


Rhett-Ying:
## 🐛 Bug

anyone has verified the partition outputs with edge feature works for distributed train?

`'paper:cites:paper/count'` is used as key of `edge_feat.dgl` while `dgl.distributed.load_partition()` supports etype only for now.

What's worse, the test code below does not take effect at all. `key` is not defined correctly. neither `part_data`.
https://github.com/dmlc/dgl/blob/eae6ce2a4f44e81eda3b876ef50fc045926912f6/tests/tools/test_dist_part.py#L271-L281

This issue is introduced by https://github.com/dmlc/dgl/pull/4623

## To Reproduce

`python3 -m pytest tests/tools/test_dist_part.py::test_part_pipeline`

## Environment

 - DGL Version (e.g., 1.0): master
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):
 - OS (e.g., Linux):
 - How you installed DGL (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version (if applicable):
 - GPU models and configuration (e.g. V100):
 - Any other relevant information:

Issue #4658, opened 2022-09-29
[Dist] add support for edge feature on partition many chunks into less parts in pipeline


Rhett-Ying:
## 🔨Work Item

**IMPORTANT:**
* This template is only for dev team to track project progress. For feature request or bug report, please use the corresponding issue templates.
* DO NOT create a new work item if the purpose is to fix an existing issue or feature request. We will directly use the issue in the project tracker.

Project tracker: https://github.com/orgs/dmlc/projects/2

## Description
For now, pipeline does not support edge feature. Once this is supported, we need to apply what we did in https://github.com/dmlc/dgl/issues/4362 onto edge features.

## Depending work items or issues
https://github.com/dmlc/dgl/issues/4361

PR #4729, opened 2022-10-18
[Dist] decouple num_chunks and num_parts for graphs with edge feature

Rhett-Ying:
## Description

This PR includes fix for https://github.com/dmlc/dgl/issues/4658 and https://github.com/dmlc/dgl/issues/4681. 
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/tests/tools/create_chunked_dataset.py
+++ b/tests/tools/create_chunked_dataset.py
@@ -15,7 +15,7 @@ from dgl.data.utils import load_graphs, load_tensors
 
 
 def create_chunked_dataset(
-    root_dir, num_chunks, include_masks=False, include_edge_data=False
+    root_dir, num_chunks, include_masks=False
 ):
     """
     This function creates a sample dataset, based on MAG240 dataset.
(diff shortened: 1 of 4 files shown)


tests/tools/create_chunked_dataset.py
```diff
@@ -15,7 +15,7 @@
 
 
 def create_chunked_dataset(
-    root_dir, num_chunks, include_masks=False, include_edge_data=False
+    root_dir, num_chunks, include_masks=False
```
jermainewang, line 18: Maybe keep this option so you can test two cases: w/ and w/o edge data

jermainewang (approved):
LGTM

tests/tools/create_chunked_dataset.py
Rhett-Ying: this is a temporary argument that was added on purpose to bypass possible test failure caused by `decouple of num_chunks and num_parts` not supported for edge feature yet. It's ok to remove it.

jermainewang: What I mean is to test two cases (w/ and w/o edge data) separately to ensure that there is no bug when the data does not have any edge data. You can ignore me if you've already tested it.

Rhett-Ying: I have tested it...

PR #4730, opened 2022-10-19
[BugFix] cugraph must be imported before torch

Rhett-Ying:
## Description

fix the bug imported by sorting module import: #4694 

as mentioned at the top of the file, cugraph must be imported before torch due to https://github.com/rapidsai/cugraph/issues/2718

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/tests/cugraph/test_basics.py
+++ b/tests/cugraph/test_basics.py
@@ -1,14 +1,9 @@
 # NOTE(vibwu): Currently cugraph must be imported before torch to avoid a resource cleanup issue.
 #    See https://github.com/rapidsai/cugraph/issues/2718
-import unittest
-
-import backend as F
 import cugraph
-import numpy as np
-import pytest
+import backend as F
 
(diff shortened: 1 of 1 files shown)


tests/cugraph/test_basics.py
```diff
@@ -1,14 +1,9 @@
 # NOTE(vibwu): Currently cugraph must be imported before torch to avoid a resource cleanup issue.
 #    See https://github.com/rapidsai/cugraph/issues/2718
-import unittest
-
-import backend as F
 import cugraph
-import numpy as np
-import pytest
+import backend as F
```
frozenbugs, line 4: can you add a todo here to explain why we need to import cugraph first, and link to the ticket to resolve it?

Rhett-Ying: at the top of this file, such note has already been there. so we don't need to add more.

Issue #4514, opened 2022-09-05
Reduce the startup overhead in DistDGL


zheng-da:
## 🔨Work Item

For a graph with 4B nodes and 30B edges, if we load the graph with 10 partitions on 10 machines, it takes more than one hour to load the graph and start distributed training. It's very painful to debug on such a large graph. We should reduce the startup overhead and reduce the startup time to around 10 minutes.

**IMPORTANT:**
* This template is only for dev team to track project progress. For feature request or bug report, please use the corresponding issue templates.
* DO NOT create a new work item if the purpose is to fix an existing issue or feature request. We will directly use the issue in the project tracker.

Project tracker: https://github.com/orgs/dmlc/projects/2

## Description

Currently, DistDGL takes a long time to load data and initialize the data for distributed training. There are multiple places we can improve to reduce the overhead:
- [x] avoid checking the loaded data (https://github.com/dmlc/dgl/blob/master/python/dgl/distributed/partition.py#L100-L119). checking the loaded graph structure is both time consuming and memory consuming. we should disable by default and only enable it for the debug mode.
- [ ] avoid converting the format (https://github.com/dmlc/dgl/blob/master/python/dgl/distributed/dist_graph.py#L351-L352). we should prepare the right format during the graph partitioning.
- [ ] avoid sorting (https://github.com/dmlc/dgl/blob/master/python/dgl/distributed/dist_graph.py#L354-L360). we should prepare the right format during the graph partitioning.

jermainewang:
The task needs some breakdown. @Rhett-Ying could you pls have a look and think of how many PRs will this require? I remember you've already started on some of them.

Rhett-Ying:
- [ ] avoid checking loaded data. 1 PR for enable/disable it. 1 PR for improving checking logic to be more memory efficient. Total: 2 PRs.
- [ ] avoid converting graph format. This depends on this open PR:  [Feature] enable to save graphs in multiple formats #[4266](https://github.com/dmlc/dgl/pull/4266) . Once it's ready, we could improve `dgl.distributed.partition_graph()` and the new distributed partition pipeline. Total: 3 PRs.
- [ ] avoid sorting. Total: 1 PR.

In total, 6 PRs are required and most of them are in small size. `[Feature] enable to save graphs in multiple formats` is probably the most time-consuming one.

nv-dlasalle:
I take it this is meant to close #4474?

Rhett-Ying:
[...]

PR #4735, opened 2022-10-20
[Dist] Reduce startup overhead: sort etypes and save in specified formats

Rhett-Ying:
## Description

Below 2 action items that are requested in https://github.com/dmlc/dgl/issues/4514 are implemented in this PR:
- [x] avoid converting the format (https://github.com/dmlc/dgl/blob/master/python/dgl/distributed/dist_graph.py#L351-L352). we should prepare the right format during the graph partitioning.
- [x] avoid sorting (https://github.com/dmlc/dgl/blob/master/python/dgl/distributed/dist_graph.py#L354-L360). we should prepare the right format during the graph partitioning.
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/dist_graph.py
+++ b/python/dgl/distributed/dist_graph.py
@@ -342,9 +342,6 @@ class DistGraphServer(KVServer):
                 if k in self.client_g.edata:
                     self.client_g.edata[k] = F.astype(
                         self.client_g.edata[k], dtype)
-            # Create the graph formats specified the users.
-            self.client_g = self.client_g.formats(graph_format)
-            self.client_g.create_formats_()
             # Sort underlying matrix beforehand to avoid runtime overhead during sampling.
             if len(etypes) > 1:
                 if 'csr' in graph_format:
@@ -353,6 +350,9 @@ class DistGraphServer(KVServer):
                 if 'csc' in graph_format:
                     self.client_g = sort_csc_by_tag(
                         self.client_g, tag=self.client_g.edata[ETYPE], tag_type='edge')
+            # Create the graph formats specified the users.
+            self.client_g = self.client_g.formats(graph_format)
+            self.client_g.create_formats_()
             if not disable_shared_mem:
                 self.client_g = _copy_graph_to_shared_mem(self.client_g, graph_name, graph_format)
 
--- a/python/dgl/distributed/dist_graph.py
+++ b/python/dgl/distributed/dist_graph.py
@@ -10,7 +10,7 @@ import numpy as np
 from ..heterograph import DGLHeteroGraph
 from ..convert import heterograph as dgl_heterograph
 from ..convert import graph as dgl_graph
-from ..transforms import compact_graphs, sort_csr_by_tag, sort_csc_by_tag
+from ..transforms import compact_graphs
 from .. import heterograph_index
 from .. import backend as F
 from ..base import NID, EID, ETYPE, ALL, is_all
@@ -342,14 +342,6 @@ class DistGraphServer(KVServer):
                 if k in self.client_g.edata:
                     self.client_g.edata[k] = F.astype(
                         self.client_g.edata[k], dtype)
-            # Sort underlying matrix beforehand to avoid runtime overhead during sampling.
-            if len(etypes) > 1:
-                if 'csr' in graph_format:
-                    self.client_g = sort_csr_by_tag(
-                        self.client_g, tag=self.client_g.edata[ETYPE], tag_type='edge')
-                if 'csc' in graph_format:
-                    self.client_g = sort_csc_by_tag(
-                        self.client_g, tag=self.client_g.edata[ETYPE], tag_type='edge')
             # Create the graph formats specified the users.
             self.client_g = self.client_g.formats(graph_format)
             self.client_g.create_formats_()
--- a/python/dgl/distributed/partition.py
+++ b/python/dgl/distributed/partition.py
@@ -589,6 +589,9 @@ def partition_graph(g, graph_name, num_parts, out_path, num_hops=1, part_method=
     ...     g, node_feats, edge_feats, gpb, graph_name, ntypes_list, etypes_list,
     ... ) = dgl.distributed.load_partition('output/test.json', 0)
     '''
+    # 'coo' is required for partition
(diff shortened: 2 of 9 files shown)


python/dgl/distributed/dist_graph.py
```diff
@@ -345,14 +345,6 @@ def __init__(self, server_id, ip_config, num_servers,
             # Create the graph formats specified the users.
             self.client_g = self.client_g.formats(graph_format)
             self.client_g.create_formats_()
-            # Sort underlying matrix beforehand to avoid runtime overhead during sampling.
```
jermainewang, line 348: BTW, removing this code may cause some breaking changes to users who are using the old partitioned graph. They will get errors like the graph is not sorted. I guess this is not avoidable but we need to let @zheng-da know about the changes.

python/dgl/distributed/partition.py
```diff
@@ -23,16 +24,25 @@
...
     '''
     for g in g_list:
         for k, dtype in RESERVED_FIELD_DTYPE.items():
             if k in g.ndata:
                 g.ndata[k] = F.astype(g.ndata[k], dtype)
             if k in g.edata:
                 g.edata[k] = F.astype(g.edata[k], dtype)
-    save_graphs(filename , g_list)
+    for g in g_list:
+        if (not sort_etypes) or (formats is None):
```
jermainewang, line 39: Would `formats=None` but graph has csr or csc so that the formats are saved without sorting?

frozenbugs:
How much overhead does this PR reduce? Give some evidence about the graph you tested?

python/dgl/distributed/dist_graph.py
Rhett-Ying: yes. you're right.

tests/tools/test_dist_part.py
```diff
@@ -26,6 +26,15 @@ def _verify_partition_data_types(part_g):
         if k in part_g.edata:
             assert part_g.edata[k].dtype == dtype
 
+def _verify_partition_formats(part_g, formats):
+    # verify saved graph formats
```
frozenbugs, line 30: nit: Verify saved graph formats.

python/dgl/distributed/partition.py
Rhett-Ying: if `formats=None`, only `COO` is saved. The case you mentioned should never happen I think.

tests/tools/test_dist_part.py
```diff
@@ -26,6 +26,15 @@ def _verify_partition_data_types(part_g):
...
+    if formats is None:
+        assert "coo" in part_g.formats()["created"]
```
frozenbugs, line 32: Where do we introduce the magic word "created"?

python/dgl/distributed/partition.py
jermainewang: Is it possible that the pass-in graph only has csr/coo formats (e.g., using `g.format` API). When the `formats` argument is None, the default behavior is to select any existing format which can be csr/coo in this case.

tests/distributed/test_partition.py
```diff
@@ -20,48 +24,67 @@ def _verify_partition_data_types(part_g):
...
+
+
 def _get_inner_node_mask(graph, ntype_id):
     if dgl.NTYPE in graph.ndata:
-        dtype = F.dtype(graph.ndata['inner_node'])
-        return graph.ndata['inner_node'] * F.astype(graph.ndata[dgl.NTYPE] == ntype_id, dtype) == 1
+        dtype = F.dtype(graph.ndata["inner_node"])
+        return (
+            graph.ndata["inner_node"]
+            * F.astype(graph.ndata[dgl.NTYPE] == ntype_id, dtype)
```
frozenbugs, line 42: Create a local variable for F.astype(graph.ndata[dgl.NTYPE] == ntype_id, dtype) to make the block more readable?

tests/distributed/test_partition.py
```diff
@@ -1,17 +1,21 @@
-import dgl
-import sys
 import os
```
frozenbugs, line 1: Let's avoid autoformat on existing code together with other code change, very hard to review :(

python/dgl/distributed/partition.py
Rhett-Ying: If the input graph has only `CSC/CSR`, partition cannot be done. `COO` is required. This is because `g.find_edges(eids)` is called, I think. And a graph with `coo` not allowed is quite rare, I think.

tests/distributed/test_partition.py
Rhett-Ying: The reason why I formatted here is I changed much code, some unnecessary code/tests are removed. So just to be clean, I apply formatting as well.

tests/tools/test_dist_part.py
Rhett-Ying: `DGLHeteroGraph.formats()`

python/dgl/distributed/partition.py
jermainewang: Then could you add a check and a proper error message to forbid that?

tests/tools/test_dist_part.py
frozenbugs: In the future we should avoid this type of magic word, instead, either using structured variable, or using global constant.

python/dgl/distributed/partition.py
Rhett-Ying: added.

tests/distributed/test_partition.py
Rhett-Ying: Use same ones from `dgl.distributed.partition` instead

Rhett-Ying:
> How much overhead does this PR reduce? Give some evidence about the graph you tested?

As shown in https://github.com/dmlc/dgl/issues/4514, for fake partition(560M nodes, 4.2B edges), r6i.metal, EBS, below parts are gone.
Function 'format_graph' executed in 61.2462s
Function 'sort_edges' executed in 49.5095s

A thorough comparison will be launched once all action items are completed. We need another PR of canonical etype be enabled entirely to fully meet the target: parition graph with `csc` saved only and load with `graph_formats=('csc')` in DistDGL.

jermainewang (approved):

PR #2747, opened 2021-03-13
[performance] Optimize the association order of AXW in GraphSAGE.

yzh119:
## Description
Use A(XW) instead of (AX)W when output feature dimensionality is greater than input feature dimensionality. This optimization was used in our GraphConv module but not in SAGEConv module: https://discuss.dgl.ai/t/gpu-kernels-while-running-graphsage-model/1449/4 .

@mufeili , @BarclayII please verify if it's equivalent to the original implementation.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage
- [x] Code is well-documented
- [x] To the my best knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [x] Related issue is referred in this PR
- [x] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

PR #4747, opened 2022-10-24
[Optimization] Optimize bias term in SageConv layer

bgawrych:
## Description
This change simply optimize adding bias term in SageConv Layer - adding bias term can be done within FullyConnected for all layers, but GCN
Change is backward compatible (same as https://github.com/dmlc/dgl/pull/2747/) however, saving/loading model parameters with state_dict was not and still is not backward compatible. Do we need have to keep this backward compatibility check as major release is coming?

Running script from https://github.com/bgawrych/dgl/tree/master/examples/pytorch/ogb/ogbn-products/graphsage on  Intel Xeon(R) Platinum 8380 gives following results:
With change (avg. 3327.742068 samples /s): 
image

Current master (avg. 3264.397611)
image

I believe for GPU this change is also beneficial as it reduces read and write operations

## Checklist


bgawrych:
@BarclayII Hi, can you review this or point someone who can review this?

BarclayII:
@dgl-bot

BarclayII:
I guess the only relevant change is in L275-277?  Is other changes relevant?
Also, setting `bias` attribute for `self.fc_self` even when `bias=False` looks strange to me.

bgawrych:
L119-L136 was modified to be compatible with written tests, where bias is accessed by 'bias' field. If _compatibility_check can be removed, this change should be clearer (as I wrote earlier it doesn't work with state_dict anyway).

BarclayII:
@dgl-bot

BarclayII:
> L119-L136 was modified to be compatible with written tests, where bias is accessed by 'bias' field. If _compatibility_check can be removed, this change should be clearer (as I wrote earlier it doesn't work with state_dict anyway).

I saw that L119-L136 is mainly moving around the code with the exception of setting `self.fc_self.bias`.  Did you set the bias solely because of one test is failing?  If so, which test are you referring to?  The problem might be how we write the test instead.

bgawrych:
I am referring to this test: https://github.com/dmlc/dgl/blob/8ae50c422b81de5734b971722d580c433d669fe8/tests/pytorch/test_nn.py#L949
and I thought it was good idea to access the bias in the same way independently of aggregation type - that's why `self.fc_self.bias = self.bias`

BarclayII:
@dgl-bot

bgawrych:
Some results from `asv run -n -e --python=same --verbose --bench model_speed.bench_sage_ns` command:

image

I've set iteration count to 140 to get more stable results
default config refers to configuration set as default in bench_sage_ns.py script, example config means sage model was configured like in script from examples (layers: 3, hidden_dim: 256, fanout: 5, 10, 15)

bgawrych:
@BarclayII Ok, I will remove this and fix the test
I tested also bigger model (512 hidden dimension and 10,15,20 fanout):
image
so I assume that 20% is not possible with this change.

I've compared also line by line performance and here are the results (left side my change and right side original):
image

itaraban (approved):

python/dgl/nn/pytorch/conv/sageconv.py
```diff
@@ -266,12 +254,12 @@ def forward(self, graph, feat, edge_weight=None):
...
                 rst = h_neigh
+                # add bias manually for GCN
+                if self.bias is not None:
+                    rst = rst + self.bias
             else:
                 rst = self.fc_self(h_self) + h_neigh
 
-            # bias term
-            if self.bias is not None:
-                rst = rst + self.bias
```
BarclayII, line 274: We are only adding bias when the aggregation type is GCN?  This doesn't look right.

bgawrych: No - for GCN fc_self linear is not performed, but BIAS is added anyway - for other aggregation types fc_self is performed in line 261 and BIAS is added within this Linear layer
`rst = self.fc_self(h_self) + h_neigh` ==> `(h_self * WEIGHTS + BIAS) + h_neigh`

BarclayII (approved):

BarclayII:
@dgl-bot

BarclayII:
@dgl-bot

PR #4748, opened 2022-10-24
[Sampling] Enable sampling with edge masks on homogeneous graph

BarclayII:
## Description
This is a subset of the PR #4633 .

Also fixes a bug where the zero probability edges are still sampled when fanout is -1.

## Checklist
- [x] Code is well-documented


tests/compute/test_sort.py
```diff
@@ -48,7 +48,7 @@ def check_sort(spm, tag_arr=None, tag_pos=None):
                 # tag value is equal to `tag_pos_ptr`
                 return False
             if tag_arr[dst[j]] > tag_arr[dst[j + 1]]:
-                # The tag should be in descending order after sorting
+                # The tag should be in ascending order after sorting
```
mufeili, line 51: Is this a breaking change or a fix for the wrong comment?

tests/compute/test_sampling.py
```diff
@@ -272,6 +272,7 @@ def _gen_neighbor_sampling_test_graph(hypersparse, reverse):
         }, {'user': card if card is not None else 4})
         g = g.to(F.ctx())
         g.edata['prob'] = F.tensor([.5, .5, 0., .5, .5, 0., 1.], dtype=F.float32)
+        g.edata['mask'] = F.tensor([True, True, False, True, True, False, True])
```
mufeili, line 275: Does this mean that users can simultaneously specify a probability distribution for sampling edges as well as a boolean mask for sampling eligibility? This seems a bit convoluted. What's the motivation for supporting such cases?

mufeili:
We might want to refactor `tests/compute/test_sampling.py` in the future. It's currently not very readable.

tests/compute/test_sampling.py
```diff
@@ -1026,12 +1059,15 @@ def test_global_uniform_negative_sampling(dtype):
...
+    test_sample_neighbors_prob()
+    test_sample_neighbors_mask()
     for args in product(['coo', 'csr', 'csc'], ['in', 'out'], [False, True]):
         test_sample_neighbors_etype_homogeneous(*args)
-    test_non_uniform_random_walk()
+    test_non_uniform_random_walk(False)
     test_uniform_random_walk(False)
     test_pack_traces()
-    test_pinsage_sampling()
+    test_pinsage_sampling(False)
```
mufeili, line 1070: Do test functions defined under `if __name__ == '__main__':` still come into effect now? @Rhett-Ying

include/dgl/aten/macro.h
```diff
@@ -171,6 +171,24 @@
...
+  } else if ((val).code == kDGLFloat && (val).bits == 32) {    \
+    typedef float DType;                                      \
+    {__VA_ARGS__}                                             \
+  } else if ((val).code == kDGLFloat && (val).bits == 64) {    \
+    typedef double DType;                                     \
+    {__VA_ARGS__}                                             \
+  } else {                                                    \
+    LOG(FATAL) << (val_name) << " can only be int8, uint8, float32 or float64"; \
+  }                                                           \
+} while (0)
```
mufeili, line 190: Two earlier comments from @jermainewang 

> Please add some docstring about how to use this macro.

> The code doesn't look like related to bool type at all.

tests/compute/test_sort.py
BarclayII: a fix for the wrong comment.

tests/compute/test_sampling.py
BarclayII: It doesn't for pytest, but it does if I run it in standalone `python` for debugging.

include/dgl/aten/macro.h
BarclayII: I renamed it to `ATEN_FLOAT_INT8_UINT8`.  Previously I named it `BOOL` because `int8` and `uint8` arrays store boolean arrays in current PyTorch/TF/etc.

src/array/cuda/rowwise_sampling_prob.cu
```diff
@@ -392,6 +393,74 @@ __global__ void _CSRRowWiseSampleReplaceKernel(
...
+  IdArray new_eid = IdArray::Empty({nnz}, idtype, ctx);
+  IdType* new_row_data = new_row.Ptr<IdType>();
+  IdType* new_col_data = new_col.Ptr<IdType>();
+  IdType* new_eid_data = new_eid.Ptr<IdType>();
+  auto stream = runtime::getCurrentCUDAStream();
+  auto device = runtime::DeviceAPI::Get(ctx);
+
+  int8_t* flags = static_cast<int8_t*>(device->AllocWorkspace(ctx, nnz));
+  int nt = dgl::cuda::FindNumThreads(nnz);
+  int nb = (nnz + nt - 1) / nt;
```
Rhett-Ying, line 430: `nnz` is `int64_t`, not sure whether overflow will happen for `int nb`.

src/array/cuda/rowwise_sampling_prob.cu
```diff
@@ -392,6 +393,74 @@ __global__ void _CSRRowWiseSampleReplaceKernel(
...
+  int tx = blockIdx.x * blockDim.x + threadIdx.x;
+  const int stride_x = gridDim.x * blockDim.x;
+  while (tx < n) {
+    output[tx] = (values[idx ? idx[tx] : tx] != criteria);
+    tx += stride_x;
+  }
+}
+
+template <DGLDeviceType XPU, typename IdType, typename DType, typename MaskGen>
+COOMatrix COOGeneralRemoveIf(COOMatrix coo, MaskGen maskgen) {
```
Rhett-Ying, line 408: can we use reference or const ref for these arguments to avoid potential copy overhead?

src/random/cpu/choice.cc
```diff
@@ -55,10 +55,22 @@ template void RandomEngine::Choice<int32_t, double>(int32_t num,
...
+                                                     FloatArray prob,
+                                                     int32_t* out, bool replace);
+template void RandomEngine::Choice<int64_t, uint8_t>(int64_t num,
+                                                     FloatArray prob,
+                                                     int64_t* out, bool replace);
 
 template <typename IdxType>
 void RandomEngine::UniformChoice(IdxType num, IdxType population, IdxType* out,
                                  bool replace) {
+  CHECK_GE(num, 0);
```
Rhett-Ying, line 72: How about throwing error message?

src/random/cpu/sample_utils.h
```diff
@@ -32,6 +32,10 @@ class BaseSampler {
   }
 };
 
+// (BarclayII 2022.9.20) Changing the internal data type of probabilities to double since
+// we are using non-uniform sampling to sample on boolean masks, where False represents
```
Rhett-Ying, line 36: what's `DType` if not changed to `double`? could you explain more why change to double in order to handle boolean mask?

Rhett-Ying: if not fixed as `double`, `DType` could be `uint8` and this will cause type mismatch?

tests/compute/test_sampling.py
Rhett-Ying: I usually run specific testcase via `-m pytest tests/compute/test_sampling.py::test_sample_xx`. why do we need to run it as standalone python script?

src/array/array.cc
```diff
@@ -555,7 +555,9 @@ COOMatrix CSRRowWiseSampling(
     // prob is pinned and rows on GPU is valid
     CHECK_VALID_CONTEXT(prob, rows);
     ATEN_CSR_SWITCH_CUDA_UVA(mat, rows, XPU, IdType, "CSRRowWiseSampling", {
-      ATEN_FLOAT_TYPE_SWITCH(prob->dtype, FloatType, "probability", {
+      CHECK(!(prob->dtype.bits == 8 && XPU == kDGLCUDA)) <<
+        "GPU sampling with masks is currently not supported yet.";
```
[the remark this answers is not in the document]
BarclayII: Yes.  Either a bool array or a float array.

src/array/cpu/rowwise_sampling.cc
```diff
@@ -27,17 +27,39 @@ inline FloatArray DoubleSlice(FloatArray array, const IdxType* idx_data,
   return ret;
 }
 
+template <typename IdxType, typename FloatType>
```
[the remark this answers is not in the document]
BarclayII: Good idea.

src/array/cpu/rowwise_topk.cc
```diff
@@ -45,7 +56,7 @@ inline PickFn<IdxType> GetTopkPickFn(int64_t k, NDArray weight, bool ascending)
       std::vector<IdxType> idx(len);
       std::iota(idx.begin(), idx.end(), off);
       std::sort(idx.begin(), idx.end(), compare_fn);
-      for (int64_t j = 0; j < k; ++j) {
+      for (int64_t j = 0; j < num_picks; ++j) {
```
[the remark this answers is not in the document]
BarclayII: Make sense.

src/array/cuda/rowwise_sampling.cu
```diff
@@ -369,6 +367,19 @@ COOMatrix CSRRowWiseSamplingUniform(CSRMatrix mat,
...
 
+template <DGLDeviceType XPU, typename IdType>
+COOMatrix CSRRowWiseSamplingUniform(
+    CSRMatrix mat, IdArray rows, const int64_t num_picks, const bool replace) {
+  if (num_picks == -1) {
+    // Basically this is UnitGraph::InEdges().
+    COOMatrix coo = CSRToCOO(CSRSliceRows(mat, rows), false);
+    IdArray sliced_rows = IndexSelect(rows, coo.row);
+    return COOMatrix(mat.num_rows, mat.num_cols, sliced_rows, coo.col, coo.data);
+  } else {
```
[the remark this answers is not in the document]
BarclayII: Which lint format are you referring to?  This seems to pass the lint test.

src/array/cuda/rowwise_sampling_prob.cu
BarclayII: This is a very good point.  We should scan the C++ code and use const references whenever possible.  I'll do this only for the new functions in this PR for now.
@frozenbugs for awareness.

src/random/cpu/sample_utils.h
BarclayII: Yes.  Let me elaborate in the comments.

tests/compute/test_sampling.py
BarclayII: Ah yeah you are right.  `__main__` is no longer needed then.  We should document this though.

src/array/cuda/rowwise_sampling_prob.cu
```diff
@@ -647,8 +717,24 @@ COOMatrix CSRRowWiseSampling(CSRMatrix mat,
   picked_col = picked_col.CreateView({new_len}, picked_col->dtype);
   picked_idx = picked_idx.CreateView({new_len}, picked_idx->dtype);
 
-  return COOMatrix(mat.num_rows, mat.num_cols, picked_row,
-      picked_col, picked_idx);
+  return COOMatrix(mat.num_rows, mat.num_cols, picked_row, picked_col, picked_idx);
+}
+
+template <DGLDeviceType XPU, typename IdType, typename FloatType>
```
Rhett-Ying, line 723: `FloatType` to `DType`?

src/array/array.cc
```diff
@@ -810,7 +812,7 @@ COOMatrix COORowWiseSampling(
     if (IsNullArray(prob)) {
       ret = impl::COORowWiseSamplingUniform<XPU, IdType>(mat, rows, num_samples, replace);
     } else {
-      ATEN_FLOAT_TYPE_SWITCH(prob->dtype, FloatType, "probability", {
+      ATEN_FLOAT_INT8_UINT8_TYPE_SWITCH(prob->dtype, FloatType, "probability", {
```
Rhett-Ying, line 815: do we need to change the `val_name`: `probability` to indicate `prob=prob or mask`?

src/array/cpu/rowwise_sampling.cc
```diff
@@ -27,17 +27,41 @@ inline FloatArray DoubleSlice(FloatArray array, const IdxType* idx_data,
...
+template <typename IdxType, typename DType>
+inline NumPicksFn<IdxType> GetSamplingNumPicksFn(
+    int64_t num_samples, FloatArray prob, bool replace) {
+  NumPicksFn<IdxType> num_picks_fn = [prob, num_samples, replace]
+    (IdxType rowid, IdxType off, IdxType len,
+     const IdxType* col, const IdxType* data) {
+      const int64_t max_num_picks = (num_samples == -1) ? len : num_samples;
+      const DType* prob_data = prob.Ptr<DType>();
+      IdxType nnz = 0;
+      for (IdxType i = off; i < off + len; ++i) {
```
Rhett-Ying, line 39: just curious, does  `i` of `IdxType` handle correctly if `prob.length() > maximum IdxType`? For example, int32_t for row/col, but num_edges has to be int64_t.
should we always use `int64_t` for indexing?

tests/compute/test_sampling.py
Rhett-Ying, line 275: `mask` should be in `int8_t` or `uint8_t. And if `F.tensor([0, 1, 0, 1, 0]) is passed in, program will crash in `ATEN_FLOAT_INT8_UINT8_TYPE_SWITCH`?
Though this is rare case(maybe mask data is loaded from numpy data file and forget to format it), is it possible to throw exception in python or always format to supported dtypes internally?

Rhett-Ying (approved):

src/array/cpu/rowwise_sampling.cc
BarclayII: Good question.  For now I think if `IdxType` is `int32`, we assume that the number of edges is also less than `2 ** 31`, because otherwise the edge IDs in CSR will overflow.  That being said, I don't know if we perform any check on this.

tests/compute/test_sampling.py
BarclayII: It will report an error if the data type is wrong.

I don't think throwing exception in Python (or correcting the format internally) is a good idea since it introduces extra overhead for rare cases.  A clear error message should be enough.  What is your concern?

Rhett-Ying: if a clear error message is given, that's enough. what does the error message look like? can user figure out that dtype of `g.edata['mask']` is incorrrect?

jermainewang (approved):

Issue #4495, opened 2022-08-31
Transform Module DropNode and DropEdge affect original graphs


mufeili:
## 🐛 Bug

Every time a `DropNode` or `DropEdge` object is called, we expect the object to return a new graph and the original graph should be kept unchanged. This is however not the case now.

## To Reproduce

```
import dgl
import dgl.transforms as T

g = dgl.rand_graph(30, 100)
num_nodes_old = g.num_nodes()
drop_node = T.DropNode()
new_g = drop_node(g)
num_nodes_new = g.num_nodes()
assert num_nodes_old == num_nodes_new
```

```
import dgl
import dgl.transforms as T

g = dgl.rand_graph(30, 100)
num_edges_old = g.num_edges()
drop_edge = T.DropEdge()
new_g = drop_edge(g)
num_edges_new = g.num_edges()
assert num_edges_old == num_edges_new
```

## Expected behavior

There should not be an assertion error.

## Environment

 - DGL Version (e.g., 1.0): 0.9
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):
 - OS (e.g., Linux):
 - How you installed DGL (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version (if applicable):
 - GPU models and configuration (e.g. V100):
 - Any other relevant information:

mufeili:
Issue priority: high

PR #4753, opened 2022-10-25
[Transform] Fix graph structure corruption with transform

mufeili:
## Description
As pointed out in #4495 , a graph transform should not corrupt the original graph structure.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/transforms/module.py
+++ b/python/dgl/transforms/module.py
@@ -1441,11 +1441,12 @@ class DropNode(BaseTransform):
         self.dist = Bernoulli(p)
 
     def __call__(self, g):
+        g = g.clone()
+
         # Fast path
         if self.p == 0:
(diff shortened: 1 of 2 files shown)


python/dgl/transforms/module.py
```diff
@@ -1397,6 +1397,7 @@ class NodeShuffle(BaseTransform):
             [ 7.,  8.]])
     """
     def __call__(self, g):
+        g = g.clone()
```
frozenbugs, line 1400: Will this be super expensive if we clone a graph?

mufeili: I think it is ok as this is typically applied to graphs of 100-1000 nodes.

jermainewang (approved):

PR #4754, opened 2022-10-25
[Dist] fix etype issue in dist part pipeline

Rhett-Ying:
## Description

Previously, I reported a bug of edge feature support in dist part pipeline: https://github.com/dmlc/dgl/issues/4681. I just found it's not fixed yet.
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/tools/distpartitioning/utils.py
+++ b/tools/distpartitioning/utils.py
@@ -377,6 +377,9 @@ def write_edge_features(edge_features, edge_file):
     edge_file     : string 
         File in which the edge information is serialized
     """
+    # TODO[Rui]: Below is a temporary fix for etype and will be
+    # further refined in the near future as we'll shift to canonical
+    # etypes entirely.
     def format_etype(etype):
         etype, name = etype.split('/')
         etype = etype.split(':')[1]
--- a/tools/distpartitioning/utils.py
+++ b/tools/distpartitioning/utils.py
@@ -377,6 +377,12 @@ def write_edge_features(edge_features, edge_file):
     edge_file     : string 
         File in which the edge information is serialized
     """
+    def format_etype(etype):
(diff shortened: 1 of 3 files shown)


tools/distpartitioning/utils.py
```diff
@@ -377,6 +377,12 @@ def write_edge_features(edge_features, edge_file):
     edge_file     : string 
         File in which the edge information is serialized
     """
+    def format_etype(etype):
+        etype, name = etype.split('/')
+        etype = etype.split(':')[1]
+        return etype + '/' + name
+    edge_features = {format_etype(etype):
+        data for etype, data in edge_features.items()}
```
jermainewang, line 385: Use a for loop will be cleaner than list comprehension for this case.

tools/distpartitioning/utils.py
```diff
@@ -377,6 +377,12 @@ def write_edge_features(edge_features, edge_file):
     edge_file     : string 
         File in which the edge information is serialized
     """
+    def format_etype(etype):
+        etype, name = etype.split('/')
+        etype = etype.split(':')[1]
+        return etype + '/' + name
```
jermainewang, line 383: I generally don't like this kind of variable shadowing. Making it difficult to read. Also, is this to convert canonical etype (in the form of `'type_1:relation:type_2'`) to etype? Is this temporary as we are shifting to canonical etype entirely?

kylasa (approved):
Code changes looks good to me.

tools/distpartitioning/utils.py
jermainewang: Sure. Let's fix them in one PR that completely drops the support for string etype.

jermainewang (approved):

PR #4756, opened 2022-10-25
ParMETIS preprocessing is failing in a multi-node cluster because of the incorrect rank of the process

kylasa:
## Description

The rank of the process was incorrect in a multi-node setting. Now updated the key to use from the environment
dictionary to retrieve the rank of the process correctly.

## Checklist
- [X] Code is well-documented

### DIFF
--- a/tools/distpartitioning/parmetis_preprocess.py
+++ b/tools/distpartitioning/parmetis_preprocess.py
@@ -26,17 +26,18 @@ def get_proc_info():
     MPI_WORLDSIZE appropriately as described above to retrieve total no. of
     processes, when needed.
 
+    MPI_LOCALRANKID is only valid on a single-node cluster. In a multi-node cluster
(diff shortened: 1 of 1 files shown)


tools/distpartitioning/parmetis_preprocess.py
```diff
@@ -26,17 +26,18 @@ def get_proc_info():
     MPI_WORLDSIZE appropriately as described above to retrieve total no. of
     processes, when needed.
 
+    MPI_LOCALRANKID is only valid on a single-node cluster. In a multi-node cluster
```
Rhett-Ying, line 29: Is `PMI_RANK` always available in different MPI implementations such as `OpenMPI`, `MPICH`?

kylasa: Yes. it is as per the Intel's documentation for the MPI support.

Rhett-Ying (approved):

Issue #3435, opened 2021-10-22
canonical_etypes should be used in dgl.distributed.partition_graph


Sherler:
## To Reproduce

1.  define two edge types with same 'middle name': (person,loves,person) ,(person, loves, pet)
2. create partition_graph using `dgl.distributed.partition_graph`
3. there will be an exception:
```
dgl._ffi.base.DGLError: Edge type "loves" is ambiguous. Please use canonical edge type in the form of (srctype, etype, dsttype)
```

## Environment

 - DGL Version (e.g., 1.0): 0.7.1
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):
 - OS (e.g., Linux):
 - How you installed DGL (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version (if applicable): 10.2
 - GPU models and configuration (e.g. V100):
 - Any other relevant information:

Rhett-Ying:
As canonical etypes are not used in `partition_graph()` yet. Let me re-open this ticket.

- [ ] use canonical etypes when partition graph including `partition_graph()` and dist partition pipeline.
- [ ] load partitions and parse canonical etypes directly in `DistGraph` instead of going through different partitions.
- [ ] throw exception if loading deprecated configuration json. update corresponding doc.

PR #4777, opened 2022-10-28
[Dist] deprecate etype and always use canonical etype for partition and load

Rhett-Ying:
## Description

1. canonical etypes are used as keys of `etypes` and `edge_map` during partition.
2. `GraphParitionBook` can only be instantiated by canonical etypes.
3. canonical and non-canonical etypes are both supported to access via `DistGraph`, `GraphPartitionBook`.

target work item: https://github.com/dmlc/dgl/issues/3435
 
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/dist_graph.py
+++ b/python/dgl/distributed/dist_graph.py
@@ -481,7 +481,6 @@ class DistGraph:
     '''
     def __init__(self, graph_name, gpb=None, part_config=None):
         self.graph_name = graph_name
-        self._gpb_input = gpb
         if os.environ.get('DGL_DIST_MODE', 'standalone') == 'standalone':
             assert part_config is not None, \
                     'When running in the standalone model, the partition config file is required'
@@ -511,7 +510,7 @@ class DistGraph:
             self._client.map_shared_data(self._gpb)
             rpc.set_num_client(1)
         else:
-            self._init()
+            self._init(gpb)
             # Tell the backup servers to load the graph structure from shared memory.
             for server_id in range(self._client.num_servers):
                 rpc.send_request(server_id, InitGraphRequest(graph_name))
@@ -534,22 +533,22 @@ class DistGraph:
         self._ntype_map = {ntype:i for i, ntype in enumerate(self.ntypes)}
         self._etype_map = {etype:i for i, etype in enumerate(self.canonical_etypes)}
 
-    def _init(self):
+    def _init(self, gpb):
         self._client = get_kvstore()
         assert self._client is not None, \
                 'Distributed module is not initialized. Please call dgl.distributed.initialize.'
         self._g = _get_graph_from_shared_mem(self.graph_name)
         self._gpb = get_shared_mem_partition_book(self.graph_name, self._g)
         if self._gpb is None:
-            self._gpb = self._gpb_input
+            self._gpb = gpb
         self._client.map_shared_data(self._gpb)
 
     def __getstate__(self):
         return self.graph_name, self._gpb
 
     def __setstate__(self, state):
-        self.graph_name, self._gpb_input = state
-        self._init()
+        self.graph_name, gpb = state
+        self._init(gpb)
 
         self._ndata_store = {}
         self._edata_store = {}
@@ -885,8 +884,8 @@ class DistGraph:
         123718280
         """
         if etype is None:
-            return sum([self._gpb._num_edges(etype) \
-                for etype in self.canonical_etypes])
+            return sum([self._gpb._num_edges(c_etype)
+                for c_etype in self.canonical_etypes])
         return self._gpb._num_edges(etype)
 
     def out_degrees(self, u=ALL):
--- a/python/dgl/distributed/dist_graph.py
+++ b/python/dgl/distributed/dist_graph.py
@@ -22,7 +22,7 @@ from .partition import load_partition, load_partition_feats, load_partition_book
 from .graph_partition_book import PartitionPolicy, get_shared_mem_partition_book
 from .graph_partition_book import HeteroDataName, parse_hetero_data_name
 from .graph_partition_book import NodePartitionPolicy, EdgePartitionPolicy
-from .graph_partition_book import _etype_tuple_to_str, _etype_str_to_tuple
+from .graph_partition_book import _etype_str_to_tuple
 from .shared_mem_utils import _to_shared_mem, _get_ndata_path, _get_edata_path, DTYPE_DICT
 from . import rpc
 from . import role
@@ -885,8 +885,8 @@ class DistGraph:
         123718280
         """
         if etype is None:
-                return sum([self._gpb._num_edges(etype) \
-                    for etype in self.canonical_etypes])
+            return sum([self._gpb._num_edges(etype) \
+                for etype in self.canonical_etypes])
         return self._gpb._num_edges(etype)
 
     def out_degrees(self, u=ALL):
--- a/python/dgl/distributed/dist_graph.py
+++ b/python/dgl/distributed/dist_graph.py
@@ -22,6 +22,7 @@ from .partition import load_partition, load_partition_feats, load_partition_book
 from .graph_partition_book import PartitionPolicy, get_shared_mem_partition_book
 from .graph_partition_book import HeteroDataName, parse_hetero_data_name
 from .graph_partition_book import NodePartitionPolicy, EdgePartitionPolicy
+from .graph_partition_book import _etype_tuple_to_str, _etype_str_to_tuple
 from .shared_mem_utils import _to_shared_mem, _get_ndata_path, _get_edata_path, DTYPE_DICT
 from . import rpc
 from . import role
@@ -355,10 +356,10 @@ class DistGraphServer(KVServer):
             self.gpb.shared_memory(graph_name)
         assert self.gpb.partid == self.part_id
         for ntype in ntypes:
-            node_name = HeteroDataName(True, ntype, None)
+            node_name = HeteroDataName(True, ntype, "")
             self.add_part_policy(PartitionPolicy(node_name.policy_str, self.gpb))
         for etype in etypes:
-            edge_name = HeteroDataName(False, etype, None)
+            edge_name = HeteroDataName(False, etype, "")
             self.add_part_policy(PartitionPolicy(edge_name.policy_str, self.gpb))
 
         if not self.is_backup_server():
@@ -381,7 +382,7 @@ class DistGraphServer(KVServer):
                 # The feature name has the following format: edge_type + "/" + feature_name to avoid
                 # feature name collision for different edge types.
                 etype, feat_name = name.split('/')
-
+                etype = _etype_str_to_tuple(etype)
                 data_name = HeteroDataName(False, etype, feat_name)
                 self.init_data(name=str(data_name), policy_str=data_name.policy_str,
                                data_tensor=edge_feats[name])
@@ -503,6 +504,7 @@ class DistGraph:
             for name in edge_feats:
                 # The feature name has the following format: edge_type + "/" + feature_name.
                 etype, feat_name = name.split('/')
+                etype = _etype_str_to_tuple(etype)
                 self._client.add_data(str(HeteroDataName(False, etype, feat_name)),
                                       edge_feats[name],
                                       EdgePartitionPolicy(self._gpb, etype=etype))
@@ -530,32 +532,7 @@ class DistGraph:
 
         # When we store node/edge types in a list, they are stored in the order of type IDs.
         self._ntype_map = {ntype:i for i, ntype in enumerate(self.ntypes)}
-        self._etype_map = {etype:i for i, etype in enumerate(self.etypes)}
-
-        # Get canonical edge types.
-        # TODO(zhengda) this requires the server to store the graph with coo format.
-        eid = []
-        for etype in self.etypes:
-            type_eid = F.zeros((1,), F.int64, F.cpu())
-            eid.append(self._gpb.map_to_homo_eid(type_eid, etype))
-        eid = F.cat(eid, 0)
-        src, dst = dist_find_edges(self, eid)
-        src_tids, _ = self._gpb.map_to_per_ntype(src)
-        dst_tids, _ = self._gpb.map_to_per_ntype(dst)
-        self._canonical_etypes = []
-        etype_ids = F.arange(0, len(self.etypes))
-        for src_tid, etype_id, dst_tid in zip(src_tids, etype_ids, dst_tids):
-            src_tid = F.as_scalar(src_tid)
-            etype_id = F.as_scalar(etype_id)
-            dst_tid = F.as_scalar(dst_tid)
-            self._canonical_etypes.append((self.ntypes[src_tid], self.etypes[etype_id],
-                                           self.ntypes[dst_tid]))
-        self._etype2canonical = {}
-        for src_type, etype, dst_type in self._canonical_etypes:
-            if etype in self._etype2canonical:
-                self._etype2canonical[etype] = ()
-            else:
-                self._etype2canonical[etype] = (src_type, etype, dst_type)
+        self._etype_map = {etype:i for i, etype in enumerate(self.canonical_etypes)}
 
     def _init(self):
         self._client = get_kvstore()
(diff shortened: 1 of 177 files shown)


python/dgl/distributed/dist_graph.py
```diff
@@ -355,10 +356,10 @@ def __init__(self, server_id, ip_config, num_servers,
             self.gpb.shared_memory(graph_name)
         assert self.gpb.partid == self.part_id
         for ntype in ntypes:
-            node_name = HeteroDataName(True, ntype, None)
+            node_name = HeteroDataName(True, ntype, "")
```
jermainewang, line 359: Have question about this change. Will come to this later.

python/dgl/distributed/dist_graph.py
```diff
@@ -568,18 +545,12 @@ def _init(self):
         self._client.map_shared_data(self._gpb)
 
     def __getstate__(self):
-        return self.graph_name, self._gpb, self._canonical_etypes
+        return self.graph_name, self._gpb
 
     def __setstate__(self, state):
-        self.graph_name, self._gpb_input, self._canonical_etypes = state
+        self.graph_name, self._gpb_input = state
```
jermainewang, line 551: This is a bad way of passing `_gpb_input` to `self._init()`. Can we do `self._init(gpb_input)` instead?

python/dgl/distributed/dist_graph.py
```diff
@@ -928,10 +885,8 @@ def num_edges(self, etype=None):
         123718280
         """
         if etype is None:
-            if len(self.etypes) == 1:
-                return self._gpb._num_edges(self.etypes[0])
-            else:
-                return sum([self._gpb._num_edges(etype) for etype in self.etypes])
+            return sum([self._gpb._num_edges(etype) \
+                for etype in self.canonical_etypes])
```
jermainewang, line 889: The variable shadowing makes it quite difficult to read. Could you change the iter variable inside the list comprehension to sth else?

jermainewang: Also, I don't think you need to add `\` if you use formatting tools like black.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -21,15 +20,23 @@
...
-
-def _str_to_tuple(s):
-    try:
-        ret = literal_eval(s)
-    except ValueError:
-        ret = s
+CANONICAL_ETYPE_DELIMITER = ":"
+
+def _etype_tuple_to_str(c_etype):
+    '''Convert canonical etype from tuple to string.
```
jermainewang, line 26: Give a simple example in the docstring.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -21,15 +20,23 @@
...
+    '''
+    assert isinstance(c_etype, tuple) and len(c_etype) == 3, \
+        "Passed-in canonical etype should be in format of (str, str, str)."
```
jermainewang, line 29: `"Passed-in ... . But got {c_etype}."`

Printing out the value will be more useful.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -21,15 +20,23 @@
...
+    return CANONICAL_ETYPE_DELIMITER.join(c_etype)
+
+def _etype_str_to_tuple(c_etype):
+    '''Convert canonical etype from tuple to string.
```
jermainewang, line 33: same. give a simple example in the docstring

python/dgl/distributed/graph_partition_book.py
```diff
@@ -793,27 +828,19 @@ def __init__(self, part_id, num_parts, node_map, edge_map, ntypes, etypes):
...
-                self._canonical_etypes[etype_id] = c_etype
-                if etype in self._etype2canonical:
-                    # If one etype maps to multiple canonical etypes, empty
-                    # tuple is used to indicate such ambiguity casued by etype.
-                    # See more details in self._to_canonical_etype().
-                    self._etype2canonical[etype] = tuple()
-                else:
-                    self._etype2canonical[etype] = c_etype
+            assert isinstance(etype, tuple) and len(etype) == 3, \
+                "Only canonical etype in tuple is supproted: (str, str, str)."
```
jermainewang, line 832: Same. Should print out the invalid input to make it easier for users to debug.

```
"Expect canonical edge type in a triplet of string, but got {etype}."
```

python/dgl/distributed/graph_partition_book.py
```diff
@@ -793,27 +828,19 @@ def __init__(self, part_id, num_parts, node_map, edge_map, ntypes, etypes):
...
+            c_etype = etype
```
jermainewang, line 833: Could you rename the loop variable at L830 to c_etype? Then you can delete L833 and change L834 to `etype = c_etype[1]`.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -1176,26 +1206,28 @@ class PartitionPolicy(object):
...
-                policy_str = NODE_PART_POLICY + ":" + DEFAULT_NTYPE
+                policy_str = NODE_PART_POLICY + POLICY_DELIMITER + DEFAULT_NTYPE
             else:
-                policy_str = EDGE_PART_POLICY + ":" + DEFAULT_ETYPE[1]
+                policy_str = EDGE_PART_POLICY + POLICY_DELIMITER + DEFAULT_ETYPE[1]
         self._policy_str = policy_str
         self._part_id = partition_book.partid
         self._partition_book = partition_book
-        self._type_name = _str_to_tuple(self.policy_str[5:])
+        self._type_name = self.policy_str[5:]
```
jermainewang, line 1228: I will suggest merge this line with L1215. E.g.,
```
policy_name, self._type_name = policy_str.split(POLICY_DDELIMITER)
```

You can also use regex to perform more rigorous matching (and checking).

python/dgl/distributed/graph_partition_book.py
```diff
@@ -1330,16 +1362,19 @@ class NodePartitionPolicy(PartitionPolicy):
...
+            NODE_PART_POLICY + POLICY_DELIMITER + ntype, partition_book
         )
 
 
 class EdgePartitionPolicy(PartitionPolicy):
     """Partition policy for edges."""
 
     def __init__(self, partition_book, etype=DEFAULT_ETYPE):
+        assert isinstance(etype, tuple) and len(etype) == 3, \
+            "Only canonical etype in tuple is supproted: (str, str, str)."
```
jermainewang, line 1374: Same suggestion to the error message.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -1363,26 +1398,32 @@ class HeteroDataName(object):
 
     def __init__(self, is_node, entity_type, data_name):
         self._policy = NODE_PART_POLICY if is_node else EDGE_PART_POLICY
+        if not is_node:
+            assert isinstance(entity_type, tuple) and len(entity_type) == 3, \
+                "Only canonical etype in tuple is supproted: (str, str, str)."
```
jermainewang, line 1403: Same suggestion to the error message.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -1363,26 +1398,32 @@ class HeteroDataName(object):
...
         self.data_name = data_name
 
     @property
     def policy_str(self):
         """concatenate policy and entity type into string"""
-        return self._policy + ":" + str(self.get_type())
+        entity_type = self.get_type()
+        if self.is_edge():
+            entity_type = _etype_tuple_to_str(entity_type)
+        return self._policy + POLICY_DELIMITER + entity_type
```
jermainewang, line 1413: Why not directly return `self._policy_str` since you've saved it at L1225.

python/dgl/distributed/partition.py
```diff
@@ -24,6 +34,43 @@
...
+
+def _load_part_config(part_config):
+    ''' Load part config and format
+    '''
+    try:
+        with open(part_config) as f:
+            part_metadata = _format_part_metadata(json.load(f),
+                _etype_str_to_tuple)
+    except AssertionError as e:
+        raise DGLError(f"Failed to load partition config due to {e}. "
```
jermainewang, line 61: I saw sometimes we are using assertion but sometimes raise error. Do we have a coherent suggestion? @frozenbugs

python/dgl/distributed/partition.py
```diff
@@ -24,6 +34,43 @@
     ETYPE: F.int32
     }
 
+def _format_part_metadata(part_metadata, formator):
+    ''' Format etypes with specified formator
+    '''
```
jermainewang, line 39: No whitespace after `'''`. New newline after a short description. Miss `.` at the end.

tools/chunk_graph.py
```diff
@@ -37,9 +40,9 @@ def _chunk_graph(g, name, ndata_paths, edata_paths, num_chunks, output_path):
     ):
         edata_paths = {g.etypes[0]: ndata_paths}
     # Then convert all edge types to canonical edge types
-    etypestrs = {etype: ":".join(etype) for etype in g.canonical_etypes}
+    etypestrs = {etype: _etype_tuple_to_str(etype) for etype in g.canonical_etypes}
```
jermainewang, line 43: I will revert this change for two reasons:
(1) Scripts under tools should try to avoid calling internal DGL utilities. Think of them as user-written scripts.
(2) The chunked graph format is different from graph partition book, though they happen to represent canonical edge type in the same way.

jermainewang (changes requested):

python/dgl/distributed/partition.py
frozenbugs: The key is to provide enough information when error happens. Solely using assert is not so good, the error msg will just be, assert failure.

But we can easily create a wrapper assert function with a string as input, such that we can simplify the code structure.

jermainewang: Probably worth looking at how other python package do logging/checking. For C++, we use GLOG.

python/dgl/distributed/graph_partition_book.py
Rhett-Ying: because we need to append the data_name as well

python/dgl/distributed/partition.py
Rhett-Ying: For me, when I need to throw an exception and to be caught by caller, I choose to raise DGLError. If I don't want to catch it, just let the exception aborts the app, use `assert` is ok for me.

jermainewang (approved):

python/dgl/distributed/graph_partition_book.py
```diff
@@ -738,14 +777,32 @@ def canonical_etypes(self):
...
+        Returns
+        -------
+        (str, str, str)
+            The corresponding canonical edge type
+        """
+        assert etype in (
+            DEFAULT_ETYPE,
+            DEFAULT_ETYPE[1],
+        ), "Base partition book only supports homogeneous graph."
+        return self.canonical_etypes
```
frozenbugs, line 797: This one looks wrong to me.
It returns canonical_etypes, while the etype is not used to get canonical_etype.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -792,28 +845,19 @@ def __init__(self, part_id, num_parts, node_map, edge_map, ntypes, etypes):
...
-                if etype in self._etype2canonical:
-                    # If one etype maps to multiple canonical etypes, empty
-                    # tuple is used to indicate such ambiguity casued by etype.
-                    # See more details in self._to_canonical_etype().
-                    self._etype2canonical[etype] = tuple()
-                else:
-                    self._etype2canonical[etype] = c_etype
+        for c_etype, etype_id in etypes.items():
+            assert isinstance(c_etype, tuple) and len(c_etype) == 3, \
+                f"Expect canonical edge type in a triplet of string, but got {c_etype}."
```
frozenbugs, line 850: over 80 chars.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -792,28 +845,19 @@ def __init__(self, part_id, num_parts, node_map, edge_map, ntypes, etypes):
...
+            etype = c_etype[1]
+            self._etypes[etype_id] = etype
+            self._canonical_etypes[etype_id] = c_etype
+            if etype in self._etype2canonical:
+                # If one etype maps to multiple canonical etypes, empty
```
frozenbugs, line 855: There are still some space before hit 80 char, let's make use of them.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -1363,26 +1414,32 @@ class HeteroDataName(object):
 
     def __init__(self, is_node, entity_type, data_name):
         self._policy = NODE_PART_POLICY if is_node else EDGE_PART_POLICY
+        if not is_node:
+            assert isinstance(entity_type, tuple) and len(entity_type) == 3, \
+                f"Expect canonical edge type in a triplet of string, but got {entity_type}."
```
frozenbugs, line 1419: over 80 chars.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -2,14 +2,13 @@
 
 import pickle
```
frozenbugs, line 3: There are multiple classes defined in this file, could be better if we can break it to multiple pyfiles.

python/dgl/distributed/graph_partition_book.py
```diff
@@ -1363,26 +1414,32 @@ class HeteroDataName(object):
...
     def is_edge(self):
         """Is this the name of edge data"""
-        return EDGE_PART_POLICY in self.policy_str
+        return self._policy == EDGE_PART_POLICY
 
     def get_type(self):
         """The type of the node/edge.
         This is only meaningful in a heterogeneous graph.
-        In homogeneous graph, type is '_N' for a node and '_E' for an edge.
+        In homogeneous graph, type is '_N' for a node and '_N:_E:_N' for an edge.
```
frozenbugs, line 1442: over 80 chars.

python/dgl/distributed/graph_partition_book.py
Rhett-Ying: max-length 100?

python/dgl/distributed/graph_partition_book.py
Rhett-Ying: It's correct as it's `BasicPartitionBook`. canonical etype is always `('_N', '_E', '_N')`

python/dgl/distributed/partition.py
```diff
@@ -24,6 +34,43 @@
     ETYPE: F.int32
     }
 
+def _format_part_metadata(part_metadata, formator):
```
frozenbugs, line 37: formator -> formatter?

Not sure whether formator is a special word in dgl, if you are talking about data-formatting, it should be formatter.

python/dgl/distributed/graph_partition_book.py
jermainewang: @Rhett-Ying No, max-length is 80.

jermainewang: @frozenbugs Look like our linter is still using max-length=100. https://github.com/dmlc/dgl/blob/master/tests/lint/pylintrc#L340

Rhett-Ying: fixed in https://github.com/dmlc/dgl/pull/4809

python/dgl/distributed/graph_partition_book.py
Rhett-Ying: fixed in https://github.com/dmlc/dgl/pull/4809

python/dgl/distributed/graph_partition_book.py
Rhett-Ying: fixed in https://github.com/dmlc/dgl/pull/4809

python/dgl/distributed/graph_partition_book.py
Rhett-Ying: fixed in https://github.com/dmlc/dgl/pull/4809

python/dgl/distributed/graph_partition_book.py
Rhett-Ying: they are put together probably because they are quite related, I think. Base class, derived classes, highly related classes. Maybe we could re-locate them in a separate PR to be clear, if you insist on this.

PR #4781, opened 2022-10-31
[Bugfix] Fix that UVA cannot work on old GPUs

### LINKED ISSUE #4726: [Bug] UVA cannot work on old GPUs (K80)
## 🐛 Bug

show 'Check failed: e == cudaSuccess || e == cudaErrorCudartUnloading: CUDA: an illegal memory access was encountered
' when I use 'python node_classification.py' .  But it is fine when use 'python node_classification.py --mide puregpu'.

1. cmd  'python node_classification.py'

## Environment

 - DGL Version :0.9
 - PyTorch :1.9
 - OS : linux
 - Python version:3.6.3
 - CUDA/cuDNN version (if applicable): 10.2
 - GPU models and configuration (e.g. V100): k80
it is fine when use 'python node_classification.py --mide puregpu'

yaox12:
## Description

Resolves #4726.

The device attribute `cudaDevAttrCanUseHostPointerForRegisteredMem` of K80 is 0, so we cannot access the host pointer directly from the device even with UVA. We should get the corresponding device pointer via `cudaHostGetDevicePointer`.

For devices whose `cudaDevAttrCanUseHostPointerForRegisteredMem` is 1, the device pointer returned by `cudaHostGetDevicePointer` will match the original host pointer.

## Checklist
- [x] Code is well-documented

## Changes
1. All functions with the `ATEN_CSR_SWITCH_CUDA_UVA` dispatcher
2. `IndexSelect()`
3. `IndexSelectCPUFromGPU()` and `IndexScatterGPUToCPU()`

### DIFF
--- a/src/array/cuda/array_index_select.cu
+++ b/src/array/cuda/array_index_select.cu
@@ -18,7 +18,7 @@ template <DGLDeviceType XPU, typename DType, typename IdType>
 NDArray IndexSelect(NDArray array, IdArray index) {
   cudaStream_t stream = runtime::getCurrentCUDAStream();
   const DType* array_data = array.Ptr<DType>();
-  if (array->ctx.device_type == kDGLCPU) {
+  if (array.IsPinned()) {
     CUDA_CALL(cudaHostGetDevicePointer(&array_data, array.Ptr<DType>(), 0));
   }
   const IdType* idx_data = static_cast<IdType*>(index->data);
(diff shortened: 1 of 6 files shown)


chang-l (approved):
In general, looks good to me. 
Btw, does it make sense to you to move `cudaHostGetDevicePointer` in `PinData()` function, or bring it closer to `PinData()` function? (e.g., maybe `PinData()` can return a different(device) pointer if the pinned data has a different device virtual address)

yaox12:
> In general, looks good to me. Btw, does it make sense to you to move `cudaHostGetDevicePointer` in `PinData()` function, or bring it closer to `PinData()` function? (e.g., maybe `PinData()` can return a different(device) pointer if the pinned data has a different device virtual address)

Then we have to change the structure of `NDArray` to add a field to save this device pointer. There is a risk of backward compatibility I think.

yaox12:
@dgl-bot

PR #4786, opened 2022-10-31
[Dist] alltoall gloo wrapper primitive returns None values in the output list of tensors which fails the torch.cat function 

kylasa:
## Description

In an all to all message, if the return value (tensor) from a node is empty tensor then None was returned to the function caller which is causing the torch.cat() to fail. Apparently torch concatenate function cannot handle None values in the input list of tensors. 

Now, with the fix, if the return'ed tensors are 0-sized tensors (please note that here we perform padding, so we need to remove padding and compare the sizes appropriately) then it instead of returning None, it does include None in the list of tensors to the invoking function.

## Checklist
- [X] Code is well-documented

### DIFF
--- a/tools/branch.info
+++ b/tools/branch.info
@@ -1,1 +1,0 @@
-flexpipeline
--- a/tools/branch.info
+++ b/tools/branch.info
@@ -1,0 +1,1 @@
+flexpipeline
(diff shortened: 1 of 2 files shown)


Rhett-Ying (commented):
why remove `dlpack` and `thrust` from `third_party`? by accident?

kylasa:
Using another PR to merge this code into the master branch. Now this PR will become obsolete

PR #4788, opened 2022-11-01
[Dist] alltoall gloo wrapper primitive returns None values in the output list of tensors which fails the torch.cat function

kylasa:
## Description

In an all to all message, if the return value (tensor) from a node is empty tensor then None was returned to the function caller which is causing the torch.cat() to fail. Apparently torch concatenate function cannot handle None values in the input list of tensors.

Now, with the fix, if the return'ed tensors are 0-sized tensors (please note that here we perform padding, so we need to remove padding and compare the sizes appropriately) then it instead of returning None, it does include None in the list of tensors to the invoking function.

## Checklist
- [X] Code is well-documented

### DIFF
--- a/tools/distpartitioning/dist_lookup.py
+++ b/tools/distpartitioning/dist_lookup.py
@@ -296,6 +296,9 @@ class DistLookupService:
 
         # Send the shuffle-global-nids to their respective ranks.
         mapped_global_nids = alltoallv_cpu(self.rank, self.world_size, shuffle_nids_list)
+        for idx in range(len(mapped_global_nids)):
+            if mapped_global_nids[idx] == None:
+                mapped_global_nids[idx] = torch.empty((0,), dtype=torch.int64)
 
         # Reorder to match global_nids (function parameter).
         global_nids_order = np.concatenate(id_list)


tools/distpartitioning/dist_lookup.py
```diff
@@ -296,6 +296,9 @@ def get_shuffle_nids(self, global_nids, my_global_nids, my_shuffle_global_nids):
 
         # Send the shuffle-global-nids to their respective ranks.
         mapped_global_nids = alltoallv_cpu(self.rank, self.world_size, shuffle_nids_list)
+        for idx in range(len(mapped_global_nids)):
```
Rhett-Ying, line 299: why not use `for idx in mapped_global_nids:` ?

kylasa: If we use `for item in mapped_global_nids:`, then we can check if item is None or not. But this will not solve our purpose. Here, we are removing `None` in the mapped_global_nids (which is a list of tensors and some can be None as well). 

we can also rewrite the code as follows: 
for idx, item in enumerate(mapped_global_nids):
  if item == None: 
     mapped_global_nids[idx] = torch.empty((0,), dtype=torch.int64)
mapped_global_nids = torch.cat(mapped_global_nids). 

Either way I am fine with the code. 
Please let me know in case you think that this for-loop needs to change. I will change it accordingly.

Rhett-Ying (approved):

PR #4790, opened 2022-11-01
[DIST] Message size to retrieve SHUFFLE_GLOBAL_NIDs is resulting in very large messages and resulting in killed process

kylasa:
## Description

When retrieving SHUFFLE_GLOBAL_NID's for src and dst end points of the local edges in the current graph partition we used to send both the source and destination end points as a single list to the distributed lookup service. For Fe_large graph this message size is becoming about 50GB+ (3.2 billion edges). This increases the pressure on the memory used to send/receive such a large message and in the case of fe_large graph the process is getting killed. 

Now, we batch the outgoing message to distributed lookup service and ensure that the in each iteration we only send a message of maximum size of 250 MILLION entries (which is about 2 GB). This will reduce the memory need to support this operation and with this change fe_large graph is able to run correctly. 

Also another optimization is to NOT send destination end points to the distributed lookup service, since current process will have associated SHUFFLE_GLOBAL_NID's. We use current process local information to retrieve these IDs. 

## Checklist
- [X] Code is well-documented

### DIFF
--- a/tools/distpartitioning/globalids.py
+++ b/tools/distpartitioning/globalids.py
@@ -98,10 +98,7 @@ def lookup_shuffle_global_nids_edges(rank, world_size, edge_data, id_lookup, nod
     # Determine the no. of times each process has to send alltoall messages.
     all_sizes = allgather_sizes([node_list.shape[0]], world_size, return_sizes=True)
     max_count = np.amax(all_sizes)
-    if max_count >= BATCH_SIZE:
-        num_splits = max_count // BATCH_SIZE + 1
-    else:
-        num_splits = 1
+    num_splits = max_count // BATCH_SIZE + 1
 
     # Split the message into batches and send.
     splits = np.array_split(node_list, num_splits)
--- a/tools/distpartitioning/globalids.py
+++ b/tools/distpartitioning/globalids.py
@@ -86,13 +86,47 @@ def lookup_shuffle_global_nids_edges(rank, world_size, edge_data, id_lookup, nod
         dictionary where keys are column names and values are numpy arrays representing all the
         edges present in the current graph partition
     '''
+    # Make sure that the outgoing message size does not exceed 2GB in size. 
+    # Even though gloo can handle upto 10GB size of data in the outgoing messages,
+    # it needs additional memory to store temporary information into the buffers which will increase
+    # the memory needs of the process. 
+    MILLION = 1000 * 1000
+    BATCH_SIZE = 250 * MILLION
     memory_snapshot("GlobalToShuffleIDMapBegin: ", rank)
-    node_list = np.concatenate([edge_data[constants.GLOBAL_SRC_ID], edge_data[constants.GLOBAL_DST_ID]])
-    shuffle_ids = id_lookup.get_shuffle_nids(node_list,
+    node_list = edge_data[constants.GLOBAL_SRC_ID]
+
+    # Determine the no. of times each process has to send alltoall messages.
+    all_sizes = allgather_sizes([node_list.shape[0]], world_size, return_sizes=True)
+    max_count = np.amax(all_sizes)
+    if max_count >= BATCH_SIZE:
+        num_splits = max_count // BATCH_SIZE + 1
+    else:
+        num_splits = 1
+
+    # Split the message into batches and send.
+    splits = np.array_split(node_list, num_splits)
+    shuffle_mappings = []
+    for item in splits:
+        logging.info(f'[Rank: {rank} SENDING : {item.shape}')
+        shuffle_ids = id_lookup.get_shuffle_nids(item,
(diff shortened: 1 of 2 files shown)


tools/distpartitioning/globalids.py
```diff
@@ -86,13 +86,47 @@ def lookup_shuffle_global_nids_edges(rank, world_size, edge_data, id_lookup, nod
...
+    BATCH_SIZE = 250 * MILLION
     memory_snapshot("GlobalToShuffleIDMapBegin: ", rank)
-    node_list = np.concatenate([edge_data[constants.GLOBAL_SRC_ID], edge_data[constants.GLOBAL_DST_ID]])
-    shuffle_ids = id_lookup.get_shuffle_nids(node_list,
+    node_list = edge_data[constants.GLOBAL_SRC_ID]
+
+    # Determine the no. of times each process has to send alltoall messages.
+    all_sizes = allgather_sizes([node_list.shape[0]], world_size, return_sizes=True)
+    max_count = np.amax(all_sizes)
+    if max_count >= BATCH_SIZE:
```
Rhett-Ying, line 101: `num_splits = max_count // BATCH_SIZE + 1` should be ok regardless of `max_count % BATCH_SIZE == 0` or `max_count >= BATCH_SIZE`.

tools/distpartitioning/globalids.py
```diff
@@ -86,13 +86,47 @@ def lookup_shuffle_global_nids_edges(rank, world_size, edge_data, id_lookup, nod
...
+        num_splits = max_count // BATCH_SIZE + 1
+    else:
+        num_splits = 1
+
+    # Split the message into batches and send.
+    splits = np.array_split(node_list, num_splits)
+    shuffle_mappings = []
+    for item in splits:
+        logging.info(f'[Rank: {rank} SENDING : {item.shape}')
```
Rhett-Ying, line 110: `NameError: name 'logging' is not defined`

tools/distpartitioning/globalids.py
kylasa: Done. Updated the code accordingly.

tools/distpartitioning/globalids.py
kylasa: Removed the logging.info line. This was only for debugging purposes.

tools/distpartitioning/globalids.py
```diff
@@ -86,13 +86,47 @@ def lookup_shuffle_global_nids_edges(rank, world_size, edge_data, id_lookup, nod
...
+    # Here retrieve SHUFFLE_GLOBAL_NODE_IDs for the destination end points of local edges.
+    uniq_ids, inverse_idx = np.unique(edge_data[constants.GLOBAL_DST_ID], return_inverse=True)
+    common, idx1, idx2 = np.intersect1d(uniq_ids, node_data[constants.GLOBAL_NID], assume_unique=True, return_indices=True)
+    assert len(common) == len(uniq_ids)
+
+    edge_data[constants.SHUFFLE_GLOBAL_DST_ID] = node_data[constants.SHUFFLE_GLOBAL_NID][idx2][inverse_idx]
+    assert len(edge_data[constants.SHUFFLE_GLOBAL_DST_ID]) == len(edge_data[constants.GLOBAL_DST_ID])
 
-    edge_data[constants.SHUFFLE_GLOBAL_SRC_ID], edge_data[constants.SHUFFLE_GLOBAL_DST_ID] = np.split(shuffle_ids, 2)
     memory_snapshot("GlobalToShuffleIDMap_AfterLookupServiceCalls: ", rank)
```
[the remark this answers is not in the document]
kylasa: Agreed. These temporary variables like uniq_ids, inverse_idx etc... will consume a couple of GB's of memory.
But when the function returns, these variables will become out of scope, and the function's stack will become obsolete as well. `gc` will take care of releasing the memory. 

In my testing, I did notice that once the function returns the memory usage decreases accordingly. So in this case, this might give us any advantage by explicitly releasing the memory. 

But if you still prefer that we should set uniq_ids = None and similarly for other variables please let me know and I will put this code. But I think this will increase the code unnecessarily. I will leave this decision upto you. I prefer the code the way it exists as of now.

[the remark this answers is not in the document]
Rhett-Ying: what I mean is when you call `memory_snapshot`, such temp variables still exist, right? And the memory occupied by them will be recorded into the snapshot.

Rhett-Ying (approved):

tools/distpartitioning/globalids.py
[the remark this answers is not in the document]
Rhett-Ying: Discussed offline and the author claims that the temp variables will not counted when generating the memory snapshot.

PR #4791, opened 2022-11-01
[Dist] Create <graph_name>_stats.txt file if it does not exist before ParMETIS execution

kylasa:
## Description

Here, we check if the <graph_name>_stats.txt does not exist then creates one in the current directory. 
<graph_name>_stats.txt file has one single line with 3 columns (num_nodes, num_edges and num_constraints). 
num_constraints is the sum of number of node types in the graph (represented as 1-hot vector) and one each for train/test/val masks. 

stats file will be created in the current working directory. 

## Checklist
- [X] Code is well-documented

### DIFF
--- a/tools/distpartitioning/parmetis_preprocess.py
+++ b/tools/distpartitioning/parmetis_preprocess.py
@@ -293,7 +293,7 @@ def gen_parmetis_input_args(params, schema_map):
     #   num_constraints = no. of node types + train_mask + test_mask + val_mask
     #   Here, (train/test/val) masks will be set to 1 if these masks exist for
     #   all the node types in the graph, otherwise these flags will be set to 0
-    assert graph_name in schema_map, "Graph name is not present in the json file"
+    assert constants.GRAPH_NAMEe in schema_map, "Graph name is not present in the json file"
     graph_name = schema_map[constants.GRAPH_NAME]
     if not os.path.isfile(f'{graph_name}_stats.txt'):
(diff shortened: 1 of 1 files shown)


tools/distpartitioning/parmetis_preprocess.py
```diff
@@ -287,6 +287,36 @@ def gen_parmetis_input_args(params, schema_map):
         schema_map[constants.STR_NUM_NODES_PER_CHUNK],
     )
 
+    # Check if <graph-name>_stats.txt exists, if not create one using metadata.
+    # Here stats file will be created in the current directory. 
+    # No. of constraints, third column in the stats file is computed as follows:
+    #   num_constraints = no. of node types + train_mask + test_mask + val_mask
+    #   Here, (train/test/val) masks will be set to 1 if these masks exist for
+    #   all the node types in the graph, otherwise these flags will be set to 0
+    assert graph_name in schema_map, "Graph name is not present in the json file"
```
Rhett-Ying, line 296: where is `graph_name` defined? should be `constants.GRAPH_NAME`?

kylasa: graph_name is present in the json file of the input graph dataset. It is one of the keys in the dictionary object created by reading the json file. Per our new dataset format, graph_name should be present in the json file for the input dataset. 

This is a typo and was interchanged the next line. You are right, it should be constants.GRAPH_NAME. Changed it accordingly.

Rhett-Ying (approved):

PR #4795, opened 2022-11-01
[Dist] Pyarrow reading large csv files memory optimization

kylasa:
## Description

Default reading of large files using pyarrow, treating it as a single chunk takes very large amounts of memory footprints. In addition, this memory lingers in the VmData space and does not return to the system memory. 

To avoid this situation and to reduce the memory footprint of pyarrow reading csv files we now use Chunked reading of csv files  using pyarrow and allowing it to read one chunk at a time. 

In addition, if storing a slice of the numpy array read from a large file is resulting in original numpy array not getting released back to the system and stays throught out the execution in the VmData space. We now do a copy.deepcopy() of the slice which will create a separate memory block and when the original numpy array gets out-of-scope it gets released to the system. 

## Checklist
- [X] Code is well-documented

### DIFF
--- a/tools/distpartitioning/dataset_utils.py
+++ b/tools/distpartitioning/dataset_utils.py
@@ -337,11 +337,18 @@ def get_dataset(input_dir, graph_name, rank, world_size, schema_map):
             if not os.path.isabs(edge_file):
                 edge_file = os.path.join(input_dir, edge_file)
             logging.info(f'Loading edges of etype[{etype_name}] from {edge_file}')
-            data_df = csv.read_csv(edge_file,
-                        read_options=pyarrow.csv.ReadOptions(autogenerate_column_names=True), 
-                        parse_options=pyarrow.csv.ParseOptions(delimiter=' '))
-            src_ids.append(data_df['f0'].to_numpy())
-            dst_ids.append(data_df['f1'].to_numpy())
+
+            read_options=pyarrow.csv.ReadOptions(use_threads=True, block_size=4096, autogenerate_column_names=True)
+            parse_options=pyarrow.csv.ParseOptions(delimiter=' ')
+            with pyarrow.csv.open_csv(edge_file, read_options=read_options, parse_options=parse_options) as reader:
+                for next_chunk in reader:
+                    if next_chunk is None:
+                        break
+
+                    next_table = pyarrow.Table.from_batches([next_chunk])
+                    src_ids.append(next_table['f0'].to_numpy())
+                    dst_ids.append(next_table['f1'].to_numpy())
+
         src_ids = np.concatenate(src_ids)
         dst_ids = np.concatenate(dst_ids)
 
--- a/tools/distpartitioning/dist_lookup.py
+++ b/tools/distpartitioning/dist_lookup.py
@@ -3,6 +3,7 @@ import os
 import numpy as np
 import pyarrow
 import torch
+import copy
 
 from pyarrow import csv
 from gloo_wrapper import alltoallv_cpu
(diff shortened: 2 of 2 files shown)


tools/distpartitioning/dist_lookup.py
```diff
@@ -75,8 +85,12 @@ def __init__(self, input_dir, ntype_names, id_map, rank, world_size):
                 end = count
             type_nid_begin.append(start)
             type_nid_end.append(end)
+
             # Slice the partition-ids which belong to the current instance.
-            partid_list.append(ntype_partids[start:end])
+            partid_list.append(copy.deepcopy(ntype_partids[start:end]))
```
Rhett-Ying, line 90: are you sure that only use `deepcopy` and `del` together, occupied memory can be released? `del` is not sufficient?

tools/distpartitioning/dataset_utils.py
```diff
@@ -337,11 +337,18 @@ def get_dataset(input_dir, graph_name, rank, world_size, schema_map):
             if not os.path.isabs(edge_file):
                 edge_file = os.path.join(input_dir, edge_file)
             logging.info(f'Loading edges of etype[{etype_name}] from {edge_file}')
-            data_df = csv.read_csv(edge_file,
-                        read_options=pyarrow.csv.ReadOptions(autogenerate_column_names=True), 
-                        parse_options=pyarrow.csv.ParseOptions(delimiter=' '))
-            src_ids.append(data_df['f0'].to_numpy())
-            dst_ids.append(data_df['f1'].to_numpy())
+
+            read_options=pyarrow.csv.ReadOptions(use_threads=True, block_size=4096, autogenerate_column_names=True)
```
Rhett-Ying, line 341: how to define best value for `block_size`? have you profiled perf with different values? `4096` is the default block size in modern computer?

Rhett-Ying: `4096` is a bit small, I think.

tools/distpartitioning/dist_lookup.py
kylasa: This is the reason for these 2 lines of code. 

Earlier, partition-id file was read and only a slice of this numpy array was stored in the dist. lookup service object. The problem is that since we are storing a slice of the numpy array, entire numpy array is stored in the memory/RAM and this was not released until the end of execution of the pipeline. Since these files are very large and can be 50GB in memory.... so this is huge chunk of memory which is used unnecessarily. 

To solve the above problem now, we do the following: 
1. after reading the partition-id file, we copy the slice of the numpy array which is owned by the current process. In this way the original numpy array which was created during this file is read will be released once the numpy array becomes out of scope. 
2. For copying purposes, we use copy.deepcopy() method. This will allocate new block of memory and copies the slice of numpy array.... so that no references of the numpy array are stored or referred anywhere. 
3. We also explicitly delete the memory by using `del` keyword for the numpy array variable. 
4. Per python documentation, `del variablename` or variable_name = None either of these ways will deallocate the memory (with all the references) behind the numpy array. 

In my testing with Fe_large graph, the partition id files can occupy 32GB of physical memory, which is RAM. By doing the deepcopy and explicitly releasing the memory by using `del` operator releases this memory. Please also note that <variable_name> = None will release the memory of the numpy array. 

Do you think there is a better way of handling this problem. ?? please let me know.

tools/distpartitioning/dataset_utils.py
kylasa: Since we are using multiple threads, I guess internally in pyarrow each thread is allowed to a chunk of the size 4096. If we do not perform chunked reading of the file.. it appears that the pyarrow is taking massive amounts of RAM to complete the reading csv files. But at the same time it is very fast, that fastest I have run so far. 

Even after reading is complete and pyarrow variables becomes out of scope, the memory consumed for reading files was NOT getting released back to the system. 

To decrease the memory foot print of pyarrow library module when reading csv files, we now use chunked reading of the files. This is reduced the total amount of memory used by pyarrow very significantly, to the order of 50GB+ for fe_large graph. 

We can use various chunk sizes, 4096, 4096*2, 4096*4 etc... Of all these block sizes I did not notice any significant change in the time to read the csv files. There is a very slight change the resident memory (RAM) used in all these instances.

Rhett-Ying (approved):

PR #4801, opened 2022-11-02
[MPOps] Add edge-wise message passing operators

jermainewang:
## Description

Part of #4758 . See the issue for what are the remaining work.

## Checklist
- [x] Code is well-documented

## Changes

Added the following new APIs
* `dgl.copy_u`
* `dgl.copy_v`
* `dgl.u_{op}_v` where op can be "add", "sub", "mul", "div", "dot".
* Finished the docstring and unit tests.
* A new typing submodule `dgl.typing` which includes four type notations: `NData`, `EData`, `NType`, `EType`.

### DIFF
--- a/python/dgl/mpops/edgewise.py
+++ b/python/dgl/mpops/edgewise.py
@@ -1,0 +1,231 @@
+"""Operators for computing edge data."""
+import sys
+
+from .. import DGLGraph
+from .. import ops
+from ..typing import NData, EData, EType
+
+__all__ = ['copy_u', 'copy_v']
+
+#######################################################
+# Edge-wise operators that fetch node data to edges
+#######################################################
+
+def copy_u(g : DGLGraph, x : NData, etype : EType = None) -> EData:
+    """Compute new edge data by fetching from source node data.
+
+    Given an input graph :math:`G(V, E)` (or a unidirectional bipartite graph
+    :math:`G(V_{src}, V_{dst}, E)`) and an input tensor :math:`X`,
+    the operator computes a tensor :math:`Y` storing the new edge data.
+    For each edge :math:`e=(u,v) \\in E`, it computes:
+
+    .. math:
+
+        Y_e = X_u
+
+    Parameters
+    ----------
+    g : DGLGraph
+        The input graph.
+    x : Tensor
+        The tensor storing the source node data. Shape :math:`(|V_{src}|, *)`.
+    etype : str or (str, str, str), optional
+        Edge type. If not specified, the input graph must have
+        only one type of edges.
+
+    Returns
+    -------
+    Tensor
+        The tensor storing the new edge data. Shape :math:`(|E|, *)`.
+
+    Examples
+    --------
+
+    **Homogeneous graph**
+
+    >>> import torch, dgl
+    >>> g = dgl.rand_graph(100, 500)  # a random graph of 100 nodes, 500 edges
+    >>> x = torch.randn(g.num_nodes(), 5)  # 5 features
+    >>> y = dgl.copy_u(g, x)
+    >>> print(y.shape)
+    (500, 5)
+
+    **Heterogeneous graph**
+
+    >>> hg = dgl.heterograph({
+    ...     ('user', 'follow', 'user'): ([0, 1, 2], [2, 3, 4]),
+    ...     ('user', 'like', 'movie'): ([3, 3, 1, 2], [0, 0, 1, 1])
+    ... })
+    >>> x = torch.randn(hg.num_nodes('user'), 5)
+    >>> y = dgl.copy_u(hg, x, etype='like')
+    >>> print(y.shape)
+    (4, 5)
+    """
+    etype_g = g if etype is None else g[etype]
+    return ops.gsddmm(etype_g, "copy_lhs", x, None)
+
+def copy_v(g : DGLGraph, x : NData, etype : EType = None) -> EData:
+    """Compute new edge data by fetching from destination node data.
+
+    Given an input graph :math:`G(V, E)` (or a unidirectional bipartite graph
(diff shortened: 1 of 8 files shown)


python/dgl/mpops/edgewise.py
```diff
@@ -0,0 +1,234 @@
...
+    >>> hg = dgl.heterograph({
+    ...     ('user', 'follow', 'user'): ([0, 1, 2], [2, 3, 4]),
+    ...     ('user', 'like', 'movie'): ([3, 3, 1, 2], [0, 0, 1, 1])
+    ... })
+    >>> x = torch.randn(hg.num_nodes('user'), 5)
+    >>> y = dgl.copy_u(hg, x, etype='like')
+    >>> print(y.shape)
+    (4, 5)
+    """
+    etype_g = g if etype is None else g[etype]
```
frozenbugs, line 64: etype_g is not very descriptive.
etyped_subgraph or subgraph_of_etype?

python/dgl/mpops/edgewise.py
```diff
@@ -0,0 +1,234 @@
...
+    return ops.gsddmm(etype_g, "copy_lhs", x, None)
+
+
+def copy_v(g: DGLGraph, x: NData, etype: EType = None) -> EData:
```
frozenbugs, line 68: ditto.

python/dgl/mpops/edgewise.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+    name = f"u_{op}_v"
+    op_verb = {
+        "add": "adding",
+        "sub": "subtracting",
+        "mul": "multiplying",
+        "div": "dividing",
+        "dot": "dot-product",
+    }
+    docstring = f"""Compute new edge data by {op_verb[op]} the source node data
```
frozenbugs, line 148: nit: 
```suggestion
    docstring = (
        f"""Compute new edge data by {op_verb[op]} the source node data and
        destination node data.
        
        Given an input graph :math:`G(V, E)` blabla blablabla blablabla
        blablabl ....
        """
    )
```

python/dgl/typing.py
```diff
@@ -0,0 +1,25 @@
...
+
+from typing import Tuple, Union
+
+try:
+    import torch
+
+    NData = torch.Tensor
+    EData = torch.Tensor
+except ImportError:
+    # When PyTorch is not available, create two class stubs to represent
```
frozenbugs, line 13: nit: there are still quite a lot of space in this line before hit 80, let's make use of them.

python/dgl/mpops/edgewise.py
frozenbugs, line 64: Is it possible that the DGL graph was create with etype "str", while the input etype is "Tuple[str, str, str]"? What will happen?

tests/pytorch/test_mpops.py
```diff
@@ -0,0 +1,195 @@
+import random
```
frozenbugs, line 1: 1. Will this automatically triggered by CI?
2. I'd suggest to make the unittest file pattern have the strict correlation with the tested files. e.g. python/dgl/mpops/edgewise.py --> tests/python/dgl/mpops/test_edgewise.py

jermainewang: Good idea. We can refactor the unit test frameworks altogether later.

python/dgl/mpops/edgewise.py
jermainewang: changed to etype_subg

python/dgl/mpops/edgewise.py
jermainewang: Triple quote string is designed for multiple lines. The (seemly strange) indentation is also intended to make it look better when users use `help(dgl.u_add_v)`. Example output below:

```
Help on function u_add_v in module dgl.mpops.edgewise:

u_add_v(g, x_node, y_node, etype=None)
    Compute new edge data by adding the source node data
    and destination node data.

    Given an input graph :math:`G(V, E)` (or a unidirectional bipartite graph
    :math:`G(V_{src}, V_{dst}, E)`) and two input tensors :math:`X` and
    :math:`Y`, the operator computes a tensor :math:`Z` storing the new edge data.
    For each edge :math:`e=(u,v) \in E`, it computes:

    .. math:

        Z_e = add(X_u, Y_v)

    If :math:`X_u` and :math:`Y_v` are vectors or high-dimensional tensors, the
    operation is element-wise and supports shape broadcasting. Read more about
    `NumPy's broadcasting semantics
    <https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html>`_.

    Parameters
    ----------
    g : DGLGraph
        The input graph.
    x_node : Tensor
        The tensor storing the source node data. Shape :math:`(|V_{src}|, *)`.
    y_node : Tensor
        The tensor storing the destination node data. Shape :math:`(|V_{dst}|, *)`.
    etype : str or (str, str, str), optional
        Edge type. If not specified, the input graph must have
        only one type of edges.
```

frozenbugs (approved):

tests/pytorch/mpops/test_edgewise.py
```diff
@@ -0,0 +1,201 @@
+import random
```
frozenbugs, line 1: Why do we need pytorch in the path?

jermainewang: Because currently DGL's unit tests are split into two sets:
* All unit tests under `tests/compute` are framework-agnostic unit tests (can run for PyTorch/MXNet/TensorFlow).
* All unit tests under `tests/pytorch` are pytorch-specific unit tests (only runnable for Pytorch).

In the future, as the part of the effort to make DGL more PyTorch-friendly, we should gradually add tests only to `tests/pytorch` and eventually remove the `pytorch` subfolder (dropping the support of the other two).

Issue #4785, opened 2022-10-31
Replace DGL PRNG with a better one


mfbalin:
## 🚀 Feature

Replace std::default_random_engine with a better proven PRNG that passes statistical tests.

## Motivation

I noticed that the DGL PRNG is just STL default random engine, which is just std::minstd_rand0. The set seed function sets the seed of the PRNG by adding the thread id to the seed. But since the minstd_rand0 is not a good quality PRNG, two threads, one with id 0 and one with id 1 for example will have seeds, seed and seed+1. The sequence of numbers generated by these two threads will be correlated with each other, they wouldn't pass the randomness tests available in the literature. I would recommend replacing the underlying random number engine with a better one, pcg is one that DGL could consider. You may ask why this is important, it is important because an experimental result could be worse than it potentially could be silently just because of the bad quality of the PRNG.

More modern PRNGs such as pcg has the capability of having streams, so that different threads can be initialized with same seed, but on a different stream. https://www.pcg-random.org/useful-features.html

Numpy for example has switched to using pcg as their default PRNG for these reasons. The new PRNG doesn't have to be PCG, there are alternatives to it. I just wanted to start a discussion about this topic. https://numpy.org/doc/stable/reference/random/upgrading-pcg64.html

mufeili:
It seems that you added that code snippet. Any thoughts? @classicsong

BarclayII:
> You may ask why this is important, it is important because an experimental result could be worse than it potentially could be silently just because of the bad quality of the PRNG.

I also mentioned in the PR #4668 ; it's better to provide evidence on this.  For instance, how much would the accuracy vary by switching the RNG alone?

mfbalin:
My use case requires using the PRNG as a function of a given random seed and a given integer t, and I expect the resulting uniform random numbers to be uncorrelated with each other for any given pair of random_seed and the integer t.

With pcg32 number of sampled vertices and edges: I seed PRNG with random seed and skip ahead t times with it.
`NS_1024_10 [(215705, 204070, 2034612), (204070, 167180, 1669726), (167180, 69542, 694811), (69542, 10331, 103009), (10331, 1024, 10023)]
LABOR-0_1024_10 [(81841, 57935, 568844), (57935, 35431, 362985), (35431, 17856, 171143), (17856, 6835, 69360), (6835, 1024, 10057)]
[...]

PR #4668, opened 2022-09-30
[Feature] (La)yer-Neigh(bor) sampling implementation

mfbalin:
## Description
CPU&GPU implementation of LABOR sampling, a drop-in replacement to Neighbor Sampling with same fanout hyperparameters. This work was done partly during a summer NVIDIA Devtech AI internship under @nv-dlasalle's mentorship.
arXiv paper can be found here: [arXiv:2210.13339](https://arxiv.org/abs/2210.13339)
To run, either replace any dgl.dataloading.NeighborSampler with dgl.dataloading.LaborSampler, or rebase #4718 onto this PR, and then:
`cd examples/pytorch/labor`
`python train_lightning_labor.py`, to compare with neighbor sampling, `python train_lightning_labor.py --sampler=neighbor`.
With 1000 batch size and 3 layers with 10,10,10 fanout on reddit, can sample 7x less vertices with same quality compared to neighbor sampling with `importance-sampling=-1` option, while `importance-sampling=0` option samples around 5x less and is faster than Neighbor Sampling in general.

Weighted sampling is available too, seems to work as expected.

same_loss_zoom
The loss curve on different datasets with same batch size. The soft edges represent the confidence interval. Number of sampled vertices can be found in Table below:
image
same_budget_ordered
Vertex sampling efficiency under the same sampling budget. A starting batch size of 1k is used and the batch size is adjusted at the end of each epoch to better match the vertex budget. The first row of plots is for the batch size and the second row of plots shows the number of vertices sampled in the last layer, $|V_3|$ of running averages. The last row shows the training loss curves.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage
- [x] Code is well-documented
- [x] To the best of my knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change

PR #4807, opened 2022-11-03
[Feature] replace dgl PRNG with pcg32

mfbalin:
## Description
A way of addressing #4785. Don't have to merge it like this, but can be just merged as well. I don't know if the existing tests rely on the underlying PRNG to pass.

## Changes
Replaced std::default_random_engine with pcg32, passed thread_id as the stream argument of pcg32.

I can close this PR if replacing the PRNG is not on the radar.

### DIFF
--- a/dgl_sparse/include/sparse/sparse_format.h
+++ b/dgl_sparse/include/sparse/sparse_format.h
@@ -81,6 +81,10 @@ std::shared_ptr<CSR> COOToCSC(const std::shared_ptr<COO>& coo);
 /** @brief Convert a CSR format to CSC format. */
 std::shared_ptr<CSR> CSRToCSC(const std::shared_ptr<CSR>& csr);
 
+/** @brief COO transposition. */
+std::shared_ptr<COO> COOTranspose(const std::shared_ptr<COO>& coo);
+
+
 }  // namespace sparse
 }  // namespace dgl
 
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -117,6 +117,11 @@ class SparseMatrix : public torch::CustomClassHolder {
    */
   void SetValue(torch::Tensor value);
 
+  /** @brief Return the transposition of the sparse matrix. It transposes the
+   * first existing sparse format by checking COO, CSR, and CSC.
+   */
+  c10::intrusive_ptr<SparseMatrix> Transpose() const;
+
  private:
   /** @brief Create the COO format for the sparse matrix internally */
   void _CreateCOO();
--- a/dgl_sparse/src/python_binding.cc
+++ b/dgl_sparse/src/python_binding.cc
@@ -24,7 +24,8 @@ TORCH_LIBRARY(dgl_sparse, m) {
       .def("coo", &SparseMatrix::COOTensors)
       .def("csr", &SparseMatrix::CSRTensors)
       .def("csc", &SparseMatrix::CSCTensors)
-      .def("set_val", &SparseMatrix::SetValue);
+      .def("set_val", &SparseMatrix::SetValue)
+      .def("transpose", &SparseMatrix::Transpose);
   m.def("create_from_coo", &CreateFromCOO)
       .def("create_from_csr", &CreateFromCSR)
       .def("create_from_csc", &CreateFromCSC)
--- a/dgl_sparse/src/sparse_format.cc
+++ b/dgl_sparse/src/sparse_format.cc
@@ -86,5 +86,11 @@ std::shared_ptr<CSR> CSRToCSC(const std::shared_ptr<CSR>& csr) {
   return CSRFromOldDGLCSR(dgl_csc);
 }
 
+std::shared_ptr<COO> COOTranspose(const std::shared_ptr<COO>& coo) {
+  auto dgl_coo = COOToOldDGLCOO(coo);
+  auto dgl_coo_tr = aten::COOTranspose(dgl_coo);
+  return COOFromOldDGLCOO(dgl_coo_tr);
+}
+
 }  // namespace sparse
 }  // namespace dgl
--- a/dgl_sparse/src/sparse_matrix.cc
+++ b/dgl_sparse/src/sparse_matrix.cc
@@ -114,6 +114,20 @@ SparseMatrix::CSCTensors() {
 
 void SparseMatrix::SetValue(torch::Tensor value) { value_ = value; }
 
+c10::intrusive_ptr<SparseMatrix> SparseMatrix::Transpose() const {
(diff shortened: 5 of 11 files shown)


frozenbugs:
Hi @mfbalin , thanks very much for your help. In general, as the discussion goes in https://github.com/dmlc/dgl/issues/4785, we don't really see a good reason to add another external dependency for dgl code.
But I do agree we can switch to MT19937, which is reportedly better than minstd_rand0, and it is provided by std.

Would you mind to switch to that in this PR?
Thanks.

mfbalin:
> Hi @mfbalin , thanks very much for your help.
> 
> As the discussion goes in #4785. I agree pcg could be better than MT19937 based on the evidence, but we don't really see a good reason to add another external dependency for dgl code. I do agree we can switch to MT19937, which is reportedly better than minstd_rand0, and it is provided by std.
> 
> Would you mind to switch to that in this PR? Thanks.

Hi, the external dependency for pcg has already been added as part of #4668. If it wants to be removed, then a short implementation header containing pcg can be added and the submodule dependency could be removed. Taking this into account, if DGL still wants to switch to mt19937, I could change the PR.

frozenbugs:
> > Hi @mfbalin , thanks very much for your help.
> > As the discussion goes in #4785. I agree pcg could be better than MT19937 based on the evidence, but we don't really see a good reason to add another external dependency for dgl code. I do agree we can switch to MT19937, which is reportedly better than minstd_rand0, and it is provided by std.
> > Would you mind to switch to that in this PR? Thanks.
> 
> Hi @frozenbugs, the external dependency for pcg has already been added as part of #4668. If it wants to be removed, then a short implementation header containing pcg can be added and the submodule dependency could be removed. Taking this into account, if DGL still wants to switch to mt19937, I could change the PR. Or if the external dependency wants to be removed for sure, I could cook up a short pcg implementation by looking at the official code available. It takes only 20 lines to implement.

Yes, please move to mt19937, and remove the dependency in #4668 if possible. If you have a strong reason to use pcg (e.g. measurable performance improvement on DGL), we can revisit the decision. Since no-one here is an expert of pcg, I'd prefer to not have the pcg implemented since it increase the maintenance cost.

mfbalin:
I have removed the submodule required for #4668 by reimplementing pcg32 in "include/dgl/pcg32.h". I have also replaced the std::default_generator to std::mt19937_64 for RandomEngine. However, the new implementation of pcg32 should also be good to go, I checked and the code compiles when I replace std::default_generator with the implementation I added for pcg32. #4668 seems to just work with new implementation also.

mfbalin:
Because of the way I am using the PRNG in #4668, I can't use mt19937 instead. That is why I had to reimplement pcg32 to remove the submodule dependency because it is required.

src/array/cpu/labor_pick.h
```diff
@@ -24,6 +24,7 @@
 
 #include <dgl/array.h>
 #include <dgl/random.h>
+#include <dgl/pcg32.h>
```
frozenbugs, line 27: Is it possible to use mt19937 in this code?

include/dgl/pcg32.h
```diff
@@ -0,0 +1,72 @@
+/*!
+ *   Copyright 2014 Melissa O'Neill <oneill@pcg-random.org>
```
frozenbugs, line 2: Is it legitimate to just copy like this?

src/array/cpu/labor_pick.h
mfbalin: It would make it much slower, and results wouldn't be theoretically guaranteed to be correct either. That code doesn't use the PRNG in the usual way, which is to seed it once and get numbers. What I do is to keep reseeding before getting a new number.

include/dgl/pcg32.h
mfbalin: I didn't directly copy. I actually basically ported it to c++ but the operator()() is mostly the same, because that is the definition of pcg32 basically. Anybody who implements pcg32 implements operator()() in the same way. For the copyright, I didn't know if I should copyright it for myself or directly use the original author's, decided to do the latter.

mfbalin: You can check the referenced file to see what I am talking about, https://github.com/imneme/pcg-c-basic/blob/master/pcg_basic.c

src/array/cpu/labor_pick.h
frozenbugs: Thanks for educating me on this, do we have any measured results to show how much it is slower?
Also, why a different prng could make it "not theoretically guaranteed to be correct"? Conceptually they are just simulating random number.

If indeed we see the performance different, yes, we should switch to pcg.
Can you revert back to your original change, which is using pcg_random.hpp in include/dgl/random.h. And I am wondering whether you can include include/dgl/random.h in src/array/cpu/labor_pick.h, instead of including pcg32.h / pcg_random.hpp.

Again, thank you for your great and in-depth contribution.

mfbalin: Usually, PRNG's are designed so that when you seed them once, every random number you sample after are independently identically distributed. But in my case, I require that the random numbers one gets by seeding with a different seed, every first random number you sample are independently identically distributed.

mfbalin: In labor_pick.h, replacing
`pcg32(random_seed, t)` line with
```
std::seed_seq seq{(uint32_t)(random_seed >> 32), (uint32_t)random_seed, (uint32_t)(t >> 32), (uint32_t)t};
std::mt19937 ng(seq);
```
Causes 1000x slowdown.

mfbalin: Can't even inspect if results are correct because the code is not terminating, as in comparing to the provided numbers in the original issue #4785.

mfbalin: The reason why it is so slow is that `sizeof(std::mt19937)` returns 5000, so each time it needs to seed, it has to reinitialize a space of kilobytes as opposed to just two numbers as in pcg32.

mfbalin: Also, I haven't run the benchmarks for it but I am certainly sure that the neighbor sampling cpu kernel will become at least 2x slower if RandomEngine switches from std::default_random_engine to std::mt19937/_64.

mfbalin: One still needs to run benchmarks to see how much slower NeighborSampler will become when std::default_random_engine is swapped for pcg32.

mfbalin:
I don't know why but requesting review again removed the other existing reviewer, it was not my intention to do so.

frozenbugs:
@dgl-bot

frozenbugs:
@dgl-bot

frozenbugs:
@dgl-bot

PR #4808, opened 2022-11-03
[Bug] Change the way thread local PRNG is seeded.

mfbalin:
## Description
This is a better way to initalize the PRNG, improving the current implementation. Addresses #4785 partly.

## Changes
Using std::seed_seq with seed and threadid is better than seed + threadid.


BarclayII:
@dgl-bot

BarclayII (approved):

BarclayII:
@dgl-bot

PR #4814, opened 2022-11-04
[Dist] enable access DistGraph.edges via canonical etype

Rhett-Ying:
## Description

we should always use canonical etypes as canonical etype is fully supported in DistDGL.
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/dist_graph.py
+++ b/python/dgl/distributed/dist_graph.py
@@ -157,7 +157,7 @@ class HeteroNodeView(object):
         return NodeSpace(data=NodeDataView(self._graph, key))
 
 class HeteroEdgeView(object):
-    """A EdgeView class to act as G.edges for a DistGraph."""
+    """An EdgeView class to act as G.edges for a DistGraph."""
     __slots__ = ['_graph']
 
     def __init__(self, graph):
--- a/python/dgl/distributed/dist_graph.py
+++ b/python/dgl/distributed/dist_graph.py
@@ -164,10 +164,9 @@ class HeteroEdgeView(object):
         self._graph = graph
 
     def __getitem__(self, key):
-        assert isinstance(key, str) or \
-            (isinstance(key, tuple) and len(key) == 3), \
-            "Expect edge type in string or a triplet of string, but got " \
-            f"{key}."
+        assert isinstance(key, str) or (
+            isinstance(key, tuple) and len(key) == 3
+        ), f"Expect edge type in string or triplet of string, but got {key}."
         return EdgeSpace(data=EdgeDataView(self._graph, key))
 
 class NodeDataView(MutableMapping):
--- a/python/dgl/distributed/dist_graph.py
+++ b/python/dgl/distributed/dist_graph.py
@@ -157,14 +157,17 @@ class HeteroNodeView(object):
         return NodeSpace(data=NodeDataView(self._graph, key))
 
 class HeteroEdgeView(object):
-    """A NodeView class to act as G.nodes for a DistGraph."""
+    """A EdgeView class to act as G.edges for a DistGraph."""
     __slots__ = ['_graph']
 
     def __init__(self, graph):
         self._graph = graph
 
     def __getitem__(self, key):
-        assert isinstance(key, str)
+        assert isinstance(key, str) or \
+            (isinstance(key, tuple) and len(key) == 3), \
+            "Expect edge type in string or a triplet of string, but got " \
+            f"{key}."
         return EdgeSpace(data=EdgeDataView(self._graph, key))
 
 class NodeDataView(MutableMapping):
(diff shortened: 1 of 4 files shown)


tests/distributed/test_dist_graph_store.py
```diff
@@ -730,7 +730,8 @@ def check_dist_graph_hetero(g, num_clients, num_nodes, num_edges):
 
     # Test reading edge data
     eids = F.arange(0, int(g.number_of_edges("r1") / 2))
-    feats1 = g.edges["r1"].data["feat"][eids]
+    c_etype = g.to_canonical_etype("r1")
+    feats1 = g.edges[c_etype].data["feat"][eids]
```
jermainewang, line 734: Is this PR a breaking change? If not, we shouldn't change the unit test but add a new test instead.

python/dgl/distributed/dist_graph.py
```diff
@@ -157,14 +157,17 @@ def __getitem__(self, key):
...
-    """A NodeView class to act as G.nodes for a DistGraph."""
+    """A EdgeView class to act as G.edges for a DistGraph."""
     __slots__ = ['_graph']
 
     def __init__(self, graph):
         self._graph = graph
 
     def __getitem__(self, key):
-        assert isinstance(key, str)
+        assert isinstance(key, str) or \
```
frozenbugs, line 167:
```suggestion
        assert isinstance(key, str) or ( 
            isinstance(key, tuple) and len(key) == 3 
        ), f"Expect edge type in string or a triplet of string, but got {key}."
```

frozenbugs: Let's avoid using \ to change newline.
Other smarter way to increase readability is to add local variable for "isinstance(key, tuple) and len(key) == 3 ".

python/dgl/distributed/graph_services.py
```diff
@@ -638,9 +637,18 @@ def sample_etype_neighbors(
...
-        fanout = F.full_1d(len(g.etypes), fanout, F.int64, F.cpu())
+        fanout = F.full_1d(len(g.canonical_etypes), fanout, F.int64, F.cpu())
     else:
-        fanout = F.tensor([fanout[etype] for etype in g.etypes], dtype=F.int64)
+        etype_ids = {etype: i for i, etype in enumerate(g.canonical_etypes)}
+        fanout_ = [None] * len(g.canonical_etypes)
+        for etype, v in fanout.items():
+            c_etype = g.to_canonical_etype(etype)
+            fanout_[etype_ids[c_etype]] = v
+        assert all(v is not None for v in fanout_), \
```
frozenbugs, line 647: ditto.

tests/distributed/test_dist_graph_store.py
Rhett-Ying: It's not a breaking change. The change I made for existing test is add more tests to cover the change here, not identical to modifying existing one. In specific here, access via `"r1"` is tested in other part in this testcase, so I think it's ok to change the access way here to cover the access via canonical etype.

frozenbugs (approved):
Approve for the good readability.

python/dgl/distributed/dist_graph.py
```diff
@@ -157,14 +157,16 @@ def __getitem__(self, key):
...
+    """A EdgeView class to act as G.edges for a DistGraph."""
     __slots__ = ['_graph']
 
     def __init__(self, graph):
         self._graph = graph
 
     def __getitem__(self, key):
-        assert isinstance(key, str)
+        assert isinstance(key, str) or (
+            isinstance(key, tuple) and len(key) == 3
```
Kh4L, line 168: How about 
```suggestion
            isinstance(key, (tuple, list)) and len(key) == 3
```

Rhett-Ying: canonical etype is always in tuple, should not be a list.

Kh4L: At first I thought that there could be cases there they might be created by the user, but that makes sense

tests/distributed/test_dist_graph_store.py
Rhett-Ying: accessing via etype and canonical etype are both added.

python/dgl/distributed/graph_services.py
```diff
@@ -638,9 +637,19 @@ def sample_etype_neighbors(
         A sampled subgraph containing only the sampled neighboring edges.  It is on CPU.
     """
     if isinstance(fanout, int):
-        fanout = F.full_1d(len(g.etypes), fanout, F.int64, F.cpu())
+        fanout = F.full_1d(len(g.canonical_etypes), fanout, F.int64, F.cpu())
     else:
-        fanout = F.tensor([fanout[etype] for etype in g.etypes], dtype=F.int64)
+        etype_ids = {etype: i for i, etype in enumerate(g.canonical_etypes)}
+        fanout_ = [None] * len(g.canonical_etypes)
```
jermainewang, line 643: fanout_ -> fanout_array  . The name is too similar to `fanout` otherwise.

PR #4826, opened 2022-11-07
[Model] Add `dgl.nn.CuGraphRelGraphConv` model

### LINKED ISSUE #4669: [FEA] Create RGCN model using cugraph-ops aggregators

tingyu66:
## Description

This PR adds a RGCN model, `dgl.nn.CuGraphRelGraphConv`, that uses the accelerated sparse aggregation primitives in cugraph-ops. The model requires `pylibcugraphops>=22.12`.

## Checklist
- [x] Code is well-documented

## Changes
- [x] New nn.Module: `dgl.nn.CuGraphRelGraphConv`
- [x] Add test that verifies the results against `dgl.nn.RelGraphConv`
- [x] Update API docs 

### Accuracy: Entity Classification

test accuracy after 100 epochs, using 20 bases

| Dataset   |  Mini-batch
| ------------ |  ------
| aifb         | ~0.92
| mutag     | ~0.58
| bgs         | ~0.70
| am          | ~0.64

## Performance
Hardware: 32-core AMD Ryzen Threadripper Pro 3975WX, NVIDIA RTX A6000
Benchmark: Mini-batch `entity_sample.py` example:
- batch_size: 100, graph idtype is int64.
- 2-layer RGCN, with fanouts: (4,4), hidden embedding dimension: 16, num_bases: 20
- The timings are for 100 epochs and the average across 3 identical runs.

|       | time_dgl (s) | time_cugraph-ops (s) |
|:-----:|:------------:|:--------------------:|
|  aifb |     4.24     |         2.61         |
| mutag |     5.65     |         3.68         |
|  bgs  |     4.19     |         2.70         |
|   am  |     19.11    |         13.71        |

During forward aggregation, most of the time (~70%) is spent on the conversion from DGL graph (in COO format) to cugraph-ops MFG type (in CSR format), indicated by `get_csr` and `create_mfg` entries in the following output. The actual forward aggregation function (agg_fwd_func) only takes 1/5 of the total time. 

```
Total Time (ns)  Instances   Avg (ns)     Med (ns)    Min (ns)  Max (ns)   StdDev (ns)   Style       Range    
---------------  ---------  -----------  -----------  --------  ---------  -----------  -------  -------------
    834,193,721        814  1,024,808.0  1,015,957.0   957,171  2,264,271     96,833.9  PushPop  total_agg_fwd
    395,471,358        814    485,837.0    474,488.0   453,493  1,577,778     78,678.5  PushPop  create_mfg   
    182,373,749        814    224,046.4    233,108.5   195,217    398,820     22,571.4  PushPop  get_csr      
    159,149,002        814    195,514.7    193,233.5   185,399    488,559     13,795.3  PushPop  agg_fwd_func
```
Hence, if DGL natively supports CSR representation of graphs, we can expect a much greater speed-up from using `dgl.nn.CuGraphRelGraphConv`.

### Notes

- Partially resolves #4669 
- This PR supersedes #4701
- Link to internal reviews https://github.com/rapidsai/dgl/pull/15.
- Closes https://github.com/rapidsai/cugraph-ops/issues/179


frozenbugs:
@dgl-bot

frozenbugs:
@dgl-bot

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+                f"but got {g.idtype}"
+            )
+        ctx.idtype = g.idtype
+
+        _in_feat = feat.shape[-1]
+        indptr, indices, edge_ids = g.adj_sparse("csc")
+        # edge_ids is in a mixed order, need to permutate incoming etypes
+        # and stash the result for backward propagation
+        ctx.edge_types_int32 = edge_types[edge_ids.long()].int()
+        # node_types are not needed by the post-variant RGCN aggregation functions
```
frozenbugs, line 82: over 80 chars.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        else:
+            raise TypeError(
+                f"Supported ID type: torch.int32 or torch.int64, "
+                f"but got {g.idtype}"
+            )
+        ctx.idtype = g.idtype
+
+        _in_feat = feat.shape[-1]
+        indptr, indices, edge_ids = g.adj_sparse("csc")
+        # edge_ids is in a mixed order, need to permutate incoming etypes
```
frozenbugs, line 79: For all comments, start with capital and end with dot.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+    r"""An accelerated relational graph convolution layer using the
+    highly-optimized aggregation primitives in cugraph-ops.
+    See :class:`dgl.nn.pytorch.conv.RelGraphConv` for mathematical model.
+
+    .. note::
+        This is an **experimental** feature.
+        Compared with :class:`dgl.nn.pytorch.conv.RelGraphConv`, this model:
+
+        * Only works on cuda devices.
+        * Only support basis-decomposition regularization.
```
frozenbugs, line 171: nit: support -> supports

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
+"""Torch Module for Relational graph convolution layer using the
```
frozenbugs, line 1: nit: make use of all the space. There are some room before hitting 80 chars.

frozenbugs: """Torch module for Relational Graph Convolution Network layer aggregation

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+    from pylibcugraphops.legacy.structure.graph_types import (
+        message_flow_graph_hg_csr_int32,
+        message_flow_graph_hg_csr_int64,
+    )
+except ModuleNotFoundError:
+    raise ModuleNotFoundError(
+        "dgl.nn.CuGraphRelGraphConv requires pylibcugraphops to be installed."
+    )
+
+CUDA_SM_PER_BLOCK = 49152
```
frozenbugs, line 25: Maybe add a comment and link to a documentation about this magic number? If possible.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+            A 1D tensor of edge types. Note that pylibcugraphops only accepts
+            edge_types in int32, so any input of other integer types will be casted
+            into int32, thus introducing some overhead. Pass in int32 tensor for
+            best performance.
+        feat : torch.Tensor, dtype=torch.float32, requires_grad=True
+            Tensor of source node features. Shape: (num_src_nodes, in_feat).
+        coeff : torch.Tensor, dtype=torch.float32, requires_grad=True
+            The coefficient matrix used in basis-decomposition regularization.
+            Shape: (num_rels, num_bases). It should be set to ``None``
+            when ``regularizer=None``.
```
frozenbugs, line 54: I am a little bit confused here. What will happen in regularizer is not "None" and coeff is also not "None"?

The CuGraphRelGraphConv calls RelGraphConvAgg, and regularizer is a property in CuGraphRelGraphConv. From the code structure POV, it is a little bit weird that a variable in a callee has a restriction depends on caller.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        ctx.device = g.device
+        if g.idtype == th.int32:
+            mfg_csr_func = message_flow_graph_hg_csr_int32
+            agg_fwd_func = agg_hg_basis_post_fwd_int32
+        elif g.idtype == th.int64:
+            mfg_csr_func = message_flow_graph_hg_csr_int64
+            agg_fwd_func = agg_hg_basis_post_fwd_int64
+        else:
+            raise TypeError(
+                f"Supported ID type: torch.int32 or torch.int64, "
```
frozenbugs, line 72: nit: make use of all the space. There are some room before hitting 80 chars.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+                feat.detach(),
+                ctx.mfg,
+                output_weight_gradient=grad_coeff,
+                weights_combination=coeff.detach()
+            )
+
+        return None, None, None, None, grad_feat, grad_coeff
+
+
+class CuGraphRelGraphConv(nn.Module):
```
mufeili, line 161: Any instructions on installing the dependencies? I remember CuGraph only support a subset of CUDA versions. Is there a easy recipe for installing all these dependencies?

mufeili:
The four datasets you mentioned are all pretty small. Have you tested on `ogbn-mag`?

mufeili:
What are the reference accuracy numbers for "Accuracy: Entity Classification"? It's not clear how good/bad these numbers are without a reference.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+
+CUDA_SM_PER_BLOCK = 49152
+
+
+class RelGraphConvAgg(th.autograd.Function):
+    r"""Custom autograd function for R-GCN aggregation layer that uses the
+    aggregation functions in cugraph-ops."""
+
+    @staticmethod
+    def forward(ctx, g, fanout, num_rels, edge_types, feat, coeff):
```
mufeili, line 33: What is `ctx`? Why did you not document it in the `Parameters` section below?

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
mufeili, line 161: Can we change from `pytorch/conv/cugraph_relgraphconv.py/CuGraphRelGraphConv` to `pytorch/conv/cugraph/relgraphconv.py/RelGraphConv`? I just find this name a bit tedious. Also we might want to handle all dependency checks for cugraph once in the same subdirectory. @jermainewang

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+                weights_combination=coeff.detach()
+            )
+
+        return None, None, None, None, grad_feat, grad_coeff
+
+
+class CuGraphRelGraphConv(nn.Module):
+    r"""An accelerated relational graph convolution layer using the
+    highly-optimized aggregation primitives in cugraph-ops.
+    See :class:`dgl.nn.pytorch.conv.RelGraphConv` for mathematical model.
```
mufeili, line 164: Instead of pointing to another NN module, I recommend either pointing to the paper or enriching the description.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+
+    .. note::
+        This is an **experimental** feature.
```
mufeili, line 167: Add an empty line above

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+            An 1D integer tensor of edge types. Shape: :math:`(|E|,)`.
+        norm : torch.Tensor, optional
+            An 1D tensor of edge norm value.  Shape: :math:`(|E|,)`.
+
+        Returns
+        -------
+        torch.Tensor
+            New node features. Shape: :math:`(|V|, D_{out})`.
+        """
+        _device = next(self.parameters()).device
```
jermainewang, line 335: Can also use `g.device`.

jermainewang (approved):
Thanks for the great effort! I'm actually quite curious about the underlying implementation. Do you have plan to write a tech blog about this? For example, I'm a little bit surprised that `RelGraphConvAgg.apply` does not need the weight matrix `self.W`. The multiplication can be deferred to after the message passing (your L377). By contrast, our implementation needs to handle it during message passing using the `TypedLinear` module. Quite curious about why those two are equivalent though your unit tests have demonstrated it.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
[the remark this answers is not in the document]
jermainewang: Let's keep it. This is to lower user's expectation about the module, e.g., the API may be changed so use it with discretion.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
jermainewang: What's the confusion here?

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+class CuGraphRelGraphConv(nn.Module):
+    r"""An accelerated relational graph convolution layer using the
+    highly-optimized aggregation primitives in cugraph-ops.
+    See :class:`dgl.nn.pytorch.conv.RelGraphConv` for mathematical model.
+
+    .. note::
+        This is an **experimental** feature.
+        Compared with :class:`dgl.nn.pytorch.conv.RelGraphConv`, this model:
```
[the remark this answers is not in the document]
jermainewang: I think this module is a special case that it is a direct optimization on our `RelGraphConv`. We can expect the users to understand the original `RelGraphConv` first, so making docstring self-contained is less a concern.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+
+        * Only works on cuda devices.
+        * Only support basis-decomposition regularization.
+        * Requires an extra argument `fanout` as input, since the current model is
+          designed for sampled-graph (message-flow-graph) use-cases.
+          Full-graph support will be added in upcoming releases.
```
mufeili, line 174: I suggest replacing L172-174 with

> Only supports message flow graphs

- You have a dedicated description of the `fanout` argument below, so you do not need to mention that here.
- "Full-graph support will be added in upcoming releases." conveys little information and it's better to not mention that at all. If users want that, they will open a feature request.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
mufeili: First, we have dependency checks like 

```
try:
    from pylibcugraphops.legacy.aggregators.node_level import (
        agg_hg_basis_post_bwd_int32,
        agg_hg_basis_post_bwd_int64,
        agg_hg_basis_post_fwd_int32,
        agg_hg_basis_post_fwd_int64,
    )
    from pylibcugraphops.legacy.structure.graph_types import (
        message_flow_graph_hg_csr_int32,
        message_flow_graph_hg_csr_int64,
    )
except ModuleNotFoundError:
    raise ModuleNotFoundError(
        "dgl.nn.CuGraphRelGraphConv requires pylibcugraphops to be installed."
    )
```

If we have a dedicated subdirectory for CuGraph, then it can be done there rather than in individual files.

Second, I think it's better to do

```
from dgl.nn.cugraph import AConv, BConv, ...
```

than

```
from dgl.nn import CuGraphAConv, CuGraphBConv
```

These are minor issues. I'm fine to wait for a few more modules to see what will happen.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+    Parameters
+    ----------
+    in_feat : int
+        Input feature size.
+    out_feat : int
+        Output feature size.
+    num_rels : int
+        Number of relations.
+    fanout : int
+        Maximum number of sampled neighbors of an destination node;
```
mufeili, line 185:
```suggestion
        Maximum number of sampled neighbors of a destination node,
```

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        i.e, maximum in degree of destination nodes
```
mufeili, line 186:
```suggestion
        i.e. maximum in degree of destination nodes
```

This is not necessarily the case and might be dependent on the sampler configuration, right?

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+
+        * Only works on cuda devices.
+        * Only support basis-decomposition regularization.
+        * Requires an extra argument `fanout` as input, since the current model is
+          designed for sampled-graph (message-flow-graph) use-cases.
+          Full-graph support will be added in upcoming releases.
+
+    Parameters
+    ----------
+    in_feat : int
```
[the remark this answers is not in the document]
mufeili: `in_feat` and `out_feat` are currently used by `dgl.nn.pytorch.conv.RelGraphConv`. Do we want to keep the consistency?

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        Input feature size.
+    out_feat : int
+        Output feature size.
+    num_rels : int
+        Number of relations.
+    fanout : int
```
mufeili, line 184: Maybe move this argument to the beginning or the end for consistency of argument order with `dgl.nn.pytorch.conv.RelGraphConv`.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        Maximum number of sampled neighbors of an destination node;
+        i.e, maximum in degree of destination nodes
+    regularizer : str, optional
+        Which weight regularizer to use ("basis" or ``None``):
+         - "basis" is for basis-decomposition.
+         - ``None`` applies no regularization.
+        Default: ``None``.
+    num_bases : int, optional
+        Number of bases. Only used when "basis" regularizer is applied.
```
mufeili, line 193:
```suggestion
        Number of bases. It comes into effect when the "basis" regularizer is applied. Otherwise, it uses number of relations (``num_rels``).
```

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        Default: ``None``.
+    bias : bool, optional
+        True if bias is added. Default: ``True``.
+    activation : callable, optional
+        Activation function. Default: ``None``.
+    self_loop : bool, optional
+        True to include self loop message. Default: ``True``.
+    dropout : float, optional
+        Dropout rate. Default: ``0.0``
+    layer_norm: bool, optional
```
mufeili, line 203:
```suggestion
    layer_norm : bool, optional
```

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        True to add layer norm. Default: ``False``
+
+    Examples
+    --------
+    >>> import dgl
+    >>> from dgl.dataloading import NeighborSampler, DataLoader
+    >>> from dgl.nn import CuGraphRelGraphConv
+    >>> import torch.nn.functional as F
+    ...
```
mufeili, line 212: This usage can be confused with line continuation.

mufeili: Can we use a built-in dataset for demonstration so that users can really run the code snippet?

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+            if num_bases is None:
+                raise ValueError(
+                    'Missing "num_bases" for basis regularization.'
+                )
+            self.W = nn.Parameter(th.Tensor(num_bases, in_feat, out_feat))
+            self.coeff = nn.Parameter(th.Tensor(num_rels, num_bases))
+            self.num_bases = num_bases
+        else:
+            raise ValueError(
+                f"Supported regularizer options: 'basis', "
```
mufeili, line 261: Also None?

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        # weight for self_loop
+        if self.self_loop:
+            self.loop_weight = nn.Parameter(th.Tensor(in_feat, out_feat))
+            nn.init.xavier_uniform_(
+                self.loop_weight, gain=nn.init.calculate_gain("relu")
+            )
+
+        self.dropout = nn.Dropout(dropout)
+
+    def reset_parameters(self):
```
mufeili, line 293: Typically `reset_parameters` allows users to reset all learnable parameters while this currently only handles `self.coeff` and `self.W`. 

- We can either do these directly in `__init__` if this is only used by `__init__`
- Or expand it to cover all learnable parameters

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+    def forward(self, g, feat, etypes, norm=None):
+        r"""Forward computation.
+
+        Parameters
+        ----------
+        g : DGLGraph
+            The graph.
+        feat : torch.Tensor
+            A 2D tensor of node features. Shape: :math:`(|V|, D_{in})`.
+        etypes : torch.Tensor or list[int]
```
mufeili, line 325: Only `torch.Tensor` is supported.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+            An 1D integer tensor of edge types. Shape: :math:`(|E|,)`.
```
mufeili, line 326:
```suggestion
            A 1D integer tensor of edge types. Shape: :math:`(|E|,)`.
```

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        norm : torch.Tensor, optional
+            An 1D tensor of edge norm value.  Shape: :math:`(|E|,)`.
```
mufeili, line 328: Ditto. The three minor issues also exist in `dgl.nn.pytorch.conv.relgraphconv`. Could you also help fix these?

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        Returns
+        -------
+        torch.Tensor
+            New node features. Shape: :math:`(|V|, D_{out})`.
+        """
+        _device = next(self.parameters()).device
+        if _device.type != "cuda":
+            raise RuntimeError(
+                f"dgl.nn.CuGraphRelGraphConv requires "
+                f"the model on device 'cuda', but got '{_device.type}'"
```
mufeili, line 339:
```suggestion
                f"the model to be on device 'cuda', but got '{_device.type}'"
```

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        if _device != etypes.device:
+            raise RuntimeError(
+                f"Expected model and etypes on the same device, "
+                f"but got '{_device}' and '{etypes.device}'"
+            )
+        if _device != feat.device:
+            raise RuntimeError(
+                f"Expected model and feature tensor on the same device, "
+                f"but got '{_device}' and '{feat.device}'"
+            )
```
mufeili, line 355: We don't do checks in L341-355 in the existing DGL NN modules. I'm not sure if they are needed.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        # check shared memory per block size
+        _sm_size = (
+            self.fanout * 2 + 3 + self.num_rels * self.num_bases
```
mufeili, line 358: Briefly explain where do `2` and `3` come from?

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+            The graph.
+        fanout : int
+            Maximum in-degree of nodes.
+        num_rels : int
+            Number of relations.
+        edge_types : torch.Tensor
+            A 1D tensor of edge types. Note that pylibcugraphops only accepts
+            edge_types in int32, so any input of other integer types will be casted
+            into int32, thus introducing some overhead. Pass in int32 tensor for
+            best performance.
```
mufeili, line 48: This is not visible to users. Perhaps you should modify the description of argument `etypes` in `CuGraphRelGraphConv`.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        feat : torch.Tensor, dtype=torch.float32, requires_grad=True
+            Tensor of source node features. Shape: (num_src_nodes, in_feat).
```
mufeili, line 50: Again, there can be inconsistencies between parameter description in this class and that in `CuGraphRelGraphConv`.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        coeff : torch.Tensor, dtype=torch.float32, requires_grad=True
```
mufeili, line 51: You can omit `dtype=torch.float32, requires_grad=True` since this is created in `CuGraphRelGraphConv`.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+            The coefficient matrix used in basis-decomposition regularization.
+            Shape: (num_rels, num_bases). It should be set to ``None``
+            when ``regularizer=None``.
+
+        Returns
+        -------
+        agg_output : torch.Tensor, dtype=torch.float32
```
mufeili, line 58: No need to specify `dtype`

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+            Aggregation output. Shape: (num_dst_nodes, num_rels * in_feat)
+            when ``regularizer=None``; Shape: (num_dst_nodes, num_bases * in_feat)
+            when ``regularizer='basis'``.
+        """
+        ctx.device = g.device
```
mufeili, line 63: This seems redundant since you already assume that in `CuGraphRelGraphConv`.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
mufeili, line 82: What does "post-variant" mean here?

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,388 @@
...
+        return agg_output
+
+    @staticmethod
+    def backward(ctx, grad_output):
+        r"""Compute the backward pass of R-GCN aggregation layer.
+
+        Parameters
+        ----------
+        grad_output : torch.Tensor, dtype=torch.float32
+            Gradient of loss function w.r.t output.
```
mufeili, line 132: Ditto for `ctx` and removing dtype of `grad_output`.

tests/cugraph/cugraph-ops/test_cugraph_relgraphconv.py
```diff
@@ -0,0 +1,90 @@
...
+    u = torch.tensor([0, 1, 0, 2, 3, 0, 4, 0, 5, 0, 6, 7, 0, 8, 9])
+    v = torch.tensor([1, 9, 2, 9, 9, 4, 9, 5, 9, 6, 9, 9, 8, 9, 0])
+    g = dgl.graph((u, v))
+    num_rels = 3
+    g.edata[dgl.ETYPE] = torch.randint(num_rels, (g.num_edges(),))
+    return g
+
+
+@pytest.mark.parametrize("regularizer", regularizers)
+def test_full_graph(regularizer):
```
mufeili, line 21: This is not supported, right?

mufeili (changes requested):
done a first pass

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: I have left some explanation there. We are working on an improvement in the next release of cugraph-ops so that this memory allocation size check actually happens in cugraph-ops.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: I have updated the doc to avoid ambiguity here. `RelGraphConvAgg` does not use the `RelGraphConvAgg` variable at all, as the regularization information is carried by coeff itself.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: Done. Have fixed all the formatting issues.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: That's a good point. Unlike what its name suggests, this module does not depend on cugraph. It only requires the `pylibcugraphops` package, which can be installed via `conda install -c nvidia pylibcugraphops=22.12` after its release.
It works with any CUDA toolkit version >= 11.

Where do you suggest to add this information?

tingyu66:
> What are the reference accuracy numbers for "Accuracy: Entity Classification"? It's not clear how good/bad these numbers are without a reference.

Reference accuracy obtained from using `dgl.nn.RelGraphConv` is summarized [here](https://github.com/dmlc/dgl/blob/master/examples/pytorch/rgcn/README.md). 

The accuracy of the new model is slightly better than the reference value. I think that is because `num_bases` is set to `num_rels` in the current DGL example.
https://github.com/dmlc/dgl/blob/ed8e9c4440c204d19a138bcc552ee476c3a7cbc1/examples/pytorch/rgcn/entity_sample.py#L17

Is that simply an oversight?

tingyu66:
> The four datasets you mentioned are all pretty small. Have you tested on `ogbn-mag`?

Not yet, I can have a try maybe later this week. Does `ogbn-mag` require extra preprocessing steps? (or is there a current example that I can refer to)

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: ctx is a context object that can be used to stash information for backward computation in pytorch.
I did not add it to docstring since ctx is not something that has to be passed in by users.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: Fixed.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: dtype removed. Regarding ctx, see comments above.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: This was some notes related to the APIs in cugraph-ops. I have cleaned up the comments.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: Fixed.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: Agree. This is redundant.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: The way we compute _sm_size (cuda shared memory allocation size) is related to our cuda kernel implementation in cugraph-ops, so I think explain the computation is kind of irrelevant here. 

As I mentioned above, we are working on an improvement in the next release of cugraph-ops so that this size check actually happens in cugraph-ops.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: `_device` denotes the device that the model (parameters) is on. It technically can be different from g.device (when users forget to copy one of them to GPU).

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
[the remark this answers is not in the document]
tingyu66: Fixed.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: This argument should be renamed as max_in_degree, as Minjie suggested.
If we use `NeighborSampler` to generate sampled graphs, max_in_degree should be equivalent to the fanout used in the sampler, correct?

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: Done, and it becomes optional now.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: > Otherwise, it uses number of relations (``num_rels``).

This is not the case in the current setup. The value is not being used when regularizer is None. I did add "It comes into effect when the "basis" regularizer is applied." for consistency with RelGraphConv.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: You are right. I updated the example with something similar to the one in RelGraphConv.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: Fixed.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: They are actually needed. cugraph-ops consists of (almost) solely computational primitives and does not contain those checks as below in DGL,

https://github.com/dmlc/dgl/blob/e3bf18255110bffe85fccdcafcb551436e173dad/python/dgl/heterograph.py#L4121-L4124

When graphs or tensors are not properly copied to cuda devices, users will see segfault without these checks.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: Fixed

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
```diff
@@ -0,0 +1,395 @@
...
+        with g.local_scope():
+            g.srcdata["h"] = feat
+            if norm is not None:
+                g.edata["norm"] = norm
+            # message passing
+            h = RelGraphConvAgg.apply(
+                g, self.num_rels, etypes, max_in_degree, feat, self.coeff
+            )
+            h = h @ self.W.view(-1, self.out_feat)
+            # apply bias and activation
```
frozenbugs, line 385: nit: ditto.

frozenbugs (approved):

tingyu66:
> Thanks for the great effort! I'm actually quite curious about the underlying implementation. Do you have plan to write a tech blog about this? For example, I'm a little bit surprised that `RelGraphConvAgg.apply` does not need the weight matrix `self.W`. The multiplication can be deferred to after the message passing (your L377). By contrast, our implementation needs to handle it during message passing using the `TypedLinear` module. Quite curious about why those two are equivalent though your unit tests have demonstrated it.

I appreciate the prompt review from the DGL team. The aggregation operation in RGCN is either sum or mean, and their linearity allows us to swap the order of operations. Instead of generating messages (applying weights) first and aggregating them later, we can aggregate/scatter features first to form a intermediate embedding (with num_rels * in_feat) for each destination node, and apply weights via dense gemm after that. We are totally open to work on a blog post after adding in full-graph support (to make it feature-complete).

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: great suggestion. The initialization calls now go into __init__.

frozenbugs:
@dgl-bot

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
mufeili: For now, you can simply add these two requirements in the doc of `CuGraphRelGraphConv`.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
mufeili: This class is not exposed to users. Meanwhile, it's still better to have a brief description for future developers.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
mufeili: I think so.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
mufeili: I see. Thanks.

mufeili:
> > What are the reference accuracy numbers for "Accuracy: Entity Classification"? It's not clear how good/bad these numbers are without a reference.
> 
> Reference accuracy obtained from using `dgl.nn.RelGraphConv` is summarized [here](https://github.com/dmlc/dgl/blob/master/examples/pytorch/rgcn/README.md).
> 
> The accuracy of the new model is slightly better than the reference value. I think that is because `num_bases` is set to `num_rels` in the current DGL example.
> 
> https://github.com/dmlc/dgl/blob/ed8e9c4440c204d19a138bcc552ee476c3a7cbc1/examples/pytorch/rgcn/entity_sample.py#L17
> 
> Is that simply an oversight?

I think this was intended.

mufeili:
> > The four datasets you mentioned are all pretty small. Have you tested on `ogbn-mag`?
> 
> Not yet, I can have a try maybe later this week. Does `ogbn-mag` require extra preprocessing steps? (or is there a current example that I can refer to)

You do not need to do that for this PR. For an example, see https://github.com/dmlc/dgl/tree/master/examples/pytorch/ogb/ogbn-mag

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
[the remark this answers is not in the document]
jermainewang: I see. We could table the renaming later.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: Instruction added.

python/dgl/nn/pytorch/conv/cugraph_relgraphconv.py
tingyu66: Fixed.

tingyu66:
I have made the suggested changes, including:

- Removed cuda shared memory size check (this is now handled in pylibcugraphops)
- Renamed `fanout` -> `max_in_degree` and make it optional. CuGraphRelGraphConv now has an (almost) identical signature as RelGraphConv
- Improve documentation and formatting

Feel free to give it a go. Thanks!

jermainewang:
@dgl-bot

mufeili (approved):

mufeili:
@dgl-bot

tingyu66:
@dgl-bot

mufeili:
@dgl-bot

tingyu66:
CI still fails because the cugraph docker image used in CI was outdated. We need to fix a version incompatibility issue internally before push a new image. Will update once that is done.

frozenbugs:
@dgl-bot

tingyu66:
Now, `test_cugraph_relgraphconv.py` have passed. CI failed due to other reasons.

mufeili:
@dgl-bot

PR #4875, opened 2022-11-10
[Sparse] SIGN Example

mufeili:
## Description

- SIGN example using `mock_sparse`, with accuracy around 0.74~0.78
- Fix two minor issues within `mock_sparse` that occur with gpu usage

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/examples/sparse/sign.py
+++ b/examples/sparse/sign.py
@@ -1,10 +1,10 @@
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
-
-from torch.optim import Adam
 from dgl.data import CoraGraphDataset
-from dgl.mock_sparse import create_from_coo, identity, diag
+from dgl.mock_sparse import create_from_coo, diag, identity
+from torch.optim import Adam
+
 
 def sign(A, X, num_hops):
     X_sign = [X]
@@ -13,21 +13,21 @@ def sign(A, X, num_hops):
         X_sign.append(X)
     return X_sign
 
+
 class Model(nn.Module):
     def __init__(self, in_size, num_hops, num_classes, hidden_size=256):
         super().__init__()
 
         num_feats = num_hops + 1
-        self.linear = nn.ModuleList([
-            nn.Linear(in_size, hidden_size) for _ in range(num_feats)
-        ])
+        self.linear = nn.ModuleList(
+            [nn.Linear(in_size, hidden_size) for _ in range(num_feats)]
+        )
         self.pred = nn.Linear(hidden_size * num_feats, num_classes)
 
     def forward(self, X):
-        X = torch.cat([
-            self.linear[i](X[i]) for i in range(len(X))
-        ], dim=1)
+        X = torch.cat([self.linear[i](X[i]) for i in range(len(X))], dim=1)
         return self.pred(F.relu(X))
+
 
 def evaluate(X, Y, model):
     model.eval()
@@ -36,7 +36,8 @@ def evaluate(X, Y, model):
         _, Y_hat = torch.max(logits, dim=1)
         return (Y_hat == Y).float().mean().item()
 
-dev = torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')
+
+dev = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
 
 data = CoraGraphDataset()
 g = data[0].to(dev)
@@ -75,7 +76,7 @@ for epoch in range(10):
     loss.backward()
     optimizer.step()
     train_acc = evaluate(X_sign_train, Y_train, model)
-    print(f'epoch {epoch} | train acc {train_acc}')
+    print(f"epoch {epoch} | train acc {train_acc}")
 
 test_acc = evaluate(X_sign_test, Y_test, model)
-print(f'test acc {test_acc}')
+print(f"test acc {test_acc}")
--- a/examples/sparse/sign.py
+++ b/examples/sparse/sign.py
@@ -47,7 +47,7 @@ A = create_from_coo(dst, src, shape=(N, N))
 
 I = identity(A.shape, device=dev)
 A_hat = A + I
-D_hat = diag(A.sum(dim=1)) ** -0.5
+D_hat = diag(A_hat.sum(dim=1)) ** -0.5
 A_hat = D_hat @ A_hat @ D_hat
 
 X = g.ndata["feat"]
--- a/examples/sparse/sign.py
+++ b/examples/sparse/sign.py
@@ -1,0 +1,81 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from torch.optim import Adam
+from dgl.data import CoraGraphDataset
+from dgl.mock_sparse import create_from_coo, identity, diag
+
+def sign(A, X, num_hops):
+    X_sign = [X]
+    for _ in range(num_hops):
+        X = A @ X
+        X_sign.append(X)
+    return X_sign
+
+class Model(nn.Module):
+    def __init__(self, in_size, num_hops, num_classes, hidden_size=256):
+        super().__init__()
+
+        num_feats = num_hops + 1
+        self.linear = nn.ModuleList([
+            nn.Linear(in_size, hidden_size) for _ in range(num_feats)
+        ])
+        self.pred = nn.Linear(hidden_size * num_feats, num_classes)
+
+    def forward(self, X):
+        X = torch.cat([
+            self.linear[i](X[i]) for i in range(len(X))
+        ], dim=1)
+        return self.pred(F.relu(X))
+
+def evaluate(X, Y, model):
+    model.eval()
+    with torch.no_grad():
+        logits = model(X)
+        _, Y_hat = torch.max(logits, dim=1)
+        return (Y_hat == Y).float().mean().item()
+
+dev = torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')
+
+data = CoraGraphDataset()
+g = data[0].to(dev)
+src, dst = g.edges()
+N = g.num_nodes()
+A = create_from_coo(dst, src, shape=(N, N))
+
+I = identity(A.shape, device=dev)
+A_hat = A + I
+D_hat = diag(A.sum(dim=1)) ** -0.5
+A_hat = D_hat @ A_hat @ D_hat
+
+X = g.ndata["feat"]
+Y = g.ndata["label"]
+train_mask = g.ndata["train_mask"]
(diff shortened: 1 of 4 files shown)


czkkkkkk:
1. I think this example will use our C++ sparse library in the end. So do we need to add a TODO to indicate we need to test this example after we have the C++ sparse library ready?
2. A README file would help to understand this example.

examples/sparse/sign.py
```diff
@@ -0,0 +1,82 @@
...
+        return (Y_pred == Y).float().mean().item()
+
+
+dev = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+
+data = CoraGraphDataset()
+g = data[0].to(dev)
+src, dst = g.edges()
+N = g.num_nodes()
+A = create_from_coo(dst, src, shape=(N, N))
```
frozenbugs, line 46: In the paper W was used to represent adjacency matrix.
I am not sure about whether the community convention is to use A for adjacency matrix.
Here are 2 suggestions I would have.
1. Use the annotation in the paper directly in our code, and add a comment say, this is not dgl convention, but we are follow the paper.
2. Use the convention, but add a comment to clarify the annotation in the paper was W.

examples/sparse/sign.py
```diff
@@ -0,0 +1,82 @@
+import torch
```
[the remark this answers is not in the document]
frozenbugs: Also mention in this example, we are implementing a 2 hop sign.

frozenbugs, line 1: https://quip-amazon.com/OWYyAbVeGhBz/DGL-Examples

For the path, let's use core/sparse/sign(_pytorch?).py

examples/sparse/sign.py
```diff
@@ -0,0 +1,82 @@
...
+        super().__init__()
+
+        num_feats = num_hops + 1
+        self.linear = nn.ModuleList(
+            [nn.Linear(in_size, hidden_size) for _ in range(num_feats)]
+        )
+        self.pred = nn.Linear(hidden_size * num_feats, num_classes)
+
+    def forward(self, X):
+        X = torch.cat([self.linear[i](X[i]) for i in range(len(X))], dim=1)
```
frozenbugs, line 28: Nesting this together made it hard to read (for me).
Not sure we can unnest the code and add comment to guide readers to understand the code better.

mufeili:
> 1. I think this example will use our C++ sparse library in the end. So do we need to add a TODO to indicate we need to test this example after we have the C++ sparse library ready?

We should. Any suggestion on how to do that?

> 2. A README file would help to understand this example.

Added

examples/sparse/sign.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def evaluate(X, Y, model):
+    model.eval()
+    with torch.no_grad():
+        logits = model(X)
+        _, Y_pred = torch.max(logits, dim=1)
+        return (Y_pred == Y).float().mean().item()
+
+
```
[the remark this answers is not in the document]
mufeili: If users need to use a different device, they can simply edit L40.

I aim to develop an extremely simple script that people can quickly understand what happens without reading any comments.

examples/sparse/sign.py
```diff
@@ -0,0 +1,82 @@
...
+data = CoraGraphDataset()
+g = data[0].to(dev)
+src, dst = g.edges()
+N = g.num_nodes()
+A = create_from_coo(dst, src, shape=(N, N))
+
+I = identity(A.shape, device=dev)
+A_hat = A + I
+D_hat = diag(A.sum(dim=1)) ** -0.5
```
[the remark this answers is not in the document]
mufeili: Nice catch, corrected

examples/sparse/sign.py
```diff
@@ -0,0 +1,82 @@
...
+A_hat = D_hat @ A_hat @ D_hat
```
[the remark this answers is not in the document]
mufeili: I think this is fine. By doing so people know that `A + I` is no longer needed.

examples/sparse/sign.py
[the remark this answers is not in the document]
mufeili: Linked the paper in the README file that @czkkkkkk suggested.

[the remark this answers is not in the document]
mufeili: The point of declaring a variable `num_hops = 2` is to tell users that we are using 2 hops.

examples/sparse/sign.py
mufeili: It has been the standard to denote the adjacency matrix by A in graph theory, see https://en.wikipedia.org/wiki/Adjacency_matrix. The SIGN paper actually also used it in various spaces like equation (4) in https://arxiv.org/pdf/2004.11198.pdf.

examples/sparse/sign.py
mufeili: Let's discuss it in the meeting next Tuesday.

czkkkkkk:
> > 1. I think this example will use our C++ sparse library in the end. So do we need to add a TODO to indicate we need to test this example after we have the C++ sparse library ready?
> 
> We should. Any suggestion on how to do that?
> 
I simply suggest to add a TODO here.

examples/sparse/sign.py
frozenbugs: In the paper A was defined as A = D−1/2WD−1/2. My understand is, A in equation (4) also means this.

My feeling is, if we also confuses with the annotation, then it would happen to many other readers.
If using A is standard, then add a comment to point out instead of A, W was used in the paper for the adjacency matrix should be enough,

examples/sparse/sign.py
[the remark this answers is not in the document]
mufeili: The page you referenced is a tutorial for new comers. The example is more for demonstrating the use of sparse matrix library. I prefer not mixing multiple objectives.

examples/sparse/sign.py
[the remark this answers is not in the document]
mufeili: Newbies should refer to a set of dedicated tutorials.

mufeili:
> > > 1. I think this example will use our C++ sparse library in the end. So do we need to add a TODO to indicate we need to test this example after we have the C++ sparse library ready?
> > 
> > 
> > We should. Any suggestion on how to do that?
> 
> I simply suggest to add a TODO here.

Added a TODO in #4860

examples/sparse/sign.py
mufeili: It's often to see inconsistencies in notations across papers. For example, the symmetric adjacency matrix normalization was probably first known by many people with the [GCN paper](https://arxiv.org/abs/1609.02907), which is way more impactful than the SIGN paper. I follow its notations in naming variables.

czkkkkkk (approved):

examples/sparse/sign.py
```diff
@@ -0,0 +1,93 @@
+"""
+Simplified version of SIGN, where we only consider symmetrically normalized
+adjacency matrix A and its powers for diffusion operators
```
frozenbugs, line 3: nit: end with dot.

examples/sparse/sign.py
```diff
@@ -0,0 +1,93 @@
...
+        X = A @ X
+        X_sign.append(X)
+    return X_sign
+
+
+class Model(nn.Module):
+    def __init__(self, in_size, num_hops, num_classes, hidden_size=256):
+        super().__init__()
+
+        # extra one for the original node features
```
frozenbugs, line 26: Extra one for the original node features.

examples/sparse/sign.py
```diff
@@ -0,0 +1,93 @@
...
+        num_feats = num_hops + 1
```
frozenbugs, line 27: num_feats is a little bit mis-leading, do we have a better name?

examples/sparse/sign.py
```diff
@@ -0,0 +1,82 @@
...
+    X_sign_train.append(feat[train_mask])
+    X_sign_test.append(feat[test_mask])
+Y_train = Y[train_mask]
+Y_test = Y[test_mask]
+
+model = Model(X.shape[1], num_hops, data.num_classes).to(dev)
+loss_func = nn.CrossEntropyLoss()
+optimizer = Adam(model.parameters(), lr=3e-3)
+
+model.train()
```
frozenbugs, line 82: What does this do?

examples/sparse/sign.py
```diff
@@ -0,0 +1,93 @@
...
+for epoch in range(10):
+    pred = model(X_sign_train)
+    loss = loss_func(pred, Y_train)
+    optimizer.zero_grad()
+    loss.backward()
+    optimizer.step()
+    train_acc = evaluate(X_sign_train, Y_train, model)
```
frozenbugs, line 89: For all example output, can we use a standard output format?

examples/sparse/sign.py
```diff
@@ -0,0 +1,93 @@
...
+    print(f"epoch {epoch} | train acc {train_acc}")
```
frozenbugs, line 90: I tried the code, it is a little bit overfit, is it a bug?


epoch 0 | train acc 0.27142858505249023
epoch 1 | train acc 1.0
epoch 2 | train acc 0.9428571462631226
epoch 3 | train acc 0.9285714030265808
epoch 4 | train acc 0.9642857313156128
epoch 5 | train acc 0.9857142567634583
epoch 6 | train acc 0.9928571581840515
epoch 7 | train acc 0.9928571581840515
epoch 8 | train acc 0.9928571581840515
epoch 9 | train acc 0.9928571581840515
test acc 0.7490000128746033

examples/sparse/sign.py
```diff
@@ -0,0 +1,93 @@
+"""
```
frozenbugs, line 1: Let's adopt the similar style (model definition + train method) as we established in https://github.com/dmlc/dgl/tree/master/tutorials/blitz.

examples/sparse/sign.py
mufeili: This is not a valid sentence.

examples/sparse/sign.py
mufeili: This is mostly for handling the different behaviors of batch normalization and dropout between training and evaluation. It's not necessary for this example though. For a reference, see https://pytorch.org/tutorials/beginner/saving_loading_models.html?highlight=eval

examples/sparse/sign.py
mufeili: This is often the case for Cora. There are only 140 training nodes in Cora after all.

mufeili:
Closed via #4908

PR #4877, opened 2022-11-11
[Sparse] Link to DGL

### LINKED ISSUE #4852: [Sparse] Link spares C++ library to DGL
## 🔨Work Item

**IMPORTANT:**

czkkkkkk:
## Description

Resolve #4852.

1. Build C++ sparse library in DGL CMakeLists.txt and link it to DGL. 
2. Move the definitions of sparse formats to a separate file.
3. Add a CSRToCOO conversion using DGL implementation with test.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/dgl_sparse/include/sparse/sparse_format.h
+++ b/dgl_sparse/include/sparse/sparse_format.h
@@ -1,0 +1,61 @@
+/**
+ *  Copyright (c) 2022 by Contributors
+ * @file sparse/sparse_format.h
+ * @brief DGL C++ sparse format header.
+ */
+#ifndef SPARSE_SPARSE_FORMAT_H_
+#define SPARSE_SPARSE_FORMAT_H_
+
+// clang-format off
+#include <sparse/dgl_headers.h>
+// clang-format on
+
+#include <torch/custom_class.h>
+#include <torch/script.h>
+
+#include <memory>
+
+namespace dgl {
+namespace sparse {
+
+/** @brief SparseFormat enumeration */
+enum SparseFormat { kCOO, kCSR, kCSC };
+
+/** @brief CSR sparse structure */
+struct CSR {
+  // CSR format index pointer array of the matrix
+  torch::Tensor indptr;
+  // CSR format index array of the matrix
+  torch::Tensor indices;
+  // The element order of the sparse format. In the SparseMatrix, we have data
+  // (value_) for each non-zero value. The order of non-zero values in (value_)
+  // may differ from the order of non-zero entries in CSR. So we store
+  // `value_indices` in CSR to indicate its relative non-zero value order to the
+  // SparseMatrix. With `value_indices`, we can retrieve the correct value for
+  // CSR, i.e., `value_[value_indices]`. If `value_indices` is not defined, this
+  // CSR follows the same non-zero value order as the SparseMatrix.
+  torch::optional<torch::Tensor> value_indices;
+};
+
+/** @brief COO sparse structure */
+struct COO {
+  // COO format row array of the matrix
+  torch::Tensor row;
+  // COO format column array of the matrix
+  torch::Tensor col;
+};
+
+/**
+ * @brief Convert a CSR format to COO format
+ * @param num_rows Number of rows of the sparse format
+ * @param num_cols Number of cols of the sparse format
+ * @param csr CSR sparse format
+ * @return COO sparse format
+ */
+std::shared_ptr<COO> CSRToCOO(
+    int64_t num_rows, int64_t num_cols, const std::shared_ptr<CSR> csr);
+
+}  // namespace sparse
+}  // namespace dgl
+
+#endif  // SPARSE_SPARSE_FORMAT_H_
--- a/dgl_sparse/src/sparse_format.cc
+++ b/dgl_sparse/src/sparse_format.cc
@@ -1,0 +1,35 @@
+/**
+ *  Copyright (c) 2022 by Contributors
+ * @file sparse_format.cc
+ * @brief DGL C++ sparse format implementations.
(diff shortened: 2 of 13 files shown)


czkkkkkk:
When linking dgl_sparse to both DGL and Libtorch, I met two major problems.
1. Diamond dependency for tensorpipe and logger. Both DGL and Pytorch have their own source code for tensorpipe and logger, leading to the same definition under the same namespace. Tensorpipe is still fine because it only raise some redefinition warnings during compilation. But the conflict of `dmlc/logging.h` and `torch/include/c10/util/logging_is_not_google_glog.h` could make the compilation fail. An example is when DGL uses the `CHECK_NOTNULL` from dmlc/logging.h, it will point to the `CHECK_NOTNULL` in Pytorch, which has a different signature.
image
2. Ambiguity of kInt, kFloat, etc. Both DGL (`dgl/runtime/c_runtime_api.h`) and Pytorch have these enumerations. We may need change our enum to `enum class {kInt, kFloat ...}`.

Changing the include order could magically solve these two issues. But we need to address them permanently. Do you have any suggestion? @jermainewang @BarclayII

dgl_sparse/include/sparse/sparse_format.h
```diff
@@ -0,0 +1,57 @@
+/**
+ *  Copyright (c) 2022 by Contributors
+ * @file sparse/sparse_format.h
+ * @brief DGL C++ sparse format header
```
frozenbugs, line 4: nit: always end with dot.

For all the comments below as well.

dgl_sparse/src/utils.h
```diff
@@ -45,6 +48,16 @@ inline static void ElementwiseOpSanityCheck(
       << ", " << B->shape()[1] << "])";
 }
 
+/** @brief Convert a DGL array to Torch tensor. */
```
frozenbugs, line 51: /** @brief Convert a Torch tensor to DGL array. */

dgl_sparse/src/utils.h
```diff
@@ -45,6 +48,16 @@ inline static void ElementwiseOpSanityCheck(
...
+inline static runtime::NDArray TorchTensorToDGLArray(torch::Tensor tensor) {
+  return runtime::DLPackConvert::FromDLPack(at::toDLPack(tensor));
+}
+
+/** @brief Convert a Torch tensor to DGL array. */
```
frozenbugs, line 56: ditto.

dgl_sparse/src/sparse_format.cc
```diff
@@ -0,0 +1,35 @@
...
+    int64_t num_rows, int64_t num_cols, const std::shared_ptr<CSR> csr) {
+  auto indptr = TorchTensorToDGLArray(csr->indptr);
+  auto indices = TorchTensorToDGLArray(csr->indices);
+  bool data_as_order = false;
+  runtime::NDArray data = aten::NullArray();
+  if (csr->value_indices.has_value()) {
+    data_as_order = true;
+    data = TorchTensorToDGLArray(csr->value_indices.value());
+  }
+  auto dgl_csr = aten::CSRMatrix(num_rows, num_cols, indptr, indices, data);
```
jermainewang, line 27: * Perhaps we could add some util functions like `CSRToOldDGLCSR`, `OldDGLCSRToCSR` as they will be quite commonly used in other operators.
* I also suggest to add the `sorted` related flags to the new CSR/COO structure. They are performance critical for certain operations.

tests/pytorch/mock_sparse2/test_sparse_matrix.py
```diff
@@ -89,3 +89,30 @@ def test_create_from_csc(dense_dim, indptr, indices, shape):
     assert torch.allclose(mat_indptr, indptr)
     assert torch.allclose(mat_indices, indices)
     assert torch.allclose(mat_val, val)
+
+@pytest.mark.parametrize("dense_dim", [None, 4])
+@pytest.mark.parametrize("indptr", [[0, 0, 1, 4], (0, 1, 2, 4)])
+@pytest.mark.parametrize("indices", [(0, 1, 2, 3), (1, 4, 3, 2)])
+@pytest.mark.parametrize("shape", [None, (3, 5)])
+def test_csr_to_coo(dense_dim, indptr, indices, shape):
```
jermainewang, line 97: Could you add tests on GPU too? For now, you could use the `ctx()` object from the test backend to get the current device being tested.

jermainewang (approved):
Mostly good. Some minor comments.

tests/pytorch/mock_sparse2/test_sparse_matrix.py
czkkkkkk: Added

dgl_sparse/src/sparse_format.cc
czkkkkkk: OK. I will have these two things in the next PR. This PR is mainly for linking to DGL.

BarclayII (approved):
I like this solution.

yaox12:
Better to add options for `BUILD_SPARSE` and `TORCH_PYTHON_INTERPS` in
https://github.com/dmlc/dgl/blob/3103f8cb362033d9b1543f4bd4b2bd9ac87a61d4/CMakeLists.txt#L38-L39

czkkkkkk:
> Better to add options for `BUILD_SPARSE` and `TORCH_PYTHON_INTERPS` in
> 
> https://github.com/dmlc/dgl/blob/3103f8cb362033d9b1543f4bd4b2bd9ac87a61d4/CMakeLists.txt#L38-L39
You are right. Will make a PR to include them.

PR #4897, opened 2022-11-15
[Sparse] Set Value and Dense

mufeili:
## Description
- major
    - Set non-zero values with `A.val = x`
    - Syntax sugar `A(x)`
    - Convert a sparse matrix into a dense one with `A.dense()`
- issues to discuss
    - Should we also support `D.dense()`?
    - Should `A(x)` and `D(x)` return a new matrix or simply the old matrix with in-place modification?
- misc
    - In message passing `AX`, rows correspond to destination nodes while columns correspond to source nodes -> fixes in the doc of `sparse_matrix.py`
    - The non-zero values created when not specified should have the same device as the edge indices -> fixes in `sparse_matrix.py`
    - Extended the unit tests to cover GPU

## Checklist
- [x] Code is well-documented

### DIFF
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -129,6 +129,12 @@ class SparseMatrix : public torch::CustomClassHolder {
   /** @return {row, col, value} tensors in the CSC format. */
   std::vector<torch::Tensor> CSCTensors();
 
+  /**
+   * @brief set non-zero values of the sparse matrix
+   * @param value Values of the sparse matrix
+   */
+  void SetValue(torch::Tensor value);
+
  private:
   /** @brief Create the COO format for the sparse matrix internally */
   void _CreateCOO();
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -130,7 +130,7 @@ class SparseMatrix : public torch::CustomClassHolder {
   std::vector<torch::Tensor> CSCTensors();
 
   /**
-   * @brief set non-zero values of the sparse matrix
+   * @brief Set non-zero values of the sparse matrix
    * @param value Values of the sparse matrix
    */
   void SetValue(torch::Tensor value);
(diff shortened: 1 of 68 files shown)


dgl_sparse/include/sparse/sparse_matrix.h
```diff
@@ -129,6 +129,12 @@ class SparseMatrix : public torch::CustomClassHolder {
   /** @return {row, col, value} tensors in the CSC format. */
   std::vector<torch::Tensor> CSCTensors();
 
+  /**
+   * @brief set non-zero values of the sparse matrix
```
frozenbugs, line 133: Set ... matrix.

frozenbugs (approved):

mufeili:
@frozenbugs What do you think of "issues to discuss"?

frozenbugs:
> @frozenbugs What do you think of "issues to discuss"?
- 
-

frozenbugs:
> @frozenbugs What do you think of "issues to discuss"?

Inplace should

tests/pytorch/mock_sparse2/test_sparse_matrix.py
```diff
@@ -2,6 +2,8 @@
 import torch
 import sys
 
+import backend as F
```
czkkkkkk, line 5: Why do we use backend here? Shall we directly use Pytorch?

python/dgl/mock_sparse2/sparse_matrix.py
```diff
@@ -21,6 +21,33 @@ def val(self) -> torch.Tensor:
...
+        """Set the non-zero values inplace.
+
+        Parameters
+        ----------
+        x : torch.Tensor, optional
+            The values of shape (nnz) or (nnz, D)
+        """
+        self.c_sparse_matrix.set_val(x)
+
+    def __call__(self, x: torch.Tensor):
```
jermainewang, line 35: I think it's my bad to include this in the first MVP. The more I think of it the more I feel the syntax is not straightforward, e.g., it may confuse users with functions. How about we remove it and come back to this after we see this is frequently used?

tests/pytorch/mock_sparse2/test_sparse_matrix.py
jermainewang: I think it is for device control L27.

jermainewang (approved):
LGTM other than the `A(x)`. Let's drop it from this MVP.

python/dgl/mock_sparse2/sparse_matrix.py
mufeili: Sure.

Issue #4887, opened 2022-11-14
[DistDGL][Bug] Node type and node feature can not use the same name.


classicsong:
## 🐛 Bug
In DistDGL, If both node type and one of its node feature have the same name, the system will crash.
The bug is related to https://github.com/dmlc/dgl/blob/ee9887d66a9eb6dc9663a5238e1874b7646acf3a/python/dgl/distributed/dist_graph.py#L172-L196

## To Reproduce

1.
1.
1.

## Environment

 - DGL Version (e.g., 1.0): NIghtly build and 0.9.0
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):
 - OS (e.g., Linux):
 - How you installed DGL (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version (if applicable):
 - GPU models and configuration (e.g. V100):
 - Any other relevant information:

classicsong:
Any ETA?

Rhett-Ying:
I will open a PR end of today

Issue #4463, opened 2022-08-24
[BUG] DistDGL crashes when a heterogeneous graph has an edge type with an edge feature has the same name of edge type.


classicsong:
## To Reproduce
You can reproduce the bug using following code:

```
import torch as th
import tempfile
from pathlib import Path
from scipy import sparse as spsp
import dgl
from dgl.distributed import partition_graph, load_partition, load_partition_book, node_split, edge_split
import socket
import os
import random
import multiprocessing as mp
import time
from dgl.distributed import DistGraphServer, DistGraph 

def start_server(rank, tmpdir, disable_shared_mem, graph_name, graph_format=['coo'],
                 keep_alive=False):
    g = DistGraphServer(rank, "rpc_ip_config.txt", 1, 1,
                        tmpdir / (graph_name + '.json'), disable_shared_mem=disable_shared_mem,
                        graph_format=graph_format, keep_alive=keep_alive)
    g.start()


def start_client(rank, tmpdir, disable_shared_mem):
    gpb = None
    if disable_shared_mem:
        _, _, _, gpb, _, _, _ = load_partition(os.path.join(tmpdir,'test_sampling.json'), rank)
    dgl.distributed.initialize("rpc_ip_config.txt")
    dist_graph = DistGraph("test_sampling", gpb=gpb)
    dist_graph.edges["r1"].data["r1"]
    dgl.distributed.exit_client()

def create_random_hetero():
    num_nodes = {'n1': 10000, 'n2': 10010, 'n3': 10020}
    etypes = [('n1', 'r1', 'n2'),
              ('n1', 'r2', 'n3'),
              ('n2', 'r3', 'n3')]
    edges = {}
    for etype in etypes:
        src_ntype, _, dst_ntype = etype
        arr = spsp.random(num_nodes[src_ntype], num_nodes[dst_ntype], density=0.001, format='coo',
                          random_state=100)
        edges[etype] = (arr.row, arr.col)
    g = dgl.heterograph(edges, num_nodes)
    g.nodes['n1'].data['n1'] = th.unsqueeze(th.arange(0, g.number_of_nodes('n1')), 1)
    g.edges['r1'].data['r1'] = th.unsqueeze(th.arange(0, g.number_of_edges('r1')), 1)
    return g

def generate_ip_config(file_name, num_machines, num_servers):
    """Get local IP and available ports, writes to file."""
    # get available IP in localhost
[...]

PR #4906, opened 2022-11-16
[Dist] init node/edge data store for Node/EdgeDataView in appropriate place

Rhett-Ying:
## Description

fix for #4887 && #4463

The direct cause of the crash is `NodeDataView` with `ntype=None` is always instantiated in `DistGraph.__init__()` and assigned to `g._ndata`. For a heterogeneous graph, instantiating with `ntype=None` will merge all data for all ntypes into single dict instead of `a dict of dict`. This is what we expect for heterogeneous graph.

as for this fix, I think we don't need to always instantiate and assign `g._ndata` in `__init__()`. It's ok to delay it.
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/distributed/dist_graph.py
+++ b/python/dgl/distributed/dist_graph.py
@@ -519,8 +519,7 @@ class DistGraph:
         self._client.map_shared_data(self._gpb)
 
     def _init_ndata_store(self):
-        '''Init node data store
-        '''
+        '''Initialize node data store.'''
         self._ndata_store = {}
         for ntype in self.ntypes:
             names = self._get_ndata_names(ntype)
@@ -541,8 +540,7 @@ class DistGraph:
                 self._ndata_store[ntype] = data
 
     def _init_edata_store(self):
-        '''Init edge data store
-        '''
+        '''Initialize edge data store.'''
         self._edata_store = {}
         for etype in self.canonical_etypes:
             names = self._get_edata_names(etype)
--- a/python/dgl/distributed/dist_graph.py
+++ b/python/dgl/distributed/dist_graph.py
@@ -522,8 +522,6 @@ class DistGraph:
 
         self._ndata_store = {}
         self._edata_store = {}
-        self._ndata = NodeDataView(self)
-        self._edata = EdgeDataView(self)
 
         self._num_nodes = 0
         self._num_edges = 0
@@ -554,8 +552,6 @@ class DistGraph:
 
         self._ndata_store = {}
         self._edata_store = {}
-        self._ndata = NodeDataView(self)
-        self._edata = EdgeDataView(self)
         self._num_nodes = 0
         self._num_edges = 0
         for part_md in self._gpb.metadata():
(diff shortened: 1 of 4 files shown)


python/dgl/distributed/dist_graph.py
```diff
@@ -600,7 +596,7 @@ def ndata(self):
             The data view in the distributed graph storage.
         """
         assert len(self.ntypes) == 1, "ndata only works for a graph with one node type."
-        return self._ndata
+        return NodeDataView(self)
```
jermainewang, line 599: It looks like creating `NodeDataView` or `EdgeDataView` will also create `DistTensor`

https://github.com/dmlc/dgl/blob/949f87cc1df76980ce2af2ac871f7f518463b2fc/python/dgl/distributed/dist_graph.py#L190-L196

How costly are they? This may be an issue because `DistGraph` is typically created only once but `ndata` and `edata` may be invoked many times.

Rhett-Ying: `DistTensor` are referenced by `g._ndata_store`, so will not be created/destroyed in each call.

Rhett-Ying: I refactored the data store init logic. this is a bit radical. pls take a look

Rhett-Ying: `self._ndata/_edata` is not defined yet because I think it's trivial to instantiate `NodeDataView` when calling `g.ndata`. And `g.nodes['_N']` and `g.ndata` have the same behavior: always instantiate a new one.
In original code, a new `NodeDataView` is always instantiated when calling `g.nodes['_N']`.

Rhett-Ying: always instantiating a new `NodeDataView` and creating `DistTensor` by iterating over existing names is more conservative than what I does now.

tests/distributed/test_dist_graph_store.py
```diff
@@ -697,12 +697,19 @@ def create_random_hetero():
         )
         edges[etype] = (arr.row, arr.col)
     g = dgl.heterograph(edges, num_nodes)
-    g.nodes["n1"].data["feat"] = F.unsqueeze(
-        F.arange(0, g.number_of_nodes("n1")), 1
-    )
-    g.edges["r1"].data["feat"] = F.unsqueeze(
-        F.arange(0, g.number_of_edges("r1")), 1
-    )
+    # assign ndata & edata.
```
jermainewang, line 700: Could you put the issue number here too?

jermainewang (approved):

jermainewang:
Approved. Some minor comments.

python/dgl/distributed/standalone_kvstore.py
```diff
@@ -58,6 +58,7 @@ def init_data(self, name, shape, dtype, part_policy, init_func, is_gdata=True):
     def delete_data(self, name):
         '''delete the data'''
         del self._data[name]
+        self._gdata_name_list.remove(name)
```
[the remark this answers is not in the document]
Rhett-Ying: This is actually a bug that is not exposed/captured before. Here's the counterpart in `kvstore`: https://github.com/dmlc/dgl/blob/4ecd2c85858879ad432273e8aa2794f8d674fd9b/python/dgl/distributed/kvstore.py#L1123

PR #4909, opened 2022-11-16
less than 2^31−1 nodes then 32-bit

hummingg:
English Version: However, if a graph contains less than 2^31−1 nodes or edges, one should use 32-bit integers as it leads to better speed and requires less memory.


Rhett-Ying:
pls target on `master` branch instead of other branches.

Rhett-Ying:
open a new PR that's based on latest official master branch is simpler

Rhett-Ying:
duplicate

PR #4917, opened 2022-11-17
[Sparse] Add HGNN example

### LINKED ISSUE #4870: [Sparse] HGNN Example
## 🔨Work Item

**IMPORTANT:**

BarclayII:
## Description
Add HGNN example (https://arxiv.org/pdf/1809.09401.pdf).
Closes #4870.

I listed a question and a small suggestion in the code and I'll list here as well:
* Make `DGLGraph.adj()` return a DGL sparse matrix instead of PyTorch sparse matrix, or do we have a convenient conversion function?
* Have an `eye()` function to construct identity sparse matrix.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/examples/sparse/hgnn.py
+++ b/examples/sparse/hgnn.py
@@ -1,3 +1,6 @@
+"""
+Hypergraph Neural Networks (https://arxiv.org/pdf/1809.09401.pdf)
+"""
 import dgl
 import dgl.data
 import dgl.mock_sparse
--- a/examples/sparse/hgnn.py
+++ b/examples/sparse/hgnn.py
@@ -1,0 +1,92 @@
+import dgl
+import dgl.data
+import dgl.mock_sparse
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torchmetrics.functional import accuracy
+import tqdm
+
+dataset = dgl.data.CoraGraphDataset()
+
+graph = dataset[0]
+# The paper created the hypergraph by "each time one vertex in the
+# graph is selected as the centroid and its connected vertices
+# are used to generate one hyperedge including the centroid
+# itself".  In this case, the incidence matrix of the hypergraph
+# is the same as the adjacency matrix of the original graph (with
+# self-loops).
+# We follow the paper assuming that the rows of the incidence matrix
+# are nodes and the columns are edges.
+
+# QUESTION: can we either (1) easily convert a PyTorch sparse matrix
+# to a DGL sparse matrix, or (2) make g.adj() a DGL sparse matrix
+# in the future?
+adj = dgl.add_self_loop(graph).adj().coalesce()
+row, col = adj.indices()
+values = adj.values()
+H = dgl.mock_sparse.create_from_coo(row, col, values)
+
+X = graph.ndata["feat"]
+Y = graph.ndata["label"]
+train_mask = graph.ndata["train_mask"]
+val_mask = graph.ndata["val_mask"]
+test_mask = graph.ndata["test_mask"]
+
+
+class HGNN(nn.Module):
+    def __init__(self, in_size, out_size, hidden_dims=16):
+        super().__init__()
+
+        self.Theta1 = nn.Linear(in_size, hidden_dims)
+        self.Theta2 = nn.Linear(hidden_dims, out_size)
+        self.dropout = nn.Dropout(0.5)
+
+    def forward(self, H, X):
+        d_V = H.sum(1)  # node degree
+        d_E = H.sum(0)  # edge degree
+        D_V_invsqrt = dgl.mock_sparse.diag(d_V ** -0.5)  # D_V ** (-1/2)
+        D_E_inv = dgl.mock_sparse.diag(d_E ** -1)  # D_E ** (-1)
+        # Minor suggestion: could we have dgl.mock_sparse.eye(N)?
+        W = dgl.mock_sparse.diag(torch.ones(d_E.shape[0]))
+
+        conv = D_V_invsqrt @ H @ W @ D_E_inv @ H.T @ D_V_invsqrt
+
+        X = conv @ self.Theta1(self.dropout(X))
+        X = F.relu(X)
+        X = conv @ self.Theta2(self.dropout(X))
+        return X
+
+
+model = HGNN(X.shape[1], dataset.num_classes)
+opt = torch.optim.Adam(model.parameters(), lr=0.001)
+best_val_acc, best_test_acc = 0, 0
+
+with tqdm.trange(500) as tq:
+    for epoch in tq:
+        # Train
+        model.train()
+        Y_hat = model(H, X)
+        loss = F.cross_entropy(Y_hat[train_mask], Y[train_mask])
+        opt.zero_grad()
+        loss.backward()
+        opt.step()
+
+        # Evaluate
+        model.eval()
+        Y_hat = model(H, X)
+        val_acc = accuracy(Y_hat[val_mask], Y[val_mask])
+        test_acc = accuracy(Y_hat[test_mask], Y[test_mask])
+        if val_acc > best_val_acc:
+            best_val_acc = val_acc
+            best_test_acc = test_acc
+
+        tq.set_postfix(
+            {
+                "Loss": f"{loss.item():.5f}",
+                "Val acc": f"{val_acc:.5f}",
+                "Test acc": f"{test_acc:.5f}",
+            },
+            refresh=False,
+        )
+print(f"Best val acc: {best_val_acc} Best test acc: {best_test_acc}")
--- a/examples/sparse/hgnn.py
+++ b/examples/sparse/hgnn.py
@@ -3,7 +3,7 @@ Hypergraph Neural Networks (https://arxiv.org/pdf/1809.09401.pdf)
 """
 import dgl
 import dgl.data
-import dgl.mock_sparse
+import dgl.mock_sparse as dglsp
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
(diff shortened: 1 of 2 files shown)


examples/sparse/hgnn.py
```diff
@@ -0,0 +1,95 @@
+"""
+Hypergraph Neural Networks (https://arxiv.org/pdf/1809.09401.pdf)
+"""
+import dgl
+import dgl.data
+import dgl.mock_sparse
```
mufeili, line 6: Is it necessary to import `dgl.data` and `dgl.mock_sparse`?

examples/sparse/hgnn.py
```diff
@@ -0,0 +1,95 @@
...
+
+graph = dataset[0]
+# The paper created the hypergraph by "each time one vertex in the
+# graph is selected as the centroid and its connected vertices
+# are used to generate one hyperedge including the centroid
+# itself".  In this case, the incidence matrix of the hypergraph
+# is the same as the adjacency matrix of the original graph (with
+# self-loops).
+# We follow the paper assuming that the rows of the incidence matrix
+# are nodes and the columns are edges.
```
mufeili, line 23: An attempt to improve the description from L16-23

```suggestion
The paper created a hypergraph from the original graph. For each node in the original graph, a hyperedge in the hypergraph is created to connect this center node and all neighbors of the center node including itself. In this case, ...
```

examples/sparse/hgnn.py
```diff
@@ -0,0 +1,95 @@
...
+dataset = dgl.data.CoraGraphDataset()
+
+graph = dataset[0]
+# The paper created the hypergraph by "each time one vertex in the
+# graph is selected as the centroid and its connected vertices
+# are used to generate one hyperedge including the centroid
+# itself".  In this case, the incidence matrix of the hypergraph
+# is the same as the adjacency matrix of the original graph (with
+# self-loops).
+# We follow the paper assuming that the rows of the incidence matrix
```
mufeili, line 22:
```suggestion
# We follow the paper and assume that the rows of the incidence matrix
```

examples/sparse/hgnn.py
mufeili, line 23:
```suggestion
# are for nodes and the columns are for edges.
```

examples/sparse/hgnn.py
```diff
@@ -0,0 +1,95 @@
...
+from torchmetrics.functional import accuracy
+import tqdm
+
+dataset = dgl.data.CoraGraphDataset()
+
+graph = dataset[0]
+# The paper created the hypergraph by "each time one vertex in the
+# graph is selected as the centroid and its connected vertices
+# are used to generate one hyperedge including the centroid
+# itself".  In this case, the incidence matrix of the hypergraph
```
mufeili, line 19: An attempt to improve the description from L16-19

```suggestion
The paper created a hypergraph from the original graph. For each node in the original graph, a hyperedge in the hypergraph is created to connect this center node and all neighbors of the center node including itself. In this case, ...
```

examples/sparse/hgnn.py
```diff
@@ -0,0 +1,95 @@
...
+# is the same as the adjacency matrix of the original graph (with
+# self-loops).
+# We follow the paper assuming that the rows of the incidence matrix
+# are nodes and the columns are edges.
+
+# QUESTION: can we either (1) easily convert a PyTorch sparse matrix
+# to a DGL sparse matrix, or (2) make g.adj() a DGL sparse matrix
+# in the future?
```
mufeili, line 27: L25-27 has been tracked by https://github.com/dmlc/dgl/issues/4876. You may remove these comments.

examples/sparse/hgnn.py
```diff
@@ -0,0 +1,95 @@
...
+adj = dgl.add_self_loop(graph).adj().coalesce()
```
mufeili, line 28: Why not use `.edges()` to access `row` and `col`? Also currently `transpose` is `False` for `.adj()`, so `adj.indices()` yields `src, dst` while strictly speaking we need `create_from_coo(dst, src)` for non-symmetric adjacency matrices.

examples/sparse/hgnn.py
```diff
@@ -0,0 +1,95 @@
...
+row, col = adj.indices()
+values = adj.values()
```
mufeili, line 30: Since the graph is not weighted, we don't need `values`. By default, 1-valued tensor will be created for non-zero values if not specified.

examples/sparse/hgnn.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def forward(self, H, X):
+        d_V = H.sum(1)  # node degree
+        d_E = H.sum(0)  # edge degree
+        D_V_invsqrt = dgl.mock_sparse.diag(d_V ** -0.5)  # D_V ** (-1/2)
+        D_E_inv = dgl.mock_sparse.diag(d_E ** -1)  # D_E ** (-1)
+        # Minor suggestion: could we have dgl.mock_sparse.eye(N)?
+        W = dgl.mock_sparse.diag(torch.ones(d_E.shape[0]))
+
+        conv = D_V_invsqrt @ H @ W @ D_E_inv @ H.T @ D_V_invsqrt
```
mufeili, line 56: L49-56 can be done in the pre-processing step.

examples/sparse/hgnn.py
```diff
@@ -0,0 +1,95 @@
...
+
+        X = conv @ self.Theta1(self.dropout(X))
+        X = F.relu(X)
+        X = conv @ self.Theta2(self.dropout(X))
+        return X
+
+
+model = HGNN(X.shape[1], dataset.num_classes)
+opt = torch.optim.Adam(model.parameters(), lr=0.001)
+best_val_acc, best_test_acc = 0, 0
```
mufeili, line 66: Maybe not necessary for a demonstration

mufeili:
What test accuracy numbers can you obtain with this script? What are the numbers in the paper? We don't need to match the numbers as long as they look fine.

examples/sparse/hgnn.py
mufeili, line 28: Maybe we can directly do `A + I` without using `add_self_loop`.

examples/sparse/hgnn.py
```diff
@@ -0,0 +1,95 @@
...
+        self.Theta1 = nn.Linear(in_size, hidden_dims)
+        self.Theta2 = nn.Linear(hidden_dims, out_size)
+        self.dropout = nn.Dropout(0.5)
+
+    def forward(self, H, X):
+        d_V = H.sum(1)  # node degree
+        d_E = H.sum(0)  # edge degree
+        D_V_invsqrt = dgl.mock_sparse.diag(d_V ** -0.5)  # D_V ** (-1/2)
+        D_E_inv = dgl.mock_sparse.diag(d_E ** -1)  # D_E ** (-1)
+        # Minor suggestion: could we have dgl.mock_sparse.eye(N)?
```
mufeili, line 53: Use `identity`?

examples/sparse/hgnn.py
BarclayII: No need to import `dgl.data`, but `dgl.mock_sparse` will throw an error without explicit importing.

examples/sparse/hgnn.py
BarclayII: If `H` is an argument in the forward function then technically I cannot put it in the preprocessing step.  So I guess the problem is whether `H` should be in the forward function or in `__init__`.

examples/sparse/hgnn.py
mufeili: I see. Perhaps either `from dgl.mock_sparse import xx` or alternatively `import dgl.mock_sparse as xx`. The existing examples follow the former option.

BarclayII:
> What test accuracy numbers can you obtain with this script? What are the numbers in the paper? We don't need to match the numbers as long as they look fine.

The test accuracy is 81.6% which matches the paper.

mufeili:
I'm good in general, though I guess @frozenbugs might have some further suggestions regarding format/style.

examples/sparse/hgnn.py
```diff
@@ -0,0 +1,91 @@
...
+
+class HGNN(nn.Module):
+    def __init__(self, H, in_size, out_size, hidden_dims=16):
+        super().__init__()
+
+        self.Theta1 = nn.Linear(in_size, hidden_dims)
+        self.Theta2 = nn.Linear(hidden_dims, out_size)
+        self.dropout = nn.Dropout(0.5)
+
+        # Compute the Laplacian with Sparse Matrix API
```
frozenbugs, line 21: Follow this style to highlight the sparse API usage?

################################################################################
# (HIGHLIGHT) Take the advantage of DGL sparse APIs to implement the feature
# pre-computation.
################################################################################

examples/sparse/hgnn.py
```diff
@@ -0,0 +1,91 @@
...
+    Y_hat = model(X)
+    val_acc = accuracy(Y_hat[val_mask], Y[val_mask])
+    test_acc = accuracy(Y_hat[test_mask], Y[test_mask])
+    return val_acc, test_acc
+
+def load_data():
+    dataset = dgl.data.CoraGraphDataset()
+
+    graph = dataset[0]
+    # The paper created a hypergraph from the original graph. For each node in
```
frozenbugs, line 55: Should this be on the top of the paper?
It sounds like a clarification for the whole implementation instead of a piece of code here.

examples/sparse/hgnn.py
```diff
@@ -0,0 +1,91 @@
...
+    Y = graph.ndata["label"]
+    train_mask = graph.ndata["train_mask"]
+    val_mask = graph.ndata["val_mask"]
+    test_mask = graph.ndata["test_mask"]
+    return H, X, Y, dataset.num_classes, train_mask, val_mask, test_mask
+
+def main():
+    H, X, Y, num_classes, train_mask, val_mask, test_mask = load_data()
+    model = HGNN(H, X.shape[1], num_classes)
+    opt = torch.optim.Adam(model.parameters(), lr=0.001)
```
frozenbugs, line 76: opt -> optimizer

We should either all use opt as abbr. or just use the full word.
I notice many inconsistency in our code and examples.

examples/sparse/hgnn.py
BarclayII: ~~Speaking of cheap neon light banners...~~

frozenbugs (approved):

PR #4919, opened 2022-11-18
[Sparse] APPNP example

mufeili:
## Description
APPNP example, the accuracy is > 0.82

## Checklist
- [x] Code is well-documented


examples/sparse/appnp.py
```diff
@@ -0,0 +1,116 @@
...
+            nn.Linear(in_size, hidden_size),
+            nn.ReLU(),
+            nn.Dropout(dropout),
+            nn.Linear(hidden_size, out_size),
+        )
+        self.num_hops = num_hops
+        self.A_dropout = nn.Dropout(dropout)
+        self.alpha = alpha
+
+    def forward(self, A, X):
```
mufeili, line 37: We might want to extend `DropEdge` for sparse matrix. @jermainewang

examples/sparse/appnp.py
```diff
@@ -0,0 +1,116 @@
...
+    D_hat = diag(A_hat.sum(dim=1)) ** -0.5
+    A_hat = D_hat @ A_hat @ D_hat
+
+    # Create APPNP model.
+    X = g.ndata["feat"]
+    in_size = X.shape[1]
+    out_size = dataset.num_classes
+    model = APPNP(in_size, out_size).to(dev)
+
+    # Kick off training.
```
frozenbugs, line 115: Let's mention that the A_hat will be updated in place in the comment.

mufeili: Instead of leaving a comment here, I left a comment in `APPNP.forward`.

frozenbugs: Let's add the following comment at *every* place that we pass in A_hat.

# Note that A_hat will be update inside the method.

frozenbugs (approved):

examples/sparse/appnp.py
mufeili: `A_hat` does not get updated right? Because we reset `A_hat.val` to its initial value in the end.

frozenbugs: Ah, I see.

examples/sparse/appnp.py
```diff
@@ -0,0 +1,117 @@
...
+            nn.Linear(in_size, hidden_size),
+            nn.ReLU(),
+            nn.Dropout(dropout),
+            nn.Linear(hidden_size, out_size),
+        )
+        self.num_hops = num_hops
+        self.A_dropout = nn.Dropout(dropout)
+        self.alpha = alpha
+
+    def forward(self, A, X):
```
frozenbugs, line 37: A -> A_hat

PR #4924, opened 2022-11-20
[Performance] Leverage hashmap to accelerate CSRSliceMatrix<kDGLCUDA, IdType>

gpzlx1:
## Description

The old `CSRSliceMatrix<kDGLCUDA, IdType>` uses BinarySearch for cols lookup with `O(E * log(#cols))` time complexity. However, `CSRSliceMatrix<kDGLCUDA, IdType>` has a severe performance drop when the `#cols` are very large. In this PR, a hashmap is leveraged instead of BinarySearch to accelerate `CSRSliceMatrix<kDGLCUDA, IdType>`, reducing the time complexity to `O(E)`. The performance of `CSRSliceMatrix<kDGLCUDA, IdType>` is improved by 1.2-49x with an acceptable memory overhead.

## Benchmark
```python
# code
import torch
import dgl
import time
import numpy as np


def bench(g, seeds, iter):
    rec_list = []
    for _ in range(iter):
        tic = time.time()
        dgl.node_subgraph(g, seeds)
        torch.cuda.synchronize()
        toc = time.time()
        rec_list.append(toc - tic)
    print("Time Cost [ms]: ", np.mean(rec_list[3:]) * 1000)


if __name__ == "__main__":
    g = RedditDataset(self_loop=True)[0].to('cuda')
    selected_size = [5000, 10000, 100000, 200000]
    for s in selected_size:
        seeds = torch.arange(s).cuda().long()
        bench(g, seeds, 10)
```

**Results:**
| #seeds             | 5000 | 10000 | 100000 | 200000  |
| ------------------ | ---- | ----- | ------ | ------- |
| Old Time Cost [ms] | 0.74 | 1.26  | 108.89 | 3015.49 |
| New Time Cost [ms] | 0.60 | 0.75  | 10.37  | 62.24   |
| SpeedUp            | 1.2  | 1.7   | 10.5   | 48.5    |

## Memory Overhead
For performance, the load factor of the hashmap is in (0.25, 0.5) and the additional memory overhead is `O(#cols)`. Because `#cols` is usually less than 1 Million on GPU, the memory overhead is not significant (less than 31MB) even at a low load factor.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/src/array/cuda/spmat_op_impl_csr.cu
+++ b/src/array/cuda/spmat_op_impl_csr.cu
@@ -4,6 +4,7 @@
  * @brief CSR operator CPU implementation
  */
 #include <dgl/array.h>
+#include <thrust/execution_policy.h>
 #include <thrust/for_each.h>
 
 #include <numeric>
@@ -607,7 +608,7 @@ CSRMatrix CSRSliceMatrix(
   IdArray hashmap_buffer = Full(-1, buffer_size, nbits, ctx);
   using it = thrust::counting_iterator<int64_t>;
   thrust::for_each(
-      it(0), it(new_ncols),
+      thrust::device.on(stream), it(0), it(new_ncols),
       [key = cols.Ptr<IdType>(), buffer = hashmap_buffer.Ptr<IdType>(),
        buffer_size] __device__(int64_t i) {
         NodeQueryHashmap<IdType> hashmap(buffer, buffer_size);
--- a/src/array/cuda/spmat_op_impl_csr.cu
+++ b/src/array/cuda/spmat_op_impl_csr.cu
@@ -4,6 +4,7 @@
  * @brief CSR operator CPU implementation
  */
 #include <dgl/array.h>
+#include <thrust/for_each.h>
 
 #include <numeric>
 #include <unordered_set>
@@ -13,6 +14,8 @@
 #include "./atomic.cuh"
 #include "./dgl_cub.cuh"
 #include "./utils.h"
+
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
 
 namespace dgl {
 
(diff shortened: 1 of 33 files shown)


yaox12:
@dgl-bot

yaox12:
@dgl-bot

src/array/cuda/spmat_op_impl_csr.cu
```diff
@@ -423,26 +426,130 @@ template std::vector<NDArray> CSRGetDataAndIndices<kDGLCUDA, int64_t>(
...
-      cuda::AtomicAdd(count + rpos, IdType(1));
+    IdType* hashmap_buffer, int64_t buffer_size, IdType* mask, IdType* count) {
+  assert(blockDim.x == WARP_SIZE);
+  assert(blockDim.y == BLOCK_WARPS);
+
+  int warp_id = threadIdx.y;
+  int laneid = threadIdx.x;
+  IdType out_row = blockIdx.x * TILE_SIZE + threadIdx.y;
+  IdType last_row =
+      MIN(static_cast<IdType>(blockIdx.x + 1) * TILE_SIZE, num_rows);
```
yaox12, line 531: There is a `min()` included by `cuda_runtime.h`. Can we use that?

src/array/cuda/spmat_op_impl_csr.cu
```diff
@@ -507,10 +608,16 @@ CSRMatrix CSRSliceMatrix(
   }
 
   // Execute SegmentMaskColKernel
-  int nb = (nnz_csr + nt - 1) / nt;
+  const int64_t num_rows = csr.num_rows;
+  constexpr int WARP_SIZE = 32;
+  constexpr int BLOCK_WARP = 256 / WARP_SIZE;
+  constexpr int TILE_SIZE = 16;
```
yaox12, line 614: How did you choose the value of `TILE_SIZE`?

yaox12 (commented):
Thanks for your contribution. I'd like to invite @ayasar70 for review since he has worked on improving `_SegmentMaskColKernel`.

src/array/cuda/spmat_op_impl_csr.cu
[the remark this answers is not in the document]
gpzlx1: Yes. Done.

src/array/cuda/spmat_op_impl_csr.cu
```diff
@@ -423,26 +426,130 @@ template std::vector<NDArray> CSRGetDataAndIndices<kDGLCUDA, int64_t>(
...
+}
+
+/**
+ * @brief A hashmap designed for CSRSliceMatrix, similar in function to set,
+ * Insert api inserts an element, Query api queries if an element is in the
+ * hashmap. For performance, it can only be created and called in the cuda
+ * kernel.
  */
 template <typename IdType>
+struct NodeQueryHashmap {
```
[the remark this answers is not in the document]
gpzlx1: In short, NodeQueryHashmap modifies the hash function and removes unnecessary memory space compared to DGL's hashmap. DGL's hashmap is designed for graph relabel and does not support element query if the element does not exist in the hashable. NodeQueryHashmap overcomes this problem.

src/array/cuda/spmat_op_impl_csr.cu
gpzlx1: When TILE_SIZE = 16, I get the best performance with my device.

src/array/cuda/spmat_op_impl_csr.cu
```diff
@@ -423,26 +424,131 @@ template std::vector<NDArray> CSRGetDataAndIndices<kDGLCUDA, int64_t>(
...
+  key = key ^ (key >> 12);
+  key = key + (key << 2);
+  key = key ^ (key >> 4);
+  key = key * 2057;
+  key = key ^ (key >> 16);
+  return key;
+}
+
+/**
+ * @brief Thomas Wang's 64 bit Mix Function
```
frozenbugs, line 446: ditto.

src/array/cuda/spmat_op_impl_csr.cu
```diff
@@ -423,26 +424,131 @@ template std::vector<NDArray> CSRGetDataAndIndices<kDGLCUDA, int64_t>(
...
+  key = key ^ (key >> 14);
+  key = (key + (key << 2)) + (key << 4);
+  key = key ^ (key >> 28);
+  key = key + (key << 31);
+  return key;
+}
+
+/**
+ * @brief A hashmap designed for CSRSliceMatrix, similar in function to set,
+ * Insert api inserts an element, Query api queries if an element is in the
```
frozenbugs, line 461: \`Insert\` to to refer to the code in the comment.

src/array/cuda/spmat_op_impl_csr.cu
frozenbugs: Add a detailed comment about those magic numbers. If needed put evidence in a google doc, and link the doc to the code.

frozenbugs:
Do you foresee any risk could be introduced by this PR?

gpzlx1:
@frozenbugs Theoretically, no risk is introduced. I have only modified the _SegmentMaskColKernel and the input and output of this kernel remains the same as the original version.

yaox12:
@dgl-bot

yaox12:
> @frozenbugs Can I use thrust in DGL?

Sure. Would it be better to use `thrust::cuda::par_nosync` as follows?
https://github.com/dmlc/dgl/blob/bf264d00fe1039eb588165aee6e414ef6ba3a08a/src/runtime/cuda/cuda_common.h#L24-L29

Since we launch thrust and _SegmentMaskColKernel on the same stream, the async policy should be safe.

gpzlx1:
> > @frozenbugs Can I use thrust in DGL?
> 
> Sure. Would it be better to use `thrust::cuda::par_nosync` as follows?
> 
> https://github.com/dmlc/dgl/blob/bf264d00fe1039eb588165aee6e414ef6ba3a08a/src/runtime/cuda/cuda_common.h#L24-L29
> 
> Since we launch thrust and _SegmentMaskColKernel on the same stream, the async policy should be safe.

Done.

yaox12:
@dgl-bot

yaox12:
LGTM. I got 1.2x - 142x speedup on A100 without obvious GPU memory overheads (both cost ~3677MB). It's impressive.

yaox12 (approved):

PR #4929, opened 2022-11-21
[Sparse] Support sparse format conversion

### LINKED ISSUE #4819: [Sparse] Support sparse format conversion.
## 🔨Work Item

**IMPORTANT:**

czkkkkkk:
## Description

Resolve #4819
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/dgl_sparse/include/sparse/sparse_format.h
+++ b/dgl_sparse/include/sparse/sparse_format.h
@@ -23,34 +23,31 @@ enum SparseFormat { kCOO, kCSR, kCSC };
 
 /** @brief COO sparse structure. */
 struct COO {
-  /** @brief the dense shape of the matrix. */
+  /** @brief The shape of the matrix. */
   int64_t num_rows = 0, num_cols = 0;
-  /** @brief COO format row array of the matrix. */
+  /** @brief COO format row indices array of the matrix. */
   torch::Tensor row;
-  /** @brief COO format column array of the matrix. */
+  /** @brief COO format column indices array of the matrix. */
   torch::Tensor col;
-  /** @brief data index tensor. When it is null, assume it is from 0 to NNZ - 1.
-   */
-  torch::optional<torch::Tensor> data;
-  /** @brief whether the row indices are sorted. */
+  /** @brief Whether the row indices are sorted. */
   bool row_sorted = false;
-  /** @brief whether the column indices per row are sorted. */
+  /** @brief Whether the column indices per row are sorted. */
   bool col_sorted = false;
 };
 
 
 /** @brief CSR sparse structure. */
 struct CSR {
-  /** @brief the dense shape of the matrix. */
+  /** @brief The dense shape of the matrix. */
   int64_t num_rows = 0, num_cols = 0;
   /** @brief CSR format index pointer array of the matrix. */
   torch::Tensor indptr;
   /** @brief CSR format index array of the matrix. */
   torch::Tensor indices;
-  /** @brief data index tensor. When it is null, assume it is from 0 to NNZ - 1.
+  /** @brief Data index tensor. When it is null, assume it is from 0 to NNZ - 1.
    */
-  torch::optional<torch::Tensor> data;
-  /** @brief whether the column indices per row are sorted. */
+  torch::optional<torch::Tensor> value_indices;
+  /** @brief Whether the column indices per row are sorted. */
   bool sorted = false;
 };
 
--- a/dgl_sparse/include/sparse/sparse_format.h
+++ b/dgl_sparse/include/sparse/sparse_format.h
@@ -18,7 +18,7 @@
 namespace dgl {
 namespace sparse {
 
-/** @brief SparseFormat enumeration */
+/** @brief SparseFormat enumeration. */
 enum SparseFormat { kCOO, kCSR, kCSC };
 
 /** @brief COO sparse structure. */
@@ -32,9 +32,9 @@ struct COO {
   /** @brief data index tensor. When it is null, assume it is from 0 to NNZ - 1.
    */
   torch::optional<torch::Tensor> data;
-  /** @brief whether the row indices are sorted */
+  /** @brief whether the row indices are sorted. */
   bool row_sorted = false;
-  /** @brief whether the column indices per row are sorted */
+  /** @brief whether the column indices per row are sorted. */
   bool col_sorted = false;
 };
 
--- a/dgl_sparse/include/sparse/sparse_format.h
+++ b/dgl_sparse/include/sparse/sparse_format.h
@@ -21,39 +21,68 @@ namespace sparse {
 /** @brief SparseFormat enumeration */
 enum SparseFormat { kCOO, kCSR, kCSC };
 
-/** @brief CSR sparse structure */
-struct CSR {
-  // CSR format index pointer array of the matrix
-  torch::Tensor indptr;
-  // CSR format index array of the matrix
-  torch::Tensor indices;
-  // The element order of the sparse format. In the SparseMatrix, we have data
-  // (value_) for each non-zero value. The order of non-zero values in (value_)
-  // may differ from the order of non-zero entries in CSR. So we store
-  // `value_indices` in CSR to indicate its relative non-zero value order to the
-  // SparseMatrix. With `value_indices`, we can retrieve the correct value for
-  // CSR, i.e., `value_[value_indices]`. If `value_indices` is not defined, this
-  // CSR follows the same non-zero value order as the SparseMatrix.
-  torch::optional<torch::Tensor> value_indices;
-};
-
-/** @brief COO sparse structure */
+/** @brief COO sparse structure. */
 struct COO {
-  // COO format row array of the matrix
+  /** @brief the dense shape of the matrix. */
+  int64_t num_rows = 0, num_cols = 0;
+  /** @brief COO format row array of the matrix. */
   torch::Tensor row;
-  // COO format column array of the matrix
+  /** @brief COO format column array of the matrix. */
   torch::Tensor col;
+  /** @brief data index tensor. When it is null, assume it is from 0 to NNZ - 1.
+   */
+  torch::optional<torch::Tensor> data;
+  /** @brief whether the row indices are sorted */
+  bool row_sorted = false;
+  /** @brief whether the column indices per row are sorted */
+  bool col_sorted = false;
 };
 
-/**
- * @brief Convert a CSR format to COO format
- * @param num_rows Number of rows of the sparse format
- * @param num_cols Number of cols of the sparse format
- * @param csr CSR sparse format
- * @return COO sparse format
- */
-std::shared_ptr<COO> CSRToCOO(
-    int64_t num_rows, int64_t num_cols, const std::shared_ptr<CSR> csr);
+
+/** @brief CSR sparse structure. */
+struct CSR {
+  /** @brief the dense shape of the matrix. */
+  int64_t num_rows = 0, num_cols = 0;
+  /** @brief CSR format index pointer array of the matrix. */
+  torch::Tensor indptr;
+  /** @brief CSR format index array of the matrix. */
+  torch::Tensor indices;
+  /** @brief data index tensor. When it is null, assume it is from 0 to NNZ - 1.
+   */
+  torch::optional<torch::Tensor> data;
+  /** @brief whether the column indices per row are sorted. */
+  bool sorted = false;
+};
+
+/** @brief Convert an old DGL COO format to a COO in the sparse library. */
+std::shared_ptr<COO> COOFromOldDGLCOO(const aten::COOMatrix& dgl_coo);
+
+/** @brief Convert a COO in the sparse library to an old DGL COO matrix. */
+aten::COOMatrix COOToOldDGLCOO(const std::shared_ptr<COO>& coo);
+
+/** @brief Convert an old DGL CSR format to a CSR in the sparse library. */
+std::shared_ptr<CSR> CSRFromOldDGLCSR(const aten::CSRMatrix& dgl_csr);
+
+/** @brief Convert a CSR in the sparse library to an old DGL CSR matrix. */
+aten::CSRMatrix CSRToOldDGLCSR(const std::shared_ptr<CSR>& csr);
+
+/** @brief Convert a CSR format to COO format. */
+std::shared_ptr<COO> CSRToCOO(const std::shared_ptr<CSR>& csr);
+
+/** @brief Convert a CSC format to COO format. */
+std::shared_ptr<COO> CSCToCOO(const std::shared_ptr<CSR>& csc);
+
+/** @brief Convert a COO format to CSR format. */
+std::shared_ptr<CSR> COOToCSR(const std::shared_ptr<COO>& coo);
+
+/** @brief Convert a CSC format to CSR format. */
+std::shared_ptr<CSR> CSCToCSR(const std::shared_ptr<CSR>& csc);
+
+/** @brief Convert a COO format to CSC format. */
+std::shared_ptr<CSR> COOToCSC(const std::shared_ptr<COO>& coo);
+
+/** @brief Convert a CSR format to CSC format. */
+std::shared_ptr<CSR> CSRToCSC(const std::shared_ptr<CSR>& csr);
 
 }  // namespace sparse
 }  // namespace dgl
--- a/dgl_sparse/src/sparse_format.cc
+++ b/dgl_sparse/src/sparse_format.cc
@@ -14,21 +14,77 @@
 namespace dgl {
 namespace sparse {
 
-std::shared_ptr<COO> CSRToCOO(
-    int64_t num_rows, int64_t num_cols, const std::shared_ptr<CSR> csr) {
-  auto indptr = TorchTensorToDGLArray(csr->indptr);
-  auto indices = TorchTensorToDGLArray(csr->indices);
-  bool data_as_order = false;
-  runtime::NDArray data = aten::NullArray();
-  if (csr->value_indices.has_value()) {
-    data_as_order = true;
-    data = TorchTensorToDGLArray(csr->value_indices.value());
-  }
-  auto dgl_csr = aten::CSRMatrix(num_rows, num_cols, indptr, indices, data);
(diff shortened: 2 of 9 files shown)


python/dgl/mock_sparse2/sparse_matrix.py
```diff
@@ -273,12 +271,12 @@ def create_from_csr(
                  values=tensor([1., 1., 1., 1., 1.]),
                  shape=(3, 3), nnz=5)
     >>> # Specify shape
-    >>> A = create_from_csr(indptr, indices, shape=(5, 3))
+    >>> A = create_from_csr(indptr, indices, shape=(3, 5))
     >>> print(A)
     SparseMatrix(indices=tensor([[0, 1, 2, 2, 2],
             [1, 2, 0, 1, 2]]),
```
mufeili, line 277: Perhaps fix the indent in printing.

python/dgl/mock_sparse2/sparse_matrix.py
```diff
@@ -273,12 +271,12 @@ def create_from_csr(
                  values=tensor([1., 1., 1., 1., 1.]),
                  shape=(3, 3), nnz=5)
     >>> # Specify shape
-    >>> A = create_from_csr(indptr, indices, shape=(5, 3))
+    >>> A = create_from_csr(indptr, indices, shape=(3, 5))
     >>> print(A)
     SparseMatrix(indices=tensor([[0, 1, 2, 2, 2],
```
mufeili, line 276: The indices here are still COO indices, right?

tests/pytorch/mock_sparse2/test_sparse_matrix.py
```diff
@@ -152,3 +159,196 @@ def test_csr_to_coo(dense_dim, indptr, indices, shape):
     assert torch.allclose(mat_row, row)
     assert torch.allclose(mat_col, col)
 
+
+@pytest.mark.parametrize("dense_dim", [None, 4])
+@pytest.mark.parametrize("indptr", [[0, 0, 1, 4], (0, 1, 2, 4)])
```
mufeili, line 164: Better to be consistent in using `[]` and `()`

tests/pytorch/mock_sparse2/test_sparse_matrix.py
```diff
@@ -152,3 +159,196 @@ def test_csr_to_coo(dense_dim, indptr, indices, shape):
...
+
+def _scatter_add(a, index, v=1):
+    index = index.tolist()
+    for i in index:
+        a[i] += v
+    return a
+
+
+@pytest.mark.parametrize("dense_dim", [None, 4])
+@pytest.mark.parametrize("row", [[0, 0, 1, 2], (0, 1, 2, 4)])
```
mufeili, line 205: Ditto

tests/pytorch/mock_sparse2/test_sparse_matrix.py
```diff
@@ -152,3 +159,196 @@ def test_csr_to_coo(dense_dim, indptr, indices, shape):
...
+    assert mat.shape == shape
+    assert mat.nnz == row.numel()
+    assert mat.dtype == val.dtype
+    assert torch.allclose(mat_val, val)
+    assert torch.allclose(mat_indptr, indptr)
+    assert torch.allclose(mat_indices, indices)
+
+
+@pytest.mark.parametrize("dense_dim", [None, 4])
+@pytest.mark.parametrize("indptr", [[0, 0, 1, 4], (0, 1, 2, 4)])
```
mufeili, line 240: Ditto

tests/pytorch/mock_sparse2/test_sparse_matrix.py
```diff
@@ -152,3 +159,196 @@ def test_csr_to_coo(dense_dim, indptr, indices, shape):
...
+    assert mat.nnz == row.numel()
+    assert mat.device == row.device
+    assert mat.dtype == val.dtype
+    assert torch.allclose(mat_val, val)
+    assert torch.allclose(mat_indptr, indptr)
+    assert torch.allclose(mat_indices, indices)
+
+
+@pytest.mark.parametrize("dense_dim", [None, 4])
+@pytest.mark.parametrize("row", [[0, 0, 1, 2], (0, 1, 2, 4)])
```
mufeili, line 281: Ditto

dgl_sparse/include/sparse/sparse_format.h
```diff
@@ -21,39 +21,68 @@ namespace sparse {
...
-  // may differ from the order of non-zero entries in CSR. So we store
-  // `value_indices` in CSR to indicate its relative non-zero value order to the
-  // SparseMatrix. With `value_indices`, we can retrieve the correct value for
-  // CSR, i.e., `value_[value_indices]`. If `value_indices` is not defined, this
-  // CSR follows the same non-zero value order as the SparseMatrix.
-  torch::optional<torch::Tensor> value_indices;
+  /** @brief data index tensor. When it is null, assume it is from 0 to NNZ - 1.
+   */
+  torch::optional<torch::Tensor> data;
+  /** @brief whether the column indices per row are sorted. */
```
mufeili, line 53: "." at the end here, but not in `COO`

mufeili:
Should we also test for the conversion between old DGL sparse formats and these new sparse formats?

dgl_sparse/src/sparse_format.cc
```diff
@@ -14,21 +14,77 @@
...
+  auto indices = TorchTensorToDGLArray(csr->indices);
+  auto data = OptionalTorchTensorToDGLArray(csr->data);
+  return aten::CSRMatrix(
+      csr->num_rows, csr->num_cols, indptr, indices, data, csr->sorted);
+}
+
+std::shared_ptr<COO> CSRToCOO(const std::shared_ptr<CSR>& csr) {
+  auto dgl_csr = CSRToOldDGLCSR(csr);
+  auto dgl_coo = aten::CSRToCOO(dgl_csr, csr->data.has_value());
+  return COOFromOldDGLCOO(dgl_coo);
```
mufeili, line 55: Have you discussed with other people about how to implement these?

mufeili, line 55: Have you discussed with other people about how to implement these?

dgl_sparse/src/utils.h
```diff
@@ -61,6 +61,24 @@ inline static torch::Tensor DGLArrayToTorchTensor(runtime::NDArray array) {
   return at::fromDLPack(runtime::DLPackConvert::ToDLPack(array));
 }
 
+/** @brief Convert a optional Torch tensor to a DGL array. */
```
mufeili, line 64:
```suggestion
/** @brief Convert an optional Torch tensor to a DGL array. */
```

dgl_sparse/src/utils.h
```diff
@@ -61,6 +61,24 @@ inline static torch::Tensor DGLArrayToTorchTensor(runtime::NDArray array) {
...
+inline static runtime::NDArray OptionalTorchTensorToDGLArray(
+    torch::optional<torch::Tensor> tensor) {
+  if (!tensor.has_value()) {
+    return aten::NullArray();
+  }
+  return TorchTensorToDGLArray(tensor.value());
+}
+
+/** @brief Convert a DGL array to an option Torch tensor. */
```
mufeili, line 73:
```suggestion
/** @brief Convert a DGL array to an optional Torch tensor. */
```

mufeili:
Done a first pass

python/dgl/mock_sparse2/sparse_matrix.py
czkkkkkk: Yes.

dgl_sparse/src/sparse_format.cc
czkkkkkk: These functions convert DGL sparse formats to/from the formats in the sparse library. The purpose is to reuse the implementations in current DGL. What is you concern?

python/dgl/mock_sparse2/sparse_matrix.py
czkkkkkk: Fixed.

dgl_sparse/include/sparse/sparse_format.h
czkkkkkk: Done

czkkkkkk:
> Should we also have independent tests for the conversion between old DGL sparse formats and these new sparse formats?

I think it is not necessary because they have the same structures except the new sparse formats use Pytorch tensors to store data.

dgl_sparse/src/sparse_format.cc
mufeili: I'm fine with reusing the existing implementations, just want to know if this is desired.

python/dgl/mock_sparse2/sparse_matrix.py
```diff
@@ -186,28 +186,28 @@ def create_from_coo(
     >>> A = create_from_coo(dst, src)
     >>> print(A)
     SparseMatrix(indices=tensor([[1, 1, 2],
-                                 [2, 4, 3]]),
-                 values=tensor([1., 1., 1.]),
-                 shape=(3, 5), nnz=3)
+            [2, 4, 3]]),
```
mufeili, line 189: Can we fix improve the printing of the matrix to have the display on the left?

mufeili (approved):

python/dgl/mock_sparse2/sparse_matrix.py
czkkkkkk: It is controlled by Pytorch. Even a simple 2-D tensor will be printed with indentation.
```
>>> a=torch.tensor([[1,2,3],[4,5,6]])
>>> print(a)
tensor([[1, 2, 3],
        [4, 5, 6]])
```

dgl_sparse/src/sparse_format.cc
czkkkkkk: I see. That is the purpose to link to libdgl.so. We can reduce our development effort and retain the same performance as the message passing interfaces.

python/dgl/mock_sparse2/sparse_matrix.py
mufeili: I actually want proper `indentation` when printing a sparse matrix, i.e.

```
SparseMatrix(indices=tensor([[1, 1, 2],
                                                    [2, 4, 3]]),
                        ...)
```

is preferred over

```
SparseMatrix(indices=tensor([[1, 1, 2],
               [2, 4, 3]]),
                        ...)
```

dgl_sparse/include/sparse/sparse_format.h
```diff
@@ -18,42 +18,71 @@
...
 namespace sparse {
 
-/** @brief SparseFormat enumeration */
+/** @brief SparseFormat enumeration. */
 enum SparseFormat { kCOO, kCSR, kCSC };
 
-/** @brief CSR sparse structure */
+/** @brief COO sparse structure. */
+struct COO {
+  /** @brief the dense shape of the matrix. */
```
jermainewang, line 26: * All comments should be a valid sentence, e.g., capitalize the first word.
* There is no difference between dense or sparse shape so just say "The shape of the matrix." is fine.

dgl_sparse/include/sparse/sparse_format.h
```diff
@@ -18,42 +18,71 @@
...
+  int64_t num_rows = 0, num_cols = 0;
+  /** @brief COO format row array of the matrix. */
```
jermainewang, line 28: `Row indices array.`

dgl_sparse/include/sparse/sparse_format.h
```diff
@@ -18,42 +18,71 @@
...
+  torch::Tensor row;
+  /** @brief COO format column array of the matrix. */
```
jermainewang, line 30: `Column indices array.`

dgl_sparse/include/sparse/sparse_format.h
```diff
@@ -18,42 +18,71 @@
...
+  torch::Tensor col;
+  /** @brief data index tensor. When it is null, assume it is from 0 to NNZ - 1.
+   */
+  torch::optional<torch::Tensor> data;
```
jermainewang, line 34: COO matrix shall not have `data` because the values are stored in the same order of row/col indices. To ensure there is no data corruption, you can check the `data` array of the old `COOMatrix` of DGL must be null after format conversion.

dgl_sparse/include/sparse/sparse_format.h
```diff
@@ -18,42 +18,71 @@
...
-  // (value_) for each non-zero value. The order of non-zero values in (value_)
-  // may differ from the order of non-zero entries in CSR. So we store
-  // `value_indices` in CSR to indicate its relative non-zero value order to the
-  // SparseMatrix. With `value_indices`, we can retrieve the correct value for
-  // CSR, i.e., `value_[value_indices]`. If `value_indices` is not defined, this
-  // CSR follows the same non-zero value order as the SparseMatrix.
-  torch::optional<torch::Tensor> value_indices;
+  /** @brief data index tensor. When it is null, assume it is from 0 to NNZ - 1.
+   */
+  torch::optional<torch::Tensor> data;
```
jermainewang, line 52: Why change it to `data`? `value_indices` is a more intuitive name.

dgl_sparse/src/sparse_matrix.cc
```diff
@@ -96,65 +96,89 @@ std::shared_ptr<CSR> SparseMatrix::CSCPtr() {
 std::vector<torch::Tensor> SparseMatrix::COOTensors() {
   auto coo = COOPtr();
   auto val = value();
+  if (coo->data.has_value()) {
+    val = val.index({coo->data.value()});
+  }
```
jermainewang, line 101: I still don't think this is the intended behavior. `COOTensors()` just need to return row and col. `CSRTensors()` returns indptr, indices and value_indices. On the user side, if they want value, they just get it by `A.val`, so it is kinda redundant to return the value tensors here.

tests/pytorch/mock_sparse2/test_sparse_matrix.py
```diff
@@ -10,15 +10,14 @@
...
-@pytest.mark.parametrize("row", [[0, 0, 1, 2], (0, 1, 2, 4)])
+@pytest.mark.parametrize("row", [(0, 0, 1, 2), (0, 1, 2, 4)])
 @pytest.mark.parametrize("col", [(0, 1, 2, 2), (1, 3, 3, 4)])
 @pytest.mark.parametrize("shape", [None, (3, 5), (5, 3)])
 def test_create_from_coo(dense_dim, row, col, shape):
     # Skip invalid matrices
-    if shape is not None and (
-        max(row) >= shape[0] or max(col) >= shape[1]
-    ):
+    if shape is not None and (max(row) >= shape[0] or max(col) >= shape[1]):
```
jermainewang, line 20: I don't like this. Can we adjust the test cases to rule out invalid combinations?

tests/pytorch/mock_sparse2/test_sparse_matrix.py
```diff
@@ -10,15 +10,14 @@
 if not sys.platform.startswith("linux"):
     pytest.skip("skipping tests on win", allow_module_level=True)
 
+
 @pytest.mark.parametrize("dense_dim", [None, 4])
-@pytest.mark.parametrize("row", [[0, 0, 1, 2], (0, 1, 2, 4)])
+@pytest.mark.parametrize("row", [(0, 0, 1, 2), (0, 1, 2, 4)])
 @pytest.mark.parametrize("col", [(0, 1, 2, 2), (1, 3, 3, 4)])
 @pytest.mark.parametrize("shape", [None, (3, 5), (5, 3)])
 def test_create_from_coo(dense_dim, row, col, shape):
```
jermainewang, line 18: I further suggest add test for autograd for the create ops. We only need to test gradient is correctly passed backwardly for the value array.

jermainewang (changes requested):

dgl_sparse/include/sparse/sparse_format.h
czkkkkkk: I added `data` in COO to make it consistent with the `COOMatrix` of DGL. Is it possible that we create a sparse format from a CSC and somehow it goes though `CSC->CSR` and `CSR->COO` conversions. Does DGL guarantee the order of the final COO follow the same value order of the CSC?

dgl_sparse/include/sparse/sparse_format.h
czkkkkkk: Just follow the name in old DGL. Will change them to `value_indices`.

dgl_sparse/src/sparse_matrix.cc
czkkkkkk: These interfaces return `value` here. Do you mean it is better to return `value_indices`?

tests/pytorch/mock_sparse2/test_sparse_matrix.py
czkkkkkk: I think it is not necessary to test autograd here. The creation of a sparse matrix does not have autograd since Pytorch only compute autograd on tensors. For example, 
```
A = create_from_coo(row, col, val)
B = create_from_coo(row, col, va1)
C = A + B
```
will compute autograd on `val + val1` because in our underlying code, we call the `val + val1` from libtorch. The sparse format creation itself does not have autograd.

dgl_sparse/include/sparse/sparse_format.h
czkkkkkk: I got you point. In the case of `CSC->CSR` and `CSR->COO`, the final COO should be ordered by the `value_indices` in the `CSR`, resulting the same value order as the original CSC.

However, I still think removing `value_indices` in the COO will make our implementation not flexible. For example, if we want to sort the COO by column, we need to create a new sparse matrix.

czkkkkkk: I am convinced to remove `value_indices` in the COO. Otherwise, it will make `mat.coo()` become more complex.

python/dgl/mock_sparse2/sparse_matrix.py
jermainewang: Agree that the current printing is not ideal though not the focus of this PR.

jermainewang (approved):

PR #4936, opened 2022-11-22
[Bug] fix the test_hgt function in test_nn.py

kunmukh:
## Description
The issue is with the mini-batch code. The block's  `num_dst_nodes` can sometimes be batch_size - 1, but the feature size is always batch_size. The test failed when the num_dst_nodes and feature sizes differed. II fixed it by adding a conditional to check if the size of the num_dst_nodes is the same as the batch size first.

## Checklist
- [x] Code is well-documented

## Changes
[x] separate input and output node types and handle them correctly in hgtconv.py
```
    if block.number_of_dst_nodes() == batch_size:
        y = m(block, x, ntype, etype)
```


tests/pytorch/test_nn.py
```diff
@@ -1519,16 +1520,17 @@ def test_hgt(idtype, in_size, num_heads):
...
     (input_nodes, output_nodes, block) = next(iter(train_loader))
     block = block[0]
     x = x[input_nodes.to(th.long)]
     ntype = ntype[input_nodes.to(th.long)]
     edge = block.edata[dgl.EID]
     etype = etype[edge.to(th.long)]
-    y = m(block, x, ntype, etype)
-    assert y.shape == (block.number_of_dst_nodes(), head_size * num_heads)
+    if block.number_of_dst_nodes() == batch_size:
+        y = m(block, x, ntype, etype)
```
yaox12, line 1532: But it means the sampled block cannot be used in training sometimes, right? If so, I believe there is a bug in the sampling algorithm.

mufeili: @BarclayII Any thoughts?

mufeili: @BarclayII Any thoughts?

BarclayII: I don't think this is the correct fix.  See #4947.

BarclayII:
Closing this PR since the fix is not correct.  The problem lies in how the unit test generates the indices, not in how the HGTConv module handles the input.  See #4947 for more details.

PR #4963, opened 2022-11-28
[Sparse] C&S Example

mufeili:
## Description
Accuracy is about 0.74~0.75

## Checklist
- [x] Code is well-documented


examples/sparse/c_and_s.py
```diff
@@ -0,0 +1,143 @@
...
+    def __init__(self, num_layers=20, alpha=0.9):
+        super().__init__()
+
+        self.num_layers = num_layers
+        self.alpha = alpha
+
+    ###########################################################################
+    # (HIGHLIGHT) Compute Label Propagation with Sparse Matrix API
+    ###########################################################################
+    @torch.no_grad()
```
jermainewang, line 24: Shall we make this a preprocessing step like SIGN?

mufeili: That's definitely feasible. A follow-up question is weather we still want to have a `CorrectAndSmooth` class or simply two functions `correct` and `smooth`.

examples/sparse/c_and_s.py
```diff
@@ -0,0 +1,143 @@
...
+        print(
+            f"Base model, In epoch {epoch}, loss: {loss:.3f}, "
+            f"val acc: {val_acc:.3f}, test acc: {test_acc:.3f}"
+        )
+
+    # Stage2: Correct and Smooth.
+    soft_label = F.softmax(logits, dim=1)
+    label = F.one_hot(label)
+    soft_label = model.correct(A_hat, label, soft_label, train_mask)
+    soft_label = model.smooth(A_hat, label, soft_label, train_mask)
```
jermainewang, line 109: Let's write them as plain functions:
```
soft_label = correct(A_hat, label, soft_label, train_mask)
soft_label = smooth(A_hat, label, soft_label, train_mask)
```

In this way, you don't need to pass the `model` object anymore and you can rename `base_model` to `model`.

examples/sparse/c_and_s.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # Create models.
+    X = g.ndata["feat"]
+    in_size = X.shape[1]
+    out_size = dataset.num_classes
+    base_model = nn.Linear(in_size, out_size).to(dev)
+    model = CorrectAndSmooth().to(dev)
+
+    # Kick off training.
+    train(base_model, model, g, A_hat, X)
```
jermainewang, line 143: How about we further split c&s with `train`? This makes it very clear that the base models are trained without graph while the post-process step needs graph.

frozenbugs (approved):

Issue #3444, opened 2021-10-26
`adj_product_graph` crashes with large number of edges.


BarclayII:
## 🐛 Bug

The following code crashes:
```python
import dgl
import torch
C = dgl.graph((torch.randint(0, 9000, (10000000,)), torch.randint(0, 9000, (10000000,))), num_nodes=9000)
C = dgl.to_simple(C)
C = C.to('cuda')
C.edata['w'] = torch.randn(C.num_edges()).cuda()
dgl.adj_product_graph(C, C, 'w')
```

Traceback looks like this, suggesting that the error is `CUSPARSE_STATUS_INSUFFICIENT_RESOURCES` thrown at the second `cusparseSpGEMM_workEstimation` call in https://github.com/dmlc/dgl/blob/579cd3eb49a43717072627440afc1b66f3938adc/src/array/cuda/csr_mm.cu#L87.  However I checked that the workspace allocated is around 100M which should be manageable.  Not sure what is the problem.

```
Traceback (most recent call last):
  File "a.py", line 18, in <module>
    dgl.adj_product_graph(C, C, 'w')
  File "/home/ubuntu/miniconda3/lib/python3.8/site-packages/dgl-0.8-py3.8-linux-x86_64.egg/dgl/transform.py", line 2641, in adj_product_graph
    C_gidx, C_weights = F.csrmm(
  File "/home/ubuntu/miniconda3/lib/python3.8/site-packages/dgl-0.8-py3.8-linux-x86_64.egg/dgl/backend/pytorch/sparse.py", line 647, in csrmm
    CSRMM.apply(gidxA, A_weights, gidxB, B_weights, num_vtypes)
  File "/home/ubuntu/miniconda3/lib/python3.8/site-packages/dgl-0.8-py3.8-linux-x86_64.egg/dgl/backend/pytorch/sparse.py", line 534, in forward
    gidxC, C_weights = _csrmm(gidxA, A_weights, gidxB, B_weights, num_vtypes)
  File "/home/ubuntu/miniconda3/lib/python3.8/site-packages/dgl-0.8-py3.8-linux-x86_64.egg/dgl/sparse.py", line 540, in _csrmm
    C, C_weights = _CAPI_DGLCSRMM(
  File "/home/ubuntu/miniconda3/lib/python3.8/site-packages/dgl-0.8-py3.8-linux-x86_64.egg/dgl/_ffi/_ctypes/function.py", line 188, in __call__
[...]

PR #4974, opened 2022-12-01
Add support for next cusparse release

chang-l:
## Description
Since next cuda release will deprecate `cusparse<t>csrgemm` functions, DGL build will fail by the time. I noticed that we choose to use old API instead of new `cusparseSpgemm` due to this failure https://github.com/dmlc/dgl/issues/3444. 

This PR re-enables  `cusparseSpgemm` API for `__CUDACC_VER_MAJOR__> 12 `, while ~~tentatively sticks to CUSPARSE_SPGEMM_ALG3~~ adaptively picks different algorithms  to alleviate memory pressure as well as avoid compromising speed a lot. As such, it can pass this case: https://github.com/dmlc/dgl/issues/3444

Note that the algorithm flags `CUSPARSE_SPGEMM_ALG3` and `CUSPARSE_SPGEMM_ALG2` will be introduced in the next release, because of lack of documentation at this moment, I can briefly describe perf. diff among cusparse spgemm algorithms (`default`, `alg2`, `alg3`):
- In terms of speed (time): `default` > `alg2` > `alg3`; here `>` means faster (better).
- In terms of mem. consumption:  `alg3` > `alg2` > `default`; here `>` means less mem. (better).

~~By using `CUSPARSE_SPGEMM_ALG3`, this PR is very conservative in terms of memory usage in trade-off of (speed)slowness.~~ 
Now, depending the number of products (returned by `cusparseSpGEMM_getNumProducts`), from small problems to large problems, `CUSPARSE_SPGEMM_DEFAULT`, `CUSPARSE_SPGEMM_ALG2` and `CUSPARSE_SPGEMM_ALG3` will be used in order. Also, to avoid algorithm switching overhead for small problems, it starts with `CUSPARSE_SPGEMM_DEFAULT` and falls back to ALG2/3 when necessary (large enough) as early as possible. As a result, for small problems, we observed very similar performance compared to the old API; however, for very large problems like https://github.com/dmlc/dgl/issues/3444, it could be 2-3X slower...

The heuristics of alg. switching are tuned based on a machine with at least 16G device memory, to ensure best dynamic range. We need to stay tuned on this spgemm use case as `cusparseSpgemm` can continuously improve over the time. 

## Checklist
- [x] Code is well-documented

### DIFF
--- a/src/array/cuda/csr_mm.cu
+++ b/src/array/cuda/csr_mm.cu
@@ -6,10 +6,11 @@
 #include <dgl/array.h>
 #include <dgl/runtime/device_api.h>
 
+#include <limits>
+
 #include "../../runtime/cuda/cuda_common.h"
 #include "./cusparse_dispatcher.cuh"
 #include "./functor.cuh"
-#include <limits>
 namespace dgl {
 
 using namespace dgl::runtime;
@@ -73,7 +74,7 @@ std::pair<CSRMatrix, NDArray> CusparseSpgemm(
   // below to estimate upperbound of num_prods to decide DEFAULT or ALG2
   int int_max = std::numeric_limits<int>::max();
   int64_t nnzAB_norm = (nnzA / A.num_rows) * (nnzB / B.num_rows);
-  int64_t dense_norm = B.num_cols;  //if denseMM, numprods = nrowA*ncolA*ncolB
+  int64_t dense_norm = B.num_cols;  // if denseMM, numprods=nrowA*ncolA*ncolB
   int64_t max_nprods = (nnzAB_norm > dense_norm) ? dense_norm : nnzAB_norm;
   max_nprods = max_nprods * A.num_rows * B.num_rows;
   cusparseSpGEMMAlg_t alg;
@@ -112,7 +113,7 @@ std::pair<CSRMatrix, NDArray> CusparseSpgemm(
     device->FreeWorkspace(ctx, workspace1);
     // rerun cusparseSpGEMM_workEstimation
     CUSPARSE_CALL(cusparseSpGEMM_workEstimation(thr_entry->cusparse_handle,
-                                                transA,transB, &alpha, matA,
+                                                transA, transB, &alpha, matA,
                                                 matB, &beta, matC, dtype, alg,
                                                 spgemmDesc, &workspace_size1,
                                                 NULL));
@@ -131,7 +132,7 @@ std::pair<CSRMatrix, NDArray> CusparseSpgemm(
     device->FreeWorkspace(ctx, workspace1);
     // rerun cusparseSpGEMM_workEstimation
     CUSPARSE_CALL(cusparseSpGEMM_workEstimation(thr_entry->cusparse_handle,
-                                                transA,transB, &alpha, matA,
+                                                transA, transB, &alpha, matA,
                                                 matB, &beta, matC, dtype, alg,
                                                 spgemmDesc, &workspace_size1,
                                                 NULL));
--- a/src/array/cuda/csr_mm.cu
+++ b/src/array/cuda/csr_mm.cu
@@ -9,7 +9,7 @@
 #include "../../runtime/cuda/cuda_common.h"
 #include "./cusparse_dispatcher.cuh"
 #include "./functor.cuh"
-
+#include <limits>
 namespace dgl {
 
 using namespace dgl::runtime;
@@ -51,8 +51,6 @@ std::pair<CSRMatrix, NDArray> CusparseSpgemm(
   IdType* dC_csrOffsets_data = dC_csrOffsets.Ptr<IdType>();
   constexpr auto idtype = cusparse_idtype<IdType>::value;
   constexpr auto dtype = cuda_dtype<DType>::value;
-  cusparseSpGEMMAlg_t alg = CUSPARSE_SPGEMM_ALG3;
-
   // Create sparse matrix A, B and C in CSR format
   CUSPARSE_CALL(cusparseCreateCsr(
       &matA, A.num_rows, A.num_cols, nnzA, A.indptr.Ptr<IdType>(),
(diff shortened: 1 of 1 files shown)


BarclayII:
I think the proposed solution is OK.  Essentially not choosing `ALG3` will make SPSPMM stop working and we don't have any other options for CUDA 12, and the solution is only limited to CUDA 12 only.

jermainewang:
> By using `CUSPARSE_SPGEMM_ALG3`, this PR is very conservative in terms of memory usage in trade-off of (speed)slowness. I wonder how bad impact of using relative slow spgemm routine (couple of secs per call for very large problems), given that `dgl.adj_product_graph` seems the only use case?

spspmm will be used in the new sparse matrix library. For example, multiplying two adjacency matrix `A @ A` to create a second-order adj matrix.

chang-l:
@jermainewang 
Thanks for the explanation. 

This PR is ready for review. @BarclayII @czkkkkkk @yaox12

src/array/cuda/csr_mm.cu
```diff
@@ -6,21 +6,21 @@
...
 namespace dgl {
 
 using namespace dgl::runtime;
 
 namespace aten {
 namespace cusparse {
 
-#if 0  // disabling CUDA 11.0+ implementation for now because of problems on
-       // bigger graphs
+#if __CUDACC_VER_MAJOR__ == 12
```
yaox12, line 21: Would it be better to use the runtime macro `CUDART_VERSION`? But to be honest, I'm not sure of their difference in switching library versions.

src/array/cuda/csr_mm.cu
```diff
@@ -110,7 +190,7 @@ std::pair<CSRMatrix, NDArray> CusparseSpgemm(
   // copy the final products to the matrix C
   CUSPARSE_CALL(cusparseSpGEMM_copy(
       thr_entry->cusparse_handle, transA, transB, &alpha, matA, matB, &beta,
-      matC, dtype, CUSPARSE_SPGEMM_DEFAULT, spgemmDesc));
+      matC, dtype, alg, spgemmDesc));
 
   device->FreeWorkspace(ctx, workspace1);
```
yaox12, line 195: Can we free `workspace1` before the actual computing?

src/array/cuda/csr_mm.cu
```diff
@@ -70,30 +70,110 @@ std::pair<CSRMatrix, NDArray> CusparseSpgemm(
...
+  // user-defined large problem size (above will use ALG3)
+  int64_t LARGE_NUM_PRODUCTS  = 800*1000*1000;
+
+  // switch to ALG2/ALG3 for medium & large problem size
+  if (alg == CUSPARSE_SPGEMM_DEFAULT && num_prods > MEDIUM_NUM_PRODUCTS) {
+    // use ALG3 for very large problem
+    if (num_prods <= LARGE_NUM_PRODUCTS)
+      alg = CUSPARSE_SPGEMM_ALG2;
+    else
+      alg = CUSPARSE_SPGEMM_ALG3;
```
yaox12, line 127: The condition operator `? :` looks better.

src/array/cuda/csr_mm.cu
chang-l: The total number of products in the SpGEMM and some other internal information are stored in `workspace1`. So, we cannot free `workspace1` until the end of SpGEMM computation.

src/array/cuda/csr_mm.cu
chang-l: I was not sure too... `CUDART_VERSION` seems get runtime version, while `__CUDACC_VER_MAJOR__ `  is defined with the major version number of nvcc when compiling. Thus, I agree `CUDART_VERSION` is better, and I will switch to `CUDART_VERSION`, also in consistent with other places in the code.

src/array/cuda/csr_mm.cu
chang-l: Updated.

yaox12 (approved):
LGTM.

czkkkkkk:
Is there a test covering this function?

chang-l:
> Is there a test covering this function?

Yes, the unit test is located within `compute/test_csrmm.py`

src/array/cuda/csr_mm.cu
```diff
@@ -70,30 +70,109 @@ std::pair<CSRMatrix, NDArray> CusparseSpgemm(
...
+    // fall back to ALG2 to estimate num_prods
+    alg = CUSPARSE_SPGEMM_ALG2;
+    device->FreeWorkspace(ctx, workspace1);
+    // rerun cusparseSpGEMM_workEstimation
+    CUSPARSE_CALL(cusparseSpGEMM_workEstimation(thr_entry->cusparse_handle,
+                                                transA, transB, &alpha, matA,
+                                                matB, &beta, matC, dtype, alg,
+                                                spgemmDesc, &workspace_size1,
+                                                NULL));
+    void* workspace1 = (device->AllocWorkspace(ctx, workspace_size1));
```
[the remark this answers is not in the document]
chang-l: Yeah... Though it runs well with various problem sizes, I think it could result in memory leak and I should not define it again locally. I will make the change later. 
Thank you for catching it.

src/array/cuda/csr_mm.cu
```diff
@@ -70,30 +70,109 @@ std::pair<CSRMatrix, NDArray> CusparseSpgemm(
...
+                                                transA, transB, &alpha, matA,
+                                                matB, &beta, matC, dtype, alg,
+                                                spgemmDesc, &workspace_size1,
+                                                workspace1));
+  } else {
+    CHECK(e == CUSPARSE_STATUS_SUCCESS) << "CUSPARSE ERROR: " << e;
+  }
+
+  // you can print num_prods, for better memory estimation
+  int64_t num_prods;
```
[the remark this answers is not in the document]
chang-l: Updated comment. 
It should be obvious that `num_prods` represents `number of intermediate products`  now.

Issue #4969, opened 2022-11-29
Possible example GCN mistake


dylanvanw:
## ❓ Question

I was looking at the example for GCN and found a possible mistake in the code. I opened this issue mostly to understand if the code I found is there with a specific reason or that it should not be there.
The example I am talking about: https://github.com/dmlc/dgl/blob/master/examples/pytorch/gcn/train.py

I noticed that there is a normalisation step on [line 97](https://github.com/dmlc/dgl/blob/master/examples/pytorch/gcn/train.py#L97) while the normalised values are not used downstream. Even if this was meant to show the normalisation step, it would not be calculating the normalisation step described in the [CGN paper](https://arxiv.org/abs/1609.02907). The paper considers both in and out degrees while the normalisation in the code only describes normalisation using the in degrees.
In the end, the normalised values are assigned to `g.ndata["norm"]` but these values are not used afterwards.

Having a normalisation step here is also unnecessary since the GraphConv layer that is used already takes care of the normalisation. https://docs.dgl.ai/en/0.9.x/_modules/dgl/nn/pytorch/conv/graphconv.html#GraphConv

Should that normalisation step still exist in the example? Is there a specific reason for that to be there that I am unaware of? If not, I can create a PR to remove it.

rudongyu:
Hi, @dylanvanw. You're right. The pre-computed norm is unnecessary there. You're appreciated to create a PR to remove it. Thanks a lot!

dylanvanw:
Hi @rudongyu! Thanks for the response. I create the PR for the changes. #4976

PR #4976, opened 2022-12-02
[Examples] GCN example correction (#4969)

dylanvanw:
Correction as mentioned in #4969 

I noticed that there is a normalisation step on line 97 while the normalised values are not used downstream. Even if this was meant to show the normalisation step, it would not be calculating the normalisation step described in the CGN paper. The paper considers both in and out degrees while the normalisation in the code only describes normalisation using the in degrees. In the end, the normalised values are assigned to g.ndata["norm"] but these values are not used afterwards.

Having a normalisation step here is also unnecessary since the GraphConv layer that is used already takes care of the normalisation. https://docs.dgl.ai/en/0.9.x/_modules/dgl/nn/pytorch/conv/graphconv.html#GraphConv

It confused me for a second thinking that I had to do the normalisation myself but this is already handled by the GraphConf.

### DIFF
--- a/examples/pytorch/gcn/train.py
+++ b/examples/pytorch/gcn/train.py
@@ -94,12 +94,6 @@ if __name__ == "__main__":
     labels = g.ndata["label"]
     masks = g.ndata["train_mask"], g.ndata["val_mask"], g.ndata["test_mask"]
 
-    # normalization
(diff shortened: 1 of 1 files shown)


rudongyu:
@dgl-bot

rudongyu (approved):

PR #4977, opened 2022-12-02
[Lint] Add a formatter GitHub Action

yaox12:
## Description

Leverage [lintrunner](https://github.com/pytorch/pytorch/wiki/lintrunner) to do format checks and suggestions for PRs via GitHub Action.

1. Only files modified in the current PR will be checked.
2. Use [UFMT](https://github.com/omnilib/ufmt), built on top of [black](https://black.readthedocs.io/) and [µsort](https://usort.readthedocs.io/) (an alternative to isort, see [here](https://usort.readthedocs.io/en/stable/why.html#comparison-to-isort) for comparison), to format Python code.
3. Use `clang-format` to format C/C++/CUDA code.

This action will post warnings/errors directly on the corresponding files :) See [here](https://github.com/yaox12/dgl/pull/57/files#diff-e22baf356d5f5ca22eb63e132344c6f3f92a94941ee5bc1ee3947929a6e5f89a) for example.

## Checklist
- [ ] Code is well-documented


.lintrunner.toml
```diff
@@ -0,0 +1,55 @@
...
+]
+is_formatter = true
+
+[[linter]]
+code = 'CLANGFORMAT'
+include_patterns = [
+    '**/*.h',
+    '**/*.cc',
+    '**/*.cuh',
+    '**/*.cu',
```
frozenbugs, line 38: should we also include:
*.c
*.cpp

frozenbugs:
I really like this!
But it seems that in the example, the warning only appears at the top of the file, and the warning is very generic.
Is it possible to have it the specific line? and call out the exact warning?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,54 @@
...
+        continue-on-error: true
+        run: |
+          # Use jq to massage the JSON lint output into GitHub Actions workflow commands.
+          jq --raw-output \
+            '"::\(if .severity == "advice" or .severity == "disabled" then "warning" else .severity end) file=\(.path),line=\(.line),col=\(.char),title=\(.code) \(.name)::" + (.description | gsub("\\n"; "%0A"))' \
+            lint.json
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}-${{ github.event_name == 'workflow_dispatch' }}
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
yaox12: Updated.

.lintrunner.toml
yaox12: Updated.

yaox12:
> I really like this! But it seems that in the example, the warning only appears at the top of the file, and the warning is very generic. Is it possible to have it the specific line? and call out the exact warning?

1. It should appear in the first line that needs to be changed now.
2. Since UFMT and clang-format are both formatters, it's hard to give exact warnings. Instead, they will give diffs before/after formatting. Click the upper right icon to see details. 
image

frozenbugs (approved):

PR #4983, opened 2022-12-05
[Example] Add UniGNN example and a hypergraph dataset example

BarclayII:
## Description
UniGNN example from the [UniGNN](https://arxiv.org/abs/2105.00956) paper.  This also shows how to run hypergraph neural networks with DGL's message passing interface.

It also adds the Cora Cocitation Hypergraph from the [HyperGCN](https://arxiv.org/abs/1809.02589) paper, which is what UniGNN uses.  It is actually different from the Cora dataset we are using.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/examples/sparse/unignn.py
+++ b/examples/sparse/unignn.py
@@ -1,0 +1,276 @@
+"""`UniGNN <https://arxiv.org/abs/2105.00956>`_
+
+This script reproduces the result from the
+`official repository <https://github.com/OneForward/UniGNN>`_
+with the following command:
+
+.. code::
+
+   python train.py --data=cocitation --dataset=cora --model-name=UniGAT
+"""
+import pickle
+import os
+import contextlib
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import tqdm
+import dgl
+import dgl.data.utils
+import dgl.nn
+import dgl.function as fn
+import numpy as np
+
+@contextlib.contextmanager
+def changedir(path):
+    try:
+        cwd = os.getcwd()
+        os.chdir(path)
+        yield
+    finally:
+        os.chdir(cwd)
+
+class HyperGCNCocitationCora(dgl.data.DGLDataset):
+    """Cora cocitation dataset from
+    `HyperGCN paper <https://github.com/malllabiisc/HyperGCN>`_.
+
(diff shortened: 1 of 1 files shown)


mufeili:
- Since you do not use sparse APIs, I prefer moving it to `examples/pytorch`.
- Why do you use a bipartite graph here?

examples/sparse/unignn.py
```diff
@@ -0,0 +1,276 @@
...
+    `HyperGCN paper <https://github.com/malllabiisc/HyperGCN>`_.
+
+    It contains a single bipartite graph, which is the star-expansion of
+    the hypergraph.  The node type ``node`` represents the nodes in the
+    original hypergraph, and the node type ``hyperedge`` represents the
+    hyperedges in the original hypergraph.  The edges between ``node``
+    and ``hyperedge`` describes whether a hyperedge contains a node
+    in the original hypergraph.
+    """
+    dataset_name = 'HyperGCNCocitationCora'
```
frozenbugs, line 44: please run black auto-reformat.

examples/sparse/unignn.py
```diff
@@ -0,0 +1,276 @@
...
+@contextlib.contextmanager
+def changedir(path):
+    try:
+        cwd = os.getcwd()
+        os.chdir(path)
+        yield
+    finally:
+        os.chdir(cwd)
+
+class HyperGCNCocitationCora(dgl.data.DGLDataset):
```
frozenbugs, line 33: Why not make it a standard data set in DGL?

BarclayII: This is to show you how a hypergraph dataset will look like.  If you guys are OK to make it a standard DGL dataset then I'll open another PR.

BarclayII:
> * Why do you use a bipartite graph here?

The reason is that we do not have a dedicated structure for hypergraph, and UniGNN is a general hypergraph model, defined in message passing form by aggregating node features onto hyperedges and then hyperedge features onto nodes.  Message passing on a bipartite graph fits this paradigm.

examples/sparse/unignn.py
mufeili: +1 for moving it to DGL core

jermainewang:
Suspended. The value of adding this example to DGL core is not clear. May need further discussion.

PR #4989, opened 2022-12-06
[Refactor] Replace third_party/nccl with PyTorch's NCCL backend

### LINKED ISSUE #4590: Replace third_party/nccl with PyTorch's NCCL backend
## 🚀 Feature
Starting in PyTorch 1.11, `send` and `recv` are supported by `torch.distributed`, which largely means we no longer need direct access to NCCL.

To do this, we should change python/dgl/cuda/nccl.py to use `torch.distributed`, and remove `src/runtime/cuda/nccl_api.h` and `src/runtime/cuda/nccl_api.cu`.

Before we remove it however, we should verify the performance of `sparse_all_to_all_push` and `spares_all_to_all_pull` do not suffer.

## Motivation

Removing our own version of NCCL will simplify code as we can use the existing communicator created by `torch.disrtibuted`, reduce build times, and decrease the likely-hood of conflicts of two versions of the libraries (and possibly multiple communicators).

yaox12:
## Description

Resolve #4590.

Removing our own version of NCCL will simplify code as we can use the existing communicator created by torch.disrtibuted, reduce build times, and decrease the likely-hood of conflicts of two versions of the libraries (and possibly multiple communicators).

### Performance

Training a model that contains a embedding lookup layer only, followed by a cross-entropy loss, with two A100-SXM4 GPUs connected by NVLink, the step time (forward + backward) under the old and new implementations are as follows. Total embedding size is (5 000 000, 128) and global batch size is 100 000.

<table>
<thead>
  <tr>
    <th rowspan="2">embedding device</th>
    <th rowspan="2">state_dict device</th>
    <th rowspan="2">state_dict dtype</th>
    <th colspan="2">step time (ms)</th>
  </tr>
  <tr>
    <th>old</th>
    <th>new</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td rowspan="4">CPU</td>
    <td rowspan="2">CPU</td>
    <td>FP32</td>
    <td>44.3</td>
    <td>40.7</td>
  </tr>
  <tr>
    <td>FP16</td>
    <td>37.6</td>
    <td>35.1</td>
  </tr>
  <tr>
    <td rowspan="2">CPU Pinned</td>
    <td>FP32</td>
    <td>28.9</td>
    <td>28.4</td>
  </tr>
  <tr>
    <td>FP16</td>
    <td>27.5</td>
    <td>27.9</td>
  </tr>
  <tr>
    <td rowspan="2">GPU</td>
    <td rowspan="2">GPU</td>
    <td>FP32</td>
    <td>8.8</td>
    <td>9.0</td>
  </tr>
  <tr>
    <td>FP16</td>
    <td>8.3</td>
    <td>8.1</td>
  </tr>
</tbody>
</table>

## Checklist
- [x] Code is well-documented


nv-dlasalle:
@classicsong since you observed good speedups when we initially integrated NCCL for the embedding updates, can you confirm that this PR does not lead to a regression for your case?

python/dgl/nn/pytorch/sparse_emb.py
```diff
@@ -187,15 +169,13 @@ def __call__(self, node_ids, device=th.device("cpu")):
...
+        if not self._comm:
             emb = self._tensor[node_ids].to(device)
         else:
-            if self.world_size > 0:
-                emb = self._comm.sparse_all_to_all_pull(
-                    node_ids, self._tensor, self._partition
-                )
-            else:
-                emb = self._tensor[node_ids]
+            # will handle self._world_size = 0 or 1
```
frozenbugs, line 175: This comment is not very clear to me.
What will handle self._world_size = 0 or 1? What will happen if self._world_size not equals 0 or 1?

Can you elaborate more?

python/dgl/nn/pytorch/sparse_emb.py
```diff
@@ -161,7 +141,9 @@ def __init__(
                 )
 
             # create local tensors for the weights
-            local_size = self._partition.local_size(self._comm.rank())
+            local_size = self._partition.local_size(
+                self._rank if self._rank > 0 else 0
```
frozenbugs, line 145: max(self._rank, 0)?

python/dgl/nn/pytorch/sparse_emb.py
```diff
@@ -361,7 +329,8 @@ def all_set_embedding(self, values):
         if self._partition:
             idxs = F.copy_to(
                 self._partition.get_local_indices(
-                    self._comm.rank(), ctx=F.context(self._tensor)
+                    self._rank if self._rank > 0 else 0,
```
frozenbugs, line 332: max(self._rank, 0)?

python/dgl/optim/pytorch/sparse_optim.py
```diff
@@ -74,7 +73,7 @@ def step(self):
         if self._first_step:
             for emb in self._params:
                 for _, data in emb._trace:
-                    if data.grad.data.device.type == "cuda":
+                    if data.grad.device.type == "cuda":
```
frozenbugs, line 76: Is this change intended?

python/dgl/partition.py
```diff
@@ -592,5 +592,21 @@ def map_to_global(self, idxs, part_id):
             )
         )
 
+    def generate_permutation(self, idxs):
+        """Create a mapping for the given indices to different partitions,
+        and a count of the number of indices per part.
```
frozenbugs, line 597: Produce a scheme that maps the given indices to separate partitions and the counts of how many indices are in each partition.

And parameter section and examples?

python/dgl/partition.py
```diff
@@ -592,5 +592,21 @@ def map_to_global(self, idxs, part_id):
             )
         )
 
+    def generate_permutation(self, idxs):
```
frozenbugs, line 595: Add unittest for this method?

python/dgl/partition.py
```diff
@@ -592,5 +592,21 @@ def map_to_global(self, idxs, part_id):
...
+        """Create a mapping for the given indices to different partitions,
+        and a count of the number of indices per part.
+
+        Return
+        ------
+        perm (Tensor): The permutation to re-order the indices by partition.
+        send_sum (Tensor): The number of indices per partition.
```
frozenbugs, line 602: what does send_sum mean?

frozenbugs, line 602: Add size of each tensor in the comment will be very helpful.

python/dgl/partition.py
```diff
@@ -592,5 +592,21 @@ def map_to_global(self, idxs, part_id):
             )
         )
 
+    def generate_permutation(self, idxs):
+        """Create a mapping for the given indices to different partitions,
+        and a count of the number of indices per part.
+
+        Return
+        ------
+        perm (Tensor): The permutation to re-order the indices by partition.
```
frozenbugs, line 601:
```
Return
------
torch.Tensor
    A tensor with shape (X, X), represents the permutation of the re-ordered indices according to the partitions.
torch.Tensor
    A tensor with shape (X, X), each item represents the number of indices belongs to each partition.
```

czkkkkkk (commented):
Overall I think the change is nice because it simplifies our NCCL call. One concern is since this PR remove the C++ NCCL interfaces, is it possible to call torch.distributed NCCL from the C++ side in the future?

python/dgl/cuda/nccl.py
```diff
@@ -1,230 +1,184 @@
...
+    """Perform an all-to-all-v operation, where by all processors send out
+    a set of indices and corresponding values. Indices and values,
+    corresponding to the current process, will copied into the output
+    arrays.
 
-class UniqueId(object):
-    """Class for allowing python code to create and communicate NCCL Unique
-    IDs, needed for creating communicators.
-    """
+    Note: This method requires 'torch.distributed.is_initialized()' and
```
czkkkkkk, line 13: What would happen if the backend is not `nccl`? Shall we raise an error here?

python/dgl/nn/pytorch/sparse_emb.py
yaox12: The following `nccl.sparse_all_to_all_pull` will handle  self._world_size = 0 or 1, where we can just return `self._tensor[node_ids]`.

python/dgl/nn/pytorch/sparse_emb.py
yaox12: Updated.

python/dgl/nn/pytorch/sparse_emb.py
yaox12: Updated.

python/dgl/optim/pytorch/sparse_optim.py
yaox12: Yes. `data.grad` is a tensor, and `data.grad.device` is enough to get its device. `tensor.data` is an outdated grammar in PyTorch.

If you look at the following lines 80-83, they all use `data.grad.device`.

python/dgl/partition.py
yaox12: Updated.

python/dgl/partition.py
yaox12: Added.
PS: `GeneratePermutation` has its C++ test.

python/dgl/partition.py
yaox12: Updated the comments here.

yaox12: Updated.

python/dgl/partition.py
yaox12: Updated.

python/dgl/cuda/nccl.py
```diff
@@ -1,230 +1,184 @@
...
-            The number of processes in this communicator.
-        """
-        return self._size
-
-
-def is_supported():
-    """Check if DGL was built with NCCL support.
+    Parameters
+    ----------
+    idx : tensor
```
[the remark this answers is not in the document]
yaox12: Updated.

python/dgl/cuda/nccl.py
```diff
@@ -1,230 +1,184 @@
...
+        The 1D set of indices to send to other processors.
+    value : tensor
```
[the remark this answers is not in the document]
yaox12: Updated.

python/dgl/cuda/nccl.py
yaox12: Added an assertation.

yaox12:
> Overall I think the change is nice because it simplifies our NCCL call. One concern is since this PR remove the C++ NCCL interfaces, is it possible to call torch.distributed NCCL from the C++ side in the future?

For dgl.sparse, yes. For vanilla DGL, we might have to import methods via TensorAdaptor if necessary.

czkkkkkk:
> > Overall I think the change is nice because it simplifies our NCCL call. One concern is since this PR remove the C++ NCCL interfaces, is it possible to call torch.distributed NCCL from the C++ side in the future?
> 
> For `dgl.sparse`, yes. For vanilla DGL, we might have to import some methods via TensorAdaptor if necessary.

Do you mean C++ DGL sparse library can call `torch.dist` from libtorch? Which libtorch APIs should be used?

yaox12:
> > > Overall I think the change is nice because it simplifies our NCCL call. One concern is since this PR remove the C++ NCCL interfaces, is it possible to call torch.distributed NCCL from the C++ side in the future?
> > 
> > 
> > For `dgl.sparse`, yes. For vanilla DGL, we might have to import some methods via TensorAdaptor if necessary.
> 
> Do you mean C++ DGL sparse library can call `torch.dist` from libtorch? Which libtorch APIs should be used?

Might be methods in [torch/csrc/distributed/c10d/ProcessGroupNCCL.hpp](https://github.com/pytorch/pytorch/blob/e70ea8d58d8da777c998d820ec3f3c2b5615c8c9/torch/csrc/distributed/c10d/ProcessGroupNCCL.hpp)

czkkkkkk:
> Might be methods in [torch/csrc/distributed/c10d/ProcessGroupNCCL.hpp](https://github.com/pytorch/pytorch/blob/e70ea8d58d8da777c998d820ec3f3c2b5615c8c9/torch/csrc/distributed/c10d/ProcessGroupNCCL.hpp)

I see. In libtorch, is it possible to use the same NCCL communicator as `torch.dist` on the Python side? Or we need to create a new `ProcessGroupNCCL`?

czkkkkkk (approved):

yaox12:
> > Might be methods in [torch/csrc/distributed/c10d/ProcessGroupNCCL.hpp](https://github.com/pytorch/pytorch/blob/e70ea8d58d8da777c998d820ec3f3c2b5615c8c9/torch/csrc/distributed/c10d/ProcessGroupNCCL.hpp)
> 
> I see. In libtorch, is it possible to use the same NCCL communicator as `torch.dist` on the Python side? Or we need to create a new `ProcessGroupNCCL`?

I guess we can pass the process group from Python side to C++ side, but I'm not 100% sure.

python/dgl/nn/pytorch/sparse_emb.py
```diff
@@ -170,9 +168,11 @@ def __call__(self, node_ids, device=th.device("cpu")):
             Target device to put the collected embeddings.
         """
         if not self._comm:
+            # embeddings are stored on the CPU
             emb = self._tensor[node_ids].to(device)
         else:
-            # will handle self._world_size = 0 or 1
+            # embeddings are stored on the GPU
+            # the following method will handle self._world_size = 0 or 1
```
frozenbugs, line 175: What's will happen if self._world_size != 0 or 1

python/dgl/cuda/nccl.py
```diff
@@ -1,230 +1,191 @@
...
+    perm = perm.long()
+
+    # Get receive splits.
+    recv_splits = torch.empty_like(send_splits)
+    dist.all_to_all_single(recv_splits, send_splits)
+
+    # Use pinned memory to speedup D2H copy.
+    recv_splits = recv_splits.to("cpu", non_blocking=True)
+    send_splits = send_splits.to("cpu", non_blocking=True)
+    # TODO(Xin): Overlap D2H copy and index selection.
```
frozenbugs, line 77: Please file a issue or link to existing issue in the following format.

# TODO(1234):

python/dgl/cuda/nccl.py
```diff
@@ -1,230 +1,191 @@
...
+    perm = perm.long()
+
+    # Get response splits.
+    resp_splits = torch.empty_like(req_splits)
+    dist.all_to_all_single(resp_splits, req_splits)
+
+    # Use pinned memory to speedup D2H copy.
+    resp_splits = resp_splits.to("cpu", non_blocking=True)
+    req_splits = req_splits.to("cpu", non_blocking=True)
+    # TODO(Xin): Overlap D2H copy and index selection.
```
frozenbugs, line 161: ditto

tests/python/common/test_partition.py
```diff
@@ -43,3 +43,11 @@ def test_get_node_partition_from_book(idtype):
     act_ids = partition.map_to_global(test_ids, 2)
     exp_ids = F.copy_to(F.tensor([6, 7, 10], dtype=idtype), F.ctx())
     assert F.array_equal(act_ids, exp_ids)
+
+    # test generate_permutation
```
frozenbugs, line 47: super nit: Test generate_permutation.

python/dgl/nn/pytorch/sparse_emb.py
yaox12: The `nccl.sparse_all_to_all_pull/push` methods are designed to perform communications across GPUs, but for world_size == 0 or 1, there is only one GPU so we don't have to do the **real** pull/push communication.

yaox12: I changed the comments to `the following method also covers self._world_size = 0 or 1`. Hope it would be clearer.

python/dgl/cuda/nccl.py
yaox12: Remove it since it's not a TODO now.

At first, I came up with two potential improvements, pinned memory and overlap. After playing with it for a while, using pinned memory (without overlapping) is fast enough. Trying to overlap the copy and index selection would introduce extra synchronization costs.

python/dgl/nn/pytorch/sparse_emb.py
frozenbugs: Ah, I see. You should just mention the explanation in the comment for both pull and push method.
e.g.
```
The following method is designed to facilitate communication across multiple GPUs and can handle situations where only one GPU is present or no GPUs (a.k.a. self._world_size = 0 or 1) are available gracefully.
```

frozenbugs (approved):

yaox12:
@dgl-bot

PR #5003, opened 2022-12-08
[Sparse] Misc fix for existing elementwise ops

mufeili:
## Description

- Fix the indent of the doc.
- Replace `DiagMatrix` with `diag` in doc examples.
- Replace `SparseMatrix` with `create_from_coo` in doc examples.
- Properly handle other types of objects in `diag_add`, `diag_sub`, `diag_mul`.
- Properly handle diagonal matrix shape in `diag_add`, `diag_sub`, `diag_mul`, `diag_div`.
- Forbid adding a diagonal matrix to sparse matrix due to the lack of support for adding sparse matrices of different sparsity.
- Forbid adding two sparse matrices of different sparsity due to the same reason above.
- Extend unit tests to properly handle diagonal matrix shape and device.
- I will continue the rest effort with additional PRs.

One question, should we import all operators in `elementwise_op.py`? For example, currently it only imports `diag_add` from `elementwise_op_diag`. If we want to expose the operators, we should also import `diag_sub`, `diag_mul`, etc.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/mock_sparse2/elementwise_op_sp.py
+++ b/python/dgl/mock_sparse2/elementwise_op_sp.py
@@ -8,14 +8,13 @@ __all__ = ["sp_add"]
 
 
 def spsp_add(A, B):
-    """ Invoke C++ sparse library for addition """
+    """Invoke C++ sparse library for addition """
     return SparseMatrix(
         torch.ops.dgl_sparse.spsp_add(A.c_sparse_matrix, B.c_sparse_matrix)
     )
 
 
-def sp_add(
-    A: SparseMatrix, B: SparseMatrix) -> SparseMatrix:
+def sp_add(A: SparseMatrix, B: SparseMatrix) -> SparseMatrix:
     """Elementwise addition.
 
     Parameters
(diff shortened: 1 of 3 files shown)


mufeili:
`lintrunner` is fine with `a * b`, but not `a ** b` and requires `a**b`. This is weird to me. @frozenbugs

python/dgl/mock_sparse2/elementwise_op_sp.py
```diff
@@ -1,31 +1,27 @@
...
+    """Invoke C++ sparse library for addition"""
     return SparseMatrix(
         torch.ops.dgl_sparse.spsp_add(A.c_sparse_matrix, B.c_sparse_matrix)
     )
 
 
-def sp_add(
-    A: Union[SparseMatrix, DiagMatrix], B: Union[SparseMatrix, DiagMatrix]
-) -> SparseMatrix:
+def sp_add(A: SparseMatrix, B: SparseMatrix) -> SparseMatrix:
```
czkkkkkk, line 16: Do we support addition between sparse matrices with different sparsities for the 12/30 milestone?

czkkkkkk (approved):

PR #5005, opened 2022-12-08
[Bugfix] Fix empty tensors may being treated as pinned

### LINKED ISSUE #5004: [Bug] `edge_subgraph` would fail with UVA when there're empty relations
## 🐛 Bug

`edge_subgraph`/`khop_in_subgraph`/`khop_out_subgraph` would fail with UVA when there're empty relations in the graph.

## To Reproduce

```python
import dgl
import torch

def create_test_heterograph2(idtype):
    """test heterograph from the docstring, with an empty relation"""
    # 3 users, 2 games, 2 developers
    # metagraph:
    #    ('user', 'follows', 'user'),
    #    ('user', 'plays', 'game'),
    #    ('user', 'wishes', 'game'),
    #    ('developer', 'develops', 'game')

    g = dgl.heterograph(
        {
            ("user", "follows", "user"): ([0, 1], [1, 2]),
            ("user", "plays", "game"): ([0, 1, 2, 1], [0, 0, 1, 1]),
            ("user", "wishes", "game"): ([0, 2], [1, 0]),
            ("developer", "develops", "game"): ([], []),
        },
        idtype=idtype,
    )
    return g

def test_uva_subgraph(idtype):
    device = torch.device("cuda:0")
    g = create_test_heterograph2(idtype)
    g.create_formats_()
    g.pin_memory_()
    indices = {"user": torch.tensor([0], dtype=idtype, device=device)}
    edge_indices = {"follows": torch.tensor([0], dtype=idtype, device=device)}

    # all the following three tests would fail
    assert g.edge_subgraph(edge_indices).device == device
    assert g.khop_in_subgraph(indices, 1)[0].device == device
    assert g.khop_out_subgraph(indices, 1)[0].device == device
    g.unpin_memory_()

test_uva_subgraph(torch.int64)
```

## Expected behavior

They should work.

## Environment

 - DGL Version (e.g., 1.0): master branch
 - OS (e.g., Linux):
 - Python version:

yaox12:
## Description

Fix #5004.

The root causes are two aspects:
1. Empty tensors might be treated as pinned though they're not.
2. When an empty CSR matrix is pinned, it has the `is_pinned=true` flag, but the underlying `indices` could be not pinned (empty). We should avoid calling `cudaHostGetDevicePointer` on it.

This PR fixes the above two issues and enhances the unit tests for them.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/include/dgl/aten/macro.h
+++ b/include/dgl/aten/macro.h
@@ -311,7 +311,7 @@
 #define CHECK_VALID_CONTEXT(VAR1, VAR2)                          \
   CHECK(                                                         \
       ((VAR1)->ctx == (VAR2)->ctx) || (VAR1).IsPinned() ||       \
-      ((VAR1).NumElements() == 0))                               \
+      ((VAR1).NumElements() == 0)) /* Let empty arrays pass */   \
       << "Expected " << (#VAR2) << "(" << (VAR2)->ctx << ")"     \
       << " to have the same device "                             \
       << "context as " << (#VAR1) << "(" << (VAR1)->ctx << "). " \
--- a/include/dgl/aten/macro.h
+++ b/include/dgl/aten/macro.h
@@ -309,7 +309,9 @@
   });
 
 #define CHECK_VALID_CONTEXT(VAR1, VAR2)                          \
-  CHECK(((VAR1)->ctx == (VAR2)->ctx) || (VAR1).IsPinned())       \
+  CHECK(                                                         \
+      ((VAR1)->ctx == (VAR2)->ctx) || (VAR1).IsPinned() ||       \
+      ((VAR1).NumElements() == 0))                               \
       << "Expected " << (#VAR2) << "(" << (VAR2)->ctx << ")"     \
       << " to have the same device "                             \
       << "context as " << (#VAR1) << "(" << (VAR1)->ctx << "). " \
--- a/src/array/cuda/array_index_select.cu
+++ b/src/array/cuda/array_index_select.cu
@@ -16,12 +16,6 @@ namespace impl {
 
 template <DGLDeviceType XPU, typename DType, typename IdType>
 NDArray IndexSelect(NDArray array, IdArray index) {
-  cudaStream_t stream = runtime::getCurrentCUDAStream();
-  const DType* array_data = array.Ptr<DType>();
-  if (array.IsPinned()) {
-    CUDA_CALL(cudaHostGetDevicePointer(&array_data, array.Ptr<DType>(), 0));
-  }
-  const IdType* idx_data = static_cast<IdType*>(index->data);
   const int64_t arr_len = array->shape[0];
   const int64_t len = index->shape[0];
   int64_t num_feat = 1;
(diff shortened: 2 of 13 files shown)


nv-dlasalle (approved):

src/array/cuda/array_index_select.cu
```diff
@@ -33,9 +27,13 @@ NDArray IndexSelect(NDArray array, IdArray index) {
 
   // use index->ctx for pinned array
   NDArray ret = NDArray::Empty(shape, array->dtype, index->ctx);
-  if (len == 0) return ret;
+  if (len == 0 || arr_len == 0) return ret;
```
chang-l, line 30: Just wondering if this checking condition is complete or not... It seems only checking the length of the first dim, right? Could it cover the case when `array->shape[i] == 0`?

src/array/cuda/csr_get_data.cu
```diff
@@ -43,19 +43,13 @@ NDArray CSRGetData(
...
-          cudaHostGetDevicePointer(&data_data, csr.data.Ptr<IdType>(), 0));
-    }
-  }
+  const IdType* indptr_data =
+      static_cast<IdType*>(cuda::GetDevicePointer(csr.indptr));
+  const IdType* indices_data =
+      static_cast<IdType*>(cuda::GetDevicePointer(csr.indices));
+  const IdType* data_data =
+      CSRHasData(csr) ? static_cast<IdType*>(cuda::GetDevicePointer(csr.data))
+                      : nullptr;
```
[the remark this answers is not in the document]
yaox12: We will check if `csr.indptr/indices/data` is pinned separately in this `GetDevicePointer()` function. A CSR matrix could hold the `is_pinned=true` flag even if it's empty, so we should avoid calling `cudaHostGetDevicePointer` on it's `indptr/indices/data` in such a case.
https://github.com/dmlc/dgl/blob/e28f07812bae274f5ba49d26629de5541c9cff60/include/dgl/aten/csr.h#L137-L145

src/array/cuda/array_index_select.cu
yaox12: I cannot imagine a tensor has other dims when the first dim is 0.

yaox12: Here is the utility to check a null array.
https://github.com/dmlc/dgl/blob/e28f07812bae274f5ba49d26629de5541c9cff60/include/dgl/aten/array_ops.h#L37

include/dgl/aten/macro.h
```diff
@@ -309,7 +309,9 @@
   });
 
 #define CHECK_VALID_CONTEXT(VAR1, VAR2)                          \
-  CHECK(((VAR1)->ctx == (VAR2)->ctx) || (VAR1).IsPinned())       \
+  CHECK(                                                         \
+      ((VAR1)->ctx == (VAR2)->ctx) || (VAR1).IsPinned() ||       \
+      ((VAR1).NumElements() == 0))                               \
```
chang-l, line 314: Do you think if it would be possible or better to check 0-element and return early [here](https://github.com/dmlc/dgl/blob/e28f07812bae274f5ba49d26629de5541c9cff60/src/array/array.cc#L120)?

chang-l (approved):

src/array/cuda/array_index_select.cu
chang-l: I see...

yaox12: I get your point. It would be better to use `array.NumElements()` here.

yaox12: Or simply `arr_len * num_feat` here.

include/dgl/aten/macro.h
yaox12: I agree. But I find we need to return early for many other functions, such as `dgl::aten::CSRSliceRows()`...

I will add a comment to this macro.

PR #5017, opened 2022-12-12
[Performance] Fix for number of threads in COOToCSR

anko-intel:
## Description
The wrong number of threads (always 1) was taken into account when selecting the COOdoCSR algorithm.
After the change, there is a significant performance improvement.
Microbenchmark (created as modification of [tests/cpp/test_spmat_coo.cc](https://gist.github.com/anko-intel/7e86c74342e3d45233c4cb849ace76eb)) gives following results on r6i.16xlarge (Intel(R) Xeon(R) Platinum 8375C) run with:
`numactl --physcpubind=0-31 ./build/runUnitTests --gtest_filter=*Big*`

index type | N | NNZ | 100%-sparsity | before [us] | after[us] | before/after
-- | -- | -- | -- | -- | -- | --
int32 | 5 111 000 | 2 612 233 | 0.00001% | 38 654 | 3 436 | 11.2
int64 | 5 111 000 | 2 612 233 | 0.00001% | 133 154 | 9 529 | 14.0
int32 | 12 000 333 | 14 400 800 | 0.00001% | 204 931 | 49 575 | 4.1
int64 | 12 000 333 | 14 400 800 | 0.00001% | 321 307 | 80 980 | 4.0
int32 | 32 768 | 9 664 | 0.0009% | 254 | 44 | 5.8
int64 | 32 768 | 9 664 | 0.0009% | 248 | 32 | 7.8
int32 | 128 603 | 148 849 | 0.0009% | 875 | 126 | 6.9
int64 | 128 603 | 148 849 | 0.0009% | 915 | 142 | 6.4
int32 | 1 024 | 630 | 0.06% | 31 | 28 | 1.1
int64 | 1 024 | 630 | 0.06% | 28 | 26 | 1.1
int32 | 1 024 | 5 243 | 0.5% | 42 | 39 | 1.1
int64 | 1 024 | 5 243 | 0.5% | 38 | 34 | 1.1

## Checklist

### DIFF
--- a/src/array/cpu/spmat_op_impl_coo.cc
+++ b/src/array/cpu/spmat_op_impl_coo.cc
@@ -639,7 +639,7 @@ N*P).
 template <DGLDeviceType XPU, typename IdType>
 CSRMatrix COOToCSR(COOMatrix coo) {
   if (!coo.row_sorted) {
-    const int64_t num_threads = omp_get_num_threads();
+    const int64_t num_threads = omp_get_max_threads();
     const int64_t num_nodes = coo.num_rows;
     const int64_t num_edges = coo.row->shape[0];
     // Besides graph density, num_threads is also taken into account. Below


yaox12:
@dgl-bot

agrabows (approved):

bgawrych (approved):

anko-intel:
@BarclayII Could you review/merge ?

src/array/cpu/spmat_op_impl_coo.cc
```diff
@@ -639,7 +639,7 @@ N*P).
 template <DGLDeviceType XPU, typename IdType>
 CSRMatrix COOToCSR(COOMatrix coo) {
   if (!coo.row_sorted) {
-    const int64_t num_threads = omp_get_num_threads();
+    const int64_t num_threads = omp_get_max_threads();
```
BarclayII, line 642: I'm more curious why/when `omp_get_num_threads` would return 1.

anko-intel: It always return 1 when it is not run in parallel region. 
Acording to the 5.0 documentation:
"Effect: The omp_get_num_threads routine returns the number of threads in the team that is executing the parallel region to which the routine region binds. If called from the sequential part of a program, this routine returns 1."

For _omp_get_max_threads_:
"Effect: The value returned by omp_get_max_threads is the value of the first element of the nthreads-var ICV of the current task. **This value is also an upper bound on the number of threads that could be used to form a new team if a parallel region without a num_threads clause were encountered after execution returns from this routine**."

BarclayII:
@dgl-bot

yaox12:
Perhaps we should update fix the `omp_get_num_threads` in corresponding tests.
https://github.com/dmlc/dgl/blob/025e43210f724ea3a5f308463166e61b4cd97399/tests/cpp/test_spmat_coo.cc#L216
and 
https://github.com/dmlc/dgl/blob/025e43210f724ea3a5f308463166e61b4cd97399/tests/cpp/test_spmat_coo.cc#L293

anko-intel:
> Perhaps we should fix the `omp_get_num_threads` in corresponding tests.
> 
> https://github.com/dmlc/dgl/blob/025e43210f724ea3a5f308463166e61b4cd97399/tests/cpp/test_spmat_coo.cc#L216
> 
> 
> and
> https://github.com/dmlc/dgl/blob/025e43210f724ea3a5f308463166e61b4cd97399/tests/cpp/test_spmat_coo.cc#L293

I've been thinking about this. 
My guess is that these checks were inserted to ensure that both versions of Unsorted*COOToCSR are tested. But it doesn't work. Fixing the number of threads for the given example could result in a situation where the number of elements/edges is greater than possible (or many edges between the same nodes). For a machine with 64 threads available and the desired value of False from the IsSparse() function, we need to insert at least 64 elements (edges) into a 4x5 matrix!
So the test should be performed on a much larger tensor, which does not seem to be a reasonable and efficient solution. From the point of view of testing, it would be best to have a method to select the COOToCSR version from the API (maybe as an additional "hint" parameter), but from the point of view of the user it seems to be unneceserry complicated. Maybe setting the number of threads from the test would be a workaround.
I'm going to do some measurements and experiments for the COOToCSR function and maybe some other conclusions/improvements will emerge. Let me  wait with the fixes in test until then.

BarclayII (approved):

BarclayII:
@anko-intel Should I wait for your test changes or can I go ahead and merge it?

anko-intel:
> @anko-intel Should I wait for your test changes or can I go ahead and merge it?

Please merge it.

PR #5021, opened 2022-12-13
[Quality of life] Fix warnings raised by the core library

yaox12:
## Description

Try to fix warnings raised by the core library. For warnings root caused by deprecated usages in unit tests, I just leave them there.

Main changes are:
1. `distutils.version.LooseVersion` -> `setuptools.extern.packaging.version.parse` because the former one is deprecated. We use the `packaging` vendored in `setuptools` in case users haven't installed it.
2. `np.bool/int` -> `np.bool_/int64`. Besides, there're many warnings on the use of `np.float` (should be either `np.float32` or `np.float64`) in unit tests, but they are not raised in the core lib so I don't fix them.
3. Some `torch.tensor` -> `torch.as_tensor` to convert tensors with different `dtype` and `device`. 

Note: I will format the changed files after this PR is approved. Otherwise, there will be too many irrelevant changes.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/apps/kg/train_pytorch.py
+++ b/apps/kg/train_pytorch.py
@@ -24,10 +24,9 @@ from torch.utils.data import DataLoader
 import torch.optim as optim
 import torch as th
 
-from distutils.version import LooseVersion
-TH_VERSION = LooseVersion(th.__version__)
-if TH_VERSION.version[0] == 1 and TH_VERSION.version[1] < 2:
-    raise Exception("DGL-ke has to work with Pytorch version >= 1.2")
+from packaging import version
+if version.parse(th.__version__) < version.parse("1.2"):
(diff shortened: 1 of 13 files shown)


frozenbugs:
I'd suggest:
1. formatting the changed files **before** this PR is send.
2. Create 3 PRs for the 3 goal instead of merge them all together for faster review.

yaox12:
> I'd suggest:
> 
> 1. formatting the changed files **before** this PR is send or **after** the PR merged, (never mix reformatting with real change in a single PR merge, when something stops working, we may look back to check PR history).
> 2. Create 3 PRs for the 3 goal instead of merge them all together for faster review.

Got it. I'll split the PR into three.

PR #5026, opened 2022-12-14
[Fix core lib warning] Remove LooseVersion

yaox12:
## Description

Replace `distutils.version.LooseVersion` with `packaging.version.parse` because the former is deprecated. Users will see the warning every time they import DGL.

If `packaging` is not installed, we use the vendored copy in `setuptools`.

## Checklist
- [ ] Code is well-documented


yaox12:
@dgl-bot

frozenbugs (approved):

frozenbugs:
@dgl-bot

frozenbugs:
Do you mean user don't need to uninstall setuptools separately or they need to?

yaox12:
> Do you mean user don't need to uninstall setuptools separately or they need to?

`packaging` is a standalone package can be installed via `pip install packaging`, and it's also vendored in `setuptools`. We can use the verdored copy when `packaing` is not installed, because users always have `setuptools`. A better way may be:
```python
try:
    from packaging import version
except ImportError:
    # If packaging isn't installed, try and use the vendored copy
    # in setuptools
    from setuptools.extern.packaging import version
```

python/dgl/utils/__init__.py
```diff
@@ -6,3 +6,10 @@
 from .internal import *
 from .pin_memory import *
 from .shared_mem import *
+
+try:
+    from packaging import version
+except ImportError:
+    # If packaging isn't installed, try and use the vendored copy
```
frozenbugs, line 13: Make full use of 80 chars before change to new line. (I think it can fit in one line with just 80 chars).
End a sentence with dot.

frozenbugs (approved):

yaox12:
@dgl-bot

PR #5034, opened 2022-12-14
[API Deprecation] Deprecate contrib module

peizhou001:
## Description
Remove entire dgl.contrib module and related tests.

## Checklist
- [ ] Code is well-documented


examples/pytorch/graphsage/advanced/train_sampling_unsupervised.py
```diff
@@ -73,7 +73,7 @@ def run(proc_id, n_gpus, args, devices, data):
     if args.data_device == 'gpu':
         nfeat = nfeat.to(device)
     elif args.data_device == 'uva':
-        nfeat = dgl.contrib.UnifiedTensor(nfeat, device=device)
+        nfeat = dgl.utils.pin_memory_inplace(nfeat)
```
yaox12, line 76: `dgl.utils.pin_memory_inplace()` returns a `ndarray` handle that holds the pinning flag. So I'd suggest either the following:
1. 
```python
nfeat_nd = dgl.utils.pin_memory_inplace(nfeat)
```
and just use `nfeat` in the following code.
2. Use PyTorch `.pin_memory()`
```python
nfeat = nfeat.pin_memory()
```
3. Use PyTorch's in-place pinning.
```python
torch.cuda.cudart().cudaHostRegister(nfeat.data_ptr(), nfeat.numel() * nfeat.element_size(), 0)
```
Generally, I prefer 2 due to the simplicity, though it requires more CPU PEAK memory during copying `nfeat` to the pinned buffer.

yaox12:
I'm OK with other changes as long as they pass the unit/example tests.

examples/pytorch/graphsage/advanced/train_sampling_unsupervised.py
peizhou001: Thanks for explaining. And I agree with you the 2end one is preferable because simplicity and clearness is at high priority in example. Let me fix it using method 2.

tests/pytorch/test_unified_tensor.py
```diff
@@ -1,85 +0,0 @@
-import dgl.multiprocessing as mp
```
jermainewang, line 1: I understand `contrib.UnifiedTensor` has been merged to the underlying of DataLoader. Are these test cases covered by other unit tests? If not, we should migrate them first.

jermainewang (approved):
Approved. @BarclayII Could you take a look at the unified tensor and dataloader part too?

tests/pytorch/test_unified_tensor.py
yaox12: I believe they have been covered by the `uva_cpu_indices` and `uva_cuda_indices` options of the dataloader tests.
https://github.com/dmlc/dgl/blob/00f2099934ca78a3c156c4222edcc3245cee4e1c/tests/pytorch/test_dataloader.py#L172

yaox12 (approved):

PR #5047, opened 2022-12-17
[CI] upgrade CI pytorch version

BarclayII:
## Description
Upgrading CI's PyTorch version to 1.12.


docker/Dockerfile.ci_gpu
```diff
@@ -1,9 +1,11 @@
 # CI docker GPU env
-FROM nvidia/cuda:10.1-cudnn7-devel-ubuntu18.04
+FROM nvidia/cuda:10.2-cudnn7-devel-ubuntu18.04
 
 ENV TZ=US
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
+RUN apt-key del 3bf863cc
```
Rhett-Ying, line 7: this looks a bit weird. could you add a comment to clarify it? It aims to fix nv image issue?

docker/Dockerfile.ci_gpu_220816
```diff
@@ -0,0 +1,41 @@
+# CI docker GPU env
```
Rhett-Ying, line 1: this is used for tensorflow and mxnet? could we keep only one docker for gpu tests?

BarclayII: Somehow Tensorflow crashes with CUDA 10.2 and I wasn't able to make it work.

docker/Dockerfile.ci_gpu
BarclayII: Added a comment.

docker/Dockerfile.ci_gpu_220816
Rhett-Ying: all right. let's keep it

Rhett-Ying (approved):
LGTM

PR #5032, opened 2022-12-14
[Sparse] Support SpMM and SDDMM

czkkkkkk:
## Description
Resolve #4855.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [ ] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [ ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

PR #5067, opened 2022-12-22
[Sparse] Softmax based on Reduction and BroadcastOp

### LINKED ISSUE #4866: [Sparse] Support softmax.
## 🔨Work Item

**IMPORTANT:**

mufeili:
## Description

Closes #4866 

- This PR implements softmax with autograd support based on GSPMM and GSDDMM following the previous implementation [here](https://github.com/dmlc/dgl/blob/master/python/dgl/backend/pytorch/sparse.py#L669).
- The implementation supports both 1D and 2D nonzero values.
- The PR adds one new method for `SpMMNoAutoGrad` and another for `SDDMMNoAutoGrad`. The new methods handle different subsets of GSpMM and GSDMM compared with the previous methods. For both ops, the union of the previous method and the new method does not cover all cases. So I decided to keep them as separate methods for the time being.
- The old edge softmax implementation invokes dedicated fused ops for CPU computation. If this is desired, I can do that after we merge this PR.

## Checklist
- [x] Code is well-documented


dgl_sparse/src/softmax.cc
```diff
@@ -0,0 +1,97 @@
...
+      torch::Tensor sparse_val);
+
+  static tensor_list backward(AutogradContext* ctx, tensor_list grad_outputs);
+};
+
+void _SoftmaxSanityCheck(
+    c10::intrusive_ptr<SparseMatrix> sparse_mat, torch::Tensor sparse_val) {
+  auto val_shape = sparse_val.sizes();
+  CHECK_EQ(val_shape[0], sparse_mat->nnz())
+      << "Softmax: the value shape does not match nnz of the sparse matrix.";
```
jermainewang, line 31: Not needed. Sparse matrix constructor has checked this.

dgl_sparse/src/softmax.cc
```diff
@@ -0,0 +1,97 @@
...
+  CHECK(sparse_val.device() == sparse_mat->device())
+      << "Softmax: sparse matrix and non-zero values should be on the same "
+         "device.";
```
jermainewang, line 34: Same. This check is not needed.

dgl_sparse/src/softmax.cc
```diff
@@ -0,0 +1,97 @@
...
+}
+
+torch::Tensor SoftmaxAutoGrad::forward(
+    AutogradContext* ctx, c10::intrusive_ptr<SparseMatrix> sparse_mat,
+    torch::Tensor sparse_val) {
+  auto sparse_val_max = SpMMNoAutoGrad(sparse_mat, sparse_val, "max");
```
jermainewang, line 40: Instead of using SpMM, I feel the new sparse reduction operators are more suitable. https://github.com/dmlc/dgl/blob/master/dgl_sparse/include/sparse/reduction.h

dgl_sparse/src/softmax.cc
```diff
@@ -0,0 +1,97 @@
...
+  auto sparse_val_exp =
+      SDDMMNoAutoGrad(sparse_mat, sparse_val, sparse_val_max, "sub").exp();
```
jermainewang, line 42: Is this operation equal to a broadcast-sub between a matrix with a vector? Maybe instead of calling it `SDDMM`, we just call it `BroadcastSub`, `BroadcastDiv`, etc.

jermainewang: Also, the grad operator of `BroadcastSub`/`BroadcastAdd` should be sparse reduction too, so supporting high-order grad shouldn't be too difficult too.

jermainewang (changes requested):

dgl_sparse/src/softmax.cc
```diff
@@ -0,0 +1,97 @@
...
+        sds - SDDMMNoAutoGrad(sparse_mat, sparse_score, accum, "mul");
+  }
+
+  return {torch::Tensor(), sparse_val_grad};
+}
+
+c10::intrusive_ptr<SparseMatrix> Softmax(
+    const c10::intrusive_ptr<SparseMatrix>& sparse_mat) {
+  auto sparse_val = sparse_mat->value();
+  _SoftmaxSanityCheck(sparse_mat, sparse_val);
```
frozenbugs, line 83: Just pass the sparse_mat and get the val in the function is a better practise.
(Since both check in _SoftmaxSanityCheck is no necessary, this could be removed).

dgl_sparse/src/softmax.cc
frozenbugs: Just FYI, this is introduced in https://github.com/dmlc/dgl/pull/5032. If rename, also think about rename SPMMNoAutoGrad.

dgl_sparse/src/matmul.cc
```diff
@@ -98,5 +139,34 @@ torch::Tensor SDDMMNoAutoGrad(
   return ret;
 }
 
+torch::Tensor SDDMMNoAutoGrad(
```
frozenbugs, line 142: It seems that by extending the function with `const std::string& op` might have changed the meaning of the function and made the name not a good fit?

(please update the PR with since #5032 has merged)

dgl_sparse/src/softmax.cc
mufeili: replaced

dgl_sparse/src/softmax.cc
mufeili: Renamed, autograd support can be left to future work

dgl_sparse/src/matmul.cc
```diff
@@ -98,6 +98,53 @@ torch::Tensor SDDMMNoAutoGrad(
...
+  const int64_t out_row = sparse_mat->nnz();
+  const std::vector<int64_t> shape({out_row, e.size(1)});
+  auto ret = torch::zeros(shape, e.options());
+
+  auto dgl_e = TorchTensorToDGLArray(e);
+  auto dgl_v = TorchTensorToDGLArray(v);
+  auto dgl_ret = TorchTensorToDGLArray(ret);
+
+  // The format for calculation will be chosen in the following order: CSR,
+  // COO. CSR is created if the sparse matrix only has CSC format.
```
jermainewang, line 113: The format selection for SDDMM should prioritize COO. The change should be easy: check if COO exists first.

dgl_sparse/src/matmul.h
```diff
@@ -53,6 +55,75 @@ torch::Tensor SDDMMNoAutoGrad(
     const c10::intrusive_ptr<SparseMatrix>& sparse_mat, torch::Tensor mat1,
     torch::Tensor mat2_tr);
 
+/**
+ * @brief Broadcast the node feature to the edges and then compute
```
jermainewang, line 59: Let's avoid use the word "node" or "edge" in the context of sparse matrix operators.

dgl_sparse/src/matmul.cc
```diff
@@ -98,6 +98,53 @@ torch::Tensor SDDMMNoAutoGrad(
...
+    aten::COOSDDMM(
+        op.c_str(), coo, dgl_e, dgl_v, dgl_ret, 1 /* Lhs target: e */,
+        0 /* rhs target: u due to transpose */);
+  }
+  return ret;
+}
+
+torch::Tensor BroadcastSubNoAutoGrad(
+    const c10::intrusive_ptr<SparseMatrix>& sparse_mat, torch::Tensor e,
+    torch::Tensor v) {
```
jermainewang, line 132: I think a better signature just needs a sparse matrix and a dense tensor. The non-zero values are naturally carried in the sparse matrix. The return will also be a sparse matrix.

jermainewang (changes requested):
Mostly good! Just some last suggestions. Will approve after they are fixed.

dgl_sparse/src/matmul.cc
mufeili: updated

dgl_sparse/src/matmul.h
mufeili: updated

dgl_sparse/src/matmul.cc
mufeili: updated

jermainewang (approved):

Issue #5068, opened 2022-12-23
When I set shuffle=False,num_workers>0 in NodeDataLoader,  I can not obtain ordered samples


zzzzzero:
When I set shuffle=False,num_workers=0 in NodeDataLoader,  I can obtain ordered samples.However,when I set shuffle=False,num_workers>0 in NodeDataLoader,   I can not  obtain ordered samples.
Here is the example:
```
import dgl
import numpy as np
sampler = dgl.dataloading.MultiLayerFullNeighborSampler(2)
import scipy.sparse as sp
spmat = sp.rand(10000, 10000, density=0.05) #
g=dgl.from_scipy(spmat)  
dataloader = dgl.dataloading.NodeDataLoader(g, np.arange(10000), sampler,batch_size=1024, shuffle=False, drop_last=False, num_workers=0)
for input_nodes, output_nodes, blocks in dataloader:
  dst_idx=blocks[-1].dstdata[dgl.NID].numpy()
  print(dst_idx[:5])
```
[0 1 2 3 4]
[1024 1025 1026 1027 1028]
[2048 2049 2050 2051 2052]
[3072 3073 3074 3075 3076]
[4096 4097 4098 4099 4100]
[5120 5121 5122 5123 5124]
[6144 6145 6146 6147 6148]
[7168 7169 7170 7171 7172]
[8192 8193 8194 8195 8196]
[9216 9217 9218 9219 9220]

```
import dgl
import numpy as np
sampler = dgl.dataloading.MultiLayerFullNeighborSampler(2)
import scipy.sparse as sp
spmat = sp.rand(10000, 10000, density=0.05) #
g=dgl.from_scipy(spmat)  
dataloader = dgl.dataloading.NodeDataLoader(g, np.arange(10000), sampler,batch_size=1024, shuffle=False, drop_last=False, num_workers=4)
for input_nodes, output_nodes, blocks in dataloader:
  dst_idx=blocks[-1].dstdata[dgl.NID].numpy()
  print(dst_idx[:5])
```
[0 1 2 3 4]
[3072 3073 3074 3075 3076]
[6144 6145 6146 6147 6148]
[8192 8193 8194 8195 8196]
[1024 1025 1026 1027 1028]
[4096 4097 4098 4099 4100]
[7168 7169 7170 7171 7172]
[9216 9217 9218 9219 9220]
[2048 2049 2050 2051 2052]
[5120 5121 5122 5123 5124]

[...]

PR #5077, opened 2022-12-27
[Bugfix] Fix worker division in the dataloader

yaox12:
## Description

Resolves https://github.com/dmlc/dgl/issues/5068.

`shuffle=False` and `num_workers>0` should give ordered results now.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/dataloading/dataloader.py
+++ b/python/dgl/dataloading/dataloader.py
@@ -119,14 +119,13 @@ def _divide_by_worker(dataset, batch_size, drop_last):
     num_samples = dataset.shape[0]
     worker_info = torch.utils.data.get_worker_info()
     if worker_info:
-        num_batches = (num_samples + (0 if drop_last else batch_size - 1)) // batch_size
-        num_batches_per_worker = num_batches // worker_info.num_workers
-        left_over = num_batches % worker_info.num_workers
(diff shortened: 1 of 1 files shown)


python/dgl/dataloading/dataloader.py
```diff
@@ -119,14 +119,13 @@ def _divide_by_worker(dataset, batch_size, drop_last):
...
-        start *= batch_size
-        end = min(end * batch_size, num_samples)
-        dataset = dataset[start:end]
+        num_batches = (
+            num_samples + (0 if drop_last else batch_size - 1)
+        ) // batch_size
+        segments = []
+        for i in range(worker_info.id, num_batches, worker_info.num_workers):
+            segments.append(dataset[i * batch_size : (i + 1) * batch_size])
+        dataset = torch.cat(segments) if segments else dataset[0:0]
```
BarclayII, line 128: If the dataset is large, would we have a lot of elements in `segments`?  And `torch.cat` will create a tensor with similar size as the original dataset (namely one over the number of workers the size of the original dataset)?

I think this implementation will introduce quite some overhead.

yaox12: You're right.

PR #5098, opened 2023-01-03
[Dist] Allow reading and writing single-column vector Parquet files.

thvasilo:
## Description
[Spark ML feature processing](https://spark.apache.org/docs/latest/ml-features) will commonly produce DataFrames, and by extension Parquet files, that contain columns whose elements are vectors instead of single values.

Example:

```python
>>> table.to_pandas()
      vector
0  [1, 2, 3]
1  [4, 5, 6]
>>> table.shape
(2, 1)
```

This PR adds the capability to the Parquet reader to read such files by transforming them to regular Numpy matrices, converting each vector to one row of the matrix.

@classicsong @Rhett-Ying  requested for review

## Checklist
- [x] Code is well-documented

### DIFF
--- a/tests/tools/utils.py
+++ b/tests/tools/utils.py
@@ -6,6 +6,7 @@ import torch
 
 import dgl
 from distpartitioning import array_readwriter
+from distpartitioning.array_readwriter.parquet import ParquetArrayParser
 from files import setdir
 
 
@@ -18,11 +19,12 @@ def _chunk_numpy_array(arr, fmt_meta, chunk_sizes, path_fmt, vector_rows=False):
         arr_chunk = arr[offset: offset + n]
         shape = arr_chunk.shape
         logging.info("Chunking %d-%d" % (offset, offset + n))
-        # If requested we write multi-column arrays as single-column vector files
-        if vector_rows and len(shape) > 1 and shape[1] > 1:
-            array_readwriter.get_array_parser(**fmt_meta).write(path, arr_chunk, vector_rows=vector_rows)
+        # If requested we write multi-column arrays as single-column vector Parquet files
+        array_parser = array_readwriter.get_array_parser(**fmt_meta)
+        if isinstance(array_parser, ParquetArrayParser) and len(shape) > 1 and shape[1] > 1:
+            array_parser.write(path, arr_chunk, vector_rows=vector_rows)
         else:
-            array_readwriter.get_array_parser(**fmt_meta).write(path, arr_chunk)
+            array_parser.write(path, arr_chunk)
         offset += n
         paths.append(path)
 
--- a/tests/tools/utils.py
+++ b/tests/tools/utils.py
@@ -9,15 +9,20 @@ from distpartitioning import array_readwriter
 from files import setdir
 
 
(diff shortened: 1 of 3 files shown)


tests/tools/utils.py
```diff
@@ -9,15 +9,20 @@
...
     offset = 0
 
     for j, n in enumerate(chunk_sizes):
         path = os.path.abspath(path_fmt % j)
         arr_chunk = arr[offset: offset + n]
+        shape = arr_chunk.shape
         logging.info("Chunking %d-%d" % (offset, offset + n))
-        array_readwriter.get_array_parser(**fmt_meta).write(path, arr_chunk)
+        # If requested we write multi-column arrays as single-column vector files
+        if vector_rows and len(shape) > 1 and shape[1] > 1:
```
thvasilo, line 22: This check can be pushed down to the parquet writer, not sure what is preferred here. 

To guarantee that any Parquet file written using the parser can be read using the parser we might need to do that.

tools/distpartitioning/array_readwriter/parquet.py
```diff
@@ -16,26 +17,44 @@ def read(self, path):
...
             logging.warning(
                 "Shape information not found in the metadata, read the data as "
                 "a 2 dim array."
             )
+        logging.info("Done reading from %s" % path)
         shape = tuple(eval(shape.decode())) if shape else arr.shape
         return arr.reshape(shape)
 
-    def write(self, path, array):
+    def write(self, path, array, vector_rows=False):
```
thvasilo, line 45: The change mostly relates to the reader, but in order to write the unit tests in a non-invasive way I added some modifications to the writer as well.

An alternative is to write a specific Parquet writer just for the tests and avoid making this modification to the writer.

Rhett-Ying:
@dgl-bot

tests/tools/utils.py
Rhett-Ying: writing a single-column is only available for `parquet`, right? we'd better add a format check here to capture any potential use `vector_rows=True` with `numpy`.

tools/distpartitioning/array_readwriter/parquet.py
Rhett-Ying: I think it's ok to add new arg here

Rhett-Ying (approved):
LGTM

tests/tools/utils.py
thvasilo: It is actually possible with Numpy too, using a [structured array](https://numpy.org/doc/stable/user/basics.rec.html), but not sure if/why one would produce such a file if they already have a regular 2D `ndarray`.

My question is more about where these checks should exist, inside `ParquetArrayParser` or only at the tests?

Rhett-Ying: only at tests should be ok as `.write()` is called here for now.

thvasilo: OK the new commit should make the writer only use vector_rows when we're writing to Parquet.

Rhett-Ying:
@dgl-bot

Rhett-Ying:
@dgl-bot

PR #5108, opened 2023-01-05
[Sparse] Remove SparseMatrix.indices and add device/dtype conversions

mufeili:
## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/sparse/diag_matrix.py
+++ b/python/dgl/sparse/diag_matrix.py
@@ -166,6 +166,163 @@ class DiagMatrix:
         """
         return DiagMatrix(self.val, self.shape[::-1])
 
+    def to(self, device=None, dtype=None):
+        """Perform matrix dtype and/or device conversion. If the target device
+        and/or dtype are already in use, the original matrix will be returned.
+
+        Parameters
+        ----------
+        device : torch.device, optional
+            This is the target device of the matrix. If not given, the current
+            device will be used.
+        dtype : torch.dtype, optional
+            This is the target data type of the matrix values. If not given,
+            the current data type will be used.
+
+        Returns
+        -------
+        DiagMatrix
+            The result matrix
+
+        Example
+        --------
+
+        >>> val = torch.ones(2)
+        >>> mat = diag(val)
+        >>> mat.to(device='cuda:0', dtype=torch.int32)
+        DiagMatrix(values=tensor([1, 1], device='cuda:0', dtype=torch.int32),
+                   size=(2, 2))
+        """
+        if device is None:
+            device = self.device
+        if dtype is None:
+            dtype = self.dtype
+
+        if device == self.device and dtype == self.dtype:
+            return self
+
+        return diag(self.val.to(device=device, dtype=dtype), self.shape)
+
+    def cuda(self):
+        """Move the matrix to GPU. If the matrix is already on GPU, the
+        original matrix will be returned.
+
+        Returns
+        -------
+        DiagMatrix
+            The matrix on GPU
+
+        Example
+        --------
+
+        >>> val = torch.ones(2)
+        >>> mat = diag(val)
+        >>> mat.cuda()
+        DiagMatrix(values=tensor([1., 1.], device='cuda:0'),
+                   size=(2, 2))
+        """
+        return self.to(device="cuda")
+
+    def cpu(self):
+        """Move the matrix to CPU. If the matrix is already on CPU, the
+        original matrix will be returned.
+
+        Returns
+        -------
+        DiagMatrix
+            The matrix on CPU
+
+        Example
+        --------
+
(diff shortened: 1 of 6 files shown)


python/dgl/sparse/diag_matrix.py
```diff
@@ -166,6 +166,163 @@ def transpose(self):
...
         return DiagMatrix(self.val, self.shape[::-1])
 
+    def to(self, device=None, dtype=None):
+        """Perform matrix dtype and/or device conversion. If the target device
+        and/or dtype are already in use, the original matrix will be returned.
+
+        Parameters
+        ----------
+        device : torch.device, optional
+            This is the target device of the matrix. If not given, the current
```
jermainewang, line 176: This is the target ... -> The target ...

python/dgl/sparse/sparse_matrix.py
```diff
@@ -194,6 +171,190 @@ def transpose(self):
...
+        """
+        if device is None:
+            device = self.device
+        if dtype is None:
+            dtype = self.dtype
+
+        if device == self.device and dtype == self.dtype:
+            return self
+
+        row, col = self.coo()
```
jermainewang, line 213: Use val_like instead of getting coo explicitly. COO may not be materialized.

mufeili: This only works if the device stays unchanged.

python/dgl/sparse/sparse_matrix.py
```diff
@@ -194,6 +172,192 @@ def transpose(self):
...
+        if device == self.device and dtype == self.dtype:
+            return self
+        elif device == self.device:
+            return val_like(self, self.val.to(dtype=dtype))
+        else:
+            row, col = self.coo()
+            row = row.to(device=device)
+            col = col.to(device=device)
+            val = self.val.to(device=device, dtype=dtype)
+            return from_coo(row, col, val, self.shape)
```
jermainewang, line 220: Still don't think this is the right way. Perhaps it needs a C++ operator, which will copy/cast whatever formats exist in the sparse matrix to the target.

python/dgl/sparse/diag_matrix.py
```diff
@@ -166,6 +167,163 @@ def transpose(self):
...
+        if dtype is None:
+            dtype = self.dtype
+
+        if device == self.device and dtype == self.dtype:
+            return self
+
+        return diag(self.val.to(device=device, dtype=dtype), self.shape)
+
+    def cuda(self):
+        """Move the matrix to GPU. If the matrix is already on GPU, the
```
frozenbugs, line 208: Will this always pick cuda:0 if multiple devices exist?

python/dgl/sparse/sparse_matrix.py
frozenbugs: I am not sure what's the difference here. And the dtype change is (likely) not a heavy ops in the training cycle, so probably not worth to optimize more?

python/dgl/sparse/sparse_matrix.py
```diff
@@ -194,6 +172,192 @@ def transpose(self):
...
+
+    def cuda(self):
```
frozenbugs, line 222: Ditto.

python/dgl/sparse/diag_matrix.py
frozenbugs: If so, add comment.

frozenbugs (approved):

python/dgl/sparse/diag_matrix.py
```diff
@@ -166,6 +167,163 @@ def transpose(self):
...
+
+        >>> val = torch.ones(2)
+        >>> mat = diag(val)
+        >>> mat.float()
+        DiagMatrix(values=tensor([1., 1.]),
+                   size=(2, 2))
+        """
+        return self.to(dtype=torch.float)
+
+    def double(self):
```
frozenbugs, line 267: not sure why do we need a list of APIs for fast dtype conversion.
The naming should be either toDouble() or asDouble() depends on whether this is a in-place conversion or not.

frozenbugs: If this is a pytorch convention then ignore my comment.

python/dgl/sparse/sparse_matrix.py
jermainewang: Discussed with @frozenbugs, we will table the optimization to post v1.0. cc @mufeili

jermainewang (approved):

python/dgl/sparse/sparse_matrix.py
frozenbugs: Filed https://github.com/dmlc/dgl/issues/5119 to track.

PR #5130, opened 2023-01-09
[DistDGL] Memory optimization to reduce memory footprint of the Dist Graph partitioning pipeline.

kylasa:
Use a python wrapper for the np.argsort() function for better usage of systems memory.

## Checklist
- [X] Code is well-documented


tools/distpartitioning/data_shuffle.py
```diff
@@ -586,6 +586,40 @@ def read_dataset(rank, world_size, id_lookup, params, schema_map):
 
     return node_tids, node_features, node_feat_tids, edge_data, edge_features, edge_tids, edge_feat_tids
 
+def reorder_data(params, world_size, data, key):
+    """
+    Auxiliary function used to sort node and edge data for the input graph.
+
+    Parameters:
+    -----------
+    params : argparser object
```
Rhett-Ying, line 595: seems only `num_parts` is needed so how about passing it directly instead of the whole `params`.

tools/distpartitioning/data_shuffle.py
```diff
@@ -755,16 +789,9 @@ def gen_dist_partitions(rank, world_size, params):
...
-    for local_part_id in range(params.num_parts//world_size):
-        idx = node_data[constants.NTYPE_ID+"/"+str(local_part_id)].argsort()
-        for k, v in node_data.items():
-            tokens = k.split("/")
-            assert len(tokens) == 2
-            if tokens[1] == str(local_part_id):
-                node_data[k] = v[idx]
-        idx = None
-    gc.collect()
+    node_data = reorder_data(params, world_size, node_data, constants.NTYPE_ID)
```
Rhett-Ying, line 792: what's the benefit of wrapping into a function? the code looks the same. less memory is used with function call?

kylasa: Problem is the numpy library itself. Whenever we call numpy functions like unique, argsort etc.. I noticed that the memory used by these numpy calls does not get released immediately. Usually it takes some time for the cleanup to happen. 

But for argsort, I noticed that if I wrap this function into a wrapper function and call the wrapper function instead of the numpy's argsort directly, then I noticed that the memory used by numpy's argsort was released back to the system immediately after the wrapper function call return to the caller.

kylasa: For some datasets, this does not mean anything. But for massively large datasets the amount of memory released back to the system can be as high as 100's of GB. Note that we are sorting billions of nodes/edges with these calls. Because of this we end up using the systems memory efficiently.

Rhett-Ying (approved):

tools/distpartitioning/data_shuffle.py
kylasa: Done.

PR #5151, opened 2023-01-11
[Sparse] Fix python doc for csc() and csr().

frozenbugs:
## Checklist
- [ ] Code is well-documented


python/dgl/sparse/sparse_matrix.py
```diff
@@ -103,7 +103,8 @@ def coo(self) -> Tuple[torch.Tensor, ...]:
         return self.c_sparse_matrix.coo()
 
     def csr(self) -> Tuple[torch.Tensor, ...]:
-        """Get the coordinate (COO) representation of the sparse matrix.
+        r"""Get the compressed sparse row (CSR) representation of the sparse
```
czkkkkkk, line 106: Do we always need to use r string for comments?

czkkkkkk (approved):

python/dgl/sparse/sparse_matrix.py
frozenbugs: not sure, as long as it renders correctly, should be fine. Some time we need r to avoid backslach escape.

PR #5157, opened 2023-01-12
Replace global cudaStream in Filter with runtime calls (fix #5153)

### LINKED ISSUE #5153: cuda_filter set cudaStream in global scope breaking multi-stream support
## 🐛 Bug

This line:
https://github.com/dmlc/dgl/blob/master/src/array/cuda/cuda_filter.cu#L21
Should be inside of the function calls so as to capture the current cuda stream.

I believe this has been an issue since multi-stream support was officially added, as what was once a `constexpr` default stream declared globally got converted to a call to get the current cuda stream by #4480 and #4503, but remained in global scope.

This bug essentially causes undefined behavior as alternate cuda streams may access uninitialized GPU memory.

nv-dlasalle:
## Description
Fix #5153 by determining the stream at function invocation

## Checklist
- [x] Code is well-documented

### DIFF
--- a/src/array/cuda/cuda_filter.cu
+++ b/src/array/cuda/cuda_filter.cu
@@ -18,8 +18,6 @@ namespace array {
 
 namespace {
 
-cudaStream_t cudaStream = runtime::getCurrentCUDAStream();
-
 template <typename IdType, bool include>
 __global__ void _IsInKernel(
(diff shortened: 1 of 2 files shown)


yaox12 (approved):
LGTM.

chang-l (approved):
Looks good to me too.

Issue #5065, opened 2022-12-22
dgl.radius_graph unexpected self loops


javiermcebrian:
## 🐛 Bug

dgl.radius_graph sometimes provides self loops when specified self_loop = False in version dgl-cu102==0.9.1 in linux

## To Reproduce

Below a test code for illustration, returning the number of self loops in the created graph.

```
import dgl
import torch
import numpy as np

def test_dgl_radius_graph_self_loops(maximum, minimum, n_nodes, radius):
    src, dst = dgl.radius_graph((maximum - minimum) * torch.rand(size=(n_nodes, 2)), r=radius, p=2, self_loop=False).edges()
    return np.sum([i==j for i, j in zip(src.tolist(), dst.tolist())])

maximum = 3000
minimum = 0
n_nodes = 27

test_dgl_radius_graph_self_loops(maximum=maximum, minimum=minimum, n_nodes=n_nodes, radius=2000)
test_dgl_radius_graph_self_loops(maximum=maximum, minimum=minimum, n_nodes=n_nodes, radius=3000)
```

## Expected behavior

I would expect no self loops in the returned graph

## Environment

 - DGL Version (e.g., 1.0): dgl-cu102==0.9.1
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): torch==1.13.0+cu117
 - OS (e.g., Linux): nvidia/cuda:10.2-cudnn7-devel-ubuntu18.04
 - How you installed DGL (`conda`, `pip`, source): pip
 - Build command you used (if compiling from source): N/A
 - Python version: 3.8
 - CUDA/cuDNN version (if applicable): CUDA 10.2 cuDNN 7
 - GPU models and configuration (e.g. V100): Tesla P100

mufeili:
It seems that I got 0 with `(src == dst).sum()`.

dmonterom:
I get the same error when using radius = 3000. It seems to be a rounding issue inside radius_graph function:
`if not self_loop:
    distances.fill_diagonal_(r + 1e-4)
edges = th.nonzero(distances <= r, as_tuple=True)`

It is weird because using radius = 3000.0001 works well.

mufeili:
Could you take a look? @ksadowski13

javiermcebrian:
@mufeili thanks, I saw PR is merged into master. Do you know in which DGL version will the fix be applied? Thanks!

mufeili:
[...]

PR #5159, opened 2023-01-12
[Fix] Radius graph returning self loops

ksadowski13:
## Description
Fixes rounding issue that caused `radius_graph` returning graph with self loops even when `self_loop=False`. It relates to https://github.com/dmlc/dgl/issues/5065.

## Checklist
- [x] Code is well-documented

## Changes
- Change epsilon value added to diagonal of distance matrix to `1` instead of `1e-4`


mufeili:
@dmonterom Could you have a try and see if this addresses your issue?

mufeili:
@ksadowski13 Could you also add a unit test for this case?

mufeili:
@dgl-bot

mufeili (approved):

PR #5148, opened 2023-01-11
[Sparse] Add more complete elementwise op

BarclayII:
## Description
This completes the following:
* Support sparse matrix subtraction: `A - B`
* Support scalar operands that are PyTorch scalar tensors instead of Python numbers in `*`, `/` and `**`.
* Support scalar operand dividing a diagonal: `v / D`.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [x] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [x] All changes have test coverage
- [x] Code is well-documented
- [x] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

PR #5160, opened 2023-01-12
[Sparse] Use NotImplemented to let Python dispatch types

BarclayII:
## Description
This is a Sub-PR of #5148 .

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/sparse/elementwise_op_diag.py
+++ b/python/dgl/sparse/elementwise_op_diag.py
@@ -43,10 +43,9 @@ def diag_add(
         )
         D1 = D1.as_sparse()
         return D1 + D2
-    raise RuntimeError(
-        "Elementwise addition between "
-        f"{type(D1)} and {type(D2)} is not supported."
-    )
+    # Python falls back to D2.__radd__(D1) then TypeError when NotImplemented
+    # is returned.
+    return NotImplemented
 
 
 def diag_sub(D1: DiagMatrix, D2: DiagMatrix) -> DiagMatrix:
@@ -78,21 +77,20 @@ def diag_sub(D1: DiagMatrix, D2: DiagMatrix) -> DiagMatrix:
             f"{D1.shape} and D2 {D2.shape} must match."
         )
         return diag(D1.val - D2.val, D1.shape)
-    raise RuntimeError(
-        "Elementwise subtraction between "
-        f"{type(D1)} and {type(D2)} is not supported."
-    )
+    # Python falls back to D2.__rsub__(D1) then TypeError when NotImplemented
+    # is returned.
+    return NotImplemented
 
 
 def diag_mul(
-    D1: Union[DiagMatrix, float, int], D2: Union[DiagMatrix, float, int]
+    D1: DiagMatrix, D2: Union[DiagMatrix, float, int]
 ) -> DiagMatrix:
     """Elementwise multiplication
 
     Parameters
     ----------
-    D1 : DiagMatrix or float or int
-        Diagonal matrix or scalar value
+    D1 : DiagMatrix
+        Diagonal matrix
     D2 : DiagMatrix or float or int
         Diagonal matrix or scalar value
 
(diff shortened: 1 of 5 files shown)


python/dgl/sparse/elementwise_op_sp.py
```diff
@@ -42,24 +42,19 @@ def sp_add(A: SparseMatrix, B: Union[DiagMatrix, SparseMatrix]) -> SparseMatrix:
...
-        B = B.as_sparse()
-    if isinstance(A, SparseMatrix) and isinstance(B, SparseMatrix):
-        return spsp_add(A, B)
-    raise RuntimeError(
-        "Elementwise addition between {} and {} is not "
-        "supported.".format(type(A), type(B))
-    )
+    # Python falls back to B.__radd__ then TypeError when NotImplemented is
+    # returned.
+    return spsp_add(A, B) if isinstance(B, SparseMatrix) else NotImplemented
```
mufeili, line 47: Wait a moment. Does this mean that `A + I` is no longer supported?

BarclayII: When Python tries `A.__add__(I)` that returns `NotImplemented`, Python will then try `I.__radd__(A)`.

python/dgl/sparse/elementwise_op_diag.py
```diff
@@ -78,21 +77,20 @@ def diag_sub(D1: DiagMatrix, D2: DiagMatrix) -> DiagMatrix:
...
-        f"{type(D1)} and {type(D2)} is not supported."
-    )
+    # Python falls back to D2.__rsub__(D1) then TypeError when NotImplemented
+    # is returned.
+    return NotImplemented
 
 
 def diag_mul(
-    D1: Union[DiagMatrix, float, int], D2: Union[DiagMatrix, float, int]
+    D1: DiagMatrix, D2: Union[DiagMatrix, float, int]
```
mufeili, line 86: Is this valid since we have both `DiagMatrix.__mul__ = diag_mul` and `DiagMatrix.__rmul__ = diag_mul`?

BarclayII: It is.
`D1 * D2 == D1.__mul__(D2) == diag_mul(D1, D2)` works normally.
`D1 * v == D1.__mul__(v) == diag_mul(D1, v)` also works.
`v * D1 == v.__mul__(D1)` will fail initially since the scalar multiplication doesn't support `DiagMatrix`.  Python will then try `D1.__rmul__(v) == diag_mul(D1, v)` which should be correct since multiplication is commutative.
`__rsub__` and `__rtruediv__` should have different routines from `__sub__` and `__truediv__` though, which is why I removed them in this PR.

python/dgl/sparse/elementwise_op_sp.py
```diff
@@ -89,14 +81,15 @@ def sp_mul(
     values=tensor([2, 4, 6]),
     shape=(3, 4), nnz=3)
     """
-    if isinstance(A, SparseMatrix) and isinstance(B, (float, int)):
+    if isinstance(B, (float, int)):
         return val_like(A, A.val * B)
-    elif isinstance(A, (float, int)) and isinstance(B, SparseMatrix):
+    elif isinstance(B, SparseMatrix):
```
mufeili, line 86: What if both `A` and `B` are SparseMatrix objects? I got the error message "TypeError: unsupported operand type(s) for *: 'SparseMatrix' and 'Tensor'", which can be confusing.

mufeili (approved):
Minor comments

python/dgl/sparse/elementwise_op_sp.py
BarclayII: In general one should raise an "unsupported operand" TypeError if the two types are incompatible for some operation.
```
>>> torch.tensor([2,3,4]) / 'aaa'
TypeError: unsupported operand type(s) for /: 'Tensor' and 'str'
```

frozenbugs (approved):

frozenbugs:
please fix lintrunner warning in test.

PR #5163, opened 2023-01-12
[Sparse] Make functions compatible with PyTorch scalar tensors

BarclayII:
## Description
A sub PR of #5148 .

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/sparse/utils.py
+++ b/python/dgl/sparse/utils.py
@@ -1,0 +1,14 @@
+"""Utilities for DGL sparse module."""
+from numbers import Number
+from typing import Union
+
+import torch
+
+
+def is_scalar(x):
+    """Check if the input is a scalar.
+    """
+    return isinstance(x, Number) or (torch.is_tensor(x) and x.dim() == 0)
+
+# Scalar type annotation
+Scalar = Union[Number, torch.Tensor]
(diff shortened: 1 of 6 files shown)


python/dgl/sparse/utils.py
```diff
@@ -0,0 +1,14 @@
+"""Utilities for DGL sparse module."""
```
frozenbugs, line 1: nit: is this sparse specific utils?

tests/pytorch/sparse/test_elementwise_op_sp.py
```diff
@@ -20,7 +20,9 @@ def all_close_sparse(A, row, col, val, shape):
     assert A.shape == shape
 
 
-@pytest.mark.parametrize("v_scalar", [2, 2.5])
+@pytest.mark.parametrize(
+    "v_scalar", [2, 2.5, torch.tensor(2), torch.tensor(2.5)]
```
mufeili, line 24: Do we expect to support things like `torch.tensor([2])`?

mufeili (approved):

tests/pytorch/sparse/test_elementwise_op_sp.py
BarclayII: I think so.  It allows us to backpropagate through the scalar coefficient at least.

python/dgl/sparse/utils.py
BarclayII: So far it's sparse specific (really PyTorch specific).  Do you have a better place for it?

PR #5178, opened 2023-01-16
[Sparse] Polish diag_matrix.py and sparse_matrix.py.

frozenbugs:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/sparse/diag_matrix.py
+++ b/python/dgl/sparse/diag_matrix.py
@@ -24,6 +24,9 @@ class DiagMatrix:
         self._val = val
         self._shape = shape
 
+    def __repr__(self):
+        return _diag_matrix_str(self)
+
     @property
     def val(self) -> torch.Tensor:
         """Returns the values of the non-zero elements.
@@ -46,9 +49,6 @@ class DiagMatrix:
         """
         return self._shape
 
-    def __repr__(self):
-        return _diag_matrix_str(self)
-
     @property
     def nnz(self) -> int:
         """Returns the number of non-zero elements in the diagonal matrix.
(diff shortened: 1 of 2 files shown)


python/dgl/sparse/diag_matrix.py
```diff
@@ -8,23 +8,7 @@
...
-        If not specified, it will be inferred from :attr:`val`, i.e.,
-        (N, N). Otherwise, :attr:`len(val)` must be equal to :attr:`min(shape)`.
-
-    Attributes
-    ----------
-    val : torch.Tensor
-        Diagonal of the matrix.
-    shape : tuple[int, int]
-        Shape of the matrix.
-    """
```
jermainewang, line 27: These should be recovered. See https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes .

frozenbugs: I am removing this since this is not the concrete list of attributes, and we also did not doc SparseMatrix.
If needed, we should add all Attributes for both SparseMatrix and DiagMatrix.
My feeling is the list could be too long to list here which makes the doc string unreadable, and we don't have a rst file relying on this.

frozenbugs: If you believe it is necessary to list all the Attributes, let's do it properly in a separated PR.

jermainewang: Oh, I'm not talking about attributes, but the docstring of `__init__` (i.e., the parameter section).

frozenbugs: Added.

jermainewang (approved):

PR #5210, opened 2023-01-19
[Bugfix] fixed leak in SpMMCreateBlocks

Kacper-Pietkun:
## Description

Fixed memory leak that happened in SpMMCreateBlocks

memory usage before fix - constantly growing with each epoch
image

memory usage after fix - flat chart
image

Charts were generated while running graphsage model (examples/pytorch/ogb/ogbn-products/graphsage/main.py) for 14 epochs and with cpu

## Checklist


bartekkuncer:
image
Images seem to be broken.

Kacper-Pietkun:
> image Images seem to be broken.

It seems fine for me.

bartekkuncer:
> > image Images seem to be broken.
> 
> It seems fine for me.

For some reason I can only see them after clicking on the image link. I guess there is a problem on my side. Sorry for the false alarm.

yaox12:
@dgl-bot

src/array/cpu/spmm_blocking_libxsmm.h
```diff
@@ -80,10 +80,16 @@ inline void SpMMCreateBlocks(
...
+    IdType *indices_block_buf = nullptr;
+    if (use_lhs) {
+      indices_block_buf = reinterpret_cast<IdType *>(
+          aligned_alloc(64, indptr[M] * sizeof(IdType)));
+    }
+    IdType *edges_block_buf = nullptr;
+    if (use_rhs) {
+      edges_block_buf = reinterpret_cast<IdType *>(
+          aligned_alloc(64, indptr[M] * sizeof(IdType)));
+    }
```
BarclayII, line 92: Could you add a comment on why this change fixes the memory leak?  As far as I remember `aligned_alloc(3)` doesn't require using `free(3)` to reclaim memory (https://linux.die.net/man/3/aligned_alloc).  Since I didn't see any `free(3)` call I think there should be other reasons.

Kacper-Pietkun: Memory, which is  allocated here:
https://github.com/dmlc/dgl/pull/5210/files#diff-09f913927175aed9edd405df276b75d0c56b07ca917a887ab5ed19344e960f07R83-R92
image

gets deallocated here:
https://github.com/dmlc/dgl/pull/5210/files#diff-09f913927175aed9edd405df276b75d0c56b07ca917a887ab5ed19344e960f07R401-R411
image

As you can see memory is deallocated only when "use_lhs" or "use_rhs" flags are equal to "true". So to avoid a leak, appropriate "if statements" have to be added before allocation.

aligned_alloc requires using free in order to avoid a memory leak according to the spec https://en.cppreference.com/w/c/memory/aligned_alloc

bgawrych (approved):

BarclayII (approved):

BarclayII:
@dgl-bot

PR #5223, opened 2023-01-25
[DistDGL][]Fix pipeline's order sensitivity to keys in the dictionaries (metadata.json)

kylasa:
Fixed the pipeline's sensitivity to the order of the keys in the "edges", "edge_type", "edge_data" dictionaries in the metadata.json file. We will always follow the order defined in the EDGE_TYPE list. Files will be read in this particular order from the disk.

## Description

When the EDGES dictionary in the input metadata.json file is out-of-order compared to EDGE_TYPE dictionary then in this case edges will be loaded in a different order (rest of the pipeline assumes that the edges are read from the disk in the order defined by EDGE_TYPE dictionary). This is the cause of inconsistency when determining the owner edges of the edge features in a given node. In Houyu's case, which is very peculiar case, where there are two edge types with equal no. of edges this inconsistency was resulting in different no. of edges and edge features in the partitioned graph objects. 

Solution to resolve this inconsistency is to enforce the order of reading edges from the disk, when the dataset is being built. Now in the changed code, we iterate over the edges in the EDGE_TYPE list and read the appropriate edge files. This will ensure that edges are ordered in non-decreasing order of edge_type in the edge_data dictionary. 

## Checklist
- [X] Code is well-documented


tools/distpartitioning/dataset_utils.py
```diff
@@ -501,8 +501,14 @@ def get_dataset(input_dir, graph_name, rank, world_size, num_parts, schema_map):
...
         edge_datadict[col] = []
 
-    for etype_name, etype_info in edge_data.items():
-        edge_info = etype_info[constants.STR_DATA]
+    # Load the edges in the order defined by STR_EDGE_TYPE. 
+    # This order is also used to assign GLOBAL_EID to edges in the function
+    # 'augment_edge_data'
+    # GLOBAL_EID is used to exchange edge features
+
+    for etype_name, etype_id in etype_name_idmap.items():
```
classicsong, line 509: The only thing changed is how etype_names are iterated. Why does changing the way it iterates over etypes matter?

Rhett-Ying (changes requested):
could you add test to elaborate the case which hit this inconsistency issue?

tools/distpartitioning/dataset_utils.py
kylasa: Hey Xiang, The issue with the search ctr graph was that the EDGE_TYPE list and the EDGE_DATA dictionary has different order of edge types. In the case of search ctr graph, the data was loaded by the nodes in the order specified by the EDGE_DATA order. where as the EDGE_TYPE order was used to assign EDGE_ID's which was creating issues because the edge_ids  were getting flipped. 

Now with this fix, we always assign EDGE_ID's based on the order mentioned in the EDGE_TYPE list.

kylasa:
Rui, Per your suggestion I have modified the way we create the chunked dataset, so that now we do not sort the keys in the input metadata.json file. With this change I also ensure that the "EDGE_DATA" and "EDGE_TYPE" keys are different most of the times. This will make sure that in every test case we run we make sure that we cover the inconsistency reported in this PR.

kylasa:
For some reason, cannot commit into this branch any longer. It appears that the base branch changed... and the git Desktop is failing to sync local and remote branches. Following error is being reported 
Screenshot 2023-02-06 at 1 24 56 PM

kylasa:
Will be opening a new branch to merge this code changes using the new PR.

kylasa:
merged into the master branch using https://github.com/dmlc/dgl/pull/5288

PR #5225, opened 2023-01-25
[DistDGL][Mem_Optimizations]Edge Ownership processes are computed on the fly when required.

kylasa:
Earlier we were storing Edge ownership processes after the dataset was retrieved from the disk. For massively large datasets, each node can handle upto 5 Billion edges, this means storing owner process-ids will consume 5 * 8 = 40GB. This memory will be hanging around until the edges are exchanged.

To reduce the memory footprint of the pipeline, we no longer store the ownership process-ids in the 'edge_data' dictionary after reading the dataset from the disk. Instead, we compute them on the fly at the time of exchanging edges.

Another optimization is not to send/receive all the messages in a one single large message. Instead we now split the total number edges into chunks, limited by 8 GB per node. And we iterate until all the chunks are exchanged.

Once all the edges are exchanged, as a sanity check, we compute the total number of edges in the system and compare it with the original value before edge shuffling, in a final assert statement before return the result to the caller.

## Checklist
- [X] Code is well-documented


tools/distpartitioning/data_shuffle.py
```diff
@@ -155,38 +156,83 @@ def exchange_edge_data(rank, world_size, num_parts, edge_data):
...
-        edge_data[constants.GLOBAL_DST_ID+"/"+str(local_part_id)] = rcvd_edge_data[:,1]
-        edge_data[constants.GLOBAL_TYPE_EID+"/"+str(local_part_id)] = rcvd_edge_data[:,2]
-        edge_data[constants.ETYPE_ID+"/"+str(local_part_id)] = rcvd_edge_data[:,3]
-        edge_data[constants.GLOBAL_EID+"/"+str(local_part_id)] = rcvd_edge_data[:,4]
+    CHUNK_SIZE = 100 * 1000 * 1000 # 100 * 8 * 5 = 1 * 4 = 8 GB/message/node 
+    num_edges = edge_data[constants.GLOBAL_SRC_ID].shape[0]
+    all_counts = allgather_sizes([num_edges], world_size, num_parts, return_sizes=True)
+    max_edges = np.amax(all_counts)
+    all_edges = np.sum(all_counts)
+    num_chunks = (max_edges // CHUNK_SIZE) + (0 if (max_edges % CHUNK_SIZE == 0) else 1)
```
Rhett-Ying, line 165: `math.ceil()` or `numpy.ceil()`?

Rhett-Ying (approved):
LGTM

tools/distpartitioning/data_shuffle.py
kylasa: Agreed with this suggestion. 
Will make this change in the next iteration of this PR, In which I intend to perform numpy optimizations for the unique function. This PR has been sitting in the queue for various reasons for the past 3 weeks (either for lintrunner reasons or for some other reasons). I will commit this PR for now.

PR #5226, opened 2023-01-26
[DistDGL][Mem_Optimizations]get_partition_ids, service provided by the distributed lookup service has high memory footprint

kylasa:
'get_partitionid' function, which is used to retrieve owner processes of the given list of global node ids, has high memory footprint. Currently this is of the order of 8x compared to the size of the input list.

For massively large datasets, this memory needs are very unrealistic and may result in OOM. In the case of CoreGraph, when retrieving owner of an edge list of size 6 Billion edges, the memory needs can be as high as 8*8*8 = 256 GB.

To limit the amount of memory used by this function, we split the size of the message sent to the distributed lookup service, so that each message is limited by the number of global node ids, which is 200 million. This reduced the memory footprint of this entire function to be no more than 0.2 * 8 * 8 = 13 GB. which is within reasonable limits.

Now since we send multiple small messages compared to one large message to the distributed lookup service, this may consume more wall-clock-time compared to earlier implementation.

## Checklist
- [X] Code is well-documented


tools/distpartitioning/dist_lookup.py
```diff
@@ -136,107 +138,134 @@ def get_partition_ids(self, global_nids):
...
-                local_order_idx.append(np.where(cond)[0])
-                global_type_nids = type_nids[cond]
-                if len(global_type_nids) <= 0:
-                    continue
-
-                local_type_nids = global_type_nids - self.type_nid_begin[tid]
-
-                assert np.all(local_type_nids >= 0)
-                assert np.all(local_type_nids <= (self.type_nid_end[tid] + 1 - self.type_nid_begin[tid]))
+        CHUNK_SIZE = 200 * 1000 * 1000
```
Rhett-Ying, line 141: how about letting user to specify it via environment variable or any other ways to have a better control on the trade-off between memory and performance?

tools/distpartitioning/dist_lookup.py
```diff
@@ -136,107 +138,134 @@ def get_partition_ids(self, global_nids):
...
+        # Determine the no. of times each process has to send alltoall messages.
+        local_rows = agg_global_nids.shape[0]
+        all_sizes = allgather_sizes([local_rows], self.world_size, self.num_parts, return_sizes=True)
+        max_count = np.amax(all_sizes)
+        num_splits = (max_count // CHUNK_SIZE) + (0 if (max_count % CHUNK_SIZE == 0) else 1)
```
Rhett-Ying, line 146: `ceil()`?

Rhett-Ying:
> get_partitionid' function, which is used to retrieve owner processes of the given list of global node ids, has high memory footprint. Currently this is of the order of 8x compared to the size of the input list.

do we know why it's `8x`?

tools/distpartitioning/dist_lookup.py
kylasa: Yes. This is what I am thinking as well. 
But we should take care of this in a separate PR. Because we had a total of 3 PRs (5226, 5225 and other PR which changes chunks the get_partitionids () function) where we are chunking the messages. I would like to add user specified command line parameters to all these 3 PRs using one PR - otherwise this will get complicated. 

Please approve this PR in the current state. I will address the issue of determining the chunk_size using another PR, so that amount of code changes are minimal and easy to do the code review.

tools/distpartitioning/dist_lookup.py
kylasa: Agreed with the suggestion. Implemented this change per suggestion.

kylasa:
Rhett Ying, 

The requirement of 8x memory is because of the temp. variables we maintain during the get_partitionid function (like type_nids, ntype_ids, owner_ids - tensors to send/receive messages). And additionally, we perform argsort(), which also consumes about 3x memory. 

And on-top of that, we cannot call get_partition_ids with a very long list. In the case of FeLarge graph this call can be processing nodes list which is as long as 3 Billion nodes. 

Because of the above reason, we need to chunk this message into multiple smaller messages so that we can control the amount of memory used by this function call. 

_Sudhir

tools/distpartitioning/dist_lookup.py
```diff
@@ -113,8 +117,9 @@ def __init__(self, input_dir, ntype_names, id_map, rank, world_size):
         self.ntype_count = np.array(ntype_count, dtype=np.int64)
         self.rank = rank
         self.world_size = world_size
+        self.num_parts = num_parts
 
-    def get_partition_ids(self, global_nids):
+    def get_partition_ids(self, agg_global_nids):
```
Rhett-Ying, line 122: why prepending `agg_` to the variable name?

tools/distpartitioning/dist_lookup.py
```diff
@@ -136,127 +141,172 @@ def get_partition_ids(self, global_nids):
...
-
-        # For each node-type, the per-type-node-id <-> partition-id mappings are
-        # stored as contiguous chunks by this lookup service.
-        # The no. of these mappings stored by each process, in the lookup service, are
-        # equally split among all the processes in the lookup service, deterministically.
-        typeid_counts = self.ntype_count[ntype_ids]
-        chunk_sizes = np.ceil(typeid_counts / self.world_size).astype(np.int64)
-        service_owners = np.floor_divide(type_nids, chunk_sizes).astype(
-            np.int64
+        CHUNK_SIZE = 200 * 1000 * 1000
```
Rhett-Ying, line 155: does the existing unit test cover the change? the size in test is larger than the chunk_size defined here? Just make sure chunk to multiple ones work as expected.

tools/distpartitioning/dist_lookup.py
kylasa: With this change now, this function now supports sending multiple smaller messages instead of one large message. So I used 'agg_', in this case, 'agg_global_nids' to indicate 'aggregate global nids' which this function is processing. When sending each smaller message out, we slice agg_global_nids to get the appropriate message slice and then send it for processing. Similarly the response's to get_partitionids is aggregated (stored) in the agg_partition_ids numpy array. 

At the end of all processing, agg_partition_ids are returned to the. caller. 

If you disagree with this naming convention, please suggest an appropriate names for the variables and I will make the necessary code changes.

tools/distpartitioning/dist_lookup.py
kylasa: Existing test case covers this case as well, but only one corner case is covered always !!!. 
To improve the test-coverage and I plan to make the chunk_sizes (they are used in 3 places in this pipeline now) to be configurable by the user itself from the command line arguments. I plan to do these changes using a separate PR. 

Once that change is done, we can run the command-line with various arguments and try testing for multiple scenarios. 

If this is acceptable, please approve the PR immediately !!!.

Rhett-Ying (approved):

PR #5227, opened 2023-01-26
[DistDGL][Robustness]Uneven distribution of input graph files for nodes/edges and features.

kylasa:
To handle unevenly sized files for nodes/edges and feature files for nodes and edges, we have to synchronize before starting large no. of messages (either one large message or a burst of messages).

## Checklist
- [X] Code is well-documented


Rhett-Ying (commented):
CI is failing, pls take a look...

tools/distpartitioning/data_shuffle.py
```diff
@@ -142,6 +146,27 @@ def gen_node_data(
 
         for local_part_id in range(num_parts // world_size):
             cond = node_partid_slice == (rank + local_part_id * world_size)
+
+        # No. of nodes in each process can differ significantly in lopsided distributions
+        # Synchronize on a per ntype basis
+        dist.barrier()
+
+        type_start, type_end = (
```
Rhett-Ying, line 154: are these 2 variables are used?

tools/distpartitioning/data_shuffle.py
```diff
@@ -142,6 +146,27 @@ def gen_node_data(
...
+        gnid_start, gnid_end = (
+            global_nid_dict[ntype_name][0, 0],
+            global_nid_dict[ntype_name][0, 1],
+        )
+
+        node_partid_slice = id_lookup.get_partition_ids(
+            np.arange(gnid_start, gnid_end, dtype=np.int64)
+        )  # exclusive
+
+        for local_part_id in range(num_parts // world_size):
```
Rhett-Ying, line 167: duplicate to line 149~150. seem previous ones should be removed.

Rhett-Ying:
what's the consequence of `do not sync`? The following steps will run with incorrect data?

Rhett-Ying (approved):
pls address the comments before merge.

tools/distpartitioning/data_shuffle.py
kylasa: This file was completely messed up during the merge conflict resolution. The same for-loop was repeated twice and with wrong indentation. So the CI was bound to fail. 

And to answer your question. Yes, these two variables are used in the for-loop. Now the code is cleaned up and you should be able to see these variables. clearly.

tools/distpartitioning/data_shuffle.py
kylasa: Lot of repeated/duplicate code was introduced after the merge conflict resolutions were made. 
It appears that the merge conflicts are very tricky and github fails more often then we thought. 

Now cleaned up all the issues, which were introduced because of the merge conflicts.

kylasa:
with the uneven file sizes assumption, we have to think of corner cases where some ranks of massively large data and some ranks might not have any data. Now when the message exchange happens, if the message exchange does not happen in lock-step fashion it will be hard to debug. 1. Some of the barrier calls are introduced so that all the ranks each major step of the algorithm in a lock-step manner. This is helping in debugging. 2. Secondly, a process may time out because some other rank is stuck in processing application logic which some ranks have already in the messaging call. I wanted to avoid this situation so I introduced barrier so that messaging calls are invoked simultaneously on all the ranks. This helps in debugging as well. 

In future, during rigorous testing with massively large datasets, some of these barrier calls might be removed. But I have to have enough proof the the application works properly without any issues in these cases. Until then I plan to leave these barrier calls.

PR #5240, opened 2023-01-30
Fix Column.record_stream(...) for unmaterialized tensors

### LINKED ISSUE #5239: [Bug][Performance] Significant overhead when `record_stream` for subgraph in dataloading
## 🐛 Bug

It seems there are many overheads observed in dataloader during feature prefetching when `use_alternate_streams=True` . When it calls `_record_stream` for every subgraph (batch), like the following: 
https://github.com/dmlc/dgl/blob/829ce109f1604ee59ac039331a3609896301e8de/python/dgl/dataloading/dataloader.py#L339
many unnecessary h2d copies (or gpu feature slicing if use_uva) are invoked from L339, due to materializing the column data(feature) within its `record_stream` method, i.e., within its `data` method, for example from here (when uva is disabled)
https://github.com/dmlc/dgl/blob/829ce109f1604ee59ac039331a3609896301e8de/python/dgl/frame.py#L246-L251
called by `record_stream` method for each column:
https://github.com/dmlc/dgl/blob/829ce109f1604ee59ac039331a3609896301e8de/python/dgl/frame.py#L551

It poses significant slow-down for both cases: `use_uva=True` and `use_uva=False`, for example, when running `example/graphsage/node_classification.py`, with uva on, `use_alternate_streams=True` causes **1.56x** more time for each epoch, while with uva off, `use_alternate_streams=True` (default) causes **1.55x** more time for each epoch.

## To Reproduce

set `use_alternate_streams=True` and run

```
python node_classification.py
```

## Environment

 - DGL Version (e.g., 1.0): 1.0
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): Pytorch 1.13
 - How you installed DGL (`conda`, `pip`, source): source
 - GPU models and configuration (e.g. V100): A5000

nvcastet:
## Description

`record_stream` should not force a column to materialize its tensor.
This PR skips `record_stream` when column is not materialized.

Fixes #5239


chang-l:
@dgl-bot

TristonC:
@jermainewang Bring this to your attention, this will improve the performance of dgl dataloader by removing overhead.

TristonC:
@nvcastet please do the lint scan as suggested.

yaox12 (approved):
LGTM.
~~Just a small question, IIUC, this is to avoid materializing features that won't be fetched when calling `record_stream` on MFGs, right?~~
Never mind, I see the issue.

yaox12:
@dgl-bot

nvcastet:
@yaox12 Yes good catch!

Can someone re-trigger CI for me?

yaox12:
@dgl-bot

yaox12:
@dgl-bot

yaox12:
@dgl-bot

yaox12:
@dgl-bot

yaox12 (approved):

nvcastet:
Thanks @yaox12 for fixing up the remaining issues!

python/dgl/frame.py
```diff
@@ -548,7 +559,14 @@ def record_stream(self, stream):
         """
         if F.get_preferred_backend() != "pytorch":
             raise DGLError("record_stream only supports the PyTorch backend.")
-        self.data.record_stream(stream)
+        # Only record streams for materialized tensors otherwise just record for the index
```
nv-dlasalle, line 562: Could there be an issue if `record_stream` gets called followed by a call to
```
@data.setter
def data(self, val):
```
Or does it make more sense that when assigning data or an index to a column, the stream associated with that column becomes the from which the data or index is associated it?

python/dgl/frame.py
```diff
@@ -548,7 +559,14 @@ def record_stream(self, stream):
...
+        if self.index is None:
+            self.data.record_stream(stream)
+        elif (
+            isinstance(self.index, _LazyIndex)
+            or F.context(self.index) != F.cpu()
+        ):
+            self.index.record_stream(stream)
```
nv-dlasalle, line 569: Should we be calling `self.storage.record_stream(stream)` whenever `index` exists?

nv-dlasalle (approved):
In general it looks good, but I'm not sure about some of the edge cases in `Column.record_stream()` and the resulting desired behavior.

python/dgl/frame.py
yaox12: How about
```python
        if (
            isinstance(self.index, _LazyIndex)
            or F.context(self.index) != F.cpu()
        ):
            self.index.record_stream(stream)
        if F.context(self.storage) != F.cpu():
            self.storage.record_stream(stream)
```

python/dgl/frame.py
```diff
@@ -548,7 +559,14 @@ def record_stream(self, stream):
         """
         if F.get_preferred_backend() != "pytorch":
             raise DGLError("record_stream only supports the PyTorch backend.")
-        self.data.record_stream(stream)
+        # Only record streams for materialized tensors otherwise just record for the index
+        if self.index is None:
+            self.data.record_stream(stream)
```
chang-l, line 564: Want to point out that fix in this PR is not enough to avoid unnecessary H2D copies by just checking if `self.index` is materialized. In non-uva mode when `self.storage` is on cpu, the extra H2D copies can be launched from here. https://github.com/dmlc/dgl/blob/aa42aaeb9f499cce5af946f7e4aed15d7661cac8/python/dgl/frame.py#L247-L248

chang-l: Sorry, it seems these H2D copies in `data` method are necessary for fetching non-`lazy-feature`... right? @nv-dlasalle @yaox12 
Should look into it more closely :)

python/dgl/frame.py
yaox12: @chang-l How do you like this modification?

nvcastet: @yaox12 @chang-l Sounds good! I added a commit with recording on storage. I think if all tests pass, it should be ready to merge.

python/dgl/frame.py
nv-dlasalle: That sounds good to me

python/dgl/frame.py
nv-dlasalle: @chang-l I think you were right, we dont want this function to cause any deferred actions to take place (`device` or `dtype`).

python/dgl/frame.py
```diff
@@ -548,7 +559,17 @@ def record_stream(self, stream):
...
+        else:
+            if (
+                isinstance(self.index, _LazyIndex)
+                or F.context(self.index) != F.cpu()
+            ):
+                self.index.record_stream(stream)
+            if F.context(self.storage) != F.cpu():
+                self.storage.record_stream(stream)
```
nv-dlasalle, line 572: If I store the main graph in GPU memory, and then generate subgraphs on a new stream, I wouldn't want to record the stream on the feature storage of the original graph.

I'm not sure of the right course of action however, as adding a deferred stream would seem to just make matters worse.

nvcastet: I feel that is the only solution if the column will be materialized later.
I don't think there is an issue of stream recording the main graph.

nv-dlasalle: You're right, I was misunderstanding the semantics of record_stream. So for correctness we want to call record stream on both the index and the storage (the only actual tensors associated with a column), but never on `.data`?

nvcastet: `.data` is a property (function call) that would trigger the costly `tensor.select` operation. So yes we want to call record stream on index and storage that will be used later if the column gets materialized.

nv-dlasalle: This looks good to me. Thanks!

python/dgl/frame.py
chang-l: I see... So, when `self.index` is `None`,  we should just have sth like below? Am I getting this correctly?   
```
if self.index is None:
    if F.context(self.storage) != F.cpu():
        self.storage.record_stream(stream)

Comment (nvcastet) [reply python/dgl/frame.py:564]: Correct.

File: python/dgl/frame.py [hunk 5]
```diff
@@ -548,7 +559,17 @@ def record_stream(self, stream):
         """
         if F.get_preferred_backend() != "pytorch":
             raise DGLError("record_stream only supports the PyTorch backend.")
-        self.data.record_stream(stream)
+        # Only record streams for materialized tensors otherwise just record for the index
+        if self.index is None:
+            self.data.record_stream(stream)
```
yaox12, line 564: Do we still need recording stream on `data`?

nvcastet: no we can still record on the `storage`. I just pushed a new commit. Let me know if that looks good.

nvcastet:
Can someone trigger CI?

yaox12:
@dgl-bot

yaox12:
@dgl-bot

yaox12:
@dgl-bot

yaox12:
@dgl-bot

yaox12:
@dgl-bot

PR #5241, opened 2023-01-31
[Performance]Add concurrent cpu id hashmap

peizhou001:
## Description
This PR add a new class `CpuIdHashMap`, which is a concurrent id hash map. Currently It is specified to be used in `ToBlockCPU` for optimization.  It maps an array of Ids which is duplicate and non-contiguous to an array which is unique and contiguous.  And also return the unique elements in original array.   The public method used outside are:

1. Init
2. Map

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/src/array/cpu/cpu_id_hash_map.cc
+++ b/src/array/cpu/cpu_id_hash_map.cc
@@ -41,6 +41,29 @@ size_t CpuIdHashMap<IdType>::Init(IdArray ids, IdArray unique_ids) {
 }
 
 template <typename IdType>
+void CpuIdHashMap<IdType>::Map(IdArray ids,
+    IdType default_val, IdArray new_ids) const {
+  const IdType* ids_data = ids.Ptr<IdType>();
+  const size_t len = static_cast<size_t>(ids->shape[0]);
+  IdType* values_data = new_ids.Ptr<IdType>();
+
+  parallel_for(0, len, grain_size, [&](int64_t s, int64_t e) {
+    for (int64_t i = s; i < e; i++) {
+        values_data[i] = mapId(ids_data[i], default_val);
+    }
+  });
+}
+
+template <typename IdType>
+CpuIdHashMap<IdType>::~CpuIdHashMap() {
+  if (_hmap != nullptr) {
+      DGLContext ctx = DGLContext{kDGLCPU, 0};
+      auto device = DeviceAPI::Get(ctx);
+      device->FreeWorkspace(ctx, _hmap);
+  }
+}
+
+template <typename IdType>
 size_t CpuIdHashMap<IdType>::fillInIds(size_t num_ids,
     const IdType* ids_data, IdArray unique_ids) {
   // Use `int16_t` instead of `bool` here. As vector<bool> is an exception
@@ -84,29 +107,6 @@ size_t CpuIdHashMap<IdType>::fillInIds(size_t num_ids,
   });
 
   return block_offset[thread_num];
-}
-
-template <typename IdType>
-CpuIdHashMap<IdType>::~CpuIdHashMap() {
-  if (_hmap != nullptr) {
-      DGLContext ctx = DGLContext{kDGLCPU, 0};
-      auto device = DeviceAPI::Get(ctx);
-      device->FreeWorkspace(ctx, _hmap);
-  }
-}
-
-template <typename IdType>
-void CpuIdHashMap<IdType>::Map(IdArray ids,
-    IdType default_val, IdArray new_ids) const {
-  const IdType* ids_data = ids.Ptr<IdType>();
-  const size_t len = static_cast<size_t>(ids->shape[0]);
-  IdType* values_data = new_ids.Ptr<IdType>();
-
-  parallel_for(0, len, grain_size, [&](int64_t s, int64_t e) {
-    for (int64_t i = s; i < e; i++) {
-        values_data[i] = mapId(ids_data[i], default_val);
-    }
-  });
 }
 
 template <typename IdType>
--- a/src/array/cpu/cpu_id_hash_map.cc
+++ b/src/array/cpu/cpu_id_hash_map.cc
@@ -1,6 +1,6 @@
 /**
  *  Copyright (c) 2019 by Contributors
- * @file dgl/cpu_id_hash_map.cc
+ * @file array/cpu/cpu_id_hash_map.cc
  * @brief Class about CPU id hash map
  */
 
--- a/src/array/cpu/cpu_id_hash_map.cc
+++ b/src/array/cpu/cpu_id_hash_map.cc
@@ -113,11 +113,11 @@ template <typename IdType>
 IdType CpuIdHashMap<IdType>::mapId(IdType id, IdType default_val) const {
   IdType pos = (id & _mask);
   IdType delta = 1;
-  while (_hmap[pos].key != kEmptyKey && _hmap[pos].key != id) {
+  while (_hmap[pos].key != k_empty_key && _hmap[pos].key != id) {
       next(&pos, &delta);
   }
 
-  if (_hmap[pos].key == kEmptyKey) {
+  if (_hmap[pos].key == k_empty_key) {
       return default_val;
   } else {
       return _hmap[pos].value;
@@ -135,7 +135,7 @@ void CpuIdHashMap<IdType>::insert(IdType id,
     std::vector<int16_t>* valid,
     size_t index) {
   IdType pos = (id & _mask), delta = 1;
-  while (!attempt_insert_at(pos, id, valid, index)) {
+  while (!attemptInsertAt(pos, id, valid, index)) {
       next(&pos, &delta);
   }
 }
@@ -151,10 +151,10 @@ void CpuIdHashMap<IdType>::set(IdType key, IdType value) {
 }
 
 template <typename IdType>
-bool CpuIdHashMap<IdType>::attempt_insert_at(int64_t pos, IdType key,
+bool CpuIdHashMap<IdType>::attemptInsertAt(int64_t pos, IdType key,
     std::vector<int16_t>* valid, size_t index) {
-  IdType empty_key = kEmptyKey;
-  IdType old_val = kEmptyKey;
+  IdType empty_key = k_empty_key;
+  IdType old_val = k_empty_key;
   COMPARE_AND_SWAP(&(_hmap[pos].key), empty_key, key, &old_val);
 
   if (old_val == empty_key) {
--- a/src/array/cpu/cpu_id_hash_map.cc
+++ b/src/array/cpu/cpu_id_hash_map.cc
@@ -15,6 +15,12 @@
 
 using namespace dgl::runtime;
 
+namespace {
+  static constexpr int64_t kEmptyKey = -1;
+  static constexpr int kGrainSize = 256;
+}  // namespace
+
+
 namespace dgl {
 namespace aten {
 
@@ -140,7 +146,8 @@ template <typename IdType>
 IdType CpuIdHashMap<IdType>::MapId(IdType id) const {
   IdType pos = (id & mask_);
   IdType delta = 1;
-  while (hmap_[pos].key != kEmptyKey && hmap_[pos].key != id) {
+  IdType empty_key = static_cast<IdType>(kEmptyKey);
+  while (hmap_[pos].key != empty_key && hmap_[pos].key != id) {
     Next(&pos, &delta);
   }
   return hmap_[pos].value;
@@ -170,7 +177,7 @@ void CpuIdHashMap<IdType>::Set(IdType key, IdType value) {
 template <typename IdType>
 bool CpuIdHashMap<IdType>::AttemptInsertAt(
     int64_t pos, IdType key, std::vector<int16_t>* valid, size_t index) {
-  IdType empty_key = kEmptyKey;
+  IdType empty_key = static_cast<IdType>(kEmptyKey);;
   IdType old_val = CompareAndSwap(&(hmap_[pos].key), empty_key, key);
 
   if (old_val == empty_key) {
--- a/src/array/cpu/cpu_id_hash_map.cc
+++ b/src/array/cpu/cpu_id_hash_map.cc
@@ -55,7 +55,7 @@ size_t CpuIdHashMap<IdType>::fillInIds(size_t num_ids,
   parallel_for(0, num_ids, grain_size, [&](int64_t s, int64_t e) {
       size_t count = 0;
       for (int64_t i = s; i < e; i++) {
-          insert_cas(ids_data[i], &valid, i);
+          insert(ids_data[i], &valid, i);
           if (valid[i]) {
               count++;
           }
@@ -77,7 +77,7 @@ size_t CpuIdHashMap<IdType>::fillInIds(size_t num_ids,
       for (int64_t i = s; i < e; i++) {
           if (valid[i]) {
               unique_ids_data[pos] = ids_data[i];
-              set_value(ids_data[i], pos);
+              set(ids_data[i], pos);
               pos = pos + 1;
           }
       }
@@ -104,13 +104,13 @@ void CpuIdHashMap<IdType>::Map(IdArray ids,
 
   parallel_for(0, len, grain_size, [&](int64_t s, int64_t e) {
     for (int64_t i = s; i < e; i++) {
-        values_data[i] = map(ids_data[i], default_val);
+        values_data[i] = mapId(ids_data[i], default_val);
     }
   });
 }
 
 template <typename IdType>
-IdType CpuIdHashMap<IdType>::map(IdType id, IdType default_val) const {
+IdType CpuIdHashMap<IdType>::mapId(IdType id, IdType default_val) const {
   IdType pos = (id & _mask);
   IdType delta = 1;
   while (_hmap[pos].key != kEmptyKey && _hmap[pos].key != id) {
@@ -131,7 +131,7 @@ void CpuIdHashMap<IdType>::next(IdType* pos, IdType* delta) const {
 }
 
 template <typename IdType>
-void CpuIdHashMap<IdType>::insert_cas(IdType id,
+void CpuIdHashMap<IdType>::insert(IdType id,
     std::vector<int16_t>* valid,
     size_t index) {
   IdType pos = (id & _mask), delta = 1;
@@ -141,13 +141,13 @@ void CpuIdHashMap<IdType>::insert_cas(IdType id,
 }
 
 template <typename IdType>
-void CpuIdHashMap<IdType>::set_value(IdType k, IdType v) {
-  IdType pos = (k & _mask), delta = 1;
-  while (_hmap[pos].key != k) {
+void CpuIdHashMap<IdType>::set(IdType key, IdType value) {
+  IdType pos = (key & _mask), delta = 1;
+  while (_hmap[pos].key != key) {
       next(&pos, &delta);
   }
 
-  _hmap[pos].value = v;
+  _hmap[pos].value = value;
 }
 
 template <typename IdType>
@@ -155,7 +155,7 @@ bool CpuIdHashMap<IdType>::attempt_insert_at(int64_t pos, IdType key,
     std::vector<int16_t>* valid, size_t index) {
   IdType empty_key = kEmptyKey;
   IdType old_val = kEmptyKey;
-  CAS(&(_hmap[pos].key), empty_key, key, &old_val);
+  COMPARE_AND_SWAP(&(_hmap[pos].key), empty_key, key, &old_val);
 
   if (old_val == empty_key) {
       (*valid)[index] = true;
--- a/src/array/cpu/cpu_id_hash_map.cc
+++ b/src/array/cpu/cpu_id_hash_map.cc
@@ -17,7 +17,7 @@ namespace dgl {
 namespace aten {
 
 template <typename IdType>
-CpuIdHashMap<IdType>::CpuIdHashMap(DGLContext ctx) : _hmap(nullptr), _ctx(ctx) {
+CpuIdHashMap<IdType>::CpuIdHashMap() : _hmap(nullptr) {
 }
 
 template <typename IdType>
@@ -30,9 +30,10 @@ size_t CpuIdHashMap<IdType>::Init(IdArray ids, IdArray unique_ids) {
     size_t capcacity = 1 << static_cast<size_t>(1 + std::log2(num * 3));
     _mask =  static_cast<IdType>(capcacity - 1);
 
-    auto device = DeviceAPI::Get(_ctx);
+    DGLContext ctx = DGLContext{kDGLCPU, 0};
+    auto device = DeviceAPI::Get(ctx);
     _hmap = static_cast<Mapping*>(
-        device->AllocWorkspace(_ctx, sizeof(Mapping) * capcacity));
+        device->AllocWorkspace(ctx, sizeof(Mapping) * capcacity));
     memset(_hmap, -1, sizeof(Mapping) * capcacity);
 
     return fillInIds(num, ids_data, unique_ids);
@@ -86,8 +87,9 @@ size_t CpuIdHashMap<IdType>::fillInIds(size_t num_ids, const IdType* ids_data, I
 template <typename IdType>
 CpuIdHashMap<IdType>::~CpuIdHashMap() {
     if (_hmap != nullptr) {
-        auto device = DeviceAPI::Get(_ctx);
-        device->FreeWorkspace(_ctx, _hmap);
+        DGLContext ctx = DGLContext{kDGLCPU, 0};
+        auto device = DeviceAPI::Get(ctx);
+        device->FreeWorkspace(ctx, _hmap);
     }
 }
 
--- a/src/array/cpu/cpu_id_hash_map.cc
+++ b/src/array/cpu/cpu_id_hash_map.cc
@@ -6,10 +6,12 @@
 
 #include "cpu_id_hash_map.h"
 
-#include <dgl/runtime/parallel_for.h>
+#include <dgl/array.h>
 #include <dgl/runtime/device_api.h>
+#include <dgl/runtime/parallel_for.h>
 
 #include <cmath>
+#include <numeric>
 
 using namespace dgl::runtime;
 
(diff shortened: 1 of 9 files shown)


tests/cpp/test_cpu_id_hash_map.cc
```diff
@@ -0,0 +1,60 @@
...
+    std::set<IdType> id_set(id_vec.begin(), id_vec.end());
+    IdArray ids = VecToIdArray(id_vec, sizeof(IdType) * 8, CTX);
+    IdArray unique_ids = NewIdArray(size, CTX, sizeof(IdType) * 8);
+    CpuIdHashMap<IdType> id_map(CTX);
+    auto unique_num = id_map.Init(ids, unique_ids);
+    unique_ids -> shape[0] = unique_num;
+    IdType* unique_id_data = unique_ids.Ptr<IdType>();
+    EXPECT_EQ(id_set.size(), unique_num);
+
+    for (size_t i = 0; i < unique_num; i++) {
```
Rhett-Ying, line 37: it it possible to check in parallel which may benefit in large size map test?

tests/cpp/test_cpu_id_hash_map.cc
```diff
@@ -0,0 +1,60 @@
...
+        id_vec[i] = static_cast<IdType>(std::rand() % range);
+    }
+}
+
+template <typename IdType, size_t size, size_t range>
+void _TestIdMap() {
+    std::vector<IdType> id_vec;
+    ConstructRandomSet(size, range, id_vec);
+    std::set<IdType> id_set(id_vec.begin(), id_vec.end());
+    IdArray ids = VecToIdArray(id_vec, sizeof(IdType) * 8, CTX);
```
Rhett-Ying, line 29: what's `CTX`?

tests/cpp/test_cpu_id_hash_map.cc
peizhou001: sure, use parallel_for for acceleration.

tests/cpp/test_cpu_id_hash_map.cc
```diff
@@ -0,0 +1,60 @@
...
+    IdType* unique_id_data = unique_ids.Ptr<IdType>();
+    EXPECT_EQ(id_set.size(), unique_num);
+
+    for (size_t i = 0; i < unique_num; i++) {
+        EXPECT_TRUE(id_set.find(unique_id_data[i]) != id_set.end());
+    }
+    
+    IdArray new_ids = NewIdArray(unique_num, CTX, sizeof(IdType) * 8);
+    IdType default_val = -1;
+    id_map.Map(unique_ids, default_val, new_ids);
```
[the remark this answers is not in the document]
peizhou001: Actual only `Map` and `Init` will be used outside, change` FillInIds -> fillInIds` to avoid misunderstanding.

tests/cpp/test_cpu_id_hash_map.cc
```diff
@@ -0,0 +1,60 @@
...
+}
+
+template <typename IdType, size_t size, size_t range>
+void _TestIdMap() {
+    std::vector<IdType> id_vec;
+    ConstructRandomSet(size, range, id_vec);
+    std::set<IdType> id_set(id_vec.begin(), id_vec.end());
+    IdArray ids = VecToIdArray(id_vec, sizeof(IdType) * 8, CTX);
+    IdArray unique_ids = NewIdArray(size, CTX, sizeof(IdType) * 8);
+    CpuIdHashMap<IdType> id_map(CTX);
```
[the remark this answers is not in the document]
peizhou001: No, It is only used in CPU. But there is another member in `DGLContext` called `device_id`, I'm not sure if it can always set to 0. If so, the argument can be removed.

tests/cpp/test_cpu_id_hash_map.cc
peizhou001: It is a static variable `static constexpr DGLContext CTX = DGLContext{kDGLCPU, 0};` in `dgl/common.h`.

tests/cpp/test_cpu_id_hash_map.cc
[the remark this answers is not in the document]
Rhett-Ying: I think `device_id` is used for cuda device choose only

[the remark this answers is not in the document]
peizhou001: I see, let me remove this parameter

src/array/cpu/cpu_id_hash_map.h
```diff
@@ -0,0 +1,151 @@
...
+#ifndef DGL_ARRAY_CPU_CPU_ID_HASH_MAP_H_
+#define DGL_ARRAY_CPU_CPU_ID_HASH_MAP_H_
+
+#include <dgl/aten/types.h>
+
+#include <vector>
+
+#ifdef _MSC_VER
+#include <intrin.h>
+#define COMPARE_AND_SWAP(ptr, oldval, newval, ret)                         \
```
frozenbugs, line 16: Do we need it in .h? what about just move to cc file if only you need this MACRO?

src/array/cpu/cpu_id_hash_map.h
```diff
@@ -0,0 +1,151 @@
...
+        reinterpret_cast<LONG*>(ptr), newval, oldval);        \
+    } else if (sizeof(newval) == 64) {                        \
+      *ret = _InterlockedCompareExchange64(                   \
+        reinterpret_cast<LONGLONG*>(ptr), newval, oldval);    \
+    } else {                                                  \
+      LOG(FATAL) << "ID can only be int32 or int64";          \
+    }                                                         \
+  } while (0)
+#elif __GNUC__
+#if __GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 1)
```
frozenbugs, line 29: As @BarclayII suggest, our minimum supported GCC version is GCC 5, so this check is unnecessary.

src/array/cpu/cpu_id_hash_map.h
frozenbugs: If we move it to cc, we can make it a function instead of a MACRO.
```
#ifdef _MSC_VER
#include <intrin.h>
#else // _MSC_VER

template <typename IdType>
static IdType CompareAndSwap(IdType* ptr, IdType old, IdType new) {
#ifdef _MSC_VER
  if (sizeof(IdType) == 32) {
    return _InterlockedCompareExchange(...)
  } else if (sizeof(IdType) == 64) {
    ...
  } else {
    ...
  }
#elif __GNUC__  // _MSC_VER
  return __sync_val_compare_and_swap(...)
#else  // _MSC_VER
  #error "CompareAndSwap is not supported on this platform."
#endif  // _MSC_VER
}
```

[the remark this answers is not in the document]
peizhou001: fixed

tests/cpp/test_cpu_id_hash_map.cc
```diff
@@ -0,0 +1,61 @@
...
+    _TestIdMap<int64_t, 1, 10>();
+    _TestIdMap<int32_t, 1000,500000>();
+    _TestIdMap<int64_t, 1000, 500000>();
+    _TestIdMap<int32_t, 50000, 1000000>();
+    _TestIdMap<int64_t, 50000, 1000000>();
+    _TestIdMap<int32_t, 100000, 40000000>();
+    _TestIdMap<int64_t, 100000, 40000000>();
+}
+
+}; // namespace
```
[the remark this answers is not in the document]
peizhou001: fixed

src/array/cpu/cpu_id_hash_map.h
```diff
@@ -0,0 +1,151 @@
...
+#error "requires GCC 4.1 or greater"
+#endif
+#define COMPARE_AND_SWAP(ptr, oldval, newval, ret) \
+  *ret = __sync_val_compare_and_swap(ptr, oldval, newval)
+#else
+#error "CAS not supported on this platform"
+#endif
```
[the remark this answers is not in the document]
peizhou001: fixed

src/array/cpu/cpu_id_hash_map.h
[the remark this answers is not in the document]
peizhou001: use old_val and new_val as new is a keyword.

src/array/cpu/cpu_id_hash_map.h
peizhou001: remove for fix

src/array/cpu/cpu_id_hash_map.h
peizhou001: Good suggestion, changed to this style.

src/array/cpu/cpu_id_hash_map.cc
```diff
@@ -0,0 +1,194 @@
...
+
+template <typename IdType>
+size_t CpuIdHashMap<IdType>::Init(IdArray ids, IdArray unique_ids) {
+  const IdType* ids_data = ids.Ptr<IdType>();
+  const size_t num = ids->shape[0];
+
+  CHECK_GT(num, 0);
+
+  size_t capcacity = 1;
+  capcacity = capcacity << static_cast<size_t>(1 + std::log2(num * 3));
```
Rhett-Ying, line 51: `std::ceil(std::log2(num * 3))`?

Rhett-Ying: and merge line 50, 51 into single line? like `const size_t capacity = 0x1 << ...`?

src/array/cpu/cpu_id_hash_map.cc
```diff
@@ -0,0 +1,194 @@
...
+      DGLContext ctx = DGLContext{kDGLCPU, 0};
+      auto device = DeviceAPI::Get(ctx);
+      device->FreeWorkspace(ctx, _hmap);
+  }
+}
+
+template <typename IdType>
+size_t CpuIdHashMap<IdType>::fillInIds(size_t num_ids,
+    const IdType* ids_data, IdArray unique_ids) {
+  // Use `int16_t` instead of `bool` here. As vector<bool> is an exception
```
Rhett-Ying, line 89: why choose 16 bits instead of 8 bits type?

src/array/cpu/cpu_id_hash_map.cc
```diff
@@ -0,0 +1,194 @@
...
+void CpuIdHashMap<IdType>::next(IdType* pos, IdType* delta) const {
+  *pos = (*pos + (*delta) * (*delta)) & _mask;
+  *delta = *delta + 1;
+}
+
+template <typename IdType>
+void CpuIdHashMap<IdType>::insert(IdType id,
+    std::vector<int16_t>* valid,
+    size_t index) {
+  IdType pos = (id & _mask), delta = 1;
```
Rhett-Ying, line 157: split into separate lines?

src/array/cpu/cpu_id_hash_map.cc
```diff
@@ -0,0 +1,194 @@
...
+
+template <typename IdType>
+void CpuIdHashMap<IdType>::next(IdType* pos, IdType* delta) const {
+  *pos = (*pos + (*delta) * (*delta)) & _mask;
+  *delta = *delta + 1;
+}
+
+template <typename IdType>
+void CpuIdHashMap<IdType>::insert(IdType id,
+    std::vector<int16_t>* valid,
```
Rhett-Ying, line 155: any reason of using pointer instead of reference which is more intuitive?

src/array/cpu/cpu_id_hash_map.cc
```diff
@@ -0,0 +1,194 @@
...
+    size_t index) {
+  IdType pos = (id & _mask), delta = 1;
+  while (!attemptInsertAt(pos, id, valid, index)) {
+      next(&pos, &delta);
+  }
+}
+
+template <typename IdType>
+void CpuIdHashMap<IdType>::set(IdType key, IdType value) {
+  IdType pos = (key & _mask), delta = 1;
```
Rhett-Ying, line 165: separate lines

src/array/cpu/cpu_id_hash_map.cc
```diff
@@ -0,0 +1,194 @@
...
+
+  if (_hmap[pos].key == k_empty_key) {
+      return default_val;
+  } else {
+      return _hmap[pos].value;
+  }
+}
+
+template <typename IdType>
+void CpuIdHashMap<IdType>::next(IdType* pos, IdType* delta) const {
```
Rhett-Ying, line 148: why not prefer reference?

src/array/cpu/cpu_id_hash_map.cc
```diff
@@ -0,0 +1,194 @@
...
+  *pos = (*pos + (*delta) * (*delta)) & _mask;
```
Rhett-Ying, line 149: add a few comments beside the code to clarify the intention of this line?

src/array/cpu/cpu_id_hash_map.cc
```diff
@@ -0,0 +1,194 @@
...
+    return _InterlockedCompareExchange(reinterpret_cast<LONG*>(ptr),
+      new_val, old_val);
+  } else if (sizeof(IdType) == 8) {
+    return _InterlockedCompareExchange64(reinterpret_cast<LONGLONG*>(ptr),
+      new_val, old_val);
+  } else {
+    LOG(FATAL) << "ID can only be int32 or int64";
+  }
+  #elif __GNUC__   // _MSC_VER
+    return __sync_val_compare_and_swap(ptr, old_val, new_val);
```
Rhett-Ying, line 33: why turn to these built-in CAS instead of previous used 1std:;atomic<>::compare_exchange_weak1 ?

src/array/cpu/cpu_id_hash_map.h
```diff
@@ -0,0 +1,147 @@
...
+
+  void insert(IdType id, std::vector<int16_t>* valid, size_t index);
+
+  void set(IdType key, IdType value);
+
+  bool attemptInsertAt(int64_t pos, IdType key,
+    std::vector<int16_t>* valid, size_t index);
+
+ private:
+  static constexpr IdType k_empty_key = static_cast<IdType>(-1);
```
Rhett-Ying, line 137: name it in capital? `K_XXX`?

src/array/cpu/cpu_id_hash_map.cc
```diff
@@ -0,0 +1,194 @@
...
+
+using namespace dgl::runtime;
+
+namespace dgl {
+namespace aten {
+
+template <typename IdType>
+IdType CpuIdHashMap<IdType>::CompareAndSwap(IdType* ptr,
+    IdType old_val, IdType new_val) {
+  #ifdef _MSC_VER
```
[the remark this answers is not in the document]
peizhou001: Fixed

src/array/cpu/cpu_id_hash_map.h
```diff
@@ -0,0 +1,147 @@
...
+   * And then insert the items in `ids` concurrently to generate the
+   * mappings, in passing returning the unique ids in `ids`. 
+   *
+   * @param ids The array of ids to be inserted as keys.
+   *
+   * @param unique_ids An id array for storing unique items in `ids`.
+   * 
+   * @return Number of unique items in input `ids`.
+   */
+  size_t Init(IdArray ids, IdArray unique_ids);
```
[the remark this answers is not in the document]
peizhou001: fixed

src/array/cpu/cpu_id_hash_map.h
```diff
@@ -0,0 +1,147 @@
+/**
+ *  Copyright (c) 2019 by Contributors
```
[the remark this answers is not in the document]
peizhou001: changed to latest

src/array/cpu/cpu_id_hash_map.h
```diff
@@ -0,0 +1,147 @@
...
+   * 
+   * @param new_ids Array for storing results.
+   *
+   */
+  void Map(IdArray ids, IdType default_val, IdArray new_ids) const;
+
+  ~CpuIdHashMap();
+
+ private:
+  IdType mapId(IdType id, IdType default_val) const;
```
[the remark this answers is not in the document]
peizhou001: CamelCase used, has changed all non-compatible code to that format.

src/array/cpu/cpu_id_hash_map.h
```diff
@@ -0,0 +1,147 @@
...
+
+  size_t fillInIds(size_t num_ids,
```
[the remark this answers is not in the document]
peizhou001: `IdArray` itself is pointer-like, so add a & is not needed.

src/array/cpu/cpu_id_hash_map.h
```diff
@@ -0,0 +1,147 @@
...
+ * R should be: 
+ * [0, 2, 5]
+**/
+template <typename IdType>
+class CpuIdHashMap {
+ public:
+  /**
+   * @brief An entry in the hashtable.
+   */
+  struct Mapping {
```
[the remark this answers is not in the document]
peizhou001: fixed

src/array/cpu/cpu_id_hash_map.cc
peizhou001: fixed

src/array/cpu/cpu_id_hash_map.cc
peizhou001: It is conflict with our rule. Lint check will return warning/error.

src/array/cpu/cpu_id_hash_map.cc
peizhou001: added

src/array/cpu/cpu_id_hash_map.cc
peizhou001: Atomic is redundant and time-cousuming to init.

peizhou001: > Atomic is redundant and time-consuming to init.

src/array/cpu/cpu_id_hash_map.h
peizhou001: fixed

src/array/cpu/cpu_id_hash_map.h
```diff
@@ -0,0 +1,147 @@
...
+ *   {key: 101, value: 4},
+ *   {key: 102, value: 5}
+ * ]
+ * Search the hashmap with array I=[98, 99, 102]:
+ * H.Map(I, -1, R) (3)
+ * R should be: 
+ * [0, 2, 5]
+**/
+template <typename IdType>
+class CpuIdHashMap {
```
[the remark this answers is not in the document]
peizhou001: Agree, because it already under the `CPU` folder. Let me remove it.

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,195 @@
...
+  const IdType* ids_data = ids.Ptr<IdType>();
+  const size_t num = static_cast<size_t>(ids->shape[0]);
+  CHECK_GT(num, 0);
+  size_t capcacity = 1;
+  capcacity = capcacity << static_cast<size_t>(1 + std::log2(num * 3));
+  mask_ = static_cast<IdType>(capcacity - 1);
+
+  DGLContext ctx = DGLContext{kDGLCPU, 0};
+  auto device = DeviceAPI::Get(ctx);
+  hmap_ = static_cast<Mapping*>(
```
frozenbugs, line 61: In general, i would prefer to have less memory ops, esp. having an exposed pointer to handle memory allocation and release. It might be error-prone and hard to debug in the future.

Consider using:
- NDArray
- Vector
- unique_ptr

src/array/cpu/cpu_id_hash_map.cc
Rhett-Ying: so only const inference is allowed? If so, probably we need to disable this rule in some way as reference is more intuitive. `*a = *a + *b ....` looks ugly. or make it as lambda or something else to escape from lint check?

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,195 @@
+/**
+ *  Copyright (c) latest by Contributors
```
Rhett-Ying, line 2: `latest` or `2023`? I don't know about it.

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,195 @@
...
+  hmap_ = static_cast<Mapping*>(
+      device->AllocWorkspace(ctx, sizeof(Mapping) * capcacity));
+  memset(hmap_, -1, sizeof(Mapping) * capcacity);
+
+  IdArray unique_ids = NewIdArray(num, ctx, sizeof(IdType) * 8);
+  return FillInIds(num, ids_data, unique_ids);
+}
+
+template <typename IdType>
+IdArray IdHashMap<IdType>::Map(const IdArray ids) const {
```
frozenbugs, line 70: Map -> MapIds

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,195 @@
...
+IdHashMap<IdType>::IdHashMap() : hmap_(nullptr), mask_(0) {}
+
+template <typename IdType>
+IdArray IdHashMap<IdType>::Init(const IdArray ids) {
+  CHECK_EQ(ids.defined(), true);
+  const IdType* ids_data = ids.Ptr<IdType>();
+  const size_t num = static_cast<size_t>(ids->shape[0]);
+  CHECK_GT(num, 0);
+  size_t capcacity = 1;
+  capcacity = capcacity << static_cast<size_t>(1 + std::log2(num * 3));
```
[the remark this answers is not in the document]
Rhett-Ying: how about preferring function/variable(probably constexpr) than MACRO?

src/array/cpu/id_hash_map.h
```diff
@@ -0,0 +1,145 @@
...
+ */
+
+#ifndef DGL_ARRAY_CPU_ID_HASH_MAP_H_
+#define DGL_ARRAY_CPU_ID_HASH_MAP_H_
+
+#include <dgl/aten/types.h>
+
+#include <vector>
+
+#ifdef _MSC_VER
```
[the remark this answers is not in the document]
peizhou001: moved to cc for fix

src/array/cpu/id_hash_map.h
```diff
@@ -0,0 +1,145 @@
...
+   * @warning Key must exist.
+   */
+  void Set(IdType key, IdType value);
+
+  bool AttemptInsertAt(
+      int64_t pos, IdType key, std::vector<int16_t>* valid, size_t index);
+
+ private:
+  /**
+   * @brief Array used to save all elelemnts in the hash table.
```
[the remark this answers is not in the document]
peizhou001: fixed

src/array/cpu/id_hash_map.h
```diff
@@ -0,0 +1,145 @@
...
+   */
+  Mapping* hmap_;
+  /**
+   * @brief Mask assisted to get the position for a key.
```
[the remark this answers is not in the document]
peizhou001: fixed

src/array/cpu/id_hash_map.h
```diff
@@ -0,0 +1,145 @@
...
+  void Set(IdType key, IdType value);
+
+  bool AttemptInsertAt(
+      int64_t pos, IdType key, std::vector<int16_t>* valid, size_t index);
+
+ private:
+  /**
+   * @brief Array used to save all elelemnts in the hash table.
+   */
+  Mapping* hmap_;
```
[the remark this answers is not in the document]
peizhou001: changed for fix

src/array/cpu/id_hash_map.h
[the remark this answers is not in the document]
peizhou001: changed

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,195 @@
...
+  mask_ = static_cast<IdType>(capcacity - 1);
+
+  DGLContext ctx = DGLContext{kDGLCPU, 0};
+  auto device = DeviceAPI::Get(ctx);
+  hmap_ = static_cast<Mapping*>(
+      device->AllocWorkspace(ctx, sizeof(Mapping) * capcacity));
+  memset(hmap_, -1, sizeof(Mapping) * capcacity);
+
+  IdArray unique_ids = NewIdArray(num, ctx, sizeof(IdType) * 8);
+  return FillInIds(num, ids_data, unique_ids);
```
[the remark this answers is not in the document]
peizhou001: moved for fix

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,195 @@
...
+
+template <typename IdType>
+IdHashMap<IdType>::IdHashMap() : hmap_(nullptr), mask_(0) {}
+
+template <typename IdType>
+IdArray IdHashMap<IdType>::Init(const IdArray ids) {
+  CHECK_EQ(ids.defined(), true);
+  const IdType* ids_data = ids.Ptr<IdType>();
+  const size_t num = static_cast<size_t>(ids->shape[0]);
+  CHECK_GT(num, 0);
```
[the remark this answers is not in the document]
peizhou001: added

src/array/cpu/id_hash_map.cc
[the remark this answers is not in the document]
peizhou001: Constexpr cannot be used here as `num` is uncertain at compiler-time.  Use an inline function for this.

src/array/cpu/id_hash_map.cc
peizhou001: 2023 looks more compatible

src/array/cpu/id_hash_map.cc
peizhou001: changed

src/array/cpu/id_hash_map.cc
peizhou001: Use `unique_ptr`

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,201 @@
...
+  };
+  hash_map_ = {nullptr, deleter};
+}
+
+template <typename IdType>
+IdArray IdHashMap<IdType>::Init(const IdArray ids) {
+  CHECK_EQ(ids.defined(), true);
+  const IdType* ids_data = ids.Ptr<IdType>();
+  const size_t num_ids = static_cast<size_t>(ids->shape[0]);
+  // Make sure `ids` is not 0 dim.
```
frozenbugs, line 75: What will happen if the ids has 2 dim?

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,201 @@
...
+  CHECK_GT(num_ids, 0);
+  size_t capcacity = GetMapSize(num_ids);
```
frozenbugs, line 77: typo: capacity

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,201 @@
...
+    }
+  });
+  return unique_ids;
+}
+
+template <typename IdType>
+IdArray IdHashMap<IdType>::MapIds(const IdArray ids) const {
+  CHECK_EQ(ids.defined(), true);
+  const IdType* ids_data = ids.Ptr<IdType>();
+  const size_t len = static_cast<size_t>(ids->shape[0]);
```
frozenbugs, line 129: Since num_ids was used in L74, so let's also use num_ids here.

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,201 @@
...
+
+  parallel_for(0, len, kGrainSize, [&](int64_t s, int64_t e) {
+    for (int64_t i = s; i < e; i++) {
+      values_data[i] = MapId(ids_data[i]);
+    }
+  });
+  return new_ids;
+}
+
+template <typename IdType>
```
frozenbugs, line 144: not sure we need to explicitly code this, remove if it is unnecessary to be added.

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,201 @@
...
+        reinterpret_cast<LONG*>(ptr), new_val, old_val);
+  } else if (sizeof(IdType) == 8) {
+    return _InterlockedCompareExchange64(
+        reinterpret_cast<LONGLONG*>(ptr), new_val, old_val);
+  } else {
+    LOG(FATAL) << "ID can only be int32 or int64";
+  }
+#elif __GNUC__  // _MSC_VER
+  return __sync_val_compare_and_swap(ptr, old_val, new_val);
+#else           // _MSC_VER
```
BarclayII, line 51: Does that mean compare and swap is not supported on Mac (clang)?

tests/cpp/test_id_hash_map.cc
```diff
@@ -0,0 +1,59 @@
...
+  IdArray new_ids = id_map.MapIds(unique_ids);
+  EXPECT_TRUE(new_ids.IsContiguous());
+}
+
+TEST(IdHashMapTest, TestIdHashMap) {
+  _TestIdMap<int32_t, 1, 10>();
+  _TestIdMap<int64_t, 1, 10>();
+  _TestIdMap<int32_t, 1000, 500000>();
+  _TestIdMap<int64_t, 1000, 500000>();
+  _TestIdMap<int32_t, 50000, 1000000>();
```
frozenbugs, line 53: Which makes it a good test case by increasing the size of the IDArray?

src/array/cpu/id_hash_map.cc
peizhou001: No，it is also supported in mac. And I have did some tests in mac and it works well.

src/array/cpu/id_hash_map.cc
peizhou001: It should crash. But as the input has been identified as an IdArray, so in general we only check if it is empty(0-dim)

src/array/cpu/id_hash_map.cc
peizhou001: fixed

src/array/cpu/id_hash_map.cc
peizhou001: changed

tests/cpp/test_id_hash_map.cc
peizhou001: When data size increase, multi thread functions are more vulnerable to mistakes. So add data column to test to test its robust.

src/array/cpu/id_hash_map.cc
peizhou001: removed

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,198 @@
...
+      Insert(ids_data[i], &valid, i);
+      count += valid[i];
+    }
+    block_offset[omp_get_thread_num() + 1] = count;
+  });
+
+  // Get ExclusiveSum of each block.
+  std::partial_sum(
+      block_offset.begin() + 1, block_offset.end(), block_offset.begin() + 1);
+  unique_ids->shape[0] = block_offset.back();
```
BarclayII, line 108: Is assigning shape values OK to do?  I feel it's very dangerous.  It's safer to compute the shape first and then allocate the `unique_ids` array.

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,198 @@
...
+    for (int64_t i = s; i < e; i++) {
+      values_data[i] = MapId(ids_data[i]);
+    }
+  });
+  return new_ids;
+}
+
+template <typename IdType>
+inline void IdHashMap<IdType>::Next(IdType* pos, IdType* delta) const {
+  // Use Quadric probing.
```
BarclayII, line 146: "Quadric" -> "quadratic"?

src/array/cpu/id_hash_map.h
```diff
@@ -0,0 +1,141 @@
...
+   * location with a given value and, only if they are the same, modifies
+   * the contents of that memory location to a new given value.
+   *
+   * @param ptr The pointer to the object to test and modify .
+   * @param old_val The value expected to be found in `ptr`.
+   * @param new_val The value to store in `ptr` if it is as expected.
+   *
+   * @return Old value pointed by the `ptr`.
+   */
+  static IdType CompareAndSwap(IdType* ptr, IdType old_val, IdType new_val);
```
BarclayII, line 82: I don't think this is the right place for this function since CAS can be applicable to other components as well.  Maybe move it to an inline function in `dgl::runtime` namespace.

src/array/cpu/id_hash_map.h
```diff
@@ -0,0 +1,141 @@
...
+   * @brief Find the mappings of given keys.
+   *
+   * @param ids The keys to map for.
+   *
+   * @return Mapping results corresponding to `ids`.
+   */
+  IdArray MapIds(const IdArray ids) const;
+
+ private:
+  void Next(IdType* pos, IdType* delta) const;
```
BarclayII, line 111: Need brief explanation on the private methods as well. (Same for other occurrences)
Here we also need to explain what `pos` and `delta` means as they are in/out parameters.
```
@param[in,out] pos ...
```

src/array/cpu/id_hash_map.h
```diff
@@ -0,0 +1,141 @@
...
+
+  IdType MapId(const IdType id) const;
+
+  void Insert(IdType id, std::vector<int16_t>* valid, size_t index);
```
BarclayII, line 115: Need to explain what is `valid`.  Also why a pointer to vector rather than `int16_t *`?  I don't think the size of the vector will change or anything.

src/array/cpu/id_hash_map.h
```diff
@@ -0,0 +1,141 @@
...
+#include <vector>
+
+namespace dgl {
+namespace aten {
+
+/**
+ * @brief A CPU targeted hashmap for mapping duplicate and non-consecutive ids
+ * in the provided array to unique and consecutive ones. It utilizes
+ * multi-threading to accelerate the insert and search speed. Currently it is
+ * only designed to be used in `ToBlockCpu` for optimizing.
```
BarclayII, line 23:
```
... only designed to be used to optimize `ToBlockCpu`, so it does
not support key deletion.
```
You can add other limitations as well.

src/array/cpu/id_hash_map.h
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * The hashmap should be used in two phases. With the first being creating the
+ * hashmap, and then init it with an id array. After that, searching any old ids
+ * to get the mappings according to your need.
```
BarclayII, line 27: I don't quite understand this explanation.  Looks like it has three phases (creation, init, and search).

Do you want to say that querying before inserting all elements is not supported?  If so you can just say that in the limitation above.

BarclayII (approved):
Minor suggestions.

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,201 @@
...
+  // when the pointer is freed.
+  auto deleter = [](Mapping* mappings) {
+    if (mappings != nullptr) {
+      DGLContext ctx = DGLContext{kDGLCPU, 0};
+      auto device = DeviceAPI::Get(ctx);
+      device->FreeWorkspace(ctx, mappings);
+    }
+  };
+  hash_map_ = {nullptr, deleter};
+}
```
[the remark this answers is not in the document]
peizhou001: We have a discussion before, the benefit of using a uniform memory allocator are listed below: 
1. It is more friendly for memory reuse. 
2. One allocator may ask a large volume amount from OS, if another allocator(e.g STL) also ask a large amount. It could cause either memory waste or frequently system call. Or in some corner cases, the second request may fail because of OOM.

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,198 @@
...
+
+template <typename IdType>
+IdArray IdHashMap<IdType>::Init(const IdArray ids) {
```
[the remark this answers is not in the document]
peizhou001: FIxed

src/array/cpu/id_hash_map.cc
```diff
@@ -0,0 +1,198 @@
...
+  hash_map_.reset(static_cast<Mapping*>(
+      DeviceAPI::Get(ctx)->AllocWorkspace(ctx, sizeof(Mapping) * capacity)));
+  memset(hash_map_.get(), -1, sizeof(Mapping) * capacity);
+
+  // This code block is to fill the ids into hash_map_.
+  IdArray unique_ids = NewIdArray(num_ids, ctx, sizeof(IdType) * 8);
+  // Use `int16_t` instead of `bool`. As vector<bool> is an exception
+  // for whom updating different elements from different threads is unsafe.
+  // see https://en.cppreference.com/w/cpp/container#Thread_safety.
+  std::vector<int16_t> valid(num_ids);
```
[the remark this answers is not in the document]
peizhou001: It make sense, use IdArray instead.

src/array/cpu/id_hash_map.cc
peizhou001: It looks safe because I find some other usages like this and the memory will not leaked. We can not allocate ahead because it is known after the data is filled in.

src/array/cpu/id_hash_map.h
peizhou001: It is a public static method so it can be reused in other modulars.

src/array/cpu/id_hash_map.h
peizhou001: Added

src/array/cpu/id_hash_map.h
```diff
@@ -0,0 +1,141 @@
...
+  void Set(IdType key, IdType value);
+
+  bool AttemptInsertAt(
+      int64_t pos, IdType key, std::vector<int16_t>* valid, size_t index);
+
+ private:
+  /**
+   * @brief Hash maps which is used to store all elements.
+   */
+  std::unique_ptr<Mapping[], std::function<void(Mapping*)>> hash_map_;
```
Rhett-Ying, line 129: why choose `Mapping[]` instead of `Mapping*`? what's the benefit of using smart pointer instead of raw pointer?

src/array/cpu/id_hash_map.h
peizhou001: added

src/array/cpu/id_hash_map.h
peizhou001: changed for remove misunderstanding.

src/array/cpu/id_hash_map.h
peizhou001: It is a specialization of `unique_ptr`,  `[]` is only supported with `T[]`.

src/array/cpu/id_hash_map.h
peizhou001: Add note and use NdArray(The elements is bool, don't use naive pointer to avoid memory operation) instead.

src/array/cpu/id_hash_map.cc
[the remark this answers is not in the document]
peizhou001: Seems `BoolArray` is unsafe in multi-thread environment, change back to vector.

PR #5299, opened 2023-02-15
[bugfix] Fix assertions in /src/runtime/workspace.h and expand unit tests for sparse optimizer

nv-dlasalle:
## Description
The unit tests in /tests/pytorch/test_optim.py only tested the case where no gradients needed to be exchanged--as a result the communication buffers were of size zero and allocated to be `null`. This would cause the assertions to fail. This PR updates the assertions to allow for `null` pointers when the size is zero, and expands to the unit test to also check the case where gradients need to be exchanged.

## Checklist
- [x] Code is well-documented


yaox12:
LGTM.

yaox12 (approved):

PR #5304, opened 2023-02-16
[Enhancement] Change id hash map

peizhou001:
## Description
As the concurrent id hash map is mainly used in `to_block `to map an id array to a new contiguous one, and current solution doesn't ensure the mapping order.  While a requirement by it is to map first Nth unique seed nodes to 0~N. This PR change the id hash map to meet the requirement.

## Checklist
- [x] Code is well-documented


src/array/cpu/concurrent_id_hash_map.cc
```diff
@@ -69,33 +69,46 @@ IdHashMap<IdType>::IdHashMap() : mask_(0) {
...
   hash_map_.reset(static_cast<Mapping*>(
-      DeviceAPI::Get(ctx)->AllocWorkspace(ctx, sizeof(Mapping) * capacity)));
+      device->AllocWorkspace(ctx, sizeof(Mapping) * capacity)));
   memset(hash_map_.get(), -1, sizeof(Mapping) * capacity);
 
   // This code block is to fill the ids into hash_map_.
   IdArray unique_ids = NewIdArray(num_ids, ctx, sizeof(IdType) * 8);
+  IdType* unique_ids_data = unique_ids.Ptr<IdType>();
+  // Fill in the first `num_seeds` ids.
+  parallel_for(0, num_seeds, kGrainSize, [&](int64_t s, int64_t e) {
```
frozenbugs, line 93: What's the common scale of num_seeds?

src/array/cpu/concurrent_id_hash_map.cc
```diff
@@ -69,33 +69,46 @@ IdHashMap<IdType>::IdHashMap() : mask_(0) {
...
+    for (int64_t i = s; i < e; i++) {
+      InsertAndSet(ids_data[i], static_cast<IdType>(i));
```
frozenbugs, line 95: Why for seed ids we don't use AttemptInsertAt?

src/array/cpu/concurrent_id_hash_map.h
```diff
@@ -24,34 +24,36 @@ namespace aten {
...
  * The hash map should be prepared in two phases before using. With the first
- * being creating the hashmap, and then init it with an id array.
+ * being creating the hashmap, and then initialize it with an id array which is
+ * divided into 2 parts: [`seed ids`, `left ids`]. In result `seed ids` are
+ * mapped to [0, num_seed_ids) and `left ids` to [num_seed_ids, num_unique_ids).
+ * Notice that mapping order is stable for `seed ids` while not for the left.
  *
- * For example, for an array A with following entries:
- * [98, 98, 100, 99, 97, 99, 101, 100, 102]
+ * For example, for an array A having 4 seed ids with following entries:
```
frozenbugs, line 32: Can you clarify what are the seed ids?

src/array/cpu/concurrent_id_hash_map.h
```diff
@@ -81,25 +83,25 @@ class IdHashMap {
...
+  ConcurrentIdHashMap(const ConcurrentIdHashMap& other) = delete;
+  ConcurrentIdHashMap& operator=(const ConcurrentIdHashMap& other) = delete;
 
   /**
-   * @brief Init the hashmap with an array of ids.
-   * Firstly allocating the memeory and init the entire space with empty key.
-   * And then insert the items in `ids` concurrently to generate the
-   * mappings, in passing returning the unique ids in `ids`.
+   * @brief Initialize the hashmap with an array of ids. The first `num_seeds`
+   * ids are unqiue and must be mapped to a contiguous array starting
```
frozenbugs, line 93: unqiue -> unique

src/array/cpu/concurrent_id_hash_map.h
```diff
@@ -24,34 +24,36 @@ namespace aten {
...
- * being creating the hashmap, and then init it with an id array.
+ * being creating the hashmap, and then initialize it with an id array which is
+ * divided into 2 parts: [`seed ids`, `left ids`]. In result `seed ids` are
+ * mapped to [0, num_seed_ids) and `left ids` to [num_seed_ids, num_unique_ids).
+ * Notice that mapping order is stable for `seed ids` while not for the left.
  *
- * For example, for an array A with following entries:
- * [98, 98, 100, 99, 97, 99, 101, 100, 102]
+ * For example, for an array A having 4 seed ids with following entries:
+ * [99, 98, 100, 97, 97, 101, 101, 102, 101]
```
frozenbugs, line 33: In the comment below you mentioned `num_seeds` ids are unique, I am assuming the first 4 are seed ids, but I see duplicated 97 in this example, it this intended?

src/array/cpu/concurrent_id_hash_map.cc
peizhou001: The scale depends on your fan-out, which usually about 1/10 of the original nodes. When input nodes is huge, it could be also very large. And parallel doesn't introduce side effects here, so keep it here should be better.

src/array/cpu/concurrent_id_hash_map.cc
peizhou001: 1. `Seed ids` mapping value is exactly its index in the array, so the key and value need to be set at the same time.
2. `Seed ids` is unique so the insertion is simpler and some checks can be removed to save efforts.

src/array/cpu/concurrent_id_hash_map.h
peizhou001: Sure, added.

src/array/cpu/concurrent_id_hash_map.h
peizhou001: fixed

src/array/cpu/concurrent_id_hash_map.h
peizhou001: Yes. `Seed ids` is unique among themselves, but it can be duplicate with other ids. So put it here may help user clarify it.

src/array/cpu/concurrent_id_hash_map.h
```diff
@@ -25,18 +25,20 @@ namespace aten {
  *
  * The hash map should be prepared in two phases before using. With the first
  * being creating the hashmap, and then initialize it with an id array which is
- * divided into 2 parts: [`seed ids`, `left ids`]. In result `seed ids` are
- * mapped to [0, num_seed_ids) and `left ids` to [num_seed_ids, num_unique_ids).
- * Notice that mapping order is stable for `seed ids` while not for the left.
+ * divided into 2 parts: [`seed ids`, `left ids`]. `Seed ids` refer to
+ * a set ids chosen as the input for sampling process and `left ids` are the
```
frozenbugs, line 29: left ids -> sampled ids.

sampled ids are the ids new sampled from the process (note the the seed ids might be sampled in the process, but not included in the sampled ids to avoid duplication)

peizhou001: Good description, adopt it in the notes.

src/array/cpu/concurrent_id_hash_map.cc
```diff
@@ -111,7 +111,9 @@ IdArray ConcurrentIdHashMap<IdType>::Init(
   parallel_for(num_seeds, num_ids, kGrainSize, [&](int64_t s, int64_t e) {
     size_t count = 0;
     for (int64_t i = s; i < e; i++) {
-      Insert(ids_data[i], &valid, i);
+      if (Insert(ids_data[i])) {
```
frozenbugs, line 114: I am assuming each i will only be accessed once:
It can be simplified to:
valid[i] = Insert(ids_data[i]);
count += valid[i];

This is actually better since the existing code in L107 assumes the valid will be initiated to 0, which might not be true for all c++ compiler.

src/array/cpu/concurrent_id_hash_map.h
```diff
@@ -57,6 +59,18 @@ namespace aten {
 template <typename IdType>
 class ConcurrentIdHashMap {
  public:
+  /**
+   * @brief The result state of an attempt to insert.
+   */
+  enum class InsertState {
```
frozenbugs, line 65: move to private section?

frozenbugs (approved):

src/array/cpu/concurrent_id_hash_map.cc
peizhou001: It makes sense. changed to this style.

src/array/cpu/concurrent_id_hash_map.h
peizhou001: changed

PR #5310, opened 2023-02-16
[DistDGL][Feature_Request]Changes in the metadata.json file for input graph dataset.

kylasa:
* Remove NUM_NODES_PER_CHUNK
* Remove NUM_EDGES_PER_CHUNK
* Remove the dependency between no. of edge files per edge type and no. of partitions
* Remove the dependency between no. of edge feature files per edge type and no. of partitions
* Remove the dependency between no. of node feature files and no. of partitions
* Add “node_type_counts”. This will be a list of integers. Each integer will represent total count of a node-type. The index in this list and the index in the “node_type” will be the same for a given node-type.
* Add “edge_type_counts”. This will be a list of integers. Each integer will represent total count of an edge-type. The index in this list and the index in the “edge_type” list will be the same for a given edge-type.

**Please note that, we still depend on the no. of edge feature files to be same as the no. of edge files for all the edge types. This dependency will also be removed at some future date.**

## Checklist
- [ X] Code is well-documented


thvasilo:
Hello @kylasa since this change affects user-facing features (such as the expected format of `metadata.json`) it should include changes in documentation such as https://github.com/dmlc/dgl/blob/master/docs/source/guide/distributed-preprocessing.rst and other parts that may have been affected.

Before changing the format do we know if there are any customers or end users that rely on the existing format? If yes, are the changes backwards compatible?

kylasa:
@thvasilo, at present there is no existing clear workflow which defines a clear path from code changes (affecting the customer facing parameters - like this PR).

I will send an email, including you as well, to Steve/Da/Minjie to let us know who is responsible for the documentation present on the docs.dgl.ai website and all the other documentation about this produce and the pipeline. I guess once we have this well-defined that resource will responsible for changing all the customer facing documentation. Lets wait from the team and figure out the next steps. 

Currently there are *NO* active customers to this pipeline. To my limited knowledge only NetpuneML and GraphStorm are the teams who are trying to integrate it their respective pipielines. 

Because of the above reason, these changes made through this PR are not backward-compatible. After this change, metadata.json file is expected to have the new keys as required by this PR.

tests/tools/test_dist_part.py
```diff
@@ -222,6 +222,13 @@ def test_chunk_graph_vector_rows(num_chunks, vector_rows):
     )
 
 
+"""
```
Rhett-Ying, line 225: If this functionality is no longer supported, do we still to keep these code or just remove them at all as a clean-up?

tests/tools/utils.py
```diff
@@ -118,6 +118,9 @@ def _chunk_graph(
     metadata["graph_name"] = name
     metadata["node_type"] = g.ntypes
 
+    # add node_type_counts
+    metadata["node_type_counts"] = [g.num_nodes(ntype) for ntype in g.ntypes]
```
Rhett-Ying, line 122: the name looks not intuitive to me. seems indicates the count of node types instead of the num of nodes for each node type. could you use a more clarified name?

tests/tools/utils.py
```diff
@@ -134,6 +137,10 @@ def _chunk_graph(
         num_nodes_per_chunk.append(num_nodes_list)
 
     metadata["edge_type"] = [etypestrs[etype] for etype in g.canonical_etypes]
+    # add edge_type_counts here.
+    metadata["edge_type_counts"] = [
```
Rhett-Ying, line 141: change the field name?

tools/dispatch_data.py
```diff
@@ -44,16 +44,11 @@ def submit_jobs(args) -> str:
...
 
     # retrieve num_parts
-    num_chunks = len(schema_map["num_nodes_per_chunk"][0])
-    num_parts = num_chunks
+    num_parts = args.num_parts
     partition_path = os.path.join(args.partitions_dir, "partition_meta.json")
     if os.path.isfile(partition_path):
         part_meta = load_partition_meta(partition_path)
         num_parts = part_meta.num_parts
-    if num_parts > num_chunks:
```
Rhett-Ying, line 53: we support `num_parts` less than `num_chunks` now?

tools/dispatch_data.py
```diff
@@ -129,6 +124,12 @@ def main():
         default="info",
         help="To enable log level for debugging purposes. Available options: (Critical, Error, Warning, Info, Debug, Notset)",
     )
+    parser.add_argument(
```
Rhett-Ying, line 127: why do we need this? `num_parts` is supposed be able to inferred from partition assignments? what if the specified one does not match with partition assignment?

tools/distpartitioning/constants.py
```diff
@@ -20,9 +20,9 @@
 PART_LOCAL_NID = "part_local_nid"
 
 STR_NODE_TYPE = "node_type"
-STR_NUM_NODES_PER_CHUNK = "num_nodes_per_chunk"
+# STR_NUM_NODES_PER_CHUNK = "num_nodes_per_chunk"
```
Rhett-Ying, line 23: how about removing instead of commenting out?

tools/distpartitioning/data_shuffle.py
```diff
@@ -794,23 +814,25 @@ def read_dataset(rank, world_size, id_lookup, params, schema_map):
         world_size,
         params.num_parts,
         schema_map,
+        ntype_counts,
     )
     logging.info(f"[Rank: {rank}] Done reading dataset {params.input_dir}")
+    dist.barrier()  # SYNCH
```
Rhett-Ying, line 820: # SYNC?

tools/distpartitioning/data_shuffle.py
```diff
@@ -794,23 +814,25 @@ def read_dataset(rank, world_size, id_lookup, params, schema_map):
...
 
     edge_data = augment_edge_data(
         edge_data, id_lookup, edge_tids, rank, world_size, params.num_parts
     )
+    dist.barrier()  # SYNCH
```
Rhett-Ying, line 825: adding `dist.barrier()` before or after the calls looks a bit tedious. How about sync within the target call?

tests/tools/utils.py
kylasa: How about "num_nodes_per_type" ??

tests/tools/utils.py
kylasa: How about "num_edges_per_type" ??

tools/dispatch_data.py
kylasa: Yes. We support no. of chunks < no. of partitions now.

tools/dispatch_data.py
kylasa: With these changes to the input metadata.json file, There is no way to infer 'num_parts' from the input metadata.json schema. 

I think we can get this information from the partitions_meta.json file, in the partitions_dir. 
Code is changed accordingly.

tools/distpartitioning/constants.py
kylasa: Done.

tools/distpartitioning/data_shuffle.py
kylasa: This is put in for the following purpose. 

We now support num_chunks < num_partitions. This means that on some ranks it is quite possible that there is no data to read from (no edge files, no edge/node feature files). In this case, these ranks the disk IO functions (get_dataset) are completed in a jiffy and on other ranks a lot of time is spent in reading files from the disk. 

In these instances, if we want all the ranks to be in sync then we have to barrier() call here. 
Otherwise, debugging  can become really hard.

tools/distpartitioning/data_shuffle.py
kylasa: As I mentioned earlier, I did for debugging purposes. 

We can do away with these barrier calls, since messaging calls themselves act as barriers, by definition. 

Also, Please note that we are targeting massively large datasets. We want to debug these datasets in as much less time as possible. In these instances, Since we ensure all the ranks run the major steps of the pipeline in a lock-step fashion, the debugging can be easier. 

I will leave it upto you. Lets chat about this in the call later this week.

tools/distpartitioning/dataset_utils.py
```diff
@@ -35,99 +40,51 @@
...
-    -------
-
-    shuffled_data: tensor
-        Shuffled node or edge data.
-    """
-    # Broadcast basic information of loaded data:
-    #   1. number of data lines
-    #   2. data dimension
-    #   3. data type
+def _broadcast_shape(data, rank, world_size, num_parts, isfeatdata, featname):
```
Rhett-Ying, line 43: `is_feat_data`, `feat_name`?

Rhett-Ying, line 43: add docstring for this function?

tools/distpartitioning/utils.py
```diff
@@ -312,23 +358,28 @@ def augment_edge_data(
     etype_offset = {}
     offset = 0
     for etype_name, tid_range in edge_tids.items():
-        assert int(tid_range[0][0]) == 0
-        assert len(tid_range) == num_parts
+        # assert int(tid_range[0][0]) == 0, f'tid_range = {tid_range}'
```
Rhett-Ying, line 361: remove instead of commenting out

tools/distpartitioning/utils.py
```diff
@@ -512,6 +563,34 @@ def write_dgl_objects(
...
+    for idx, typename in enumerate(names):
+        gnid_end += counts[typename]
+        tid_dict[typename] = [[0, counts[typename]]]
+        gid_dict[typename] = np.array([gnid_start, gnid_end]).reshape([1, 2])
+        gnid_start = gnid_end
+
+    return tid_dict, gid_dict
+
+
+def get_typecounts(ntypes, ntype_counts):
```
Rhett-Ying, line 585: the function name is not intuitive.

tests/tools/utils.py
```diff
@@ -118,11 +118,13 @@ def _chunk_graph(
     metadata["graph_name"] = name
     metadata["node_type"] = g.ntypes
 
+    # add node_type_counts
+    metadata["num_nodes_per_type"] = [g.num_nodes(ntype) for ntype in g.ntypes]
```
Rhett-Ying, line 122: pls update the doc page as well: https://github.com/dmlc/dgl/blob/master/docs/source/guide/distributed-preprocessing.rst#example-mag240m-lsc

Rhett-Ying (changes requested):
CI is failing yet. pls fix it

tools/distpartitioning/dataset_utils.py
kylasa: Done.

kylasa: Done.

tools/distpartitioning/parmetis_preprocess.py
```diff
@@ -113,80 +116,34 @@ def convert_to_numpy_and_write_back(data_df):
...
-        elif edges_format == constants.STR_PARQUET:
-            data_df = pq.read_table(edge_data_files[rank])
-            data_df = data_df.rename_columns(["f0", "f1"])
-        else:
-            raise NotImplementedError(f"Unknown edge format {edges_format}")
+        # handle any no. of files case here.
+        file_idxes = generate_read_list(len(edge_data_files), params.num_parts)
+
+        for idx in file_idxes[rank]:
+            if edges_format == constants.STR_CSV:
```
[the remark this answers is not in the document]
kylasa: Done.

tools/distpartitioning/utils.py
kylasa: Done.

tools/distpartitioning/utils.py
kylasa: Changed to 'get_ntype_counts_map', as suggested.

tests/tools/utils.py
kylasa: Will update the documents page as soon as I know the process and who is the POC for updating the webpage.

tests/tools/test_dist_part.py
kylasa: Done.

tools/distpartitioning/constants.py
```diff
@@ -38,3 +36,8 @@
 STR_GRAPH_NAME = "graph_name"
 STR_NODE_FEATURES = "node_features"
 STR_EDGE_FEATURES = "edge_features"
+
+STR_NODE_TYPE_COUNTS = "num_nodes_per_type"
```
Rhett-Ying, line 40: `STR_NUM_NODES_PER_TYPE`?

Rhett-Ying (approved):

tools/distpartitioning/constants.py
kylasa: Done.

PR #5311, opened 2023-02-17
[Feature] Add `state_dict`, `load_state_dict`, `param_groups` to `dgl.optim.SparseGradOptimizer`

### LINKED ISSUE #5270: Add `state_dict`, `load_state_dict`, `param_groups` to `dgl.optim.SparseGradOptimizer`
## 🔨Work Item

- [ ] Add `state_dict`
- [ ] Add `load_state_dict`
- [ ] Add `param_groups`

## Description

`dgl.optim.SparseGradOptimizer` is very useful for training large-scale model-parallel node embeddings. However, it didn't follow PyTorch's optimizer APIs, and thus requires extra effort for debugging and checkpointing.

With this work item, we hope `dgl.optim.SparseGradOptimizer` to emulate PyTorch optimizers' behaviors.

## Depending work items or issues

This may depend on https://github.com/dmlc/dgl/pull/4989 to use PyTorch's NCCL communicator.

yaox12:
## Description

Resolves #5270

1. Add `state_dict`, `load_state_dict` to `dgl.optim.SparseGradOptimizer`
2. Add `param_groups` to `dgl.optim.SparseGradOptimizer`. Note that different from torch's optimizer, our `param_groups` doesn't contain parameters because getting the whole embedding is very expensive. It still contains other attributes, e.g., lr, betas, eps, for debugging.
3. To support the above functionalities, we further add `_all_get_optm_state` and `_all_set_optm_state` to `dgl.nn.NodeEmbedding`. Ideally, the optimizer states should be attached to the optimizer instead of the embedding, however, to gather multi-GPU states, we have to leverage the TCPStore and communicator of the embedding, so I just follow the current design.
4. Add unit tests.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/nn/pytorch/sparse_emb.py
+++ b/python/dgl/nn/pytorch/sparse_emb.py
@@ -349,7 +349,12 @@ class NodeEmbedding:  # NodeEmbedding
             th.distributed.barrier()
 
     def _all_get_tensor(self, shared_name, tensor, shape):
-        """A helper function to get model-parallel tensors."""
+        """A helper function to get model-parallel tensors.
+
+        This method must and only need to be called in multi-GPU DDP training.
+        For now, it's only used in ``all_get_embedding`` and
+        ``_all_get_optm_state``.
+        """
         # create a shared memory tensor
         if self._rank == 0:
             # root process creates shared memory
@@ -467,7 +472,6 @@ class NodeEmbedding:  # NodeEmbedding
             # stored in CPU memory
             if self._rank <= 0:
                 for state, new_state in zip(self._optm_state, states):
-                    print(state.shape, new_state.shape)
                     state[:] = F.copy_to(
                         new_state, ctx=F.context(self._tensor)
                     )[:]
--- a/python/dgl/nn/pytorch/sparse_emb.py
+++ b/python/dgl/nn/pytorch/sparse_emb.py
@@ -348,6 +348,34 @@ class NodeEmbedding:  # NodeEmbedding
         if th.distributed.is_initialized():
             th.distributed.barrier()
 
+    def _all_get_tensor(self, shared_name, tensor, shape):
+        # create a shared memory tensor
+        if self._rank == 0:
+            # root process creates shared memory
+            val = create_shared_mem_array(
+                shared_name,
+                shape,
+                tensor.dtype,
+            )
+            self._store.set(shared_name, shared_name)
+        else:
+            self._store.wait([shared_name])
+            val = get_shared_mem_array(
+                shared_name,
+                shape,
+                tensor.dtype,
+            )
+        # need to map indices and slice into existing tensor
+        idxs = self._partition.map_to_global(
+            F.arange(0, tensor.shape[0], ctx=F.context(tensor)),
+            self._rank,
+        ).to(val.device)
+        val[idxs] = tensor.to(val.device)
+
+        # wait for all processes to finish
+        th.distributed.barrier()
+        return val
+
     def all_get_embedding(self):
         """Return a copy of the embedding stored in CPU memory. If this is a
         multi-processing instance, the tensor will be returned in shared
@@ -367,35 +395,11 @@ class NodeEmbedding:  # NodeEmbedding
                 # non-multiprocessing
                 return self._tensor.to(th.device("cpu"))
             else:
-                # create a shared memory tensor
-                shared_name = self._name + "_gather"
-                if self._rank == 0:
-                    # root process creates shared memory
-                    emb = create_shared_mem_array(
-                        shared_name,
-                        (self._num_embeddings, self._embedding_dim),
-                        self._tensor.dtype,
-                    )
-                    self._store.set(shared_name, shared_name)
-                else:
-                    self._store.wait([shared_name])
-                    emb = get_shared_mem_array(
-                        shared_name,
-                        (self._num_embeddings, self._embedding_dim),
-                        self._tensor.dtype,
-                    )
-                # need to map indices and slice into existing tensor
-                idxs = self._partition.map_to_global(
-                    F.arange(
-                        0, self._tensor.shape[0], ctx=F.context(self._tensor)
-                    ),
-                    self._rank,
-                ).to(emb.device)
-                emb[idxs] = self._tensor.to(emb.device)
-
-                # wait for all processes to finish
-                th.distributed.barrier()
-                return emb
+                return self._all_get_tensor(
+                    f"{self._name}_gather",
+                    self._tensor,
+                    (self._num_embeddings, self._embedding_dim),
+                )
         else:
             # already stored in CPU memory
             return self._tensor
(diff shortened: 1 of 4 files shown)


python/dgl/nn/pytorch/sparse_emb.py
```diff
@@ -376,6 +376,36 @@ def all_set_embedding(self, values):
         if th.distributed.is_initialized():
             th.distributed.barrier()
 
+    def _all_get_tensor(self, shared_name, tensor, shape):
+        """A helper function to get model-parallel tensors."""
+        # create a shared memory tensor
+        if self._rank == 0:
```
peizhou001, line 382: If the embedding is created in shared memory, will it has a legal `rank` number?

python/dgl/nn/pytorch/sparse_emb.py
```diff
@@ -376,6 +376,36 @@ def all_set_embedding(self, values):
...
+        # need to map indices and slice into existing tensor
+        idxs = self._partition.map_to_global(
+            F.arange(0, tensor.shape[0], ctx=F.context(tensor)),
+            self._rank,
+        ).to(val.device)
+        val[idxs] = tensor.to(val.device)
+
+        self._store.delete_key(shared_name)
+        # wait for all processes to finish
+        th.distributed.barrier()
```
peizhou001, line 406: Ditto, will distributed always inited?

python/dgl/optim/pytorch/sparse_optim.py
```diff
@@ -474,6 +474,40 @@ def zero_grad(self):
...
+
+        NOTE: This method must be called by all processes sharing the
+        underlying embedding, or it may result in a deadlock.
+
+        Returns
+        -------
+        dictionary of optimizer states
+            The optimizer states stored in CPU memory.
+        """
+        return {emb.name: emb._all_get_optm_state() for emb in self._params}
```
peizhou001, line 491: Do we need also store basic hyper parameters like `lr`, `beta`?

tests/python/pytorch/optim/test_optim.py
```diff
@@ -590,6 +589,86 @@ def test_multiprocess_sparse_adam_zero_step_cuda_tensor(num_workers):
     assert F.allclose(dgl_weight, torch_weight)
 
 
+def start_sparse_adam_state_dict_worker(
```
peizhou001, line 592: Is shared memory embedding also tested?

python/dgl/nn/pytorch/sparse_emb.py
```diff
@@ -376,6 +376,36 @@ def all_set_embedding(self, values):
...
+        if self._rank == 0:
+            # root process creates shared memory
+            val = create_shared_mem_array(
+                shared_name,
+                shape,
+                tensor.dtype,
+            )
+            self._store.set(shared_name, shared_name)
+        else:
+            self._store.wait([shared_name])
```
peizhou001, line 391: Will this blocked when the case is not multi-gpu?

python/dgl/nn/pytorch/sparse_emb.py
peizhou001, line 382: Have this covered the case about single device?

python/dgl/nn/pytorch/sparse_emb.py
peizhou001, line 406: Will this cause error when th uninitialized?

python/dgl/nn/pytorch/sparse_emb.py
```diff
@@ -376,6 +376,36 @@ def all_set_embedding(self, values):
...
+                tensor.dtype,
+            )
+        # need to map indices and slice into existing tensor
+        idxs = self._partition.map_to_global(
+            F.arange(0, tensor.shape[0], ctx=F.context(tensor)),
+            self._rank,
+        ).to(val.device)
+        val[idxs] = tensor.to(val.device)
+
+        self._store.delete_key(shared_name)
```
peizhou001, line 404: Why delete it here? will it cause the `val` be revoked?

python/dgl/nn/pytorch/sparse_emb.py
yaox12: This function is abstracted from https://github.com/dmlc/dgl/blob/45153fc07833c84fd0bd8fcdeaf4bd4e796b8d7c/python/dgl/nn/pytorch/sparse_emb.py#L398-L426 so that it can be reused to gather optimizer states.

It's an internal helper function that should only be called in multi-GPU DDP training. The single-GPU and CPU cases are routed in upper level functions such as `all_get_embedding()` and `_all_get_optm_state()`. I'll update the docstring.

python/dgl/nn/pytorch/sparse_emb.py
yaox12: Same as above.

python/dgl/nn/pytorch/sparse_emb.py
yaox12: Same as above.

python/dgl/nn/pytorch/sparse_emb.py
yaox12: > Why delete it here?

Say we try to save the checkpoint file at the end of every epoch, without deleting this key, from the second epoch and so on, processes with rank > 0 will try to get the shared array before rank 0 has created it, because the key is already in the TCPStore.

> Will it cause the val be revoked?

No. The object stored in `_store` is just a string.

python/dgl/optim/pytorch/sparse_optim.py
yaox12: Updated.

python/dgl/nn/pytorch/sparse_emb.py
peizhou001: Got you, it make sense if the case is just for DDP.

python/dgl/nn/pytorch/sparse_emb.py
peizhou001: I see. It means each call will allocate a new address of memory. It should be ok as the local `state_dict` in pytorch should also return a new dict every time.

python/dgl/optim/pytorch/sparse_optim.py
peizhou001: Could we make the format uniformed with [dist embedding state dict](https://github.com/dmlc/dgl/blob/master/python/dgl/distributed/optim/pytorch/sparse_optim.py#L66-L82)? So that user won't be confused to get  two different formats.

yaox12: Some thoughts on this:
1. The current format follows PyTorch optimizer's style (keys are "state" and "param_groups"):
```
{
    "state": {
        emb_name: state_tensor,
        ...
    },
    "param_groups": {
        "lr": ,
        ...
    }
}
```
2. Different from the **local** state dict of dist embedding, the state dict here doesn't have "ids", so I feel the following structure is a bit complicated for it:
```
{
    "params": {
        ...,
    },
    "emb_states": {
        emb_name: {
            "ids": ...,  # don't need it 
            "state": {   # the example in your link misses this key
                state_name: state_tensor, 
                ...,
            },
        },
    }
}
```

peizhou001: I see, follow pytorch style is a good choice.

peizhou001 (approved):

yaox12:
@dgl-bot

PR #5334, opened 2023-02-20
[Fix] removed pragma omp for in SpMMCreateBlocks

Kacper-Pietkun:
## Description
Removed "#pragma omp for", because it was used without parallel keyword.
Adding parallel keyword to the pragma made loop run slower than without it. It was slower because this loop runs for a small number of iterations.

Number of iterations can be calculated with following formula:
$\dfrac{T*C(M-1)}{M} + 1$
where:
- $C$ is a constant NUM_BLOCKS_PER_THREAD = 20
- $T$ is a result of omp_get_max_threads function
- $M$ is csr.num_rows

$M$ is neglectable when it comes to the number of loop's iterations.
$C$ is a constant.
Number of iterations depends mainly on $T$, however it will not make loop run for a very big number of iterations. For example, when $T = 144$, there will be no more iterations than 2900 for any $M$. 

I tried using parallel keyword and runtime::parallel_for function, but in both cases loop was executing much slower than in a situation without "#pragma omp for" (runtime::parallel_for function was slightly faster than just adding parallel keyword). Profiling was performed on a GraphSage (examples/pytorch/ogb/ogbn-products/graphsage) with ogbn-products and ogbn-papers100M datasets.

## Checklist


bgawrych (approved):

bartekkuncer (approved):

frozenbugs:
@dgl-bot

peizhou001:
Hi @Kacper-Pietkun, it make sense to remove the 'opm for' as the iteration space is small. Could you provide more data about how much speed gain did you get?

Kacper-Pietkun:
> Hi @Kacper-Pietkun, it make sense to remove the `opm for` as the iteration space is small. Could you provide more data about how much speed gain did you get?

Hello @peizhou001, my change doesn't speed up the code, because it is already executed on only one thread (because there is no parallel keyword before the pragma omp for, thus there is no parallelism). The goal of this PR is to just clean the code.

peizhou001 (approved):

Issue #5295, opened 2023-02-14
Fix print statements in distDGL


isratnisa:
### Misleading print statements
1. When training on large graphs, distDGL takes a while  to load the dataset and convert it to different storage formats (up to 1.5 hours in my experience). It does not show any log in the mean time which makes the user think that either the code got stuck or there is some issue with socket connection. 

For example, RGCN on papers100M dataset waits after this step for a long time:

```
[21:50:21] /opt/dgl/src/rpc/rpc.cc:159: Receiver with NetType~socket is created.
[21:50:21] /opt/dgl/src/rpc/rpc.cc:159: Receiver with NetType~socket is created.
```
 
We can add more statements such as:
```
Loading dataset...
Converting graph format...
Storing graph or embedding tables...
```
etc.

### Too many optional messages

Also, the current code prints too many optional messages like padding info (`Pad nids from 69525 to 69524`), partition info (`part 0, train: 69525, val: 8685, test: 5506`), etc. Most of the messages should be printed only if a user turns on the verbose mode.

@Rhett-Ying

Rhett-Ying:
confirmed. thanks for reporting this...

isratnisa:
Also, I think the current code prints too many optional messages like padding info (`Pad nids from 69525 to 69524`), partition info (`part 0, train: 69525, val: 8685, test: 5506`), etc. Most of the messages should be printed only if a user turns on the verbose mode.

Rhett-Ying:
@isratnisa  could you help fix it? just print what you think that are essential messages.

Rhett-Ying:
> 

These log are printed by the example script? `//examples/pytorch/grpahsage/dist/trian_dist.py`? Actually there's no `verbose` option in DistDGL for now.

PR #5350, opened 2023-02-21
[DistDGL] add log when load graph and data

Rhett-Ying:
## Description

fix for https://github.com/dmlc/dgl/issues/5295
## Checklist
- [ ] Code is well-documented


python/dgl/distributed/partition.py
```diff
@@ -294,10 +301,24 @@ def load_partition_feats(
...
     if load_edges:
-        edge_feats = load_tensors(relative_to_config(part_files["edge_feats"]))
+        feat_path = relative_to_config(part_files["edge_feats"])
+        print(
+            f"Start to load edge data from {feat_path} which is "
+            f"{os.path.getsize(feat_path)} bytes. It may take non-trivial "
+            "time for large data."
+        )
+        edge_feats = load_tensors(feat_path)
+        print("Finished loading node data...")
```
isratnisa, line 321: Typo `print("Finished loading edge data...")`

python/dgl/distributed/partition.py
```diff
@@ -190,7 +190,14 @@ def load_partition(part_config, part_id, load_feats=True):
     assert (
         "part_graph" in part_files
     ), "the partition does not contain graph structure."
-    graph = load_graphs(relative_to_config(part_files["part_graph"]))[0][0]
+    partition_path = relative_to_config(part_files["part_graph"])
+    print(
+        f"Start to load partition from {partition_path} which is "
+        f"{os.path.getsize(partition_path)} bytes. It may take non-trivial "
+        "time for large partition."
+    )
```
isratnisa, line 198: Do we need to say `It may take non-trivial time for large data`? Users should expect it from the byte size IMHO.

python/dgl/distributed/partition.py
```diff
@@ -190,7 +190,14 @@ def load_partition(part_config, part_id, load_feats=True):
...
+    graph = load_graphs(partition_path)[0][0]
+    print("Finished loading partition...")
```
isratnisa, line 200: I think we don't need `...` here as we are not waiting for anything.

python/dgl/distributed/partition.py
```diff
@@ -294,10 +301,24 @@ def load_partition_feats(
     ), "the partition does not contain edge feature."
     node_feats = None
     if load_nodes:
-        node_feats = load_tensors(relative_to_config(part_files["node_feats"]))
+        feat_path = relative_to_config(part_files["node_feats"])
+        print(
+            f"Start to load node data from {feat_path} which is "
+            f"{os.path.getsize(feat_path)} bytes. It may take non-trivial "
+            "time for large data."
+        )
```
isratnisa, line 309: ditto

python/dgl/distributed/partition.py
```diff
@@ -294,10 +301,24 @@ def load_partition_feats(
...
+        node_feats = load_tensors(feat_path)
+        print("Finished loading node data...")
```
isratnisa, line 311: Same. May be no `...`

python/dgl/distributed/partition.py
```diff
@@ -294,10 +301,24 @@ def load_partition_feats(
...
     edge_feats = None
     if load_edges:
-        edge_feats = load_tensors(relative_to_config(part_files["edge_feats"]))
+        feat_path = relative_to_config(part_files["edge_feats"])
+        print(
+            f"Start to load edge data from {feat_path} which is "
+            f"{os.path.getsize(feat_path)} bytes. It may take non-trivial "
+            "time for large data."
```
isratnisa, line 318: Ditto

isratnisa (changes requested):
Thank you for taking care of the issue. I have left some comments.

python/dgl/distributed/partition.py
Rhett-Ying: let's remove it then.

python/dgl/distributed/partition.py
Rhett-Ying: removed

PR #5366, opened 2023-02-22
[Feature] Import PyTorch's `pin_memory()` method for DGL graph structure

chang-l:
## Description
This PR imports pytorch's `CachingHostAllocator` [[ref] ](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/cuda/CachingHostAllocator.cpp)and its associated functions (like `CachingHostAllocator_recordEvent` and ``CachingHostAllocator_emptyCache``) to DGL's `tensoradaptor`. It enhances/enriches DGL's native memory pinning strategy (`cudaHostRegister`), by using backend's pinned memory pool (via`cudaHostAlloc`) to avoid overhead from `cudaHostRegister` and frequent memory allocations/deallocations.

While `cudaHostRegister` (`DGLGraph.pin_memory_`) is more suitable to pin large data (like entire feature/graph) ONCE for future frequent GPU accesses (no memory allocation involved but with the compromising of speed),  pytorch's pinned memory pool (`DGLGraph(Index).pin_memory`) is better for objects (small memory blocks) that could be created/destroyed often (like DGLGraph/feature for batches).

Based on this PR, it would be easier to implement `pin_memory()` method for DGLGraph object, and let PyTorch's dataloader pick it up when `pin_memory=True` to enable async H2D copying of (sub)graph structure (or feature if needed). Complete dataloader support should be included in a separate PR.

## Checklist
- [x] Code is well-documented

## Changes
C++ (DGL core):
- Imported `getCachingHostAllocator`, `CachingHostAllocator_recordEvent` and `CachingHostAllocator_emptyCache` to completely support pytorch's cachingHostAllocator.
- Added/propagated `pin_memory*` methods to all necessary classes(ndarry/csr/coo/unitgraph/heterograph).
- Added `RecordedCopyDataFromTo` to fuse data copy (H2D) and following `CachingHostAllocator_recordEvent` to manage the (pinned)resource consistently with[backend PyTorch](https://github.com/pytorch/pytorch/blob/5f24b2b1f062e930c41b576e44195f9912dbc958/aten/src/ATen/cuda/CachingHostAllocator.cpp#L128-L137).
- Empty the caching pool if calling `cudaHostRegister` to release resources/prepare enough page-locked memory needed by `cudaHostRegister`

Python:
- Added outplace `pin_memory()` for `heterograph_index`
- Unit test

## To-dos
_(maybe need separate PRs)_

Python:
- Add outplace `pin_memory()` for `heterograph` or other equivalents to integrate with dataloader.


chang-l:
@jermainewang  Pls help add other reviewers. Thanks.

tests/python/common/test_heterograph.py
```diff
@@ -1255,6 +1255,57 @@ def test_pin_memory_(idtype):
     assert not g3.is_pinned()
 
 
+@unittest.skipIf(
```
frozenbugs, line 1258: Maybe create a test_heterograph_index.py and put the test case in it? In the future, we can gradually move all test related to heterograph_index.py to it.

python/dgl/heterograph_index.py
```diff
@@ -254,6 +254,22 @@ def copy_to(self, ctx):
...
+              The graph structure must be on CPU to be pinned.
+              If the graph struture is already pinned, the function returns a graph
+              structure with the same storage.
+
+        Returns
+        -------
+        HeteroGraphIndex
+            The pinned graph index.
+        """
+        # if bool(_CAPI_DGLHeteroIsPinned(self)) : return self
```
frozenbugs, line 270: Remove the code if we don't need it.

src/runtime/ndarray.cc
```diff
@@ -206,6 +211,46 @@ void NDArray::CopyFromTo(DGLArray* from, DGLArray* to) {
...
+  NDArray ret = Internal::Create(shape, dtype, ctx);
+  // setup memory content
+  size_t size = GetDataSize(ret.data_->dl_tensor);
+  // size_t alignment = GetDataAlignment(ret.data_->dl_tensor);
+  if (size > 0) {
+    ret.data_->dl_tensor.data = DeviceAPI::Get(kDGLCUDA)->AllocPinnedDataSpace(
+        size, &(ret.data_->pyt_ctx), &(ret.data_->pyt_raw_deleter));
+    CHECK(
+        ret.data_->pyt_ctx != nullptr && ret.data_->pyt_raw_deleter != nullptr)
+        << "Can not return proper CachingHostAllocator";
```
frozenbugs, line 248: Maybe just "Data allocation failed."? From the code, I can't tell the failed is caused by CachingHostAllocator or other issue.

src/runtime/ndarray.cc
```diff
@@ -206,6 +211,46 @@ void NDArray::CopyFromTo(DGLArray* from, DGLArray* to) {
...
+  DGLContext ctx = from->ctx.device_type != kDGLCPU ? from->ctx : to->ctx;
+  CHECK(ctx.device_type == kDGLCUDA)
+      << "Can not record event if not cuda device";
+  DeviceAPI::Get(kDGLCUDA)->RecordedCopyDataFromTo(
+      from->data, static_cast<size_t>(from->byte_offset), to->data,
+      static_cast<size_t>(to->byte_offset), from_size, from->ctx, to->ctx,
+      from->dtype, pyt_ctx);
+}
+
+NDArray NDArray::PinnedEmpty(
```
frozenbugs, line 236: Maybe better name than PinnedEmpty.
AllocEmptyPinnedMemory?

include/dgl/runtime/tensordispatch.h
```diff
@@ -149,6 +149,37 @@ class TensorDispatcher {
     auto entry = entrypoints_[Op::kRecordStream];
     FUNCCAST(tensoradapter::RecordStream, entry)
     (ptr, static_cast<cudaStream_t>(stream), device_id);
+#endif  // DGL_USE_CUDA
+  }
```
frozenbugs, line 153: Add a blank line to give the function some space to breathe.

tensoradapter/pytorch/torch.cpp
```diff
@@ -55,6 +56,49 @@ TA_EXPORTS void RecordStream(void* ptr, cudaStream_t stream, int device_id) {
...
+
+ private:
+  std::unique_ptr<void, c10::DeleterFnPtr> ptr_;
+};
+
+TA_EXPORTS void* CUDARawHostAlloc(
+    size_t nbytes, void** ctx, void** raw_deleter) {
+  auto data_ptr = at::cuda::getCachingHostAllocator()->allocate(nbytes);
+  auto raw = data_ptr.get();
+  *ctx = data_ptr.get_context();  // raw  ctx ptr for record event
```
frozenbugs, line 73:
```
// Return the raw ctx ptr for recording event.
*ctx = data_ptr.get_context();
```

tensoradapter/pytorch/torch.cpp
```diff
@@ -55,6 +56,49 @@ TA_EXPORTS void RecordStream(void* ptr, cudaStream_t stream, int device_id) {
...
+
+  auto* data_deleter = new CUDAHostDeleter(
```
frozenbugs, line 75: ditto.

tensoradapter/pytorch/torch.cpp
```diff
@@ -55,6 +56,49 @@ TA_EXPORTS void RecordStream(void* ptr, cudaStream_t stream, int device_id) {
...
+      data_ptr.move_context());  // transfer ownership to raw_deleter
+  *raw_deleter = static_cast<void*>(data_deleter);
+  return raw;
+}
+
+// every single CUDARawHostAlloc has an unique CUDAHostDeleter obj
```
frozenbugs, line 81: // Designated CUDAHostDeleter for CUDARawHostAlloc.

tests/python/common/test_heterograph_index.py
```diff
@@ -0,0 +1,97 @@
...
+import scipy.sparse as ssp
+import test_utils
+from dgl import DGLError
+from scipy.sparse import rand
+from test_utils import get_cases, parametrize_idtype
+from utils import assert_is_identical_hetero
+
+
+def create_test_heterograph(idtype):
+    # test heterograph from the docstring, plus a user -- wishes -- game relation
```
frozenbugs, line 22: No need to mention the heterograph. is from the docstring, the unit test should be self-contained.

tests/python/common/test_heterograph_index.py
```diff
@@ -0,0 +1,97 @@
...
+@unittest.skipIf(
+    F._default_context_str == "cpu", reason="Need gpu for this test"
+)
+@unittest.skipIf(
+    dgl.backend.backend_name != "pytorch",
+    reason="Pinning graph outplace only supported for PyTorch",
+)
+@parametrize_idtype
+def test_pin_memory(idtype):
+    # TODO: rewrite this test case to accept different graphs so we
```
frozenbugs, line 54: Do you have an issue to track the TODO? Usually I would suggest we follow TODO(#{ISSUE_ID}): blabla.

tests/python/common/test_heterograph_index.py
```diff
@@ -0,0 +1,97 @@
...
+    #  can test reverse graph and batched graph
```
frozenbugs, line 55: just 1 space between # and the first char, even in a newline.

tests/python/common/test_heterograph_index.py
```diff
@@ -0,0 +1,97 @@
...
+    g = create_test_heterograph(idtype)
+    g.nodes["user"].data["h"] = F.ones((3, 5))
+    g.nodes["game"].data["i"] = F.ones((2, 5))
+    g.edges["plays"].data["e"] = F.ones((4, 4))
+    g = g.to(F.cpu())
+    assert not g.is_pinned()
+
+    # pin a CPU graph
```
frozenbugs, line 63: `# Test pinning a CPU graph.`

tests/python/common/test_heterograph_index.py
```diff
@@ -0,0 +1,97 @@
...
+    g._graph.pin_memory()
+    assert not g.is_pinned()
+    g._graph = g._graph.pin_memory()
+    assert g.is_pinned()
+    assert g.device == F.cpu()
+
+    # it's fine to clone with new formats, but new graphs are not pinned
```
frozenbugs, line 70: This comment is a little bit confusing to me. Can you simplify it to a pure statement?

tests/python/common/test_heterograph_index.py
```diff
@@ -0,0 +1,97 @@
...
+    assert not g.formats("csr").is_pinned()
+    # 'coo' formats is already created and thus not cloned
+    assert g.formats("coo").is_pinned()
+
+    g1 = g.to(F.cuda())
+    # error pinning a GPU graph
+    with pytest.raises(DGLError):
+        g1._graph.pin_memory()
+
+    # test pin empty homograph
```
frozenbugs, line 83: `# Test pinning an empty homograph.`

tests/python/common/test_heterograph_index.py
```diff
@@ -0,0 +1,97 @@
...
+    g2 = dgl.graph(([], []))
+    g2._graph = g2._graph.pin_memory()
+    assert g2.is_pinned()
+
+    # test pin heterograph with 0 edge of one relation type
```
frozenbugs, line 88: ditto.

tests/python/common/test_heterograph_index.py
```diff
@@ -0,0 +1,97 @@
...
+    # it's fine to clone with new formats, but new graphs are not pinned
+    # >>> g.formats()
+    # {'created': ['coo'], 'not created': ['csr', 'csc']}
+    assert not g.formats("csc").is_pinned()
+    assert not g.formats("csr").is_pinned()
+    # 'coo' formats is already created and thus not cloned
+    assert g.formats("coo").is_pinned()
+
+    g1 = g.to(F.cuda())
+    # error pinning a GPU graph
```
frozenbugs, line 79:
```
# Test pinning a GPU graph will cause error raised.
g1 = g.to(F.cuda())
with pytest.raises(DGLError):
```

frozenbugs:
Hi Chang, sorry being picky on the code quality. Overall the PR looks very good to me, just some small tweak to make it shiny.

chang-l:
No worries @frozenbugs  I am always happy to iterate and improve in every aspect.

Thank you for taking the time and review this PR. I will address them all tomorrow.

src/runtime/cuda/cuda_device_api.cc
```diff
@@ -214,10 +232,33 @@ class CUDADeviceAPI final : public DeviceAPI {
...
     return true;
   }
 
+  void* AllocPinnedDataSpace(
+      size_t nbytes, void** ctx, void** deleter) override {
+    // prevent users from pinning empty tensors or graphs
+    if (nbytes == 0) return nullptr;
+    TensorDispatcher* td = TensorDispatcher::Global();
+    CHECK(td->IsAvailable())
+        << "CachingHost allocator only available with TensorAdatpor";
```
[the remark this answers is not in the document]
chang-l: Sry for my typos. Fixed.

src/runtime/ndarray.cc
```diff
@@ -206,6 +211,46 @@ void NDArray::CopyFromTo(DGLArray* from, DGLArray* to) {
...
+      << "DGLArrayCopyFromTo: The size must exactly match";
+
+  CHECK(
+      from->ctx.device_type == to->ctx.device_type ||
+      from->ctx.device_type == kDGLCPU || to->ctx.device_type == kDGLCPU)
+      << "Can not copy across different ctx types directly";
+
+  // Use the context that is *not* a cpu context to get the correct device
+  // api manager.
+  DGLContext ctx = from->ctx.device_type != kDGLCPU ? from->ctx : to->ctx;
```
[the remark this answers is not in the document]
chang-l: >It seems the whole logic here (L220-L229) assumes that:

>   There are only 2 device_type: kDGLCUDA, and kDGLCPU.
>   the only invalid combination is both equal to kDGLCPU.

I don't think it assumes there are only 2 device types. I guess that's why these checkers avoid using `kDGLCUDA` explicitly. 
If we decide to simply the code logic to support only two device types (CUDA and CPU), then I agree the checkers in all copy functions can be simplified.

src/runtime/ndarray.cc
chang-l: Updated comment. Actually, the previous call `AllocPinnedDataSpace` changes the two pointers `pyt_ctx` and `pyt_raw_deleter`; and this is to check the return status of `AllocPinnedDataSpace` (CUDACachingAllocator).

src/runtime/ndarray.cc
chang-l: There is a similar function `Empty` in the same file. `PinnedEmpty` is very similar to that function but the empty ndarray returned within pinned memory. The naming convention within `PinnedEmpty` function can also map to `Empty` function, i.e., `Empty` calls `AllocDataSpace`, while `PinnedEmpty` calls `AllocPinnedDataSpace`

tests/python/common/test_heterograph_index.py
chang-l: Let me just remove this comment as it seems not related to this test...

include/dgl/runtime/ndarray.h
```diff
@@ -182,6 +182,8 @@ class NDArray {
    *       IsPinned: directly return;
    *       kDGLCUDA: invalid, will throw an error.
    */
+
+  inline NDArray PinMemory();
```
yaox12, line 186: Can you add a specific comment for `PinMemory()`? For example, "Return a copy of the array in pinned memory".

include/dgl/runtime/ndarray.h
```diff
@@ -229,6 +231,9 @@ class NDArray {
    */
   DGL_DLL static NDArray Empty(
       std::vector<int64_t> shape, DGLDataType dtype, DGLContext ctx);
+
+  DGL_DLL static NDArray PinnedEmpty(
```
yaox12, line 235: Ditto.

include/dgl/runtime/ndarray.h
```diff
@@ -286,6 +291,8 @@ class NDArray {
   DGL_DLL static void CopyFromTo(DGLArray* from, DGLArray* to);
   DGL_DLL static void CopyFromTo(
       DGLArray* from, DGLArray* to, DGLStreamHandle stream);
+  DGL_DLL static void RecordedCopyFromTo(
```
yaox12, line 294: Ditto.

src/graph/unit_graph.cc
```diff
@@ -140,6 +140,11 @@ class UnitGraph::COO : public BaseHeteroGraph {
     return COO(meta_graph_, adj_.CopyTo(ctx));
   }
 
+  COO PinMemoryOutPlace() {
```
yaox12, line 143: Pls add a brief comment.

src/graph/unit_graph.cc
```diff
@@ -535,6 +540,11 @@ class UnitGraph::CSR : public BaseHeteroGraph {
     }
   }
 
+  CSR PinMemoryOutPlace() {
```
yaox12, line 543: Ditto.

src/graph/unit_graph.h
```diff
@@ -214,6 +214,8 @@ class UnitGraph : public BaseHeteroGraph {
    */
   void PinMemory_() override;
 
+  HeteroGraphPtr PinMemoryOutPlace();
```
yaox12, line 217: Ditto.

yaox12 (commented):
LGTM except for some comments issues.

src/runtime/ndarray.cc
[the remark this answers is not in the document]
frozenbugs: But the current implementation only works for kDGLCUDA, and kDGLCPU.
Check at L220, make sure "either from or to is kDGLCPU" or both are same.
And Check L227, get the non kDGLCPU, and L228 checks the non kDGLCPU one equals kDGLCUDA.

The logic equals to the code I suggested.

src/runtime/ndarray.cc
```diff
@@ -206,6 +213,46 @@ void NDArray::CopyFromTo(DGLArray* from, DGLArray* to) {
...
+void NDArray::RecordedCopyFromTo(DGLArray* from, DGLArray* to, void* pyt_ctx) {
+  size_t from_size = GetDataSize(*from);
+  size_t to_size = GetDataSize(*to);
+  CHECK_EQ(from_size, to_size)
+      << "DGLArrayCopyFromTo: The size must exactly match.";
+
+  CHECK(
+      from->ctx.device_type == to->ctx.device_type ||
+      from->ctx.device_type == kDGLCPU || to->ctx.device_type == kDGLCPU)
+      << "Can not copy across different ctx types directly.";
```
nv-dlasalle, line 225: Why can't I copy from one GPU to another? Also, the error message here only captures the first line of the check.

chang-l: This checker doesn't block GPU-GPU copy. GPU-GPU copy is the copy across the same ctx type.
It blocks the copy across 2 non-host device types, which I guess is one of the heuristics in the code. In most runs, other than cpu, we only have one device (CUDA). Thus, maybe we can simplify the logic here...

src/graph/unit_graph.cc
```diff
@@ -1259,6 +1269,37 @@ HeteroGraphPtr UnitGraph::CopyTo(HeteroGraphPtr g, const DGLContext& ctx) {
   }
 }
 
+HeteroGraphPtr UnitGraph::PinMemoryOutPlace() {
```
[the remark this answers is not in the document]
chang-l: Thanks for catching it. I will update it.

python/dgl/heterograph_index.py
```diff
@@ -254,6 +254,21 @@ def copy_to(self, ctx):
         """
         return _CAPI_DGLHeteroCopyTo(self, ctx.device_type, ctx.device_id)
 
+    def pin_memory(self):
+        """Pin this graph to the page-locked memory.
+
+        NOTE: This is an outplace method using the backend pinned memory pool.
+              The graph structure must be on CPU to be pinned.
+              If the graph struture is already pinned, the function returns a graph
+              structure with the same storage.
```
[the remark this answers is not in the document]
chang-l: It will raise an exception from [here](https://github.com/chang-l/dgl/blob/d4fe3281c543a22bb20d1420120b04cb83803948/src/runtime/ndarray.cc#L240)

src/runtime/ndarray.cc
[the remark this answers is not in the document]
chang-l: >The logic equals to the code I suggested.

I agree. Unlike more general function `CopyFromTo` , here, I can simplify the logic as this is dedicated for cuda.

include/dgl/runtime/ndarray.h
```diff
@@ -182,6 +182,8 @@ class NDArray {
    *       IsPinned: directly return;
    *       kDGLCUDA: invalid, will throw an error.
    */
+
```
frozenbugs, line 185: move the empty line from L185 to L177

include/dgl/runtime/ndarray.h
```diff
@@ -454,7 +464,19 @@ inline void NDArray::CopyFrom(DGLArray* other) {
 
 inline void NDArray::CopyFrom(const NDArray& other) {
   CHECK(other.data_ != nullptr);
-  CopyFrom(&(other.data_->dl_tensor));
+  // only one NDArray is pinned
+  // they are on diff devices
```
frozenbugs, line 468: Can you elaborate more? What are 'they'?

include/dgl/runtime/ndarray.h
```diff
@@ -464,7 +486,18 @@ inline void NDArray::CopyTo(DGLArray* other) const {
 
 inline void NDArray::CopyTo(const NDArray& other) const {
   CHECK(other.data_ != nullptr);
-  CopyTo(&(other.data_->dl_tensor));
+  // only one NDArray is pinned and they are on diff devices
```
frozenbugs, line 489: ditto.

include/dgl/aten/coo.h
```diff
@@ -141,6 +143,14 @@ struct COOMatrix {
    *       kDGLCUDA: invalid, will throw an error.
    *       The context check is deferred to pinning the NDArray.
    */
+  inline COOMatrix PinMemory() {
+    if (is_pinned) return *this;
+    return COOMatrix(
```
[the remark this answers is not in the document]
frozenbugs: Does UnpinMemory_ works for cases pinned by this method?

include/dgl/runtime/device_api.h
```diff
@@ -98,6 +98,11 @@ class DeviceAPI {
       const void* from, size_t from_offset, void* to, size_t to_offset,
       size_t num_bytes, DGLContext ctx_from, DGLContext ctx_to,
       DGLDataType type_hint) = 0;
+
+  virtual void RecordedCopyDataFromTo(
```
frozenbugs, line 102: Add doc string.

include/dgl/runtime/device_api.h
```diff
@@ -152,6 +157,9 @@ class DeviceAPI {
    * @return false when pinning an empty tensor. true otherwise.
    */
   DGL_DLL virtual bool PinData(void* ptr, size_t nbytes);
+  DGL_DLL virtual void* AllocPinnedDataSpace(
```
frozenbugs, line 160: Add doc string.

src/runtime/cpu_device_api.cc
```diff
@@ -63,6 +65,11 @@ class CPUDeviceAPI final : public DeviceAPI {
         static_cast<const char*>(from) + from_offset, size);
   }
 
+  void RecordedCopyDataFromTo(
+      void* from, size_t from_offset, void* to, size_t to_offset, size_t size,
+      DGLContext ctx_from, DGLContext ctx_to, DGLDataType type_hint,
+      void* pyt_ctx) final {}
```
frozenbugs, line 71: Should we raise an exception if this is called?

src/runtime/cpu_device_api.cc
```diff
@@ -84,16 +91,18 @@ struct CPUWorkspacePool : public WorkspacePool {
 
 void* CPUDeviceAPI::AllocWorkspace(
     DGLContext ctx, size_t size, DGLDataType type_hint) {
-  TensorDispatcher* td = TensorDispatcher::Global();
-  if (td->IsAvailable()) return td->CPUAllocWorkspace(size);
+  TensorDispatcher* tensor_dispatcher = TensorDispatcher::Global();
+  if (tensor_dispatcher->IsAvailable())
+    return tensor_dispatcher->CPUAllocWorkspace(size);
```
frozenbugs, line 96:
```
  if (tensor_dispatcher->IsAvailable()) {
    return tensor_dispatcher->CPUAllocWorkspace(size);
  }
```

src/runtime/cpu_device_api.cc
```diff
@@ -84,16 +91,18 @@ struct CPUWorkspacePool : public WorkspacePool {
...
   return dmlc::ThreadLocalStore<CPUWorkspacePool>::Get()->AllocWorkspace(
       ctx, size);
 }
 
 void CPUDeviceAPI::FreeWorkspace(DGLContext ctx, void* data) {
-  TensorDispatcher* td = TensorDispatcher::Global();
-  if (td->IsAvailable()) return td->CPUFreeWorkspace(data);
+  TensorDispatcher* tensor_dispatcher = TensorDispatcher::Global();
+  if (tensor_dispatcher->IsAvailable())
+    return tensor_dispatcher->CPUFreeWorkspace(data);
```
frozenbugs, line 105: ditto

src/runtime/cuda/cuda_device_api.cc
```diff
@@ -107,9 +107,10 @@ class CUDADeviceAPI final : public DeviceAPI {
       DGLDataType type_hint) final {
     SetDevice(ctx);
     // Redirect to PyTorch's allocator when available.
-    TensorDispatcher* td = TensorDispatcher::Global();
-    if (td->IsAvailable())
-      return td->CUDAAllocWorkspace(nbytes, getCurrentCUDAStream());
+    TensorDispatcher* tensor_dispatcher = TensorDispatcher::Global();
+    if (tensor_dispatcher->IsAvailable())
+      return tensor_dispatcher->CUDAAllocWorkspace(
+          nbytes, getCurrentCUDAStream());
```
frozenbugs, line 113: ditto

src/runtime/cuda/cuda_device_api.cc
```diff
@@ -119,8 +120,9 @@ class CUDADeviceAPI final : public DeviceAPI {
 
   void FreeDataSpace(DGLContext ctx, void* ptr) final {
     SetDevice(ctx);
-    TensorDispatcher* td = TensorDispatcher::Global();
-    if (td->IsAvailable()) return td->CUDAFreeWorkspace(ptr);
+    TensorDispatcher* tensor_dispatcher = TensorDispatcher::Global();
+    if (tensor_dispatcher->IsAvailable())
+      return tensor_dispatcher->CUDAFreeWorkspace(ptr);
```
frozenbugs, line 125: ditto.

include/dgl/runtime/ndarray.h
BarclayII: Yeah I don't quite understand why we need a specific `RecordedCopyFromTo` under this particular condition.  Looks like this will be called when one array is pinned by PyTorch and the other is not, *and* one is on CPU and the other is on GPU.  Moreover, it doesn't seem to matter *which one* is on CPU or pinned by PyTorch, so I'm more confused.

include/dgl/runtime/tensordispatch.h
```diff
@@ -152,6 +152,38 @@ class TensorDispatcher {
 #endif  // DGL_USE_CUDA
   }
 
+  inline void* CUDAAllocHostWorkspace(
+      size_t nbytes, void** ctx, void** deleter) {
+#ifdef DGL_USE_CUDA
+    auto entry = entrypoints_[Op::kCUDARawHostAlloc];
+    return FUNCCAST(tensoradapter::CUDARawHostAlloc, entry)(
+        nbytes, ctx, deleter);
+#endif  // DGL_USE_CUDA
```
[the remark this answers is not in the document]
BarclayII: Agreed.  Same for other functions.

include/dgl/runtime/tensordispatch.h
```diff
@@ -180,6 +215,10 @@ class TensorDispatcher {
     static constexpr int kCUDARawDelete = 3;
     static constexpr int kCUDACurrentStream = 4;
     static constexpr int kRecordStream = 5;
+    static constexpr int kCUDARawHostAlloc = 6;
+    static constexpr int kCUDARawHostDelete = 7;
+    static constexpr int kCUDARecordHostAlloc = 8;
+    static constexpr int kCUDAHostAllocEmptyCache = 8;
```
BarclayII, line 221: Two 8's?

src/graph/heterograph.h
```diff
@@ -241,6 +241,9 @@ class HeteroGraph : public BaseHeteroGraph {
    */
   void PinMemory_() override;
 
+  /** @brief Return a new pinned graph managed by backend allocator. */
```
BarclayII, line 244: What does it mean by "backend allocator"?  Could I say "Return a new graph pinned by the backend"?

src/runtime/cpu_device_api.cc
BarclayII: I agree.  This function should never be called and therefore it's better to call `BUG_IF_FAIL`.

src/runtime/cuda/cuda_device_api.cc
```diff
@@ -214,10 +234,33 @@ class CUDADeviceAPI final : public DeviceAPI {
   bool PinData(void* ptr, size_t nbytes) override {
     // prevent users from pinning empty tensors or graphs
     if (ptr == nullptr || nbytes == 0) return false;
+    TensorDispatcher* tensor_dispatcher = TensorDispatcher::Global();
+    // Minimize the pinned memory pool allocated by backend (via tensoradapter)
+    // to preserve enough memory for DGL inherited in-place pin-memory operation
+    if (tensor_dispatcher->IsAvailable()) {
+      tensor_dispatcher->CUDAHostAllocEmptyCache();
```
BarclayII, line 241: The function is named `AllocEmptyCache` but the comment says "Minimize the pinned memory pool".  At a first glance I don't understand the connection.  I need to look into the tensor adapter code to figure out this is actually "empty the cache in host allocator" rather than "allocating an empty cache".  So it's better to change the name to `CUDAHostAllocatorEmptyCache`.

Also, does it mean we are emptying the cache every time a graph object is pinned?  Would it incur additional overhead?

src/runtime/cuda/cuda_device_api.cc
```diff
@@ -214,10 +234,33 @@ class CUDADeviceAPI final : public DeviceAPI {
...
     return true;
   }
 
+  void* AllocPinnedDataSpace(
+      size_t nbytes, void** ctx, void** deleter) override {
+    // prevent pinning empty tensors or graphs
+    if (nbytes == 0) return nullptr;
+    TensorDispatcher* tensor_dispatcher = TensorDispatcher::Global();
+    CHECK(tensor_dispatcher->IsAvailable())
+        << "CachingHost allocator only available with TensorAdapter.";
```
BarclayII, line 253: So `graph.pin_memory()` will only work if tensor adapter is available?  I think a better error message is needed if this is the case, like telling the users to update the PyTorch version to what we support.  @frozenbugs what do you think?

src/runtime/cuda/cuda_device_api.cc
```diff
@@ -214,10 +234,33 @@ class CUDADeviceAPI final : public DeviceAPI {
...
+    return tensor_dispatcher->CUDAAllocHostWorkspace(nbytes, ctx, deleter);
+  }
+
+  void FreePinnedDataSpace(void** deleter) override {
+    TensorDispatcher* tensor_dispatcher = TensorDispatcher::Global();
+    if (tensor_dispatcher->IsAvailable()) {
+      tensor_dispatcher->CUDAFreeHostWorkspace(deleter);
+    }
```
BarclayII, line 261: What will it do if tensor adapter is not available?

src/runtime/ndarray.cc
```diff
@@ -159,7 +164,8 @@ NDArray NDArray::EmptyShared(
     const std::string& name, std::vector<int64_t> shape, DGLDataType dtype,
     DGLContext ctx, bool is_create) {
   NDArray ret = Internal::Create(shape, dtype, ctx);
-  // setup memory content
+  // setup memory content (i.e., allocation), other infos are setup in
+  // Internal::Create
```
BarclayII, line 168: What does "other info" refer to?

tensoradapter/include/tensoradapter.h
```diff
@@ -65,6 +65,12 @@ cudaStream_t CUDACurrentStream();
  * @param device_id Device of the tensor.
  */
 void RecordStream(void* ptr, cudaStream_t stream, int device_id);
+
+void* CUDARawHostAlloc(size_t nbytes, void** ctx, void** raw_deleter);
+void CUDARawHostDelete(void** raw_deleter);
+void CUDARecordHostAlloc(
+    void* data, void* ctx, cudaStream_t stream, int device_id);
+void CUDAHostAllocEmptyCache();
```
BarclayII, line 73: `CUDAHostAllocatorEmptyCache`.

BarclayII (changes requested):

include/dgl/runtime/ndarray.h
chang-l: Yeah, the comment is not clear. Here, I want the if condition to check (1) it is a copy between two diff devices and (2) one of the array (cpu) is pinned by PyTorch. 
Given that the array on gpu can never be pinned (will raise an exception), i.e., `pinned_by_pyt_` always `false` if on gpu, the second constraint can be simplified to `data1.pinned_by_pyt_ != data2.pinned_by_pyt_`.

I agree it compromises the readability a lot. So I updated by (1) splitting the if condition (2) improving the comment.

include/dgl/aten/coo.h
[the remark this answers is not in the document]
chang-l: Sorry. I didn't pay enough attention on docstring when I developed the feature :( 

I fixed the docstring for `PinMemory()` method. 

>Does UnpinMemory_ works for cases pinned by this method?

No. If pinned by PyTorch (`PinMemory`) , we don't need to 'unpin' it, as the pinning memory pool is cached (unless we call `emptycache` explicitly). Instead, the pinned memory is 'freed' (actually just flushed data) by `FreePinnedDataSpace` call

include/dgl/runtime/tensordispatch.h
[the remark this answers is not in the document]
chang-l: How about organizing all cuda-related functions into the preprocessor? This is also consistent with `tensoradapter/torch.cpp`

src/runtime/ndarray.cc
chang-l: Sorry for the ambiguity.  I mean other infos like strides, shape, etc, have been set up in Internal::Create already, so only the size is assigned below. 
Updated.

src/runtime/cuda/cuda_device_api.cc
chang-l: It should raise an error... I updated this function `FreePinnedDataSpace` to be consistent with `AllocPinnedDataSpace`.

src/runtime/cuda/cuda_device_api.cc
chang-l: >So graph.pin_memory() will only work if tensor adapter is available? 

Yes.

I updated the msg to `CachingHostAllocator is not available in the current backend PyTorch. Please update the PyTorch version to 1.11+`.
We can circulate from here.

src/runtime/cuda/cuda_device_api.cc
chang-l: >So it's better to change the name to CUDAHostAllocatorEmptyCache.

Thanks. I updated accordingly.

> Also, does it mean we are emptying the cache every time a graph object is pinned? Would it incur additional overhead?

Yes, we need to empty the cache using cudaHostFree() when a graph is pinned by DGL, so it would add extra overhead. I think it is fine.  In DGL, pinning/unpinning a graph via `cudaHostRegister`/`cudaHostUnRegister` is very expensive, so we normally do it once for the entire graph structure and feature storage at the beginning. 
Compared to the cost of pinning the entire graph, the overhead of empty cache should be small. And `cudaHostRegister` should not be invoked often, otherwise, we should cache it via `CUDARawHostAlloc`.  It is also necessary to do so, as we need to isolate two pinned pools and reserve enough space for `cudaHostRegister`, in preparation for the entire graph/feature.

src/graph/heterograph.h
chang-l: Thanks. I improved this part of docstring.

include/dgl/runtime/tensordispatch.h
chang-l: Sorry for my typo. This is scary at first glance... I guess `reinterpret_cast` 'saved' me to make it work even they shared the same entry point.
Thanks, I fixed it.

src/runtime/cpu_device_api.cc
chang-l: Updated.

include/dgl/aten/coo.h
```diff
@@ -56,14 +56,16 @@ struct COOMatrix {
   /** @brief constructor */
   COOMatrix(
```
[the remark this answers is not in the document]
chang-l: @nv-dlasalle   Thanks for the suggestion. I updated this part accordingly. 
@frozenbugs  No, this is not checked in `CheckValidity`. Currently, all three arrays need to be pinned, otherwise, new pinned COO/CSR matrix will be allocated.

include/dgl/runtime/ndarray.h
chang-l: Yeah.. it makes sense and it's more like pytorch's passing tensor options to `empty` function. 
Here, since we may only have one option `pin_memory` and other empty-like function (`empty_shared`) is also separated from `empty`, maybe it's better to put it in another PR, in case we have more `empty` like functions in future.

chang-l:
@dgl-bot

src/graph/heterograph.cc
```diff
@@ -277,6 +277,24 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DGLContext& ctx) {
       hgindex->meta_graph_, rel_graphs, hgindex->num_verts_per_type_));
 }
 
+HeteroGraphPtr HeteroGraph::PinMemory(HeteroGraphPtr g) {
+  bool allpinned = true;
+  std::vector<HeteroGraphPtr> rel_graphs;
+  auto hgindex = std::dynamic_pointer_cast<HeteroGraph>(g);
+  CHECK_NOTNULL(hgindex);
+  for (auto subg : hgindex->relation_graphs_) {
```
frozenbugs, line 285: I'd prefer to have 2 loops here. 
The first just check whether all pined.
If not, loop again and push.
The loop should be way cheaper than push_back ops.

frozenbugs:
Overall the PR is still a little bit oversize which made it very hard to review, in the future, we should try break it up to smaller PRs.

For example, for this PR, we can break it out like creating RecordedCopyDataFromTo in the first PR, and then implement PinMemory for NSArray.

src/runtime/cuda/cuda_device_api.cc
chang-l: >If this is to be used for pinning sampled subgraphs, would it kind of cancel the benefit of asynchronous transfer of subgraph structures?

When pinning sampled subgraphs, we should never call this function `PinData`, which uses `cudaHostRegister`. Instead, for small subgraphs (pinning/destroying frequently), we should use `CUDARawHostAlloc` imported by this PR.

To sum up, 
(1) in UVA mode, we pin the entire graph and feature ONCE with DGL's `PinData()` function via `dgl.graph.pin_memory_()`. Before pinning, we clean the backend's pinned mem pool by `CUDAHostAllocatorEmptyCache`.
(2) in non-UVA mode, we can pin the subgraph for each batch with a different function `AllocPinnedDataSpace()` via `dgl.graph._graph.pin_memory()` and it shares the same pinned memory pool managed by PyTorch. It does not need to empty cache 'manually' by DGL, but handled by PyTorch.

python/dgl/heterograph_index.py
```diff
@@ -254,6 +254,21 @@ def copy_to(self, ctx):
         """
         return _CAPI_DGLHeteroCopyTo(self, ctx.device_type, ctx.device_id)
 
+    def pin_memory(self):
```
frozenbugs, line 257: Please update the python doc of pin_memory and pin_memory_, and also the c++ documentation for NDArray.

src/runtime/cuda/cuda_device_api.cc
BarclayII: I see.  Because `PinData` is pinning the data in-place with `CUDAHostRegister`, we have to empty the cache before pinning, and `CUDAHostRegister` is not called often anyway.

I think we should rename this `PinData` function as it is now ambiguous: the name could refer to both in-place pinning and out-of-place pinning.  Something more explicit like `PinDataInPlace` or `RegisterHostMemory` should be good.

include/dgl/aten/coo.h
```diff
@@ -56,16 +56,21 @@ struct COOMatrix {
...
         num_cols(ncols),
         row(rarr),
         col(carr),
         data(darr),
         row_sorted(rsorted),
-        col_sorted(csorted),
-        is_pinned(pinned_mem) {
+        col_sorted(csorted) {
+    is_pinned = aten::IsNullArray(row) ? true : row.IsPinned();
+    if (!aten::IsNullArray(col)) {
```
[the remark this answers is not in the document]
BarclayII: I think this logic is just checking whether all `row`, `col`, `data` are either null arrays or pinned arrays.  So I think this is clearer:
```C++
is_pinned = (aten::IsNullArray(row) || row.IsPinned()) && \
            (aten::IsNullArray(col) || col.IsPinned()) && \
            (aten::IsNullArray(data) || data.IsPinned());
```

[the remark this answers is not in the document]
BarclayII: I think this logic is just checking whether all `row`, `col`, `data` are either null arrays or pinned arrays.  So I think this is more straightforward:
```C++
is_pinned = (aten::IsNullArray(row) || row.IsPinned()) && \
            (aten::IsNullArray(col) || col.IsPinned()) && \
            (aten::IsNullArray(data) || data.IsPinned());
```

include/dgl/aten/csr.h
```diff
@@ -60,6 +60,13 @@ struct CSRMatrix {
         indices(iarr),
         data(darr),
         sorted(sorted_flag) {
+    is_pinned = aten::IsNullArray(indptr) ? true : indptr.IsPinned();
+    if (!aten::IsNullArray(indices)) {
+      is_pinned = is_pinned && indices.IsPinned();
+    }
+    if (!aten::IsNullArray(data)) {
+      is_pinned = is_pinned && data.IsPinned();
+    }
```
BarclayII, line 69: Same.

include/dgl/runtime/ndarray.h
```diff
@@ -174,6 +174,13 @@ class NDArray {
    * @brief Return a new array with a copy of the content.
    */
   inline NDArray Clone() const;
+
+  /**
+   * @brief Return a copy of the NDArray in pinned (page-locked) memory.
+   * @note This is an out-place method, and the current context has to be
```
BarclayII, line 180: `out-of-place`

include/dgl/runtime/ndarray.h
```diff
@@ -229,6 +236,16 @@ class NDArray {
    */
   DGL_DLL static NDArray Empty(
       std::vector<int64_t> shape, DGLDataType dtype, DGLContext ctx);
+
+  /**
+   * @brief Create an empty NDArray in pinned memory.
+   * @param shape The shape of the new array.
+   * @param dtype The data type of the new array.
+   * @param ctx The context of the Array.
+   * @return The created Array
```
BarclayII, line 245: Capitalized `Array`?

src/graph/unit_graph.cc
```diff
@@ -140,7 +140,8 @@ class UnitGraph::COO : public BaseHeteroGraph {
     return COO(meta_graph_, adj_.CopyTo(ctx));
   }
 
-  COO PinMemoryOutPlace() {
+  /** @brief Copy the adj_ to pinned memory: COOMatrix of the COO graph. */
```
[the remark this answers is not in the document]
BarclayII: I think the term `COOMatrix` should be fine here.  `adjacency matrix` could also imply CSR.

src/runtime/ndarray.cc
```diff
@@ -159,7 +167,7 @@ NDArray NDArray::EmptyShared(
     const std::string& name, std::vector<int64_t> shape, DGLDataType dtype,
     DGLContext ctx, bool is_create) {
   NDArray ret = Internal::Create(shape, dtype, ctx);
-  // setup memory content
+  // setup memory content (i.e., allocation)
```
BarclayII, line 170: Why not just say "allocate memory"?

src/runtime/ndarray.cc
```diff
@@ -175,7 +183,7 @@ NDArray NDArray::EmptyShared(
 NDArray NDArray::Empty(
     std::vector<int64_t> shape, DGLDataType dtype, DGLContext ctx) {
   NDArray ret = Internal::Create(shape, dtype, ctx);
-  // setup memory content
+  // setup memory content (i.e., allocation)
```
BarclayII, line 186: Ditto

src/runtime/ndarray.cc
```diff
@@ -206,6 +214,41 @@ void NDArray::CopyFromTo(DGLArray* from, DGLArray* to) {
...
 }
 
+void NDArray::RecordedCopyFromTo(DGLArray* from, DGLArray* to, void* pyt_ctx) {
+  size_t from_size = GetDataSize(*from);
+  size_t to_size = GetDataSize(*to);
+  CHECK_EQ(from_size, to_size)
+      << "DGLArrayCopyFromTo: The size must exactly match.";
+
+  CHECK(from->ctx.device_type != to->ctx.device_type)
+      << "Recoding event is only called for the copy between CPU and GPU";
```
BarclayII, line 224: End with dot.

src/runtime/ndarray.cc
```diff
@@ -206,6 +214,41 @@ void NDArray::CopyFromTo(DGLArray* from, DGLArray* to) {
...
+  CHECK_EQ(ctx.device_type, kDGLCPU) << "Only NDArray on CPU can be pinned";
+  NDArray ret = Internal::Create(shape, dtype, ctx);
+  // setup memory content (i.e., allocation)
+  size_t size = GetDataSize(ret.data_->dl_tensor);
+  if (size > 0) {
+    ret.data_->dl_tensor.data = DeviceAPI::Get(kDGLCUDA)->AllocPinnedDataSpace(
+        size, &(ret.data_->pyt_ctx), &(ret.data_->pyt_raw_deleter));
+    CHECK(
+        ret.data_->pyt_ctx != nullptr && ret.data_->pyt_raw_deleter != nullptr)
+        << "Failed to allocate memory from PyTorch's CUDACachingAllocator.";
```
BarclayII, line 246: Is it possible to display the actual reason of the failure here?  Something like error code?

include/dgl/aten/coo.h
```diff
@@ -133,6 +140,18 @@ struct COOMatrix {
...
 
+  /** @brief Return a copy of this matrix in pinned (page-locked) memory. */
+  inline COOMatrix PinMemory() {
+    if (is_pinned) return *this;
+    auto new_coo = COOMatrix(
+        num_rows, num_cols, row.PinMemory(), col.PinMemory(),
+        aten::IsNullArray(data) ? data : data.PinMemory(), row_sorted,
+        col_sorted);
+    CHECK(new_coo.is_pinned)
+        << "The new allocated COOMatrix is not pinned correctly";
```
BarclayII, line 151: Is it possible to display the reason why it fails?

include/dgl/aten/csr.h
```diff
@@ -126,6 +133,17 @@ struct CSRMatrix {
...
   }
 
+  /** @brief Return a copy of this matrix in pinned (page-locked) memory. */
+  inline CSRMatrix PinMemory() {
+    if (is_pinned) return *this;
+    auto new_csr = CSRMatrix(
+        num_rows, num_cols, indptr.PinMemory(), indices.PinMemory(),
+        aten::IsNullArray(data) ? data : data.PinMemory(), sorted);
+    CHECK(new_csr.is_pinned)
+        << "The new allocated CSRMatrix is not pinned correctly";
```
BarclayII, line 143: Same.

src/runtime/ndarray.cc
```diff
@@ -68,8 +68,16 @@ void NDArray::Internal::DefaultDeleter(NDArray::Container* ptr) {
   } else if (ptr->dl_tensor.data != nullptr) {
     // if the array is still pinned before freeing, unpin it.
     if (ptr->pinned_by_dgl_) UnpinContainer(ptr);
-    dgl::runtime::DeviceAPI::Get(ptr->dl_tensor.ctx)
-        ->FreeDataSpace(ptr->dl_tensor.ctx, ptr->dl_tensor.data);
+    if (ptr->pinned_by_pyt_) {
+      DeviceAPI::Get(kDGLCUDA)->FreePinnedDataSpace(&(ptr->pyt_raw_deleter));
+      CHECK(ptr->pyt_raw_deleter == nullptr);
```
BarclayII, line 73: Just to make sure: this is an assertion, right?

include/dgl/aten/coo.h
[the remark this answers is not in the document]
frozenbugs: this is a little bit error-prone.
Some c++ compiler may executing the logic operator from right to left. In this case if row is null, row.IsPinned() will throw exception.

include/dgl/runtime/ndarray.h
```diff
@@ -464,16 +491,22 @@ inline void NDArray::CopyFrom(DGLArray* other) {
...
-          other.data_->dl_tensor.ctx.device_type) {
+  // copy between two devices
+  if (data_->dl_tensor.ctx.device_type !=
+      other.data_->dl_tensor.ctx.device_type) {
     CHECK(data_ != nullptr);
     auto to_ctx_type = data_->dl_tensor.ctx.device_type;
-    auto ptr = (to_ctx_type == kDGLCPU ? data_ : other.data_);
-    void* pyt_ctx = ptr->pyt_ctx;  // get the correct ctx
-    RecordedCopyFromTo(&(other.data_->dl_tensor), &(data_->dl_tensor), pyt_ctx);
+    auto cpu_data_ = (to_ctx_type == kDGLCPU ? data_ : other.data_);
```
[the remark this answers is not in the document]
chang-l: It's actually necessary to have a `CopyFrom` when `cpu_data->pinned_by_pyt_` is false, falling back to normal copy operation without recording.

include/dgl/runtime/device_api.h
```diff
@@ -160,6 +189,28 @@ class DeviceAPI {
    */
   DGL_DLL virtual void UnpinData(void* ptr);
 
+  /**
+   * @brief Pin host memory using PyTorch CachingHostAllocator.
```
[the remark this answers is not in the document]
chang-l: Sure. It allocates pinned memory. I updated the comment.

include/dgl/aten/coo.h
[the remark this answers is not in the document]
chang-l: @frozenbugs  Unlike bit-wise operators, I think `&&` and `||` are guaranteed to evaluate from left to right, according to C++ standards. See this thread https://stackoverflow.com/a/5683030

include/dgl/runtime/tensordispatch.h
```diff
@@ -134,6 +134,66 @@ class TensorDispatcher {
     auto entry = entrypoints_[Op::kCUDACurrentStream];
     return FUNCCAST(tensoradapter::CUDACurrentStream, entry)();
   }
+
+  /**
+   * @brief Allocate a piece of pinned CPU memory via PyTorch
+   * CachingHostAllocator Used in CUDADeviceAPI::AllocPinnedDataSpace().
```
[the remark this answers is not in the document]
chang-l: > If this is the only place we need to function, why not define a local function?

I guess this is for putting all imported tensoradaptor functions in the same dispatcher/header file?

> We don't need to call out where the method is used in the c++ doc. 

Okay, maybe I can put it in `@note`.

include/dgl/runtime/tensordispatch.h
```diff
@@ -134,6 +134,66 @@ class TensorDispatcher {
     auto entry = entrypoints_[Op::kCUDACurrentStream];
     return FUNCCAST(tensoradapter::CUDACurrentStream, entry)();
   }
+
```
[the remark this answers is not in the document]
chang-l: > Maybe the easier way is to create a google doc to illustrate the diagram of the logic, 

In fact, the implementation logic of `CachingHostAllocator` is very detailed in the comment of pytorch src code. I think we don't need to create a doc to rephrase their design...

I tried to describe the main idea behind it briefly (maybe not brief enough 😞) and hope it can make future developers' life easier when dealing with this allocator.

Will slightly update again.

include/dgl/aten/coo.h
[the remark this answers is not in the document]
frozenbugs: I see, thx!

src/runtime/ndarray.cc
chang-l: lol... I guess I was trying to be consistent with other 'setup's in `Internal::Create` function. Updated.

src/runtime/ndarray.cc
chang-l: I didn't find such info at pytorch side. I can add more info in this err msg, by printing the two ptrs, sth like 
```
<< "The allocation failed in PyTorch's CachingHostAllocator. "
<< "The returned context pointer is " << pyt_ctx_ 
<< " and the function deleter is " << pyt_raw_deleter_;

File: include/dgl/aten/coo.h (hunk 59 above)
Comment (chang-l) [reply include/dgl/aten/coo.h:151]: I think it can only fail when `data.PinMemory().is_pinned()` is false  (or `row`, `col`). Here, unless I scan every ndarray to print out the one (ones) that fails, it can hardly print further useful info for developers and users...

File: src/runtime/ndarray.cc (hunk 61 above)
Comment (chang-l) [reply src/runtime/ndarray.cc:73]: Yes, it is an assertion, as I assign `pyt_raw_deleter` to nullptr when the deleter is called (see `torch.cpp` L85).

File: include/dgl/runtime/ndarray.h [hunk 66]
```diff
@@ -174,6 +174,13 @@ class NDArray {
    * @brief Return a new array with a copy of the content.
    */
   inline NDArray Clone() const;
+
+  /**
+   * @brief Return a copy of the NDArray in pinned (page-locked) memory.
+   * @note This is an out-place method, and the current context has to be
+   * kDGLCPU. Otherwise, it will throw an error.
+   */
+  inline NDArray PinMemory();
```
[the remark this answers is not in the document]
chang-l: `PinMemory_` is DGL in-house pinning strategy, which only uses `cudaHostRegister`. 
`PinMemory` utilizes PyTorch's pinning memory pool (cached), to avoid overhead of frequently pinning/unpin memory.

python/dgl/heterograph_index.py
chang-l: Updated.

chang-l:
Sorry for the delay. Ready for the next round @frozenbugs @BarclayII @yaox12

src/graph/heterograph.cc
```diff
@@ -277,6 +277,29 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DGLContext& ctx) {
       hgindex->meta_graph_, rel_graphs, hgindex->num_verts_per_type_));
 }
 
+HeteroGraphPtr HeteroGraph::PinMemory(HeteroGraphPtr g) {
+  bool allpinned = true;
```
frozenbugs, line 281: this local variable is not used.

src/graph/heterograph.cc
```diff
@@ -277,6 +277,29 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DGLContext& ctx) {
...
+  auto hgindex = std::dynamic_pointer_cast<HeteroGraph>(g);
```
frozenbugs, line 282: hgindex -> casted_ptr

src/graph/heterograph.cc
```diff
@@ -277,6 +277,29 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DGLContext& ctx) {
...
+  CHECK_NOTNULL(hgindex);
+  auto rel_g = hgindex->relation_graphs_;
```
frozenbugs, line 284: let's avoid unconventional abbr.

Just use relation_graphs.

src/graph/heterograph.cc
```diff
@@ -277,6 +277,29 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DGLContext& ctx) {
...
+
+  auto it = std::find_if_not(
+      rel_g.begin(), rel_g.end(), [](auto& subg) { return subg->IsPinned(); });
```
frozenbugs, line 287: Just to check are all graphs in relation_graphs_ subgraph? I am asking to make sure the subg is a proper variable name.

src/graph/heterograph.cc
```diff
@@ -277,6 +277,29 @@ HeteroGraphPtr HeteroGraph::CopyTo(HeteroGraphPtr g, const DGLContext& ctx) {
...
+  // all pinned, return itself.
```
frozenbugs, line 288: Every comment should be a sentence.
`// All underlying relation graphs are pinned, return the input hetero-graph directly.`

src/graph/heterograph.cc
frozenbugs: I see a comment mentions "underlying relation graphs", I'd suggest to make it consistent.

subg -> underlying_graph

src/graph/heterograph.h
```diff
@@ -249,6 +249,15 @@ class HeteroGraph : public BaseHeteroGraph {
    */
   void UnpinMemory_();
 
+  /**
+   * @brief Copy the current graph to pinned memory managed by
+   *     PyTorch CachingHostAllocator for each relation graph.
+   * @note If any of the underlying relation graphs are already pinned, the
+   *     function simply uses its original copy. If all pinned, it returns
+   *     itself.
```
frozenbugs, line 257: nit: gramma issue. it returns itself, the it means the function, but the return is the original graph.

from ChatGPT with manual tweak:
If any of the underlying relation graphs are already pinned, the function will utilize their existing copies. If all of them are pinned, the function will return the original input hetero-graph directly.

src/graph/unit_graph.cc
```diff
@@ -1259,6 +1277,37 @@ HeteroGraphPtr UnitGraph::CopyTo(HeteroGraphPtr g, const DGLContext& ctx) {
   }
 }
 
+HeteroGraphPtr UnitGraph::PinMemory() {
+  CSRPtr pinned_in_csr, pinned_out_csr;
+  COOPtr pinned_coo;
+  if (this->in_csr_->defined() && !this->in_csr_->IsPinned()) {
```
frozenbugs, line 1283: nit, let's avoid **if (not ...) {...} else {...}**, flip it if possible.
which means: swap condition 1 and 2.

if (this->in_csr_->defined() && this->in_csr_->IsPinned()) {
} else if (this->in_csr_->defined()) {
} else {
}

same for the others.

src/runtime/ndarray.cc
```diff
@@ -68,8 +68,16 @@ void NDArray::Internal::DefaultDeleter(NDArray::Container* ptr) {
   } else if (ptr->dl_tensor.data != nullptr) {
     // if the array is still pinned before freeing, unpin it.
     if (ptr->pinned_by_dgl_) UnpinContainer(ptr);
-    dgl::runtime::DeviceAPI::Get(ptr->dl_tensor.ctx)
-        ->FreeDataSpace(ptr->dl_tensor.ctx, ptr->dl_tensor.data);
+    if (ptr->pinned_by_pyt_) {
+      DeviceAPI::Get(kDGLCUDA)->FreePinnedDataSpace(&(ptr->pyt_raw_deleter_));
+      CHECK(ptr->pyt_raw_deleter_ == nullptr);
+      // reset the flag and storage ctx ptr for completeness
```
frozenbugs, line 74: ditto.

src/runtime/ndarray.cc
```diff
@@ -206,6 +214,44 @@ void NDArray::CopyFromTo(DGLArray* from, DGLArray* to) {
...
+      from->data, static_cast<size_t>(from->byte_offset), to->data,
+      static_cast<size_t>(to->byte_offset), from_size, from->ctx, to->ctx,
+      from->dtype, pyt_ctx);
+}
+
+NDArray NDArray::PinnedEmpty(
+    std::vector<int64_t> shape, DGLDataType dtype, DGLContext ctx) {
+  CHECK_EQ(ctx.device_type, kDGLCPU) << "Only NDArray on CPU can be pinned";
+  NDArray ret = Internal::Create(shape, dtype, ctx);
+  // setup memory content (i.e., allocation)
```
frozenbugs, line 239: ditto.

tensoradapter/pytorch/torch.cpp
```diff
@@ -55,6 +56,50 @@ TA_EXPORTS void RecordStream(void* ptr, cudaStream_t stream, int device_id) {
...
+};
+
+TA_EXPORTS void* CUDARawHostAlloc(
+    size_t nbytes, void** ctx, void** raw_deleter) {
+  auto data_ptr = at::cuda::getCachingHostAllocator()->allocate(nbytes);
+  auto raw = data_ptr.get();
+  // Return the raw ctx ptr for recording event.
+  *ctx = data_ptr.get_context();
+
+  // Transfer ownership to raw_deleter
```
frozenbugs, line 76: super nit, end with dot.

src/runtime/ndarray.cc
BarclayII: I see, you can leave it.

include/dgl/aten/coo.h
BarclayII: OK NVM.

include/dgl/runtime/ndarray.h
```diff
@@ -454,7 +512,25 @@ inline void NDArray::CopyFrom(DGLArray* other) {
...
+      // cudaMemcpyAsync as long as the memory is pinned by PyTorch.
+      void* pyt_ctx = cpu_data->pyt_ctx_;
+      RecordedCopyFromTo(
+          &(other.data_->dl_tensor), &(data_->dl_tensor), pyt_ctx);
+    } else {
+      CopyFrom(&(other.data_->dl_tensor));
+    }
+  } else {
+    CopyFrom(&(other.data_->dl_tensor));
+  }
```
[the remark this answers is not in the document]
chang-l: Thanks. Updated.

include/dgl/runtime/ndarray.h
```diff
@@ -483,6 +577,17 @@ inline NDArray NDArray::Clone() const {
   return this->CopyTo(dptr->ctx);
 }
 
+inline NDArray NDArray::PinMemory() {
+  CHECK(data_ != nullptr);
+  const DGLArray* dptr = operator->();
```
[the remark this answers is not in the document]
chang-l: Okay.

include/dgl/runtime/tensordispatch.h
```diff
@@ -134,6 +134,69 @@ class TensorDispatcher {
...
+  inline void CUDAFreeHostWorkspace(void** deleter) {
+    auto entry = entrypoints_[Op::kCUDARawHostDelete];
+    FUNCCAST(tensoradapter::CUDARawHostDelete, entry)(deleter);
+  }
+
+  /**
+   * @brief Invoke the record_event function call from PyTorch
+   *     CachingHostAllocator.
+   * @note This function assoicates a CUDA stream (used by a copy kernel) to the
+   *     pinned data. In the free path of this data (with
```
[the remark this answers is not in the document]
chang-l: By `free path of this data`, I mean, calling the 'deallocator' within `CUDAFreeHostWorkspace`. It does not necessarily mean free the memory though, as it also flushes the data for future other usage.

python/dgl/heterograph_index.py
```diff
@@ -254,10 +254,30 @@ def copy_to(self, ctx):
...
+        """Copies the graph structure to pinned memory, if it's not already pinned.
+
+        NOTE: This is similar to PyTorch's Tensor.pin_memory() method but customized for
+              graph structure.
+              Since it shares the pin_memory allocator with PyTorch, the lifecycle of
+              the returned graph structure is also managed by PyTorch.
+              Note that if a batch includes DGL graph structure object (HeteroGraphIndex),
+              PyTorch's DataLoader memory pinning logic will recognize this function and
+              automatically invoke this function when pin_memory=True.
```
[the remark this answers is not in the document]
chang-l: I see. Thanks.

src/graph/heterograph.cc
chang-l: >Just to check are all graphs in relation_graphs_ subgraph?

Yes.
> I'd suggest to make it consistent.

Okay.

src/graph/heterograph.h
chang-l: Updated.

frozenbugs (commented):
LGTM, overall this is a giant PR which we should try to avoid if we can.

frozenbugs (approved):

BarclayII (approved):
I'm good.  Thanks a lot!

PR #5376, opened 2023-02-23
[Utils] Node homophily measure

mufeili:
## Description
I'll implement the rest two measures once everyone is comfortable with the practice in this PR.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/homophily.py
+++ b/python/dgl/homophily.py
@@ -1,0 +1,56 @@
+"""Utils for tacking graph homophily and heterophily"""
+from . import backend as F
+from . import function as fn
+
+__all__ = ['node_homophily']
+
+def node_homophily(graph, y):
+    """Homophily measure from `Geom-GCN: Geometric Graph Convolutional Networks
+    <https://arxiv.org/abs/2002.05287>`__
+
+    We follow the practice of a later paper `Large Scale Learning on
+    Non-Homophilous Graphs: New Benchmarks and Strong Simple Methods
+    <https://arxiv.org/abs/2110.14446>`__ to call it node homophily.
+
+    Mathematically it is defined as follows:
+
+    .. math::
+      \frac{1}{|\mathcal{V}|} \sum_{v \in \mathcal{V}} \frac{ | \{ (u,v) : u
+      \in \mathcal{N}(v) \wedge y_v = y_u \} |  } { |\mathcal{N}(v)| }
+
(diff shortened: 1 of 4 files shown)


python/dgl/homophily.py
```diff
@@ -0,0 +1,57 @@
+"""Utils for tacking graph homophily and heterophily"""
+from . import backend as F, function as fn
+
+__all__ = ["node_homophily"]
+
+
+def node_homophily(graph, y):
+    r"""Homophily measure from `Geom-GCN: Geometric Graph Convolutional Networks
```
frozenbugs, line 8: measurement

python/dgl/homophily.py
```diff
@@ -0,0 +1,57 @@
...
+    <https://arxiv.org/abs/2002.05287>`__
+
+    We follow the practice of a later paper `Large Scale Learning on
```
frozenbugs, line 11: avoid using `later`, `new` in code, the code will be here for long, but the paper will not be new any more :)

tests/python/common/test_homophily.py
```diff
@@ -0,0 +1,19 @@
...
+import backend as F
+
+import dgl
+from test_utils import parametrize_idtype
+
+
+@unittest.skipIf(dgl.backend.backend_name == "tensorflow", reason="Skip TF")
+@parametrize_idtype
+def test_node_homophily(idtype):
+    # NOTE: If you want to update this test case, remember to update the
```
frozenbugs, line 12: If we want to introduce this practice, let's follow this format:

```
# IfChangeThenChange: python/dgl/homophily.py
# Update the docstring example.
```

frozenbugs: Added in the conventions:
https://docs.google.com/document/d/1NGXPg4U67ZpMDVxxkb-SprEx11d8zqNB_0cLweTL90Y/edit#

frozenbugs (approved):

python/dgl/homophily.py
mufeili: Measure is the term used in this kind of papers: https://arxiv.org/abs/2110.14446

python/dgl/homophily.py
mufeili: It emphasizes the fact that the term was not introduced in the original paper.

tests/python/common/test_homophily.py
mufeili: updated

PR #5382, opened 2023-02-24
[Utils] Edge and LINKX homophily measure

mufeili:
## Checklist
- [x] Code is well-documented


python/dgl/homophily.py
```diff
@@ -52,6 +58,139 @@ def node_homophily(graph, y):
...
+    >>> graph = dgl.graph(([0, 1, 2, 3], [1, 2, 0, 4]))
+    >>> y = torch.tensor([0, 0, 0, 0, 1])
+    >>> dgl.linkx_homophily(graph, y)
+    0.19999998807907104
+    """
+    with graph.local_scope():
+        # Compute |{u\in N(v): y_v = y_u}| for each node v.
+        src, dst = graph.edges()
+        # Handle the case where graph is of dtype int32.
+        src = src.long()
```
frozenbugs, line 155: why not?
```
src = F.astype(src, F.int64)
```
I guess they are equivalent on PyTorch?

Create a local method? 
```
def edges_astype(graph)
        src, dst = graph.edges()
        # Handle the case where graph is of dtype int32.
        src = F.astype(src, F.int64)
        dst = F.astype(dst, F.int64)
        return src, dst
```

Or better name?

python/dgl/homophily.py
```diff
@@ -52,6 +58,139 @@ def node_homophily(graph, y):
...
+
+    Mathematically it is defined as follows:
+
+    .. math::
+      \frac{1}{C-1} \sum_{k=1}^{C} \max \left(0, \frac{\sum_{v\in C_k}|\{u\in
+      \mathcal{N}(v): y_v = y_u \}|}{\sum_{v\in C_k}|\mathcal{N}(v)|} -
+      \frac{|\mathcal{C}_k|}{|\mathcal{V}|} \right)
+
+    where :math:`C` is the number of node classes, :math:`C_k` is the set of
+    nodes that belong to class k, :math:`\mathcal{N}(v)` is the predecessors of
```
frozenbugs, line 125: super nit: `is the predecessor` or `are the predecessors`

frozenbugs (approved):

python/dgl/homophily.py
mufeili: For the first question, you are right. The reason for the alternative is that this method will only work for PyTorch and therefore I turned to more PyTorch-native syntax in case users want to check and understand the implementation.

mufeili: I've addressed the second question.

python/dgl/homophily.py
mufeili: resolved

python/dgl/homophily.py
```diff
@@ -1,12 +1,25 @@
...
-__all__ = ["node_homophily"]
+try:
+    import torch
+except ImportError:
+    pass
+
+__all__ = ["node_homophily", "edge_homophily", "linkx_homophily"]
+
+
+def get_long_edges(graph):
```
frozenbugs, line 13: should internal helper function start with _?

mufeili: I'm fine either way. For this file, it should be clear that only the functions included in `__all__` are external.

python/dgl/homophily.py
```diff
@@ -45,13 +59,140 @@ def node_homophily(graph, y):
...
+        class_graph.ndata["same_class_deg"] = torch.cat(
+            [graph.ndata["same_class_deg"], class_placeholder], dim=0
+        )
+        class_graph.update_all(
+            fn.copy_u("same_class_deg", "m"), fn.sum("m", "class_deg_aggr")
+        )
+
+        # Similarly, compute \sum_{v\in C_k}|N(v)| for all k in parallel.
+        class_graph.ndata["deg"] = torch.cat([deg, class_placeholder], dim=0)
+        class_graph.update_all(fn.copy_u("deg", "m"), fn.sum("m", "deg_aggr"))
```
jermainewang, line 187: I feel the implementation is over complicated and creating a new graph maybe costly too when the input graph is large. Will the following be simpler/more-readable?

1. Compute the node heterophily without averaging it across all the nodes.
2. Have a loop over the number of classes and each iteration slices the nodes of one class and compute the class-wise heterophily rate.
3. Aggregate across all the classes.

As the number of classes is usually small, the loop won't be too slow.

tests/python/common/test_homophily.py
```diff
@@ -17,3 +17,29 @@ def test_node_homophily(idtype):
...
+    dgl.backend.backend_name != "pytorch", reason="Only support PyTorch for now"
+)
+@parametrize_idtype
+def test_linkx_homophily(idtype):
+    # IfChangeThenChange: python/dgl/homophily.py
+    # Update the docstring example.
+    device = F.ctx()
+    graph = dgl.graph(([0, 1, 2, 3], [1, 2, 0, 4]), device=device)
+    y = F.tensor([0, 0, 0, 0, 1])
+    assert dgl.linkx_homophily(graph, y) == 0.19999998807907104
```
jermainewang, line 45: Please do not use == for float value comparison. Use `math.isclose`.

jermainewang (changes requested):

python/dgl/homophily.py
```diff
@@ -15,8 +29,8 @@ def node_homophily(graph, y):
     Mathematically it is defined as follows:
 
     .. math::
-      \frac{1}{|\mathcal{V}|} \sum_{v \in \mathcal{V}} \frac{ | \{ (u,v) : u
-      \in \mathcal{N}(v) \wedge y_v = y_u \} |  } { |\mathcal{N}(v)| }
+      \frac{1}{|\mathcal{V}|} \sum_{v \in \mathcal{V}} \frac{ | \{u
+      \in \mathcal{N}(v): y_v = y_u \} |  } { |\mathcal{N}(v)| }
```
rudongyu, line 33: A comma at the end of the equation is required to connect the following clause (the same for the other 2 functions).

python/dgl/homophily.py
```diff
@@ -45,13 +59,140 @@ def node_homophily(graph, y):
...
+    0.19999998807907104
+    """
+    with graph.local_scope():
+        # Compute |{u\in N(v): y_v = y_u}| for each node v.
+        # Handle the case where graph is of dtype int32.
+        src, dst = get_long_edges(graph)
+        # Compute y_v = y_u for all edges.
+        graph.edata["same_class"] = (y[src] == y[dst]).float()
+        graph.update_all(
+            fn.copy_e("same_class", "m"), fn.mean("m", "same_class_deg")
```
rudongyu, line 162: The aggregator here should be 'sum', right?

python/dgl/homophily.py
```diff
@@ -1,12 +1,26 @@
...
+    pass
+
+__all__ = ["node_homophily", "edge_homophily", "linkx_homophily"]
+
+
+def get_long_edges(graph):
+    """Internal function for getting the edges of a graph as long tensors"""
+    src, dst = graph.edges()
+    src = F.astype(src, F.int64)
+    dst = F.astype(dst, F.int64)
```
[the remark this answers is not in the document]
mufeili: changed to only support PyTorch and use PyTorch-native APIs

python/dgl/homophily.py
mufeili: You are right.

python/dgl/homophily.py
mufeili: simplified

python/dgl/homophily.py
frozenbugs: Sound good.

nit: Maybe rename to get_edges_long, more natural.

python/dgl/homophily.py
```diff
@@ -45,13 +57,121 @@ def node_homophily(graph, y):
...
+    where :math:`C` is the number of node classes, :math:`C_k` is the set of
+    nodes that belong to class k, :math:`\mathcal{N}(v)` are the predecessors
+    of node :math:`v`, :math:`y_v` is the class of node :math:`v`, and
+    :math:`\mathcal{V}` is the set of nodes.
+
+    Parameters
+    ----------
+    graph : DGLGraph
+        The graph.
+    y : Tensor
```
frozenbugs, line 135: torch.Tensor

and others.

python/dgl/homophily.py
```diff
@@ -45,13 +57,121 @@ def node_homophily(graph, y):
...
+        )
+
+        deg = graph.in_degrees().float()
+        num_nodes = graph.num_nodes()
+        value = 0
+        num_classes = y.max(dim=0).values.item() + 1
+
+        for k in range(num_classes):
+            # Get the nodes that belong to class k.
+            class_mask = y == k
```
frozenbugs, line 170: nit: class_mask = (y == k)

tests/python/common/test_homophily.py
```diff
@@ -16,4 +19,32 @@ def test_node_homophily(idtype):
...
+    )
+    y = F.tensor([0, 0, 0, 0, 1])
+    assert math.isclose(dgl.edge_homophily(graph, y), 0.75)
+
+
+@unittest.skipIf(
+    dgl.backend.backend_name != "pytorch", reason="Only support PyTorch for now"
+)
+@parametrize_idtype
+def test_linkx_homophily(idtype):
```
frozenbugs, line 44: Is there any conner case you need to handle?
e.g. there was a max(0, xxxx)
Should we check the 0 cases?

python/dgl/homophily.py
```diff
@@ -45,13 +57,121 @@ def node_homophily(graph, y):
...
+        src, dst = get_long_edges(graph)
+        # Compute y_v = y_u for all edges.
+        graph.edata["same_class"] = (y[src] == y[dst]).float()
+        graph.update_all(
+            fn.copy_e("same_class", "m"), fn.sum("m", "same_class_deg")
+        )
+
+        deg = graph.in_degrees().float()
+        num_nodes = graph.num_nodes()
+        value = 0
```
jermainewang, line 165: nit: I will usually separate loop-invariant variables with loop-mutable variables. For example:
```python
deg = graph.in_degrees().float()
num_nodes = graph.num_nodes()
num_classes = y.max(dim=0).values.item() + 1

value = 0
for k in range(num_classes):
    ...
    value += ...
```

python/dgl/homophily.py
```diff
@@ -45,13 +57,121 @@ def node_homophily(graph, y):
...
+        num_classes = y.max(dim=0).values.item() + 1
+
+        for k in range(num_classes):
+            # Get the nodes that belong to class k.
+            class_mask = y == k
+            same_class_deg_k = graph.ndata["same_class_deg"][class_mask].sum()
+            deg_k = deg[class_mask].sum()
+            # Value for a null model.
+            null_value = class_mask.sum() / num_nodes
+            value += max(0, same_class_deg_k / deg_k - null_value)
```
jermainewang, line 175: nit: I will try to align the code implementation with the math. E.g.,
```python
num_nodes_k = class_mask.sum()
value += max(0, same_class_deg_k / deg_k - num_nodes_k / num_nodes)
```
In general, I'm not sure whether the term "null model" or "null value" is a common sense and ppl can understand it without referring to the paper.

tests/python/common/test_homophily.py
jermainewang: +1

tests/python/common/test_homophily.py
```diff
@@ -16,4 +19,32 @@ def test_node_homophily(idtype):
         ([1, 2, 0, 4], [0, 1, 2, 3]), idtype=idtype, device=device
     )
     y = F.tensor([0, 0, 0, 0, 1])
-    assert dgl.node_homophily(graph, y) == 0.6000000238418579
+    assert math.isclose(dgl.node_homophily(graph, y), 0.6000000238418579)
```
jermainewang, line 22: Let's make the check looser to avoid spurious CI failing.

jermainewang (changes requested):

python/dgl/homophily.py
```diff
@@ -1,12 +1,24 @@
...
+except ImportError:
+    pass
+
+__all__ = ["node_homophily", "edge_homophily", "linkx_homophily"]
+
+
+def get_long_edges(graph):
+    """Internal function for getting the edges of a graph as long tensors."""
+    src, dst = graph.edges()
+    return src.long(), dst.long()
```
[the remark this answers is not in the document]
mufeili: I'm fine either way. Maybe you two can start a fight :) @frozenbugs

python/dgl/homophily.py
mufeili: I initially did what you suggested, and then the lint check failed.

tests/python/common/test_homophily.py
mufeili: I think the current cases are sufficient.

python/dgl/homophily.py
```diff
@@ -45,13 +57,121 @@ def node_homophily(graph, y):
...
+    """
+    with graph.local_scope():
+        # Compute |{u\in N(v): y_v = y_u}| for each node v.
+        # Handle the case where graph is of dtype int32.
+        src, dst = get_long_edges(graph)
+        # Compute y_v = y_u for all edges.
+        graph.edata["same_class"] = (y[src] == y[dst]).float()
+        graph.update_all(
+            fn.copy_e("same_class", "m"), fn.sum("m", "same_class_deg")
+        )
```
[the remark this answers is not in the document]
mufeili: How so? You convert the graph to a sparse matrix and call `AX`. I don't think there are significant differences.

tests/python/common/test_homophily.py
mufeili: done

python/dgl/homophily.py
mufeili: done

python/dgl/homophily.py
mufeili: done

python/dgl/homophily.py
mufeili: done

python/dgl/homophily.py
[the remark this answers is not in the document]
jermainewang:
```python
with graph.local_scope():
    # Handle the case where graph is of dtype int32.
    src, dst = get_long_edges(graph)
    # Compute y_v = y_u for all edges.
    graph.edata["same_class"] = (y[src] == y[dst]).float()
    graph.update_all(
        fn.copy_e("same_class", "m"), fn.mean("m", "same_class_deg")
    )
    return graph.ndata["same_class_deg"].mean(dim=0).item()
```
v.s.
```python
A = graph.adj
same_class = (y[A.row] == y[A.col]).float()
same_class_avg = dglsp.val_like(A, same_class).smean(dim=1)
return same_class_avg.mean(dim=0).item()
```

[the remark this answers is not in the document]
jermainewang: v.s. in the new message passing API style

```python
src, dst = get_long_edges(graph)
same_class = (y[src] == y[dst]).float()
same_class_avg = dgl.mpops.copy_e_mean(g, same_class)
return same_class_avg.mean(dim=0).item()
```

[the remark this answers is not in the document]
mufeili: Still, it's quite subtle. I'm fine either way.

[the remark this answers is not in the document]
jermainewang: My opinion is to go with the math formulation: If the model is described in node-wise/edge-wise computation then message passing is the way to goal; otherwise, use sparse. In this case, the definition is in node/edge so message passing is more suitable. You can see that although the sparse APIs are shorter, it doesn't align well with the definition, e.g., the use of `val_like` and `smean` is not straightforward.

jermainewang (approved):
LGTM. Do you want to remove the [DoNotMerge] tag?

PR #5385, opened 2023-02-27
[DistDGL][UserEx]Sync parmetis_wrapper with changes in metadata.json

kylasa:
1. In the preprocess.py, make sure that num_partitions is defined as input argument. Also, align 'input_dir' with the input dataset. schema_file is assumed to be located inside the input_dir. Also, graph_stats.txt file is assumed to be present in the input_dir.

2. Use DGL_HOME environment variable so that parmetis_wrapper command can be run anywhere.

## Checklist
- [X] Code is well-documented

### DIFF
--- a/tools/distpartitioning/parmetis_wrapper.py
+++ b/tools/distpartitioning/parmetis_wrapper.py
@@ -24,7 +24,7 @@ def check_dependencies():
         mpi_install
     ), "Could not locate the following dependency: MPI. Please install it and try again."
 
-    dgl_path = os.environ.get("DGL_HOME")
+    dgl_path = os.environ.get("DGL_HOME", "")
     assert os.path.isdir(
         dgl_path
     ), "Environment variable DGL_HOME not found. Please define the DGL installation path"
--- a/tools/distpartitioning/parmetis_wrapper.py
+++ b/tools/distpartitioning/parmetis_wrapper.py
@@ -24,6 +24,11 @@ def check_dependencies():
         mpi_install
     ), "Could not locate the following dependency: MPI. Please install it and try again."
 
+    dgl_path = os.environ.get("DGL_HOME")
+    assert os.path.isdir(
+        dgl_path
+    ), "Environment variable DGL_HOME not found. Please define the DGL installation path"
+
 
 def run_parmetis_wrapper(params):
     """Function to execute all the steps needed to run ParMETIS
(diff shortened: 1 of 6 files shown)


tools/distpartitioning/parmetis_wrapper.py
```diff
@@ -24,6 +24,11 @@ def check_dependencies():
         mpi_install
     ), "Could not locate the following dependency: MPI. Please install it and try again."
 
+    dgl_path = os.environ.get("DGL_HOME")
+    assert os.path.isdir(
+        dgl_path
+    ), "Environment variable DGL_HOME not found. Please define the DGL installation path"
```
Rhett-Ying, line 30: `DGL repository path`?

tools/distpartitioning/parmetis_wrapper.py
```diff
@@ -50,10 +54,11 @@ def run_parmetis_wrapper(params):
     # Trigger pre-processing step to generate input files for ParMETIS.
     preproc_cmd = (
         f"mpirun -np {num_partitions} -hostfile {params.hostfile} "
-        f"python3 tools/distpartitioning/parmetis_preprocess.py "
+        f"python3 $DGL_HOME/tools/distpartitioning/parmetis_preprocess.py "
```
Rhett-Ying, line 57: Have you verified this works? do users need to add `//tools` or any other directories into `PYTHONPATH`?

Rhett-Ying: something like this ? https://github.com/dmlc/dgl/blob/c396942d068d8c8c181162b0c9af4391ca404962/tests/scripts/task_distributed_test.sh#L39

tools/distpartitioning/parmetis_wrapper.py
kylasa: Yes. This is the DGL repository or DGL installation path. 

Typically, DGL_HOME is put in the '.profile' file, when I install DGL. 
What do you think about this change. This will allow us to run parmetis_wrapper.py from any location other than the dgl installation location.

tools/distpartitioning/parmetis_wrapper.py
```diff
@@ -24,6 +24,11 @@ def check_dependencies():
         mpi_install
     ), "Could not locate the following dependency: MPI. Please install it and try again."
 
+    dgl_path = os.environ.get("DGL_HOME")
```
Rhett-Ying, line 27: `os.environ.get("DGL_HOME", "")`. `os.path.isdir()` does not work with `None`

Rhett-Ying (approved):
Pls address the new comment before merge.

tools/distpartitioning/parmetis_wrapper.py
kylasa: Done.

PR #5390, opened 2023-02-28
Support for no. of chunks smaller than no. of partitions.

kylasa:
Following changes are made with this PR.
1. Code changes for handling no. of chunks smaller than no. of partitions
2. Adding new test cases, which were previously deleted, for no. of chunks smaller than no. of partitions.
3. Also adding test cases, where multiple partitions are handled by a single process.

## Checklist
- [ X] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [X] Code is well-documented

### DIFF
--- a/tools/distpartitioning/data_shuffle.py
+++ b/tools/distpartitioning/data_shuffle.py
@@ -24,6 +24,7 @@ from globalids import (
 from gloo_wrapper import allgather_sizes, alltoallv_cpu, gather_metadata_json
 from utils import (
     augment_edge_data,
+    DATA_TYPE_ID,
     get_edge_types,
     get_etype_featnames,
     get_gid_offsets,
@@ -36,6 +37,7 @@ from utils import (
     memory_snapshot,
     read_json,
     read_ntype_partition_files,
+    REV_DATA_TYPE_ID,
     write_dgl_objects,
     write_metadata_json,
 )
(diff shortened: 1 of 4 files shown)


tools/distpartitioning/data_shuffle.py
```diff
@@ -436,14 +439,58 @@ def exchange_feature(
         assert np.all(global_eids == data[constants.GLOBAL_EID][idx1])
         partid_slice = id_lookup.get_partition_ids(global_dst_nids)
 
+    # determine the shape of the feature-data
+    # this is needed to so that ranks where feature-data is not present
+    # should use the correct shape for sending the padded vector.
+    # exchange length here.
+    feat_dim_len = 0
+    if featdata_key != None:
```
Rhett-Ying, line 447: `is not None`

tools/distpartitioning/data_shuffle.py
```diff
@@ -436,14 +439,58 @@ def exchange_feature(
...
+        feat_dim_len = len(featdata_key.shape)
+    all_lens = allgather_sizes(
+        [feat_dim_len], world_size, num_parts, return_sizes=True
+    )
+    if all_lens[0] <= 0:
```
Rhett-Ying, line 452: any testcase to cover this scenario?

tools/distpartitioning/data_shuffle.py
```diff
@@ -436,14 +439,58 @@ def exchange_feature(
...
+
     for idx in range(world_size):
         cond = partid_slice == (idx + local_part_id * world_size)
         gids_per_partid = gids_feat[cond]
         tids_per_partid = tids_feat[cond]
         local_idx_partid = local_idx[cond]
 
         if gids_per_partid.shape[0] == 0:
-            feats_per_rank.append(torch.empty((0, 1), dtype=torch.float))
+            # feats_per_rank.append(torch.empty((0, 1), dtype=torch.float))
```
Rhett-Ying, line 485: remove it

Rhett-Ying (approved):

tools/distpartitioning/data_shuffle.py
kylasa: Since we support no. of chunks < no. of partitions, we will now run into situations when reading dataset from the disk some processes (when handling more than one partition per process) will not have any data in partitions other than '0'. In these cases, we use a sentinal value of node and edge_feature_tids as [(0,0)] to indicate that there is no data in these cases. In such cases we will hit this code, as pointed above. 

The existing test cases, where no. of partitions more than one per process will cover this corner case.

tools/distpartitioning/data_shuffle.py
kylasa: Done.

tools/distpartitioning/data_shuffle.py
kylasa: Done.

PR #5391, opened 2023-03-01
[DistDGL][Robustness]Replacing numpy's unique with custom implementation

kylasa:
numpy's unique is very memory expensive. It is about 8x times the input size. For large graphs this will be an issue. Consider the case where we have 3 billion edges, in this case, this function will consume memory which is as high as 550Gb. This is limiting the size of the input graph we can partition on a single machine. 

Replacing numpy's unique with memory efficient custom implementation. which is atmost 3x memory expensive. 

## Checklist
- [X] Code is well-documented
- [] Related issue is referred in this PR


kylasa:
https://github.com/dmlc/dgl/issues/5395 is opened to update the numpy version. Once this is fixed CI tests are expected to pass.

tools/distpartitioning/convert_partition.py
```diff
@@ -21,6 +21,94 @@
 from utils import get_idranges, memory_snapshot, read_json
 
 
+def _get_unique_invidx(srcids, dstids, uniq_nids):
```
Rhett-Ying, line 24: pls add dedicated testcase for this newly added function.

tools/distpartitioning/convert_partition.py
```diff
@@ -18,7 +18,129 @@
...
+        a list of integers. These are inverse indices, which will be indices
+        from the unique elements list specifying the elements from the
+        input array, srcids
+    numpy array :
+        a list of integers. These are inverse indices, which will be indices
+        from the unique elements list specifying the elements from the
+        input array, dstids
+
+    """
+    mask = np.in1d(srcids, nids, invert=True, kind="table")
```
Rhett-Ying, line 70: `kind=` requires `numpy >=1.24` which not available in CI.

tests/tools/test_convert_partition.py
```diff
@@ -0,0 +1,64 @@
...
+    srcids = np.random.randint(0, num_nodes, (num_edges,))
+    dstids = np.random.randint(nid_begin, nid_end, (num_edges,))
+
+    # generate unique node-ids for any
+    # partition. This list should be sorted.
+    # This is equivilant to shuffle_nids in a partition
+    unique_nids = np.arange(nid_begin, nid_end)
+
+    # invoke the test target
+    uniques, idxes, src_ids, dst_ids = _get_unique_invidx(
```
classicsong, line 33: Can you compare the output with a implementation using np.unique?

tests/tools/test_convert_partition.py
```diff
@@ -0,0 +1,64 @@
...
+
+from tools.distpartitioning import utils
+
+from tools.distpartitioning.convert_partition import _get_unique_invidx
+
+
+@pytest.mark.parametrize(
+    "num_nodes, num_edges, nid_begin, nid_end",
+    [
+        [4000, 40000, 0, 1000],
```
classicsong, line 15: Can you add test cases as [1,1,0,1], [4000,0,1000,4000], etc. to test the corner cases.

tools/distpartitioning/convert_partition.py
kylasa: Testing the new functionality without kind.

tests/tools/test_convert_partition.py
kylasa: Done.

tests/tools/test_convert_partition.py
kylasa: I don't understand. For every call do you want me to compare the workaround solution with numpy's unique ??

tests/tools/test_convert_partition.py
kylasa: [4000, 0, 1000, 4000] cannot be a valid test case, since we always need a non-empty edge-list to process. One of the return parameters of this function is the edge-list itself, which is over-written with the return parameters.

tests/tools/test_convert_partition.py
```diff
@@ -4,25 +4,115 @@
...
+        [4000, 40000, 2000, 3000],
+        [4000, 40000, 3000, 4000],
+        [4000, 100, 0, 1000],
+        [4000, 100, 1000, 2000],
+        [4000, 100, 2000, 3000],
+        [4000, 100, 3000, 4000],
+        [1, 1, 0, 1],
+    ],
+)
+def test_get_unique_invidx_with_numpy(num_nodes, num_edges, nid_begin, nid_end):
```
classicsong, line 26: What is the difference between test_get_unique_invidx_with_numpy and test_get_unique_invidx?

tools/distpartitioning/convert_partition.py
```diff
@@ -65,44 +78,46 @@ def _get_unique_invidx(srcids, dstids, nids):
...
-        f"It is not a super-set. "
-        f"len(dstids_int_nids) = {len(dstids_int_nids)}, len(nids) = {len(nids)}"
-    )
-
-    # validation check here.
-    assert np.all(gold_uniques.sort() == uniques.sort()), (
-        f"Our assumption about gold_unique is NOT right. "
-        f"len(gold_uniques) == {len(gold_uniques)} "
-        f"len(uniques) == {len(uniques)}"
+    # build uniques and idxes, first and second return parameters
```
classicsong, line 117: Did you change the implementation?

tools/distpartitioning/convert_partition.py
```diff
@@ -65,44 +78,46 @@ def _get_unique_invidx(srcids, dstids, nids):
...
+    uniques = np.concatenate([uniques, nids])
+    idxes = np.concatenate(
+        [idxes, len(srcids) + len(dstids) + np.arange(len(nids))]
```
classicsong, line 120: Can you prepare a google doc describing the details of the `unique` implementation?

tests/tools/test_convert_partition.py
kylasa: test_get_unique_invidx_with_numpy - is the one you requested. Comparing the results with numpy's unique results. and test_get_unique_invidx is the test case for testing our work-around solution (here I validate the results from the workaround solution). I like to keep them separated.

tools/distpartitioning/convert_partition.py
kylasa: In a way yes. This newer implementation is optimized for wall-clock time. In the earlier version, which was used for benchmarking purposes on felarge and webgraphs, we used to compute the list of uniques in a step by step manner - and validating them as well. Now I removed the validation part, since the benchmarking results were highly reduced when compared to the numpy's unique implementation.

tools/distpartitioning/convert_partition.py
kylasa: The doc string and the comments in the code are very extensive. If you follow these comments, you should be able understand the implementation fairly easily. But if you prefer a quip document, I will put the same text in the quip document and share it with a larger group.

kylasa: https://quip-amazon.com/x5KSAK0FamJE/Dist-Graph-Partitioning-Pipeline-Numpy-Unique-Issues

classicsong (approved):

zheng-da:
The test script doesn't run in CI?

Issue #4507, opened 2022-09-02
[bug] dgl.dataloading.DataLoader forwards args to torch.utils.data.DataLoader but many are not valid


nv-dlasalle:
## 🐛 Bug

Currently, the DGL's DataLoader is a child of PyTorch's DataLoader, and passes `kwargs` to it, however DGL's dataloader does several things under the hood which breaks many of these options, and no indication is given to the user what is valid and what is not. Further more, CI covers a fraction of possible options users my pass in.

Current problematic options:
* `sampler` and `batch_sampler`: Specifying either of these results in errors about `IterableDataset`. Our documentation explicitly references `sampler` inside the description for the `use_ddp` argument.
* `collate_fn`: This results in a duplicate keyword error.
* `pin_memory`: This results in a duplicate keyword error, as the super is called with `pin_memory=self.pin_prefetcher`.
* `pin_memory_device` (new in 1.12): This will only have effect on the input to the _PrefetchingIterator, which may not affect the data as returned to the user.
* `generator`: Has no effect.
* `timeout`: While this will work relatively correctly with respect to the workers, when a prefetching thread is active, it is possible to block on the dataloader for longer than the specified timeout, which would be unexpected from the user's perspective.
* `worker_init_fn`: This results in a duplicate keyword error, however this appears just to be a bug as we attempt to correctly wrap the passed in function, but fail to delete it from the `kwargs`.
* `batch_size=None`: While it would be odd for a user to do this given we don't provide users a way to manually batch data, our documentation does address it when it is valid for PyTorch's DataLoader.

We also have the issue of when PyTorch's DataLoader is updated, how to maintain compatibility. For example, the option `pin_memory_device` was just added in 1.12.

## Proposed fix

I think the best way to fix this, is to remove the `kwargs` parameter to our DataLoader, and instead capture every argument explicitly. This way we can limit, verify, and test the different argument combinations passed in. It will also make reading our documentation easier.

## Alternative fix

We could try to keep documentation on what parameters are valid, but we may need separate documentation per PyTorch version.

llCurious:
Any updates?
I also have the problem of passing `worker_init_fn`.

I manage to remove the **kwargs and the duplicate keyword error is gone. However, the `worker_init_fn` does not seem to work.

BarclayII:
> Any updates? I also have the problem of passing `worker_init_fn`.
> 
[...]

PR #5420, opened 2023-03-03
[Bugfix] Fix duplicate worker_init_fn argument when provided in DataLoader

BarclayII:
Addresses one of the problems in #4507 as requested.

### DIFF
--- a/tests/python/pytorch/dataloading/test_dataloader.py
+++ b/tests/python/pytorch/dataloading/test_dataloader.py
@@ -91,6 +91,7 @@ def test_saint(num_workers, mode):
     for sg in dataloader:
         pass
 
+
 @parametrize_idtype
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


tests/python/pytorch/dataloading/test_dataloader.py
```diff
@@ -622,6 +622,26 @@ def test_edge_dataloader_excludes(
             break
 
 
+def dummy_worker_init_fn(worker_id):
+    pass
+
+
+def test_dataloader_worker_init_fn():
```
frozenbugs, line 629: It seems that the test case added here does not cover the fix.

BarclayII: It does.  Previously it will throw an error even if `worker_init_fn` is an empty function (as long as it's not None).

frozenbugs (approved):

frozenbugs (approved):

Issue #5367, opened 2023-02-23
[Sparse] Remove Python DiagMatrix class and represent it as a SparseMatrix.


czkkkkkk:
## 🔨Work Item

**IMPORTANT:**
* This template is only for dev team to track project progress. For feature request or bug report, please use the corresponding issue templates.
* DO NOT create a new work item if the purpose is to fix an existing issue or feature request. We will directly use the issue in the project tracker.

Project tracker: https://github.com/orgs/dmlc/projects/2

There are two PRs for this issue.
1. Support C++ Diag sparse format. (#5432)
2. Remove DiagMatrix in Python and enable the Python sparse matrix with C++ Diag sparse matrix.

PR #5432, opened 2023-03-07
[Sparse] Support Diag sparse format in C++

czkkkkkk:
## Description

This is one of the PR for the issue #5367 . 

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/dgl_sparse/include/sparse/sparse_format.h
+++ b/dgl_sparse/include/sparse/sparse_format.h
@@ -19,7 +19,7 @@ namespace dgl {
 namespace sparse {
 
 /** @brief SparseFormat enumeration. */
-enum SparseFormat { kCOO, kCSR, kCSC };
+enum SparseFormat { kCOO, kCSR, kCSC, kDiag };
 
 /** @brief COO sparse structure. */
 struct COO {
@@ -48,6 +48,11 @@ struct CSR {
   torch::optional<torch::Tensor> value_indices;
   /** @brief Whether the column indices per row are sorted. */
   bool sorted = false;
+};
+
+struct Diag {
+  /** @brief The dense shape of the matrix. */
+  int64_t num_rows = 0, num_cols = 0;
 };
 
 /** @brief Convert an old DGL COO format to a COO in the sparse library. */
@@ -89,6 +94,21 @@ std::shared_ptr<CSR> COOToCSC(const std::shared_ptr<COO>& coo);
 
 /** @brief Convert a CSR format to CSC format. */
 std::shared_ptr<CSR> CSRToCSC(const std::shared_ptr<CSR>& csr);
+
+/** @brief Convert a Diag format to COO format. */
+std::shared_ptr<COO> DiagToCOO(
+    const std::shared_ptr<Diag>& diag,
+    const c10::TensorOptions& indices_options);
+
+/** @brief Convert a Diag format to CSR format. */
+std::shared_ptr<CSR> DiagToCSR(
+    const std::shared_ptr<Diag>& diag,
+    const c10::TensorOptions& indices_options);
+
+/** @brief Convert a Diag format to CSC format. */
+std::shared_ptr<CSR> DiagToCSC(
+    const std::shared_ptr<Diag>& diag,
+    const c10::TensorOptions& indices_options);
 
 /** @brief COO transposition. */
 std::shared_ptr<COO> COOTranspose(const std::shared_ptr<COO>& coo);
--- a/dgl_sparse/src/elemenwise_op.cc
+++ b/dgl_sparse/src/elemenwise_op.cc
@@ -22,6 +22,10 @@ c10::intrusive_ptr<SparseMatrix> SpSpAdd(
     const c10::intrusive_ptr<SparseMatrix>& A,
     const c10::intrusive_ptr<SparseMatrix>& B) {
   ElementwiseOpSanityCheck(A, B);
+  if (A->HasDiag() && B->HasDiag()) {
+    return SparseMatrix::FromDiagPointer(
+        A->DiagPtr(), A->value() + B->value(), A->shape());
+  }
   auto torch_A = COOToTorchCOO(A->COOPtr(), A->value());
   auto torch_B = COOToTorchCOO(B->COOPtr(), B->value());
   auto sum = (torch_A + torch_B).coalesce();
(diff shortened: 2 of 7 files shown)


dgl_sparse/include/sparse/sparse_format.h
```diff
@@ -90,6 +95,21 @@ std::shared_ptr<CSR> COOToCSC(const std::shared_ptr<COO>& coo);
 /** @brief Convert a CSR format to CSC format. */
 std::shared_ptr<CSR> CSRToCSC(const std::shared_ptr<CSR>& csr);
 
+/** @brief Convert a Diag format to COO format. */
+std::shared_ptr<COO> DiagToCOO(
```
frozenbugs, line 99: Why do we need this 3 conversion? In which case, they will be used?

czkkkkkk: They are used for operators that do not have implementation on Diag format, e.g., SpMM.

frozenbugs: What's the implication of the performance here if we convert diag to COO/CSR/CSV for operators?
Will it strongly decrease the spmm performance?

This solution looks good to me, but let's make sure we understand the trade-off we are making here.

frozenbugs (approved):

dgl_sparse/include/sparse/sparse_format.h
czkkkkkk: I simply follow our current implementation to ensure no performance regression. Currently, we also convert the DiagMatrix to SparseMatrix for SpMM on the Python side.

dgl_sparse/src/spspmm.cc
```diff
@@ -116,10 +116,44 @@ tensor_list SpSpMMAutoGrad::backward(
...
+      auto new_val =
+          lhs_mat->value().index({slice}) * rhs_mat->value().index({slice});
+      new_val = torch::constant_pad_nd(
+          new_val, {0, new_diag_len - common_diag_len}, 0);
+      return SparseMatrix::FromDiag(new_val, {m, p});
+    } else {
+      // Diag @ Sparse
+      auto row = rhs_mat->Indices().index({0});
+      auto val = lhs_mat->value().index_select(0, row) * rhs_mat->value();
+      return SparseMatrix::ValLike(rhs_mat, val);
```
jermainewang, line 140: Is this implementation correct? The index_select should use `A.row` according to https://github.com/dmlc/dgl/blob/master/python/dgl/sparse/matmul.py#L172.

dgl_sparse/src/spspmm.cc
```diff
@@ -116,10 +116,44 @@ tensor_list SpSpMMAutoGrad::backward(
...
+  }
+}
+
 c10::intrusive_ptr<SparseMatrix> SpSpMM(
     const c10::intrusive_ptr<SparseMatrix>& lhs_mat,
     const c10::intrusive_ptr<SparseMatrix>& rhs_mat) {
   _SpSpMMSanityCheck(lhs_mat, rhs_mat);
+  if (lhs_mat->HasDiag() || rhs_mat->HasDiag()) {
+    return DiagSpSpMM(lhs_mat, rhs_mat);
+  }
```
jermainewang, line 156: Same comment on if-branch code style.

dgl_sparse/src/elemenwise_op.cc
```diff
@@ -22,6 +22,10 @@ c10::intrusive_ptr<SparseMatrix> SpSpAdd(
     const c10::intrusive_ptr<SparseMatrix>& A,
     const c10::intrusive_ptr<SparseMatrix>& B) {
   ElementwiseOpSanityCheck(A, B);
+  if (A->HasDiag() && B->HasDiag()) {
+    return SparseMatrix::FromDiagPointer(
+        A->DiagPtr(), A->value() + B->value(), A->shape());
+  }
```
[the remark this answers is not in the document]
frozenbugs: The current implementation is better.
For all functions, it could have multiple early return checks, so it is highly recommended to do:
```
if (...) {
  return ...
}
if (...) {
  return ...
}
blablabla
blablabla
blablabla
return ...
```

dgl_sparse/src/spspmm.cc
```diff
@@ -116,10 +116,44 @@ tensor_list SpSpMMAutoGrad::backward(
   return {torch::Tensor(), lhs_val_grad, torch::Tensor(), rhs_val_grad};
 }
 
+c10::intrusive_ptr<SparseMatrix> DiagSpSpMM(
+    const c10::intrusive_ptr<SparseMatrix>& lhs_mat,
+    const c10::intrusive_ptr<SparseMatrix>& rhs_mat) {
+  if (lhs_mat->HasDiag()) {
+    if (rhs_mat->HasDiag()) {
```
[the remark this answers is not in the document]
frozenbugs: +1

dgl_sparse/src/spspmm.cc
czkkkkkk: I think so. `rhs_mat->Indices().index({0})` returns the row coordinates of the SparseMatrix without data copy.

dgl_sparse/src/sparse_format.cc
```diff
@@ -99,6 +99,45 @@ std::shared_ptr<CSR> CSRToCSC(const std::shared_ptr<CSR>& csr) {
...
+  auto indices = torch::arange(nnz, indices_options).repeat({2, 1});
+  return std::make_shared<COO>(
+      COO{diag->num_rows, diag->num_cols, indices, true, true});
+}
+
+std::shared_ptr<CSR> DiagToCSR(
+    const std::shared_ptr<Diag>& diag,
+    const c10::TensorOptions& indices_options) {
+  int64_t nnz = std::min(diag->num_rows, diag->num_cols);
+  auto indptr = torch::arange(nnz + 1, indices_options);
```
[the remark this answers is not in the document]
czkkkkkk: Updated

dgl_sparse/src/spspmm.cc
jermainewang: I see. `rhs_mat->Indices()` returns the COO index. How is it different with `rhs_mat->COOPointer()->indices`? Do we want to have two interfaces?

czkkkkkk: They are the same. But we need `Indices()` exposed to python to avoid data copy of the COO tensor.

jermainewang (approved):

PR #5457, opened 2023-03-15
Fix CI Test case for graph partitioning validation script

kylasa:
## Description

CI Tests for verify_partitions.py script, which is used for validation of partitioned graph objects, was executed using os.system("string") command. This resulted in running the verification script, but as a system command (as a separate process, out of the context of the CI Test process). Because of this reason, even though the verify_partitions.py was failing in the newly spawned process it was NOT caught by the CI Test process leading us to believe that the CI Test passed successfully!. 

This above behavior is corrected now with this PR. In this PR, the verify_partitions.py is NO longer executed as a system command. Now, for the CI test cases it is run directly by invoking the main function from the verify_partitions.py. This will make sure that all the errors happening in verify_partitions.py are correctly caught and reported by the CI test case which is running that function.

## Checklist
- [X] Code is well-documented

### DIFF
--- a/tests/tools/test_dist_part.py
+++ b/tests/tools/test_dist_part.py
@@ -26,6 +26,7 @@ from tools.verification_utils import (
     verify_partition_data_types,
     verify_partition_formats,
 )
+from tools.verify_partitions import validate_results
 
 
 def _test_chunk_graph(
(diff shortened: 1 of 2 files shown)


frozenbugs:
Good, now we successfully trigger the error in the CI. Please close this PR and merge it with the fix and verify the error is fixed.

Issue #5454, opened 2023-03-15
Creating output buffers via `torch.empty()` instead of `torch.zeros()` when possible


yaox12:
## Description

In cases where the output buffers don't have to be zeros, it's better to use `torch.empty()` to create output buffers instead of `torch.zeros()` in favor of better performance.

For example, in `segmentMM`,
https://github.com/dmlc/dgl/blob/f5ddb11448335c45b16fb1d0561c490d937ab8f5/python/dgl/backend/pytorch/sparse.py#L973
creating a buffer for the output matrix C can be significantly time-consuming for large graphs.

The following figure shows the profiling timeline for RGCN on MAG240M. As we can see, `aten::zeros` takes almost 1/3 of the total time of `segmentMM`.

![image](https://user-images.githubusercontent.com/3831900/225225116-df33a374-fe35-437d-96ea-9d745a67936d.png)

## Additional context

Pay attention to ops that require the output buffer to be strict zero, for example, `SpMMCoo` if I understand correctly.

Rhett-Ying:
could you help optimize it

mufeili:
Can we close this issue now? @yaox12

PR #5462, opened 2023-03-17
[Performance] Creating out buffers for `segment_mm`|`sddmm` via `torch.empty()`

yaox12:
## Description

As stated in https://github.com/dmlc/dgl/issues/5454, creating out buffers via `torch.empty()` instead of `torch.zeros()` usually give a better performance.

This PR covers the changes for `segment_mm` and `sddmm`. `spmm` is much more complicated as the replacement can only happen in a few code paths. So I'd like to put it in another PR.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/backend/pytorch/sparse.py
+++ b/python/dgl/backend/pytorch/sparse.py
@@ -970,7 +970,7 @@ class SEGMENTMM(th.autograd.Function):
     def forward(ctx, A, B, seglen_A):
         if B.dim() != 3:
             raise ValueError("segment_mm expects B to be a 3D tensor.")
-        C = th.zeros((A.shape[0], B.shape[2]), device=A.device, dtype=A.dtype)
+        C = th.empty((A.shape[0], B.shape[2]), device=A.device, dtype=A.dtype)
         C = _segment_mm(A, B, C, seglen_A)
         ctx.backward_cache = A, B, seglen_A
         return C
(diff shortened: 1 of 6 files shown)


yaox12:
@dgl-bot

python/dgl/backend/pytorch/sparse.py
```diff
@@ -970,7 +970,7 @@ class SEGMENTMM(th.autograd.Function):
     def forward(ctx, A, B, seglen_A):
         if B.dim() != 3:
             raise ValueError("segment_mm expects B to be a 3D tensor.")
-        C = th.zeros((A.shape[0], B.shape[2]), device=A.device, dtype=A.dtype)
+        C = th.empty((A.shape[0], B.shape[2]), device=A.device, dtype=A.dtype)
```
Rhett-Ying, line 973: shall we fix for `GATHERMM` too?

yaox12: The GatherMM kernel updates the output matrix C in an in-place manner. So it requires the output buffer being initialized as zeros.
https://github.com/dmlc/dgl/blob/097fd6dbb1a2ec306438edd1b94a29ec393fa0fa/src/array/cuda/gather_mm.cu#L124-L128

Rhett-Ying (approved):

PR #5465, opened 2023-03-17
[Sparse] Fix bugs in Diag format conversions

czkkkkkk:
## Description

`arange_out` is not applicable because it changes the shape of a tensor.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/dgl_sparse/src/sparse_format.cc
+++ b/dgl_sparse/src/sparse_format.cc
@@ -113,7 +113,8 @@ std::shared_ptr<CSR> DiagToCSR(
     const c10::TensorOptions& indices_options) {
   int64_t nnz = std::min(diag->num_rows, diag->num_cols);
   auto indptr = torch::full(diag->num_rows + 1, nnz, indices_options);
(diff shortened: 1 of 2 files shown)


dgl_sparse/src/sparse_format.cc
```diff
@@ -113,7 +113,8 @@ std::shared_ptr<CSR> DiagToCSR(
     const c10::TensorOptions& indices_options) {
```
frozenbugs, line 113: not related to this PR, why do we need to pass indices_options in instead of create it here?

frozenbugs: why other conversion does not need the option. e.g. COOToCSC

czkkkkkk: This is used to decide the dtype and devices of the format to convert. The Diag sparse format only have `num_rows` and `num_cols` without those information.

frozenbugs (approved):

PR #5466, opened 2023-03-17
[Optimization] Optimize bias term in GatConv layer

DominikaJedynak:
## Description
This PR optimizes bias calculation in GatConv layer when using residual connection by delegating bias to fc in this scenario:
Before: rst = rst + self.res_fc(h_dst) + self.bias = rst + (weight * h_dst) + self.bias
After: rst = rst + self.res_fc(h_dst) with bias = rst + (weight * h_dst + bias)

Change is partially similar to https://github.com/dmlc/dgl/pull/4747.

Measured on Intel Xeon(R) Platinum 8375C (r6i.16xlarge)
Current master:
Screenshot (62)

After the change:
Screenshot (54)

## Checklist


jermainewang:
@dgl-bot

python/dgl/nn/pytorch/conv/gatconv.py
```diff
@@ -170,21 +170,23 @@ def __init__(
...
+                if bias:
+                    self.bias = nn.Parameter(
+                        th.FloatTensor(size=(num_heads * out_feats,))
+                    )
         else:
             self.register_buffer("res_fc", None)
+            if bias:
+                self.bias = nn.Parameter(
+                    th.FloatTensor(size=(num_heads * out_feats,))
+                )
```
jermainewang, line 189: Let's simplify this branch to make it more readable. How about:
```python
self.has_res_fc = False
self.has_bias = False
if residual:
    if self._in_dst_feats != out_feats * num_heads:
        self.res_fc = nn.Linear(
            self._in_dst_feats, num_heads * out_feats, bias=bias
        )
        self.has_res_fc = True
    else:
        self.res_fc = Identity()

if bias and not self.has_res_fc:
    self.bias = nn.Parameter(
        th.FloatTensor(size=(num_heads * out_feats,))
    )
    self.has_bias = True
```

python/dgl/nn/pytorch/conv/gatconv.py
```diff
@@ -208,10 +210,12 @@ def reset_parameters(self):
             nn.init.xavier_normal_(self.fc_dst.weight, gain=gain)
         nn.init.xavier_normal_(self.attn_l, gain=gain)
         nn.init.xavier_normal_(self.attn_r, gain=gain)
-        if self.bias is not None:
+        if hasattr(self, "bias"):
```
jermainewang, line 213: `if self.has_bias:`

python/dgl/nn/pytorch/conv/gatconv.py
```diff
@@ -208,10 +210,12 @@ def reset_parameters(self):
...
             nn.init.constant_(self.bias, 0)
         if isinstance(self.res_fc, nn.Linear):
             nn.init.xavier_normal_(self.res_fc.weight, gain=gain)
+            if self.res_fc.bias is not None:
```
jermainewang, line 217: not needed anymore

python/dgl/nn/pytorch/conv/gatconv.py
```diff
@@ -338,7 +342,7 @@ def forward(self, graph, feat, get_attention=False):
                 )
                 rst = rst + resval
             # bias
-            if self.bias is not None:
+            if hasattr(self, "bias"):
```
jermainewang, line 345: `if self.has_bias:`

jermainewang (changes requested):
Hi @DominikaJedynak , I've left some suggestions on how to make the code more readable. Thanks for the contribution!

python/dgl/nn/pytorch/conv/gatconv.py
jermainewang: Similarly, we can change L338 to use the flag we've saved.
```python
if self.has_res_fc:
    ...
```

python/dgl/nn/pytorch/conv/gatconv.py
DominikaJedynak: Why isn't it needed anymore? We still can have both situations when fc has no or has bias inside, which might need to be reset.

python/dgl/nn/pytorch/conv/gatconv.py
DominikaJedynak: @jermainewang Thank you for your suggestion. I have updated the code but I am concerned that the names of the flags might be misleading: eg. the flag has_bias is set to false when actually bias is present but inside nn.Linear. What do you think of changing eg. has_bias for has_separate_bias and has_res_fc for has_linear_res?

python/dgl/nn/pytorch/conv/gatconv.py
jermainewang: I see you. You are right.

jermainewang (approved):

jermainewang:
@dgl-bot

python/dgl/nn/pytorch/conv/gatconv.py
jermainewang: Or `has_explicit_bias`?

DominikaJedynak:
@dgl-bot

PR #5165, opened 2023-01-12
[Config] Enable libxsmm by default for AVX cpu

daniil-sizov:
## Description
Enable libxsmm by default for AVX cpu. Improves performance for AVX by leveraging libxsmm

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [x] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [x] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

## Changes
- [x] Removed unused xbyak code (breaking build before the change)

PR #5470, opened 2023-03-20
[Cleanup] Cleanup unused CMake options

yaox12:
## Description

Remove the following CMake options used anywhere:
- `USE_AVX` that was removed in https://github.com/dmlc/dgl/pull/5165.
- `USE_NCCL` that was removed in https://github.com/dmlc/dgl/pull/4989.
- `USE_FP16` that was removed in https://github.com/dmlc/dgl/pull/4648.

## Checklist
- [ ] Code is well-documented


benchmarks/scripts/build_dgl_asv.sh
```diff
@@ -11,11 +11,7 @@ pip install -r /asv/torch_gpu_pip.txt
 # 'CUDA_TOOLKIT_ROOT_DIR' is always required for sparse build as torch1.13.1+cu116 is installed.
 CMAKE_VARS="-DUSE_OPENMP=ON -DBUILD_TORCH=ON -DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda"
 if [[ $DEVICE == "gpu" ]]; then
-    CMAKE_VARS="-DUSE_CUDA=ON -DUSE_NCCL=ON $CMAKE_VARS"
-fi
-arch=`uname -m`
-if [[ $arch == *"x86"* ]]; then
```
Rhett-Ying, line 17: is AVX available on non-x86 machines?

cmake/config.cmake
```diff
@@ -40,14 +40,11 @@ set(BUILD_CPP_TEST OFF)
...
-set(USE_AVX ON)
-
 # Whether to build PyTorch plugins.
 set(BUILD_TORCH ON)
 
 # Whether to enable CUDA kernels compiled with TVM.
 set(USE_TVM OFF)
 
-# Whether to enable fp16 to support mixed precision training.
-set(USE_FP16 OFF)
```
Rhett-Ying, line 53: pls remove `USE_FP16` from `//docs/source/guide_ko/mixed_precision.rst` as well

CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ dgl_option(USE_HDFS "Build with HDFS support" OFF) # Set env HADOOP_HDFS_HOME if
 dgl_option(REBUILD_LIBXSMM "Clean LIBXSMM build cache at every build" OFF) # Set env HADOOP_HDFS_HOME if needed
 dgl_option(USE_EPOLL "Build with epoll for socket communicator" ON)
 dgl_option(TP_BUILD_LIBUV "Build libuv together with tensorpipe (only impacts Linux)" ON)
+dgl_option(BUILD_TORCH "Build the PyTorch plugin" ON)
```
Rhett-Ying, line 36: I think in previous code, it's **OFF* in default as `//cmake/config.cmake` is not copied to build directory in default.

cmake/config.cmake
yaox12: Removed. I copy the ko doc to google translate and remove paragraphs about building from source with USE_FP16.

benchmarks/scripts/build_dgl_asv.sh
yaox12: `USE_AVX` was removed in #5165. I guess https://github.com/dmlc/dgl/blob/866c70dace72c8d73e2011b8881fcf0a6eea4be7/src/runtime/config.cc#L20 will return false on non-AVX CPUs.

Rhett-Ying: `-DUSE_AVX=OFF` is specified when building package for osx. But does `USE_AVX` really take effect as it's already removed. @BarclayII do you have any comments on this?

Rhett-Ying: confirmed offline, `USE_AVX` does not take effect now. so it's ok to remove it at all.

Rhett-Ying (approved):

PR #5473, opened 2023-03-20
Bug: TypedStorage is deprecated

tanmoyio:


yaox12:
I don't think simply replace `storage()` with `untyped_storage()` will solve this problem, because
1. `untyped_storage().size()` returns the number of bytes instead of a number of elements.
2. `untyped_storage()` is only available since PyTorch 2.0 (was `_storage()`), while DGL supports PyTorch 1.12+.
3. There're other `.storage()` in DGL.

jermainewang:
@tanmoyio Could you provide more context? What are the potential issue without this fix?

yaox12:
I opened PR #5656 to fix it. So close this.

Issue #4293, opened 2022-07-25
Add versioning to all DGLDatasets to force reloading when codes are changed


jermainewang:
## 🚀 Feature

Add versioning to all DGLDatasets to detect:
* Changes to the raw dataset files stored in S3. If so, re-download the dataset.
* Changes to the preprocessing code. If so, ignore the previous local cache and reprocess the data.

## Motivation

Brought up by #3987 which asks to revert the default reordering behavior of DGL builtin datasets. The issue is that even if we've implemented the request, users may still load cached datasets from local disk which may not reflect the latest change. Therefore, we require some versioning mechanism to detect those changes.

## Alternatives

Use a different dataset folder whenever DGL updates. This could cause excessive disk storage use.

mufeili:
| Dataset | Current versioning in cached files | Desired versioning behavior |
| -------- | --------------------------------- | ---------------------------- |
| Yelp       |

PR #5477, opened 2023-03-21
[Dataset] Chameleon

mufeili:
## Description
This PR adds the variant of Chameleon as introduced in [Geom-GCN: Geometric Graph Convolutional Networks](https://arxiv.org/abs/2002.05287).

A few thoughts and suggested practice:
- Handling changes of the data source. In general, this should be very rare. If the data source indeed changes, the url will need to be changed as well. This PR appends a hash value based on URL to the data directory so that the data can be re-downloaded and processed by checking if the directory exists.
- DGLGraph construction
    - If the graph is relatively small, then constructing it from the raw data files is very efficient as in this PR. In this case, I suggest not caching any DGL-specific data like DGLGraph. This can save a lot of trouble like if the graph is constructed in the desired way.
    - If the graph is very large, then we should have some file to track the configuration associated as I suggested in #4293.
- Test and regression
    - For small graphs as in this PR, unit tests are cheap enough.
    - For larger graphs, then we can have very preliminary unit tests or even no unit tests at all and instead rely on regression tests.

## Checklist
- [x] Code is well-documented


python/dgl/data/wiki_network.py
jermainewang, line 1: Suggest to avoid using framework-agnostic backend for new datasets. Make them PyTorch only.

python/dgl/data/wiki_network.py
```diff
@@ -0,0 +1,186 @@
...
+            test_masks += [F.zerocopy_from_numpy(f["test_mask"])]
+        self._g.ndata["train_mask"] = F.astype(
+            F.stack(train_masks, dim=1), F.data_type_dict["bool"]
+        )
+        self._g.ndata["val_mask"] = F.astype(
+            F.stack(val_masks, dim=1), F.data_type_dict["bool"]
+        )
+        self._g.ndata["test_mask"] = F.astype(
+            F.stack(test_masks, dim=1), F.data_type_dict["bool"]
+        )
```
jermainewang, line 89: The general question is whether the logic of data processing / graph construction should belong to DGL core codebase or not.

python/dgl/data/dgl_dataset.py
```diff
@@ -221,6 +220,15 @@ def _get_hash(self):
...
         return hash_func.hexdigest()[:8]
 
+    def _get_hash_url(self):
+        """Compute the hash of the url"""
+        if self._url is None:
+            return ""
+        else:
+            hash_func = hashlib.sha1()
+            hash_func.update(str(self._url).encode("utf-8"))
+            return "_" + hash_func.hexdigest()[:8]
```
frozenbugs, line 230: the `"_" +` should not be appended here since the function is named. _get_hash_url

python/dgl/data/wiki_network.py
```diff
@@ -0,0 +1,186 @@
+"""
+Wikipedia page-page networks on three topics: chameleons, crocodiles and
+squirrels
```
frozenbugs, line 3: end with dot.

frozenbugs, line 3: I don't see crocodiles and squirrels related code.

python/dgl/data/wiki_network.py
```diff
@@ -0,0 +1,186 @@
...
+            raw_dir=raw_dir,
+            hash_key=(url),
+            force_reload=force_reload,
+            verbose=verbose,
+            transform=transform,
+        )
+
+    def process(self):
+        """Load and process the data."""
+        # node features and labels
```
frozenbugs, line 50: \# Process node features and labels.

python/dgl/data/wiki_network.py
```diff
@@ -0,0 +1,186 @@
...
+            data = [[int(v) for v in r.split("\t")] for r in data]
+        COO = F.swapaxes(F.tensor(data, dtype=F.data_type_dict["int64"]), 0, 1)
+        dst = F.squeeze(F.narrow_row(COO, start=0, stop=1), dim=0)
+        src = F.squeeze(F.narrow_row(COO, start=1, stop=2), dim=0)
+
+        self._g = graph((src, dst), num_nodes=F.shape(features)[0])
+        self._g.ndata["feat"] = features
+        self._g.ndata["label"] = labels
+
+        # split
```
frozenbugs, line 73: \# Split blablabla.

python/dgl/data/wiki_network.py
```diff
@@ -0,0 +1,186 @@
...
+
+    @property
+    def num_classes(self):
+        return self._num_classes
+
+
+class ChameleonDataset(WikiNetworkDataset):
+    """Wikipedia page-page network on chameleons from `Multi-scale Attributed
+    Node Embedding <https://arxiv.org/abs/1909.13021>`__ and later processed by
+    `Geom-GCN: Geometric Graph Convolutional Networks
```
frozenbugs, line 115: Why Geom-GCN is related? Is this dataset designated for Geom-GCN?

python/dgl/data/wiki_network.py
mufeili: Done

python/dgl/data/wiki_network.py
mufeili: Upon the offline discussion, the general principle will be including the logic of data processing in the core DGL codebase.

python/dgl/data/dgl_dataset.py
mufeili: renamed to `_get_hash_url_suffix`

python/dgl/data/wiki_network.py
mufeili: removed for now, will add later in another PR after we merge this one

python/dgl/data/wiki_network.py
mufeili: Geom-GCN introduced this variant of the dataset, including turning the task from node regression into node classification, modifying node features, and introducing these dataset splits

frozenbugs: Maybe reword it. Currently working sounds like this dataset is just used by Geom-GCN.
e.g. later processed by --> introduced by.
```
Wikipedia page-page network on chameleons from `Multi-scale Attributed
    Node Embedding <https://arxiv.org/abs/1909.13021>`__, introduced by
    `Geom-GCN: Geometric Graph Convolutional Networks
```

frozenbugs (approved):

jermainewang (approved):

python/dgl/data/wiki_network.py
mufeili: did "later processed by" -> "modified by"

PR #5479, opened 2023-03-22
[Enhancement]Set default graph dataloader thread number

peizhou001:
## Description
According to our test, graph dataloader default start up a lot of threads for `disjoint_union` operation, unfortunately, it doesn't bring up speed up whereas may run out of the CPU resource. This PR add a default grain size according to experience to limit the default thread number for graph data loader.

## Checklist
- [ ] Code is well-documented


jermainewang:
@peizhou001 Remember to assign reviewers or nobody will see it!

czkkkkkk:
* What is the downside of running out of CPU threads? I feel they will be wasted if they are not fully utilized.
* Setting the number of threads may depend on the machine. Does the performance degrade if a machine only have 8 cores but we use 32 threads? I'd suggest to set the thread numbers through a system call, such as [hardward_concurrency](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency).

src/array/cpu/disjoint_union.cc
```diff
@@ -85,32 +85,38 @@ COOMatrix DisjointUnionCoo(const std::vector<COOMatrix>& coos) {
...
-
-      if (has_data) {
-        for (IdType j = 0; j < coo.row->shape[0]; j++) {
-          const auto d = (!COOHasData(coo)) ? j : edges_dat[j];
-          res_dat_data[prefix_elm[i] + j] = d + prefix_elm[i];
+  // 32 is a number obtained from experience. If a user set the grain size
+  // explicitly via env, use that value(`default_grain_size`) instead.
+  size_t grain_size = std::getenv("DGL_PARALLEL_FOR_GRAIN_SIZE")
+                          ? runtime::default_grain_size()
+                          : 32;
```
frozenbugs, line 92: Consider creating a ./include/env_variable.h, and include the header here, instead of calling std::getenv everywhere, which makes the output hard to predict.

```
static const char* kDGLParallelForGrainSize = std::getenv("DGL_PARALLEL_FOR_GRAIN_SIZE");
```

src/array/cpu/disjoint_union.cc
```diff
@@ -85,32 +85,38 @@ COOMatrix DisjointUnionCoo(const std::vector<COOMatrix>& coos) {
...
+  dgl::runtime::parallel_for(
+      0, coos.size(), grain_size, [&](IdType b, IdType e) {
+        for (IdType i = b; i < e; ++i) {
```
frozenbugs, line 95: I think the *for* should indent with **l** in **dgl::runtime**

peizhou001:
> * What is the downside of running out of CPU threads? I feel they will be wasted if they are not fully utilized.
> * Setting the number of threads may depend on the machine. Does the performance degrade if a machine only have 8 cores but we use 32 threads? I'd suggest to set the thread numbers through a system call, such as [hardward_concurrency](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency).

The downside task is graph classification in single machine multiple GPU. A single task in this function is tiny, so using one thread one task is actually wistful. According my test, there is no explicit performance improvement between 4 and 64 threads.

It make sense to use a ratio of the CPU combined with a constant number.

src/array/cpu/disjoint_union.cc
peizhou001: make sense, changed for fix.

src/array/cpu/disjoint_union.cc
peizhou001: fixed.

src/array/cpu/disjoint_union.cc
```diff
@@ -85,32 +85,38 @@ COOMatrix DisjointUnionCoo(const std::vector<COOMatrix>& coos) {
...
-      }
-
-      if (has_data) {
-        for (IdType j = 0; j < coo.row->shape[0]; j++) {
-          const auto d = (!COOHasData(coo)) ? j : edges_dat[j];
-          res_dat_data[prefix_elm[i] + j] = d + prefix_elm[i];
+  // 32 is a number obtained from experience. If a user set the grain size
+  // explicitly via env, use that value(`default_grain_size`) instead.
+  size_t grain_size = std::getenv("DGL_PARALLEL_FOR_GRAIN_SIZE")
+                          ? runtime::default_grain_size()
```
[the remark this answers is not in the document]
peizhou001: Yes, we need a different flag because different number should be given in different scenario.

src/array/cpu/disjoint_union.cc
peizhou001: Not very clear, could you explain more?

frozenbugs, line 93:
```suggestion
  dgl::runtime::parallel_for(
      0, coos.size(), grain_size, [&](IdType b, IdType e) {
    for (IdType i = b; i < e; ++i) {
```

frozenbugs: See the suggestion below.

include/dgl/runtime/parallel_for.h
```diff
@@ -26,11 +27,13 @@ namespace {
...
   size_t grain_size;
 
-  DefaultGrainSizeT() {
-    auto var = std::getenv("DGL_PARALLEL_FOR_GRAIN_SIZE");
+  DefaultGrainSizeT() : DefaultGrainSizeT(1) {}
+
+  explicit DefaultGrainSizeT(size_t default_grain_size) {
+    auto var = dgl::kDGLParallelForGrainSize;
 
     if (!var) {
```
frozenbugs, line 35: nit: avoid if not else, flip it to if else.

frozenbugs (approved):

src/array/cpu/disjoint_union.cc
peizhou001: fixed

include/dgl/runtime/parallel_for.h
peizhou001: fixed

PR #5497, opened 2023-03-24
[Feature] Add bfloat16 support for CPU

itaraban:
## Description

Enabling BFloat16 data type for CPU.
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/include/dgl/aten/macro.h
+++ b/include/dgl/aten/macro.h
@@ -157,7 +157,7 @@
       LOG(FATAL) << (val_name) << " can't be float16 on CPU";               \
     } else if (                                                             \
         XPU == kDGLCPU && (val).bits == 16 && (val).code == kDGLBfloat) {   \
-        typedef BFloat16 FloatType;                                         \
+      typedef BFloat16 FloatType;                                           \
       { __VA_ARGS__ }                                                       \
     } else {                                                                \
       LOG(FATAL) << (val_name)                                              \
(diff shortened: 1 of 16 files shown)


include/dgl/aten/macro.h
```diff
@@ -152,8 +152,13 @@
...
       { __VA_ARGS__ }                                                       \
-    } else if (XPU == kDGLCPU) {                                            \
-      LOG(FATAL) << (val_name) << " can only be float32 or float64 on CPU"; \
+    } else if (                                                             \
+        XPU == kDGLCPU && (val).bits == 16 && (val).code == kDGLFloat) {    \
+      LOG(FATAL) << (val_name) << " can't be float16 on CPU";               \
+    } else if (                                                             \
+        XPU == kDGLCPU && (val).bits == 16 && (val).code == kDGLBfloat) {   \
+        typedef BFloat16 FloatType;                                         \
+      { __VA_ARGS__ }                                                       \
```
frozenbugs, line 161: Can you run lintrunner -a?
I think there are some issue about different new line characters, which we can't see from eyes about lint catches it.

include/dgl/runtime/bfloat16.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include <cmath>
+
+class BFloat16 {
+  uint16_t val;
+
+ public:
+  constexpr BFloat16() : val(0) {}
+  // disabling "explicit" warning in lint
+  // implicit usage on costructor is expected
```
frozenbugs, line 17: nit:
```suggestion
  // Disable lint "explicit" warning, since implicit usage on constructor is
  // expected.
```

src/array/cpu/segment_reduce.cc
```diff
@@ -69,6 +75,14 @@ template void SegmentReduce<kDGLCPU, int64_t, double>(
     const std::string& op, NDArray feat, NDArray offsets, NDArray out,
     NDArray arg);
 
+template<> void ScatterAdd<kDGLCPU, int32_t, BFloat16>(
```
frozenbugs, line 78: Why do we need `<>` here?

src/array/cpu/segment_reduce.cc
```diff
@@ -78,6 +92,18 @@ template void ScatterAdd<kDGLCPU, int32_t, double>(
 template void ScatterAdd<kDGLCPU, int64_t, double>(
     NDArray feat, NDArray arg, NDArray out);
 
+template<> void UpdateGradMinMax_hetero<kDGLCPU, int32_t, BFloat16>(
```
frozenbugs, line 95: ditto.

src/array/cpu/spmm.h
```diff
@@ -51,18 +53,20 @@ void SpMMSumCsrNaive(
     for (auto rid = b; rid < e; ++rid) {
       const IdType row_start = indptr[rid], row_end = indptr[rid + 1];
       DType* out_off = O + rid * dim;
-      for (IdType j = row_start; j < row_end; ++j) {
-        const IdType cid = indices[j];
-        const IdType eid = has_idx ? edges[j] : j;
-        for (int64_t k = 0; k < dim; ++k) {
+      for (int64_t k = 0; k < dim; ++k) {
+        AccType akk = 0.;
```
frozenbugs, line 57: what's `akk`, let's avoid unconventional abbr.

src/array/cpu/spmm.h
```diff
@@ -42,6 +42,8 @@ template <typename IdType, typename DType, typename Op>
 void SpMMSumCsrNaive(
     const BcastOff& bcast, const CSRMatrix& csr, const DType* X, const DType* W,
     DType* O) {
+  typedef typename std::conditional<
+      std::is_same<DType, BFloat16>::value, float, DType>::type AccType;
```
frozenbugs, line 46: Why not make it a file-wide typedef?

include/dgl/runtime/bfloat16.h
itaraban: thanks for suggestion, changed.

include/dgl/aten/macro.h
itaraban: changed

src/array/cpu/segment_reduce.cc
itaraban: To partially specialize template, and avoid creating common template for this function for BFloat16

src/array/cpu/spmm.h
itaraban: thanks, changed

src/array/cpu/spmm.h
itaraban: It depends on DType and appears only in two Naive methods. I don't have good idea right now how to move it in pretty way.

src/array/cpu/segment_reduce.cc
frozenbugs: I don't see difference better this ScatterAdd with the one on L80. Why they are different? Should we be consistent?

src/array/cpu/spmm.h
frozenbugs: Try 
```
template <typename DType>
using AccType = typename std::conditional<std::is_same<DType, BFloat16>::value, float, DType>::type;
```

It should work with c++ 11.

itaraban: thanks, changed

frozenbugs:
There is still one open comment, please address, overall the code style looks great! Thanks.
Add @BarclayII to double check.

frozenbugs:
> There is still one open comment, please address, overall the code style looks great! Thanks. Add @BarclayII to double check.

LGTM from my side, @BarclayII feel free to approve if this PR looks good to you.

include/dgl/runtime/bfloat16.h
```diff
@@ -0,0 +1,68 @@
...
+    return *this;
+  }
+
+  operator float() const {
+    union {
+      float f;
+      uint16_t raw[2];
+    };
+    raw[0] = 0;
+    raw[1] = val;
```
BarclayII, line 63: Is it assuming big endian byte order?  If it's little endian I think this should be
```
raw[0] = val;
```
Also, could we have a test case on converting between bfloat16 and float32?

BarclayII (approved):
LGTM

itaraban:
@dgl-bot

PR #5506, opened 2023-03-28
[Sparse] Support strided tensor when calling old DGL APIs

### LINKED ISSUE #5502: [Sparse] SpMM does not handle strided torch tensor
## 🐛 Bug

SpMM does not handle strided torch tensor, leading to wrong computation result.

## To Reproduce

```python
import dgl.sparse as dglsp
import torch

dev = 'cuda:0'

i = torch.tensor([[1, 0, 1, 2],
                  [3, 0, 2, 1]])
v = torch.arange(4).float() + 1
A = dglsp.spmatrix(i, v, shape=(4, 4))
X = torch.arange(40).reshape(4, 5, 2).float()

print(A.to_dense())
print(X[:,:,0])

print('Correct result')
print(A.to_dense() @ X[:,:,0])

print('This is incorrect.')
print(A @ X[:,:,0])

print('This is correct.')
print(A @ X[:,:,0].clone())
```
Output:
```
tensor([[2., 0., 0., 0.],
        [0., 0., 3., 1.],
        [0., 4., 0., 0.],
        [0., 0., 0., 0.]])
tensor([[ 0.,  2.,  4.,  6.,  8.],
        [10., 12., 14., 16., 18.],
        [20., 22., 24., 26., 28.],
        [30., 32., 34., 36., 38.]])
Correct result
tensor([[  0.,   4.,   8.,  12.,  16.],
        [ 90.,  98., 106., 114., 122.],
        [ 40.,  48.,  56.,  64.,  72.],
        [  0.,   0.,   0.,   0.,   0.]])
This is incorrect.
tensor([[ 0.,  2.,  4.,  6.,  8.],
        [45., 49., 53., 57., 61.],
        [20., 24., 28., 32., 36.],
        [ 0.,  0.,  0.,  0.,  0.]])
This is correct.
tensor([[  0.,   4.,   8.,  12.,  16.],
        [ 90.,  98., 106., 114., 122.],
        [ 40.,  48.,  56.,  64.,  72.],
        [  0.,   0.,   0.,   0.,   0.]])
```

## Expected behavior

Regardless of whether `Tensor.clone()` is called (which converts a strided tensor to a non-strided one), the result should always be the same.

## Environment

 - DGL Version (e.g., 1.0): 1.0.1+cu113
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): torch 1.12.1+cu113
 - OS (e.g., Linux): Linux
 - Python version: 3.7.5
 - CUDA/cuDNN version (if applicable): 11.3

## Additional context

Perhaps many other operators have the same issue.

czkkkkkk:
## Description
Resolve #5502.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/tests/python/pytorch/sparse/test_matmul.py
+++ b/tests/python/pytorch/sparse/test_matmul.py
@@ -13,6 +13,7 @@ from .utils import (
     rand_coo,
     rand_csc,
     rand_csr,
+    rand_stride,
     sparse_matrix_to_dense,
     sparse_matrix_to_torch_sparse,
 )
@@ -30,6 +31,7 @@ def test_spmm(create_func, shape, nnz, out_dim):
     else:
         X = torch.randn(shape[1], requires_grad=True, device=dev)
 
+    X = rand_stride(X)
     sparse_result = matmul(A, X)
     grad = torch.randn_like(sparse_result)
     sparse_result.backward(grad)
@@ -56,6 +58,7 @@ def test_bspmm(create_func, shape, nnz):
     dev = F.ctx()
     A = create_func(shape, nnz, dev, 2)
     X = torch.randn(shape[1], 10, 2, requires_grad=True, device=dev)
+    X = rand_stride(X)
 
     sparse_result = matmul(A, X)
     grad = torch.randn_like(sparse_result)
--- a/tests/python/pytorch/sparse/utils.py
+++ b/tests/python/pytorch/sparse/utils.py
@@ -1,7 +1,7 @@
 import numpy as np
 import torch
 
-from dgl.sparse import from_coo, from_csc, from_csr, SparseMatrix, spmatrix
+from dgl.sparse import from_csc, from_csr, SparseMatrix, spmatrix
 
 np.random.seed(42)
 torch.random.manual_seed(42)
@@ -15,7 +15,7 @@ def clone_detach_and_grad(t):
 
 def rand_stride(t):
     """Add stride to the last dimension of a tensor."""
-    stride = np.random.randint(1, 4)
+    stride = np.random.randint(2, 4)
     ret = torch.stack([t] * stride, dim=-1)[..., 0]
     ret = ret.detach()
     if torch.is_floating_point(t):
--- a/tests/python/pytorch/sparse/utils.py
+++ b/tests/python/pytorch/sparse/utils.py
@@ -1,7 +1,7 @@
 import numpy as np
 import torch
 
-from dgl.sparse import from_coo, from_csc, from_csr, SparseMatrix
+from dgl.sparse import from_coo, from_csc, from_csr, SparseMatrix, spmatrix
 
 np.random.seed(42)
 torch.random.manual_seed(42)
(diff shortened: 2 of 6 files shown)


jermainewang:
1. Double check all other operators.
2. What is the best practice moving forward? Shall we have a safe guard utility to check/handle strided cases?
3. Add proper unit tests. I'm curious if ChatGPT can help us convert our UTs automatically.

frozenbugs:
+1 for unit test, for all operators, we should have a unit test for strided cases. Is there any other cases / types of tensor.

czkkkkkk:
> * Double check all other operators.
> * What is the best practice moving forward? Shall we have a safe guard utility to check/handle strided cases?

I added `.contiguous` when converting a Pytorch tensor to a DGL array. Then we don't need to handle strided tensors in any operator.

> * Add proper unit tests. I'm curious if ChatGPT can help us convert our UTs automatically.
Added. The github copilot provides useful hints when writing tests.

czkkkkkk:
> Is there any other cases / types of tensor.

I don't think so. Stride is a very special case.

tests/python/pytorch/sparse/test_matmul.py
```diff
@@ -30,6 +31,7 @@ def test_spmm(create_func, shape, nnz, out_dim):
     else:
         X = torch.randn(shape[1], requires_grad=True, device=dev)
 
+    X = rand_stride(X)
```
jermainewang, line 34: This means we are only testing strided tensor while not testing normal tensors, though I feel it is a minor issue.

jermainewang (approved):

tests/python/pytorch/sparse/test_matmul.py
czkkkkkk: Actually the stride is a random number in range [1, 4]. It is possible that the stride is one.

tests/python/pytorch/sparse/utils.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 import torch
 
-from dgl.sparse import from_coo, from_csc, from_csr, SparseMatrix
+from dgl.sparse import from_coo, from_csc, from_csr, SparseMatrix, spmatrix
```
frozenbugs, line 4: remove from_coo

tests/python/pytorch/sparse/test_matmul.py
jermainewang: Oh, we should avoid randomness in unit tests as much as possible. Otherwise, it is possible that some developers got lucky but later others fail the CI pipeline due to non-related errors. Please fix this.

jermainewang (changes requested):
Just realized that the stride is a random int from 1 to 4. This is bad practice because stride=1 means no stride. Please fix the number in UT.

czkkkkkk:
> Just realized that the stride is a random int from 1 to 4. This is bad practice because stride=1 means no stride. Please fix the number in UT.

Done

jermainewang (approved):

PR #5519, opened 2023-04-03
[DO NOT MERGE][Sparse] Support spsp_mul on CPU.

czkkkkkk:
## Description

This PR is one of the work items in #5368. 

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/dgl_sparse/src/cpu/matrix_ops_impl.cc
+++ b/dgl_sparse/src/cpu/matrix_ops_impl.cc
@@ -1,9 +1,10 @@
 /**
  *  Copyright (c) 2023 by Contributors
- * @file cpu/matrix_ops_impl_cpu.cc
- * @brief DGL C++ implementation of matrix operators.
+ * @file cpu/matrix_ops_impl.cc
+ * @brief DGL CPU implementation of matrix operators.
  */
 #include "../matrix_ops_impl.h"
+
 #include "../utils.h"
 
 namespace dgl {
@@ -33,8 +34,9 @@ COOIntersectionImpl(
   };
   std::unordered_map<std::pair<IdType, IdType>, IdType, decltype(pair_hash_fn)>
       pair_map(lhs_len, pair_hash_fn);
-  for (int64_t k = 0; k < lhs_len; ++k)
+  for (int64_t k = 0; k < lhs_len; ++k) {
     pair_map.emplace(std::make_pair(lhs_row_ptr[k], lhs_col_ptr[k]), k);
+  }
 
   for (int64_t i = 0; i < rhs_len; ++i) {
     const IdType row_id = rhs_row_ptr[i], col_id = rhs_col_ptr[i];
--- a/python/dgl/sparse/elementwise_op_sp.py
+++ b/python/dgl/sparse/elementwise_op_sp.py
@@ -90,8 +90,8 @@ def sp_sub(A: SparseMatrix, B: SparseMatrix) -> SparseMatrix:
 def sp_mul(A: SparseMatrix, B: Union[SparseMatrix, Scalar]) -> SparseMatrix:
     """Elementwise multiplication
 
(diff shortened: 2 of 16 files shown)


czkkkkkk:
@dgl-bot

python/dgl/sparse/elementwise_op_sp.py
```diff
@@ -83,8 +90,8 @@ def sp_sub(A: SparseMatrix, B: SparseMatrix) -> SparseMatrix:
 def sp_mul(A: SparseMatrix, B: Union[SparseMatrix, Scalar]) -> SparseMatrix:
     """Elementwise multiplication
 
-    If :attr:`B` is a sparse matrix, both :attr:`A` and :attr:`B` must be
-    diagonal matrices.
+    Note that if both :attr:`A` and :attr:`B` are sparse matrices, it currently
+    only support multiplication on CPU.
```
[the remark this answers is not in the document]
czkkkkkk: Not a regression. If both matrices are diagonal, it can run on GPU. I changed this description.

frozenbugs (approved):

dgl_sparse/src/cpu/matrix_ops_impl.cc
```diff
@@ -0,0 +1,69 @@
...
+
+template <c10::DeviceType XPU, typename IdType>
+std::tuple<std::shared_ptr<COO>, torch::Tensor, torch::Tensor>
+COOIntersectionImpl(
+    const std::shared_ptr<COO> &lhs, const std::shared_ptr<COO> &rhs) {
+  const int64_t lhs_len = lhs->indices.size(1);
+  const int64_t rhs_len = rhs->indices.size(1);
+
+  torch::Tensor lhs_row = lhs->indices.index({0}).contiguous();
+  const IdType *lhs_row_ptr = static_cast<IdType *>(lhs_row.data_ptr());
```
jermainewang, line 21: Could provide a utility function to get data pointers, e.g., `const IdType *lhs_row_ptr = utils::GetPtr<IdType>(lhs->indices.index({0}))`.

dgl_sparse/src/cpu/matrix_ops_impl.cc
```diff
@@ -0,0 +1,69 @@
...
+  std::vector<IdType> ret_lhs_indices, ret_rhs_indices;
+
+  auto pair_hash_fn = [](const std::pair<IdType, IdType> &pair) {
+    return std::hash<IdType>()(pair.first) ^ std::hash<IdType>()(pair.second);
+  };
+  std::unordered_map<std::pair<IdType, IdType>, IdType, decltype(pair_hash_fn)>
+      pair_map(lhs_len, pair_hash_fn);
+  for (int64_t k = 0; k < lhs_len; ++k) {
+    pair_map.emplace(std::make_pair(lhs_row_ptr[k], lhs_col_ptr[k]), k);
+  }
```
jermainewang, line 39: If we don't consider operator efficiency. We can simply do the following (inspired by https://discuss.pytorch.org/t/intersection-between-to-vectors-tensors/50364/10?u=minjie_wang)

1. Hash the integer pairs to one integer (e.g., by `i * N + j` where `N` is the number of rows/columns).
2. Concatenate lhs and rhs into one vector.
3. Compute unique and the unique indices.
4. Unhash the integer to integer pairs and recover the indices.

This way you may get an operator that supports both CPU/GPU.

czkkkkkk: It is doable and not very slow. I think we can use this implementation. Then do you think we need to keep the current code structure for adding a new operator?

jermainewang:
A new implementation was merged in #5543 . We leave this PR as a future reference for adding custom C++ sparse operators.

mufeili:
PR closed as it's only for reference.

Issue #5522, opened 2023-04-05
Add multi-machine support for DDP dataloader


chang-l:
## 🚀 Feature

Currently, DGL dataloader only works with DDP in a single host, because dataloader uses shared memory to store indices like below: https://github.com/dmlc/dgl/blob/a244de579cbd964912aa46446a0a9b2a2f595c87/python/dgl/dataloading/dataloader.py#L262

The DGL dataloader currently does not support multiprocessing in a real distributed system, such as a multi-node architecture, unlike PyTorch's dataloader which offers this capability. While DistDGL provides a solution, it can pose a significant challenge for users due to its high level of complexity in code-base/setup/run.

I think that removing the dependency of DGL's dataloader on shared memory would be beneficial because: (1) it would improve the completeness of the dataloader, just like PyTorch; (2) it would allow users to scale up the number of GPUs at other machines without having to rely on DistDGL.

## Motivation

Extend the current dataloader (DDP) to support multi-node arch, just like PyTorch.

## Alternatives

DistDGL

## Pitch

In cases where the graph and features can fit into host memory, the DGL's dataloader should allow the multi-GPU example code to run smoothly on multi-node architectures with minimal modifications to the training script (should just be the setting in 
 `dist.init_process_group`)

For example, currently, if we run the following at each host:
```
torchrun  --nproc_per_node=2 --nnodes=2 --node_rank=xx --master_addr=XXX --master_port=1234  multi_gpu_node_classification.py --gpu 0,1,2,4

```
it would fail and print:
```
  File "multi_gpu_node_classification.py", line 118, in train
    train_dataloader = DataLoader(
  File "/usr/local/lib/python3.8/dist-packages/dgl/dataloading/dataloader.py", line 836, in __init__
    self.dataset = create_tensorized_dataset(
  File "/usr/local/lib/python3.8/dist-packages/dgl/dataloading/dataloader.py", line 562, in create_tensorized_dataset
    return DDPTensorizedDataset(indices, batch_size, drop_last, ddp_seed, shuffle)
  File "/usr/local/lib/python3.8/dist-packages/dgl/dataloading/dataloader.py", line 214, in __init__
    self._indices = call_once_and_share(
  File "/usr/local/lib/python3.8/dist-packages/dgl/multiprocessing/pytorch.py", line 104, in call_once_and_share
    result = get_shared_mem_array(name, shape, dtype)
[...]

PR #5523, opened 2023-04-05
[Feature] Add multi-machine support for dataloader

chang-l:
## Description

This is a draft PR to address https://github.com/dmlc/dgl/issues/5522. 

## Checklist
- [ ] Code is well-documented

## Changes

- In `multiprocessing/pytorch.py`
Instead of the root process, process with local_rank=0 at every host creates the shared memory. 

- In `dataloading/dataloader.py`
Broadcast the shuffling seed; and create the identical index array in the shared memory at every machine.


mufeili:
@chang-l Do you have any reviewers in mind?

chang-l:
@mufeili  Thanks for reminding. I've added reviewers.  Please feel free to add others.

chang-l:
Currently, the main concerns are:

(1) this PR relies env variable `LOCAL_RANK` to decide if one process needs to allocate shared memory and  shuffle the indices or not, i.e., only the process with `LOCAL_RANK=0` will allocate/shuffle the array at its system shared memory.
The `LOCAL_RANK` variable is set by PyTorch's distributed package. When users launch a multiprocessing job with `torchrun/torch.distributed.launch`, env variable `LOCAL_RANK` will be available and usually used to set the cuda device id. 

So, there is a chance that when users change `LOCAL_RANK` for whatever reason, it may break the code...

(2) In dataloader, it seems we only use shared memory for indices, which should not consume too much memory... would it be possible to just use local memory (at process) to store the indices? It will create many copies of indices at each process, but it may be acceptable...

(3) not sure how PyTorch resolves this kind of issue (DDP dataloader in multiple machines). I need to have another look at PyTorch's dataloader implementation...

I need your inputs @BarclayII @jermainewang @nv-dlasalle @Rhett-Ying   Thanks!

nvcastet:
The assumption of `LOCAL_RANK` being set will make it hard to be portable in any environment out of the box. pytorch does not assume this variable is set in their distributed stack from what I know.

> (2) In dataloader, it seems we only use shared memory for indices, which should not consume too much memory... would it be possible to just use local memory (at process) to store the indices? It will create many copies of indices at each process, but it may be acceptable...

Why many copies of indices at each process and not just one per process instead of a global one?

chang-l:
> Why many copies of indices at each process and not just one per process instead of a global one?

@nvcastet Sry, it was my typo and I meant many copies of indices as each process needs to keep one.

chang-l:
I dig a little deeper and found some heuristics/background behind this issue... It seems to me that there are four alternatives when shuffling the indices with multi-GPU (DDP):

**- Alternative I:  Not using shared memory at all.** 
This is how PyTorch shuffle the dataset's indices. which causes each worker process to maintain a complete copy of indices, resulting in a total memory consumption of **`O(N * Ndevice * Nworkers)`** bytes when `persistent_workers=False`. It easily becomes intractable as the number of edges/nodes increases (e.g., with 1B edges, 4 devices per node, and `num_worker=4`, it consumes 128G RAM) @nvcastet 

**- Alternative II:  Using [`torch.shared_memory_()`](https://pytorch.org/docs/stable/generated/torch.Tensor.share_memory_.html) on each training (parent) process to share with its worker processes**
This approach can be seen as a compromise, where it still duplicates indices but bounds to **`O(N * Ndevice)`** bytes at each host. It could still be memory inefficient and hurts the scalability, especially for GNN applications. See this issue https://github.com/dmlc/dgl/issues/3357 and this PR https://github.com/dmlc/dgl/pull/3614
@nv-dlasalle @yaox12 

**- Alternative III: Using shared memory at each host/node and sync through backend distributed pkg (this PR)**
This is based on the current implementation in DGL which only works for single machine. Ideally, in future, PyTorch can provide a different share memory storage that can support `torch.distributed.launch`, like this [RFC](https://github.com/pytorch/pytorch/issues/64932). But it is seems backlogged and no est. time yet.

**- Alternative IV: Decomposing indices to each training process and shared with its worker processes**
This approach also addresses this issue https://github.com/dmlc/dgl/issues/3357 but meanwhile is also compatible with multi-node architecture. Any reasons why this was not considered earlier? @BarclayII @nv-dlasalle @yaox12

Rhett-Ying:
Hi @chang-l , as for your **Alternative IV**, do you mean to split `indices` into `num_nodes * num_devices * num_workers` parts and each part has no intersection?

nvcastet:
Thanks Chang for doing a list of alternatives. I agree that doing *Alternative I* does not make sense since even native PyTorch uses shared memory when the training process (one per device) shares the indices with its dataloader workers.
*Alternative III* runs into the problem of assuming an environmental variable being set for the local rank which native PyTorch does not require to scale from one node to multi-node.
*Alternative IV* seems the most promising in my opinion, I do not think we need to split indices to the level of dataloader workers but simply to the number of training processes (which is equals to the number of devices). Then each device process shuffle N/Ndevices indices and feeds its dataloader worker(s) (only when `num_workers > 0`).

chang-l:
@Rhett-Ying 
Yeah, I meant decompose the indices to each training process (`num_nodes * num_devices_per_node` in this context).

PR #5532, opened 2023-04-07
[Bug] Merge rather replace when find reverse eids

peizhou001:
## Description
When calling `as_edge_prediction_sampler` with `exclude=reverse_types`, both original eids and reverse eids should be removed from latest MFG, but currently only  reverse eids are removed, so this PR is used to fix this bug.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/dataloading/base.py
+++ b/python/dgl/dataloading/base.py
@@ -281,17 +281,9 @@ def _find_exclude_eids_with_reverse_types(g, eids, reverse_etype_map):
     }
     exclude_eids.update(
         {
-            v: F.unique(
-                F.cat(
-                    (
-                        exclude_eids[k],
-                        exclude_eids.get(
-                            v, F.tensor([], dtype=exclude_eids[k].dtype)
-                        ),
-                    ),
-                    dim=0,
-                )
-            )
+            v: F.unique(F.cat((exclude_eids[k], exclude_eids[v]), dim=0))
+            if v in exclude_eids
+            else exclude_eids[k]
             for k, v in reverse_etype_map.items()
             if k in exclude_eids
         }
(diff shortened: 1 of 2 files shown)


Rhett-Ying:
@dgl-bot

python/dgl/dataloading/base.py
```diff
@@ -281,9 +281,11 @@ def _find_exclude_eids_with_reverse_types(g, eids, reverse_etype_map):
...
     exclude_eids.update(
         {
-            reverse_etype_map[k]: v
-            for k, v in exclude_eids.items()
-            if k in reverse_etype_map
+            v: F.unique(F.cat((exclude_eids[k], exclude_eids[v]), dim=0))
+            if v in exclude_eids
+            else exclude_eids[k]
+            for k, v in reverse_etype_map.items()
+            if k in exclude_eids
```
BarclayII, line 288: Hmm...  This comprehension looks too complicated.  I guess it's more readable to decompose it into a for loop.
```python
for k, v in reverse_etype_map.items():
    if k in exclude_eids:
        exclude_eids[v] = ...
```

cc @frozenbugs

BarclayII (approved):
I don't have any problems other than the comment above.

python/dgl/dataloading/base.py
peizhou001: It make sense, change for fix.

python/dgl/dataloading/base.py
```diff
@@ -279,13 +279,14 @@ def _find_exclude_eids_with_reverse_types(g, eids, reverse_etype_map):
...
         for k, v in reverse_etype_map.items()
     }
-    exclude_eids.update(
-        {
-            reverse_etype_map[k]: v
-            for k, v in exclude_eids.items()
-            if k in reverse_etype_map
-        }
-    )
+    for k, v in reverse_etype_map.items():
```
frozenbugs, line 282: I am not sure whether the _find_exclude_eids_with_reverse_types clearly states what inside the method.

peizhou001: yes, it is exactly what the function do.

Issue #5509, opened 2023-03-29
[Dataset] MovieLens


LspongebobJH:
## 🚀 Feature

MovieLens is a heterogeneous rating graph, assembled by GroupLens Research from the https://movielens.org/. This dataset provides user ratings on movies with side information of users and movies. `users` and `movies` are modeled by nodes and `rates` are ground truth labels for edges between `users` and `movies`. The task on movielens is usually edge(relation) classification. There are two prevalent (sub-)datasets,  movielens-100k and movielens-1m, used by four examples in dgl, IGMC, GCMC, TAHIN and PinSAGE. A brief summary of these two datasets is as follows (statistics from IGMC paper):
| Dataset | Users | Items | Ratings | Density | Rating types |
| :---: | :---: | :---: | :---: | :---: | :---: |
| ML-100K | 943 | 1,682 | 100,000 | 0.0630 |1, 2, 3, 4, 5|
| ML-1M | 6,040 | 3,706 | 1,000,209 | 0.0447 |1, 2, 3, 4, 5|

The specific usage of movielens in IGMC, GCMC, TAHIN, PinSAGE and implementations of PyG are slightly different. A simple comparison is as follows:
(Assume that the default mode is to construct initial node features with side information, and the only relation in the dataset is user-movie)

IGMC: Side information is discarded. Initial node features are constructed by a local one-hot encoding. The one-hot encoding is used to differentiate users and movies, and reflect distance between neighbors with the target user-movie pair. 
GCMC: default
TAHIN: Side information is used to construct heterogenous nodes. For example, relations provided by TAHIN involve user-age, user-occupation, movie-genre, etc, where age, occupation and genre are side information of users and movies.
PinSAGE: default
PyG: default

There are some slight differences among examples above. These differences mainly consist in 1) how to process/embed text features, such as movie titles, user occupations, etc. 2) using which dataset. 
1) For example, GCMC utilizes Glove embeddings, PinSAGE suggest using FastText embeddings, PyG uses SentenceTransformer API to load pre-trained Transformer from HuggingFace and embed texts with the selected Transformer.
[...]

PR #3606, opened 2021-12-23
[Model] Simplify IGMC

LspongebobJH:
## Description

Simplify IGMC codes

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage
- [x] Code is well-documented
- [x] To the best of my knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

## Changes
- [x] Simplify IGMC codes

cc @mufeili

PR #5567, opened 2023-04-18
[Dataset] MovieLens

LspongebobJH:
## Description

This PR is to provide an unified API for MovieLens datasets. 
- This PR is proposed due to the issue[1].
- The API is based on the IGMC implementations of MovieLens dataset[2]. 
- The text preprocessing steps follow the implementation of GCMC[3].
- Note that this API is not compatible with TAHIN[4], since TAHIN has completely different logic of preprocessing MovieLens dataset. Refer to issue[1] for more details.

## Checklist
- [x] Code is well-documented

## Reference
[1] Related issue: [[Dataset] MovieLens](https://github.com/dmlc/dgl/issues/5509#issue-1645311160)
[2] Related PR: [[Model] Simplify IGMC](https://github.com/dmlc/dgl/pull/3606#issue-1087590263)
[3] Related example: [GCMC example](https://github.com/dmlc/dgl/tree/master/examples/pytorch/gcmc)
[4] Related example: [TAHIN example](https://github.com/dmlc/dgl/tree/master/examples/pytorch/TAHIN)

cc @mufeili


mufeili:
Try following the practice in https://github.com/dmlc/dgl/pull/5511.

LspongebobJH:
> Try following the practice in #5511.

Ok

LspongebobJH:
Hi @mufeili ,
I've already refactored codes following the practice in https://github.com/dmlc/dgl/pull/5511

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,630 @@
...
+import dgl
+from dgl.data import DGLDataset
+
+from dgl.data.utils import (
+    download,
+    load_graphs,
+    load_info,
+    save_graphs,
+    save_info,
+)
```
mufeili, line 21: Use relative imports to be consistent with the current DGL practice.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,630 @@
+"""MovieLens dataset"""
+
+import os
+import re
+import zipfile
+
+import numpy as np
+import pandas as pd
+import torch as th
+from sklearn.model_selection import train_test_split
```
mufeili, line 10: I don't think `sklearn` is currently a dependency of DGL. See if you can use the utils from https://github.com/dmlc/dgl/blob/master/python/dgl/data/utils.py instead.

mufeili:
@dgl-bot

python/dgl/data/movielens.py
LspongebobJH: This issue and the one below are solved in the latest commit. There is one point. I was trying to make movielens support multiple platforms. However, I use torchtext to preprocess text embeddings following GCMC practice. This may lead to movielens depending on PyTorch since torchtext might depend on PyTorch.

python/dgl/data/movielens.py
LspongebobJH: the same as above.

python/dgl/data/movielens.py
LspongebobJH: Another question is, I'm not quite familiar with unit test of dgl if we need to test movielens (I write a test script for movielens following actor dataset PR. But I didn't know how to test it). Could you guide me to some documents of the unit test pipeline in dgl?

mufeili: For the first issue, it's fine to only support PyTorch. You can follow the practice in https://github.com/dmlc/dgl/blob/master/python/dgl/homophily.py

mufeili: For a local test, you can simply build DGL from source and test the code snippet. You don't necessarily need to launch the unit test pipeline locally.

python/dgl/data/movielens.py
mufeili: I didn't get it. Could you elaborate more?

mufeili:
@dgl-bot

python/dgl/data/movielens.py
mufeili: got it

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,646 @@
...
+    "Thriller",
+    "War",
+    "Western",
+]
+GENRES_ML_1M = GENRES_ML_100K[1:]
+GENRES_ML_10M = GENRES_ML_100K + ["IMAX"]
+
+
+class MovieLensDataset(DGLDataset):
+    r"""MovieLens dataset.
```
mufeili, line 50: Is there a paper link?

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,646 @@
...
+
+    The datasets consist of user ratings for movies and incorporate additional user/movie information in the form of features. 
+    The data are stored in undirected bipartite graph structure, where nodes represent users and movies, edges store ratings
```
mufeili, line 53:
```suggestion
    The nodes represent users and movies, and the edges store ratings
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,646 @@
...
+    that users assign to movies.
+
+
```
mufeili, line 56: remove a line

python/dgl/data/movielens.py
LspongebobJH: Yes. I add reference and related website link in the latest commit.

python/dgl/data/movielens.py
LspongebobJH: > For the first issue, it's fine to only support PyTorch. You can follow the practice in https://github.com/dmlc/dgl/blob/master/python/dgl/homophily.py

In the lastest commit, I've already follow the practice in the link to update PyTorch support.

LspongebobJH: > For a local test, you can simply build DGL from source and test the code snippet. You don't necessarily need to launch the unit test pipeline locally.

I've already built DGL from source and briefly test MovieLens. Seems MovieLens work well. 
```python
import dgl

print(dgl.__version__)

from dgl.data import MovieLensDataset

dataset = MovieLensDataset('ml-100k', valid_ratio=0.2, test_ratio=0.1)
train_g, valid_g, test_g = dataset[0]
info, feat = dataset.info, dataset.feat

train_ratings, valid_ratings, test_ratings = \
    info['train_rating_pairs'], info['valid_rating_pairs'], info['test_rating_pairs']
user_feat, movie_feat = \
    feat['user_feat'], feat['movie_feat']

pass
# I have tested the shape of variables above, and some statistics of graphs, 
# such as number of nodes and edges. No error.
```

python/dgl/data/movielens.py
mufeili: Follow the practice in https://github.com/dmlc/dgl/pull/5511

mufeili:
@dgl-bot

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,658 @@
...
+    -----
+    - When using MovieLens, users are required to install :obj:`torchtext>=0.15.1`, which provides :obj:`Glove` API to preprocess text embeddings.
+    - The number of edges is doubled to form an undirected(bidirected) graph structure.
+    - Each time when the dataset is loaded with a different setting of training, validation and testing ratio, the param
+    :obj:`"force_reload"` should be set to :obj:`"True"`. Otherwise the previous split of dataset will be loaded and not be
+    updated in the storage.
+
+    Examples
+    --------
+    >>> dataset = MovieLens()
```
mufeili, line 144:
```suggestion
    >>> dataset = MovieLensDataset()
```

mufeili, line 144: This should be `MovieLensDataset`. Add code for importing the class. Make sure that users can run this example without errors.

python/dgl/data/movielens.py
LspongebobJH: fix in the latest commit

python/dgl/data/movielens.py
LspongebobJH: fix in the latest commit

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,678 @@
...
+
+def check_pytorch():
+    """Check if PyTorch is the backend."""
+    if HAS_TORCH is False:
+        raise ModuleNotFoundError(
+            "MovieLensDataset requires PyTorch to be the backend."
+        )
+
+class MovieLensDataset(DGLDataset):
+    r"""MovieLens[1] dataset for edge prediction tasks. The raw datasets are extracted from
```
mufeili, line 63:
```suggestion
    r"""MovieLens dataset for edge prediction tasks. The raw datasets are extracted from 
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,678 @@
...
+            user_features = np.zeros(
+                shape=(user_data.shape[0], 1), dtype=np.float32
+            )
+        else:
+            raise NotImplementedError
+        return user_features
+
+    def _process_movie_fea(self, movie_data):
+        import torchtext
+        from torchtext.data.utils import get_tokenizer
```
mufeili, line 498: Is it possible to pre-extract the embeddings to some files and upload all data to DGL's s3 bucket? In other words, we can use our own URLs rather than the raw URLs.

LspongebobJH: Yes. But I think we still need the tokenizer of torchtext.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,678 @@
...
+
+        # Old torchtext-legacy API commented below
+        # TEXT = torchtext.legacy.data.Field(tokenize='spacy', tokenizer_language='en_core_web_sm')
+        tokenizer = get_tokenizer(
+            "spacy", language="en_core_web_sm"
+        )  # new API (torchtext 0.9+)
+        embedding = torchtext.vocab.GloVe(
+            name="840B", dim=300, cache=self.text_cache
+        )
+
```
LspongebobJH, line 508: @mufeili There's one more issue that I suddenly found. torchtext tokenizer (spacy) requires users to download something in advance through `python -m spacy download en_core_web_sm`. I will check later what it is exactly. I think users would find it inconvenient.

python/dgl/data/movielens.py
mufeili: Why is this the case? You can pre-process all features. After that there is no need for tokenizer and torchtext.

LspongebobJH: I'm a little confused. Can we just preprocess graphs and features in advance, upload them to s3, and let API download them when users use this dataset? If true, API would not process raw data the first time users use it. Then do we need the function process() if no preprocessing is required at the user side? Also, I suppose that implementations of all datasets can follow this paradigm, right? So when do we need process()?

mufeili: First, the graphs and features can be stored in various forms. To handle different forms of data, it can still be beneficial to have a `process` function.

Second, this can be a trade-off. When we expose all pre-processing code, the implementation is less opaque and users can understand how the dataset is processed. On the other hand, this can introduce complex dependencies as in this case.

LspongebobJH: Ok. I got your point. Here only the movie features require torchtext. I think we can preprocess movie features, upload them to s3, remove those preprocessing codes of movie features from the API, and leave other parts unchanged. What do you think?

mufeili: Let's first upload all files including the preprocessed features to s3 and then walk through the rest implementation. Ping me offline if you need help in uploading the files.

LspongebobJH: I have updated the code logic. Movie features would be extracted from somewhere, such as s3. Other code logics are kept unchanged. In the latest commit, I assume that movie features are stored in a certain source folder, and will be copied to the folder storing preprocessed stuff. The latest commit remove dependencies on some third-party frameworks, such as torchtext.

.gitignore
```diff
@@ -168,3 +168,6 @@ config.cmake
 
 # model file
 *.pth
+
+# third_parth modules
+third_party/*
```
mufeili, line 173: Which logic introduced this?

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    def __init__(
+        self,
+        name,
+        valid_ratio,
+        test_ratio=None,
+        raw_dir=None,
+        force_reload=None,
+        verbose=None,
+        transform=None,
```
mufeili, line 195: Since data splitting is random, add an optional argument for fixing the random seed.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    - Ratings: 10,000,054 (0.5, 1, 1.5, ..., 4.5, 5.0)
+
+    Parameters
+    ----------
+    name: str
+        Dataset name. (:obj:`"ml-100k"`, :obj:`"ml-1m"`, :obj:`"ml-10m"`). 
+    valid_ratio: int
+        Ratio of validation samples out of the whole dataset. Should be in (0.0, 1.0).
+    test_ratio: int
+        Ratio of testing samples out of the whole dataset. Should be in (0.0, 1.0) for ml-1m and ml-10m. This parameter is invalid
```
mufeili, line 97: I assume `valid_ratio` + `test_ratio` need to be smaller than 1?

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+        when :obj:`name` is :obj:`"ml-100k"`, since it testing samples are pre-specified.
```
mufeili, line 98:
```suggestion
        when :obj:`name` is :obj:`"ml-100k"`, since its testing samples are pre-specified.
```

python/dgl/data/movielens.py
mufeili, line 97: Is this required when name is `ml-1m` or `ml-10m`?

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    - Movies: 10,677
+    - Ratings: 10,000,054 (0.5, 1, 1.5, ..., 4.5, 5.0)
+
+    Parameters
+    ----------
+    name: str
+        Dataset name. (:obj:`"ml-100k"`, :obj:`"ml-1m"`, :obj:`"ml-10m"`). 
+    valid_ratio: int
+        Ratio of validation samples out of the whole dataset. Should be in (0.0, 1.0).
+    test_ratio: int
```
mufeili, line 96:
```suggestion
    test_ratio: int, optional
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+        Ratio of testing samples out of the whole dataset. Should be in (0.0, 1.0) for ml-1m and ml-10m. This parameter is invalid
+        when :obj:`name` is :obj:`"ml-100k"`, since it testing samples are pre-specified.
+        Default: None
+    raw_dir : str
```
mufeili, line 100:
```suggestion
    raw_dir : str, optional
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+        Raw file directory to download/contain the input data directory.
```
mufeili, line 101:
```suggestion
        Raw file directory to download/store the input data directory.
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+        Default: ~/.dgl/
+    force_reload : bool
```
mufeili, line 103:
```suggestion
    force_reload : bool, optional
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+        Whether to reload the dataset. Default: False
```
mufeili, line 104:
```suggestion
        Whether to re-download and -process the dataset. Default: False
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    verbose : bool
```
mufeili, line 105:
```suggestion
    verbose : bool, optional
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+        Whether to print out progress information. Default: True.
```
mufeili, line 106:
```suggestion
        Whether to print progress information. Default: True.
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    reverse_edge : bool
+        Whether to add reverse edges in graph. Default: True.
```
mufeili, line 108: Do you ever need `reverse_edge=False`? If not, remove it. Also currently there isn't an argument for this below.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+        transformed before every access.
+
+    Attributes
+    ----------
+    train_graph: dgl.DGLGraph
+        Training graph
+    valid_graph: dgl.DGLGraph
+        Validation graph
+    test_graph: dgl.DGLGraph
+        Testing graph
```
mufeili, line 121: I suggest not exposing `train_graph`, `valid_graph`, and `test_graph`. Just let users do

```
train_g, val_g, test_g = dataset[0]
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    user_feat: torch.Tensor
+        User features
+    movie_feat: torch.Tensor
+        Movie features
```
mufeili, line 125: Why not store these in `train_graph`, `valid_graph`, `test_graph`? What are their shapes?

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    info: dict
+        Store training, validation and testing rating pairs.
```
mufeili, line 127: Also return this with `dataset[0]`?

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+
+    Notes
+    -----
+    - The number of edges is doubled to form an undirected(bidirected) graph structure.
+    - Each time when the dataset is loaded with a different setting of training, validation and testing ratio, the param
+    :obj:`"force_reload"` should be set to :obj:`"True"`. Otherwise the previous split of dataset will be loaded and not be
+    updated in the storage.
```
mufeili, line 134: How much time does it take for processing this dataset? If this is very fast, then it's better to simply re-process from scratch every time.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+        User features
+    movie_feat: torch.Tensor
+        Movie features
+    info: dict
+        Store training, validation and testing rating pairs.
+
+    Notes
+    -----
+    - The number of edges is doubled to form an undirected(bidirected) graph structure.
+    - Each time when the dataset is loaded with a different setting of training, validation and testing ratio, the param
```
mufeili, line 132:
```suggestion
    - Each time when the dataset is loaded with a different setting of training, validation and test ratio, the param
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    :obj:`"force_reload"` should be set to :obj:`"True"`. Otherwise the previous split of dataset will be loaded and not be
```
mufeili, line 133:
```suggestion
    :obj:`force_reload` should be set to :obj:`True`. Otherwise the previous split of the dataset will be loaded and the new split will not come into effect.
```

Remove L134.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    updated in the storage.
+
+    Examples
+    --------
+    >>> from dgl.data import MovieLensDataset
+    >>>
```
mufeili, line 139: You don't need ">>>" for empty lines.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+
+    Examples
+    --------
+    >>> from dgl.data import MovieLensDataset
+    >>> 
+    >>> dataset = MovieLensDataset(name='ml-100k', valid_ratio=0.2)
+    >>> train_g, valid_g, test_g = dataset[0]
+    >>> train_g
+    Graph(num_nodes=2625, num_edges=128000,
+        ndata_schemes={}
```
mufeili, line 144:
```suggestion
         ndata_schemes={}
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+        edata_schemes={'_ID': Scheme(shape=(), dtype=torch.int64), 'etype': Scheme(shape=(), dtype=torch.float32)})
```
mufeili, line 145:
```suggestion
         edata_schemes={'_ID': Scheme(shape=(), dtype=torch.int64), 'etype': Scheme(shape=(), dtype=torch.float32)})
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    >>>
+    >>> # get ratings of edges in the training graph
+    >>> ratings = train_g.edata['etype']
```
mufeili, line 148: Why not replace etype with rate/rating then?

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    >>> ratings
+    tensor([3., 3., 2.,  ..., 4., 4., 2.])
+    >>>
+    >>> # get training, validation and testing rating pairs
+    >>> train_rating, valid_rating, test_rating = \
+    ...     dataset.info['train_rating_pairs'], dataset.info['valid_rating_pairs'], dataset.info['test_rating_pairs']
```
mufeili, line 154: What's the shape of train_rating_pairs? What are the semantic meanings of the dimensions?

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    >>> train_rating[0] # node index of users in training rating pairs
+    tensor([614, 772, 531,  ..., 674, 639, 740])
+    >>> train_rating[1] # node index of movies in training rating pairs
+    tensor([1236,  954, 1487,  ..., 1842, 1631, 1168])
```
mufeili, line 158:
```suggestion
    tensor([1236, 954, 1487, ..., 1842, 1631, 1168])
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    >>> # get ratings of edges in the training graph
+    >>> ratings = train_g.edata['etype']
+    >>> ratings
+    tensor([3., 3., 2.,  ..., 4., 4., 2.])
+    >>>
+    >>> # get training, validation and testing rating pairs
+    >>> train_rating, valid_rating, test_rating = \
+    ...     dataset.info['train_rating_pairs'], dataset.info['valid_rating_pairs'], dataset.info['test_rating_pairs']
+    >>> train_rating[0] # node index of users in training rating pairs
+    tensor([614, 772, 531,  ..., 674, 639, 740])
```
mufeili, line 156:
```suggestion
    tensor([614, 772, 531, ..., 674, 639, 740])
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+            [0.9600, 1.0000, 0.0000,  ..., 0.0000, 0.0000, 0.0000],
+            [0.4400, 0.0000, 0.0000,  ..., 0.0000, 0.0000, 0.0000]])
+
+    """
+
+    _url = {
+        "ml-100k": "http://files.grouplens.org/datasets/movielens/ml-100k.zip",
+        "ml-1m": "http://files.grouplens.org/datasets/movielens/ml-1m.zip",
+        "ml-10m": "http://files.grouplens.org/datasets/movielens/ml-10m.zip",
+    }
```
mufeili, line 185: I assume these will be replaced by our s3 urls?

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+try:
+    import torch
+except ImportError:
+    HAS_TORCH = False
+else:
+    HAS_TORCH = True
+
+def check_pytorch():
+    """Check if PyTorch is the backend."""
+    if HAS_TORCH is False:
```
mufeili, line 57:
```suggestion
    if not HAS_TORCH:
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+                        movie_genres[i, genre_map[ele]] = 1.0
+                    else:
+                        print(
+                            "genres not found, filled with unknown: {}".format(
+                                genres
+                            )
+                        )
+                        movie_genres[i, genre_map["unknown"]] = 1.0
+            for idx, genre_name in enumerate(self.genres):
+                assert idx == genre_map[genre_name]
```
mufeili, line 592: No need for L592 now

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+        download(self.url, path=zip_file_path)
+
+        with zipfile.ZipFile(zip_file_path, "r") as archive:
+            archive.extractall(path=self.raw_dir)
+
+        '''
+        TODO: Here are codes downloading movie features from dgl s3. 
+        Right now codes are used for testing only.
+        '''
+        dir = os.path.join(os.path.expanduser("~"), ".dgl", "movie_feat")
```
mufeili, line 239:
```suggestion
        dir = os.path.join(self.raw_dir, "movie_feat")
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+        shutil.copyfile(os.path.join(dir, self.name + '_movie_feat.pkl'), 
+                        os.path.join(self.raw_path, self.name + '_movie_feat.pkl'))
+
+    def process(self):
+        print("Starting processing {} ...".format(self.name))
```
mufeili, line 244:
```suggestion
        print(f"Starting processing {self.name} ...")
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+            names=["user_id", "movie_id", "rating", "timestamp"],
+            dtype={
+                "user_id": np.int32,
+                "movie_id": np.int32,
+                "ratings": np.float32,
+                "timestamp": np.int64,
+            },
+            engine="python",
+        )
+        rating_data = rating_data.sample(frac=1).reset_index(drop=True)
```
mufeili, line 614: `sample(frac=1)` is also random.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+            col_name="id",
+            reserved_ids_set=set(all_rating_data["user_id"].values),
+        )
+        movie_data = self._drop_unseen_nodes(
+            data_df=movie_data,
+            col_name="id",
+            reserved_ids_set=set(all_rating_data["movie_id"].values),
+        )
+
+        self.user_feat = tensor(self._process_user_fea(user_data))
```
mufeili, line 305: suggest renaming "fea" to "feat"

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+        # reindex u and v, v nodes start after u
+        num_user = len(self._global_user_id_map)
+        train_v_indices += num_user
+        valid_v_indices += num_user
+        test_v_indices += num_user
+
+        self.train_rating_pairs = (
+            tensor(train_u_indices, dtype=data_type_dict["int64"]),
+            tensor(train_v_indices, dtype=data_type_dict["int64"]),
+        )
```
mufeili, line 340: You can simply use `torch.LongTensor` since you are using PyTorch anyway. Same for all other usage of framework-agnostic APIs.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+    def has_cache(self):
+        if os.path.exists(self.graph_path) and os.path.exists(self.user_feat_path) and os.path.exists(self.movie_feat_path):
+            return True
+        return False
+
+    def save(self):
+        save_graphs(
+            self.graph_path,
+            [self.train_graph, self.valid_graph, self.test_graph],
+        )
```
mufeili, line 389: I think other stuff like `self.info` can also be saved with `save_graphs`

python/dgl/data/movielens.py
LspongebobJH: Yes. After we upload these three zip files to s3, we could replace these links with s3 urls.

python/dgl/data/movielens.py
LspongebobJH: fix

python/dgl/data/movielens.py
LspongebobJH: Ok. fix

python/dgl/data/movielens.py
LspongebobJH: Oh yes. fix.

python/dgl/data/movielens.py
LspongebobJH: Yes. I add more specifications in the doc and some tests in codes.

python/dgl/data/movielens.py
LspongebobJH: fix

python/dgl/data/movielens.py
LspongebobJH: user_feat and movie_feat have different feature dimensions. 
ml-100k, user_feat: 23, movie_feat: 320
ml-1m, user_feat: 23, movie_feat: 319
ml-10m, user_feat: 1, movie_feat: 321

python/dgl/data/movielens.py
LspongebobJH: Actually slow. Especially slow for ml-1m and 10m. If necessary I can provide an approximate time profile.

python/dgl/data/movielens.py
LspongebobJH: fix

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,644 @@
...
+    def __init__(
+        self,
+        name,
+        valid_ratio,
+        test_ratio=None,
+        raw_dir=None,
+        force_reload=None,
+        verbose=None,
+        transform=None,
+        random_state=None
```
mufeili, line 188: 0 above and "None" here.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,644 @@
...
+        Ratio of validation samples out of the whole dataset. Should be in (0.0, 1.0). 
+    test_ratio: int, optional
+        Ratio of testing samples out of the whole dataset. Should be in (0.0, 1.0) for :obj:`"ml-1m"` and :obj:`"ml-10m"`. This parameter is invalid
+        when :obj:`name` is :obj:`"ml-100k"`, since its testing samples are pre-specified. 
+        Default: None
+    raw_dir : str, optional
+        Raw file directory to download/store the data.
+        Default: ~/.dgl/
+    force_reload : bool, optional
+        Whether to re-download(if the dataset has not been downloaded) and re-process the dataset.
```
mufeili, line 104: "(if the dataset has not been downloaded)" this is not needed since the dataset will be downloaded in that case anyway

this is more about if the dataset on s3 has been updated

python/dgl/data/movielens.py
mufeili: In that case, should we return a heterogeneous graph instead?

python/dgl/data/movielens.py
mufeili: @LspongebobJH Any thoughts?

python/dgl/data/movielens.py
mufeili: Could you give me a rough number? You don't need to be too precise.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,644 @@
...
+    movie_feat: torch.Tensor
+        Movie features
+    info: dict
+        Store training, validation and testing rating pairs.
+
+    Notes
+    -----
+    - Each time when the dataset is loaded with a different setting of training, validation and test ratio, the parameter
+    :obj:`force_reload` should be set to :obj:`True`. Otherwise the previous split of the dataset will be loaded and the new split will not come into effect.
+    updated in the storage.
```
mufeili, line 128: Regarding

```
    - Each time when the dataset is loaded with a different setting of training, validation and test ratio, the parameter
    :obj:`force_reload` should be set to :obj:`True`. Otherwise the previous split of the dataset will be loaded and the new split will not come into effect.
    updated in the storage.
```

Can we handle this automatically as described [here](https://github.com/dmlc/dgl/issues/4293)? See the "Proposal" section.

python/dgl/data/movielens.py
mufeili: @LspongebobJH To clarify, you can still have empty lines, but you don't need ">>>" at the beginning.

python/dgl/data/movielens.py
mufeili: @LspongebobJH not handled

python/dgl/data/movielens.py
mufeili: @LspongebobJH ditto

python/dgl/data/movielens.py
LspongebobJH: Sorry but I don't get why the dataset will be re-downloaded with force_reload if the raw datasets have already been downloaded in the local storage? I have checked codes in `dgl_dataset.py` and see there are logics checking if there are existing datasets in local before downloading them with force_reload.

LspongebobJH: And could you elaborate "_this is more about if the dataset on s3 has been updated_"? I did't get this point.

python/dgl/data/movielens.py
LspongebobJH: Oh here are some conversations that I haven't had time to check yet. I leave them unresolved to remind me of checking them these days. I will mark _resolve conversation_ after I resolve them. The same for other unresolved conversations.

mufeili: I see. Thanks.

python/dgl/data/movielens.py
mufeili: > Sorry but I don't get why the dataset will be re-downloaded with force_reload if the raw datasets have already been downloaded in the local storage? I have checked codes in dgl_dataset.py and see there are logics checking if there are existing datasets in local before downloading them with force_reload.

This is not right. See https://github.com/dmlc/dgl/blob/master/python/dgl/data/dgl_dataset.py#L186. `force_reload` will force re-downloading.

mufeili: > And could you elaborate "this is more about if the dataset on s3 has been updated"? I did't get this point.

It's possible that the url stays unchanged, but the data source corresponding to the url has been changed and updated. In that case, it's necessary to force re-downloading. This is rare though.

LspongebobJH: > > Sorry but I don't get why the dataset will be re-downloaded with force_reload if the raw datasets have already been downloaded in the local storage? I have checked codes in dgl_dataset.py and see there are logics checking if there are existing datasets in local before downloading them with force_reload.
> 
> This is not right. See https://github.com/dmlc/dgl/blob/master/python/dgl/data/dgl_dataset.py#L186. `force_reload` will force re-downloading.

But look at the implementations of `self._download()`, https://github.com/dmlc/dgl/blob/4c01563245c46aadcd35d0869a8fb5d7962408d2/python/dgl/data/dgl_dataset.py#L159, dgll will test whether dataset has been downloaded first. If yes, then return.

mufeili: My bad. You are right. Then it always requires a new URL as raw_path has a suffix based on the URL.

python/dgl/data/movielens.py
LspongebobJH: Emm I'm also thinking about this. Unlike classic heterogeneous graph, movielens has no complicated canonical edge types or meta-paths that might carry rich information, and the numbers of node and edge types are too small(2 and 1) to exhibit diversity. I don't think users would regard it as a heterogeneous graph. Instead, they would think it's a homogeneous graph with edge classification as the target task since the only problem that brings us to heterogeneous graph is feature dimension, which I think is a minor problem(Two MLPs to change dimensions are enough). I didn't see the necessity to make movielens a more complicated structure. Furthermore, keeping it as homogeneous graph is helpful to directly apply GNN examples in DGL to movielens (maybe with two MLPs). Heterogeneous graph would make `HeteroGraphConv` API unavoidable(I guess, may not  be true). I think users are more familiar with the former option. Hence I think keeping movielens as simple homogeneous graphs is better

python/dgl/data/movielens.py
LspongebobJH: Oh really? I was supposed that save_graphs() can only be used to save dgl graphs and other objects have to be saved by save_info() :)

python/dgl/data/movielens.py
mufeili: DGL's heterogeneous graph data structure is meant to cover the bipartite graphs that you mentioned. Users can choose to convert it into a homogeneous graph if they want. You can leave the issue to them. Also, don't our current examples have the issue of different feature dimensions? How do they handle this issue?

python/dgl/data/movielens.py
mufeili: I think so, but you can have a try to verify that.

LspongebobJH: Oh I guess we can store them in `labels` param of save_graphs()?

LspongebobJH: > I think other stuff like `self.info` can also be saved with `save_graphs`.

I've tested and it won't work. save_graphs() requires input to be a list of DGLGraphs. We can't input self.info into `labels` param since it requires to be a dict with Tensor as values.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,640 @@
...
+            verbose=verbose,
+            transform=transform,
+        )
+
+    def download(self):
+        zip_file_path = os.path.join(self.raw_dir, self.name + ".zip")
+        download(self.url, path=zip_file_path)
+
+        with zipfile.ZipFile(zip_file_path, "r") as archive:
+            archive.extractall(path=self.raw_dir)
```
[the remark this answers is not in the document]
LspongebobJH: Just curious. I see that the default value of overwrite of extract_archive is False, but its document says the default value is True?
https://github.com/dmlc/dgl/blob/d64f7c4808ae249e9df5aec2bcc439f4b7dbb713/python/dgl/data/utils.py#L240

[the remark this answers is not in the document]
LspongebobJH: > I think you can import and use `extract_archive`?

fix in the latest commit.

python/dgl/data/movielens.py
LspongebobJH: fix

python/dgl/data/movielens.py
LspongebobJH: ml-100k: 1s
ml-1m: 10s
ml-10m: 192s(3min+)
Emm I think for ml-100k and ml-1m such period is acceptable. But for ml-10m such period may be too long?

python/dgl/data/movielens.py
LspongebobJH: One way is to store valid_ratio and test_ratio in a separate file when saving graphs. This file would be very small and can be quickly loaded. Each time when movielens API gonna load graphs, it loads this small file at first, check whether loaded ratio match with current ratio specified by users in API arguments. If yes, directly load graphs in local disk. If not, go to process(). What do you think?

python/dgl/data/movielens.py
mufeili: You are right. I agree with you.

python/dgl/data/movielens.py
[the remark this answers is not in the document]
mufeili: > Just curious. I see that the default value of overwrite of extract_archive is False, but its document says the default value is True?

It's possible that the doc is wrong or outdated.

python/dgl/data/movielens.py
mufeili: That sounds good to me.

LspongebobJH: fix in the latest commit. I've implemented `check_version()` to check whether the current valid_ratio and test_ratio are identical to the last setting. If not, dataset would be re-processed. The ratio setting would be stored in a local file specified by `self.version_path`. I also override `__repr__()` to correctly display information of dataset, such as current ratio and local storage path.

python/dgl/data/movielens.py
LspongebobJH: fix in the latest commit

python/dgl/data/movielens.py
LspongebobJH: Yeah you are right. Give me some time to refactor some codes.

mufeili: Sure. Thanks.

python/dgl/data/movielens.py
mufeili: Thanks. I' ll try it once we update the s3 file. Let's keep this unresolved for now.

python/dgl/data/movielens.py
LspongebobJH: fix in the latest commit. Convert graph structure to heterogeneous graphs and remove those items that are unnecessary to be stored in local.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,625 @@
...
+    >>> train_g, valid_g, test_g = dataset[0]
+    >>> train_g
+    Graph(num_nodes={'movie': 1617, 'user': 943},
+          num_edges={('movie', 'movie-user', 'user'): 64000, ('user', 'user-movie', 'movie'): 64000},
+          metagraph=[('movie', 'user', 'movie-user'), ('user', 'movie', 'user-movie')])
+
+    >>> # get ratings of edges in the training graph.
+    >>> # Note that we only store ratings in 'user-movie' edges since 'user-movie' and 'movie-user' are
+    >>> # just inverse edges and share the same set of ratings.
+    >>> ratings = train_g.edata['rate']['user-movie']
```
mufeili, line 136:
```suggestion
    >>> ratings = train_g.edges["user-movie"].data["rate"]
```
The advantage of this syntax is that you can also update features by `train_g.edges["user-movie"].data["rate"] = new_ratings` but not with `train_g.edata['rate']['user-movie'] = new_ratings`.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,625 @@
...
+    >>> ratings
+    tensor([3., 3., 2.,  ..., 4., 4., 2.])
+
+    >>> # get input features of users and movies respectively
+    >>> train.ndata['feat']['user'] # user features
```
mufeili, line 141: ditto

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,625 @@
...
+            self.construct_g(valid_rating_pairs, valid_rating_values, user_feat, movie_feat), \
+            self.construct_g(test_rating_pairs, test_rating_values, user_feat, movie_feat)
+
+        self.train_graph.edata.pop('_ID')
+        self.valid_graph.edata.pop('_ID')
+        self.test_graph.edata.pop('_ID')
+
+        self.train_graph.ndata.pop('_ID')
+        self.valid_graph.ndata.pop('_ID')
+        self.test_graph.ndata.pop('_ID')
```
mufeili, line 351: L345-351 can be done inside `construct_g`.

mufeili:
@LspongebobJH Have you tried your latest code with an end-to-end example on MovieLens to verify its correctness?

LspongebobJH:
> @LspongebobJH Have you tried your latest code with an end-to-end example on MovieLens to verify its correctness?

Not yet. I'll let you know after I finish testing.

LspongebobJH:
@mufeili The MovieLens API has been tested. I put testing codes in a script file `test_movielens.py`. Furthermore, the example codes in the doc of MovieLens have also been tested.

mufeili:
@LspongebobJH Have you tested on the model examples that use MovieLens?

LspongebobJH:
> @LspongebobJH Have you tested on the model examples that use MovieLens?

@mufeili Oh not yet. I think I can try on IGMC model? We just switched from IGMC PR to the work of MovieLens before. I can try to replace implementations of MovieLens in IGMC PR with the API that we implement in the current PR and see how the model perform.

mufeili:
@LspongebobJH Thank you for the great efforts! Could you send me the data in private? I'll test the API on my side and then upload the data to the s3 bucket if everything works well.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,584 @@
...
+import os
+import zipfile
+import shutil
+
+import numpy as np
+import pandas as pd
+
+
+from torch import LongTensor, Tensor
+from ..convert import graph, heterograph
```
mufeili, line 11:
```suggestion
from ..convert import heterograph
```

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,574 @@
...
+    - When :obj:`name` is :obj:`"ml-100k"`, the :obj:`test_ratio` is invalid, and the training ratio is equal to 1-:obj:`valid_ratio`.
+    When :obj:`name` is :obj:`"ml-1m"` or :obj:`"ml-10m"`, the :obj:`test_ratio` is valid,
+    and the training ratio is equal to 1-:obj:`valid_ratio`-:obj:`test_ratio`.
+    - The number of edges is doubled to form an undirected(bidirected) graph structure.
+
+    Examples
+    --------
+    >>> from dgl.data import MovieLensDataset
+    >>> dataset = MovieLensDataset(name='ml-100k', valid_ratio=0.2)
+    >>> train_g, valid_g, test_g = dataset[0]
```
mufeili, line 127: `dataset[0]` now only returns a single graph rather than three graphs.

mufeili: Thanks.

python/dgl/data/movielens.py
```diff
@@ -0,0 +1,593 @@
...
+    >>> g = dataset[0]
+    >>> g
+    Graph(num_nodes={'movie': 1682, 'user': 943},
+          num_edges={('movie', 'movie-user', 'user'): 100000, ('user', 'user-movie', 'movie'): 100000},
+          metagraph=[('movie', 'user', 'movie-user'), ('user', 'movie', 'user-movie')])
+
+    >>> # get ratings of edges in the training graph.
+    >>> rate = g.edges['user-movie'].data['rate'] # or rate = g.edges['movie-user'].data['rate']
+    >>> rate
+    tensor([5., 5., 3.,  ..., 3., 5., 3.])
```
mufeili, line 136: I got "tensor([5., 5., 3.,  ..., 3., 3., 5.])"

mufeili:
@LspongebobJH  I've uploaded the data files to DGL's s3 bucket. Could you try it and update the code if it works? You may want to first rename the old files to avoid overwriting.

mufeili:
Unit tests failed due to

```
train_graph, valid_graph, test_graph = movielens[0]
```

Please also check the other part of the unit test.

mufeili:
@dgl-bot

mufeili (approved):

emduc:
Hi ! Sorry to dig this out. 

I am trying to make sense of [this example](https://github.com/dmlc/dgl/blob/master/examples/pytorch/pinsage/process_movielens1m.py) and I would love to know how the features are preprocessed in the current MovieLensDataset, in this slightly frustrating not to be able to access movie titles and not having any idea of the item feats are computed. Could the PINSage example work with the given MovieLensDataset? Am I better off using the example preprocessing and ignore this implementation?

PR #5572, opened 2023-04-19
[LOG] check whether cpu_affinity is avaible on the platform

Zhanghyi:
## Description
cpu_affinity is not avalible on drawin. Refer to https://psutil.readthedocs.io/en/latest/index.html#psutil.Process.cpu_affinity. This fix may save 5 mininutes for macos users like me.
image
image

### DIFF
--- a/python/dgl/dataloading/dataloader.py
+++ b/python/dgl/dataloading/dataloader.py
@@ -1034,7 +1034,11 @@ class DataLoader(torch.utils.data.DataLoader):
         )
 
     def __iter__(self):
-        if self.device.type == "cpu" and not self.cpu_affinity_enabled:
+        if (
+            self.device.type == "cpu"
+            and hasattr(psutil.Process, "cpu_affinity")
+            and not self.cpu_affinity_enabled
+        ):
             link = "https://docs.dgl.ai/tutorials/cpu/cpu_best_practises.html"
             dgl_warning(
                 f"Dataloader CPU affinity opt is not enabled, consider switching it on "


mufeili:
What will happen now? Could you elaborate more about "save 5 mininutes"?

Zhanghyi:
> What will happen now? Could you elaborate more about "save 5 mininutes"?

Sure. As a macos user, when i use dgl dataloader, there is a warning: Dataloader CPU affinity opt is not enabled, consider switching it on (see enable_cpu_affinity() or CPU best practices for DGL https://docs.dgl.ai/tutorials/cpu/cpu_best_practises.html). And I adjust the code according to the doc, there will be an error telling me psutil.Process doesn't have method 'cpu_affinity'. I looked up the error and finally found out that cpu_affinity is not available on macos(drawin). This change avoid printing this warning to macos users.

Zhanghyi:
> > What will happen now? Could you elaborate more about "save 5 mininutes"?
> 
> Sure. As a macos user, when i use dgl dataloader, there is a warning: Dataloader CPU affinity opt is not enabled, consider switching it on (see enable_cpu_affinity() or CPU best practices for DGL https://docs.dgl.ai/tutorials/cpu/cpu_best_practises.html). And I adjust the code according to the doc, there will be an error telling me psutil.Process doesn't have method 'cpu_affinity'. I looked up the error and finally found out that cpu_affinity is not available on macos(drawin). This change avoid printing this warning to macos users.

The status quo is that macos users can never eliminate the warning(kinda annoying) and will probably spend some time trying to enable the cpu_affinity on macos. BTW, in cpu inference scenario on linux, I appreciate it a lot that using cpu_affinity did speedup 1.5x in my case.

mufeili (approved):
Thanks. I'm good.

mufeili:
@dgl-bot

PR #5578, opened 2023-04-20
[Distributed] Ensure round-robin edge file downloads, reduce logging, other improvements.

thvasilo:
## Description

This PR changes the edge structure file download order so that files are downloaded by each rank in a round-robin way, to be consistent with the expectations of ParMETIS and allow us to run ParMETIS in environments without a shared filesystem (e.g. SageMaker without EFS).

We also move lots of logging from the INFO level to DEBUG to reduce console clutter when running jobs and tests.

Finally we fix the tests for `tests/tools/test_parmetis_preproc.py`.

## Checklist
- [x] Code is well-documented


thvasilo:
Note that there some unrelated tests that do not work due to some missing constants, but the test suite should run successfully, e.g.

```
  File "tools/verify_partitions.py", line 246, in <module>
    _validate_results(params)
  File "tools/verify_partitions.py", line 187, in _validate_results
    node_partids = get_node_partids(params.partitions_dir, graph_schema)
  File "/home/ubuntu/repos/dgl/tools/verification_utils.py", line 207, in get_node_partids
    graph_schema[constants.STR_NODE_TYPE_COUNTS],
AttributeError: module 'constants' has no attribute 'STR_NODE_TYPE_COUNTS'
```

Those tests should be fixed in a separate PR.

Requesting review from @classicsong

tools/distpartitioning/parmetis_preprocess.py
```diff
@@ -76,50 +97,50 @@ def gen_edge_files(rank, schema_map, params):
...
+
+        global_src_id = data_f0 + ntype_gnid_offset[src_ntype_name][0, 0]
+        global_dst_id = data_f1 + ntype_gnid_offset[dst_ntype_name][0, 0]
+        cols = [global_src_id, global_dst_id]
+        col_names = ["global_src_id", "global_dst_id"]
+
+        out_file_name = Path(edge_data_files[idx]).stem.split(".")[0]
+        out_file = os.path.join(
+            outdir, etype_name, f"edges_{out_file_name}.csv"
+        )
```
thvasilo, line 116: @kylasa note that we change the directory where edge files are stored, including the edge type in the directory structure. Previously if two edge types shared the same file names they'd end up overwriting each other's files.

So where previously for 
```
{
	"edge_type_1": 
		{"data": ["file_1.parquet", "file_2.parquet"]},
	"edge_type_2": 
		{"data": ["file_1.parquet", "file_2.parquet"]},
}
```
we'd end up with just two files under `f"{outdir}/edges_file_1.parquet"` and `f"{outdir}/edges_file_2.parquet"`,
now we get the correct: `f"{outdir}/edge_type_1/edges_file_1.parquet"`, `f"{outdir}/edge_type_1/edges_file_2.parquet"` etc.

Rhett-Ying:
@dgl-bot

tools/distpartitioning/utils.py
```diff
@@ -701,11 +705,47 @@ def generate_read_list(num_files, world_size):
...
 
     Examples
     --------
     >>> tools.distpartitionning.utils.generate_read_list(10, 4)
     [array([0, 1, 2]), array([3, 4, 5]), array([6, 7]), array([8, 9])]
     """
     return np.array_split(np.arange(num_files), world_size)
+
+
+def generate_roundrobin_read_list(num_files, world_size):
```
Rhett-Ying, line 722: how about combining these 2 different generation method into 1 function with an argument like `strategy=`?

tools/distpartitioning/utils.py
```diff
@@ -701,11 +705,47 @@ def generate_read_list(num_files, world_size):
...
+        e.g. rank 0 reads the indexed in read_list[0], rank 1 the
+        ones in read_list[1] etc.
+
+    Examples
+    --------
+    >>> tools.distpartitionning.utils.generate_roundrobin_read_list(10, 4)
+    [[0, 4, 8], [1, 5, 9], [2, 6], [3, 7]]
+    """
+    assignment_lists = [[] for _ in range(world_size)]
+    for rank, part_idx in zip(cycle(range(world_size)), range(num_files)):
```
Rhett-Ying, line 748: just FYI which avoid importing `itertools`.
```
for i, num in enumerate(range(num_files)):
    subarrays[i % world_size].append(num)
```

tools/distpartitioning/utils.py
```diff
@@ -689,7 +690,10 @@ def map_partid_rank(partid, world_size):
 
 
 def generate_read_list(num_files, world_size):
-    """Generate the file IDs to read for each rank.
+    """
```
Rhett-Ying, line 693: how about re-naming this function to indicate what kind of split strategy is used?

Rhett-Ying (approved):
generally look good to me. please make sure CI including lint check pass. If something blocks CI, pls file another dedicated PR to resolve it.

tools/distpartitioning/utils.py
thvasilo: Can we have a decision here? I think I also prefer the two methods.

tools/distpartitioning/utils.py
thvasilo: Noted, but I still find the current implementation more readable. If there's no objection I'd like to keep it as is.

tools/distpartitioning/utils.py
thvasilo: Resolving to keep conversation in one thread.

tools/distpartitioning/parmetis_preprocess.py
```diff
@@ -43,6 +44,26 @@ def get_proc_info():
         return 0
 
 
+def get_world_size():
```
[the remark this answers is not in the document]
thvasilo: I wouldn't mind, but then we should move `get_proc_info()` (lines 23-44) as well.

Actually I'm curious why we don't use `torch.distributed.get_world_size()`  and `torch.distributed.get_rank()` for these, they seem to be used in most of DGL distributed.

tools/distpartitioning/parmetis_preprocess.py
```diff
@@ -76,50 +97,50 @@ def gen_edge_files(rank, schema_map, params):
...
+        col_names = ["global_src_id", "global_dst_id"]
+
+        out_file_name = Path(edge_data_files[idx]).stem.split(".")[0]
+        out_file = os.path.join(
+            outdir, etype_name, f"edges_{out_file_name}.csv"
+        )
+        os.makedirs(os.path.dirname(out_file), exist_ok=True)
+
+        options = csv.WriteOptions(include_header=False, delimiter=" ")
+        options.delimiter = " "
```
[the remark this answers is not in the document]
thvasilo: Thanks good catch.

Rhett-Ying:
@dgl-bot

tools/distpartitioning/utils.py
Rhett-Ying: sure, let's keep 2 methods.

tools/distpartitioning/parmetis_preprocess.py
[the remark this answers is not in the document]
thvasilo: @classicsong can we get a decision here? The options are:

* Move get_world_size() and get_proc_info() to utils.py.
* Remove these functions and use `torch.distributed.get_world_size()` and `torch.distributed.get_rank()` in another PR (as they might need further testing).
* Do nothing, keep both these functions in their current location.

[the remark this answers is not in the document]
classicsong: move Move get_world_size() and get_proc_info() to utils.py.
As we launch parmetis_preprocess.py using mpirun. So we use mpirun env to get world size.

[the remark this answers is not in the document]
thvasilo: Sounds good. Let's do this in a separate PR to avoid causing trouble with the existing codebase. Once we have better code coverage we can look into refactoring.

classicsong (approved):
LGTM

tools/distpartitioning/parmetis_preprocess.py
```diff
@@ -360,23 +386,23 @@ def run_preprocess_data(params):
...
         An instance of argparser class which stores command line arguments.
     """
-    logging.info(f"Starting to generate ParMETIS files...")
+    logging.info("Starting to generate ParMETIS files...")
     rank = get_proc_info()
 
     assert os.path.isdir(
         params.input_dir
-    ), f"Please check `input_dir` argument."
+    ), f"Please check `input_dir` argument: {params.input_dit}."
```
thvasilo, line 394: @thvasilo Fix in upcoming PR.

Issue #5592, opened 2023-04-24
benchmark tests crashed after PR 5366(commit: 62b5f50)


Rhett-Ying:
## 🐛 Bug

`//benchmarks/benchmarks/model_speed/bench_rgcn_hetero_ns.py` works well on commit `27b008b` which is just before `62b5f50` but crahsed on `62b5f50`.
```
               load ogbn-mag
               finish loading ogbn-mag
               start training...
               4 DL workers are assigned to cpus [0, 1, 2, 3], main process will use cpus [4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
               Traceback (most recent call last):
                 File "/home/ubuntu/.local/lib/python3.8/site-packages/asv/benchmark.py", line 1293, in main_run_server
                   main_run(run_args)
                 File "/home/ubuntu/.local/lib/python3.8/site-packages/asv/benchmark.py", line 1167, in main_run
                   result = benchmark.do_run()
                 File "/home/ubuntu/.local/lib/python3.8/site-packages/asv/benchmark.py", line 573, in do_run
                   return self.run(*self._current_params)
                 File "/home/ubuntu/.local/lib/python3.8/site-packages/asv/benchmark.py", line 859, in run
                   return self.func(*param)
                 File "/home/ubuntu/workspace/dgl_1/benchmarks/benchmarks/model_speed/bench_rgcn_hetero_ns.py", line 365, in track_time
                   for step, (input_nodes, seeds, blocks) in enumerate(loader):
                 File "/home/ubuntu/workspace/dgl_1/python/dgl/dataloading/dataloader.py", line 619, in __next__
                   self._next_non_threaded()
                 File "/home/ubuntu/workspace/dgl_1/python/dgl/dataloading/dataloader.py", line 592, in _next_non_threaded
                   batch = next(self.dataloader_it)
                 File "/home/ubuntu/.local/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 628, in __next__
                   data = self._next_data()
                 File "/home/ubuntu/.local/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1316, in _next_data
                   idx, data = self._get_data()
                 File "/home/ubuntu/.local/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1282, in _get_data
                   success, data = self._try_get_data()
[...]

PR #5609, opened 2023-04-25
[Bugfix] Fix the uninitialized pin_memory flag for empty graph

chang-l:
## Description
This is to fix https://github.com/dmlc/dgl/issues/5592. For future/own reference, I will detail the root cause in the following.

The root cause is when the first relational graph is empty, the entire heterograph could return `is_pinned() = true` even if they are never being pinned by DGL or PyTorch at all. 

This is because (1) the empty unit graph would return `is_pinned = True` due to the following init. logic: https://github.com/dmlc/dgl/blob/ae8cbde5cfa699ff8d9e774ae6acd2cb8f7b451a/include/dgl/aten/coo.h#L67-L69 and (2) the first relational graph determines if the entire hetero-graph is pinned. 
https://github.com/dmlc/dgl/blob/ae8cbde5cfa699ff8d9e774ae6acd2cb8f7b451a/src/graph/heterograph.h#L58

As a result, when the sub-graph is pickled/unpickled by workers/main process after sampling, the code will always try to pin the (sub)graph as long as its `relation_graphs_[0]` is an empty graph (`coo.data`, `coo.row` and `coo.col` are NullArray). See here https://github.com/dmlc/dgl/blob/ae8cbde5cfa699ff8d9e774ae6acd2cb8f7b451a/src/graph/pickle.cc#L64 is for sub-graph pickling, which will always write `True` if `graph->relation_graphs_[0]` is empty. And then here https://github.com/dmlc/dgl/blob/ae8cbde5cfa699ff8d9e774ae6acd2cb8f7b451a/src/graph/pickle.cc#L156-L158 will pin the unpickled graph even if it never gets pinned before.

## Checklist
- [ ] Code is well-documented

## Changes
- If its a null/empty graph, always init. the `is_pinned` to false
- Update tests to cover this case

### DIFF
--- a/include/dgl/aten/coo.h
+++ b/include/dgl/aten/coo.h
@@ -158,7 +158,6 @@ struct COOMatrix {
              "with the above stacktrace.";
       return new_coo;
     }
-    is_pinned = true;
     return *this;
   }
 
--- a/include/dgl/aten/coo.h
+++ b/include/dgl/aten/coo.h
@@ -158,6 +158,7 @@ struct COOMatrix {
              "with the above stacktrace.";
       return new_coo;
     }
+    is_pinned = true;
     return *this;
   }
 
--- a/include/dgl/aten/coo.h
+++ b/include/dgl/aten/coo.h
@@ -64,8 +64,7 @@ struct COOMatrix {
         data(darr),
         row_sorted(rsorted),
         col_sorted(csorted) {
-    if (!aten::IsNullArray(row) || !aten::IsNullArray(col) ||
-        !aten::IsNullArray(data)) {
+    if (!IsEmpty()) {
       is_pinned = (aten::IsNullArray(row) || row.IsPinned()) &&
                   (aten::IsNullArray(col) || col.IsPinned()) &&
                   (aten::IsNullArray(data) || data.IsPinned());
@@ -130,6 +129,11 @@ struct COOMatrix {
     CHECK_NO_OVERFLOW(row->dtype, num_cols);
   }
 
+  inline bool IsEmpty() const {
+    return aten::IsNullArray(row) && aten::IsNullArray(col) &&
+           aten::IsNullArray(data);
+  }
+
   /** @brief Return a copy of this matrix on the give device context. */
   inline COOMatrix CopyTo(const DGLContext& ctx) const {
     if (ctx == row->ctx) return *this;
@@ -141,8 +145,7 @@ struct COOMatrix {
 
   /** @brief Return a copy of this matrix in pinned (page-locked) memory. */
   inline COOMatrix PinMemory() {
-    if (!aten::IsNullArray(row) || !aten::IsNullArray(col) ||
-        !aten::IsNullArray(data)) {
+    if (!IsEmpty()) {
       if (is_pinned) return *this;
       auto new_coo = COOMatrix(
           num_rows, num_cols, row.PinMemory(), col.PinMemory(),
@@ -168,8 +171,7 @@ struct COOMatrix {
    *       The context check is deferred to pinning the NDArray.
    */
   inline void PinMemory_() {
-    if (!aten::IsNullArray(row) || !aten::IsNullArray(col) ||
-        !aten::IsNullArray(data)) {
+    if (!IsEmpty()) {
       if (is_pinned) return;
       row.PinMemory_();
       col.PinMemory_();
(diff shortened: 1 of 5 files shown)


chang-l:
@Rhett-Ying  Can you please try this PR to see if it can fix your crash? Also, what PyTorch version did you use for testing?

Rhett-Ying:
@dgl-bot run r6i.16xlarge bench_rgcn_hetero dmlc/1.0.x

Rhett-Ying:
@chang-l   pytorch: 1.13.1

Rhett-Ying:
@chang-l confirmed. it works now.

include/dgl/aten/coo.h
```diff
@@ -64,9 +64,12 @@ struct COOMatrix {
         data(darr),
         row_sorted(rsorted),
         col_sorted(csorted) {
-    is_pinned = (aten::IsNullArray(row) || row.IsPinned()) &&
-                (aten::IsNullArray(col) || col.IsPinned()) &&
-                (aten::IsNullArray(data) || data.IsPinned());
+    if (!aten::IsNullArray(row) || !aten::IsNullArray(col) ||
```
Rhett-Ying, line 67: how about wrapping these `IsNullArray()` check into a member function of `COOMatrix`: `IsEmpty()`to remove code redundancy?

chang-l:
> pytorch: 1.13.1

Thanks. In fact, I was not able to reproduce the crash with PyTorch > 1.12 (built/run with cuda) yesterday... I guess you ran it with CPU-only? (of course it is a bug even if it ran through with cuda)
Anyway, glad it fixes the issue. I will finalize this PR later today :)

Rhett-Ying:
@chang-l yes, I hit this issue when run on CPU.

Rhett-Ying:
@dgl-bot run r6i.16xlarge bench_rgcn_hetero dmlc/1.0.x

include/dgl/aten/coo.h
```diff
@@ -138,16 +145,21 @@ struct COOMatrix {
...
+          aten::IsNullArray(data) ? data : data.PinMemory(), row_sorted,
+          col_sorted);
+      CHECK(new_coo.is_pinned)
+          << "An internal DGL error has occured while trying to pin a COO "
+             "matrix. Please file a bug at "
+             "'https://github.com/dmlc/dgl/issues' "
+             "with the above stacktrace.";
+      return new_coo;
+    }
+    is_pinned = true;
```
yaox12, line 161: `is_pinned` will be set to true even if `IsEmpty() == true`, is it expected?

chang-l: You're right. `is_pinned` should not be set for this pinning method :( 
Thank you!

include/dgl/aten/coo.h
```diff
@@ -159,13 +170,17 @@ struct COOMatrix {
...
+    if (!IsEmpty()) {
+      if (is_pinned) return;
+      row.PinMemory_();
+      col.PinMemory_();
+      if (!aten::IsNullArray(data)) {
+        data.PinMemory_();
+      }
+      is_pinned = true;
     }
     is_pinned = true;
```
yaox12, line 182: Also here.

include/dgl/aten/csr.h
```diff
@@ -151,13 +162,17 @@ struct CSRMatrix {
...
+    if (!IsEmpty()) {
+      if (is_pinned) return;
+      indptr.PinMemory_();
+      indices.PinMemory_();
+      if (!aten::IsNullArray(data)) {
+        data.PinMemory_();
+      }
+      is_pinned = true;
     }
     is_pinned = true;
```
yaox12, line 174: And here.

chang-l: This is probably necessary I think, as it is an in-place method. We should flag it even if it is empty when we called this function.

include/dgl/aten/coo.h
chang-l: Given the test failures, now, I start thinking I probably should set `is_pinned ` to True even if `IsEmpty() == true`, for both in-place and out-of-place methods. If graph is empty, we should just change the flag in these functions. Then, out-of-place method becomes the same as in-place method as there are no memory movements, but just flag changing.

Rhett-Ying (approved):

Rhett-Ying:
@chang-l  are we ready to merge this PR?

PR #5617, opened 2023-04-26
[Feature] Decompose dataset to each process to add multi-node support for dataloader

chang-l:
## Description
This is a PR to address https://github.com/dmlc/dgl/issues/5522 with [alternative IV ](https://github.com/dmlc/dgl/pull/5523#issuecomment-1512273586), i.e., decompose (1D) the dataset to each process to reduce memory overhead. Note that there are two different decompositions depending on if `drop_last=True`.

This PR is still WIP as I need to add some tests... 

## Checklist
- [ ] Code is well-documented
- [] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

## Changes
Added several functions to decompose the `id_tensor` and `indices` to each process. Now, `id_tensor` stores the original dataset at each partition (process), while `indices` stores the local indexes at each proc.

### DIFF
--- a/python/dgl/dataloading/dataloader.py
+++ b/python/dgl/dataloading/dataloader.py
@@ -24,7 +24,6 @@ from ..batch import batch as batch_graphs
 from ..distributed import DistGraph
 from ..frame import LazyFeature
 from ..heterograph import DGLGraph
-from ..multiprocessing import call_once_and_share
 from ..storages import wrap_storage
 from ..utils import (
     dtype_of,
@@ -132,6 +131,57 @@ def _get_id_tensor_from_mapping(indices, device, keys):
         offset += length
     return id_tensor
 
+def _split_to_local_id_tensor_from_mapping(indices, keys, bounds):
+    dtype = dtype_of(indices)
+    device = next(iter(indices.values())).device
+    num_samples = bounds[1] - bounds[0]
+    id_tensor = torch.empty(num_samples, 2, dtype=dtype, device=device)
+
+    index_offset = 0
+    split_id_offset = 0
+    for i, k in enumerate(keys):
+        if k not in indices:
+            continue
+        index = indices[k]
+        length = index.shape[0]
+        index_offset2 = index_offset + length #[offset : offset2)
+        lb = max(bounds[0], index_offset)
+        ub = min(bounds[1], index_offset2)
+        if ub > lb:
+            split_id_offset2 = split_id_offset + (ub - lb)
+            assert split_id_offset2 <= num_samples
+            id_tensor[split_id_offset:split_id_offset2, 0] = i
+            id_tensor[split_id_offset:split_id_offset2, 1] = index[lb-index_offset:ub-index_offset]
+            split_id_offset += (ub - lb)
+            if (split_id_offset2 == num_samples):
+                break
+        index_offset = index_offset2
+    # padding locally when drop_last = False
+    if split_id_offset2 < num_samples:
+        assert (bounds[1] - index_offset2) == (num_samples - split_id_offset2)
+        padding_size = bounds[1] - index_offset2
+        id_tensor[split_id_offset2:, 0] = id_tensor[:padding_size, 0]
+        id_tensor[split_id_offset2:, 1] = id_tensor[:padding_size, 1]
+    # make sure id_tensor is completely filled
+    assert(split_id_offset2 == num_samples or split_id_offset2 + padding_size == num_samples)
+    return id_tensor
+
+def _split_to_local_id_tensor(indices, bounds):
+    dtype = dtype_of(indices)
+    device = indices.device
+    num_samples = bounds[1] - bounds[0]
+    id_tensor = torch.empty(num_samples, dtype=dtype, device=device)
+
+    # if needs padding when drop_last = False
+    if bounds[1] > len(indices):
+        reminder = len(indices) - bounds[0]
+        id_tensor[0:reminder] = indices[bounds[0]:]
+        padding_size = bounds[1] - len(indices)
+        # padding id_tensor from start
+        id_tensor[reminder:] = indices[:padding_size]
+    else:
+        id_tensor = indices[bounds[0]:bounds[1]]
+    return id_tensor
 
 def _divide_by_worker(dataset, batch_size, drop_last):
     num_samples = dataset.shape[0]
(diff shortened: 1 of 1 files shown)


python/dgl/dataloading/dataloader.py
```diff
@@ -133,6 +132,61 @@ def _get_id_tensor_from_mapping(indices, device, keys):
...
+
+def _split_to_local_id_tensor(indices, bounds):
+    dtype = dtype_of(indices)
+    device = indices.device
+    num_samples = bounds[1] - bounds[0]
+    id_tensor = torch.empty(num_samples, dtype=dtype, device=device)
+
+    # if needs padding when drop_last = False
+    if bounds[1] > len(indices):
+        reminder = len(indices) - bounds[0]
```
nvcastet, line 180: nit: `remainder`

python/dgl/dataloading/dataloader.py
```diff
@@ -133,6 +132,61 @@ def _get_id_tensor_from_mapping(indices, device, keys):
...
+        id_tensor[0:reminder] = indices[bounds[0] :]
+        padding_size = bounds[1] - len(indices)
+        # padding id_tensor from start
```
nvcastet, line 183: Do we need to pad `self._id_tensor`?
I thought we did not need to since at line 321, we make sure to pad `self._indices` with indices less than `self.local_num_indices`.

nvcastet (commented):
Thank you for the great PR.
Just one question about padding.

python/dgl/dataloading/dataloader.py
chang-l: Right. `self._id_tensor` may not need to be padded. I guess I was trying to keep the same size of `id_tensor` at every proc and did the padding for safety... I will simplify this part.

BarclayII:
One question: how would the users call DataLoader while avoiding storing the entire ID set on a single machine?

chang-l:
>One question: how would the users call DataLoader while avoiding storing the entire ID set on a single machine?

@BarclayII To clairfy, this PR assumes the entire index datasets (e.g., train_idx) are stored/duplicated in each machine when calling the dataloader (people can use shared memory to store the datasets to save memory). So, with this PR, the native DGL (without DistDGL) should work out-of-the-box with `torchrun`/`torch.distributed.launch`, assuming each machine has a copy of graph structure/features/index datasets.

In future, if users have to store id datasets distributedly across nodes, unlike distributed sampling or distributed feature fetching (given index datasets are already distributed, it is reasonable to assume graph/feature must also be stored distributedly), the modification based on this PR should be much more straightforward. We just need to ensure the input `indices` is decomposed to every local process instead of all processes, i.e., just changing the function call `_decompose_one_dimension(length, world_size, rank, drop_last):` --> `_decompose_one_dimension(length, local_world_size, local_rank, drop_last):`.

python/dgl/dataloading/dataloader.py
nv-dlasalle: My understanding is we need padding here for the case where the dataset is not shuffled, and `321` for when the dataset is shuffled (to ensure the replicated indices change with each shuffle).

nv-dlasalle (approved):
Overall looks good. Minor comments that optional to address.

python/dgl/dataloading/dataloader.py
nvcastet: Ah the padding here is more for the non-shuffling case? But we could have the same behavior as un the shuffling case where we limit the index range inside the padded `self._indices` to be less than `self.local_num_indices` so that the padded section of `_id_tensor` is never used.

python/dgl/dataloading/dataloader.py
```diff
@@ -229,64 +293,39 @@ def __init__(self, indices, batch_size, drop_last, ddp_seed, shuffle):
...
-        # with the same size as the indices so we can trim it later.
-        self.shared_mem_size = (
-            self.total_size if not self.drop_last else len_indices
+        self.local_bounds = _decompose_one_dimension(
+            len_indices, self.num_replicas, self.rank, drop_last
         )
-        self.num_indices = len_indices
+        self.num_samples = self.local_bounds[1] - self.local_bounds[0]
+        self.local_num_indices = self.num_samples
+        if not drop_last and self.local_bounds[1] > len_indices:
```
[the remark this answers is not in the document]
chang-l: Thanks. I updated.

python/dgl/dataloading/dataloader.py
chang-l: Yes, we need to either pad `self._indices` or `self._id_tensor` for shuffling. In general, it doesn't matter as it would not affect memory and computation efficiency...

In the latest commit, I padded `self._indices` and leave `self._id_tensor` uninitialized, to make it consistent between `shuffle=True` or `False`

jermainewang:
@BarclayII @peizhou001 Do you have other comments?

python/dgl/dataloading/dataloader.py
```diff
@@ -133,6 +132,50 @@ def _get_id_tensor_from_mapping(indices, device, keys):
     return id_tensor
 
 
+def _split_to_local_id_tensor_from_mapping(indices, keys, bounds):
```
BarclayII, line 135: Could we make `bounds` two arguments instead?  Like `lower_bound` and `upper_bound`?  I think they are two scalars?  `bounds[1]` and `bounds[0]` are less readable.

python/dgl/dataloading/dataloader.py
```diff
@@ -133,6 +132,50 @@ def _get_id_tensor_from_mapping(indices, device, keys):
...
+                lower - index_offset : upper - index_offset
+            ]
+            split_id_offset += upper - lower
+            if split_id_offset2 == num_samples:
+                break
+        index_offset = index_offset2
+    return id_tensor
+
+
+def _split_to_local_id_tensor(indices, bounds):
```
BarclayII, line 165: Same here.

python/dgl/dataloading/dataloader.py
```diff
@@ -229,64 +282,48 @@ def __init__(self, indices, batch_size, drop_last, ddp_seed, shuffle):
...
-        else:
-            self.num_samples = math.ceil(len_indices / self.num_replicas)
-        self.total_size = self.num_samples * self.num_replicas
-        # If drop_last is True, we create a shared memory array larger than the number
-        # of indices since we will need to pad it after shuffling to make it evenly
-        # divisible before every epoch.  If drop_last is False, we create an array
-        # with the same size as the indices so we can trim it later.
-        self.shared_mem_size = (
-            self.total_size if not self.drop_last else len_indices
+        self.local_bounds = _decompose_one_dimension(
```
BarclayII, line 285: In this case you can do
```
self.local_lower_bound, self.local_upper_bound = _decompose_one_dimension(
...
)
```

BarclayII (approved):
I'm OK with a code style suggestion.

python/dgl/dataloading/dataloader.py
chang-l: Sure. Thanks

PR #5644, opened 2023-05-01
[Tutorial] Add a small test for GAT exercise

BarclayII:


notebooks/sparse/quickstart.ipynb
```diff
@@ -1307,6 +1307,21 @@
       },
       "execution_count": null,
       "outputs": []
+    },
+    {
+      "cell_type": "code",
+      "source": [
+        "# Test:\n",
+        "# Let's use the symmetric A created above.\n",
+        "X = torch.randn(6, 20)\n",
```
frozenbugs, line 1316: Is it possible to not use randn, such that the result is deterministic?

BarclayII: Since we are coding a NN layer whose parameters are randomly initialized, the result won't be deterministic anyway.

Issue #4501, opened 2022-09-01
Support CI stage for CUDA 11.x


isratnisa:
## 🚀 Feature
Current CI is based on only CUDA 10.2. We need to support later versions such as 11.0 to resolve and test issues (e.g., [issue 3444](https://github.com/dmlc/dgl/issues/3444))

## Motivation
Issue [3444](https://github.com/dmlc/dgl/issues/3444)

## Alternatives
Disable CSRGEMM APIs in CUDA 11.0 and used the obsolete CSRGEMM from CUDA 10.2 APIs.

BarclayII:
PyTorch 1.13 now deprecated CUDA 10.2 support so the priority needs to be adjusted a little.

mfbalin:
Are there any plans on deprecating CUDA 10.2 support for DGL? DGL dependencies such as thrust and CUB dropped support for it in their latest releases. For example, #5648 requires updating thrust and CUB.

Rhett-Ying:
Hi @mfbalin  what's the minimum CUDA version required by #5648 ?

mfbalin:
Hi @Rhett-Ying, that PR seems to successfully compile on the nvcr.io/nvidia/pytorch:20.12-py3 container which has CUDA 11.1.1. I am not sure if that is the lowest version that will compile though.

Rhett-Ying:
Thanks for your info. We'll discuss the deprecation and get back to you as soon as possible.

mfbalin:
@Rhett-Ying any news on CUDA 10.2 support/deprecation?

Rhett-Ying:
@mfbalin We have added this into our special topic and will discuss it this Thursday.

mfbalin:
Lowest supported cuda version is 11 now. Reopen if unresolved.

PR #5648, opened 2023-05-02
[Performance][CUDA] Faster CSRToCOO

mfbalin:
## Description
CUB recently implemented a DeviceCopy::Batched algorithm, which can be used to implement the DeviceRunLength::Decode algorithm, which in turn can be used to implement the CSRToCOO for the GPU in a more efficient way for the 64-bit code path. The current implementation suffers from O(E log N) complexity, which this PR reduces to O(E).

My motivation was that CSRToCOO is called inside the labor sampler algorithm, so improving the performance will also improve Labor Sampler and make its complexity truly linear in the number of edges.

My experimental evaluations show that the Labor Sampler runtime consistently improves with this change.

## Checklist
- [x] Code is well-documented


mfbalin:
@BarclayII @yaox12, could you take a look? This PR requires using the latest thrust and cub versions.

yaox12:
1. Has this feature been included in any CUB release version?
2. CUB 2.x introduced the libcu++ dependency. Do you know if it's compatible with CUDA 10.2, since DGL still supports it?

yaox12:
@dgl-bot

mfbalin:
@yaox12
1. Not yet, I just recently contributed DeviceCopy to CUB in this [PR](https://github.com/NVIDIA/cub/pull/675), I guess it will be included in the next release. I don't know the timeline of the next release though.
2. It looks like thrust includes cub and libcu++ as submodule dependencies however I don't know if the libcu++ version that comes with thrust is compatible with CUDA 10.2. What I was able to find is that libcu++ was first released for CUDA 10.2. I don't know if they dropped CUDA 10.2 support in later releases.

yaox12:
@dgl-bot

mfbalin:
It looks like the libcudacxx headers can not be found even though I added it to the cuda include directories, what else needs to be done so that those headers can be found?

Error msg: `/root/jenkins/workspace/dgl_PR-5648@2/third_party/thrust/thrust/detail/type_traits.h:27:10: fatal error: cuda/std/type_traits: No such file or directory`

yaox12:
@dgl-bot

mfbalin:
It looks like the following header can not be found this time: `fatal error: cuda_bf16.h: No such file or directory`

@yaox12 Isn't this header supposed to come with CUDA? Do you have any insights?

jermainewang:
@yaox12  Would you like to review the code?

czkkkkkk:
From the above discussion, this PR may depends on the CUDA version. @mfbalin can you keep the current implementation and use the new code when CUDA version is greater than 11.0?

> Not yet, I just recently contributed DeviceCopy to CUB in this https://github.com/NVIDIA/cub/pull/675, I guess it will be included in the next release. I don't know the timeline of the next release though.

Does it mean this PR depends on a unreleased version of CUB? If so, we cannot merge this PR for now.

mfbalin:
The new CUB and Thrust releases require CUDA 11.0 from my understanding. That is why I was waiting for a reply in #4501 to see if CUDA 10.2 support would be maintained or not. Otherwise, additional effort is required to choose the thrust version according to the CUDA version and maintain multiple versions of the code. I created this PR to show that future thrust/CUB versions have benefits for the DGL code base so that maintainers can decide how to proceed. Even if this PR is not merged now, it could be merged in the future.

jermainewang:
@frozenbugs @Rhett-Ying for the CUDA 10.2 issue in CI.

czkkkkkk:
Hi @mfbalin, we do not deprecate CUDA 10.2 for now. Let's keep this PR open and merge it after the deprecation.

mfbalin:
@czkkkkkk CUDA 10.2 is not in the CI anymore if I am not mistaken with #5977. Are we going to wait for months before the next release of thrust/cub or is it possible to merge this PR before then?

yaox12:
@dgl-bot

src/array/cuda/csr2coo.cu
```diff
@@ -80,14 +78,33 @@ COOMatrix CSRToCOO<kDGLCUDA, int64_t>(CSRMatrix csr) {
...
+  thrust::counting_iterator<int64_t> iota(0);
+
+  auto input_buffer = thrust::make_transform_iterator(iota, RepeatIndex{});
+  auto output_buffer = thrust::make_transform_iterator(
+      iota, OutputBufferIndexer<int64_t>{
+                csr.indptr.Ptr<int64_t>(), ret_row.Ptr<int64_t>()});
+  auto buffer_sizes = thrust::make_transform_iterator(
+      iota, AdjacentDifference<int64_t>{csr.indptr.Ptr<int64_t>()});
+
+  constexpr int64_t max_copy_at_once = std::numeric_limits<int32_t>::max();
```
czkkkkkk, line 93: `max_copy_at_once` is a pretty large number. Why it is necessary?

CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ if(USE_CUDA)
   message(STATUS "Use external CUB/Thrust library for a consistent API and performance.")
   cuda_include_directories(BEFORE "${CMAKE_SOURCE_DIR}/third_party/thrust")
   cuda_include_directories(BEFORE "${CMAKE_SOURCE_DIR}/third_party/thrust/dependencies/cub")
+  cuda_include_directories(BEFORE "${CMAKE_SOURCE_DIR}/third_party/thrust/dependencies/libcudacxx/include")
```
czkkkkkk, line 53: Here introduces more dependencies on thrust. Could you try whether you can use the thrust library in CUDA?

mfbalin: They come automatically when the thrust submodule in DGL is updated. New versions of thrust come with its own cub and libcudacxx.

src/array/cuda/csr2coo.cu
mfbalin: Because the cub algorithm only works with int32_t types when it comes to sizes. If we want to copy more items, then we call the algorithm multiple times.

mfbalin: For example, if we had a graph 3 billion nodes, then the loop would run multiple times.

CMakeLists.txt
mfbalin: The CUB function I am using in this PR became part of CUB very recently. Thus, it will be available only with cuda 12. Since dgl supports cuda 11, DGL needs to continue to depend on thrust as a submodule.

mfbalin:
I am not sure if this new code path or the existing int32_t specialization is faster. One would need to do a benchmark to see if the cusparse csr2coo  is better or cub is better.
Cusparse code path: https://github.com/dmlc/dgl/blob/078be4369411e72365248409e39a544ed4451e86/src/array/cuda/csr2coo.cu#L44

CMakeLists.txt
czkkkkkk: I see.

czkkkkkk (approved):

CMakeLists.txt
frozenbugs: Can you put a TODO here to remind us remove this once we deprecating CUDA 11?

frozenbugs:
@mfbalin can you help us take a look at the failing test?

mfbalin:
Partition had a minor bug due to the 1 off specification of histogramEven call parameters. Tests should all pass now.

mfbalin:
Actually, we might want to fix this bug in a separate PR so that if there is any problem with the new thrust version (on the off change), we can take this PR back but the bug fix is still stays.

frozenbugs:
> Actually, we might want to fix this bug in a separate PR so that if there is any problem with the new thrust version (on the off change), we can take this PR back but the bug fix is still stays.

Solid judgement :)

frozenbugs:
> Partition had a minor bug due to the 1 off specification of histogramEven call parameters. Tests should all pass now.

Thanks for fixing, any reason this was not exposed without your optimization?

mfbalin:
The histogram has bins, let's say we are computing a histogram for bins with values 0 to 6 (7 bins in total). Without the fix, the bin boundaries were [0, 8) (8 exclusive) vs [0, 7). Documentation says that CUB evenly distributed the values in the range to bins. But 8 doesn't divide 7. So without the fix, with the new update, 0 and 1 was in the first bin. Without the update 6 and 7 were in the last bin (Which didn't expose the bug). I am speculating that this caused a rounding error and a change in the implementation in CUB might have exposed this bug.

PR #5657, opened 2023-05-06
[GraphBolt] add ItemSet

Rhett-Ying:

### DIFF
--- a/python/dgl/dataloading2/itemset.py
+++ b/python/dgl/dataloading2/itemset.py
@@ -36,7 +36,7 @@ class ItemSet:
         raise NotImplementedError
 
 
-class DictItemSet(ItemSet):
+class DictItemSet:
     r"""Itemset wrapping multiple itemsets with keys.
 
     Each item is retrieved by iterating over each itemset and returned with
@@ -54,3 +54,9 @@ class DictItemSet(ItemSet):
         for key, itemset in self._itemsets.items():
             for item in itemset:
                 yield {key: item}
+
+    def __getitem__(self, _):
+        raise NotImplementedError
+
+    def __len__(self):
+        raise NotImplementedError
--- a/python/dgl/dataloading2/itemset.py
+++ b/python/dgl/dataloading2/itemset.py
@@ -1,0 +1,55 @@
+__all__ = [
+    "ItemSet",
+    "DictItemSet"
+]
+
+
+class ItemSet:
+    r"""An iterable itemset.
+
+    All itemsets that represent an iterable of items should subclass it. Such
+    form of itemset is particularly useful when items come from a stream. This
+    class requires each input itemset to be iterable.
+
+    Parameters
+    ----------
+    items: Iterable or Tuple[Iterable]
+    """
+    def __init__(self, items):
+        if isinstance(items, tuple):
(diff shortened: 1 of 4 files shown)


python/dgl/dataloading2/itemset.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Parameters
+    ----------
+    itemsets: Dict[str, ItemSet]
+    """
+
+    def __init__(self, itemsets):
+        self._itemsets = itemsets
+
+    def __iter__(self):
```
frozenbugs, line 53: By default, it will return items of one key first and then next key?

By any chance it is a desired behavior?

tests/python/pytorch/dataloading2/test_itemset.py
```diff
@@ -0,0 +1,115 @@
...
+        assert heads[i] == head
+        assert tails[i] == tail
+        assert_close(neg_tails[i], negs)
+
+
+def test_DictItemSet():
+    # Node or edge IDs
+    ids = {
+        ("user", "like", "item"): dgl.ItemSet(torch.arange(0, 5)),
+        ("user", "follow", "user"): dgl.ItemSet(torch.arange(5, 10)),
```
frozenbugs, line 49: It is not really related to the test, but usually test should reflect the actual use case.

In our design for hetoro graph, for the train set, should the id of different node all start from zero or continuous?

python/dgl/dataloading2/itemset.py
Rhett-Ying: this behavior is what we discussed before: just iterate over key by key.

tests/python/pytorch/dataloading2/test_itemset.py
Rhett-Ying: train ids are always required to start from zero? and what about val/test ids which will `ItemSet` as well?

frozenbugs: I see, correct. But making it continuous [(0,5) then (5, 10)] is a little bit confusing.

tests/python/pytorch/dataloading2/test_itemset.py
```diff
@@ -0,0 +1,115 @@
+import dgl
+import torch
+from torch.testing import assert_close
+
+
+def test_ItemSet():
+    # Node or edge IDs.
```
frozenbugs, line 7: In principle, we should break each test case such as:

test_ItemSet_Node_Ids():

test_ItemSet_Graph_Ids():

test_ItemSet_Node_Pairs():

e.t.c.

frozenbugs:
Overall LGTM, only nit on tests.

tests/python/pytorch/dataloading2/test_itemset.py
Rhett-Ying: sure, let's use `(0,5)`

tests/python/pytorch/dataloading2/test_itemset.py
Rhett-Ying: changed as suggested.

frozenbugs (approved):

PR #5685, opened 2023-05-13
[Performance Improvement] Make GPU sampling and to_block use pinned memory to decrease required synchronization

nv-dlasalle:
## Description
This uses the pinned memory from #5366 to improve GPU->CPU communicating tensor sizes during sampling and to_block. This improves the performance of UVA and GPU sampling.

## Checklist
- [x] Code is well-documented


src/graph/transform/cuda/cuda_to_block.cu
```diff
@@ -174,13 +177,16 @@ struct CUDAIdsMapper {
...
+          {num_ntypes}, DGLDataTypeTraits<int64_t>::dtype,
+          DGLContext{kDGLCPU, 0});
+      CUDA_CALL(cudaMemcpyAsync(
+          new_len_tensor->data, count_lhs_device,
+          sizeof(*num_nodes_per_type.data()) * num_ntypes,
+          cudaMemcpyDeviceToHost, stream));
+      CUDA_CALL(cudaEventRecord(copyEvent, stream));
 
-      // Wait for the node counts to finish transferring.
       device->FreeWorkspace(ctx, count_lhs_device);
```
yaox12, line 190: Is there a risk that `count_lhs_device` will be freed before the copying is finished?

nv-dlasalle: As the copy is happening on the current cuda stream retrieved by `runtime::getCurrentCUDAStream()`, this allocation should not be available for re-use until the copy operation finishes.

Issue #5545, opened 2023-04-12
DGLGraph.formats with list input doesn't work


mfbalin:
## To Reproduce

```
g = g.formats(['csc', 'coo'])
g.formats()
```

Should create both csc and coo formats. But the output is:
`{'created': ['coo'], 'not created': ['csc']}`

https://colab.research.google.com/drive/1cbLt_NaWNI3X5R3ekVmLPaIrrt4iJI_g?usp=sharing

## Expected behavior

Both formats to be created.

## Environment

 - DGL Version (e.g., 1.0): 1.0
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): Pytorch
 - OS (e.g., Linux):
 - How you installed DGL (`conda`, `pip`, source): pip
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version (if applicable):
 - GPU models and configuration (e.g. V100):
 - Any other relevant information:

mufeili:
@frozenbugs @jermainewang @BarclayII 

Currently `DGLGraph.formats(formats)` always only creates a single sparse format even if multiple formats are passed. The format gets chosen based on some logic in C++. The rest formats appear only to be "allowed formats" and appear in `'not created'` when you invoke `DGLGraph.formats()`. I've tried DGL 0.6, and this was already the case with DGL 0.6. Do you still remember the context and expected behavior of `DGLGraph.formats(formats)`?

jermainewang:
According to the [API doc](https://docs.dgl.ai/generated/dgl.DGLGraph.formats.html#dgl.DGLGraph.formats), "Get a cloned graph with the specified sparse format(s)", it should return a graph with all the formats specified by the users materialized. In this regard, the current implementation is buggy.

mufeili:
@jermainewang Then do we want to fix the implementation or simply update the doc?

mfbalin:
I think the behaviour of the code needs to be corrected, otherwise what would be the point of passing a list of formats to be created?

I had a nasty memory bug in a multi-gpu code, each process recreated formats after mp.spawn and I fixed it by using DGLGraph.create_formats_() because DGLGraph.formats is not working as intended. But this is not optimal because a user might not need all the formats and may want to save memory.

jermainewang:
+1 w/ @mfbalin

mufeili:
@mfbalin @jermainewang got it, added to backlog

keli-wen:
[...]

PR #5708, opened 2023-05-18
[Sparse] Solved the problem that DGLGraph.formats with list input doesn't work 

keli-wen:
## Description

I have resolved the previous [issue#5545 -- DGLGraph.formats with list input doesn't work](https://github.com/dmlc/dgl/issues/5545). I identified that the issue was in the `UnitGraph::GetGraphInFormat` function within `src/graph/unit_graph.cc`. Now, the `DGLGraph.formats` function can correctly create the corresponding sparse format.

## Checklist
- [ ] Code is well-documented


mufeili:
@dgl-bot

mufeili:
@dgl-bot

frozenbugs:
@dgl-bot

mufeili:
Could you rename the unit test [here](https://github.com/dmlc/dgl/blob/master/tests/python/common/test_heterograph.py#L2527) to `test_formats` and expand it to cover the cases previously DGL failed to handle?

mufeili:
@Rhett-Ying Do you think if we need to trigger a regression test for this change?

frozenbugs:
> @Rhett-Ying Do you think if we need to trigger a regression test for this change?

Why we have concern about regression after we update unit test? Does this PR affect performance?

frozenbugs:
@dgl-bot

frozenbugs:
The error likely related to this line.

https://github.com/dmlc/dgl/blob/53714ca83112f0747061b53b4f1d3e0b5fd80611/tests/distributed/test_distributed_sampling.py#L33

Since the fix, the formats time will increase, which caused the timeout.

keli-wen:
> 



> The error likely related to this line.
> 
> https://github.com/dmlc/dgl/blob/53714ca83112f0747061b53b4f1d3e0b5fd80611/tests/distributed/test_distributed_sampling.py#L33
> 
> Since the fix, the formats time will increase, which caused the timeout.

Actually, when I checked the log, I found that the error **might** come from:

`dgl._ffi.base.DGLError: [15:59:30] /root/jenkins/workspace/dgl_PR-5708/src/array/cpu/./rowwise_pick.h:340: Check failed: (j + 1 == len) || (et[et_idx[j]] <= et[et_idx[j + 1]]): Edge type is not sorted. Please sort in advance or specify 'rowwise_etype_sorted' as false.`

Because I noticed that after this error occurred, it started to time out.

Before the repair of the `formats` function, there was actually only one Edge type, so could it be possible that this CHECK has not been functioning before the repair of the `formats` function?

PR #5700, opened 2023-05-16
[GraphBolt] Add csc sampling graph python side code

peizhou001:
## Description
This is the python code about csc sampling graph.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [x] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [x] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

PR #5711, opened 2023-05-19
[GraphBolt] Reorganzie graphbolt module

peizhou001:
## Description
This PR rename the folder name of graphbolt and remove the currenltly unsused files inside it.

## Checklist
- [ ] Code is well-documented


python/dgl/graphbolt/__init__.py
```diff
@@ -1,3 +1,4 @@
 """GraphBolt"""
 
+from .graph_storage import *
```
Rhett-Ying, line 3: does `graph_storage` exist?

peizhou001: For later PR usage

Rhett-Ying: then it's better to add this at that time.

Rhett-Ying (approved):
overall LGTM. How about replace `base.py` with `__init__.py` to mark it as a module?

python/dgl/graphbolt/__init__.py
peizhou001: Both would be ok, according to the comment in #5700 , we move the `graphbolt` level changes in this PR.

python/dgl/__init__.py
```diff
@@ -48,7 +47,6 @@
     set_node_lazy_features,
     set_src_lazy_features,
 )
-from .dataloading2 import *
```
Rhett-Ying, line 51: this results in CI failure, I think.

python/dgl/graphbolt/__init__.py
frozenbugs: Only add code when the code is valid, if the graph_storage does not even exist, why import?

frozenbugs (approved):

python/dgl/__init__.py
Rhett-Ying: I think it's better to update `dgl.ItemSet` in tests to `graphbolt.ItemSet` in this PR.

peizhou001: fixed

python/dgl/graphbolt/__init__.py
peizhou001: Make sense, have dropped the changes.

Issue #5710, opened 2023-05-19
`dgl.linkx_homophily` calculation fails for `value=0`.


galogm:
## 🐛 Bug

`dgl.linkx_homophily` calculation fails for `value=0`.

## To Reproduce

```py
import dgl

dataset = dgl.data.TexasDataset('./data')
graph = dataset[0]

# remove self-loop and make the graph undirected
graph = dgl.remove_self_loop(graph)
graph = dgl.to_bidirected(graph, copy_ndata=True)

linkx_h = dgl.linkx_homophily(graph, graph.ndata['label'])
```

## Expected behavior

```log
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "${path_to_venv}/.venv/lib/python3.7/site-packages/dgl/homophily.py", line 189, in linkx_homophily
    return value.item() / (num_classes - 1)
AttributeError: 'int' object has no attribute 'item'
```

## Environment

 - DGL Version (e.g., 1.0): 1.1.0
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): 1.12.1+cu113
 - OS (e.g., Linux): Linux
 - How you installed DGL (`conda`, `pip`, source): pip
 - Build command you used (if compiling from source): none
 - Python version: 3.7.16
 - CUDA/cuDNN version (if applicable): 11.3
 - GPU models and configuration (e.g. V100): none
 - Any other relevant information: none

## Additional context

The initial `value` of linkx_homophily needs to be `tensor(0.0)` rather than `int(0)`.

![image](https://github.com/dmlc/dgl/assets/65325063/b4a60993-ee30-439c-8a82-60597b6910d9)

mufeili:
Thank you for the report. I confirm the bug. Could you help fix it by opening a PR?

PR #5720, opened 2023-05-22
[Bugfix] `dgl.linkx_homophily` calculation fails when all class homophily=0

galogm:
## Description

Fail to calculate `dgl.linkx_homophily` when all class homophily = 0 due to inpropriate type of initial value.

image

```log
  File "${path_to_venv}/.venv/lib/python3.7/site-packages/dgl/homophily.py", line 189, in linkx_homophily
    return value.item() / (num_classes - 1)
AttributeError: 'int' object has no attribute 'item'
```

Check issue #5710 for details.

## Checklist
- [x] Code is well-documented

## Changes

- [x] Fix1: Change the initial value from `int(0)` to `torch.tensor(0.0)`. As the `homophily` module has already tried to import torch, it feels safe to user `torch.tensor` as the initial value.
- [x] Test1: Add unit test for the fix.

### DIFF
--- a/python/dgl/homophily.py
+++ b/python/dgl/homophily.py
@@ -177,7 +177,7 @@ def linkx_homophily(graph, y):
         num_nodes = graph.num_nodes()
         num_classes = y.max(dim=0).values.item() + 1
 
-        value = 0
+        value = torch.tensor(0.0)
         for k in range(num_classes):
             # Get the nodes that belong to class k.
             class_mask = y == k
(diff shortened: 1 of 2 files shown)


python/dgl/homophily.py
```diff
@@ -177,7 +177,7 @@ def linkx_homophily(graph, y):
         num_nodes = graph.num_nodes()
         num_classes = y.max(dim=0).values.item() + 1
 
-        value = 0
+        value = torch.tensor(0.0)
```
mufeili, line 180:
```suggestion
        value = torch.tensor(0.0).to(graph.device)
```

mufeili:
I'm good other than the minor comment above as the graph can be on a GPU.

mufeili (approved):

mufeili:
@dgl-bot

Issue #5443, opened 2023-03-09
Support converting heterogeneous graph to NetworkX graph


jermainewang:
## 🚀 Feature

Extend [`dgl.to_networkx()`](https://docs.dgl.ai/generated/dgl.to_networkx.html?highlight=to_networkx#dgl.to_networkx) or add a new API `dgl.to_networkx_hetero` to support converting a DGL heterogeneous graph to a NetworkX graph.

## Motivation

Requested by user community: https://discuss.dgl.ai/t/k-hop-path-between-two-nodes-for-a-heterogeneous-graph/3448

## Alternatives

No alternative. Users need manual implementation.

### Heterogeneous graph in NetworkX
The key design decision is how to store a heterogeneous graph in NetworkX. Below are two proposals:

To create a user-item graph with two relations "follow" and "like":

**Idea 1** Set the "type" attributes of nodes and edges.

```python
>>> import networkx as nx
>>> G = nx.DiGraph()
>>> G.add_node(0, type="user")
>>> G.add_node(1, type="user")
>>> G.add_node(2, type="user")
>>> G.add_node(3, type="item")
>>> G.add_node(4, type="item")
>>> G.add_edge(0, 1, type="follow")
>>> G.add_edge(0, 2, type="follow")
>>> G.add_edge(2, 0, type="follow")
>>> G.add_edge(0, 3, type="like")
>>> G.add_edge(0, 4, type="like")
>>> G.add_edge(2, 4, type="like")
```

**Idea 2** Set the "triples" edge attribute ([rdflib's practice](https://rdflib.readthedocs.io/en/stable/_modules/rdflib/extras/external_graph_libs.html))

```python
>>> import networkx as nx
>>> G = nx.DiGraph()
>>> G.add_edge(0, 1, triples=("user", "follow", "user"))
>>> G.add_edge(0, 2, triples=("user", "follow", "user"))
>>> G.add_edge(2, 0, triples=("user", "follow", "user"))
>>> G.add_edge(0, 3, triples=("user", "like", "item"))
>>> G.add_edge(0, 4, triples=("user", "like", "item"))
>>> G.add_edge(2, 4, triples=("user", "like", "item"))
```

### API behavior

```python
import dgl
hg = ...  # DGL heterogeneous graph
hg_nx = dgl.to_networkx(hg)   # Converted to a networkx graph
```

### Handle node/edge attributes

```python
hg = ...  # DGL heterogeneous graph
[...]

PR #5726, opened 2023-05-22
[Feature] Support heterogeneous graphs in the `to_networkx` method

nvitucci:
## Description
I've seen there has been no further action on https://github.com/dmlc/dgl/issues/5443, so I went on to raise my own. I've leveraged most of the code from the `to_homogeneous` method as I quite liked the solution.

## Checklist

- [x] Code is well-documented

## Changes
- The `to_networkx` method is now split into two `_to_networkx_homogeneous` and `_to_networkx_heterogeneous` private methods. The docstring is updated accordingly.
- The `_to_networkx_heterogeneous` adds both node types as a node `label` attribute and edge canonical types as the `triples` property per the suggestion in the [original issue](https://github.com/dmlc/dgl/issues/5443). I am not fully convinced of the `triples` attribute though, but this can be further discussed either here or in the issue, depending on your preference.

## Notes

This PR does not yet include a complete test because I have noticed that the only related test code is the `_test_nx_conversion` function in `tests/python/common/function/test_basics.py`. I have anyway added some test code in the updated `to_networkx` method docstring, and I am ready to add a full test. What would be the preference here?


nvitucci:
@dgl-bot

mufeili:
@AymericBasset, for awareness

mufeili:
Could you add unit tests following the steps below?

1. Create a file `tests/python/common/test_convert.py`.
2. Add a function `test_to_networkx` in this file. You can move the old unit tests [here](https://github.com/dmlc/dgl/blob/master/tests/python/common/function/test_basics.py#L202) to the new file and further augment them.

nvitucci:
@mufeili I've moved the existing test in the new module and enabled it, but it fails even without my changes. The offending lines are the following:

```
nxg = nx.cycle_graph(5)
nxg.remove_nodes_from([0, 4])
for u in nxg.nodes():
    nxg.nodes[u]["h"] = F.tensor([u])
for u, v, d in nxg.edges(data=True):
    d["h"] = F.tensor([u, v])

g = dgl.from_networkx(nxg, node_attrs=["h"], edge_attrs=["h"])
```

For the time being I've disabled it again as it is not directly relevant to this PR. Is this a known issue that led to disable it in the first place?

mufeili:
> @mufeili I've moved the existing test in the new module and enabled it, but it fails even without my changes with `dgl._ffi.base.DGLError: Expect edge_id_attr_name and edge_attrs to be None when nx_graph is undirected, got None and ['h']`. The offending lines are the following:
> 
> ```
> nxg = nx.cycle_graph(5)
> nxg.remove_nodes_from([0, 4])
> for u in nxg.nodes():
>     nxg.nodes[u]["h"] = F.tensor([u])
> for u, v, d in nxg.edges(data=True):
>     d["h"] = F.tensor([u, v])
> 
> g = dgl.from_networkx(nxg, node_attrs=["h"], edge_attrs=["h"])
> ```
> 
> For the time being I've disabled it again as it is not directly relevant to this PR. Is this a known issue that led to disable it in the first place?

Thank you for the report. I think this issue is irrelevant to this PR as this PR focuses on `to_networkx`. Could you open a separate issue so that we can take a deeper look later?

tests/python/common/test_convert.py
```diff
@@ -0,0 +1,179 @@
+from collections import defaultdict as ddict
+
+import dgl
+import networkx as nx
+
+import backend as F
+
```
mufeili, line 7: Define a single `test_to_networkx` so that we have a one-on-one correspondence between the unit test function and the user facing API.

nvitucci: Sounds good. Should this contain only the test code from the new `test_nx_convert_heterogeneous` (maybe with a TODO mentioning the other existing test, which I'd move back to `test_basics` until the `DGLError` issue is solved)?

mufeili:
@nvitucci Lint check failed

```
python/dgl/convert.py:1809:0: C0301: Line too long (103/100) (line-too-long)
python/dgl/convert.py:1810:0: C0301: Line too long (101/100) (line-too-long)
python/dgl/convert.py:1813:0: C0301: Line too long (104/100) (line-too-long)
python/dgl/convert.py:1814:0: C0301: Line too long (103/100) (line-too-long)
```

python/dgl/convert.py
```diff
@@ -1688,36 +1775,52 @@ def to_networkx(g, node_attrs=None, edge_attrs=None):
...
+        ... })
+        ... g.ndata['n'] = {
+        ...     'game': torch.zeros(5, 1),
+        ...     'user': torch.ones(4, 1)
+        ... }
+        ... g.edata['e'] = {
+        ...     ('user', 'follows', 'user'): torch.zeros(2, 1),
+        ...     'plays': torch.ones(2, 1)
+        ... }
+    >>> nx_g = to_networkx(g, node_attrs=['n'], edge_attrs=['e'])
```
mufeili, line 1794:
```suggestion
    >>> nx_g = dgl.to_networkx(g, node_attrs=['n'], edge_attrs=['e'])
```

python/dgl/convert.py
```diff
@@ -1688,36 +1775,52 @@ def to_networkx(g, node_attrs=None, edge_attrs=None):
...
+    >>> nx_g.nodes(data=True)
+    NodeDataView({0: {'label': 'game', 'n': tensor([0.])},
```
mufeili, line 1796: What if there are features called `'label'`? For example, `g.ndata['label'] = ...`. It might be better to expose an argument for users to specify the node feature name to use for node types.

nvitucci:
> @nvitucci Lint check failed
> 
> ```
> python/dgl/convert.py:1809:0: C0301: Line too long (103/100) (line-too-long)
> python/dgl/convert.py:1810:0: C0301: Line too long (101/100) (line-too-long)
> python/dgl/convert.py:1813:0: C0301: Line too long (104/100) (line-too-long)
> python/dgl/convert.py:1814:0: C0301: Line too long (103/100) (line-too-long)
> ```

You're right, I couldn't run the linter after the change - will fix later.

python/dgl/convert.py
nvitucci: True, it would be better to make it configurable. Perhaps it is still convenient to keep a default name, for example `type` (more common in triple-based approaches) rather than `label`?

mufeili:
@dgl-bot

nvitucci:
@dgl-bot

nvitucci:
@mufeili Do I always need to tag the dgl-bot manually?

mufeili:
> @mufeili Do I always need to tag the dgl-bot manually?

@nvitucci Sorry for the unpleasant experience. Currently only core developers are granted the permission to trigger it.

python/dgl/convert.py
mufeili: Yes. That sounds good to me.

python/dgl/convert.py
```diff
@@ -1688,36 +1793,54 @@ def to_networkx(g, node_attrs=None, edge_attrs=None):
     >>> nx_g.edges(data=True)
     OutMultiEdgeDataView([(1, 1, {'id': 0, 'h1': tensor([1.]), 'h2': tensor([0., 0.])}),
                           (2, 3, {'id': 1, 'h1': tensor([1.]), 'h2': tensor([0., 0.])})])
+
+    With a heterogeneous graph:
+
+    >>> g = dgl.heterograph({
+        ...     ('user', 'follows', 'user'): (torch.tensor([0, 1]), torch.tensor([1, 2])),
```
mufeili, line 1800: "..." needs to be aligned with ">>>". Same for all the usage below.

python/dgl/convert.py
```diff
@@ -1643,8 +1643,107 @@ def _batcher(lst):
...
+            )
+
+    return nx_graph
+
+
+def to_networkx(
+    g,
+    node_attrs=None,
+    edge_attrs=None,
+    ntype_attr="type",
```
mufeili, line 1743: I suggest changing to "ntype" to be consistent with the terminology used by DGL.

python/dgl/convert.py
```diff
@@ -1643,8 +1643,107 @@ def _batcher(lst):
...
+    etype_attr="triples",
```
mufeili, line 1744: Similarly, I suggest changing to "etype" here.

python/dgl/convert.py
```diff
@@ -1653,11 +1752,15 @@ def to_networkx(g, node_attrs=None, edge_attrs=None):
...
     ----------
     g : DGLGraph
-        A homogeneous graph.
+        A homogeneous or heterogeneous graph.
     node_attrs : iterable of str, optional
         The node attributes to copy from ``g.ndata``. (Default: None)
     edge_attrs : iterable of str, optional
         The edge attributes to copy from ``g.edata``. (Default: None)
+    ntype_attr : str, optional
+        The node attribute to write the node type to. (Default: "type")
```
mufeili, line 1761: I suggest changing to

```
The name of the node attribute to store the node types in the NetworkX object.
```

python/dgl/convert.py
```diff
@@ -1653,11 +1752,15 @@ def to_networkx(g, node_attrs=None, edge_attrs=None):
...
+    etype_attr : str, optional
+        The edge attribute to write the edge type to. (Default: "triples")
```
mufeili, line 1763: Ditto. Also, will you store the edge type or the canonical edge type?

nvitucci:
> > @mufeili Do I always need to tag the dgl-bot manually?
> 
> @nvitucci Sorry for the unpleasant experience. Currently only core developers are granted the permission to trigger it. I'll trigger that for you once I finish a review. I will greatly appreciate it if you can wait till I finish a round of review to save some back-and-forth efforts.

Thanks! No worries - I thought I had to do it myself, but since it's not the case I'll leave you to trigger it whenever it's most convenient.

python/dgl/convert.py
nvitucci: You're right, I'm not sure why that happened. Fixed.

python/dgl/convert.py
nvitucci: Sounds good. Done.

python/dgl/convert.py
nvitucci: Sounds good to me. Done.

python/dgl/convert.py
nvitucci: My understanding is that, according to the original conversation, the canonical type (as a triple) is preferred here. I also think this is more flexible, as the edge type name can always be extracted from the canonical type if that's needed.

To summarize, the approach here is to save the node type as the `ntype` attribute on NetworkX nodes and the (canonical) edge type as the `etype` attribute on NetworkX edges. Is this still a good assumption?

python/dgl/convert.py
```diff
@@ -1688,36 +1793,54 @@ def to_networkx(g, node_attrs=None, edge_attrs=None):
...
+                          (9, 6, {'id': 0, 'triples': ('user', 'follows', 'topic')}),
+                          (9, 7, {'id': 1, 'triples': ('user', 'follows', 'topic')}),
+                          (9, 10, {'id': 3, 'triples': ('user', 'follows', 'user'), 'e': tensor([0.])}),
+                          (11, 4, {'id': 5, 'triples': ('user', 'plays', 'game'), 'e': tensor([1.])})])
     """
     if g.device != F.cpu():
         raise DGLError(
             "Cannot convert a CUDA graph to networkx. Call g.cpu() first."
         )
     if not g.is_homogeneous:
```
mufeili, line 1838: It might the code more readable to do `if g.is_homogeneous:` instead.

python/dgl/convert.py
```diff
@@ -1643,8 +1643,107 @@ def _batcher(lst):
...
+        for _, _, attr in nx_graph.edges(data=True):
+            eid = attr["id"]
+            feat_dict = g._get_e_repr(0, eid)
+            attr.update(
+                {key: F.squeeze(feat_dict[key], 0) for key in edge_attrs}
+            )
+    return nx_graph
+
+
+def _to_networkx_heterogeneous(
```
mufeili, line 1672: Will it significantly simplify the implementation by leveraging https://docs.dgl.ai/generated/dgl.to_homogeneous.html#dgl.to_homogeneous? Intuitively, there might be a significant overlapping between the logic here and the logic in `to_homogeneous`.

python/dgl/convert.py
nvitucci: I agree. Fixed.

mufeili:
@dgl-bot

python/dgl/convert.py
nvitucci: There is certainly some overlap between the two. I think the main difference is that `to_homogeneous` creates a new graph, which is not necessary while creating a NetworkX graph IMHO (although my first approach was to use _exactly_ `to_homogeneous` and adjust the indices in order to create the NetworkX graph). Also, as far as I can tell, `to_homogeneous` is not being explicitly tested.

Would you like me to look for ways to extract common code?

mufeili: Then, will it suffice to simply call `to_homogeneous`, make some adjustments, and then call the `to_networkx` implementation for homogeneous graphs? This might significantly simplify your implementation. Over the years, an important lesson I've learnt is that we should try avoiding adding new code unless it's really necessary. In particular, there can be nasty maintenance efforts.

nvitucci: It would do, at the expense of creating an intermediate graph object. Do you consider this an acceptable compromise?

nvitucci: Ok then, I'll update the code to make use of `to_homogeneous` as much as possible. 

I think it would make sense to add some note in the docstring saying that the conversion creates a homogeneous graph internally and, as such, incurs extra memory cost. What do you think?

mufeili: I don't think you need to add that kind of note. Many of DGL's modules/utilities can incur extra memory cost. Creating a NetworkX object may also incur extra memory cost.

nvitucci: Fine by me. Can you also confirm/reject the [assumption](https://github.com/dmlc/dgl/pull/5726#discussion_r1206340811) on etypes?

nvitucci: Interestingly, `to_homogeneous` does not seem to work right away with `ndata` and `edata`. I'll investigate.

nvitucci: From further investigation, I've found that `to_homogeneous` does not work in two cases, namely:
- when a node/edge feature is not assigned to all node/edge types;
- when its shape is different across node/edge types. 

A minimal example:

```
g = dgl.heterograph(
    {
        ("user", "follows", "user"): (
            torch.tensor([0, 1, 1, 2, 2, 2]),
            torch.tensor([0, 0, 1, 1, 2, 2]),
        ),
        ("user", "plays", "game"): (
            torch.tensor([0, 1]),
            torch.tensor([0, 1]),
        ),
    }
)
g.nodes["user"].data["h"] = torch.ones(3, 1)
g.edges["plays"].data["h"] = torch.ones(2, 1)
```

Both `dgl.to_homogeneous(g, ndata=["h"])` and `dgl.to_homogeneous(g, edata=["h"])` fail with `KeyError: 'h'` (the issue is in the `combine_frames` function of the `heterograph` module).

Adding feature `h` on both the `game` node type and the `follows` edge type with a different shape (second dimension of the `torch.ones` vector being 2 instead of 1) fails as well in the same method. For example, the following:

```
g.nodes["game"].data["h"] = torch.ones(2, 2)
dgl.to_homogeneous(g, ndata=["h"])
```

fails with `DGLError: Cannot concatenate column h with shape Scheme(shape=(1,), dtype=torch.float32) and shape Scheme(shape=(2,), dtype=torch.float32)`, and the same happens with the following:

```
g.edges["follows"].data["h"] = torch.ones(6, 2)
dgl.to_homogeneous(g, edata=["h"])
```

Therefore, `to_homogeneous` cannot be used as a direct replacement if this is the desired behaviour.

Am I correct in assuming that both these scenarios are valid use cases?

python/dgl/convert.py
mufeili: Yes. You just need to make that point clear in the documentation.

python/dgl/convert.py
mufeili: I'm fine either way. Will you encounter any of these cases in your scenario?

Option one, we can choose not to support these two cases for this PR and wait to see if there will be many feature requests. We might want to improve the warning messages to communicate better.

Option two, we support these two cases. For this route, we can use `to_homogeneous` to renumber the node/edge IDs and process the node/edge feature separately.

nvitucci: I have no direct use case at the moment, but I think they are reasonable scenarios as typically different node/edge types have different attributes assigned. 

Unless you specifically want to rule these out because of different assumptions (perhaps related to uniformity?), then option two seems more natural.

nvitucci: I've refactored the internal function to use `to_homogeneous` and process ndata/edata separately, with a comment to explain the choice.

mufeili:
Why did you make changes related to `libxsmm`? If this was done by accident, could you revert the changes related to it?

python/dgl/convert.py
```diff
@@ -1643,8 +1643,98 @@ def _batcher(lst):
...
+    etypes = g.canonical_etypes
+
+    for hom_nid, ndata in enumerate(zip(hom_g.ndata[NID], hom_g.ndata[NTYPE])):
+        orig_nid, ntype = ndata
+        attrs = {ntype_attr: ntypes[ntype]}
+
+        if node_attrs is not None:
+            assert (
+                ntype_attr not in node_attrs
+            ), f"'{ntype_attr}' already used as node type attribute"
```
mufeili, line 1693:
```suggestion
            ), f"'{ntype_attr}' already used as node type attribute, please provide a different ntype_attr"
```

nvitucci:
> Why did you make changes related to `libxsmm`? If this was done by accident, could you revert the changes related to it?

It must have happened while updating from master. I've reverted it.

python/dgl/convert.py
```diff
@@ -1643,8 +1643,98 @@ def _batcher(lst):
...
+                    for key in node_attrs
+                    if key in feat_dict
+                }
+            )
+
+        nx_graph.add_node(hom_nid, **attrs)
+
+    for hom_eid, edata in enumerate(zip(hom_g.edata[EID], hom_g.edata[ETYPE])):
+        orig_eid, etype = edata
+        attrs = {"id": hom_eid, etype_attr: etypes[etype]}
```
mufeili, line 1708: Why did you include "hom_eid" in "id", but not for the nodes? Also, "id" again becomes a reserved key that users should not use. Is this necessary?

python/dgl/convert.py
```diff
@@ -1643,8 +1643,98 @@ def _batcher(lst):
...
+
+        if edge_attrs is not None:
+            assert (
+                etype_attr not in edge_attrs
+            ), f"'{etype_attr}' already used as edge type attribute"
```
mufeili: ditto

tests/python/common/test_convert.py
```diff
@@ -0,0 +1,100 @@
+import backend as F
+import dgl
+
+
+def get_nodes_by_ntype(nodes, ntype):
+    return dict((k, v) for k, v in nodes.items() if v["ntype"] == ntype)
+
+
+def get_edges_by_etype(edges, etype):
+    return [e for e in edges if e[2]["etype"] == etype]
```
mufeili, line 10: Leave a comment explaining `e = (.., ..., ...)` so that we can better understand the magic number 2 here.

python/dgl/convert.py
nvitucci: I've done this for compatibility with what the [current method](https://docs.dgl.ai/en/latest/generated/dgl.to_networkx.html) does for homogeneous graphs, but actually I meant to ask this too. What do you think?

mufeili: I see. Thanks. Maybe for now we just add an extra argument to handle this and maintain backward compatibility.

tests/python/common/test_convert.py
nvitucci: I've added a method since this was used in several places.

python/dgl/convert.py
mufeili: I meant adding an extra argument "eid_attr" and set its default value to "id".

python/dgl/convert.py
```diff
@@ -1830,7 +1833,7 @@ def to_networkx(
         return _to_networkx_homogeneous(g, node_attrs, edge_attrs)
```
mufeili, line 1833: Also do this for homogeneous graphs?

python/dgl/convert.py
```diff
@@ -1643,8 +1643,99 @@ def _batcher(lst):
...
+        nx_graph.add_node(hom_nid, **attrs)
+
+    for hom_eid, edata in enumerate(zip(hom_g.edata[EID], hom_g.edata[ETYPE])):
+        orig_eid, etype = edata
+        attrs = {eid_attr: hom_eid, etype_attr: etypes[etype]}
+
+        if edge_attrs is not None:
+            assert (
+                etype_attr not in edge_attrs
+            ), f"'{etype_attr}' already used as edge type attribute, please provide a different value for etype_attr"
```
mufeili, line 1713: Also check `eid_attr`

python/dgl/convert.py
mufeili: Maybe no :) I'm not very convinced why do we need to do this for now. Let's just do this for edges for backward compatibility.

nvitucci: I think we only have it because of the way the attributes are added in the homogeneous version. I agree we keep it for compatibility only.

python/dgl/convert.py
nvitucci: Sure, I can add this, but the test for the homogeneous (current) version is still disabled. I'd recommend to leave this out to leave the existing code as unchanged as possible, so that it's easy to make changes later if needed.

python/dgl/convert.py
```diff
@@ -1643,8 +1643,102 @@ def _batcher(lst):
...
+            assert (
+                eid_attr not in edge_attrs
+            ), f"'{eid_attr}' already used as edge ID attribute, please provide a different value for etype_attr"
```
mufeili, line 1716:
```suggestion
            ), f"'{eid_attr}' already used as edge ID attribute, please provide a different value for eid_attr"
```

python/dgl/convert.py
nvitucci: Of course. Done.

mufeili:
@dgl-bot

mufeili:
I'm good. Thanks for the great efforts.

nvitucci:
> I'm good. Thanks for the great efforts!

Thank you very much for the assistance, @mufeili - it was great to pair with you!

mufeili:
Lint check failed

```
python/dgl/convert.py:1696:0: C0301: Line too long (117/100) (line-too-long)
python/dgl/convert.py:1716:0: C0301: Line too long (117/100) (line-too-long)
python/dgl/convert.py:1719:0: C0301: Line too long (111/100) (line-too-long)
python/dgl/convert.py:1759:0: C0301: Line too long (105/100) (line-too-long)
python/dgl/convert.py:1761:0: C0301: Line too long (115/100) (line-too-long)
python/dgl/convert.py:1763:0: C0301: Line too long (108/100) (line-too-long)
python/dgl/convert.py:1827:0: C0301: Line too long (101/100) (line-too-long)
python/dgl/convert.py:1831:0: C0301: Line too long (102/100) (line-too-long)
python/dgl/convert.py:1832:0: C0301: Line too long (101/100) (line-too-long)
```

nvitucci:
> Lint check failed
> 
> ```
> python/dgl/convert.py:1696:0: C0301: Line too long (117/100) (line-too-long)
> python/dgl/convert.py:1716:0: C0301: Line too long (117/100) (line-too-long)
> python/dgl/convert.py:1719:0: C0301: Line too long (111/100) (line-too-long)
> python/dgl/convert.py:1759:0: C0301: Line too long (105/100) (line-too-long)
> python/dgl/convert.py:1761:0: C0301: Line too long (115/100) (line-too-long)
> python/dgl/convert.py:1763:0: C0301: Line too long (108/100) (line-too-long)
> python/dgl/convert.py:1827:0: C0301: Line too long (101/100) (line-too-long)
> python/dgl/convert.py:1831:0: C0301: Line too long (102/100) (line-too-long)
> python/dgl/convert.py:1832:0: C0301: Line too long (101/100) (line-too-long)
> ```

Strange, I didn't catch this with `ufmt`. I'll fix shortly.

mufeili:
@dgl-bot

mufeili:
```
    def test_to_networkx():
        # TODO: adapt and move code from the _test_nx_conversion function in
        # tests/python/common/function/test_basics.py to here
        # (pending resolution of https://github.com/dmlc/dgl/issues/5735).
        g = dgl.heterograph(
            {
                ("user", "follows", "user"): ([0, 1], [1, 2]),
                ("user", "follows", "topic"): ([1, 1], [1, 2]),
                ("user", "plays", "game"): ([0, 3], [3, 4]),
            }
        )
    
        n1 = F.randn((5, 3))
        n2 = F.randn((4, 2))
        e1 = F.randn((2, 3))
        e2 = F.randn((2, 2))
    
>       g.ndata["n"] = {"game": n1, "user": n2}
> FAILED tests/python/common/test_convert.py::test_to_networkx - dgl._ffi.base.DGLError: Cannot assign node feature "n" on device /gpu:0 to a graph on device /cpu:0. Call DGLGraph.to() to copy the graph to the same device.
```

mufeili:
> > ```
> >     def test_to_networkx():
> >         # TODO: adapt and move code from the _test_nx_conversion function in
> >         # tests/python/common/function/test_basics.py to here
> >         # (pending resolution of https://github.com/dmlc/dgl/issues/5735).
> >         g = dgl.heterograph(
> >             {
> >                 ("user", "follows", "user"): ([0, 1], [1, 2]),
> >                 ("user", "follows", "topic"): ([1, 1], [1, 2]),
> >                 ("user", "plays", "game"): ([0, 3], [3, 4]),
> >             }
> >         )
> >     
> >         n1 = F.randn((5, 3))
> >         n2 = F.randn((4, 2))
> >         e1 = F.randn((2, 3))
> >         e2 = F.randn((2, 2))
> >     
> > >       g.ndata["n"] = {"game": n1, "user": n2}
> > > FAILED tests/python/common/test_convert.py::test_to_networkx - dgl._ffi.base.DGLError: Cannot assign node feature "n" on device /gpu:0 to a graph on device /cpu:0. Call DGLGraph.to() to copy the graph to the same device.
> > ```
> 
> I am not sure what to do here. What do you recommend?

The CI involves tests on CPU and GPU. On GPU instances, `F.randn` creates tensors on GPU. However, your graph is always on CPU.

nvitucci:
> > > ```
> > >     def test_to_networkx():
> > >         # TODO: adapt and move code from the _test_nx_conversion function in
> > >         # tests/python/common/function/test_basics.py to here
> > >         # (pending resolution of https://github.com/dmlc/dgl/issues/5735).
> > >         g = dgl.heterograph(
> > >             {
> > >                 ("user", "follows", "user"): ([0, 1], [1, 2]),
> > >                 ("user", "follows", "topic"): ([1, 1], [1, 2]),
> > >                 ("user", "plays", "game"): ([0, 3], [3, 4]),
> > >             }
> > >         )
> > >     
> > >         n1 = F.randn((5, 3))
> > >         n2 = F.randn((4, 2))
> > >         e1 = F.randn((2, 3))
> > >         e2 = F.randn((2, 2))
> > >     
> > > >       g.ndata["n"] = {"game": n1, "user": n2}
> > > > FAILED tests/python/common/test_convert.py::test_to_networkx - dgl._ffi.base.DGLError: Cannot assign node feature "n" on device /gpu:0 to a graph on device /cpu:0. Call DGLGraph.to() to copy the graph to the same device.
> > > ```
> > 
> > 
> > I am not sure what to do here. What do you recommend?
> 
> The CI involves tests on CPU and GPU. On GPU instances, `F.randn` creates tensors on GPU. However, your graph is always on CPU. For an example of testing on different graph ID types and device, see https://github.com/dmlc/dgl/blob/master/tests/python/pytorch/nn/conv/test_gatedgcnconv.py.

I'm giving it a try with the latest commit.

mufeili:
@dgl-bot

nvitucci:
@mufeili I see now it fails on the `if g.device != F.cpu()` check. I am wondering if this method should be supported at all on GPU graphs?

mufeili:
@nvitucci I think for now it's fine to disable the unit tests for GPU.

mufeili:
@dgl-bot

PR #5730, opened 2023-05-23
[Contrib][Sampling] Temporal neighbor sampling

jermainewang:
## Description

The PR adds a new API `temporal_sample_neighbors` under `dgl.contrib.sampling`.

cc @frozenbugs 

## Checklist
- [x] Code is well-documented

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,6 +154,7 @@ file(GLOB_RECURSE DGL_SRC_1
   src/api/*.cc
   src/graph/*.cc
   src/scheduler/*.cc
+  src/contrib/graph/*.cc
 )
 
 list(APPEND DGL_SRC ${DGL_SRC_1})
--- a/include/dgl/runtime/container.h
+++ b/include/dgl/runtime/container.h
@@ -679,6 +679,29 @@ inline std::vector<T> ListValueToVector(const List<Value>& list) {
   return ret;
 }
 
+/**
+ * @brief Helper function to convert a Map<Value, Value> object to an unordered map.
+ * @tparam K key type
+ * @tparam V value type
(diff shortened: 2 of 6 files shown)


jermainewang:
> HGT proposes that during sampling, as soon as a neighbor without timestamp is accessed from a seed node, we assign the seed node's timestamp to it (Alg. 2, Line 6 in https://arxiv.org/pdf/2003.01332.pdf).

This is the same to including them during sampling. Should be a easy patch.

> Another possible option is not to require the neighbors on each layer has timestamps earlier than the nodes on the next layer, but instead just require all the sampled nodes with a timestamp in all the layers is no later than the seed nodes at the last layer.

I will suggest to create a new sampling operator so we keep both options open for experiments. Also, what if there are multiple seed nodes? Does this require the sampled nodes to have timestamp smaller than *any* of the seed nodes? or depending on their connectivity?

BarclayII:
> what if there are multiple seed nodes?

In this case, a seed node should have its own computation graph.

jermainewang:
> 

I see. That will be a more fundamental change to the existing pipeline.

BarclayII:
> > HGT proposes that during sampling, as soon as a neighbor without timestamp is accessed from a seed node, we assign the seed node's timestamp to it (Alg. 2, Line 6 in https://arxiv.org/pdf/2003.01332.pdf).
> 
> This is the same to including them during sampling. Should be a easy patch.

Do you want to include this patch in this PR or open up a separate PR?

jermainewang:
Let's suspend this PR until we are crystal clear about the request.

mfbalin:
@jermainewang Does the temporal sampling functionality cover what is implemented here?

PR #5737, opened 2023-05-24
[Graphbolt] Add SharedMemory

czkkkkkk:
## Description

We add the class of shared memory without handling unexpected termination first.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [ ] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [ ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

Issue #5690, opened 2023-05-15
[GraphBolt] New Storage API : create_heterograph_from_shared_memory


peizhou001:

czkkkkkk:
This issue is breakdown into two PRs.
1. Support shared memory in the C++ side.
2. Provide Python interface to copy/load graph and add unittests.

PR #5738, opened 2023-05-25
[Graphbolt] Read/write CSCSamplingGraph on shared memory.

czkkkkkk:
## Description
One of the work items in #5690. <strike> This PR depends on #5737. Please review #5737 first. </strike>

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/include/graphbolt/csc_sampling_graph.h
+++ b/graphbolt/include/graphbolt/csc_sampling_graph.h
@@ -167,7 +167,9 @@ class CSCSamplingGraph : public torch::CustomClassHolder {
 
   /**
    * @brief Maximum number of bytes used to serialize the metadata of the
-   * member tensors, including tensor shape and dtype.
+   * member tensors, including tensor shape and dtype. The constant is estimated
+   * by multiplying the number of tensors in this class and the maximum number
+   * of bytes used to serialize the metadata of a tensor (4 * 8192 for now).
    */
   static constexpr int64_t SERIALIZED_METAINFO_SIZE_MAX = 32768;
 
--- a/graphbolt/include/graphbolt/csc_sampling_graph.h
+++ b/graphbolt/include/graphbolt/csc_sampling_graph.h
@@ -11,7 +11,8 @@
 #include <string>
 #include <vector>
 
-#include "sampled_subgraph.h"
+#include "./sampled_subgraph.h"
+#include "./shared_memory.h"
 
 namespace graphbolt {
 namespace sampling {
@@ -44,11 +45,19 @@ class CSCSamplingGraph : public torch::CustomClassHolder {
    * present.
    * @param type_per_edge A tensor representing the type of each edge, if
    * present.
+   * @param tensor_meta_shm The shared memory used to hold the meta information
+   * of the tensors associated with this class. Nullptr means that the graph is
+   * not on shared memory.
+   * @param tensor_data_shm The shared memory used to old the data of the
+   * tensors associated with this class. Nullptr means that the graph is not on
+   * shared memory.
    */
   CSCSamplingGraph(
       torch::Tensor& indptr, torch::Tensor& indices,
       const torch::optional<torch::Tensor>& node_type_offset,
-      const torch::optional<torch::Tensor>& type_per_edge);
+      const torch::optional<torch::Tensor>& type_per_edge,
+      SharedMemoryPtr&& tensor_meta_shm = nullptr,
+      SharedMemoryPtr&& tensor_data_shm = nullptr);
 
   /**
    * @brief Create a homogeneous CSC graph from tensors of CSC format.
@@ -115,6 +124,24 @@ class CSCSamplingGraph : public torch::CustomClassHolder {
   c10::intrusive_ptr<SampledSubgraph> InSubgraph(
       const torch::Tensor& nodes) const;
 
+  /**
+   * @brief Copy the graph to shared memory.
+   * @param shared_memory_name The name of the shared memory.
+   *
+   * @return A new CSCSamplingGraph object on shared memory.
+   */
+  c10::intrusive_ptr<CSCSamplingGraph> CopyToSharedMemory(
+      const std::string& shared_memory_name);
+
+  /**
+   * @brief Load the graph from shared memory.
+   * @param shared_memory_name The name of the shared memory.
+   *
+   * @return A new CSCSamplingGraph object on shared memory.
+   */
+  static c10::intrusive_ptr<CSCSamplingGraph> LoadFromSharedMemory(
+      const std::string& shared_memory_name);
+
  private:
   /** @brief CSC format index pointer array. */
   torch::Tensor indptr_;
@@ -137,6 +164,20 @@ class CSCSamplingGraph : public torch::CustomClassHolder {
    * edge types. The length of it is equal to the number of edges.
    */
   torch::optional<torch::Tensor> type_per_edge_;
+
+  /**
+   * @brief Maximum number of bytes used to serialize the metadata of the
+   * member tensors, including tensor shape and dtype.
+   */
+  static constexpr int64_t SERIALIZED_METAINFO_SIZE_MAX = 32768;
+
+  /**
+   * @brief Shared memory used to hold the tensor meta information and data of
+   * this class. By storing its shared memory objects, the graph controls the
+   * resources of shared memory, which will be released automatically when the
+   * graph is destroyed.
+   */
+  SharedMemoryPtr tensor_meta_shm_, tensor_data_shm_;
 };
 
 }  // namespace sampling
--- a/graphbolt/include/graphbolt/csc_sampling_graph.h
+++ b/graphbolt/include/graphbolt/csc_sampling_graph.h
@@ -45,19 +45,11 @@ class CSCSamplingGraph : public torch::CustomClassHolder {
    * present.
    * @param type_per_edge A tensor representing the type of each edge, if
    * present.
-   * @param tensor_meta_shm The shared memory used to hold the meta information
-   * of the tensors associated with this class. Nullptr means that the graph is
-   * not on shared memory.
-   * @param tensor_data_shm The shared memory used to old the data of the
-   * tensors associated with this class. Nullptr means that the graph is not on
-   * shared memory.
    */
   CSCSamplingGraph(
       torch::Tensor& indptr, torch::Tensor& indices,
       const torch::optional<torch::Tensor>& node_type_offset,
-      const torch::optional<torch::Tensor>& type_per_edge,
-      SharedMemoryPtr&& tensor_meta_shm = nullptr,
-      SharedMemoryPtr&& tensor_data_shm = nullptr);
+      const torch::optional<torch::Tensor>& type_per_edge);
 
   /**
    * @brief Create a homogeneous CSC graph from tensors of CSC format.
@@ -143,6 +135,20 @@ class CSCSamplingGraph : public torch::CustomClassHolder {
       const std::string& shared_memory_name);
 
  private:
+  /**
+   * @brief Build a CSCSamplingGraph from shared memory tensors.
+   *
+   * @param shared_memory_tensors A tuple of two share memory objects holding
+   * tensor meta information and data respectively, and a vector of optional
+   * tensors on shared memory.
+   *
+   * @return A new CSCSamplingGraph on shared memory.
+   */
+  static c10::intrusive_ptr<CSCSamplingGraph> BuildGraphFromSharedMemoryTensors(
+      std::tuple<
+          SharedMemoryPtr, SharedMemoryPtr,
+          std::vector<torch::optional<torch::Tensor>>>&& shared_memory_tensors);
+
   /** @brief CSC format index pointer array. */
   torch::Tensor indptr_;
 
--- a/graphbolt/src/csc_sampling_graph.cc
+++ b/graphbolt/src/csc_sampling_graph.cc
@@ -120,45 +120,42 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
           : torch::nullopt);
 }
 
+static c10::intrusive_ptr<CSCSamplingGraph> BuildGraphFromSharedMemoryTensors(
+    SharedMemoryTensors&& shared_memory_tensors) {
+  const auto& tensors = std::get<2>(shared_memory_tensors);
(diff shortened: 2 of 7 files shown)


graphbolt/src/shm/utils.cc
```diff
@@ -0,0 +1,146 @@
...
+ * @brief Load tensors data from shared memory.
+ * @param shared_memory_name The name of shared memory.
+ * @param tensor_metas The meta info of tensors, including tensor shape and
+ * dtype.
+ *
+ * @return A pair of shared memory holing the tensors and the tensors.
+ */
+static std::pair<SharedMemoryPtr, std::vector<torch::Tensor>>
+_LoadTensorsDataFromSharedMemory(
+    const std::string& shared_memory_name,
```
frozenbugs, line 70: not used?

czkkkkkk: Used by `LoadTensorsFromSharedMemory`.

czkkkkkk:
@dgl-bot

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -115,5 +120,46 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+          shm_name_with_prefix, tensors, SERIALIZED_METAINFO_SIZE_MAX);
+  return c10::make_intrusive<CSCSamplingGraph>(
+      shared_tensors[0], shared_tensors[1],
+      shared_tensors.size() > 2
+          ? torch::optional<torch::Tensor>(shared_tensors[2])
+          : torch::optional<torch::Tensor>(),
+      shared_tensors.size() > 3
+          ? torch::optional<torch::Tensor>(shared_tensors[3])
+          : torch::optional<torch::Tensor>(),
+      std::move(tensor_meta_shm), std::move(tensor_data_shm));
```
frozenbugs, line 144: After the renaming in graphbolt/src/shared_memory_utils.h, I would create a helper function in this cc file to covert SharedMemoryTensors to c10::make_intrusive<CSCSamplingGraph>.

graphbolt/src/shared_memory_utils.cc
```diff
@@ -0,0 +1,146 @@
...
+#include "./shared_memory_utils.h"
+
+#include <graphbolt/serialize.h>
+#include <graphbolt/shared_memory.h>
+
+namespace graphbolt {
+namespace sampling {
+
+static SharedMemoryPtr CopyTorchArchiveToSharedMemory(
+    const std::string& shared_memory_name, int64_t memory_size,
```
frozenbugs, line 16: same here, memory_size -> size, since under this scope, it is clear that the size is archive memory size.

frozenbugs: but in CopyTensorsToSharedMemory, it makes sense to use meta_memory_size to clarify.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -115,5 +120,46 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
           : torch::nullopt);
 }
 
+c10::intrusive_ptr<CSCSamplingGraph> CSCSamplingGraph::CopyToSharedMemory(
+    const std::string& shared_memory_name) {
+  auto shm_name_with_prefix = std::string("graphbolt_") + shared_memory_name;
```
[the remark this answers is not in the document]
czkkkkkk: Yes. Removed.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -115,5 +120,46 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+  std::vector<torch::Tensor> tensors = {indptr_, indices_};
+  if (node_type_offset_.has_value()) {
+    tensors.push_back(node_type_offset_.value());
+    tensors.push_back(type_per_edge_.value());
```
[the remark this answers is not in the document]
czkkkkkk: Done

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -120,45 +120,42 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+                         : torch::nullopt,
+      std::move(std::get<0>(shared_memory_tensors)),
+      std::move(std::get<1>(shared_memory_tensors)));
+}
+
 c10::intrusive_ptr<CSCSamplingGraph> CSCSamplingGraph::CopyToSharedMemory(
     const std::string& shared_memory_name) {
-  auto shm_name_with_prefix = std::string("graphbolt_") + shared_memory_name;
   std::vector<torch::Tensor> tensors = {indptr_, indices_};
+  TORCH_CHECK(
```
czkkkkkk, line 139: @frozenbugs I add a check here to avoid the case when the node_type_offset_ is None and type_per_edge_ is not None.

graphbolt/include/graphbolt/csc_sampling_graph.h
```diff
@@ -137,6 +164,20 @@ class CSCSamplingGraph : public torch::CustomClassHolder {
    * edge types. The length of it is equal to the number of edges.
    */
   torch::optional<torch::Tensor> type_per_edge_;
+
+  /**
+   * @brief Maximum number of bytes used to serialize the metadata of the
+   * member tensors, including tensor shape and dtype.
+   */
+  static constexpr int64_t SERIALIZED_METAINFO_SIZE_MAX = 32768;
```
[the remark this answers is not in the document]
frozenbugs: Mention this in the doc, and clarify it is calculated based what, such that if ppl change this class, they know how to reestimate the constant

graphbolt/src/csc_sampling_graph.cc
frozenbugs: Why we decided to not support a graph with type_per_edge but not node_type_offset?

czkkkkkk:
@dgl-bot

graphbolt/src/shared_memory_utils.cc
```diff
@@ -0,0 +1,248 @@
...
+    } else {
+      optional_tensors.push_back(torch::nullopt);
+    }
+  }
+  return std::make_tuple(
+      std::move(shm), std::move(tensors), std::move(optional_tensors));
+}
+
+SharedMemoryTensors CopyTensorsToSharedMemory(
+    const std::string& name, const std::vector<torch::Tensor>& tensors,
```
frozenbugs, line 141: make every tensor optional can save a lot of code.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -115,5 +120,30 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+
+c10::intrusive_ptr<CSCSamplingGraph> CSCSamplingGraph::CopyToSharedMemory(
+    const std::string& shared_memory_name) {
+  auto shared_memory_tensors = CopyTensorsToSharedMemory(
+      shared_memory_name, {indptr_, indices_},
+      {node_type_offset_, type_per_edge_}, SERIALIZED_METAINFO_SIZE_MAX);
+  return BuildGraphFromSharedMemoryTensors(std::move(shared_memory_tensors));
+}
+
+c10::intrusive_ptr<CSCSamplingGraph> CSCSamplingGraph::LoadFromSharedMemory(
```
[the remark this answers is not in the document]
czkkkkkk: This is the definition of a static member function. So no need to use `static`

graphbolt/include/graphbolt/csc_sampling_graph.h
```diff
@@ -44,11 +45,19 @@ class CSCSamplingGraph : public torch::CustomClassHolder {
...
+   * tensors associated with this class. Nullptr means that the graph is not on
+   * shared memory.
    */
   CSCSamplingGraph(
       torch::Tensor& indptr, torch::Tensor& indices,
       const torch::optional<torch::Tensor>& node_type_offset,
-      const torch::optional<torch::Tensor>& type_per_edge);
+      const torch::optional<torch::Tensor>& type_per_edge,
+      SharedMemoryPtr&& tensor_meta_shm = nullptr,
+      SharedMemoryPtr&& tensor_data_shm = nullptr);
```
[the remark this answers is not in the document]
czkkkkkk: I'd prefer the second solution, although it needs to define `BuildGraphFromSharedMemoryTensors` as a private static member function. The first solution may have overloading conflict since the only difference of the two constructors is from the default arguments.

graphbolt/src/shared_memory_utils.cc
czkkkkkk: OK

graphbolt/include/graphbolt/csc_sampling_graph.h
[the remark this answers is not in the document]
frozenbugs: Make sense.

graphbolt/src/shared_memory_utils.h
```diff
@@ -0,0 +1,69 @@
...
+ * 2. Copy tensors to shared memory `shared_memory_name + "_data"`, which can be
+ * loaded by other processes with meta info.
+ *
+ * @param name The name of shared memory.
+ * @param optional_tensors The optional tensors to copy.
+ * @param max_meta_memory_size The maximum size of meta memory.
+ *
+ * @return A tuple of tensor meta shared memory, tensor data shared memory, and
+ * shared optional tensors.
+ */
```
frozenbugs, line 48: 1. Since every tensor is optional_tensor, so it is OK to replace every optional_tensor to just tensor.
2. It is important to mention that the order of tensors loaded from LoadTensorsFromSharedMemory will be exactly the same as the tensors copied from CopyTensorsToSharedMemory.

frozenbugs (approved):

graphbolt/src/shared_memory_utils.h
czkkkkkk: Done

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -115,5 +117,35 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
           : torch::nullopt);
 }
 
+c10::intrusive_ptr<CSCSamplingGraph>
+CSCSamplingGraph::BuildGraphFromSharedMemoryTensors(
+    std::tuple<
```
Rhett-Ying, line 122: why not use `SharedMemoryTensors` defined in `shared_memory_utils.h`? for clarity?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -115,5 +117,35 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+        SharedMemoryPtr, SharedMemoryPtr,
+        std::vector<torch::optional<torch::Tensor>>>&& shared_memory_tensors) {
```
Rhett-Ying, line 124: why `std::vector<>` is used instead of `std::tuple<>` or explicitly defined struct? access via index like `0`, `1` looks weird to me. Is this the best practice? just curious.

graphbolt/src/shared_memory_utils.cc
```diff
@@ -0,0 +1,178 @@
...
+      metas.push_back({true, shape, dtype});
+    } else {
+      metas.push_back({false, {}, torch::ScalarType::Undefined});
+    }
+  }
+  SharedMemoryPtr data_shm;
+  std::vector<torch::optional<torch::Tensor>> ret_tensors;
+  std::tie(data_shm, ret_tensors) =
+      LoadTensorsDataFromSharedMemory(name + "_data", metas);
+  return std::make_tuple(
```
Rhett-Ying, line 173: I am wondering if we could utilize `std::tuple_cat()` here. or we could avoid explicit definition of `data_shm` and `ret_tensors` which looks a bit unnecessary via `auto&& tensor_data = ...` and access via `std::pair<>.first/second`?

graphbolt/src/csc_sampling_graph.cc
czkkkkkk: Yes, I think it is more clear for a header file. And also `shared_memory_utils.h` is not in the include directory and I don't want to move it.

graphbolt/src/csc_sampling_graph.cc
czkkkkkk: `vector` is used here considering we may add more tensors in CSCSamplingGraph while the tensors loaded from shared memory are also in a vector.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -115,5 +117,35 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+  auto& optional_tensors = std::get<2>(shared_memory_tensors);
```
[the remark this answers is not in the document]
czkkkkkk: Agree. `std::tuple` is not a good solution. Another solution is to use a `struct` but it may overcomplicate things.

graphbolt/src/csc_sampling_graph.cc
Rhett-Ying: but if we add more tensors, we have to manually update in line 127 with `optional_tensors[4/5/6]`? Is it possible to remove such explicit index access? a new constructor or variadic template(parameters pack)?

graphbolt/src/csc_sampling_graph.cc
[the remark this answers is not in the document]
Rhett-Ying: for tuple_size <=3, I think it's acceptable to use tuple. we could turn to struct when tuple size increase in the future.

graphbolt/src/shared_memory_utils.cc
czkkkkkk: I just learned `std::tuple_cat()`. It is a great function to simplify code in this situation, which involves tuple or pair as returned value. We may optimize this code in the future.

graphbolt/src/csc_sampling_graph.cc
czkkkkkk: I feel the index access of a vector is unavoidable because we always need to load the tensors of a class from a vector. What do you mean by a new constructor or variadic template?

Rhett-Ying: i don't know yet. could the size of vector be determined in compile stage?

PR #5739, opened 2023-05-25
[Model] Implemented PGExplainer for Heterogeneous graph

ndbaker1:
## Description

Implemented PGExplainer (https://arxiv.org/pdf/2011.04573.pdf) for Heterogeneous Graphs.

## Checklist
- [x] Code is well-documented

## Changes

Added a new class HeteroPGExplainer for Heterogeneous graph explanation.

The output of the model has been manually verified on MUTAG dataset. A visualization script was created to visualize the mutag dataset (external ref: https://gist.github.com/ndbaker1/68b956f9232acc948116e90d3aad14af). Unfortunately, there's no ground truth for the explanation, so we cannot quantify the model's performance. Instead, we manually investigated the results and confirmed that they aligned well with human intuition.

Inline documentation is added following the original style, and an example of using the module is included.


mufeili:
I have a high-level question. Is it sufficient to call `to_homogeneous` first and then use the previous PGExplainer implementation for homogeneous graphs? If not, what will be the gaps?

ndbaker1:
> I have a high-level question. Is it sufficient to call `to_homogeneous` first and then use the previous PGExplainer implementation for homogeneous graphs? If not, what will be the gaps?

I believe the gap in this approach is that the Heterograph classifier will not be usable by existing PGExplainer when performing model evaluations

The approach I've gone with has been to store the edge mask in its homograph format in order to keep the internals of PGExplainer intact, and use the context from the `EIDS` data to reobtain the heterograph edge mask

mufeili:
@dgl-bot

python/dgl/nn/pytorch/explain/pgexplainer.py
```diff
@@ -358,3 +360,382 @@ def explain_graph(self, graph, feat, tmp=1.0, training=False, **kwargs):
         self.clear_masks()
 
         return (probs, edge_mask) if training else (probs.data, edge_mask)
+
+
+class HeteroPGExplainer(nn.Module):
```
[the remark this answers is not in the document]
ndbaker1: makes sense, it should save lots on repetition since only `explain_graph` and `train_step` require changes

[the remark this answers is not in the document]
ndbaker1: before this, do you approve of removing the `feat` parameter completely from [`set_masks`](https://github.com/dmlc/dgl/blob/147f6d243532e52c082623463d69cf444030112c/python/dgl/nn/pytorch/explain/pgexplainer.py#L67-L85) in `PGExplainer` and replacing the `num_nodes` lines with:

```python
num_nodes, _ = feat.shape
```

with

```python
num_nodes = graph.num_nodes()
```

this would make the function usable within the hetero impl. as well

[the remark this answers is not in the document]
mufeili: Sure.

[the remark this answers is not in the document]
ndbaker1: done 👍

mufeili:
@dgl-bot

mufeili:
Lint check failed

```
python/dgl/nn/pytorch/explain/pgexplainer.py:391:4: W0235: Useless super delegation in method '__init__' (useless-super-delegation)
```

python/dgl/nn/pytorch/explain/pgexplainer.py
```diff
@@ -358,3 +357,243 @@ def explain_graph(self, graph, feat, tmp=1.0, training=False, **kwargs):
...
+        ...
+        ...         if embed:
+        ...             return h
+        ...
+        ...         with g.local_scope():
+        ...             g.ndata["h"] = h
+        ...             hg = 0
+        ...             for ntype in g.ntypes:
+        ...                 hg = hg + dgl.mean_nodes(g, "h", ntype=ntype)
+        ...             return th.sigmoid(self.fc(hg))
```
mufeili, line 512: Using `th.sigmoid` here may not be appropriate. In particular, `out_feats` is not necessarily 2.

python/dgl/nn/pytorch/explain/pgexplainer.py
```diff
@@ -358,3 +357,243 @@ def explain_graph(self, graph, feat, tmp=1.0, training=False, **kwargs):
...
+        super(HeteroPGExplainer, self).__init__(
+            model,
+            num_features,
+            coff_budget,
+            coff_connect,
+            sample_bias,
+        )
+
+    def train_step(self, graph, feat, tmp, **kwargs):
+        r"""Compute the loss of the explanation network
```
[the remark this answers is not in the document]
ndbaker1: cool, done

python/dgl/nn/pytorch/explain/pgexplainer.py
ndbaker1: you are right I forgot to fix this, done now.

python/dgl/nn/pytorch/explain/pgexplainer.py
```diff
@@ -358,3 +356,227 @@ def explain_graph(self, graph, feat, tmp=1.0, training=False, **kwargs):
...
+        ...                 etype: {"edge_weight": mask} for etype, mask in edge_weight.items()
+        ...             }
+        ...             h = self.conv(g, h, mod_kwargs=mod_kwargs)
+        ...         else:
+        ...             h = self.conv(g, h)
+        ...
+        ...         if embed:
+        ...             return h
+        ...
+        ...         with g.local_scope():
```
mufeili, line 490: Could you also update the example for homogeneous graphs regarding the use of `if embed:` and `with g.local_scope():`?

python/dgl/nn/pytorch/explain/pgexplainer.py
```diff
@@ -358,3 +356,227 @@ def explain_graph(self, graph, feat, tmp=1.0, training=False, **kwargs):
...
+        >>> num_classes = 2
+        >>> g = dgl.heterograph({("user", "plays", "game"): ([0, 1, 1, 2], [0, 0, 1, 1])})
+        >>> g.nodes["user"].data["h"] = th.randn(g.num_nodes("user"), input_dim)
+        >>> g.nodes["game"].data["h"] = th.randn(g.num_nodes("game"), input_dim)
+
+        >>> transform = dgl.transforms.AddReverse()
+        >>> g = transform(g)
+
+        >>> # define and train the model
+        >>> model = Model(input_dim, hidden_dim, num_classes, g.canonical_etypes)
```
mufeili, line 509: Maybe you don't need `hidden_dim` for a demo. Simply using `input_dim` can be sufficient.

tests/python/pytorch/nn/test_nn.py
```diff
@@ -1839,6 +1839,64 @@ def forward(self, g, h, embed=False, edge_weight=None):
     probs, edge_weight = explainer.explain_graph(g, feat)
 
 
+@pytest.mark.parametrize("g", get_cases(["hetero"], exclude=["zero-degree"]))
```
mufeili, line 1842: Does this test cover batched graphs?

mufeili:
@dgl-bot

python/dgl/nn/pytorch/explain/pgexplainer.py
```diff
@@ -358,3 +356,227 @@ def explain_graph(self, graph, feat, tmp=1.0, training=False, **kwargs):
...
+        self.elayers = self.elayers.to(graph.device)
+
+        pred = self.model(graph, feat, embed=False, **kwargs).argmax(-1).data
+
+        prob, hetero_edge_mask = self.explain_graph(
+            graph, feat, tmp=tmp, training=True, **kwargs
+        )
+
+        # store edge mask in homogeneous format
+        self.edge_mask = torch.concat(list(hetero_edge_mask.values()))
```
mufeili, line 422: Is this guaranteed to be the order of the node/edge types between the conversion of homogeneous graphs and heterogeneous graphs?

python/dgl/nn/pytorch/explain/pgexplainer.py
```diff
@@ -358,3 +356,227 @@ def explain_graph(self, graph, feat, tmp=1.0, training=False, **kwargs):
...
+
+        >>> # Explain the graph
+        >>> feat = g.ndata.pop("h")
+        >>> probs, edge_mask = explainer.explain_graph(g, feat)
+        """
+        self.model = self.model.to(graph.device)
+        self.elayers = self.elayers.to(graph.device)
+
+        embed = self.model(graph, feat, embed=True, **kwargs)
+        embed_flat = torch.concat(list(embed.values()))
```
mufeili, line 540: ditto as above

mufeili, line 540: For safely, it might be better to detach the embeddings from the computation graph by `.data`.

python/dgl/nn/pytorch/explain/pgexplainer.py
```diff
@@ -358,3 +356,227 @@ def explain_graph(self, graph, feat, tmp=1.0, training=False, **kwargs):
...
+        values = emb.reshape(-1)
+
+        values = self.concrete_sample(values, beta=tmp, training=training)
+        self.sparse_mask_values = values
+
+        reverse_eids = homo_graph.edge_ids(row, col).long()
+        edge_mask = (values + values[reverse_eids]) / 2
+
+        self.clear_masks()
+        self.set_masks(homo_graph, edge_mask)
```
mufeili, line 560: Do you still need to call `self.clear_masks()` and `self.set_masks` inside `explain_graph`? Is it sufficient to call them inside `train_step`?

mufeili (commented):
done a pass, some comments

python/dgl/nn/pytorch/explain/pgexplainer.py
ndbaker1: sure thing, I will standardize the usage of this across examples and tests

tests/python/pytorch/nn/test_nn.py
ndbaker1: sorry has me a bit confused, should I try to cover all possible heterographs in `get_cases`?

something like this:
```suggestion
@pytest.mark.parametrize("g", get_cases(["hetero", "bipartite", "block"], exclude=["block-bipartite"]))
```

python/dgl/nn/pytorch/explain/pgexplainer.py
ndbaker1: done

ndbaker1: I have updated it to let `to_homogeneous` handle the merge like you mentioned

python/dgl/nn/pytorch/explain/pgexplainer.py
ndbaker1: Initially I relied on the determinism of the edge dict, but based on the convs. below I think it makes more sense to retain `edge_mask` from the `explain_graph` step and avoid the redundant conversion.

let me know what you think of the changes once they are visible

python/dgl/nn/pytorch/explain/pgexplainer.py
ndbaker1: I don't see the need for the `clear_masks` calls as well.

with respect to the changes related to an [above comment](https://github.com/dmlc/dgl/pull/5739#discussion_r1214815720), I can keep the final clear based on the `training` param.

ndbaker1: I also think the `clear_masks` call is only necessary when there is no training, so following changes from [convo. above](https://github.com/dmlc/dgl/pull/5739#discussion_r1214815720) I think it should suffice to clear the mask when it is not in training setting

ndbaker1: I don't personally see any cases where `set_masks` is necessary inside the class because the steps from `explain_graph` shouldn't yeild a `None` for `edge_mask`?, but don't want to be too eager about pruning code.

tests/python/pytorch/nn/test_nn.py
mufeili: From the documentation you added, the module supports a batched graph, one that is created with [dgl.batch](https://docs.dgl.ai/generated/dgl.batch.html#dgl.batch). However, previously the unit test did not cover a batched heterogeneous graph. I was suggesting the expansion of the unit test to cover batched heterogeneous graphs. You don't necessarily need to use `get_cases`. It's fine to directly create graphs inside the unit test function.

python/dgl/nn/pytorch/explain/pgexplainer.py
mufeili: What if we reply on `to_homogeneous` to perform this kind of conversion? In this way, it's guaranteed that the order of edge-specific information is consistent with the result of `to_homogeneous`.

python/dgl/nn/pytorch/explain/pgexplainer.py
```diff
@@ -348,13 +346,225 @@ def explain_graph(self, graph, feat, tmp=1.0, training=False, **kwargs):
...
+            The temperature parameter fed to the sampling procedure.
+        kwargs : dict
+            Additional arguments passed to the GNN model.
+
+        Returns
+        -------
+        Tensor
+            A scalar tensor representing the loss.
+        """
+        return super().train_step(graph, feat, tmp=tmp, **kwargs)
```
mufeili, line 412: Do you still need L390-412 then?

python/dgl/nn/pytorch/explain/pgexplainer.py
ndbaker1: will make sure the conversions are always obtained using `to_homogeneous` 👍

tests/python/pytorch/nn/test_nn.py
ndbaker1: oh i understand now, sorry for the confusion. It makes sense to me to keep a case for this so let me try to make an addition to `get_cases`

ndbaker1: added

python/dgl/nn/pytorch/explain/pgexplainer.py
mufeili: I see. I don't have a better solution. Let's keep it as so for now.

mufeili:
@dgl-bot

python/dgl/nn/pytorch/explain/pgexplainer.py
```diff
@@ -348,13 +346,225 @@ def explain_graph(self, graph, feat, tmp=1.0, training=False, **kwargs):
...
+        >>> for epoch in range(20):
+        ...     tmp = float(init_tmp * np.power(final_tmp / init_tmp, epoch / 20))
+        ...     loss = explainer.train_step(g, g.ndata["h"], tmp)
+        ...     optimizer_exp.zero_grad()
+        ...     loss.backward()
+        ...     optimizer_exp.step()
+
+        >>> # Explain the graph
+        >>> feat = g.ndata.pop("h")
+        >>> probs, edge_mask = explainer.explain_graph(g, feat)
```
mufeili, line 521: One last comment. I got 

```
{('game',
  'rev_plays',
  'user'): tensor([3.0763e-04, 1.3422e-04, 9.6319e-07, 2.5353e-05],
        grad_fn=<IndexBackward0>),
 ('user',
  'plays',
  'game'): tensor([3.0763e-04, 1.3422e-04, 9.6319e-07, 2.5353e-05],
        grad_fn=<IndexBackward0>)}
```

for `edge_mask`. In other words, the reverse edges have the same importance as the forward edges. Is this expected from the perspective of `PGExplainer`?

kunmukh: `
To explain predictions made by a GNN model, we divide the original input graph Go into two subgraphs: Go = Gs + ∆G, where Gs presents the underlying subgraph that makes important contributions to GNN’s predictions, which is the expected explanatory graph, and ∆G consists of the remaining task-irrelevant edges for predictions made by the GNN
`

From the above line in the paper, I understand that when we select an edge as important (e.g., be part of `Gs`), we also select its reverse edge.  Therefore, the reverse edges are expected to have the same importance as the forward edges of the selected edges.

ndbaker1: let me review the reference material for any other possible remarks

ndbaker1: code-wise this is a result of the following logic within the `explain_graph` call, which obtains symmetric edge weights:

```python
reverse_eids = homo_graph.edge_ids(row, col).long()
edge_mask = (values + values[reverse_eids]) / 2
```

If I had to rationalize this, I would say it is due to the implicit requirement that every node has a corresponding embedding, which explains the addition of reverse-edges in order to guarantee global message-passing coverage. Then as a result we assume that these edges share the same score?

mufeili: @ndbaker1 @kunmukh Is this from a prior implementation? I did not find this in the paper.

kunmukh: In the original implementation, https://github.com/flyingdoog/PGExplainer/blob/master/codes/Explainer.py#L27, I saw this line `sym_mask = (sym_mask + tf.transpose(sym_mask)) / 2`, where they calculate a symmetric mask by adding sym_mask with its transpose and dividing the result by 2. This ensures that the mask is symmetric along the diagonal.

Since the mask is symmetric along the diagonal, the edges and reverse edges have `edge_mask = (values + values[reverse_eids]) / 2` . 

@ndbaker1, please correct me if I missed something.

mufeili: I see. Thanks. I'm good then.

mufeili (approved):

PR #5743, opened 2023-05-26
[GraphBolt] add in_subgraph() API at the python level

Rhett-Ying:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/graph_storage/csc_sampling_graph.py
+++ b/python/dgl/graphbolt/graph_storage/csc_sampling_graph.py
@@ -189,6 +189,8 @@ class CSCSamplingGraph:
         SampledSubgraph
             The in subgraph.
         """
+        # Ensure nodes is 1-D tensor.
+        assert nodes.dim() == 1, "Nodes should be 1-D tensor."
         # Ensure that there are no duplicate nodes.
         assert len(torch.unique(nodes)) == len(
             nodes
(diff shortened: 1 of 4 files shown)


python/dgl/graphbolt/graph_storage/csc_sampling_graph.py
```diff
@@ -173,6 +173,26 @@ def metadata(self) -> Optional[GraphMetadata]:
...
+        nodes : torch.Tensor
+            The nodes to form the subgraph which are type agnostic.
+
+        Returns
+        -------
+        SampledSubgraph
+            The in subgraph.
+        """
+        # Remove duplicate nodes.
+        nodes = torch.unique(nodes)
```
frozenbugs, line 193: In which case user will call in_subgraph with duplicated nodes? Is it a valid query?

Rhett-Ying: just in case of user pass in duplicate node in mistake.

frozenbugs: If it is a user mistake, let's catch it and throw an error rather than fix it silently.

Rhett-Ying: updated.

python/dgl/graphbolt/graph_storage/csc_sampling_graph.py
```diff
@@ -173,6 +173,28 @@ def metadata(self) -> Optional[GraphMetadata]:
...
+        ----------
+        nodes : torch.Tensor
+            The nodes to form the subgraph which are type agnostic.
+
+        Returns
+        -------
+        SampledSubgraph
+            The in subgraph.
+        """
+        # Remove duplicate nodes.
```
frozenbugs, line 192:
```suggestion
        # Ensure that there are no duplicate nodes.
```

frozenbugs (approved):

PR #5757, opened 2023-06-01
[GraphBolt] Add common macro

peizhou001:
## Description
Add `macro.h` used for common macros in graphbolt, currently only one used to dispatch type.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/src/macro.h
+++ b/graphbolt/src/macro.h
@@ -1,0 +1,51 @@
+/**
+ *  Copyright (c) 2023 by Contributors
+ * @file graphbolt/src/macro.h
+ * @brief Common macros for graphbolt package.
+ */
+
+#ifndef GRAPHBOLT_MACRO_H_
+#define GRAPHBOLT_MACRO_H_
+
+#include <torch/torch.h>
+
+/**
+ * Dispatch according to torch scalar type.
+ *
+ * ATEN_ID_TYPE_SWITCH(type_per_edge.dtype(), DataType, {
+ *   // Now DataType is the type corresponding to data type in type_per_edge.
+ *   // For instance, one can do this for a CPU array:
+ *   DataType *data = static_cast<DataType *>(type_per_edge.data_ptr());
+ * });
+ */
+#define ATEN_DATA_TYPE_SWITCH(val, DataType, ...)                           \
+  do {                                                                      \
+    if ((val) == torch::kInt8 || (val) == torch::kUInt8 ||                  \
+        (val) == torch::kBool) {                                            \
+      typedef int8_t DataType;                                              \
+      { __VA_ARGS__ }                                                       \
+    } else if ((val) == torch::kInt16) {                                    \
+      typedef int16_t DataType;                                             \
+      { __VA_ARGS__ }                                                       \
+    } else if ((val) == torch::kInt32) {                                    \
+      typedef uint32_t DataType;                                            \
+      { __VA_ARGS__ }                                                       \
+    } else if ((val) == torch::kInt64) {                                    \
+      typedef uint64_t DataType;                                            \
+      { __VA_ARGS__ }                                                       \
+    }                                                                       \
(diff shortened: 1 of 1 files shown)


graphbolt/src/macro.h
```diff
@@ -0,0 +1,51 @@
...
+      { __VA_ARGS__ }                                                       \
+    } else if ((val) == torch::kInt32) {                                    \
+      typedef uint32_t DataType;                                            \
+      { __VA_ARGS__ }                                                       \
+    } else if ((val) == torch::kInt64) {                                    \
+      typedef uint64_t DataType;                                            \
+      { __VA_ARGS__ }                                                       \
+    }                                                                       \
+  }                                                                         \
+  else if ((val) == torch::kFloat) {                                        \
```
frozenbugs, line 38: This block is not correctly aligned, something wrong with the braces.

graphbolt/src/macro.h
```diff
@@ -0,0 +1,51 @@
...
+    typedef float ProbType;                                                 \
+    { __VA_ARGS__ }                                                         \
+  }                                                                         \
+  else if ((val) == torch::kDouble) {                                       \
+    typedef double ProbType;                                                \
+    { __VA_ARGS__ }                                                         \
+    else {                                                                  \
+      LOG(FATAL) << "Probs can only be bool, uint8, int8, float or double"; \
```
Rhett-Ying, line 46: `LOG(FATAL)` is defined in `third_party/dmlc-core/include/dmlc/logging.h`?  Is it a best practice to use such a macro here which apparently does not include `logging.h` here.

graphbolt/src/macro.h
```diff
@@ -0,0 +1,51 @@
+/**
```
Rhett-Ying, line 1: this file is placed under `//src`, so it will not be exported as header file on purpose?

graphbolt/src/macro.h
```diff
@@ -0,0 +1,51 @@
...
+ *   // Now DataType is the type corresponding to data type in type_per_edge.
+ *   // For instance, one can do this for a CPU array:
+ *   DataType *data = static_cast<DataType *>(type_per_edge.data_ptr());
+ * });
+ */
+#define ATEN_DATA_TYPE_SWITCH(val, DataType, ...)                           \
+  do {                                                                      \
+    if ((val) == torch::kInt8 || (val) == torch::kUInt8 ||                  \
+        (val) == torch::kBool) {                                            \
+      typedef int8_t DataType;                                              \
```
Rhett-Ying, line 25: picky: `using DataType = int8_t;` instead of `typedef` which is less modern?

graphbolt/src/macro.h
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Dispatch according to torch scalar type.
+ *
+ * ATEN_ID_TYPE_SWITCH(type_per_edge.dtype(), DataType, {
+ *   // Now DataType is the type corresponding to data type in type_per_edge.
+ *   // For instance, one can do this for a CPU array:
+ *   DataType *data = static_cast<DataType *>(type_per_edge.data_ptr());
+ * });
+ */
+#define ATEN_DATA_TYPE_SWITCH(val, DataType, ...)                           \
```
frozenbugs, line 21: What's ATEN_?

Rhett-Ying, line 21: just curious, please feel free to ignore this comment. personally I don't like this macro which define a `DataType` in runtime with multiple `if-else`. I am wondering if we could avoid this macro and make calls that rely on `DataType` to be template-like which works well with `tensor.dtype()` automatically.

jermainewang:
Why don't we use PyTorch's default macro? E.g., `AT_DISPATCH_FLOATING_TYPES` . See https://pytorch.org/tutorials/advanced/cpp_extension.html

PR #5758, opened 2023-06-01
[Graphbolt]Add threading utils

peizhou001:
## Description
Add 'threading_utils.h', currently only add one function to compute thread number.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/include/graphbolt/threading_utils.h
+++ b/graphbolt/include/graphbolt/threading_utils.h
@@ -1,0 +1,30 @@
+/**
+ *  Copyright (c) 2023 by Contributors
+ * @file graphbolt/utils.h
+ * @brief Threading utils in graphbolt.
+ */
+#ifndef GRAPHBOLT_THREADING_UTILS_H_
+#define GRAPHBOLT_THREADING_UTILS_H_
+
+#include <omp.h>
+
+namespace graphbolt {
+namespace runtime {
+
+inline size_t compute_num_threads(size_t begin, size_t end, size_t grain_size) {
+#ifdef _OPENMP
+  if (omp_in_parallel() || end - begin <= grain_size || end - begin == 1)
+    return 1;
+
+  return std::min(
+      static_cast<int64_t>(torch::get_num_threads()),
+      torch::divup(end - begin, grain_size));
+#else
+  return 1;
+#endif
+}
+
+}  // namespace runtime
+}  // namespace graphbolt
+
+#endif  // GRAPHBOLT_THREADING_UTILS_H_


graphbolt/include/graphbolt/threading_utils.h
```diff
@@ -0,0 +1,30 @@
...
+namespace graphbolt {
+namespace runtime {
+
+inline size_t compute_num_threads(size_t begin, size_t end, size_t grain_size) {
+#ifdef _OPENMP
+  if (omp_in_parallel() || end - begin <= grain_size || end - begin == 1)
+    return 1;
+
+  return std::min(
+      static_cast<int64_t>(torch::get_num_threads()),
```
frozenbugs, line 20: I don't see torch related include, this is correct?

graphbolt/include/graphbolt/threading_utils.h
```diff
@@ -0,0 +1,30 @@
...
+ * @file graphbolt/utils.h
+ * @brief Threading utils in graphbolt.
+ */
+#ifndef GRAPHBOLT_THREADING_UTILS_H_
+#define GRAPHBOLT_THREADING_UTILS_H_
+
+#include <omp.h>
+
+namespace graphbolt {
+namespace runtime {
```
Rhett-Ying, line 12: as we're adding more and more files in this path, I think we need to split them into further sub-directories, especially the ones lie in different namespaces. for this one, it's in namespace `runtime` while some of other existing files are in namespace `sampling`.

graphbolt/include/graphbolt/threading_utils.h
frozenbugs: That would be the ideal case. But nowadays most of the c++ code are tested along with Python, so we don't have this requirement.

But all the code should be able to build at least.

Rhett-Ying, line 20: the return type of function is  `size_t` while it's cast to `int64_t` here. Mismatch. I don't think they are the same though `size_t` is implementation-defined.

graphbolt/include/graphbolt/threading_utils.h
frozenbugs, line 1: Does this need to be in graphbolt/include? should it be in just src?

graphbolt/include/graphbolt/threading_utils.h
```diff
@@ -0,0 +1,30 @@
...
+
+inline size_t compute_num_threads(size_t begin, size_t end, size_t grain_size) {
+#ifdef _OPENMP
+  if (omp_in_parallel() || end - begin <= grain_size || end - begin == 1)
+    return 1;
+
+  return std::min(
+      static_cast<int64_t>(torch::get_num_threads()),
+      torch::divup(end - begin, grain_size));
```
Rhett-Ying, line 21: same question, could we use `std::ceil` instead?

graphbolt/include/graphbolt/threading_utils.h
```diff
@@ -0,0 +1,30 @@
...
+#ifndef GRAPHBOLT_THREADING_UTILS_H_
+#define GRAPHBOLT_THREADING_UTILS_H_
+
+#include <omp.h>
+
+namespace graphbolt {
+namespace runtime {
+
+inline size_t compute_num_threads(size_t begin, size_t end, size_t grain_size) {
+#ifdef _OPENMP
```
Rhett-Ying, line 15: will we assume `GraphBolt` utilize OPENMP for parallel in most of time? Does `torch::parallel_for` utilize OpenMP? Will we prefer to use `torch::parallel_for`? If so, do we need to add `#ifdef` here?

graphbolt/include/graphbolt/threading_utils.h
```diff
@@ -0,0 +1,30 @@
...
+ */
+#ifndef GRAPHBOLT_THREADING_UTILS_H_
+#define GRAPHBOLT_THREADING_UTILS_H_
+
+#include <omp.h>
+
+namespace graphbolt {
+namespace runtime {
+
+inline size_t compute_num_threads(size_t begin, size_t end, size_t grain_size) {
```
Rhett-Ying, line 14: the naming style is what we're following now?

graphbolt/include/graphbolt/threading_utils.h
```diff
@@ -0,0 +1,30 @@
...
+#ifdef _OPENMP
+  if (omp_in_parallel() || end - begin <= grain_size || end - begin == 1)
+    return 1;
+
+  return std::min(
+      static_cast<int64_t>(torch::get_num_threads()),
+      torch::divup(end - begin, grain_size));
+#else
+  return 1;
```
Rhett-Ying, line 23: curious only: I am not aware of the scenario of calling this function(at least from this PR), so I don't know if it's correct to return `1` for `#else`.

PR #5759, opened 2023-06-01
[Graphbolt] Avoid transitive inclusion (include what you need)

czkkkkkk:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/src/shared_memory_utils.cc
+++ b/graphbolt/src/shared_memory_utils.cc
@@ -8,6 +8,12 @@
 
 #include <graphbolt/serialize.h>
 #include <graphbolt/shared_memory.h>
+#include <torch/torch.h>
+
+#include <cstring>
+#include <string>
+#include <tuple>
+#include <vector>
 
 namespace graphbolt {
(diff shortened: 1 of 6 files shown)


graphbolt/src/shared_memory_utils.cc
```diff
@@ -8,6 +8,12 @@
 
 #include <graphbolt/serialize.h>
 #include <graphbolt/shared_memory.h>
+#include <torch/torch.h>
+
+#include <cstring>
+#include <string>
+#include <tuple>
```
Rhett-Ying, line 15: so the guideline is whenever we need a header, we always include it explicitly even though it's already included in the corresponding header of current source file?

Rhett-Ying (approved):

graphbolt/src/shared_memory_utils.cc
czkkkkkk: Yes. Follow the guideline in [google c++ coding style](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use).

PR #5769, opened 2023-06-02
[Sparse] formats api doc checkin

keli-wen:
## Description
A new API has been designed for the `formats` and `create_formats_` function. This PR first presents the high-level Python design, with the specific design detailed in the function's docstring.

We aim to ensure backward compatibility while enabling:
- `formats` to set `allowed_formats` and designate the `create_formats` for sparse matrices.
- `create_formats_` to also designate the `create_formats` for sparse matrices.

This PR is related to [Issues: DGLGraph.formats with list input doesn't work #5545](https://github.com/dmlc/dgl/issues/5545).

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/heterograph.py
+++ b/python/dgl/heterograph.py
@@ -6065,6 +6065,88 @@ class DGLGraph(object):
         self._node_frames = old_nframes
         self._edge_frames = old_eframes
 
+    def new_formats(self, allowed_formats=None, create_formats=None):
+        r"""Get a cloned graph with the specified allowed sparse format(s)
+        and create specific sparse matrices allowed for the graph or query
+        for the usage status of sparse formats.
+
+        The API copies both the graph structure and the features.
+
+        If both ``allowed_formats`` and ``create_formats`` are None, the
+        API returns a dict recording the usage status of sparse formats.
+        In any other case, a DGLGraph is returned. The format(s) of the
+        resulting cloned graph is depends on the ``allowed_formats``,
+        ``create_formats`` and the format(s) of the original graph:
+
+        * If ``create_formats`` isn't None, the cloned graph's sparse
+          formats will be consistent with ``create_formats``.
+        * Otherwise:
+          * If the original graph's format is a subset of ``allowed_formats``,
+            the cloned graph's sparse formats will remain consistent with the
+            original.
+          * If a format in the original graph is also in ``allowed_formats``,
+            any formats not in ``allowed_formats`` will be removed in the cloned
+            graph.
+          * If the original graph has no formats in ``allowed_formats``, a new
+            format will be made for the cloned graph. This will be in the order
+            of ``'coo'``, ``'csr'``, ``'csc'``, regardless of the
+            ``allowed_formats`` order.
+
+        Parameters
+        ----------
+        allowed_formats : str or list of str or None
+
+            * If ``allowed_formats`` is None, return the usage status of sparse
+              formats.
+            * Otherwise, it can be a string ``'coo'``, ``'csr'``, or ``'csc'``
+              or a sublist of them, specifying the allowed sparse formats to
+              use.
+
+        create_formats : str or list of str or None
+
+            * If ``create_formats`` is None:
+              * If ``allowed_formats`` is None, return the usage status of
+                sparse formats.
+              * Otherwise, follow the rules above to create sparse matrices for
+                the cloned graph.
+            * Otherwise, it can be ``'coo'``, ``'csr'``, or ``'csc'`` or a
+              sublist of them, specifying the sparse matrices to create. If
+              ``allowed_formats`` is None or there are some formats in the list
+              that are not allowed for the graph, an error will be raised.
+
+        Returns
+        -------
+        dict or DGLGraph
+
+            * If ``allowed_formats`` is None, the result will be a dict
+              recording the usage status of sparse formats.
+            * Otherwise, a DGLGraph will be returned, which is a clone of the
+              original graph with the specified allowed sparse format(s) and
+              created sparse format(s).
+        """
+        pass
+
+    def new_create_formats_(self, create_formats=None):
+        r"""Create specified sparse matrices allowed for the graph.
+
+        By default, sparse matrices for a graph are created only when necessary.
+        In some cases, such as in a multi-process data loader, we may want to
+        create them immediately. This API allows for that.
+
+        Parameters
+        ----------
+        create_formats : str or list of str or None
+
+            * If ``create_formats`` is None, create all sparse matrices allowed
+              for the graph.
+            * Otherwise, it can be ``'coo'``, ``'csr'``, or ``'csc'`` or a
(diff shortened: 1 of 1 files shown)


frozenbugs:
@mufeili, this PR is just to make sure that the usage of the two api are as expected, can you take a look.
We will merge this PR first and then work on the implementation and override the existing ones in a upcoming PR.

mufeili:
What do you mean by "designate"?

python/dgl/heterograph.py
```diff
@@ -6065,6 +6065,88 @@ def local_scope(self):
         self._node_frames = old_nframes
         self._edge_frames = old_eframes
 
+    def new_formats(self, allowed_formats=None, create_formats=None):
+        r"""Get a cloned graph with the specified allowed sparse format(s)
+        and create specific sparse matrices allowed for the graph or query
+        for the usage status of sparse formats.
```
mufeili, line 6071: Do we still want to overload a single API for both setting and logging?

python/dgl/heterograph.py
```diff
@@ -6065,6 +6065,88 @@ def local_scope(self):
         self._node_frames = old_nframes
         self._edge_frames = old_eframes
 
+    def new_formats(self, allowed_formats=None, create_formats=None):
```
mufeili, line 6068: Another question is if it suffices to return a graph with a randomly selected allowed format rather than require users to specify `create_formats`.

python/dgl/heterograph.py
```diff
@@ -6065,6 +6065,88 @@ def local_scope(self):
...
+          * If the original graph's format is a subset of ``allowed_formats``,
+            the cloned graph's sparse formats will remain consistent with the
+            original.
+          * If a format in the original graph is also in ``allowed_formats``,
+            any formats not in ``allowed_formats`` will be removed in the cloned
+            graph.
+          * If the original graph has no formats in ``allowed_formats``, a new
+            format will be made for the cloned graph. This will be in the order
+            of ``'coo'``, ``'csr'``, ``'csc'``, regardless of the
+            ``allowed_formats`` order.
```
mufeili, line 6093: L6081 - 6093 appear too complex for me. Can we simplify the behaviors?

python/dgl/heterograph.py
```diff
@@ -6065,6 +6065,88 @@ def local_scope(self):
...
+
+            * If ``create_formats`` is None, create all sparse matrices allowed
+              for the graph.
+            * Otherwise, it can be ``'coo'``, ``'csr'``, or ``'csc'`` or a
+              sublist of them, specifying the sparse matrices to create. If
+              there are some formats already created, they will be skipped.
+              If there are some formats in the list that are not allowed for
+              the graph, an error will be raised.
+        """
+        pass
```
mufeili, line 6148: What are the advantages of this API over first specifying the allowed formats with one API and then create all allowed formats with another API?

python/dgl/heterograph.py
frozenbugs: Instead of having a api, maybe it is better to have an attribute for formats?

python/dgl/heterograph.py
frozenbugs: the formats() method are designed for out-place reset, and create_formats_ is designed for in-place creation, that's why we have 2 different apis

python/dgl/heterograph.py
mufeili: Yes, could be.

python/dgl/heterograph.py
mufeili: Still, that's already what `g.create_formats_()` was doing right? The only differences are that this new API allows creating for a subset of allowed formats that's not necessarily a subset of the allowed formats, which might raise trigger additional errors. The benefits seem to be quite small in my opinion. Why not stick to `g.create_formats_()` for now and add a new API if this becomes really necessary?

keli-wen: Indeed, but I'm considering that under the condition of `create_formats = None`, the original behavior of `g.create_formats_()` will not be disrupted. 

For those who wish to modify in place, `create_formats` might be quite helpful. As for accidental errors, such as when some format(s) in `create_formats` are not in the graph's `allowed_formats`, I believe these types of exceptions are easy to catch and return with a clear error message.

python/dgl/heterograph.py
keli-wen: That's correct. I believe this aspect requires further consideration. While drafting the document, I also felt this part was overly complex. Perhaps a new API would make the entire design clearer. However, this needs more discussion. @frozenbugs

python/dgl/heterograph.py
mufeili: Yes. I just want to point out that when there is not a clear customer, it may not be worth the efforts to expand the current API.

python/dgl/heterograph.py
frozenbugs: It will be more confusion for user if we just say "a subset of the allowed formats will be created", how can I know which allowed format will be created.

What about: always create all allowed format, if create_formats == None?
if create_formats is a subset of allowed_formats, then create the create_formats
Otherwise, raise an error.

python/dgl/heterograph.py
frozenbugs: Oh, sorry for the confusion, this PR is just for code review on the API doc string. Eventually, the old formats and create_formats_ will be replace by these 2 apis.

We will never name a API newblabla, new will be old right after it get merged :)

python/dgl/heterograph.py
mufeili: It actually doesn't matter. Many ops create allowed sparse formats on demand. None of these ops mention the creation of particular sparse formats in their docs and most users never complain about them. If a sparse format is allowed, then it may be created. That's it. Easy.

python/dgl/heterograph.py
mufeili: I understand. The point is that you may not even need to modify that API at all.

mufeili:
@keli-wen Sounds good to me! I also recommend implementing them in separate PRs.

python/dgl/heterograph.py
```diff
@@ -6065,6 +6065,88 @@ def local_scope(self):
...
+
+            * If ``allowed_formats`` is None, the result will be a dict
+              recording the usage status of sparse formats.
+            * Otherwise, a DGLGraph will be returned, which is a clone of the
+              original graph with the specified allowed sparse format(s) and
+              created sparse format(s).
+        """
+        pass
+
+    def new_create_formats_(self, create_formats=None):
```
Rhett-Ying, line 6129: these APIs are the main ones user use to create format or formatting? If yes, we'd better add benchmark tests for these APIs. just refer to existing one: https://github.com/dmlc/dgl/blob/master/benchmarks/benchmarks/api/bench_format_conversion.py

keli-wen:
Finally, this problem is solved in the following two PR.

#5851
#5859

PR #5771, opened 2023-06-02
[GraphBolt] Add sample etype neighbors

peizhou001:
## Description
Enable sample neighbors per etype, depend on #5770.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [ ] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [ ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

PR #5774, opened 2023-06-03
[GraphBolt] Add probs for neighbor sampling

peizhou001:
## Description
Enable sampling with probs, based on #5771 

## Checklist
- [ ] Code is well-documented


graphbolt/src/csc_sampling_graph.cc
```diff
@@ -122,9 +122,12 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
 
 c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighbors(
-    const torch::Tensor& nodes, int64_t fanout, bool replace) const {
+    const torch::Tensor& nodes, const std::vector<int64_t>& fanouts,
+    bool replace, const torch::optional<torch::Tensor>& probs) const {
   const int64_t num_nodes = nodes.size(0);
-
+  // If true, perform sampling for each edge type of each node, otherwise just
+  // sample once for each node with no regard of edge types.
+  bool consider_etype = (fanouts.size() > 1);
```
jermainewang, line 130: const

jermainewang: Do we check `fanouts.size() == num_etypes` for this case?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -198,21 +208,60 @@ c10::intrusive_ptr<CSCSamplingGraph> CSCSamplingGraph::LoadFromSharedMemory(
 
 torch::Tensor Pick(
     int64_t offset, int64_t num_neighbors, int64_t fanout, bool replace,
-    const torch::TensorOptions& options) {
+    const torch::TensorOptions& options,
+    const torch::optional<torch::Tensor>& probs) {
```
jermainewang, line 212: I see we are dropping the original design that passes in a pick UDF. I wonder if that design will make the code cleaner? In fact, we could treat the logic to handle with or without replacement, uniform or non-uniform sampling or parts of the pick UDF logic. For example,

```c++
c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighbors(
    const torch::Tensor& nodes, const std::vector<int64_t>& fanouts,
    PickUDF pick_udf) const {
  // The pick UDF will handle the detailed logics to sample on a segment while
  // this function handles the outer parallelism.
  ...

  // the big loop
  torch::parallel_for(0, num_nodes, 32, [&](size_t b, size_t e) {
    for (size_t i = b; i < e; ++i) {
      const auto nid = nodes[i].item<int64_t>();
      TORCH_CHECK(
          nid >= 0 && nid < NumNodes(),
          "The seed nodes' IDs should fall within the range of the graph's "
          "node IDs.");
      const auto offset = indptr_[nid].item<int64_t>();
      const auto num_neighbors = indptr_[nid + 1].item<int64_t>() - offset;

      if (num_neighbors == 0) {
        // Initialization is performed here because all tensors will be
        // concatenated in the master thread, and having an undefined tensor
        // during concatenation can result in a crash.
        picked_neighbors_per_node[i] = torch::tensor({}, indptr_.options());
        continue;
      }

      if (consider_etype) {
        picked_neighbors_per_node[i] = PickEtype(
            offset, num_neighbors, fanouts, indptr_.options(),
            type_per_edge_.value(), pick_udf);  // just pass in the UDF.
      } else {
        picked_neighbors_per_node[i] = Pick(
            offset, num_neighbors, fanouts[0], indptr_.options(),
            pick_udf);  // Just pass-in the UDF
      }
      num_picked_neighbors_per_node[i + 1] =
          picked_neighbors_per_node[i].size(0);
    }
  });  // End of the thread.
  ...
}
```

And then we can just invoke the function with different UDFs for different scenarios (instead of having mingled logics). For example,

```c++
// For uniform sampling
SampleNeighbors(nodes, fanouts, UniformPickUDF);
// For non-uniform sampling
SampleNeighbors(nodes, fanouts, NonUniformPickUDF);
```

And the each UDF will handle the details such as whether it falls back to picking all neighbors, etc. For example,

```c++
void UniformPickUDF(...) {
  // You can bind any variables to it that the `SampleNeighbors` function does not need to know.

  if ((fanout == -1) || (num_neighbors <= fanout && !replace) {
    do_the_fast_path();
  } else {
    do_the_normal_work();
  }
}

void NonUniformPickUDF(...) {
  // You can bind any variables to it that the `SampleNeighbors` function does not need to know.
  auto positive_prop = (probs.value().slice(0, offset, offset + num_neighbors) > 0);
  if ((fanout == -1) || (num_neighbors <= positive_prop.size() && !replace) {
    // Note you could have different fast path condition than uniform sampling.
    do_the_fast_path();
  } else {
    do_the_normal_work();
  }
}
```

jermainewang: The central idea is that if we believe the function `SampleNeighbors` only handles the parallelism of outer-loop as well as locating the segment belonging to the same type, then arguments like `replace`, `prob`, `fanouts` are actually all irrelevant.

jermainewang: If you think this requires too much a change, one way to quickly makes it read better is to restructure this `Pick` function so that uniform and non-uniform sampling logics are in two blocks. E.g.,

```c++
torch::Tensor Pick(
    const torch::optional<torch::Tensor>& probs) {
  if (probs.has_value()) {
    return NonUniformPick(offset, num_neighbors, fanout, replace, options, probs);
  } else {
    return UniformPick(offset, num_neighbors, fanout, replace, options, probs);
  }
}
```

peizhou001: Make sense, the design of using a UDF make the code and logic more clear, as well as help clarify the responsibility of the 'sample_neighbors' code block, It is a good choice. But as the change is too much, and as the principle of 'PR should be short and only do one thing', we can first adopt the 'quick way' to add the probs logic and make the refactor in the future.

graphbolt/src/csc_sampling_graph.cc
peizhou001: Add this logic.

graphbolt/include/graphbolt/csc_sampling_graph.h
```diff
@@ -217,6 +222,60 @@ class CSCSamplingGraph : public torch::CustomClassHolder {
...
+ * Otherwise, each value can be selected only once.
+ * @param options Tensor options specifying the desired data type of the result.
+ * @param probs Optional tensor containing the (unnormalized) probabilities
+ * associated with each neighboring edge of a node in the original graph. It
+ * must be a 1D floating-point tensor with the number of elements equal to the
+ * number of edges in the graph.
+ *
+ * @return A tensor containing the picked neighbors.
+ */
+torch::Tensor NonUniformPick(
```
frozenbugs, line 274: It probably worth to add an comment about how the probs is used in the NonUniformPick. The implementation is non-trivial.

python/dgl/graphbolt/graph_storage/csc_sampling_graph.py
```diff
@@ -273,8 +279,27 @@ def sample_neighbors(
             assert len(self.metadata.edge_type_to_id) == fanouts.size(
                 0
             ), "Fanouts should have the same number of elements as etypes."
+        if probs is not None:
+            assert probs.dim() == 1, "Probs should be 1-D tensor."
+            assert (
+                probs.size(0) == self.num_edges
+            ), "Probs should have the \
+                same number of elements as the number of edges."
```
frozenbugs, line 287: nit
```suggestion
            ), "Probs should have the same number of elements as the number of \
                 edges."
```

peizhou001: Fixed

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -221,17 +223,51 @@ torch::Tensor Pick(
...
 }
 
+torch::Tensor NonUniformPick(
+    int64_t offset, int64_t num_neighbors, int64_t fanout, bool replace,
+    const torch::TensorOptions& options,
+    const torch::optional<torch::Tensor>& probs) {
+  torch::Tensor picked_neighbors;
+  auto local_probs = probs.value().slice(0, offset, offset + num_neighbors);
+  auto positive_probs_mask = (local_probs > 0);
+  if ((fanout == -1) || (num_neighbors <= fanout && !replace)) {
```
jermainewang, line 240: Is this condition correct? Should it be `num_neighbors <= positive_probs_mask.size()`?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -221,17 +223,51 @@ torch::Tensor Pick(
...
+    picked_neighbors = torch::arange(offset, offset + num_neighbors, options);
+    picked_neighbors =
+        torch::masked_select(picked_neighbors, positive_probs_mask);
+  } else {
+    auto num_positive_probs = positive_probs_mask.size(0);
+    if (num_positive_probs == 0) return torch::tensor({}, options);
```
jermainewang, line 246: Move this before the short-cut condition. It should be the very first condition.

graphbolt/include/graphbolt/csc_sampling_graph.h
peizhou001: added

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -221,17 +223,51 @@ torch::Tensor Pick(
...
-      picked_neighbors = torch::randperm(num_neighbors, options) + offset;
-      picked_neighbors = picked_neighbors.slice(0, 0, fanout);
+      picked_neighbors = torch::randperm(num_neighbors, options);
+      picked_neighbors = picked_neighbors.slice(0, 0, fanout) + offset;
     }
   }
   return picked_neighbors;
 }
 
+torch::Tensor NonUniformPick(
```
jermainewang, line 233: Mark it as inline

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -210,7 +212,7 @@ c10::intrusive_ptr<CSCSamplingGraph> CSCSamplingGraph::LoadFromSharedMemory(
   return BuildGraphFromSharedMemoryTensors(std::move(shared_memory_tensors));
 }
 
-torch::Tensor Pick(
+torch::Tensor UniformPick(
```
jermainewang, line 215: mark it as inline

graphbolt/include/graphbolt/csc_sampling_graph.h
jermainewang: BTW, do we need this function declaration in the header? Can't we just make it internal?

graphbolt/src/csc_sampling_graph.cc
peizhou001: right, let me fix

graphbolt/src/csc_sampling_graph.cc
peizhou001: Fixed

graphbolt/src/csc_sampling_graph.cc
peizhou001: Fixed

graphbolt/src/csc_sampling_graph.cc
peizhou001: Fixed

python/dgl/graphbolt/graph_storage/csc_sampling_graph.py
```diff
@@ -273,8 +279,27 @@ def sample_neighbors(
...
+                torch.bfloat16,
+                torch.float32,
+                torch.float64,
+            ], "Probs should have a floating-point or boolean data type."
+            # Note probs will be passed as input for 'torch.multinomial'
+            # in deeper stack, which doesn't support 'torch.half' and
+            # torch.bool' data types.To avoid crashes, convert 'probs' to
+            # 'bfloat16' data type.
+            if probs.dtype in [torch.bool, torch.float16]:
+                probs = probs.to(torch.bfloat16)
```
jermainewang, line 300: Put this to c++ because it is tied to your underlying implementation instead of the python operator semantics.

jermainewang (approved):

python/dgl/graphbolt/graph_storage/csc_sampling_graph.py
peizhou001: Fixed

graphbolt/include/graphbolt/csc_sampling_graph.h
peizhou001: Remove .h declare for fix

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -221,17 +231,52 @@ torch::Tensor Pick(
...
-      picked_neighbors = torch::randperm(num_neighbors, options) + offset;
-      picked_neighbors = picked_neighbors.slice(0, 0, fanout);
+      picked_neighbors = torch::randperm(num_neighbors, options);
+      picked_neighbors = picked_neighbors.slice(0, 0, fanout) + offset;
     }
   }
   return picked_neighbors;
 }
 
+inline torch::Tensor NonUniformPick(
```
frozenbugs, line 241: Consider also copy the doc string to here as well? Since you already wrote pretty good doc string in the header file.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -210,7 +220,7 @@ c10::intrusive_ptr<CSCSamplingGraph> CSCSamplingGraph::LoadFromSharedMemory(
   return BuildGraphFromSharedMemoryTensors(std::move(shared_memory_tensors));
 }
 
-torch::Tensor Pick(
+inline torch::Tensor UniformPick(
```
frozenbugs, line 223: Consider also copy the doc string to here as well? Since you already wrote pretty good doc string in the header file.

graphbolt/src/csc_sampling_graph.cc
peizhou001: Sure, let me make a copy.

graphbolt/src/csc_sampling_graph.cc
peizhou001: Sure, let me make a copy.

PR #5803, opened 2023-06-07
[GraphBolt] support DictItemSet in MinibatchSampler

Rhett-Ying:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/minibatch_sampler.py
+++ b/python/dgl/graphbolt/minibatch_sampler.py
@@ -39,7 +39,7 @@ class MinibatchSampler(IterDataPipe):
 
     Examples
     --------
-    1. Node/edge IDs.
+    1. Node IDs.
     >>> import torch
     >>> from dgl import graphbolt as gb
     >>> item_set = gb.ItemSet(torch.arange(0, 10))
@@ -101,18 +101,17 @@ class MinibatchSampler(IterDataPipe):
     >>> list(data_pipe)
     [tensor([1, 2, 3, 4]), tensor([5, 6, 7, 8]), tensor([ 9, 10])]
 
-    7. Heterogeneous node/edge IDs.
+    7. Heterogeneous node IDs.
     >>> ids = {
-    ...     ("user", "like", "item"): gb.ItemSet(torch.arange(0, 5)),
-    ...     ("user", "follow", "user"): gb.ItemSet(torch.arange(5, 10)),
+    ...     "user": gb.ItemSet(torch.arange(0, 5)),
+    ...     "item": gb.ItemSet(torch.arange(0, 5)),
     ... }
     >>> item_set = gb.DictItemSet(ids)
     >>> minibatch_sampler = gb.MinibatchSampler(item_set, 4)
     >>> list(minibatch_sampler)
-    [{('user', 'like', 'item'): tensor([0, 1, 2, 3])},
-    {('user', 'follow', 'user'): tensor([5, 6, 7]),
-        ('user', 'like', 'item'): tensor([4])},
-    {('user', 'follow', 'user'): tensor([8, 9])}]
+    [{'user': tensor([0, 1, 2, 3])},
+    {'item': tensor([0, 1, 2]), 'user': tensor([4])},
+    {'item': tensor([3, 4])}]
 
     8. Heterogeneous node pairs.
     >>> node_pairs = (torch.arange(0, 5), torch.arange(5, 10))
--- a/python/dgl/graphbolt/minibatch_sampler.py
+++ b/python/dgl/graphbolt/minibatch_sampler.py
@@ -104,62 +104,68 @@ class MinibatchSampler(IterDataPipe):
     7. Heterogeneous node IDs.
     >>> ids = {
     ...     "user": gb.ItemSet(torch.arange(0, 5)),
-    ...     "item": gb.ItemSet(torch.arange(0, 5)),
+    ...     "item": gb.ItemSet(torch.arange(0, 6)),
     ... }
     >>> item_set = gb.DictItemSet(ids)
     >>> minibatch_sampler = gb.MinibatchSampler(item_set, 4)
     >>> list(minibatch_sampler)
     [{'user': tensor([0, 1, 2, 3])},
     {'item': tensor([0, 1, 2]), 'user': tensor([4])},
-    {'item': tensor([3, 4])}]
+    {'item': tensor([3, 4, 5])}]
 
     8. Heterogeneous node pairs.
-    >>> node_pairs = (torch.arange(0, 5), torch.arange(5, 10))
+    >>> node_pairs_like = (torch.arange(0, 5), torch.arange(0, 5))
+    >>> node_pairs_follow = (torch.arange(0, 6), torch.arange(6, 12))
     >>> item_set = gb.DictItemSet({
-    ...     ("user", "like", "item"): gb.ItemSet(node_pairs),
-    ...     ("user", "follow", "user"): gb.ItemSet(node_pairs),
+    ...     ("user", "like", "item"): gb.ItemSet(node_pairs_like),
+    ...     ("user", "follow", "user"): gb.ItemSet(node_pairs_follow),
     ... })
     >>> minibatch_sampler = gb.MinibatchSampler(item_set, 4)
     >>> list(minibatch_sampler)
-    [{('user', 'like', 'item'): [tensor([0, 1, 2, 3]), tensor([5, 6, 7, 8])]},
-    {('user', 'follow', 'user'): [tensor([0, 1, 2]), tensor([5, 6, 7])],
-        ('user', 'like', 'item'): [tensor([4]), tensor([9])]},
-    {('user', 'follow', 'user'): [tensor([3, 4]), tensor([8, 9])]}]
+    [{('user', 'like', 'item'): [tensor([0, 1, 2, 3]), tensor([0, 1, 2, 3])]},
+    {('user', 'like', 'item'): [tensor([4]), tensor([4])],
+     ('user', 'follow', 'user'): [tensor([0, 1, 2]), tensor([6, 7, 8])]},
+    {('user', 'follow', 'user'): [tensor([3, 4, 5]), tensor([ 9, 10, 11])]}]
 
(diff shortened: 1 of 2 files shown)


python/dgl/graphbolt/minibatch_sampler.py
```diff
@@ -108,11 +100,72 @@ class MinibatchSampler(IterDataPipe):
...
+    >>> ids = {
+    ...     ("user", "like", "item"): gb.ItemSet(torch.arange(0, 5)),
+    ...     ("user", "follow", "user"): gb.ItemSet(torch.arange(5, 10)),
+    ... }
+    >>> item_set = gb.DictItemSet(ids)
+    >>> minibatch_sampler = gb.MinibatchSampler(item_set, 4)
+    >>> list(minibatch_sampler)
+    [{('user', 'like', 'item'): tensor([0, 1, 2, 3])},
+    {('user', 'follow', 'user'): tensor([5, 6, 7]),
+        ('user', 'like', 'item'): tensor([4])},
```
frozenbugs, line 114: Shouldn't it be like this?
```suggestion
    {('user', 'like', 'item'): tensor([4])},
        ('user', 'follow', 'user'): tensor([5, 6, 7]),
```

python/dgl/graphbolt/minibatch_sampler.py
```diff
@@ -108,11 +100,72 @@ class MinibatchSampler(IterDataPipe):
     >>> data_pipe = data_pipe.map(add_one)
     >>> list(data_pipe)
     [tensor([1, 2, 3, 4]), tensor([5, 6, 7, 8]), tensor([ 9, 10])]
+
+    7. Heterogeneous node/edge IDs.
+    >>> ids = {
+    ...     ("user", "like", "item"): gb.ItemSet(torch.arange(0, 5)),
+    ...     ("user", "follow", "user"): gb.ItemSet(torch.arange(5, 10)),
```
frozenbugs, line 107: When we ask users to provide dataset for heterogeneous training.

For different node types, do we ask users to provide ids always from 0, or globally continuous?

python/dgl/graphbolt/minibatch_sampler.py
```diff
@@ -108,11 +100,72 @@ class MinibatchSampler(IterDataPipe):
     >>> data_pipe = data_pipe.map(add_one)
     >>> list(data_pipe)
     [tensor([1, 2, 3, 4]), tensor([5, 6, 7, 8]), tensor([ 9, 10])]
+
+    7. Heterogeneous node/edge IDs.
```
frozenbugs, line 104: Heterogeneous node IDs.

edge IDs is not recommended in our new design, let's use node ID in example.

python/dgl/graphbolt/minibatch_sampler.py
Rhett-Ying: I don't think we have any assumption or limitation on node ids. ids always could be arbitrary.

python/dgl/graphbolt/minibatch_sampler.py
Rhett-Ying: No. `('user', 'follow', 'user')` is iterated and placed to the mini-batch first, then `('user', 'like', 'item')`. But anyway, `dict` is not ordered.

python/dgl/graphbolt/minibatch_sampler.py
Rhett-Ying: what about homogeneous edge IDs? not recommended either?

frozenbugs: No, not recommended either.

python/dgl/graphbolt/minibatch_sampler.py
frozenbugs: The itemset should not have any assumption, but as example, it should show what's we recommend to users.

python/dgl/graphbolt/minibatch_sampler.py
Rhett-Ying: why eid is not recommended in graphbolt?

python/dgl/graphbolt/minibatch_sampler.py
```diff
@@ -108,11 +100,71 @@ class MinibatchSampler(IterDataPipe):
...
+    >>> ids = {
+    ...     "user": gb.ItemSet(torch.arange(0, 5)),
+    ...     "item": gb.ItemSet(torch.arange(0, 5)),
+    ... }
+    >>> item_set = gb.DictItemSet(ids)
+    >>> minibatch_sampler = gb.MinibatchSampler(item_set, 4)
+    >>> list(minibatch_sampler)
+    [{'user': tensor([0, 1, 2, 3])},
+    {'item': tensor([0, 1, 2]), 'user': tensor([4])},
+    {'item': tensor([3, 4])}]
```
frozenbugs, line 114:
```suggestion
    {'item': tensor([3, 4, 5])}]
```

python/dgl/graphbolt/minibatch_sampler.py
```diff
@@ -108,11 +100,71 @@ class MinibatchSampler(IterDataPipe):
...
+    {('user', 'follow', 'user'): [tensor([0, 1, 2]), tensor([5, 6, 7])],
+        ('user', 'like', 'item'): [tensor([4]), tensor([9])]},
+    {('user', 'follow', 'user'): [tensor([3, 4]), tensor([8, 9])]}]
+
+    9. Heterogeneous node pairs and labels.
+    >>> node_pairs_labels = (
+    ...     torch.arange(0, 5), torch.arange(5, 10), torch.arange(10, 15))
+    >>> item_set = gb.DictItemSet({
+    ...     ("user", "like", "item"): gb.ItemSet(node_pairs_labels),
+    ...     ("user", "follow", "user"): gb.ItemSet(node_pairs_labels),
```
frozenbugs, line 134: ditto

python/dgl/graphbolt/minibatch_sampler.py
```diff
@@ -108,11 +100,71 @@ class MinibatchSampler(IterDataPipe):
...
+    {('user', 'follow', 'user'):
+        [tensor([3, 4]), tensor([8, 9]), tensor([13, 14])]}]
+
+    10. Heterogeneous head, tail and negative tails.
+    >>> head_tail_neg_tails = (
+    ...     torch.arange(0, 5), torch.arange(5, 10),
+    ...     torch.arange(10, 20).reshape(-1, 2))
+    >>> item_set = gb.DictItemSet({
+    ...     ("user", "like", "item"): gb.ItemSet(head_tail_neg_tails),
+    ...     ("user", "follow", "user"): gb.ItemSet(head_tail_neg_tails),
```
frozenbugs, line 152: ditto.

frozenbugs (approved):

jermainewang:
Some questions about heterogeneous item set.

1. Since the input is a dictionary, how will we decide the order when `shuffle=False`? For example, given `{ 'user' : ..., 'item' : ...}`, which of the "user" or "item" ids will be iterated first?
2. In the above example, if all the "user" ids have been consumed, will the returned mini-batch dictionary still contain the key `'user'` or not? If yes, what is shape/dtype of the corresponding tensor?

Rhett-Ying:
> Some questions about heterogeneous item set.
> 
>     1. Since the input is a dictionary, how will we decide the order when `shuffle=False`? For example, given `{ 'user' : ..., 'item' : ...}`, which of the "user" or "item" ids will be iterated first?
> 
>     2. In the above example, if all the "user" ids have been consumed, will the returned mini-batch dictionary still contain the key `'user'` or not? If yes, what is shape/dtype of the corresponding tensor?

For 1, A number of items(this is controlled by buffer size) will be iterated and saved into a buffer, then pick one of them from the buffer when `shuffle=True` for each iteration. So in the case you mentioned, `user` will be iterated first if `len('user')` is greater than `buffer_size`(10000 in default). If not greater than `buffer_size`, then `item` may be picked up for each iteration. Here's the implementation which could help understand the shuffle logic: https://github.com/pytorch/pytorch/blob/6ac3352a37ead6bae0de9fe4c45b618f8e82dc78/torch/utils/data/datapipes/iter/combinatorics.py#L122-L135
So an enhancement is to dynamically specify the `buffer_size` to control the iteration order.

For 2, it only returns the ones picked up for the current mini-batch, so `No` if `user` has already been ran out.

@jermainewang

python/dgl/graphbolt/minibatch_sampler.py
```diff
@@ -108,11 +100,71 @@ class MinibatchSampler(IterDataPipe):
...
+    >>> list(minibatch_sampler)
+    [{'user': tensor([0, 1, 2, 3])},
+    {'item': tensor([0, 1, 2]), 'user': tensor([4])},
+    {'item': tensor([3, 4])}]
+
+    8. Heterogeneous node pairs.
+    >>> node_pairs = (torch.arange(0, 5), torch.arange(5, 10))
+    >>> item_set = gb.DictItemSet({
+    ...     ("user", "like", "item"): gb.ItemSet(node_pairs),
+    ...     ("user", "follow", "user"): gb.ItemSet(node_pairs),
```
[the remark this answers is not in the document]
Rhett-Ying: I changed to 2 different node_pairs

python/dgl/graphbolt/minibatch_sampler.py
```diff
@@ -1,27 +1,19 @@
...
+from functools import partial
+from typing import Optional
 
 from torch.utils.data import default_collate
 from torchdata.datapipes.iter import IterableWrapper, IterDataPipe
 
 from ..batch import batch as dgl_batch
 from ..heterograph import DGLGraph
-from .itemset import ItemSet
+from .itemset import DictItemSet, ItemSet
```
jermainewang, line 12: Small suggestion: perhaps rename `DictItemSet` to `ItemSetDict` because the former is not a subclass of `ItemSet`.

jermainewang (approved):

python/dgl/graphbolt/minibatch_sampler.py
Rhett-Ying: pls let me change it in another PR to be clear.

PR #5844, opened 2023-06-09
[GraphBolt] define __len__ for ItemSet/ItemSetDict

Rhett-Ying:
## Description

1. define `__len__` to enable `len()` if possible. This definition will not result in crash if call with `list(item_set)`. Referred from https://github.com/pytorch/pytorch/blob/6acb8d3d1cf577b63cd11bf88c627d3e86669901/torch/utils/data/datapipes/iter/combinatorics.py#L47-L51 What's more, enabling length could make it possible to do optimization: slice batch of items from original data directly instead of iterating one by one.
2. remove size check for tuple items as `.size()` is available in `torch.Tensor` only, not general enough as `Iterable` is required only for the input items.
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/itemset.py
+++ b/python/dgl/graphbolt/itemset.py
@@ -1,6 +1,6 @@
 """GraphBolt Itemset."""
 
-from typing import Dict, Iterable, Iterator, Tuple
+from typing import Dict, Iterable, Iterator, Sized, Tuple
 
 __all__ = ["ItemSet", "ItemSetDict"]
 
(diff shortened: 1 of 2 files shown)


python/dgl/graphbolt/itemset.py
```diff
@@ -49,9 +49,6 @@ class requires each input itemset to be iterable.
 
     def __init__(self, items: Iterable or Tuple[Iterable]) -> None:
         if isinstance(items, tuple):
-            assert all(
-                items[0].size(0) == item.size(0) for item in items
-            ), "Size mismatch between items in tuple."
```
frozenbugs, line 54: So we don't need to check the tuple of tensor have the same size anymore?

Rhett-Ying: No. remove size check for tuple items as .size() is available in torch.Tensor only, not general enough as Iterable is required only for the input items.

frozenbugs (approved):

PR #5845, opened 2023-06-09
[Graphbolt] Add FeatureStore.

### LINKED ISSUE #5823: [Graphbolt] Load and fetch from CPU memory.

czkkkkkk:
## Description
Resolve #5823 .
## Checklist
- [ ] Code is well-documented


czkkkkkk:
@frozenbugs , I made a draft following the code in our design doc. 

There is still one thing to decide: how to handle the heterogeneous information of features. In a graph, a feature array has a feature name and belongs to a specific type of nodes. I'd suggest to use `InMemoryFeatureStore` as a key-value store. When fetching/storing features, `FeatureFetcher` uses `(ntype, feat_name)` as the key to access features.

czkkkkkk:
@frozenbugs , this PR only includes FeatureStore now.

python/dgl/graphbolt/feature_store.py
```diff
@@ -0,0 +1,166 @@
...
+            of the feature are read. If None, the entire feature is returned.
+
+        Returns
+        -------
+        torch.Tensor
+            The read feature.
+        """
+        raise NotImplementedError
+
+    def add_feature(self, key: str, value: torch.Tensor):
```
frozenbugs, line 29: How can you add feature without ids? When do you need to add feature?

python/dgl/graphbolt/feature_store.py
```diff
@@ -0,0 +1,166 @@
...
+import torch
+
+
+class FeatureStore:
+    r"""Base class for feature store."""
+
+    def __init__(self):
+        pass
+
+    def read_feature(self, key: str, ids: torch.Tensor = None):
```
frozenbugs, line 11: redundant _feature in each method, just read / add / update or maybe just get / set? not sure why we need add.

czkkkkkk: I'm not sure whether we need to `add_feature` (which also adds a new key) in the future. If not, all the keys of the FeatureStorage are initialized in the constructor. Since it is not clear, I will remove this interface now.

Comparing `set/get` and `read/update`, I prefer the latter because if we use some ids to access a feature, we only read or update a part of the feature, instead of setting or getting the whole feature.

frozenbugs: What's the API name for tensor?

frozenbugs: Sounds good.

frozenbugs: read_feature -> read is enough, FeatureStore's read method implies read_feature.

python/dgl/graphbolt/feature_store.py
```diff
@@ -0,0 +1,151 @@
...
+        feature is identified by a unique key, and its value is specified using
+        a tensor.
+
+        Parameters
+        ----------
+        key : str
+            The key that uniquely identifies the feature in the feature store.
+        value : torch.Tensor
+            The updated value of the feature.
+        ids : torch.Tensor, optional
```
frozenbugs, line 44: the length of the id should match the length of value?

python/dgl/graphbolt/feature_store.py
```diff
@@ -0,0 +1,151 @@
...
+            The index of the feature. If specified, only the specified indices
+            of the feature will be updated. If None, the entire feature will be
```
frozenbugs, line 46: This comment is confusing. My understand is, the value will be n vectors, and ids will be n id, and row i of value will be update to featurestore[key][id[i]].

python/dgl/graphbolt/feature_store.py
czkkkkkk: Right!

python/dgl/graphbolt/feature_store.py
czkkkkkk: They should match. I will update the comment and add an assert.

python/dgl/graphbolt/feature_store.py
czkkkkkk: Which part is confusing? If `ids` is given, it works exactly as what you described.

tests/python/pytorch/graphbolt/test_feature_store.py
```diff
@@ -0,0 +1,34 @@
+import torch
+from dgl import graphbolt as gb
+
+
+def test_feature_store():
```
frozenbugs, line 5: test_in_memory_feature_store

frozenbugs:
Overall LGTM, need to clarify the behavior of update method.

tests/python/pytorch/graphbolt/test_feature_store.py
```diff
@@ -0,0 +1,34 @@
...
+    )
+    assert torch.equal(
+        feature_store.read_feature("a", torch.tensor([1, 1])),
+        torch.tensor([2, 2]),
+    )
+    assert torch.equal(
+        feature_store.read_feature("c", torch.tensor([1])),
+        torch.tensor([[4, 5, 6]]),
+    )
+    try:
```
Rhett-Ying, line 24: better practice of expect exception in pytest: `with pytest.raises(AssertionError)` ?

python/dgl/graphbolt/feature_store.py
czkkkkkk: I changed the description.

```
        ids : torch.Tensor, optional
            The indices of the feature to update. If specified, only the
            specified indices of the feature will be updated. For the feature,
            the `ids[i]` row is updated to `value[i]`. So the indices and value
            must have the same length. If None, the entire feature will be
            updated.
```

tests/python/pytorch/graphbolt/test_feature_store.py
czkkkkkk: Cool! I updated the code.

Issue #5807, opened 2023-06-08
[DistDGL] deprecate `tensorpipe` in RPC


Rhett-Ying:
## 🔨Work Item

**IMPORTANT:**
* This template is only for dev team to track project progress. For feature request or bug report, please use the corresponding issue templates.
* DO NOT create a new work item if the purpose is to fix an existing issue or feature request. We will directly use the issue in the project tracker.

Project tracker: https://github.com/orgs/dmlc/projects/2

## Description
As [tensorpipe](https://github.com/pytorch/tensorpipe) is not actively maintained and sometime DistDGL hangs such as https://github.com/dmlc/dgl/issues/4267, we'd like to deprecate and always use `socket` for RPC.

PR #5876, opened 2023-06-15
[DistDGL] remove tensorpipe from third_party and no build

Rhett-Ying:
## Description

part of #5807 
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,6 @@ dgl_option(USE_S3 "Build with S3 support" OFF)
 dgl_option(USE_HDFS "Build with HDFS support" OFF) # Set env HADOOP_HDFS_HOME if needed
 dgl_option(REBUILD_LIBXSMM "Clean LIBXSMM build cache at every build" OFF) # Set env HADOOP_HDFS_HOME if needed
 dgl_option(USE_EPOLL "Build with epoll for socket communicator" ON)
-dgl_option(TP_BUILD_LIBUV "Build libuv together with tensorpipe (only impacts Linux)" ON)
 dgl_option(BUILD_TORCH "Build the PyTorch plugin" OFF)
 dgl_option(BUILD_SPARSE "Build DGL sparse library" ON)
 dgl_option(BUILD_GRAPHBOLT "Build Graphbolt library" OFF)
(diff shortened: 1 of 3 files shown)


src/rpc/net_type.h
```diff
@@ -22,7 +22,7 @@ struct RPCBase {
   virtual void Finalize() = 0;
 
   /**
-   * @brief Communicator type: 'socket', 'tensorpipe', etc
+   * @brief Communicator type such as 'socket'.
```
frozenbugs, line 25: We only support socket, right? Can we remove it?

Rhett-Ying: yes but I choose to keep it as I'd like to keep this API for potential extention. From functional perspective, the whole class(not only this API) could be removed as only `socket` is supported for now. If we decide to remove such logic thoroughly, I prefer to do it in a separate PR.

frozenbugs: Yes, let's remove them all since we only have one option. If we need future extension we may redesign it.

frozenbugs (approved):

CMakeLists.txt
```diff
@@ -223,33 +220,11 @@ if((NOT MSVC) AND USE_LIBXSMM)
...
-  string(REPLACE "${WNOERROR_FLAGS}" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
-
-  # Avoid exposing third-party symbols when using DGL as a library.
-  if (NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wl,--exclude-libs,ALL")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wl,--exclude-libs,ALL")
-  endif()
-endif(NOT MSVC)
+# Avoid exposing third-party symbols when using DGL as a library.
+if((NOT MSVC) AND (NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin"))
```
frozenbugs, line 224: If NOT (A or B)?

Rhett-Ying: I think current one is a bit readable than `if(NOT (MSVC OR (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")))`.

czkkkkkk (approved):

PR #5882, opened 2023-06-17
[Performance][CUDA] Sorting for indices for UVM code path.

mfbalin:
Regular accesses over the PCI-e are faster than nonregular ones. Since the bottleneck is the PCI-e bandwidth, a sort overhead on the GPU is very negligible but the returns can be up to 4x faster when fetching mag240M edge types with UVA (16ms -> 4ms). Normal feature fetches are also faster overall for mag240M training by around 30%.


mfbalin:
@yaox12, could you take a look? The patch is minimal so should be fairly easy to review.

yaox12:
@dgl-bot

mfbalin:
Failed tests look related to ogb dataset loaders.

yaox12:
@dgl-bot

yaox12 (approved):
LGTM.

yaox12:
@dgl-bot

src/array/cuda/uvm/array_index_select_uvm.cuh
```diff
@@ -34,9 +35,10 @@ __global__ void IndexSelectMultiKernelAligned(
         ((uint64_t)(&array[in_row * num_feat]) % CACHE_LINE_SIZE) /
         sizeof(DType);
     col = col - idx_offset;
+    const auto out_row_ = perm ? perm[out_row] : out_row;
```
nv-dlasalle, line 38: Can we differentiate more between `out_row` and `out_row_`? Maybe `out_row_index` and `out_row` respectively? In many places in dgl a variable suffixed with `_` denotes a member variable.

nv-dlasalle (approved):
Looks good to me--one minor comment.

frozenbugs:
Is it possible to add some unit test for this change and related code?

mfbalin:
If there is already a test for correctness checking of UVA indexing, they should do the job. This patch is to improve the performance and the behavior should be unchanged.

frozenbugs:
> If there is already a test for correctness checking of UVA indexing, they should do the job. This patch is to improve the performance and the behavior should be unchanged. I was getting inconsistent performance measurements before this change, sometimes less data transferred was faster than more data transferred, etc.

I see, thanks for the clarification, if this is a performance critical pass, maybe worth to work with @Rhett-Ying to add benchmark test, feel free to make the decision based on your judgement.

Rhett-Ying:
@dgl-bot

Rhett-Ying:
@mfbalin Hi, how do you measure the performance improvement of this PR? existing [DGL benchmarks ](https://github.com/dmlc/dgl/tree/master/benchmarks/benchmarks)or self-customized ones?

mfbalin:
> @mfbalin Hi, how do you measure the performance improvement of this PR? existing [DGL benchmarks ](https://github.com/dmlc/dgl/tree/master/benchmarks/benchmarks)or self-customized ones?

@nv-dlasalle and I are finalizing a paper. The runtime measurements were made with my own training scripts.

Rhett-Ying:
@mfbalin Does the below code snippet measure the target performance improvement?
```
import torch
import dgl

feat_data = torch.rand(1000, 10).pin_memory()
idx = torch.randint(0, 1000, (1000,)).to('cuda')

# t0
sliced_data = dgl.utils.gather_pinned_tensor_rows(feat_data, idx)
# t1
```

mfbalin:
> @mfbalin Does the below code snippet measure the target performance improvement?
> 
> ```
> import torch
> import dgl
> 
> feat_data = torch.rand(1000, 10).pin_memory()
> idx = torch.randint(0, 1000, (1000,)).to('cuda')
> 
> # t0
> sliced_data = dgl.utils.gather_pinned_tensor_rows(feat_data, idx)
> # t1
> ```

You might want to try larger sizes though. Like if the scenario was ogbn-products, then 2M for feat_data and idx would be of size 500k.

nv-dlasalle:
> > @mfbalin Does the below code snippet measure the target performance improvement?
> > ```
> > import torch
> > import dgl
> > 
> > feat_data = torch.rand(1000, 10).pin_memory()
> > idx = torch.randint(0, 1000, (1000,)).to('cuda')
> > 
> > # t0
> > sliced_data = dgl.utils.gather_pinned_tensor_rows(feat_data, idx)
> > # t1
> > ```
> 
> You might want to try larger sizes though. Like if the scenario was ogbn-products, then 2M for feat_data and idx would be of size 200k.
Is ogbn-products big enough to observe a performance improvement on? Or do we need to target even larger so that there are significant TLB misses?

mfbalin:
When I was fetching mag240M edge types, it made an enormous difference (4x). The exact scenario was that the array length is 3.4e9, the dtype is uint8, and I was fetching around 600k of them at a time.

nvcastet:
Thanks a lot for creating the PR @mfbalin. I can confirm that I experienced the same benefits as you did by sorting the indices before doing CPU->GPU UVA transfers (GPU TLB misses are considerably reduced).

yaox12:
I think we can also merge this PR and add the benchmark later.

PR #5884, opened 2023-06-17
[Bugfix] Fix 5873

### LINKED ISSUE #5873: multi_update_all gives infinity on isolated nodes when using min reducer with different output keys
## 🐛 Bug

The title is quite long so I will directly demonstrate the bug below.

## To Reproduce

```python
import dgl
import torch
g = dgl.heterograph({
    ('A', 'AB', 'B'): ([0,1,2,3],[0,0,1,1]),
    ('C', 'CB', 'B'): ([0,1,2,3],[2,2,3,3])
})
g.nodes['A'].data['x'] = torch.randn(4, 16)
g.nodes['C'].data['x'] = torch.randn(4, 16)
g.multi_update_all({
    'AB': (dgl.function.copy_u('x', 'm'), dgl.function.min('m', 'a1')),
    'CB': (dgl.function.copy_u('x', 'm'), dgl.function.min('m', 'a2'))
    }, cross_reducer='min')
# One of the following gives infinity while the other does not.
print(g.nodes['B'].data['a1'])
print(g.nodes['B'].data['a2'])
```

## Expected behavior

Neither of the outputs should give infinity.

## Environment

 - OS (e.g., Linux):
 - Python version:

## Additional context

The reason is that in https://github.com/dmlc/dgl/blob/3a0bcc0c15cd0ce445318382887cd977979514df/python/dgl/heterograph.py#L5261-L5271
we are only replacing infinities for the first output key, assuming that the reducers all produce the same output key.  In practice we should use a for loop.

Credit to @zhhhhahahaha for finding this bug.

BarclayII:
Fixes #5873


mufeili:
Perhaps it's better to also update the docstring.

tests/python/common/test_heterograph-update-all.py
```diff
@@ -334,6 +334,37 @@ def _print_error(a, b):
                 _test(lhs, rhs, binary_op, reducer)
 
 
+def test_issue_5873():
```
frozenbugs, line 337: bad name, be specific about what you are testing, e.g. test multiple blablabla.

frozenbugs (approved):

BarclayII:
@dgl-bot

PR #5885, opened 2023-06-18
[Performance][CUDA] Labor UVA optimization

mfbalin:
## Description
Labor Sampler can perform UVA sampling. However, since it has to access the indptr and indices arrays (especially indices) very often, it slows down by a lot compared to NeighborSampler slowdown. This PR focuses on optimizing the accesses to indptr and indices arrays in a way that doesn't bring any regressions to the pure GPU scenario whereas improving training throughput with fanout 10,10,10 by more than 10 percent on ogbn-products training with `use_uva=True`. With this change, `ogbn-products` UVA sampling runtime for labor went from 18ms to 13ms on an A100 machine during multi-GPU training. On other multi-GPU machines with V100s, the sampling time is more than 2x faster.

Also fixed a bug for the weighted case. Added some tests to this PR so that we can detect if things are working properly. Below are the profiles of the old and the new code for LABOR-0 UVA. The new version has the extra OneHopExtractorAlignedKernel that copies the indices array in an aligned manner if it is pinned prior to the main sampling kernel.

old
new

## Checklist
- [ ] Code is well-documented


mfbalin:
@BarclayII could you take a look? On machines with multiple GPUs where PCI-e bandwidth is highly contested, this PR should improve things by quite a lot.

src/array/cpu/labor_pick.h
```diff
@@ -124,7 +123,9 @@ auto compute_importance_sampling_probabilities(
         var_1 = 0;
         if (weighted) {
           for (auto j = indptr[rid]; j < indptr[rid + 1]; j++)
-            var_1 += A[j] * A[j] / std::min(ONE, c * ps[j - indptr[rid]]);
+            var_1 += A[j] > 0
```
mfbalin, line 126: this check is needed to handle 0 probabilities properly.

src/array/cuda/labor_sampling.cu
```diff
@@ -247,21 +294,19 @@ __global__ void _CSRRowWiseLayerSampleDegreeKernel(
...
 
-    const IdType degree = indptr[row + 1] - in_row_start;
+    const IdType degree = subindptr[out_row + 1] - out_row_start;
 
     if (degree > 0) {
       // stands for k in in arXiv:2210.13339, i.e. fanout
       const auto k = min(num_picks, degree);
       // slightly better than NS
-      const FloatType d_ = ds ? ds[row] : degree;
+      const FloatType d_ = ds ? ds[out_row] : degree;
```
mfbalin, line 306: ds and d2s has shape num_rows so this was bugged.

mfbalin:
@frozenbugs, @yaox12, @nv-dlasalle  I am adding some overdue tests after my initial contribution and also a specialized code path when we detect the sampling is performed over UVA.

yaox12:
@dgl-bot

yaox12:
@dgl-bot

mfbalin:
Is there a problem with the CI again? I don't see how the distributed tests are related to these changes.

yaox12:
@dgl-bot

tests/python/common/sampling/test_sampling.py
```diff
@@ -681,6 +681,117 @@ def _test3(p, replace):
         assert subg["flips"].num_edges() == 4
 
 
+def _test_sample_labors(hypersparse, prob):
+    g, hg = _gen_neighbor_sampling_test_graph(hypersparse, False)
+
+    def _test1(p):
```
frozenbugs, line 687: Please be specific on the naming about what case is tested here.

tests/python/common/sampling/test_sampling.py
```diff
@@ -681,6 +681,117 @@ def _test3(p, replace):
         assert subg["flips"].num_edges() == 4
 
 
+def _test_sample_labors(hypersparse, prob):
```
frozenbugs, line 684: Is labor sampling GPU only or for both GPU and CPU?

mfbalin: It supports CPU sampling, GPU UVA sampling, and pure UVA sampling, basically, everything NeighborSampler does. This PR mostly improves the performance of the GPU UVA sampling scenario.

mfbalin: My intention was to make the change from NeighborSampler to LaborSampler as seamless as possible, they even have the same hyperparameters.

tests/python/common/sampling/test_sampling.py
mfbalin: I copied the test from the corresponding ones for NeighborSampling. I am not exactly sure about the exact testing scenario. The behaviors of the two samplers are pretty similar so I could do that. It is actually a bit hard to test LaborSampler because given a fanout of 2, it returns 2 neighbors per vertex in expectation. So it is hard to test for this expectation.

src/array/cuda/labor_sampling.cu
frozenbugs: Thanks for catching this, can you help us why this was not caught by the existing unit test?

mfbalin: In my initial PR, there were no unit tests, and I wasn't reminded about them. I tested the implementation with example training scripts myself. But it turns out there are some scenarios with weights and different edge types with minor bugs. So, I wanted to fix them in this PR and do a sanity check so that at least the different scenarios work properly. This doesn't test for full correctness as that would require statistical testing on the degrees of the sampled subgraphs.

mfbalin: I could potentially contribute the statistical testing part in another PR if you think that is okay.

tests/python/common/sampling/test_sampling.py
frozenbugs: I would have some concerns if the author can't provide well defined unit test cases, the algorithm is pretty complicated already :)
If we really want to make LaborSampler an "awesome version" of NeighborSampler and eventually replacing NeighborSampling in most situations, we should make sure we have 100% confident on the functional correctness and performance measurement by have concrete unit tests and regression tests.

frozenbugs: I do agree that the existing test is not perfect already, we should make it better.

mfbalin: If there was even the slightest demand for Labor Sampler, I would have the motivation to make it %100 robust production quality. As it is, it is almost there as I believe the performance of the code is already optimized and most of the functionality will be tested for with this PR. The statistical testing for importance_sampling=0 scenario is easily doable properly but it is a bit challenging to test for importance_sampling!=0. I believe that some end to end tests where we swap NeighborSampling with LaborSampling with different hyperparameters would also be a good addition to check for accuracy of convergence etc.

mfbalin: image
These are my results from the paper by just replacing NeighborSampler with LaborSampler without any other change.

mfbalin: You can count on me to assist with any issues related to this code. If anybody opens an issue, I will make sure that it is properly resolved or clarifications will be made.

mfbalin: I will go ahead and look at the testing code to see what is going on. I will be adding comments about what is being tested for in different parts soon.

Rhett-Ying:
@frozenbugs `dgl.sampling.sample_labors()` is not covered in existing DGL benchmarks, we need to add it in order to do performance comparison

Rhett-Ying:
@dgl-bot

Rhett-Ying:
@dgl-bot

src/array/cuda/labor_sampling.cu
frozenbugs: statistical testing is pretty hard, might worth to do it offline, and link the result in PR, but building a CI for it is not easy.

mfbalin: It is actually not that hard. If there are `n` seed nodes in `S`, then the expected number of edges sampled will be `\sum_{s \in S} \min(k, d_s) ~ nk`. Then, the actual number of sampled edges will deviate by around the square root of the expected number. So, we might not need something complicated to test `importance_sampling=0`. If there are a lot of seed nodes, that is already enough to do the statistical test.

frozenbugs (approved):

frozenbugs:
Let's get this PR merged first, and let's discuss next week to find a path to integrate it in graphbolt. @mfbalin

mfbalin:
I would also like to make a PR including the kappa feature I talked about in my presentation. It modifies only the random number generation logic and adds the kappa parameter to the sampler API. However, it won't be useful if #4341 is not merged.

Rhett-Ying:
@caojy1998 FYI. We need to add benchmark for `sample_labor()` into our new benchmark framework.

caojy1998:
> @caojy1998 FYI. We need to add benchmark for `sample_labor()` into our new benchmark framework.

OK, got it.

mfbalin:
Is the lack of a benchmark script blocking the merge of this PR or are we waiting for another reviewer's approval?

frozenbugs:
no, not blocking, let's run the CI, and I will merge it.

PR #5887, opened 2023-06-19
[TESTS] Make pytorch test_nn.py fully reproducible

anko-intel:
## Description
To make a comparison "before/after" doable,  there is a need to have the same internal results (matrices) -> the same seed for each run should be used for all randomized functions.


anko-intel:
@dgl-bot run

frozenbugs (approved):

PR #5914, opened 2023-06-28
[Graphbolt] Support on-disk feature store.

### LINKED ISSUE #5811: On Disk Node & Edge feature & file loader

czkkkkkk:
## Description

Resolve #5811.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/feature_store.py
+++ b/python/dgl/graphbolt/feature_store.py
@@ -49,17 +49,19 @@ class FeatureStore:
         raise NotImplementedError
 
 
-class InMemoryFeatureStore(FeatureStore):
-    r"""In-memory key-value feature store, where the key is a string and value
-    is Pytorch tensor."""
+class TorchBasedFeatureStore(FeatureStore):
+    r"""Torch based key-value feature store, where the key are strings and
+    values are Pytorch tensors."""
 
     def __init__(self, feature_dict: dict):
-        """Initialize an in-memory feature store.
+        """Initialize a torch based feature store.
 
         The feature store is initialized with a dictionary of tensors, where the
         key is the name of a feature and the value is the tensor. The value can
         be multi-dimensional, where the first dimension is the index of the
         feature.
+
+        Note that the values can be in memory or on disk.
 
         Parameters
         ----------
@@ -74,7 +76,7 @@ class InMemoryFeatureStore(FeatureStore):
         ...     "item": torch.arange(0, 6),
         ...     "rel": torch.arange(0, 6).view(2, 3),
         ... }
-        >>> feature_store = InMemoryFeatureStore(feature_dict)
+        >>> feature_store = TorchBasedFeatureStore(feature_dict)
         >>> feature_store.read("user", torch.tensor([0, 1, 2]))
         tensor([0, 1, 2])
         >>> feature_store.read("item", torch.tensor([0, 1, 2]))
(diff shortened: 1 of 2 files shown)


tests/python/pytorch/graphbolt/test_feature_store.py
```diff
@@ -1,13 +1,28 @@
+import numpy as np
 import pytest
 import torch
 from dgl import graphbolt as gb
 
 
-def test_in_memory_feature_store():
+def to_on_disk_tensor(name, t):
+    path = f"/tmp/{name}.npy"
```
Rhett-Ying, line 8: please use `with tempfile.TemporaryDirectory() as test_dir` for temp dir/file

python/dgl/graphbolt/feature_store.py
```diff
@@ -85,18 +87,35 @@ def __init__(self, feature_dict: dict):
...
             f"but got {type(feature_dict)}."
         )
         for k, v in feature_dict.items():
             assert isinstance(
                 k, str
-            ), f"Key in InMemoryFeatureStore must be str, but got {k}."
+            ), f"Key in TorchBasedFeatureStore must be str, but got {k}."
             assert isinstance(v, torch.Tensor), (
-                f"Value in InMemoryFeatureStore must be torch.Tensor,"
+                f"Value in TorchBasedFeatureStore must be torch.Tensor,"
```
Rhett-Ying, line 118: if data in numpy.Array, or other array-like format, will you create other specialized class such as `NumpyBasedxxx` to handle it? I am wondering if we need to convert implicitly?

tests/python/pytorch/graphbolt/test_feature_store.py
```diff
@@ -1,13 +1,28 @@
...
+    t = t.numpy()
+    np.save(path, t)
+    t = torch.as_tensor(np.load(path, mmap_mode="r+"))
```
Rhett-Ying, line 11: no RAM is consumed at all until `t` is accessed by indices?

tests/python/pytorch/graphbolt/test_feature_store.py
```diff
@@ -1,13 +1,28 @@
...
 import pytest
 import torch
 from dgl import graphbolt as gb
 
 
-def test_in_memory_feature_store():
+def to_on_disk_tensor(name, t):
+    path = f"/tmp/{name}.npy"
+    t = t.numpy()
+    np.save(path, t)
```
Rhett-Ying, line 10: save as numpy instead of `torch.save()` looks a bit not intuitive here. the reason using numpy here is only numpy supports reference to disk tensor?

tests/python/pytorch/graphbolt/test_feature_store.py
```diff
@@ -1,13 +1,28 @@
+import numpy as np
 import pytest
 import torch
 from dgl import graphbolt as gb
 
 
-def test_in_memory_feature_store():
+def to_on_disk_tensor(name, t):
```
frozenbugs, line 7: Other than unittest, we should also test the performance and add E2E tests benchmark for canonical solutions in the regression framework #5916.

python/dgl/graphbolt/feature_store.py
```diff
@@ -85,18 +87,35 @@ def __init__(self, feature_dict: dict):
...
         >>> feature_store.read("user", torch.tensor([0, 1, 2]))
         tensor([1, 1, 1])
+
+        >>> import numpy as np
+        >>> user = np.arange(0, 5)
+        >>> item = np.arange(0, 6)
+        >>> np.save("/tmp/user.npy", user)
+        >>. np.save("/tmp/item.npy", item)
+        >>> feature_dict = {
+        ...     "user": torch.as_tensor(np.load("/tmp/user.npy",
```
frozenbugs, line 97: How does the np.load work, does it load all file to memory or it does some smart dispatching?

python/dgl/graphbolt/feature_store.py
czkkkkkk: I think we only support torch based feature store. If users have numpy arrays, they need to convert them to torch tensors.

tests/python/pytorch/graphbolt/test_feature_store.py
czkkkkkk: Yes. And only the returned tensor of `read` consumes memory.

tests/python/pytorch/graphbolt/test_feature_store.py
czkkkkkk: Yes. I cannot find a interface of Pytorch that supports on-disk memory. And it is a common practice to use numpy memory mapped array as the backend of Pytorch on-disk tensor.

python/dgl/graphbolt/feature_store.py
czkkkkkk: If we specify the mmap_mode, it keeps the file on disk and reads the data when necessary. See the mode description in `memmap` (https://numpy.org/doc/stable/reference/generated/numpy.memmap.html#numpy.memmap).

python/dgl/graphbolt/feature_store.py
Rhett-Ying: then why we create a dedicated `TorchBasedxxx` instead of `FeatureStorage` directly? will we add more derived and different kinds of feature storage implementations later?

tests/python/pytorch/graphbolt/test_feature_store.py
Rhett-Ying: then could we add some comments here to help people understand it?

tests/python/pytorch/graphbolt/test_feature_store.py
czkkkkkk: Is there any instructions to write the benchmark?

python/dgl/graphbolt/feature_store.py
frozenbugs: I see, thanks. Then we just need to make sure we understand the performance since this will be our recommended canonical solution.

tests/python/pytorch/graphbolt/test_feature_store.py
```diff
@@ -1,13 +1,32 @@
...
+
+import numpy as np
 import pytest
 import torch
 from dgl import graphbolt as gb
 
 
-def test_in_memory_feature_store():
+def to_on_disk_tensor(name, t):
+    with tempfile.TemporaryDirectory() as test_dir:
```
Rhett-Ying, line 11: I'm not sure if it works as file will be deleted out of `with` but returned value `t` is still referencing to disk tensor? file deletion will fail?

frozenbugs:
LGTM, @Rhett-Ying please go head with the approval after your comment is addressed.

tests/python/pytorch/graphbolt/test_feature_store.py
czkkkkkk: It is weird because it still work... Maybe it is because the content on the file is still there. I will change the code to avoid this issue.

python/dgl/graphbolt/feature_store.py
czkkkkkk: I think so. `FeatureStore` can be extended easily by users.

tests/python/pytorch/graphbolt/test_feature_store.py
czkkkkkk: Added a comment. This comment is also necessary when I write `load_feature_store`.

Rhett-Ying (approved):
LGTM

czkkkkkk:
@dgl-bot

PR #5901, opened 2023-06-25
[GraphBolt] DataLoader end-to-end example

BarclayII:
## Description
This provides an end-to-end DataLoader example using DataPipes.  See the test case for end-to-end usage.

It does subgraph sampling with multiprocessing, and feature fetching with daemonic Python thread.  Please see if you can understand the logic better.

Closes #5831 .

The logic for CUDA stream event recording is from PR #4250.  So I'm requesting the author @yaox12 for review as well to see if my understanding of stream event recording is correct.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [x] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [x] All changes have test coverage
- [x] Code is well-documented
- [x] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change
- [x] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

PR #5918, opened 2023-06-28
[GraphBolt] Datapipe for copying to given device

BarclayII:
## Description
This is a sub-PR of #5901 .

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/copy_to.py
+++ b/python/dgl/graphbolt/copy_to.py
@@ -1,0 +1,42 @@
+"""Graph Bolt CUDA-related Data Pipelines"""
+
+import torch
+from torchdata.datapipes.iter import IterDataPipe
+
+from ..utils import recursive_apply
+
+
+def _to(x, device):
+    return x.to(device) if hasattr(x, "to") else x
+
+
+class CopyTo(IterDataPipe):
+    """DataPipe that transfers each element yielded from the previous DataPipe
+    to the given device.
+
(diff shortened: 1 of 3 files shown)


python/dgl/graphbolt/copy_to.py
```diff
@@ -0,0 +1,42 @@
...
+from torchdata.datapipes.iter import IterDataPipe
+
+from ..utils import recursive_apply
+
+
+def _to(x, device):
+    return x.to(device) if hasattr(x, "to") else x
+
+
+class CopyTo(IterDataPipe):
```
frozenbugs, line 13: Why not also use Mapper?

BarclayII: I think the current implementation is more straightforward.  If I use a `Mapper` then I'll have to write it like:
```python
def __init__(self, dp, device):
    self.device = device
    super().__init__(dp, self.map_func)
def map_func(self, data):
    return recursive_apply(data, _to, self.device)
```

frozenbugs: This is 5 lines compare to existing 8 lines of code, it is not bad.

BarclayII: But I don't think it's as readable as 8 lines of code.  A for loop is cleaner than passing in a function IMO.

frozenbugs (approved):

PR #5925, opened 2023-06-28
Removing warnings in the EEG-GNN example.

drivanov:
## Description

The purpose of the PR is to correct the following two warnings:

- `DtypeWarning: Columns (0) have mixed types. Specify dtype option on import or set low_memory=False.`

- `UserWarning: The memmapped array [[ -572.13196116  -363.93541683  -739.53754843 ...  -357.44040591
  -1722.60122587 -1294.1573575 ]] loaded from the file psd_features_data_X is not not bytes aligned. This may cause segmentation faults if this memmapped array is used in some libraries like BLAS or PyTorch. To get rid of this warning, regenerate your pickle file with joblib >= 1.2.0. See https://github.com/joblib/joblib/issues/563 for more details
  warnings.warn(message)`

## Checklist
- [x] Code is well-documented

## Changes

The first warning has been fixed with the `low_memory=False` option.
The second one was fixed by regenerating the pickle file.


mufeili:
Previously, how did you trigger these two warnings?

frozenbugs:
@dgl-bot

jermainewang:
> Previously, how did you trigger these two warnings?

@drivanov Friendly nudge.

frozenbugs:
@dgl-bot

drivanov:
First warning
```
DtypeWarning: Columns (0) have mixed types. Specify dtype option on import or set low_memory=False.
```
was generated because `low_memory=False` was not used in line 81 of  `eeg-gcnn/main.py`:
```
    _DATASET_INDEX = pd.read_csv("master_metadata_index.csv", low_memory=False)
```

Second one:
```
UserWarning: The memmapped array [[ -572.13196116  -363.93541683  -739.53754843 ...  -357.44040591   -1722.60122587 -1294.1573575 ]] loaded from the file psd_features_data_X is not not bytes aligned. ...
```
was there because the size of the `psd_features_data_X` file was 86528478 and it doesn't match the memory mapped array.

examples/pytorch/eeg-gcnn/main.py
```diff
@@ -91,6 +92,12 @@
     # set up input and targets from files
     memmap_x = f"psd_features_data_X"
     memmap_y = f"labels_y"
+    if os.path.getsize(memmap_x) == 86528478:
```
frozenbugs, line 95: why 86528478 is so special?

drivanov: In fact, 86528478 is the length of the unaligned `psd_features_data_X`. I agree with you, it would be much better to get rid of this constant.  Because the alignment of files is very fast, I implemented
```
def load_memory_mapped_array(file_name):
    # Since `file_name` is not always byte-aligned,
    # let's fix this to avoid the warning.
    joblib.dump(joblib.load(file_name), file_name)
    return load(file_name, mmap_mode="r")
```
and started doing it all the time, no matter the file size.

examples/pytorch/eeg-gcnn/main.py
```diff
@@ -13,6 +15,14 @@
 from sklearn.model_selection import train_test_split
 from torch.utils.data import WeightedRandomSampler
 
+
+def load_memory_mapped_array(file_name):
+    # Since `file_name` is not always byte-aligned,
+    # let's fix this to avoid the warning.
```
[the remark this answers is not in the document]
drivanov: Sorry, I can't find a good explanation why ` joblib.load + joblib.dump` solves this problem. It looks like `joblib.load` is able to read unaligned files and `joblib.dump` and `joblib.dump` always creates properly aligned files.

From what I can see in debugger, the initial size of file "psd_features_data_X" is 86528478. After 
```
       joblib.dump(joblib.load(file_name), file_name)
```
it becomes 86528481. The values of the elements of the mapped arrays `x` and `y`  are the same in both cases:

- before my changes:
 ```
       x = load(memmap_x, mmap_mode="r")
       y = load(memmap_y mmap_mode="r")
 ```
- with my changes:
 ```
       x = load_memory_mapped_array(f"psd_features_data_X")
       y = load_memory_mapped_array(f"labels_y")
```

I am not sure, just another thought... Perhaps `psd_features_data_X` was created with some previous version of `joblib.dump` that had the memory alignment bug mentioned in the [Release 1.2.0](https://joblib.readthedocs.io/en/latest/developing.html) notes:

- **Fix memory alignment bug for pickles containing numpy arrays. This is especially important when loading the pickle with mmap_mode != None as the resulting numpy.memmap object would not be able to correct the misalignment without performing a memory copy. This bug would cause invalid computation and segmentation faults with native code that would directly access the underlying data buffer of a numpy array, for instance C/C++/Cython code compiled with older GCC versions or some old OpenBLAS written in platform specific assembly.**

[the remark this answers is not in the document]
Comment (frozenbugs) [reply examples/pytorch/eeg-gcnn/main.py:21]: I see, thanks for the explanation, looks like the root cause is the original data prepared by the example is slightly polluted. Let's comment it here, see my recommendation blow.

Comment (frozenbugs) [inline examples/pytorch/eeg-gcnn/main.py:21]:
```suggestion
    # Due to a legacy problem related to memory alignment in joblib [1], the
    # data provided in the example may not be byte-aligned. This can be risky
    # when loading with mmap_mode. To fix the issue, load and re-dump the data.
    # [1] https://joblib.readthedocs.io/en/latest/developing.html#release-1-2-0
```

examples/pytorch/eeg-gcnn/main.py
```diff
@@ -13,6 +15,14 @@
 from sklearn.model_selection import train_test_split
 from torch.utils.data import WeightedRandomSampler
 
+
+def load_memory_mapped_array(file_name):
```
frozenbugs, line 19: Internal method.
```suggestion
def _load_memory_mapped_array(file_name):
```

examples/pytorch/eeg-gcnn/main.py
```diff
@@ -1,5 +1,7 @@
 import argparse
 
+import joblib
```
frozenbugs, line 3: duplicate with L12, either remove L12 and use joblib.load for existing load method, or change L12 to `from joblib import dump, load`, and change joblib.dump -> dump.

frozenbugs (approved):

examples/pytorch/eeg-gcnn/main.py
drivanov: DONE. Thanks for the very good suggestions!

frozenbugs:
@dgl-bot

PR #5929, opened 2023-06-29
[Performance] Add NCCL support 

isratnisa:
## Checklist
- [ ] Code is well-documented


yaox12 (approved):
It seems we have to check `torch.distributed.is_initialized()` before calling `th.distributed.get_backend()`. Otherwise LGTM,

python/dgl/distributed/optim/pytorch/sparse_optim.py
```diff
@@ -323,7 +327,11 @@ def step(self):
                                 idx_j = idx_i[mask]
                                 grad_j = grad_i[mask]
                                 idx_split_size.append(
-                                    th.tensor([idx_j.shape[0]], dtype=th.int64)
+                                    th.tensor(
+                                        [idx_j.shape[0]],
+                                        dtype=th.int64,
+                                        device=device,
```
Rhett-Ying, line 333: what if multiple GPUs are being used during train, do we need to specify device id like `cuda:0/1/2/3` here?

python/dgl/distributed/optim/pytorch/sparse_optim.py
```diff
@@ -256,9 +256,9 @@ def step(self):
         of the embeddings involved in a mini-batch to DGL's servers and update the embeddings.
         """
         with th.no_grad():
+            device = "cuda" if th.distributed.get_backend() == "nccl" else "cpu"
```
Rhett-Ying, line 259: we'd better wrap with `torch.device()`?

python/dgl/distributed/optim/pytorch/sparse_optim.py
```diff
@@ -336,40 +344,65 @@ def step(self):
...
-                    alltoall_cpu(
-                        self._rank,
-                        self._world_size,
-                        gather_list,
-                        idx_split_size,
+                        th.empty(
+                            [self._world_size], dtype=th.int64, device=device
+                        ).chunk(self._world_size)
                     )
+                    if th.distributed.get_backend() == "nccl":
```
Rhett-Ying, line 351: how about wrapping these if-else into a function like `dgl.all_to_all()` which is more brief in code or you need to add such logic in multiple places.

Rhett-Ying (changes requested):
dedicated unit test is required

python/dgl/distributed/optim/pytorch/sparse_optim.py
isratnisa: I think you are right. Changed device to   `device = th.device(f"cuda:{self._rank}")`

python/dgl/distributed/optim/pytorch/sparse_optim.py
isratnisa: Fixed

python/dgl/distributed/optim/pytorch/sparse_optim.py
isratnisa: Added.

Rhett-Ying (commented):
could you add test cases into https://github.com/dmlc/dgl/blob/master/tests/python/pytorch/distributed/optim/test_dist_optim.py? and please resolve the lint error first.

Rhett-Ying (approved):
please fix the lint issue before merge

PR #5964, opened 2023-07-07
[GraphBolt] Add sampled sub graph

peizhou001:
## Description
Add sampled sub graph, which should be the common result graph for sampling used in graphbolt.

## Checklist
- [ ] Code is well-documented


python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
```diff
@@ -0,0 +1,153 @@
...
+
+class SampledSubGraph:
+    r"""Class for sampled subgraph."""
+
+    def __init__(
+        self,
+        node_pairs,
+        reverse_column_node_ids=None,
+        reverse_row_node_ids=None,
+        reverse_edge_ids=None,
```
frozenbugs, line 12: Add type.

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
```diff
@@ -0,0 +1,156 @@
...
+    ):
+        """ ""Initialize the SampledSubGraph object. In the context of a
+        heterogeneous graph, each field should be of `Dict` type. For
+        homogeneous graphs, each field should correspond to its respective
+        value type."
+
+        Parameters
+        ----------
+        node_pairs :  Tuple[Tensor] or Dict[(str, str, str), Tuple[Tensor]]
```
frozenbugs, line 23: Do we prefer COO here or CSC?

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
```diff
@@ -0,0 +1,156 @@
...
+            Row's reverse node ids in the original graph. A graph structure
+            can be treated as a coordinated row and column pair, and this is
+            the mapped ids of the row.
+            - If `reverse_row_node_ids` is a tensor: It represents the
+                original node ids.
+            - If `reverse_row_node_ids` is a dictionary: The keys should be
+                node type and the values should be corresponding original
+                heterogeneous node ids.
+        reverse_edge_ids: Optional[Tensor or Dict[(str, str, str), Tensor]]
+            Reverse edge ids in the original graph.This is useful when edge
```
frozenbugs, line 49:
```suggestion
            Reverse edge ids in the original graph. This is useful when edge
```

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
```diff
@@ -0,0 +1,156 @@
...
+            features are needed.
```
frozenbugs, line 50: Clarify when it will be present and when it will be absent.

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
```diff
@@ -0,0 +1,156 @@
+"""Heterogeneous sampled subgraph."""
+# pylint: disable= invalid-name
+from collections.abc import Mapping
+
+
+class SampledSubGraph:
```
frozenbugs, line 6: Instead of making it a class, let's make it a API first, and then implementation.

frozenbugs: And move it to graphbolt/ top level folder.

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
peizhou001: added

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
```diff
@@ -0,0 +1,156 @@
...
+    r"""Class for sampled subgraph."""
+
+    def __init__(
+        self,
+        node_pairs,
+        reverse_column_node_ids=None,
+        reverse_row_node_ids=None,
+        reverse_edge_ids=None,
+    ):
+        """ ""Initialize the SampledSubGraph object. In the context of a
```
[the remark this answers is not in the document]
peizhou001: fixed

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
```diff
@@ -0,0 +1,156 @@
...
+        heterogeneous graph, each field should be of `Dict` type. For
+        homogeneous graphs, each field should correspond to its respective
+        value type."
```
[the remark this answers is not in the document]
peizhou001: fixed

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
```diff
@@ -0,0 +1,156 @@
...
+        Parameters
+        ----------
+        node_pairs :  Tuple[Tensor] or Dict[(str, str, str), Tuple[Tensor]]
+            Node pairs representing source-destination edges.
+            - If `node_pairs` is a tuple: It should be in the format ('u', 'v')
+                representing source and destination pairs.
+            - If `node_pairs` is a dictionary: The keys should be edge type and
+                the values should be corresponding node pairs. The ids inside
+                is heterogeneous ids.
+        reverse_column_node_ids : Optional[Tensor or Dict[str, Tensor]]
```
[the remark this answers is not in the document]
peizhou001: Yes, this is optional. Because some algorithm will not compact columns, we can't assume it is compressed. 
If the sampled graph is of format `CSC`,  it is the input seed nodes.

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
```diff
@@ -0,0 +1,156 @@
...
+            Column's reverse node ids in the original graph. A graph structure
+            can be treated as a coordinated row and column pair, and this is
+            the mapped ids of the column.
+            - If `reverse_column_node_ids` is a tensor: It represents the
+                original node ids.
+            - If `reverse_column_node_ids` is a dictionary: The keys should be
+                node type and the values should be corresponding original
+                heterogeneous node ids.
+        reverse_row_node_ids : Optional[Tensor or Dict[str, Tensor]]
```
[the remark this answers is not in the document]
peizhou001: added

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
peizhou001: Both can be supported in this structure, because both `COO` and  `CSC` are 2 tensors. In our canonical case, the subgraph isn't compacted. Therefore, we opt for `COO` over `CSC` to avoid numerous gaps that would occur with `CSC`.

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
peizhou001: added

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
peizhou001: fixed

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
peizhou001: I'm not quite understand how to make it as an API, because this struct is needed as the result of sampling.

python/dgl/graphbolt/__init__.py
```diff
@@ -14,7 +14,7 @@
 from .dataset import *
 from .dataloader import *
 from .subgraph_sampler import *
-
+from .sampled_sub_graph import *
```
czkkkkkk, line 17: Better to use `.sampled_subgraph`?

python/dgl/graphbolt/sampled_sub_graph.py
```diff
@@ -0,0 +1,172 @@
...
+class SampledSubGraph:
+    r"""Class for sampled subgraph."""
+
+    def __init__(
+        self,
+        node_pairs: Union[
+            Dict[Tuple[str, str, str], Tuple[torch.tensor, torch.tensor]],
+            Tuple[torch.tensor, torch.tensor],
+        ],
+        reverse_column_node_ids: Union[
```
czkkkkkk, line 18: I feel the word `reverse` is not straightforward. Can we use `original`?

python/dgl/graphbolt/sampled_sub_graph.py
```diff
@@ -0,0 +1,172 @@
...
+            Dict[str, torch.tensor], torch.tensor
+        ] = None,
+        reverse_row_node_ids: Union[
```
czkkkkkk, line 21: Since we are not in the scope of a `CSC` graph, it is strange to use `row` and `column`. Can we use `src` and `dst`?

python/dgl/graphbolt/sampled_sub_graph.py
```diff
@@ -0,0 +1,172 @@
...
+            row IDs match these compacted ones.
+        reverse_edge_ids: Optional[Tensor or Dict[(str, str, str), Tensor]]
+            Reverse edge ids in the original graph. This is useful when edge
+            features are needed.
+            - If `reverse_edge_ids` is a tensor: It represents the
+                original edge ids.
+            - If `reverse_edge_ids` is a dictionary: The keys should be
+                edge type and the values should be corresponding original
+                heterogeneous edge ids.
+
```
czkkkkkk, line 72: Better to give an example here.

python/dgl/graphbolt/sampled_sub_graph.py
```diff
@@ -0,0 +1,172 @@
+"""Heterogeneous sampled subgraph."""
+# pylint: disable= invalid-name
+from collections.abc import Mapping
+from typing import Dict, Tuple, Union
+
+import torch
+
+
+class SampledSubGraph:
```
czkkkkkk, line 9: Since this class is purely for storing data, can we use `@dataclass` or `NamedTuple`?

BarclayII: Also change `SubGraph` to `Subgraph`.

python/dgl/graphbolt/sampled_sub_graph.py
```diff
@@ -0,0 +1,172 @@
...
+    r"""Class for sampled subgraph."""
+
+    def __init__(
+        self,
+        node_pairs: Union[
+            Dict[Tuple[str, str, str], Tuple[torch.tensor, torch.tensor]],
```
BarclayII, line 15: The typing annotation should be `torch.Tensor` rather than `torch.tensor`, which is a function.

python/dgl/graphbolt/sampled_sub_graph.py
```diff
@@ -0,0 +1,172 @@
...
+    ):
+        """Initialize the SampledSubGraph object. In the context of a
+        heterogeneous graph, each field should be of `Dict` type. For
+        homogeneous graphs, each field should correspond to its respective
+        value type.
+
+        Parameters
+        ----------
+        node_pairs :  Tuple[Tensor, Tensor] or Dict[(str, str, str), Tuple[Tensor, Tensor]]
+            Node pairs representing source-destination edges.
```
BarclayII, line 36: Would `SampleSubgraph` support subgraphs of a multigraph?  If so, how do we distinguish multi-edges for a subgraph from a multigraph if we only use node pairs?  If not, are we going to make a new class to support it?

BarclayII (changes requested):
I'm concerned about using node pairs to represent edges.  If we only consider simple graphs, we should state so in this class.  Even so, I wonder what we are going to do for subgraphs of multi-graphs.

python/dgl/graphbolt/sampled_sub_graph.py
frozenbugs: Yes, we support multigraph as well, there is a optional "reverse_edge_ids" that can be used to distinguish multiple edges.

python/dgl/graphbolt/__init__.py
peizhou001: make sense, fixed

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
frozenbugs: See this https://github.com/dmlc/dgl/blob/master/python/dgl/graphbolt/dataset.py as an example.

python/dgl/graphbolt/sampled_sub_graph.py
peizhou001: fixed

python/dgl/graphbolt/sampled_sub_graph.py
peizhou001: fixed

python/dgl/graphbolt/sampled_sub_graph.py
peizhou001: Actually we don't want to limit the format of the sub graph, it could be a `CSC`, `CSR` or `COO`, and all of them can be represented as 2 tensors. From this perspective,  it's better to use `row` and `column`?

frozenbugs (commented):
Can you separate it to 2 PRs? 1 for refactoring, the other for adding API.

python/dgl/graphbolt/graph_storage/sampled_sub_graph.py
peizhou001: Changed to API style.

python/dgl/graphbolt/sampled_sub_graph.py
peizhou001: Maybe `reverse` can intuitively reflect the mapping processing?

python/dgl/graphbolt/sampled_sub_graph.py
peizhou001: Sure, added one example

python/dgl/graphbolt/__init__.py
```diff
@@ -45,4 +45,4 @@ def load_graphbolt():
 
 load_graphbolt()
 
-SampledSubgraph = torch.classes.graphbolt.SampledSubgraph
+C_SampledSubgraph = torch.classes.graphbolt.SampledSubgraph
```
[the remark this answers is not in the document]
peizhou001: Yes, it will be removed from here.

python/dgl/graphbolt/sampled_subgraph.py
```diff
@@ -0,0 +1,70 @@
...
+        the mapped ids of the column.
+        - If `reverse_column_node_ids` is a tensor: It represents the
+            original node ids.
+        - If `reverse_column_node_ids` is a dictionary: The keys should be
+            node type and the values should be corresponding original
+            heterogeneous node ids.
+        If present, it means column IDs are compacted, and `node_pairs`
+        column IDs match these compacted ones.
+        """
+        raise NotImplementedError
```
frozenbugs, line 42: nit: Since the following 3 methods are optional, return None instead of raise error?

python/dgl/graphbolt/impl/csc_sampling_graph_sampled_subgraph.py
```diff
@@ -0,0 +1,70 @@
...
+from dataclasses import dataclass
+from typing import Dict, Tuple, Union
+
+import torch
+
+from ..sampled_subgraph import SampledSubGraph
+
+
+@dataclass
+class CSCSamplingGraphSampledSubgraph(SampledSubGraph):
```
frozenbugs, line 12: SampledSubgraphImpl

python/dgl/graphbolt/impl/csc_sampling_graph_sampled_subgraph.py
frozenbugs, line 1: sampled_subgraph_impl.py

frozenbugs (approved):

python/dgl/graphbolt/impl/csc_sampling_graph_sampled_subgraph.py
peizhou001: make sense, changed for fix.

python/dgl/graphbolt/sampled_subgraph.py
peizhou001: Make sense, changed for fix.

python/dgl/graphbolt/sampled_subgraph.py
```diff
@@ -0,0 +1,70 @@
+"""Graphbolt sampled subgraph."""
+# pylint: disable= invalid-name
+from typing import Dict, Tuple
+
+import torch
+
+
+class SampledSubGraph:
```
czkkkkkk, line 8: Do we use `SampledSubgraph` or `SampledSubGraph`?

PR #5969, opened 2023-07-07
Cpu docker tcmalloc

daniil-sizov:
## Description

tcmalloc significantly improves performance for large datasets on CPU

Graphsage standalone, ogbn-products (s/epoch):
Intel(R) Xeon(R) Gold 6152 CPU @ 2.10GHz
| no tcmalloc | tcmalloc |
| ------------- | --------- |
| 64.8298       | 54.7999  |

Graphsage distributed, ogbn-papers100M (s/epoch):
r6i.4xlarge, OMP_NUM_THREADS=6
| no tcmalloc | tcmalloc |
| ------------- | --------- |
| 59.9632       | 44.5427  |

## Checklist


docker/Dockerfile.ci_cpu
```diff
@@ -13,6 +13,10 @@ RUN bash /install/ubuntu_install_core.sh
 COPY install/ubuntu_install_build.sh /install/ubuntu_install_build.sh
 RUN bash /install/ubuntu_install_build.sh
 
+# tcmalloc
+RUN apt-get install -y libgoogle-perftools4
+ENV LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libtcmalloc.so.4:$LD_PRELOAD
```
Rhett-Ying, line 18: Once prepend the tcmalloc to `LD_PRELOAD`, will `dgllib.so` always use tcmalloc instead glic for memory allocation automatically?

Rhett-Ying: do we need to link tcmalloc when building dgllib.so?

itaraban: Yes, `tcmalloc` will be used for all glibc functions for PyTorch and DGL(and other).
Unfortunately, linkage with tcmalloc doesn't work properly, that's why we propose to use `PRELOAD`

Rhett-Ying: I see. But updating the docker image(we need to build new docker image with the changed dockerfile) benefit CI tests only.If `tcmalloc` proves to always outperform `glibc`, we'd better mention such performance improvement in user guide? Then user could try with `tcmalloc` if they want. Or if DGL tries to force user to always use tcmalloc, DGL lib should be built/linked with it directly?

daniil-sizov: Good idea, added to the user guide.
Unfortunately it is seems not possible to build DGL with tcmalloc since pytorch is not built with it. Mismatch in memory allocators causes crashes

frozenbugs:
@Rhett-Ying is out for this week, if not an emergent PR, let's way for Rhett coming back next monday.

Rhett-Ying (approved):

PR #5972, opened 2023-07-10
[GraphBolt] init graph topology

Rhett-Ying:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/impl/ondisk_dataset.py
+++ b/python/dgl/graphbolt/impl/ondisk_dataset.py
@@ -88,7 +88,7 @@ class OnDiskDataset(Dataset):
         """Return the test set."""
         return self._test_sets
 
-    def graph(self) -> CSCSamplingGraph:
+    def graph(self) -> object:
         """Return the graph."""
         return self._graph
 
--- a/python/dgl/graphbolt/impl/ondisk_dataset.py
+++ b/python/dgl/graphbolt/impl/ondisk_dataset.py
@@ -3,9 +3,11 @@
 from typing import Dict, List, Tuple
 
 from ..dataset import Dataset
+
+from ..graph_storage import CSCSamplingGraph, load_csc_sampling_graph
 from ..itemset import ItemSet, ItemSetDict
 from ..utils import read_data, tensor_to_tuple
-from .ondisk_metadata import OnDiskMetaData, OnDiskTVTSet
+from .ondisk_metadata import OnDiskGraphTopology, OnDiskMetaData, OnDiskTVTSet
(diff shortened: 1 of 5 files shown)


python/dgl/graphbolt/impl/ondisk_dataset.py
```diff
@@ -82,14 +88,22 @@ def test_sets(self) -> List[ItemSet] or List[ItemSetDict]:
...
         """Return the feature."""
         return self._feature
 
+    def _load_graph(
+        self, graph_topology: OnDiskGraphTopology
+    ) -> CSCSamplingGraph:
+        """Load the graph topology."""
+        if graph_topology is None:
+            return None
+        return load_csc_sampling_graph(graph_topology.path)
```
frozenbugs, line 105: In yaml, we have type: CSCSamplingGraph, we should check the type first, such that in the future when we support new type, the old dgl version can failure properly.

python/dgl/graphbolt/impl/ondisk_dataset.py
```diff
@@ -82,14 +88,22 @@ def test_sets(self) -> List[ItemSet] or List[ItemSetDict]:
         """Return the test set."""
         return self._test_sets
 
-    def graph(self) -> object:
+    def graph(self) -> CSCSamplingGraph:
```
frozenbugs, line 91: I think we can still be very soft here to keep it as object.

python/dgl/graphbolt/impl/ondisk_dataset.py
Rhett-Ying: when we support new type, the old one(type) is supposed to work well?
pydantic is helping us verify the enum value automatically.

python/dgl/graphbolt/impl/ondisk_dataset.py
Rhett-Ying: sure as more types might be supported later

python/dgl/graphbolt/impl/ondisk_dataset.py
frozenbugs: I see, the I have less concern. nit: still worth to check the OnDiskGraphTopologyType = CSC_SAMPLING before using load_csc_sampling_graph(graph_topology.path), and return none for others.

frozenbugs (approved):

python/dgl/graphbolt/impl/ondisk_dataset.py
Rhett-Ying: sure.

PR #5985, opened 2023-07-11
Improving the Capsule Network example.

drivanov:
## Description

By using `dgl.graph()` instead of the previously used `dgl.DGLGraph()` we not only remove the warning:
```
/usr/local/lib/python3.10/dist-packages/dgl/heterograph.py:92: DGLWarning: Recommend creating graphs by `dgl.graph(data)` instead of `dgl.DGLGraph(data)`.
```

but we also make the Capsule Network example approximately 10.7% faster". 

Here is the execution time for 5 epochs:

```
          Before:           After:     
real    16m 42.777s      14m 55.046s
user    17m 37.111s      15m 49.642s
```

Changes to `python/dgl/convert.py` make it easier to remove similar warnings for other examples.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/.lintrunner.toml
+++ b/.lintrunner.toml
@@ -39,6 +39,7 @@ include_patterns = [
     '**/*.cu',
 ]
 exclude_patterns = [
+    'third_party/**',
 ]
 init_command = [
     'python3',
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -227,6 +227,21 @@ if((NOT MSVC) AND (NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin"))
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wl,--exclude-libs,ALL")
 endif()
 
+# Compile gpu_cache
+if(USE_CUDA)
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DUSE_GPU_CACHE")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DUSE_GPU_CACHE")
+  # Manually build gpu_cache because CMake always builds it as shared
+  file(GLOB gpu_cache_src
+    third_party/HugeCTR/gpu_cache/src/nv_gpu_cache.cu
+  )
+  cuda_add_library(gpu_cache STATIC ${gpu_cache_src})
+  target_include_directories(gpu_cache PRIVATE "third_party/HugeCTR/gpu_cache/include")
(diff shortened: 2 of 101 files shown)


frozenbugs:
@dgl-bot

frozenbugs:
@dgl-bot

jermainewang:
Hi @drivanov , thanks for the contribution. I like your fix to remove those warnings. My suggestion is to keep the changes to the capsule network example and try to avoid modifying the core library code such as `convert.py`. Could you help update the PR as follows:
* Revert the changes to `convert.py`.
* Refactor the `init_graph` function to first generate an edge list (e.g., two integer list `u` and `v`) and then pass them to `dgl.graph` call. This is the intended way of constructing a graph in DGL right now.

Thanks!

drivanov:
Hi @jermainewang, In fact, both integer lists **u** and **v** that you talked about in this example must be empty. So the simplest solution to this problem would be to replace the call
```
g = dgl.DGLGraph()
```
in `init_graph` by one of the following calls
```
g = dgl.graph(data=([], []), num_nodes=0)
g = dgl.graph(data=([], []))
g = dgl.graph(data=[])
g = dgl.graph([])
```

Unfortunately, it's not possible to just use
```
g = dgl.graph()
```
because `dgl.graph(...)` requires positional argument `data`.
The changes I made to  `convert.py` remove this limitation and make it easier to remove similar warnings for other examples. As I could see, the `dgl.DGLGraph()` call is used in 49 different places and I think it would be easier to replace
```
g = dgl.DGLGraph()
```
by 
```
g = dgl.graph()
```
in these places.

jermainewang:
@drivanov What I mean is to change the `init_graph` function to something like follows:
```python
def init_graph(in_nodes, out_nodes, f_size, device="cpu"):
    src, dst = [], []
    in_indx = list(range(in_nodes))
    out_indx = list(range(in_nodes, in_nodes + out_nodes))
    # add edges use edge broadcasting
    for u in in_indx:
        src += [u] * len(out_indx)
        dst += out_indx
    ...   # other logics
    g = dgl.graph((src,dst))  # dgl.graph once;
    ...   # other logics
```

Basically, replace all `add_nodes` and `add_edges` with logics for manipulating the src and dst list directly.

drivanov:
@jermainewang: As you previously suggested, I 
- reverted my changes to `convert.py`
- refactored the `init_graph` function.

With these changes, I see a 7x (700%) runtime improvement, and with my initial changes, I only saw a 1.1x (10%) runtime improvement.

frozenbugs:
@dgl-bot

jermainewang:
@drivanov Nice work! Thanks for the contribution!

jermainewang (approved):

PR #5986, opened 2023-07-12
[Graphbolt][Feature] LABOR-0 implementation into sample_neighbors with tests

mfbalin:
## Description
Implemented LABOR-0 versions of uniform and nonuniform pick functions. Sharing almost the whole code path with neighbor sampling with the only addition of an additional flag and the double use of the replace parameter in the implementation details to be random seed when labor is enabled instead of the replace parameter.

Sequential Poisson Sampling is used to deterministically sample fanout neighbors as described in the labor arxiv paper. https://arxiv.org/pdf/2210.13339.pdf Section A.3. Basically indistinguishable from neighbor sampling (dgl.sampling.sample_labors samples fanout neighbors in expectation unlike here.) except for the more overlapping and smaller sampled (by number of unique vertices) subgraph. If neighborhoods of seed vertices have zero overlap with each other, then is exactly equivalent to neighbor sampling.

Runtime complexity for fanout `k` and `d` neighbors is O(d + k log(d / k)) = O(d) with O(k) space complexity, but a two pass algorithm with space complexity O(d) and runtime complexity O(d) is also possible.

## Checklist
- [x] Code is well-documented


graphbolt/include/graphbolt/csc_sampling_graph.h
```diff
@@ -203,6 +203,97 @@ class CSCSamplingGraph : public torch::CustomClassHolder {
       const std::string& shared_memory_name);
 
  private:
+  template <bool labor>
```
frozenbugs, line 206: Avoid moving such that it is easier for the reviewers to see the diff?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -207,6 +198,34 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighbors(
...
+    // stack, which doesn't support 'torch.half' and 'torch.bool' data types. To
+    // avoid crashes, convert 'probs_or_mask' to 'float32' data type.
+    if (probs_or_mask.value().dtype() == torch::kBool ||
+        probs_or_mask.value().dtype() == torch::kFloat16) {
+      probs_or_mask = probs_or_mask.value().to(torch::kFloat32);
+    }
+  }
+  if (labor) {
+    TORCH_CHECK(
+        !replace, "Sampling with replacement is not supported for labor.");
```
frozenbugs, line 218: There are 2 concerns here:
1. labor sampling is not yet a widely received sampling algorithm, it might be confusing to user who have no clue about what's labor.
2. The code here are intertwined heavily, creating a lot of if-else, which is not very readability friendly.

I'd suggest:
PR#1. Create a `def sample_neighbors_labor(...)`, put a clear documentation without implementation and add empty unit tests.

PR#2. Create a new C++ implementation, instead of overwriting SampleNeighbors, create SampleNeighborsLabor(...), same for pick function, and update the unit tests coorespondingly.

mfbalin: So we don't mind copying the whole of Sample Neighbors code along with PickByEType binary search logic? If I separate out the labor code, it will be an extra maintenance burden and fall behind the neighbor sampler when it comes to functionalities as the neighbor sampler code is updated. This code strictly aims to be %100 compatible with NeighborSampling. Even the existing tests can't detect if labor is enabled, they all just pass.

My plan was to have another version with a separate API where the %100 compatibility is not followed (Implementing the expected fanout algorithm with importance sampling as in dgl.sampling.sample_labors).

About 1: It is possible to change the bool parameter's name to layer, since layer sampling algorithms have been around for 5+ years, users should be more familiar with them. We can explain in the documentation that what enables layer sampling functionality for NeighborSampler is the Labor algorithm.

About 2: Would it be acceptable if I made the code more readable by eliminating the extra ifs but still using templates to share the code? It is possible to do that. I just want this code to be maintained just as well as neighbor sampling.

mfbalin: I will shortly push a much more readable version to illustrate what I mean.

mfbalin: I had to use only 1 extra if expression potentially making it less readable, and it is to dispatch Neighbor or Labor from SampleNeighbors by passing the right template parameter to SampleNeighborsImpl. The code should read the same now, the other existing functions are untouched except that they are templated now. What do you think?

I also changed the outer looking API parameter name to layer which the users should already know about.

mfbalin: My reasoning is that if somebody for example decided to optimize the multithreading logic for neighbor sampling, the benefits will automatically transfer to labor as well, the goal is to keep their quality on par so that labor is always production ready too.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -129,22 +130,13 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
           : torch::nullopt);
 }
 
-c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighbors(
+template <sampler_t sampler>
+c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
     const torch::Tensor& nodes, const std::vector<int64_t>& fanouts,
     bool replace, bool return_eids,
-    torch::optional<std::string> probs_name) const {
+    const torch::optional<torch::Tensor>& probs_or_mask,
```
frozenbugs, line 137: This change probably due to poor conflict resolving.

graphbolt/include/graphbolt/csc_sampling_graph.h
```diff
@@ -16,6 +16,20 @@
 namespace graphbolt {
 namespace sampling {
 
+enum sampler_t { NEIGHBOR, LABOR };
+
+template <sampler_t sampler>
```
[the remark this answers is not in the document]
frozenbugs: or `template <SamplerType Sampler>`

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -472,5 +498,95 @@ torch::Tensor PickByEtype(
...
+        }));
+    return picked_neighbors;
+  } else {
+    return LaborPick<false>(
+        offset, num_neighbors, fanout, args.random_seed, options, probs_or_mask,
+        args.indices);
+  }
+}
+
+template <bool nonuniform, typename fp_t = float>
```
frozenbugs, line 524:
```suggestion
template <bool NonUniform, typename T = float>
```

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -472,5 +498,95 @@ torch::Tensor PickByEtype(
...
+inline torch::Tensor LaborPick(
```
frozenbugs, line 525: Can you add elaboration about how Labor Pick works? Similar to UniformPick / NonUniformPick above.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -472,5 +498,95 @@ torch::Tensor PickByEtype(
...
+    int64_t offset, int64_t num_neighbors, int64_t fanout, int64_t random_seed,
+    const torch::TensorOptions& options,
+    const torch::optional<torch::Tensor>& probs_or_mask,
+    const torch::Tensor& indices) {
+  fanout = fanout < 0 ? num_neighbors : std::min(fanout, num_neighbors);
+  if (!nonuniform && fanout >= num_neighbors) {
+    return torch::arange(offset, offset + num_neighbors, options);
+  }
+  torch::Tensor heap_tensor = torch::empty({fanout * 2}, torch::kInt32);
+  // Assuming max_degree of a vertex is <= 4 billion
```
frozenbugs, line 535: nit
```suggestion
  // Assuming max_degree of a vertex is <= 4 billion.
```

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -472,5 +498,95 @@ torch::Tensor PickByEtype(
...
+      nonuniform ? probs_or_mask.value().data_ptr<fp_t>() + offset : nullptr;
+  AT_DISPATCH_INTEGRAL_TYPES(
+      indices.scalar_type(), "LaborPickMain", ([&] {
+        const scalar_t* local_indices_data =
+            indices.data_ptr<scalar_t>() + offset;
+        for (uint32_t i = 0; i < fanout; ++i) {
+          const auto t = local_indices_data[i];
+          pcg32 ng(random_seed, t);
+          std::uniform_real_distribution<float> uni;
+          auto rnd = uni(ng);  // r_t
```
frozenbugs, line 548: what does r_t mean?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -472,5 +498,95 @@ torch::Tensor PickByEtype(
...
+        for (uint32_t i = fanout; i < num_neighbors; ++i) {
+          const auto t = local_indices_data[i];
+          pcg32 ng(random_seed, t);
+          std::uniform_real_distribution<float> uni;
+          auto rnd = uni(ng);  // r_t
+          if constexpr (nonuniform) {
+            rnd = local_probs_data[i] > 0
+                      ? (float)(rnd / local_probs_data[i])
+                      : std::numeric_limits<float>::infinity();
+          }  // r_t / \pi_t
```
frozenbugs, line 568: Create a inline method for this, such that you don't need to copy paste code?

graphbolt/src/csc_sampling_graph.cc
frozenbugs: Thanks for the quick action, the implementation now is much more readable. I made some comment inline.

For the python level API, let me discuss with my team to see whether people prefer keep a single api with option or keep separate apis.

graphbolt/src/csc_sampling_graph.cc
mfbalin: https://arxiv.org/pdf/2210.13339.pdf Page 13 algorithm 1 has r_t \leq c_s \pi_t. But here, I used Page 14, A3: Sampling a fixed number of neighbors.

mfbalin: This code has heap functions etc. which are logarithmic but the complexity analysis yields a linear complexity :).

mfbalin:
The name of the labor paper is Layer-Neighbor Sampling -- Defusing Neighborhood Explosion in GNNs. So it is Neighbor Sampling with layer sampling :).

graphbolt/src/csc_sampling_graph.cc
mfbalin: The name of the labor paper is Layer-Neighbor Sampling -- Defusing Neighborhood Explosion in GNNs. So it is Neighbor Sampling with layer sampling :). So it would factually be correct for SampleNeighbors to have a layer bool parameter. But, of course, it is your and your team's decision.

graphbolt/src/csc_sampling_graph.cc
mfbalin: maybe you can check the names of the new inline functions.

frozenbugs:
Per discussion with my team, we prefer to create separated python api instead of fuse them together, here is the example code https://github.com/dmlc/dgl/pull/6002/files that you can copy from. The reason we have internal _sample_neighbor is due to distdgl has a legacy dependency, for sample_labor, we don't need to follow the pattern.

mfbalin:
How do you think the outer api should look? I want to have a version of labor that is identical to the neigbor sampling api in terms of api and behaviour. But I also want to have a separate api that is more advanced and not identical in behaviour (such as importance sampling, kappa feature, layer_dependency or possibly sampling an expected number of vertices instead of deterministic).

mfbalin:
I implemented sampling with replacement as well. Full functionality matching sample_neighbors available now.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -207,6 +200,34 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighbors(
       subgraph_reverse_edge_ids, subgraph_type_per_edge);
 }
 
+c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighbors(
+    const torch::Tensor& nodes, const std::vector<int64_t>& fanouts,
+    bool replace, bool labor, bool return_eids,
```
frozenbugs, line 205: I still see the mix-usage of layer (in header file) and labor (here), since the apis in python are clarified, maybe change them all to labor or layer or layer_neighbor? either should be fine now.

frozenbugs:
> I implemented sampling with replacement as well. Full functionality matching sample_neighbors available now.

Thanks for the quick follow up, the replacement code looks good, but in the future let's avoid extending the scope of a PR while it is being reviewed. Let's get a PR approved and merged before we extending the scope. For the new request, you can check out from the PR and work from there, and git rebase once the PR is merged.

frozenbugs:
> How do you think the outer api should look? I want to have a version of labor that is identical to the neigbor sampling api in terms of api and behaviour. But I also want to have a separate api that is more advanced and not identical in behaviour (such as importance sampling, kappa feature, layer_dependency or possibly sampling an expected number of vertices instead of deterministic).

For the version of labor that is identical to the neighbor sampling api, what's the difference then? Is there any benefit of having a labor version identical to the neighbor sampling?

mfbalin:
> > How do you think the outer api should look? I want to have a version of labor that is identical to the neigbor sampling api in terms of api and behaviour. But I also want to have a separate api that is more advanced and not identical in behaviour (such as importance sampling, kappa feature, layer_dependency or possibly sampling an expected number of vertices instead of deterministic).
> 
> For the version of labor that is identical to the neighbor sampling api, what's the difference then? Is there any benefit of having a labor version identical to the neighbor sampling?

Since users are familiar with neighbor sampler and its behavior, I thought this implementation is a true drop-in replacement so the switch would be easy for the users.

But I believe that neither sampling with replacement or without replacement are good for weighted sampling. Thus I want to implement Poisson sampling as well. Importance sampling will be supported only for poisson sampling also.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -424,10 +445,12 @@ inline torch::Tensor NonUniformPick(
...
 }
 
-torch::Tensor Pick(
+template <>
+torch::Tensor Pick<SamplerType::NEIGHBOR>(
     int64_t offset, int64_t num_neighbors, int64_t fanout, bool replace,
     const torch::TensorOptions& options,
-    const torch::optional<torch::Tensor>& probs_or_mask) {
+    const torch::optional<torch::Tensor>& probs_or_mask,
+    SamplerArgs<SamplerType::NEIGHBOR> args) {
```
peizhou001, line 453: ` SamplerArgs<SamplerType::NEIGHBOR>&`

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -436,11 +459,12 @@ torch::Tensor Pick(
   }
 }
 
+template <SamplerType S>
 torch::Tensor PickByEtype(
     int64_t offset, int64_t num_neighbors, const std::vector<int64_t>& fanouts,
     bool replace, const torch::TensorOptions& options,
     const torch::Tensor& type_per_edge,
-    const torch::optional<torch::Tensor>& probs_or_mask) {
+    const torch::optional<torch::Tensor>& probs_or_mask, SamplerArgs<S> args) {
```
peizhou001, line 467: Ditto

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -472,5 +496,233 @@ torch::Tensor PickByEtype(
...
+    return LaborPick<false, true>(
+        offset, num_neighbors, fanout, options, probs_or_mask, args);
+  } else {  // replace = false
+    return LaborPick<false, false>(
+        offset, num_neighbors, fanout, options, probs_or_mask, args);
+  }
+}
+
+template <typename T>
+inline T labor_uniform_random(int64_t random_seed, int64_t t) {
```
peizhou001, line 529: Move this to `random.h` under graphbolt folder as this seems a general function.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -472,5 +496,233 @@ torch::Tensor PickByEtype(
...
+  pcg32 ng(random_seed, t);
+  std::uniform_real_distribution<T> uni;
+  return uni(ng);
+}
+
+template <typename T>
+inline T invcdf(T u, int64_t n, T rem) {
```
peizhou001, line 536: Could you give a more meaningful name about the argument or add a explained formula ?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -472,5 +496,233 @@ torch::Tensor PickByEtype(
...
+  constexpr T one = 1;
+  return rem * (one - std::pow(one - u, one / n));
+}
+
+template <typename T>
+inline T labor_jth_sorted_uniform_random(
+    int64_t random_seed, int64_t t, int64_t c, int64_t j, T& rem, int64_t n) {
+  const auto u = labor_uniform_random<T>(random_seed, t + j * c);
+  rem -= invcdf(u, n, rem);
```
peizhou001, line 545: could n be 0?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -472,5 +496,233 @@ torch::Tensor PickByEtype(
...
+  AT_DISPATCH_INTEGRAL_TYPES(
+      args.indices.scalar_type(), "LaborPickMain", ([&] {
+        const scalar_t* local_indices_data =
+            args.indices.data_ptr<scalar_t>() + offset;
+        if constexpr (Replace) {
+          // [Algorithm] @mfbalin
+          // Use a max-heap to get rid of the big random numbers and filter the
+          // smallest fanout of them. Implements arXiv:2210.13339 Section A.3.
+          // Unlike sampling without replacement below, the same item can be
+          // included fanout many times in our sample. Thus, we sort and pick
```
peizhou001, line 602: fanout times or many times?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -472,5 +496,233 @@ torch::Tensor PickByEtype(
...
+          // the smallest fanout random numbers out of num_neighbors * fanout
+          // of them. Each item has fanout many random numbers in the race and
+          // the smallest fanout of them get picked. Instead of generating
+          // fanout * num_neighbors random numbers and increase the complexity,
+          // I devised an algorithm to generate the fanout numbers for an item
+          // in a sorted manner on demand, meaning we continue generating random
+          // numbers for an item if it has been sampled that many times already.
```
peizhou001, line 609: even if?

graphbolt/src/csc_sampling_graph.cc
mfbalin: int64_t offset, int64_t num_neighbors are not passed by reference. This is because they are cheap to copy. You can treat this class same as those arguments. Any expensive members can be references in the class instead.

mfbalin: Making it a reference will make labor version of the code slower.

graphbolt/src/csc_sampling_graph.cc
mfbalin: Same as above.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -472,5 +496,233 @@ torch::Tensor PickByEtype(
...
+              } else if (rnd < heap_data[0].first) {
+                std::pop_heap(heap_data, heap_data + fanout);
+                heap_data[fanout - 1] = std::make_pair(rnd, i);
+                std::push_heap(heap_data, heap_data + fanout);
+              } else {
+                rem_data[i] = -1;
+              }
+            }
+          }
+          for (uint32_t i = 0; i < num_neighbors; ++i) {
```
peizhou001, line 644: L623 - 641 is basically same as L644 - 663, can we try to share the code?

graphbolt/src/csc_sampling_graph.cc
mfbalin: No, it would cause a division by 0 in the formula above. Should I place a note?

graphbolt/src/csc_sampling_graph.cc
mfbalin: only if :)

frozenbugs:
@dgl-bot

graphbolt/src/csc_sampling_graph.cc
peizhou001: I see, it's okay to pass the argument by value . But a little confused why it is slower.

graphbolt/src/csc_sampling_graph.cc
mfbalin: We can make this argument layer since it enables layer sampling functionality for neighbor. But the underlying algorithm powering this is labor, so we can keep the implementation details labor if you would like.

frozenbugs:
> > > How do you think the outer api should look? I want to have a version of labor that is identical to the neigbor sampling api in terms of api and behaviour. But I also want to have a separate api that is more advanced and not identical in behaviour (such as importance sampling, kappa feature, layer_dependency or possibly sampling an expected number of vertices instead of deterministic).
> > 
> > 
> > For the version of labor that is identical to the neighbor sampling api, what's the difference then? Is there any benefit of having a labor version identical to the neighbor sampling?
> 
> Since users are familiar with neighbor sampler and its behavior, I thought this implementation is a true drop-in replacement so the switch would be easy for the users.
> 
> But I believe that neither sampling with replacement or without replacement are good for weighted sampling. Thus I want to implement Poisson sampling as well. Importance sampling will be supported only for poisson sampling also.
> 
> That is why, I initially wanted to add a bool layer parameter to sample_neighbors. My plan now is to expose sample_layer_neighbors with same API as a drop-in replacement to sample_neighbors while sample_labors will be similar to dgl.sampling.sample_labors. Does that make sense? Please let me know how best I can present Labor to users.

It makes sense to introduce multiple APIs if needed, we need to make sure that when we create new API, we make it clear to users in which case / why they should chose the new APIs for sampling. Currently I am a little bit confused here:

1. My understand of introduce the labor sampling is to get higher overlap when sampling such that we can leverage cache to reduce the feature fetching speed, if that's the case, why do we need 2 different labor API?

2. You mentioned that a version of Labor sampling and neighbor sampling as identical, why do we need an identical version, is it faster?

graphbolt/src/csc_sampling_graph.cc
mfbalin: If you look at the specializations for neighbor and labor, for neighbor the class has no members (for now) and for labor, it has 2 int64_t arguments and one reference to a torch::Tensor. The function arguments are passed by copy for int64_t values and by reference for torch::optional or potentially expensive items. If we pass the class by copy, then it will pass the two int64_t members by copy, matching the existing code, and it will pass the reference by copy, which is basically passing by reference. (passing reference by copy is like copying the pointer.). So passing the args by copy does the exact same thing as if members of it were being passed separately.

mfbalin: If we pass args by reference, then it will be as if we are passing the torch::tensor by reference twice and passing int64_t values by reference, which is slower than passing by copy.

mfbalin:
> > > > How do you think the outer api should look? I want to have a version of labor that is identical to the neigbor sampling api in terms of api and behaviour. But I also want to have a separate api that is more advanced and not identical in behaviour (such as importance sampling, kappa feature, layer_dependency or possibly sampling an expected number of vertices instead of deterministic).
> > > 
> > > 
> > > For the version of labor that is identical to the neighbor sampling api, what's the difference then? Is there any benefit of having a labor version identical to the neighbor sampling?
> > 
> > 
> > Since users are familiar with neighbor sampler and its behavior, I thought this implementation is a true drop-in replacement so the switch would be easy for the users.
> > But I believe that neither sampling with replacement or without replacement are good for weighted sampling. Thus I want to implement Poisson sampling as well. Importance sampling will be supported only for poisson sampling also.
> > That is why, I initially wanted to add a bool layer parameter to sample_neighbors. My plan now is to expose sample_layer_neighbors with same API as a drop-in replacement to sample_neighbors while sample_labors will be similar to dgl.sampling.sample_labors. Does that make sense? Please let me know how best I can present Labor to users.
> 
> It makes sense to introduce multiple APIs if needed, we need to make sure that when we create new API, we make it clear to users in which case / why they should chose the new APIs for sampling. Currently I am a little bit confused here:
> 
> 1. My understand of introduce the labor sampling is to get higher overlap when sampling such that we can leverage cache to reduce the feature fetching speed, if that's the case, why do we need 2 different labor API?
> 2. You mentioned that a version of Labor sampling and neighbor sampling as identical, why do we need an identical version, is it faster?

1. Both APIs of labor will have higher overlap compared to neighbor. What changes is the sampling method. There are 3 different sampling methods you can use, sampling without replacement, with replacement, poisson sampling (flipping coins). Here, I implemented sampling with and without replacement for labor. dgl.sampling.sample_labors implements poisson sampling.
2. By identical, I mean the API, input arguments, output structure. When a user says sample me fanout edges with replacement, they will get exactly that, with or without labor, when the API is identical. However, if they use labor, they will get more overlap. However, dgl.sampling.sample_labors is not identical because it uses poisson sampling (hence get fanout edges in expectation, not deterministic), not sampling with or without replacement that users are more familiar with.

So, what I mean by having two different labor APIs is that one that looks exactly like neighbor (replace parameter etc.). The other one will look like dgl.sampling.sample_labors, exposing more advanced functionalities.

graphbolt/src/csc_sampling_graph.cc
mfbalin: took out the common part into a lambda function

frozenbugs: helper is always a bad name, can you make it mode contextual? It is better to be long and precise than short and confusing.

mfbalin: Loop body? It had no name before so not sure what to name it.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -472,5 +496,206 @@ torch::Tensor PickByEtype(
...
+          // and each modification takes O(log(fanout)). So the total complexity
+          // is O((fanout + num_neighbors) log(fanout)). It is possible to
+          // decrease the logarithmic factor down to
+          // O(log(min(fanout, num_neighbors))).
+          torch::Tensor remaining =
+              torch::ones({num_neighbors}, torch::kFloat32);
+          float* rem_data = remaining.data_ptr<float>();
+          auto heap_end = heap_data;
+          const auto init_count = (num_neighbors + fanout - 1) / num_neighbors;
+          auto helper = [&](auto t, auto j, auto i) {
```
frozenbugs, line 603: avoid auto / i / j / t in a function definition, even it is a lambda function

graphbolt/src/csc_sampling_graph.cc
mfbalin: Oh, I can call it sample_neighbor_i_with_index_t_jth_time.

frozenbugs: Yes, it sounds like a reasonable.

frozenbugs:
@dgl-bot

peizhou001 (approved):
The logic LGTM, just fix other comments for merge.

PR #5996, opened 2023-07-13
[Graphbolt] Add exclude_edges impl and unittest.

czkkkkkk:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/impl/sampled_subgraph_impl.py
+++ b/python/dgl/graphbolt/impl/sampled_subgraph_impl.py
@@ -68,6 +68,63 @@ class SampledSubgraphImpl(SampledSubgraph):
             ), "Nodes in pairs should be of type torch.Tensor."
 
 
+def _to_reverse_ids(node_pair, reverse_row_node_ids, reverse_column_node_ids):
+    u, v = node_pair
+    if reverse_row_node_ids is not None:
+        u = reverse_row_node_ids[u]
+    if reverse_column_node_ids is not None:
+        v = reverse_column_node_ids[v]
+    return (u, v)
+
+
+def _relabel_homo_edges(lhs_edges, rhs_edges):
+    """Relabel the source nodes and destination nodes independently for two set
+    of edges."""
+    src = torch.cat([lhs_edges[0], rhs_edges[0]])
+    dst = torch.cat([lhs_edges[1], rhs_edges[1]])
+    _, src = torch.unique(src, return_inverse=True)
+    _, dst = torch.unique(dst, return_inverse=True)
+    lhs_edges = (src[: lhs_edges[0].numel()], dst[: lhs_edges[0].numel()])
+    rhs_edges = (src[lhs_edges[0].numel() :], dst[lhs_edges[0].numel() :])
+    return lhs_edges, rhs_edges
+
+
+def _exclude_homo_edges(lhs_edges, rhs_edges):
+    lhs_edges, rhs_edges = _relabel_homo_edges(lhs_edges, rhs_edges)
+    # Compact edges to integers to use torch.isin to exclude edges
+    dst_range = lhs_edges[0].numel() + rhs_edges[0].numel()
+    lhs_val = lhs_edges[0] * dst_range + lhs_edges[1]
+    rhs_val = rhs_edges[0] * dst_range + rhs_edges[1]
+    mask = ~torch.isin(lhs_val, rhs_val)
+    return torch.nonzero(mask, as_tuple=True)[0]
+
+
+def _slice_subgraph(subgraph: SampledSubgraphImpl, index: torch.Tensor):
+    """Slice the subgraph according to the index."""
+
+    def _index_select(obj, index):
+        if obj is None:
+            return None
+        if isinstance(obj, torch.Tensor):
+            return obj[index]
+        if isinstance(obj, tuple):
+            return tuple(_index_select(v, index) for v in obj)
+        # Handle the case when obj is a dictionary
+        assert isinstance(obj, dict)
+        assert isinstance(index, dict)
+        ret = {}
+        for k, v in obj.items():
+            ret[k] = _index_select(v, index[k])
+        return ret
+
+    return SampledSubgraphImpl(
+        node_pairs=_index_select(subgraph.node_pairs, index),
+        reverse_column_node_ids=subgraph.reverse_column_node_ids,
+        reverse_row_node_ids=subgraph.reverse_row_node_ids,
+        reverse_edge_ids=_index_select(subgraph.reverse_edge_ids, index),
+    )
+
+
 def exclude_edges(
     subgraph: SampledSubgraphImpl,
     edges: Union[
@@ -94,5 +151,39 @@ def exclude_edges(
     SampledSubgraphImpl
         The sampled subgraph with the excluded edges.
     """
-    # TODO(zhenkun): Implement this.
-    raise NotImplementedError
+    assert isinstance(subgraph.node_pairs, tuple) == isinstance(edges, tuple), (
+        "The sampled subgraph and the edges to exclude should be both "
+        "homogeneous or both heterogeneous."
+    )
+    # Three steps to exclude edges:
+    # 1. Convert the node pairs to the reverse ids.
+    # 2. Exclude the edges and get the index of the edges to keep.
+    # 3. Slice the subgraph according to the index.
+    if isinstance(subgraph.node_pairs, tuple):
+        reverse_edges = _to_reverse_ids(
+            subgraph.node_pairs,
+            subgraph.reverse_row_node_ids,
+            subgraph.reverse_column_node_ids,
+        )
+        index = _exclude_homo_edges(reverse_edges, edges)
+        return _slice_subgraph(subgraph, index)
+    else:
+        index = {}
+        for etype, pair in subgraph.node_pairs.items():
+            reverse_row_node_ids = (
+                None
+                if subgraph.reverse_row_node_ids is None
+                else subgraph.reverse_row_node_ids.get(etype[0])
+            )
+            reverse_column_node_ids = (
+                None
+                if subgraph.reverse_column_node_ids is None
+                else subgraph.reverse_column_node_ids.get(etype[2])
+            )
+            reverse_edges = _to_reverse_ids(
+                pair,
+                reverse_row_node_ids,
+                reverse_column_node_ids,
+            )
+            index[etype] = _exclude_homo_edges(reverse_edges, edges.get(etype))
+        return _slice_subgraph(subgraph, index)
--- a/python/dgl/graphbolt/impl/sampled_subgraph_impl.py
+++ b/python/dgl/graphbolt/impl/sampled_subgraph_impl.py
@@ -77,25 +77,24 @@ def _to_reverse_ids(node_pair, reverse_row_node_ids, reverse_column_node_ids):
     return (u, v)
 
 
-def _relabel_homo_edges(lhs_edges, rhs_edges):
-    """Relabel the source nodes and destination nodes independently for two set
-    of edges."""
-    src = torch.cat([lhs_edges[0], rhs_edges[0]])
-    dst = torch.cat([lhs_edges[1], rhs_edges[1]])
-    _, src = torch.unique(src, return_inverse=True)
-    _, dst = torch.unique(dst, return_inverse=True)
-    lhs_edges = (src[: lhs_edges[0].numel()], dst[: lhs_edges[0].numel()])
-    rhs_edges = (src[lhs_edges[0].numel() :], dst[lhs_edges[0].numel() :])
-    return lhs_edges, rhs_edges
+def _relabel_two_arrays(lhs_array, rhs_array):
+    """Relabel two arrays into a consecutive range starting from 0."""
+    concated = torch.cat([lhs_array, rhs_array])
+    _, mapping = torch.unique(concated, return_inverse=True)
+    return mapping[: lhs_array.numel()], mapping[lhs_array.numel() :]
 
 
-def _exclude_homo_edges(lhs_edges, rhs_edges):
-    lhs_edges, rhs_edges = _relabel_homo_edges(lhs_edges, rhs_edges)
-    # Compact edges to integers to use torch.isin to exclude edges
-    dst_range = lhs_edges[0].numel() + rhs_edges[0].numel()
-    lhs_val = lhs_edges[0] * dst_range + lhs_edges[1]
-    rhs_val = rhs_edges[0] * dst_range + rhs_edges[1]
-    mask = ~torch.isin(lhs_val, rhs_val)
+def _exclude_homo_edges(edges, edges_to_exclude):
+    """Return the indices of edges that are not in edges_to_exclude."""
+    # 1. Relabel edges.
+    src, src_to_exclude = _relabel_two_arrays(edges[0], edges_to_exclude[0])
+    dst, dst_to_exclude = _relabel_two_arrays(edges[1], edges_to_exclude[1])
+    # 2. Compact the edges to integers.
+    dst_max_range = dst.numel() + dst_to_exclude.numel()
+    val = src * dst_max_range + dst
+    val_to_exclude = src_to_exclude * dst_max_range + dst_to_exclude
+    # 3. Use torch.isin to get the indices of edges to keep.
+    mask = ~torch.isin(val, val_to_exclude)
     return torch.nonzero(mask, as_tuple=True)[0]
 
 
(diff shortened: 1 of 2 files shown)


czkkkkkk:
@dgl-bot

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -94,5 +151,39 @@ def exclude_edges(
...
         The sampled subgraph with the excluded edges.
     """
-    # TODO(zhenkun): Implement this.
-    raise NotImplementedError
+    assert isinstance(subgraph.node_pairs, tuple) == isinstance(edges, tuple), (
+        "The sampled subgraph and the edges to exclude should be both "
+        "homogeneous or both heterogeneous."
+    )
+    # Three steps to exclude edges:
+    # 1. Convert the node pairs to the reverse ids.
```
frozenbugs, line 159:
```suggestion
    # 1. Convert the node pairs to the original ids if they are compacted.
```

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -68,6 +68,63 @@ def __post_init__(self):
...
+    _, src = torch.unique(src, return_inverse=True)
+    _, dst = torch.unique(dst, return_inverse=True)
+    lhs_edges = (src[: lhs_edges[0].numel()], dst[: lhs_edges[0].numel()])
+    rhs_edges = (src[lhs_edges[0].numel() :], dst[lhs_edges[0].numel() :])
+    return lhs_edges, rhs_edges
+
+
+def _exclude_homo_edges(lhs_edges, rhs_edges):
+    lhs_edges, rhs_edges = _relabel_homo_edges(lhs_edges, rhs_edges)
+    # Compact edges to integers to use torch.isin to exclude edges
```
frozenbugs, line 94:
```suggestion
    # Compact edges to integers to use torch.isin to exclude edges.
```

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -68,6 +68,63 @@ def __post_init__(self):
...
+    """Slice the subgraph according to the index."""
+
+    def _index_select(obj, index):
+        if obj is None:
+            return None
+        if isinstance(obj, torch.Tensor):
+            return obj[index]
+        if isinstance(obj, tuple):
+            return tuple(_index_select(v, index) for v in obj)
+        # Handle the case when obj is a dictionary
```
frozenbugs, line 112: ditto

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -68,6 +68,63 @@ def __post_init__(self):
...
+    dst = torch.cat([lhs_edges[1], rhs_edges[1]])
+    _, src = torch.unique(src, return_inverse=True)
+    _, dst = torch.unique(dst, return_inverse=True)
+    lhs_edges = (src[: lhs_edges[0].numel()], dst[: lhs_edges[0].numel()])
+    rhs_edges = (src[lhs_edges[0].numel() :], dst[lhs_edges[0].numel() :])
+    return lhs_edges, rhs_edges
+
+
+def _exclude_homo_edges(lhs_edges, rhs_edges):
+    lhs_edges, rhs_edges = _relabel_homo_edges(lhs_edges, rhs_edges)
```
frozenbugs, line 93: Why do we need to relabel before we exclude? It is a little bit confusing.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -68,6 +68,63 @@ def __post_init__(self):
...
+            return tuple(_index_select(v, index) for v in obj)
+        # Handle the case when obj is a dictionary
+        assert isinstance(obj, dict)
+        assert isinstance(index, dict)
+        ret = {}
+        for k, v in obj.items():
+            ret[k] = _index_select(v, index[k])
+        return ret
+
+    return SampledSubgraphImpl(
```
frozenbugs, line 120: Is the returned SampledSubgraphImpl compacted or not or depends on the input SampledSubgraphImpl?

tests/python/pytorch/graphbolt/impl/test_sampled_subgraph_impl.py
```diff
@@ -0,0 +1,121 @@
...
+            assert key in rhs
+            _assert_container_equal(value, rhs[key])
+
+
+@pytest.mark.parametrize("is_reverse", [True, False])
+def test_exclude_edges_homo(is_reverse):
+    node_pairs = (torch.tensor([0, 2, 3]), torch.tensor([1, 4, 2]))
+    if is_reverse:
+        reverse_row_node_ids = torch.tensor([10, 15, 11, 24, 9])
+        reverse_column_node_ids = torch.tensor([10, 15, 11, 24, 9])
```
frozenbugs, line 31: @peizhou001 do we compact row and column respectively or altogether?

tests/python/pytorch/graphbolt/impl/test_sampled_subgraph_impl.py
```diff
@@ -0,0 +1,121 @@
...
+    subgraph = SampledSubgraphImpl(
+        node_pairs,
+        reverse_column_node_ids,
+        reverse_row_node_ids,
+        reverse_edge_ids,
+    )
+    if is_reverse:
+        excluded_edges = (torch.tensor([11]), torch.tensor([9]))
+    else:
+        excluded_edges = (torch.tensor([2]), torch.tensor([4]))
```
frozenbugs, line 44: by any chance user will use compacted edge id?

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -68,6 +68,63 @@ def __post_init__(self):
...
+def _to_reverse_ids(node_pair, reverse_row_node_ids, reverse_column_node_ids):
+    u, v = node_pair
+    if reverse_row_node_ids is not None:
+        u = reverse_row_node_ids[u]
+    if reverse_column_node_ids is not None:
+        v = reverse_column_node_ids[v]
+    return (u, v)
+
+
+def _relabel_homo_edges(lhs_edges, rhs_edges):
```
BarclayII, line 80: Why do we have two sets of edges during relabeling?

peizhou001: According to previous discussion, some sampling algorithm natually have column compressed from a view of "columnwise pick", but row is not compressed.

tests/python/pytorch/graphbolt/impl/test_sampled_subgraph_impl.py
peizhou001: We compact them together. But according to our previous discussion, row and column may not be both compressed in some algorithm.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -94,5 +151,39 @@ def exclude_edges(
     SampledSubgraphImpl
         The sampled subgraph with the excluded edges.
```
peizhou001, line 152: without?

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -94,5 +151,39 @@ def exclude_edges(
...
     """
-    # TODO(zhenkun): Implement this.
-    raise NotImplementedError
+    assert isinstance(subgraph.node_pairs, tuple) == isinstance(edges, tuple), (
+        "The sampled subgraph and the edges to exclude should be both "
+        "homogeneous or both heterogeneous."
```
peizhou001, line 156: Remove the second both?

peizhou001, line 156: Remove the second both?

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -68,6 +68,63 @@ def __post_init__(self):
...
+    lhs_edges = (src[: lhs_edges[0].numel()], dst[: lhs_edges[0].numel()])
+    rhs_edges = (src[lhs_edges[0].numel() :], dst[lhs_edges[0].numel() :])
+    return lhs_edges, rhs_edges
+
+
+def _exclude_homo_edges(lhs_edges, rhs_edges):
+    lhs_edges, rhs_edges = _relabel_homo_edges(lhs_edges, rhs_edges)
+    # Compact edges to integers to use torch.isin to exclude edges
+    dst_range = lhs_edges[0].numel() + rhs_edges[0].numel()
+    lhs_val = lhs_edges[0] * dst_range + lhs_edges[1]
```
peizhou001, line 96: Can I understand first element in tuple as high number and the second as low number with  `dst_range` as radix?
If this is the case, will it cause error because both` lhs_edges[1]` and `rhs_edges[1]` can be larger than `dst_range`?

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -68,6 +68,63 @@ def __post_init__(self):
...
+    rhs_val = rhs_edges[0] * dst_range + rhs_edges[1]
+    mask = ~torch.isin(lhs_val, rhs_val)
+    return torch.nonzero(mask, as_tuple=True)[0]
```
peizhou001, line 99: Another choice it to return mask directly and use `[mask]` for filter. Both methonds all okay for me, you can choose one you perfer.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
peizhou001: I'm also a little confused as both ids are already converted to non-compacted format. And `unique `should be a relatively costly op.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -68,6 +68,63 @@ def __post_init__(self):
             ), "Nodes in pairs should be of type torch.Tensor."
 
 
+def _to_reverse_ids(node_pair, reverse_row_node_ids, reverse_column_node_ids):
```
peizhou001, line 71: ` reverse_row_node_ids` is not the compacted counterpart of  `node_pair[0]`, same for `reverse_column_node_ids`. For example:

```
origin_node_pair = ([100, 50, 50, 100], ...)
after compact 
node_pair = ([1, 0, 0, 1], ...)
reverese_row_node_ids = [50, 100], means 50 -> 0 and 100 -> 1.
```

peizhou001 (changes requested):

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
czkkkkkk: This is also my understanding. The code below is to get the `origin_node_pair` of your example. It works because `reverse_row_node_ids[node_pair] = [100, 50, 50, 100]`.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
czkkkkkk: This to to map the node ids of two set of edges to a consecutive range from 0. I concatenate them in this function.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
czkkkkkk: This is to convert each edge to a single value to apply `torch.isin`. Another solution is to compact an edge `(u, v)` to `(u << 32 + v)`. But it does not work if some node id exceed `2^32`. Therefore I further compact the nodes ids of all edges to make their range smaller.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
czkkkkkk: They would not be larger than `lhs_edges[0].numel() + rhs_edges[0].numel()` since they are compacted.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
czkkkkkk: Yeah. I think it does not affect the performance.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
czkkkkkk: It depends on the input subgraph. I will add explanation about this.

tests/python/pytorch/graphbolt/impl/test_sampled_subgraph_impl.py
czkkkkkk: My implementation does not assume they are compacted together. I will add tests on the cases of single compaction.

tests/python/pytorch/graphbolt/impl/test_sampled_subgraph_impl.py
czkkkkkk: Yeah. This should be removed.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
BarclayII: I see.  I would suggest (1) rename `lhs_edges` and `rhs_edges` to more informative names like `node_id_pairs`, `node_id_pairs_to_exclude`.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -68,6 +68,63 @@ def __post_init__(self):
...
+
+def _relabel_homo_edges(lhs_edges, rhs_edges):
+    """Relabel the source nodes and destination nodes independently for two set
+    of edges."""
+    src = torch.cat([lhs_edges[0], rhs_edges[0]])
+    dst = torch.cat([lhs_edges[1], rhs_edges[1]])
+    _, src = torch.unique(src, return_inverse=True)
+    _, dst = torch.unique(dst, return_inverse=True)
+    lhs_edges = (src[: lhs_edges[0].numel()], dst[: lhs_edges[0].numel()])
+    rhs_edges = (src[lhs_edges[0].numel() :], dst[lhs_edges[0].numel() :])
```
BarclayII, line 88: Try to avoid reusing variables for different purposes, e.g. in L85-86, `src` could be something like `src_relabeled`.

tests/python/pytorch/graphbolt/impl/test_sampled_subgraph_impl.py
czkkkkkk: I found the `if` branch cannot be removed in the tests. For the case of `is_reverse == False`, it means the subgraph is not compacted and the edge `(2, 4)` means an edge between original ids.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
czkkkkkk: Good point. I use `edges` and `edges_to_exclude` now.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
czkkkkkk: I rewrote this function to make it only relabel two arrays instead of two edge sets.

frozenbugs (approved):

tests/python/pytorch/graphbolt/impl/test_sampled_subgraph_impl.py
```diff
@@ -0,0 +1,136 @@
...
+    reverse_edge_ids = torch.Tensor([5, 9, 10])
+    subgraph = SampledSubgraphImpl(
+        node_pairs,
+        reverse_column_node_ids,
+        reverse_row_node_ids,
+        reverse_edge_ids,
+    )
+    src_to_exclude = torch.tensor([11]) if reverse_row else torch.tensor([2])
+    dst_to_exclude = torch.tensor([9]) if reverse_column else torch.tensor([4])
+    excluded_edges = (src_to_exclude, dst_to_exclude)
```
frozenbugs, line 48: For excluded_edges, should them all be original hetero id?

tests/python/pytorch/graphbolt/impl/test_sampled_subgraph_impl.py
```diff
@@ -0,0 +1,136 @@
...
+    result = exclude_edges(subgraph, excluded_edges)
```
frozenbugs, line 49:
```suggestion
    result = exclude_edges(subgraph, edges_to_exclude)
```

PR #6005, opened 2023-07-15
[Graphbolt] Gpu cache feature store

mfbalin:
## Description
Graphbolt GPU feature is implemented. I think a generic FeatureStore that can accept any kind of Features is needed so that the user can cache some of the features, such as features but not labels.

### DIFF
--- a/python/dgl/graphbolt/feature_store.py
+++ b/python/dgl/graphbolt/feature_store.py
@@ -12,9 +12,35 @@ class Feature:
         pass
 
     def read(self, ids: torch.Tensor = None):
+        """Read from the feature.
+
+        Parameters
+        ----------
+        ids : torch.Tensor, optional
+            The index of the feature. If specified, only the specified indices
+            of the feature are read. If None, the entire feature is returned.
+
+        Returns
+        -------
+        torch.Tensor
+            The read feature.
+        """
         pass
 
     def update(self, value: torch.Tensor, ids: torch.Tensor = None):
+        """Update the feature.
+
+        Parameters
+        ----------
+        value : torch.Tensor
+            The updated value of the feature.
+        ids : torch.Tensor, optional
+            The indices of the feature to update. If specified, only the
+            specified indices of the feature will be updated. For the feature,
+            the `ids[i]` row is updated to `value[i]`. So the indices and value
+            must have the same length. If None, the entire feature will be
+            updated.
+        """
         pass
 
 
--- a/python/dgl/graphbolt/feature_store.py
+++ b/python/dgl/graphbolt/feature_store.py
@@ -2,7 +2,20 @@
 
 import torch
 
-__all__ = ["FeatureStore"]
+__all__ = ["Feature", "FeatureStore"]
+
+
+class Feature:
+    r"""Base class for feature."""
+
+    def __init__(self):
+        pass
+
+    def read(self, ids: torch.Tensor = None):
+        pass
+
+    def update(self, value: torch.Tensor, ids: torch.Tensor = None):
+        pass
 
 
 class FeatureStore:
--- a/python/dgl/graphbolt/feature_store.py
+++ b/python/dgl/graphbolt/feature_store.py
@@ -25,7 +25,7 @@ class Feature:
         torch.Tensor
             The read feature.
         """
-        pass
+        raise NotImplementedError
 
     def update(self, value: torch.Tensor, ids: torch.Tensor = None):
         """Update the feature.
@@ -41,7 +41,7 @@ class Feature:
             must have the same length. If None, the entire feature will be
             updated.
         """
-        pass
+        raise NotImplementedError
 
 
 class FeatureStore:
--- a/python/dgl/graphbolt/feature_store.py
+++ b/python/dgl/graphbolt/feature_store.py
@@ -2,46 +2,7 @@
 
 import torch
 
-__all__ = ["Feature", "FeatureStore"]
-
-
-class Feature:
-    r"""Base class for feature."""
-
-    def __init__(self):
-        pass
-
-    def read(self, ids: torch.Tensor = None):
-        """Read from the feature.
-
-        Parameters
-        ----------
-        ids : torch.Tensor, optional
-            The index of the feature. If specified, only the specified indices
-            of the feature are read. If None, the entire feature is returned.
-
-        Returns
-        -------
-        torch.Tensor
-            The read feature.
-        """
-        raise NotImplementedError
-
-    def update(self, value: torch.Tensor, ids: torch.Tensor = None):
-        """Update the feature.
-
-        Parameters
-        ----------
-        value : torch.Tensor
-            The updated value of the feature.
-        ids : torch.Tensor, optional
-            The indices of the feature to update. If specified, only the
-            specified indices of the feature will be updated. For the feature,
-            the `ids[i]` row is updated to `value[i]`. So the indices and value
-            must have the same length. If None, the entire feature will be
-            updated.
-        """
-        raise NotImplementedError
+__all__ = ["FeatureStore"]
 
 
 class FeatureStore:
--- a/python/dgl/graphbolt/impl/gpu_cached_feature_store.py
+++ b/python/dgl/graphbolt/impl/gpu_cached_feature_store.py
@@ -1,4 +1,4 @@
-"""Torch-based feature store for GraphBolt."""
+"""GPUcached feature store for GraphBolt."""
 import torch
 
 from dgl.cuda import GPUCache
@@ -44,6 +44,7 @@ class GPUCachedFeatureStore(FeatureStore):
             f"but got {type(fallback_store)}."
         )
         self._fallback_store = fallback_store
+        # we query the underlying feature store to learn the feature dimension
         self.cache_size = cache_size
         feat0 = fallback_store.read(torch.tensor([0]))
         self.item_shape = feat0.shape[1:]
--- a/python/dgl/graphbolt/impl/gpu_cached_feature_store.py
+++ b/python/dgl/graphbolt/impl/gpu_cached_feature_store.py
@@ -25,11 +25,12 @@ class GPUCachedFeatureStore(FeatureStore):
         Examples
         --------
         >>> import torch
-        >>> torch_feat = torch.arange(0, 5)
+        >>> torch_feat = torch.arange(0, 8)
+        >>> cache_size = 5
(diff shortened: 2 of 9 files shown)


src/runtime/cuda/gpu_cache.cu
```diff
@@ -0,0 +1,171 @@
...
+        *rv = GpuCacheRef32(
+            std::make_shared<GpuCache<unsigned int>>(num_items, num_feats));
+      else
+        *rv = GpuCacheRef64(std::make_shared<GpuCache<long long>>(  // NOLINT
+            num_items, num_feats));
+    });
+
+DGL_REGISTER_GLOBAL("cuda._CAPI_DGLGpuCacheQuery")
+    .set_body([](DGLArgs args, DGLRetValue *rv) {
+      IdArray keys = args[1];
```
nv-dlasalle, line 129: Do we need a check here that the keys are GPU accessible (GPU memory or pinned)? Or alternatively in the Query() function above?

src/runtime/cuda/gpu_cache.cu
```diff
@@ -0,0 +1,171 @@
...
+        ret.push_back(Value(MakeValue(std::get<2>(result))));
+      }
+
+      *rv = ret;
+    });
+
+DGL_REGISTER_GLOBAL("cuda._CAPI_DGLGpuCacheReplace")
+    .set_body([](DGLArgs args, DGLRetValue *rv) {
+      IdArray keys = args[1];
+      NDArray values = args[2];
```
nv-dlasalle, line 154: Do we need a check here that the keys and values are GPU accessible (GPU memory or pinned)? Or alternatively in the Replace() function above?

nv-dlasalle (approved):
Looks good to me. I would request though we add some checks so users will be notified by DGL that their input tensors are on the wrong device, rather than receiving an error from HugeCTR or CUDA.

src/runtime/cuda/gpu_cache.cu
mfbalin: Sorry about leaving changes from #4341 here. Only the graphbolt python file is added in this PR. I will address these in #4341.

mfbalin: Committed the fixes in #4341.

mfbalin:
@BarclayII Do you want to take a look as well?

python/dgl/graphbolt/impl/gpu_cached_feature_store.py
```diff
@@ -0,0 +1,106 @@
...
+
+__all__ = ["GPUCachedFeatureStore"]
+
+
+class GPUCachedFeatureStore(FeatureStore):
+    r"""GPU cached feature store wrapping a fallback feature store."""
+
+    def __init__(self, fallback_store: FeatureStore, cache_size: int):
+        """Initialize GPU cached feature store with a given fallback.
+        Places the GPU cache to torch.cuda.current_device().
```
frozenbugs, line 16: How can user set which device to load to? Will different GPUs share the same cache?

mfbalin: `torch.cuda.set_device(device)` before constructing the gpu cached feature store will ininitialize the GPU cache at the currently set device. Do we want the capability to pass the device, and if it is `None`, we fall back to the current_device?

mfbalin: I think `dgl.cuda.GPUCache` doesn't have an option to take in a device argument when constructing the GPU cache. It simply constructs it at the currently set CUDA device in the process, i.e. `CUDA_CALL(cudaGetDevice(&cuda_device));`

mfbalin: A single GPUCache instance belongs to a single GPU only. GPUs can't share the same cache. A wrapper for a multi-GPU cache could be implemented though.

frozenbugs: I see, we can focus on single GPU cases for this PR, for multiple GPU use case, let's re-visit. Please put a comment on the header of the file to clarify it is designed for Single GPU usage.

frozenbugs: Currently each GPU have a dataloader process, if torch.cuda.current_device() can recognize the device assigned to each process, then it should work properly.

mfbalin: https://github.com/dmlc/dgl/blob/b6f5ba9a809fcd2e5b2c440f538c1ccc965a9e59/examples/multigpu/node_classification_sage.py#L242 sets the default device so it will just work out of the box.

frozenbugs:
@mfbalin Are you still working on this PR?

frozenbugs:
I see, we already have UVA work on our todolist, will kick of the work soon.

mfbalin:
@frozenbugs I see that there is a feature store and a feature now. Feature store can hold multiple features. But there is no feature class to derive from. I need this functionality to add GPUCachedFeature. We also need to have a Generic FeatureStore implementation that can have different kinds of features inside of it so that some of the features can be the cached kind.

python/dgl/graphbolt/feature_store.py
```diff
@@ -2,7 +2,46 @@
 
 import torch
 
-__all__ = ["FeatureStore"]
+__all__ = ["Feature", "FeatureStore"]
```
frozenbugs, line 5: Can you do it in a separated PR?

frozenbugs: I am doing it in https://github.com/dmlc/dgl/pull/6204, feel free to remove.

tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
frozenbugs, line 1: create a test_gpu_cached_feature_store.py and move the test there.

frozenbugs, line 1: create a test_gpu_cached_feature.py and move the test there.

python/dgl/graphbolt/impl/gpu_cached_feature_store.py
frozenbugs, line 1: rename to python/dgl/graphbolt/impl/gpu_cached_feature.py

python/dgl/graphbolt/impl/gpu_cached_feature_store.py
```diff
@@ -0,0 +1,112 @@
...
+        tensor([0, 1, 2])
+        >>> feature.update(torch.ones(3, dtype=torch.long),
+        ... torch.tensor([0, 1, 2]))
+        >>> feature.read(torch.tensor([0, 1, 2, 3]))
+        tensor([1, 1, 1, 3])
+        """
+        super(GPUCachedFeature, self).__init__()
+        assert isinstance(fallback_feature, Feature), (
+            f"fallback_feature must be FeatureStore, "
+            f"but got {type(fallback_feature)}."
```
frozenbugs, line 44:
```suggestion
            f"The fallback_feature must be an instance of Feature, but got "
            f"{type(fallback_feature)}."
```

python/dgl/graphbolt/impl/gpu_cached_feature_store.py
```diff
@@ -0,0 +1,112 @@
...
+        )
+        self._fallback_feature = fallback_feature
+        # we query the underlying feature store to learn the feature dimension
+        self.cache_size = cache_size
```
frozenbugs, line 48:
```suggestion
        self.cache_size = cache_size
        # Fetching the feature dimension from the underlying feature.
```

frozenbugs (approved):

tests/python/pytorch/graphbolt/impl/test_gpu_cached_feature.py
```diff
@@ -0,0 +1,47 @@
...
+    reason="GPUCachedFeature requires a GPU.",
+)
+def test_gpu_cached_feature():
+    a = torch.tensor([1, 2, 3]).to("cuda").float()
+    b = torch.tensor([[1, 2, 3], [4, 5, 6]]).to("cuda").float()
+
+    feat_store_a = gb.GPUCachedFeature(gb.TorchBasedFeature(a), 2)
+    feat_store_b = gb.GPUCachedFeature(gb.TorchBasedFeature(b), 1)
+
+    assert torch.equal(feat_store_a.read(), a)
```
frozenbugs, line 21:
```suggestion
    # Test read the entire feature.
    assert torch.equal(feat_store_a.read(), a)
```

tests/python/pytorch/graphbolt/impl/test_gpu_cached_feature.py
```diff
@@ -0,0 +1,47 @@
...
+    assert torch.equal(feat_store_b.read(), b)
+    assert torch.equal(
```
frozenbugs, line 23:
```suggestion

    # Test read with ids.
    assert torch.equal(
```

tests/python/pytorch/graphbolt/impl/test_gpu_cached_feature.py
```diff
@@ -0,0 +1,47 @@
...
+    )
+    assert torch.equal(
+        feat_store_a.read(torch.tensor([1, 1]).to("cuda")),
+        torch.tensor([2.0, 2.0]).to("cuda"),
+    )
+    assert torch.equal(
+        feat_store_b.read(torch.tensor([1]).to("cuda")),
+        torch.tensor([[4.0, 5.0, 6.0]]).to("cuda"),
+    )
+    feat_store_a.update(
```
frozenbugs, line 35:
```suggestion

    # Test update the entire feature.
    feat_store_a.update(
```

tests/python/pytorch/graphbolt/impl/test_gpu_cached_feature.py
```diff
@@ -0,0 +1,47 @@
...
+        torch.tensor([0.0, 1.0, 2.0]).to("cuda"),
+        torch.tensor([0, 1, 2]).to("cuda"),
+    )
+    assert torch.equal(
+        feat_store_a.read(), torch.tensor([0.0, 1.0, 2.0]).to("cuda")
+    )
+    feat_store_a.update(
```
frozenbugs, line 42:
```suggestion

    # Test update with ids.
    feat_store_a.update(
```

PR #6017, opened 2023-07-18
[Graphbolt] Modify `preprocess_ondisk_dataset` function.

keli-wen:
## Description

Let the function return a raw `str` type instead of `pathlib.Path`.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/impl/ondisk_dataset.py
+++ b/python/dgl/graphbolt/impl/ondisk_dataset.py
@@ -137,9 +137,9 @@ def preprocess_ondisk_dataset(input_config_path: str) -> str:
         ):
             # Always save the feature in numpy format.
             out_feature["format"] = "numpy"
-            out_feature["path"] = processed_dir_prefix / feature[
-                "path"
-            ].replace("pt", "npy")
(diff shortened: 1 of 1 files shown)


frozenbugs (approved):

python/dgl/graphbolt/impl/ondisk_dataset.py
```diff
@@ -209,7 +209,7 @@ def preprocess_ondisk_dataset(input_config_path: str) -> str:
     with open(output_config_path, "w") as f:
         yaml.dump(output_config, f)
     print("Finish preprocessing the on-disk dataset.")
-    return output_config_path
+    return str(output_config_path)
```
Rhett-Ying, line 212: this only print the path of yaml file as string, but does not print the paths within the yaml file content. Please add dedicated test to make sure all the paths in yaml file is printed as raw string instead of `Path` object.

keli-wen: I understand what you mean, I have now changed all `Path` to `str`, but I have a question: why the paths I saved before are all normal.

PR #5717, opened 2023-05-19
[GraphBolt] Sample etype neighbors c code

peizhou001:
## Description
`SampleEtypeNeighbors` API  C side code.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [ ] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [ ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

PR #6030, opened 2023-07-20
[Graphbolt] Add `cat` optimization for UniformPick

keli-wen:
## Description

We've optimized the `SampleNeighborsImpl` function in `csc_sampling_graph.cc`. This PR primarily resolves this [Issue#5817 Optimize CSC neighbor sampler performance.](https://github.com/dmlc/dgl/issues/5817)

The original algorithm, when conducting multithreaded Neighbor Sampling, would place all sampled results into a general vector of Tensor. We want to optimize this part.

### Problem Analysis

Let's consider a scenario where we have $n$ `Tensors`(same size) and $m$ `threads`, and we need to concatenate all `Tensors` into one large `Tensor`. We consider two different strategies:

- Strategy **A**: Each `thread` uses `torch::cat` to combine up to $\frac{n}{m}$ `Tensors` into a `big Tensor`, then `torch::cat` is used again to combine the $m$ `big Tensors` into the `result`.
- Strategy **B**: Each `thread` only returns a `std::vector<Tensor>`, and in the main program, the results from each thread are combined into a `std::vector<Tensor> all_tensors`. Finally, `torch::cat` is used to combine the `all_tensors` with $n$ elements into the `result`. (currently strategy)

A necessary condition for Strategy **A** to outperform Strategy **B** is:

- **[Assumption 1]**: The performance of `torch::cat` when merging **fewer large-Tensors** is superior to that of merging **many small-Tensors**. We conducted a simple test to verify this: (where `small` represents a group of small tensors and `big` represents a group of large tensors)

| `small_num` | `small_size` | `big_num` | `big_size` | `device` |  `small`  |   `big`   |     `gap`     |
| :---------: | :----------: | :-------: | :--------: | :------: | :-------: | :-------: | :-----------: |
|   $10000$   |    $2000$    |   $20$    | $1000000$  |  `CPU`   | $1.1438s$ | $0.3369s$ |  $3.4\times$  |
|   $10000$   |    $2000$    |   $200$   |  $100000$  |  `CPU`   | $1.1321s$ | $0.3941s$ |  $2.9\times$  |
|   $10000$   |    $2000$    |   $20$    | $1000000$  |  `GPU`   | $0.1146s$ | $0.0020s$ | $57.3\times$  |
|   $10000$   |    $2000$    |   $200$   |  $100000$  |  `GPU`   | $0.1133s$ | $0.0037s$ | $30.62\times$ |

We found **[Assumption 1]** to be true, so we proceeded with the next optimization step.

Based on our performance tests, we observed that **if the results obtained by each thread are first stored within the thread and `cat` is performed once within the thread to obtain a smaller Tensor, and finally `cat` is performed externally on the Tensors obtained from all threads, the efficiency will be higher.** In other words, Strategy A is superior. The detailed performance test experiment is as follows: 

### Time Performance Test

exp_cat

exp_cat_time_table

We found that except in very few cases, this optimization could bring about a performance improvement of around $2$ times. **For red cells in Graph3**, when the data is too large, it is possible that multiple `cat` operations within a thread may lead to performance degradation.

### Memory Usage Test

We performed memory tests using `Massif` in `Valgrind`, and for the sake of test accuracy, the entire test was conducted in a C++ environment, looping each test 5 times. (Massif is a tool in Valgrind, **used to measure the amount of heap memory used by an application**)

exp_cat_memo

exp_cat_memo_table

We only performed memory tests on two commonly used graphs. **For Graph1**, there was also an improvement in terms of memory. **For Graph2**, since there were only $10$ sampled nodes, our strategy did not significantly impact it. 

Therefore, this optimization is effective.

### DIFF
--- a/graphbolt/src/csc_sampling_graph.cc
+++ b/graphbolt/src/csc_sampling_graph.cc
@@ -141,50 +141,64 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
   // If true, perform sampling for each edge type of each node, otherwise just
   // sample once for each node with no regard of edge types.
   bool consider_etype = (fanouts.size() > 1);
-  std::vector<torch::Tensor> picked_neighbors_per_node(num_nodes);
+  const int64_t num_threads = torch::get_num_threads();
+  std::vector<torch::Tensor> picked_neighbors_per_thread(num_threads);
   torch::Tensor num_picked_neighbors_per_node =
       torch::zeros({num_nodes + 1}, indptr_.options());
 
+  // Calculate GrainSize for parallel_for.
+  int64_t grain_size = (num_nodes + num_threads - 1) / num_threads;
   AT_DISPATCH_INTEGRAL_TYPES(
       indptr_.scalar_type(), "parallel_for", ([&] {
-        torch::parallel_for(0, num_nodes, 32, [&](scalar_t b, scalar_t e) {
-          const scalar_t* indptr_data = indptr_.data_ptr<scalar_t>();
-          for (scalar_t i = b; i < e; ++i) {
-            const auto nid = nodes[i].item<int64_t>();
-            TORCH_CHECK(
-                nid >= 0 && nid < NumNodes(),
-                "The seed nodes' IDs should fall within the range of the "
-                "graph's node IDs.");
-            const auto offset = indptr_data[nid];
-            const auto num_neighbors = indptr_data[nid + 1] - offset;
+        torch::parallel_for(
+            0, num_nodes, grain_size, [&](scalar_t begin, scalar_t end) {
+              const auto indptr_options = indptr_.options();
+              const scalar_t* indptr_data = indptr_.data_ptr<scalar_t>();
+              // Get current thread id.
+              auto t_id = torch::get_thread_num();
+              int64_t cur_grain_size = end - begin;
+              std::vector<torch::Tensor> picked_neighbors_cur_thread(
+                  cur_grain_size);
 
-            if (num_neighbors == 0) {
-              // To avoid crashing during concatenation in the master thread,
-              // initializing with empty tensors.
-              picked_neighbors_per_node[i] =
-                  torch::tensor({}, indptr_.options());
-              continue;
-            }
+              for (scalar_t i = begin; i < end; ++i) {
+                const auto nid = nodes[i].item<int64_t>();
+                TORCH_CHECK(
+                    nid >= 0 && nid < NumNodes(),
+                    "The seed nodes' IDs should fall within the range of the "
+                    "graph's node IDs.");
+                const auto offset = indptr_data[nid];
+                const auto num_neighbors = indptr_data[nid + 1] - offset;
 
-            if (consider_etype) {
-              picked_neighbors_per_node[i] = PickByEtype(
-                  offset, num_neighbors, fanouts, replace, indptr_.options(),
-                  type_per_edge_.value(), probs_or_mask, args);
-            } else {
-              picked_neighbors_per_node[i] = Pick(
-                  offset, num_neighbors, fanouts[0], replace, indptr_.options(),
-                  probs_or_mask, args);
-            }
-            num_picked_neighbors_per_node[i + 1] =
-                picked_neighbors_per_node[i].size(0);
-          }
-        });  // End of the thread.
+                if (num_neighbors == 0) {
+                  // Initialization is performed here because all tensors will
+                  // be concatenated in the current thread, and having an
+                  // undefined tensor during concatenation can result in a
+                  // crash.
+                  picked_neighbors_cur_thread[i - begin] =
+                      torch::tensor({}, indptr_options);
+                  continue;
+                }
+
+                if (consider_etype) {
+                  picked_neighbors_cur_thread[i - begin] = PickByEtype(
+                      offset, num_neighbors, fanouts, replace, indptr_options,
+                      type_per_edge_.value(), probs_or_mask, args);
+                } else {
+                  picked_neighbors_cur_thread[i - begin] = Pick(
+                      offset, num_neighbors, fanouts[0], replace,
+                      indptr_options, probs_or_mask, args);
+                }
+                num_picked_neighbors_per_node[i + 1] =
+                    picked_neighbors_cur_thread[i - begin].size(0);
+              }
+              picked_neighbors_per_thread[t_id] =
+                  torch::cat(picked_neighbors_cur_thread);
+            });  // End of parallel_for.
       }));
-
   torch::Tensor subgraph_indptr =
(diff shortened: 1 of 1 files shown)


frozenbugs:
What does neighbors_per_etype mean?

keli-wen:
> What does neighbors_per_etype mean?

According to the generate code:

```py
def random_graph_with_fixed_neighbors(num_nodes,
                                      neighbors_per_etype,
                                      num_ntypes,
                                      num_etypes):
    neighbors_per_node = num_etypes * neighbors_per_etype
    csc_indptr = torch.arange((num_nodes+1)) * neighbors_per_node
    num_edges = num_nodes * neighbors_per_node
    indices = torch.randint(0, num_nodes, (num_edges,))
    type_per_edge = torch.arange(num_etypes).repeat_interleave(
        neighbors_per_etype
    ).repeat(num_nodes)

    node_type_offset = torch.sort(
        torch.randint(0, num_nodes, (num_ntypes + 1,))
    )[0]
    node_type_offset[0] = 0
    node_type_offset[-1] = num_nodes

    # Add the metadata to the graph.
    ntypes = { f"n{i}" : i for i in range(num_ntypes) }
    etypes = {
        (f"n{(i % num_ntypes) + 1}", f"e{i}", f"n{(i // num_etypes) + 1}") : i
            for i in range(num_etypes)
    }
    metadata = gb.GraphMetadata(ntypes, etypes)

    return gb.from_csc(csc_indptr, indices, node_type_offset, type_per_edge, None, metadata)
```

`neighbors_per_etype` denote the number of neighbors of each edge type of the node.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -141,50 +141,64 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
...
-                "graph's node IDs.");
-            const auto offset = indptr_data[nid];
-            const auto num_neighbors = indptr_data[nid + 1] - offset;
+        torch::parallel_for(
+            0, num_nodes, grain_size, [&](scalar_t begin, scalar_t end) {
+              const auto indptr_options = indptr_.options();
+              const scalar_t* indptr_data = indptr_.data_ptr<scalar_t>();
+              // Get current thread id.
+              auto t_id = torch::get_thread_num();
+              int64_t cur_grain_size = end - begin;
```
frozenbugs, line 159:
```suggestion
              int64_t local_grain_size = end - begin;
```

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -141,50 +141,64 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
...
+              std::vector<torch::Tensor> picked_neighbors_cur_thread(
```
frozenbugs, line 160:
```suggestion
              std::vector<torch::Tensor> local_picked_neighbors(
```

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -141,50 +141,64 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
...
-            num_picked_neighbors_per_node[i + 1] =
-                picked_neighbors_per_node[i].size(0);
-          }
-        });  // End of the thread.
-      }));
+                if (num_neighbors == 0) {
+                  // Initialization is performed here because all tensors will
+                  // be concatenated in the current thread, and having an
+                  // undefined tensor during concatenation can result in a
+                  // crash.
```
frozenbugs, line 176:
```suggestion
                  // To avoid crashing during concatenation in the master thread,
                  // initializing with empty tensors.
```

frozenbugs (approved):

graphbolt/src/csc_sampling_graph.cc
keli-wen: done

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -141,50 +141,64 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
...
   // sample once for each node with no regard of edge types.
   bool consider_etype = (fanouts.size() > 1);
-  std::vector<torch::Tensor> picked_neighbors_per_node(num_nodes);
+  const int64_t num_threads = torch::get_num_threads();
+  std::vector<torch::Tensor> picked_neighbors_per_thread(num_threads);
   torch::Tensor num_picked_neighbors_per_node =
       torch::zeros({num_nodes + 1}, indptr_.options());
 
+  // Calculate GrainSize for parallel_for.
+  int64_t grain_size = (num_nodes + num_threads - 1) / num_threads;
```
peizhou001, line 150: Change `grain_size` to a default value.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -141,50 +141,64 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
...
-                "The seed nodes' IDs should fall within the range of the "
-                "graph's node IDs.");
-            const auto offset = indptr_data[nid];
-            const auto num_neighbors = indptr_data[nid + 1] - offset;
+        torch::parallel_for(
+            0, num_nodes, grain_size, [&](scalar_t begin, scalar_t end) {
+              const auto indptr_options = indptr_.options();
+              const scalar_t* indptr_data = indptr_.data_ptr<scalar_t>();
+              // Get current thread id.
+              auto t_id = torch::get_thread_num();
```
peizhou001, line 158: Use `thead_id` to follow your naming convention.

graphbolt/src/csc_sampling_graph.cc
frozenbugs: This is not addressed.

mfbalin:
So the speedup this PR is achieving over the existing implementation may be due to the possibility that `torch.cat` doesn't have a parallel implementation. Profiling the code would reveal whether this is the case. Otherwise, doing two levels of `torch.cat` operations only increases the work. Each edge is concatenated two times.

peizhou001:
> 

That's true, and the speedup is not very much as you can see in the diagram. According to our test, it performs better when the data volume is small, maybe because cat huge tensors can leverage bulk copy or vector instruction (just suspect)? so it is a tradeoff to decide whether merge it. For rollout cat, it could be a backup improvement for future.

frozenbugs:
Hi @mfbalin, thanks for your suggestion, we will further optimize the performance by allocating the memory first to avoid concatenation. We will merge this PR for now since there are some measurable improvement.

peizhou001:
> neighbors

It can also be calculated when `probs_or_mask` exists, I have ever implmented it in #5717 .

PR #4718, opened 2022-10-17
[Example] Labor sampling

mfbalin:
## Description
Adding the example files for the labor sampling PR #4668, thus depends on that PR. Requires #6033 to run. The run command:
`python train_lightning.py --dataset=ogbn-products --use-uva --cache-size=500000 --batch-dependency=64`

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] Code is well-documented
- [x] To the best of my knowledge, examples are either not affected by this change,
      or have been fixed to be compatible with this change
- [x] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

PR #6033, opened 2023-07-22
[Feature] Dataloader gpu cache

mfbalin:
Branched off of #4341 to utilize gpu cache inside dataloader to cache feature copies. #4718 depends on this feature. Only the dataloader is modified in this PR. Should be merged after #4341.

@frozenbugs

### DIFF
--- a/python/dgl/dataloading/dataloader.py
+++ b/python/dgl/dataloading/dataloader.py
@@ -403,7 +403,7 @@ def _prefetch_update_feats(
                     )
                     values = F.astype(values, F.dtype(missing_values))
                     F.scatter_row_inplace(values, missing_index, missing_values)
-                    # reshape the flattened result to match the original shape.
+                    # Reshape the flattened result to match the original shape.
                     F.reshape(values, (values.shape[0],) + item_shape)
                     values.__cache_miss__ = missing_keys.shape[0] / ids.shape[0]
                     feats[tid, key] = values
--- a/python/dgl/dataloading/dataloader.py
+++ b/python/dgl/dataloading/dataloader.py
@@ -345,7 +345,7 @@ def _numel_of_shape(shape):
 
 def _init_gpu_caches(graph, gpu_caches):
     if not hasattr(graph, "_gpu_caches"):
-        graph._gpu_caches = {}, {}
+        graph._gpu_caches = {"node": {}, "edge": {}}
     if gpu_caches is None:
         return
     if not isinstance(gpu_caches, tuple):
@@ -356,13 +356,17 @@ def _init_gpu_caches(graph, gpu_caches):
             for key in frame.keys():
                 if key in gpu_caches[i] and gpu_caches[i][key] > 0:
                     column = frame._columns[key]
-                    if (key, type_) not in graph._gpu_caches[i]:
+                    node_or_edge = ["node", "edge"][i]
+                    if (key, type_) not in graph._gpu_caches[node_or_edge]:
                         cache = GPUCache(
                             gpu_caches[i][key],
                             _numel_of_shape(column.shape),
                             graph.idtype,
                         )
-                        graph._gpu_caches[i][key, type_] = cache, column.shape
+                        graph._gpu_caches[node_or_edge][key, type_] = (
+                            cache,
+                            column.shape,
+                        )
 
 
 def _prefetch_update_feats(
@@ -399,6 +403,7 @@ def _prefetch_update_feats(
                     )
                     values = F.astype(values, F.dtype(missing_values))
                     F.scatter_row_inplace(values, missing_index, missing_values)
+                    # reshape the flattened result to match the original shape.
                     F.reshape(values, (values.shape[0],) + item_shape)
                     values.__cache_miss__ = missing_keys.shape[0] / ids.shape[0]
                     feats[tid, key] = values
@@ -428,7 +433,7 @@ def _prefetch_for_subgraph(subg, dataloader):
         NID,
         dataloader.device,
         dataloader.pin_prefetcher,
-        dataloader.graph._gpu_caches[0],
+        dataloader.graph._gpu_caches["node"],
     )
     _prefetch_update_feats(
         edge_feats,
@@ -438,7 +443,7 @@ def _prefetch_for_subgraph(subg, dataloader):
         EID,
         dataloader.device,
         dataloader.pin_prefetcher,
-        dataloader.graph._gpu_caches[1],
+        dataloader.graph._gpu_caches["edge"],
     )
     return _PrefetchedGraphFeatures(node_feats, edge_feats)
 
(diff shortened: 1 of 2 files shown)


python/dgl/dataloading/dataloader.py
```diff
@@ -336,8 +339,41 @@ def __len__(self):
         ) // self.batch_size
 
 
+def _numel_of_shape(shape):
+    return reduce(operator.mul, shape, 1)
+
+
+def _init_gpu_caches(graph, gpu_caches):
+    if not hasattr(graph, "_gpu_caches"):
+        graph._gpu_caches = {}, {}
```
peizhou001, line 348: A tuple of dict is a little confusing, can we replace it to a dict of dict with meaningful keys? like {"node": {}, "edge": {}}?

python/dgl/dataloading/dataloader.py
```diff
@@ -336,8 +339,41 @@ def __len__(self):
...
+    if gpu_caches is None:
+        return
+    if not isinstance(gpu_caches, tuple):
+        gpu_caches = gpu_caches, {}
+    for i, frames in enumerate([graph._node_frames, graph._edge_frames]):
+        for tid, frame in enumerate(frames):
```
peizhou001, line 354: what tid means?

python/dgl/dataloading/dataloader.py
```diff
@@ -336,8 +339,41 @@ def __len__(self):
...
+            type_ = [graph.ntypes, graph.canonical_etypes][i][tid]
```
peizhou001, line 355: why naming a variable with suffix `_`?  In convention it indicates an internal member variable.

python/dgl/dataloading/dataloader.py
```diff
@@ -351,9 +387,25 @@ def _prefetch_update_feats(
...
+                    values, missing_index, missing_keys = cache.query(ids)
+                    missing_values = get_storage_func(parent_key, type_).fetch(
+                        missing_keys, device, pin_prefetcher
+                    )
+                    cache.replace(
+                        missing_keys, F.astype(missing_values, F.float32)
+                    )
+                    values = F.astype(values, F.dtype(missing_values))
+                    F.scatter_row_inplace(values, missing_index, missing_values)
+                    F.reshape(values, (values.shape[0],) + item_shape)
```
peizhou001, line 402: Why need `reshape` here?

peizhou001: My understanding is it store a flatten version of the desired data, if that is the case, can we add a description about it?

python/dgl/dataloading/dataloader.py
```diff
@@ -376,6 +428,7 @@ def _prefetch_for_subgraph(subg, dataloader):
         NID,
         dataloader.device,
         dataloader.pin_prefetcher,
+        dataloader.graph._gpu_caches[0],
```
peizhou001, line 431: 0 and 1 is a little confusing, how about use`_gpu_cache[node]` and `_gpu_cache[edge]` as mentioned above?

python/dgl/dataloading/dataloader.py
```diff
@@ -385,6 +438,7 @@ def _prefetch_for_subgraph(subg, dataloader):
         EID,
         dataloader.device,
         dataloader.pin_prefetcher,
+        dataloader.graph._gpu_caches[1],
```
peizhou001, line 441: Ditto.

python/dgl/dataloading/dataloader.py
```diff
@@ -791,6 +845,11 @@ class DataLoader(torch.utils.data.DataLoader):
         Whether to pin the feature tensors into pinned memory.
 
         Default: True if the graph is on CPU and :attr:`device` is CUDA.  False otherwise.
+    gpu_cache : dict or tuple(dict, dict), optional
```
peizhou001, line 848: Add link `https://github.com/NVIDIA-Merlin/HugeCTR/blob/main/gpu_cache/ReadMe.md` for further reference.

python/dgl/dataloading/dataloader.py
```diff
@@ -336,8 +339,41 @@ def __len__(self):
...
+def _numel_of_shape(shape):
+    return reduce(operator.mul, shape, 1)
+
+
+def _init_gpu_caches(graph, gpu_caches):
+    if not hasattr(graph, "_gpu_caches"):
+        graph._gpu_caches = {}, {}
+    if gpu_caches is None:
+        return
+    if not isinstance(gpu_caches, tuple):
```
peizhou001, line 351: Is it implicitly indicates when `gpu_cahches` is a single, it must be a node cache?

mfbalin: Yes, I assumed that would be the most used case. It is a single dict in this case which looks very clean.

python/dgl/dataloading/dataloader.py
mfbalin: I am following the variable naming convention in other parts of the code in this file.

python/dgl/dataloading/dataloader.py
mfbalin: An example usage of this feature is in the labor example PR #4718.

python/dgl/dataloading/dataloader.py
peizhou001: I see, so that should be fine.

python/dgl/dataloading/dataloader.py
```diff
@@ -791,6 +850,13 @@ class DataLoader(torch.utils.data.DataLoader):
         Whether to pin the feature tensors into pinned memory.
 
         Default: True if the graph is on CPU and :attr:`device` is CUDA.  False otherwise.
+    gpu_cache : dict or tuple(dict, dict), optional
```
peizhou001, line 853: `tuple(dict, dict) `-> `dict[dict]`?

peizhou001 (approved):
LGTM, please fix the left simple issue and request @frozenbugs for review.

python/dgl/dataloading/dataloader.py
```diff
@@ -351,9 +391,26 @@ def _prefetch_update_feats(
                         "Found a LazyFeature with no ID specified, "
                         "and the graph does not have dgl.NID or dgl.EID columns"
                     )
-                feats[tid, key] = get_storage_func(parent_key, type_).fetch(
-                    column.id_ or default_id, device, pin_prefetcher
-                )
+                ids = column.id_ or default_id
```
frozenbugs, line 394: id or ids?

python/dgl/dataloading/dataloader.py
```diff
@@ -351,9 +391,26 @@ def _prefetch_update_feats(
...
+                    values = F.astype(values, F.dtype(missing_values))
+                    F.scatter_row_inplace(values, missing_index, missing_values)
+                    # reshape the flattened result to match the original shape.
+                    F.reshape(values, (values.shape[0],) + item_shape)
+                    values.__cache_miss__ = missing_keys.shape[0] / ids.shape[0]
+                    feats[tid, key] = values
+                else:
+                    feats[tid, key] = get_storage_func(parent_key, type_).fetch(
+                        ids, device, pin_prefetcher
+                    )
```
frozenbugs, line 413: I would expecting that cache needs to be updated if hit misses, why it is not happening here?

python/dgl/dataloading/dataloader.py
```diff
@@ -351,9 +391,26 @@ def _prefetch_update_feats(
...
+                    values, missing_index, missing_keys = cache.query(ids)
+                    missing_values = get_storage_func(parent_key, type_).fetch(
+                        missing_keys, device, pin_prefetcher
+                    )
+                    cache.replace(
+                        missing_keys, F.astype(missing_values, F.float32)
+                    )
+                    values = F.astype(values, F.dtype(missing_values))
+                    F.scatter_row_inplace(values, missing_index, missing_values)
+                    # reshape the flattened result to match the original shape.
```
frozenbugs, line 406:
```suggestion
                    # Reshape the flattened result to match the original shape.
```

frozenbugs (approved):

PR #6056, opened 2023-07-28
[NN] Add EGT Layer

ZHITENGLI:
## Description

We'd like to add EGTLayer to the graph transformer family in DGL, which is the basic unit of Edge-augmented Graph Transformer (EGT).

## Checklist
- [x] Code is well-documented


rudongyu:
@dgl-bot

python/dgl/nn/pytorch/gt/egt.py
```diff
@@ -0,0 +1,178 @@
...
+
+    Parameters
+    ----------
+    ndim : int
+        Node embedding dimension.
+    edim : int
+        Edge embedding dimension.
+    num_heads : int
+        Number of attention heads, by which :attr: `ndim` is divisible.
+    num_vns : int
```
rudongyu, line 21: directly use "num_virtual_nodes" for clarity?

python/dgl/nn/pytorch/gt/egt.py
```diff
@@ -0,0 +1,178 @@
...
+        Number of virtual nodes.
+    dropout : float, optional
+        Dropout probability. Default: 0.0.
+    attn_dropout : float, optional
+        Attention dropout probability. Default: 0.0.
+    activation : callable activation layer, optional
+        Activation function. Default: nn.ELU().
+    ffn_multiplier : float, optional
```
rudongyu, line 29: ffn_hidden_size_multiplier? Shall we elaborate more on how it affects the hidden size of feed forward layers for node and edge update?

python/dgl/nn/pytorch/gt/egt.py
```diff
@@ -0,0 +1,178 @@
...
+        activation=nn.ELU(),
+        ffn_multiplier=2.0,
+        edge_update=True,
+    ):
+        super().__init__()
+        self.num_heads = num_heads
+        self.num_vns = num_vns
+        self.edge_update = edge_update
+
+        assert not (ndim % num_heads)
```
rudongyu, line 71: add a description for assertion error

python/dgl/nn/pytorch/gt/egt.py
```diff
@@ -0,0 +1,178 @@
...
+                nn.Linear(edim, edge_inner_dim),
+                activation,
+                nn.Linear(edge_inner_dim, edim),
+                nn.Dropout(dropout),
+            )
+
+    def forward(self, h, e, mask=None):
+        """Forward computation. Note: :attr:`h` and :attr:`e` should be padded
+        with embedding of virtual nodes if :attr:`num_vns` > 0, while
+        :attr:`mask` should be padded with `0` values for virtual nodes.
```
rudongyu, line 106: the padding should be in the beginning or at the end?

python/dgl/nn/pytorch/gt/egt.py
```diff
@@ -0,0 +1,178 @@
...
+
+        Parameters
+        ----------
+        h : torch.Tensor
+            A 3D input tensor. Shape: (batch_size, N, :attr:`ndim`), where N
+            is the sum of maximum number of nodes and number of virtual nodes.
```
rudongyu, line 112:
```suggestion
            is the sum of the maximum number of nodes and the number of virtual nodes.
```

python/dgl/nn/pytorch/gt/egt.py
```diff
@@ -0,0 +1,178 @@
...
+
+        h_r1 = h
+        e_r1 = e
+
+        h_ln = self.mha_ln_h(h)
+        e_ln = self.mha_ln_e(e)
+        QKV = self.QKV(h_ln)
+        E = self.E(e_ln)
+        G = self.G(e_ln)
+        shp = QKV.shape
```
rudongyu, line 138: I'd recommend to use meaningful names. e.g.,
```
bsz, N, _ = QKV.shape
...
```

python/dgl/nn/pytorch/gt/egt.py
```diff
@@ -0,0 +1,178 @@
...
+
+        assert not (ndim % num_heads)
+        self.dot_dim = ndim // num_heads
+        self.mha_ln_h = nn.LayerNorm(ndim)
+        self.mha_ln_e = nn.LayerNorm(edim)
+        self.E = nn.Linear(edim, num_heads)
+        self.QKV = nn.Linear(ndim, ndim * 3)
+        self.G = nn.Linear(edim, num_heads)
+        self.attn_dropout = nn.Dropout(attn_dropout)
+        self.O_h = nn.Linear(ndim, ndim)
```
rudongyu, line 79: Could we use module names like "edge_input_linear, node_input_linear, gate_linear, etc." for readability?

rudongyu (changes requested):
I've done a pass.

python/dgl/nn/pytorch/gt/egt.py
ZHITENGLI: Since `ffn_multiplier` is set to 1.0 for all configs, I'd like to remove it later.

python/dgl/nn/pytorch/gt/egt.py
ZHITENGLI: Actually it doesn't matter since users can define which nodes to be used in the readout function in EGT model. The official implementation of EGT is to put it at the beginning.

python/dgl/nn/pytorch/gt/egt.py
rudongyu: ok, fair enough.

python/dgl/nn/pytorch/gt/egt.py
rudongyu: It seems that the `degree_scalars` of virtual nodes is set to 1.0 at the beginning. Does it suggest that padding at the beginning is mandatory?

rudongyu: btw, do you think it's a better solution to include a VirtualNode module like what EGT does? Will it reduce user's burden and improve usability?

ZHITENGLI: > It seems that the `degree_scalars` of virtual nodes is set to 1.0 at the beginning. Does it suggest that padding at the beginning is mandatory?

Yes, I was wrong. It should be in the beginning.

ZHITENGLI: > btw, do you think it's a better solution to include a VirtualNode module like what EGT does? Will it reduce user's burden and improve usability?

I think so, but this VirtualNode module may not be universal since its function is not only to generate virtual node embedding, but also to general positional encodings or attention bias related to the virtual nodes. While PEs and attention bias are different in different models.

python/dgl/nn/pytorch/gt/egt.py
```diff
@@ -0,0 +1,178 @@
...
+            invalid positions are indicated by `-inf`.
+            Shape: (batch_size, N, N). Default: None.
+
+        Returns
+        -------
+        h : torch.Tensor
+            The output node embedding. Shape: (batch_size, N, :attr:`feat_size`).
+        e : torch.Tensor, optional
+            The output edge embedding. Shape: (batch_size, N, N, :attr:`edge_feat_size`).
+            It is only returned when :attr:`edge_update` is True.
```
rudongyu, line 125:
```suggestion
            It is returned only if :attr:`edge_update` is True.
```

python/dgl/nn/pytorch/gt/egt.py
```diff
@@ -0,0 +1,178 @@
...
+        """
+
```
rudongyu, line 127:
```suggestion

```

rudongyu (commented):
some minors

rudongyu:
@dgl-bot

rudongyu:
@dgl-bot

rudongyu:
@dgl-bot

rudongyu (approved):
Thanks for your great effort!

PR #6058, opened 2023-07-28
[Misc] Simplify the data path's assignment in python utils

songqing:
## Description

1 The assignment of data path name of different datasets are duplicated, and it may be easy to misspelled and may be missed somewhere when changing it.

## Checklist
- [.] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [.] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [.] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [.] All changes have test coverage
- [.] Code is well-documented
- [.] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### DIFF
--- a/examples/pytorch/dimenet/qm9.py
+++ b/examples/pytorch/dimenet/qm9.py
@@ -130,11 +130,19 @@ class QM9(QM9Dataset):
             verbose=verbose,
         )
 
+    @property
+    def graph_path(self):
+        return f"{self.save_path}/dgl_graph.bin"
+
+    @property
+    def line_graph_path(self):
+        return f"{self.save_path}/dgl_line_graph.bin"
+
     def has_cache(self):
         """step 1, if True, goto step 5; else goto download(step 2), then step 3"""
-        graph_path = f"{self.save_path}/dgl_graph.bin"
-        line_graph_path = f"{self.save_path}/dgl_line_graph.bin"
-        return os.path.exists(graph_path) and os.path.exists(line_graph_path)
+        return os.path.exists(self.graph_path) and os.path.exists(
+            self.line_graph_path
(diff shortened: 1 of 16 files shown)


songqing:
@BarclayII Could you help to review this PR?

frozenbugs:
@dgl-bot

frozenbugs:
Thanks for the fixing, can you break 2 unrelated changes to 2 PRs?

For extract_archive, instead of changing the default value to True, what about change the documentation to False since our customer may already have code that works without setting the overwrite, we don't want to accidentally break them.

songqing:
> Thanks for the fixing, can you break 2 unrelated changes to 2 PRs?
> 
> For extract_archive, instead of changing the default value to True, what about change the documentation to False since our customer may already have code that works without setting the overwrite, we don't want to accidentally break them.

OK, I have removed the change of extract_archive().

frozenbugs:
@dgl-bot

frozenbugs (approved):

songqing:
CI failed, but it seems that the fail has nothing to do with my PR, as the UT test_gat with GPU succeed, but the UT failed with  CPU. And the UT passed in my local environment.

frozenbugs:
@dgl-bot

PR #6070, opened 2023-08-01
Improving the DEEPWALK example.

drivanov:
## Description

In the current PR we eliminate the following warning:
```
/usr/local/lib/python3.10/dist-packages/dgl/heterograph.py:92: DGLWarning: Recommend creating graphs by `dgl.graph(data)` instead of `dgl.DGLGraph(data)`.
```
We have also changed the output of information about training time for batches. Now this information is displayed on the same line of the screen for all batches.

## Checklist
- [x] Code is well-documented


examples/pytorch/ogb/deepwalk/deepwalk.py
```diff
@@ -247,26 +247,29 @@ def fast_train(self):
                 if i > 0 and i % self.args.print_interval == 0:
                     if self.args.print_loss:
                         print(
-                            "Batch %d training time: %.2fs loss: %.4f"
+                            "\rBatch %d training time: %.2fs loss: %.4f"
```
frozenbugs, line 250: why this change?

frozenbugs: revert?

drivanov: Without this change, we do have more than 1100 lines (1138, to be exact) similar to the following one:
```
Batch 93700, training time: 1.49s
```
I think it would be better to just output this information to the same line of the screen and change this line about every 1.5 seconds.

frozenbugs: Usually for example in DGL, we output newline without erase the previous. And we use i % self.args.print_interval = 0 to reduce the frequency if needed.

drivanov: Reverted

frozenbugs:
Thanks for taking time to make DGL example code excellent!

PR #6086, opened 2023-08-02
[Graphbolt] SampleNeighbors code refactor: add pick_fn

RamonZhou:
## Description
Extract the logic of picking out of the main process, reducing the number of params of `SampleNeighborsImpl()`.
The params of the outer layer hasn't changed.
In the future, `pick_fn` will be modified again to put the picked neighbors directly into preallocated tensors.
Plan: https://docs.google.com/document/d/1O7OuUAMOWbaNsPKXKJYfdWiktD9qRlgc8NSmdKmxDUY/edit?usp=sharing.

## Checklist
- [ ] Code is well-documented

## Changes
- Add `GetPickFn()`.


peizhou001 (approved):

frozenbugs:
@mfbalin Can you take a look at this change and the plan for the optimization in the Description section.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -232,11 +265,17 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighbors(
...
+            fanouts, replace, indptr_.options(), type_per_edge_, probs_or_mask,
+            args));
   } else {
     SamplerArgs<SamplerType::NEIGHBOR> args;
     return SampleNeighborsImpl(
-        nodes, fanouts, replace, return_eids, probs_or_mask, args);
+        nodes, return_eids,
+        GetPickFn(
+            fanouts, replace, indptr_.options(), type_per_edge_, probs_or_mask,
+            args));
```
frozenbugs, line 278:
```
PickFn pickFn;
if (layer) {
    ...
    pickFn = 
} else {
    ...
    pickFn = 
}
return SampleNeighborsImpl(nodes, return_eids, pickFn);
```

frozenbugs:
LGTM.

graphbolt/include/graphbolt/csc_sampling_graph.h
```diff
@@ -222,12 +224,8 @@ class CSCSamplingGraph : public torch::CustomClassHolder {
       const std::string& shared_memory_name);
 
  private:
-  template <SamplerType S>
```
mfbalin, line 225: Using std::function to dispatch the appropriate function is going to be slower than the current code. I would have preferred the use of templates. Compared to templates, the advantage is the smaller generated code size. What happens when we use std::function is the dispatch happens at runtime and the compiler can't optimize PickFn and the code calling PickFn together. It becomes slower than simply using if expressions to call one function or the other. You can see this link: https://stackoverflow.com/questions/67615330/why-stdfunction-is-too-slow-is-cpu-cant-utilize-instruction-reordering

However, if PickFn takes a very long time, this overhead can simply be ignored. In our case, it is called `nodes.size()` times. I personally wouldn't want to pay this overhead `nodes.size()` times.

The solution would be to employ the use of templates, depending on whether `fanouts.size() > 1`, we can dispatch the appropriate `PickFn` at compile time. For this to work, instead of defining `PickFn` globally, we need to make it a template parameter.

graphbolt/include/graphbolt/csc_sampling_graph.h
```diff
@@ -31,6 +31,8 @@ struct SamplerArgs<SamplerType::LABOR> {
   int64_t num_nodes;
 };
 
+using PickFn = std::function<torch::Tensor(int64_t, int64_t)>;
```
mfbalin, line 34: This type would need to be a template type parameter to get fastest code.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -131,16 +131,54 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+    return [&fanouts, replace, &options, &probs_or_mask, args](
+               int64_t offset, int64_t num_neighbors) {
+      return Pick(
+          offset, num_neighbors, fanouts[0], replace, options, probs_or_mask,
+          args);
+    };
+  }
+}
+
 c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
```
mfbalin, line 179: PickFn needs to be a template to get fastest code.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -232,11 +265,17 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighbors(
         static_cast<int64_t>(0), std::numeric_limits<int64_t>::max());
     SamplerArgs<SamplerType::LABOR> args{indices_, random_seed, NumNodes()};
     return SampleNeighborsImpl(
```
mfbalin, line 267: If my template suggestion is used, then the dispatch would happen here with an `if` expression.

mfbalin (commented):
Hi all, here are my thoughts on the current refactoring efforts. The use of std::function is similar to the use of dynamic polymorphism and the use of virtual functions. My suggestion is to avoid using it and employ templates instead. I assume that we want max performance from the sampling algorithms implemented here, but I also know that readability and maintainability are big concerns as well. Since the benchmarks aren't available in the repo, does this change introduce any regressions in terms of performance because I would expect it to be slightly slower?

I am simply commenting instead of requesting changes. You guys can make the final decision about which approach to take.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -131,16 +131,54 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+ */
 template <SamplerType S>
+PickFn GetPickFn(
+    const std::vector<int64_t>& fanouts, bool replace,
+    const torch::TensorOptions& options,
+    const torch::optional<torch::Tensor>& type_per_edge,
+    const torch::optional<torch::Tensor>& probs_or_mask, SamplerArgs<S> args) {
+  // If fanouts.size() > 1, perform sampling for each edge type of each node,
+  // otherwise just sample once for each node with no regard of edge types.
+  if (fanouts.size() > 1) {
```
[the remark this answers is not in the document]
mfbalin: Then, we can make the return type of GetPickFn simply `auto`, and don't have to use `std::function`.

graphbolt/include/graphbolt/csc_sampling_graph.h
peizhou001: That's a good suggestion! We'll give it a try to see if there are any unforeseen obstacles. If everything goes smoothly, we can adopt this solution without any issues.

graphbolt/src/csc_sampling_graph.cc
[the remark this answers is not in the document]
mfbalin: However, the fastest solution would be to do the if only at dispatching and no more if expressions after that.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -227,16 +271,39 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighbors(
...
+          GetPickFn(
+              fanouts, replace, indptr_.options(), type_per_edge_,
+              probs_or_mask, args));
+    } else {
+      return SampleNeighborsImpl(
+          nodes, return_eids,
+          GetPickFnByEtype(
+              fanouts, replace, indptr_.options(), type_per_edge_,
+              probs_or_mask, args));
+    }
```
RamonZhou, line 306: Is there any good way to reduce the repeated code? The problem is that the type of `args` in the two branches are different. The return types of `GetPickFn` and `GetPickFnByEtype` are also different.

mfbalin: It is good for now. However, if one more template parameter is going to be added in the future, having 2^3 options, then we can consider either using macros (`#define`) or the use of template metaprogramming tricks to loop over all options using recursion. An example is here: https://stackoverflow.com/questions/52286202/dynamic-dispatch-to-template-function-c. There may be better examples or ways to do it than this one.

However, it is not getting out of hand for now. We can consider reducing code size later if needed. The bad thing about using templates is the dispatch can become ugly. The good thing is the excellent performance (since everything happens at compile time) and that only the dispatch is ugly, the main code is almost identical :).

mfbalin: However, we might want to consider the following solution also: if having one more if condition at runtime doesn't bother us when it comes to performance, which it shouldn't as we do the same for the replace or weight parameters, then it is possible modify `GetPickFn` to move the `if (ByEtype)` check to inside of the lambda to eliminate `GetPickFn`'s `ByEtype` template argument. Then, a single lambda will be returned and we don't have to dispatch with respect to `ByEtype` anymore as the returned lambda will do the if check at runtime. In this case, since the if condition always succeeds or fails, it is easy for the branch predictor to get it right every time so shouldn't have any noticeable runtime impact.

mfbalin (approved):
If we want to reduce the ugliness of dispatch, then `GetPickFn` can return a single lambda and we can remove its template argument `ByEtype`. The if condition will need to be moved to inside of the lambda function in this case.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -131,16 +131,53 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+ * without replacement. If True, a value can be selected multiple times.
+ * Otherwise, each value can be selected only once.
+ * @param options Tensor options specifying the desired data type of the result.
+ * @param type_per_edge A tensor representing the type of each edge, if
+ * present.
+ * @param probs_or_mask Optional tensor containing the (unnormalized)
+ * probabilities associated with each neighboring edge of a node in the original
+ * graph. It must be a 1D floating-point tensor with the number of elements
+ * equal to the number of edges in the graph.
+ * @param args Contains labor specific arguments.
```
frozenbugs, line 149:
```suggestion
 * @param args Contains sampling algorithm specific arguments.
```

frozenbugs (approved):

peizhou001 (approved):

graphbolt/src/csc_sampling_graph.cc
RamonZhou: This can't be easily done since `GetPickFn` now returns a lambda and each lambda function has its unique type. So the return values of these two `GetPickFn` have different types.

PR #6097, opened 2023-08-04
Revert "[Misc] clean up remove --inplace."

frozenbugs:
Reverts dmlc/dgl#6096


frozenbugs:
reverted, suspicious since it might break the CI.

PR #6101, opened 2023-08-04
[Graphbolt] Add num_pick_fn into neighbor sampling

RamonZhou:
## Description
Similar to the `pick_fn`, add `num_pick_fn` as an UDF into the sampling process. `num_pick_fn` returns the number of neighbors to be picked for the given node.
This function will be used when pre-allocating tensors in the next PR. For now, it's only used in a check.
This is the Step 2 in https://docs.google.com/document/d/1O7OuUAMOWbaNsPKXKJYfdWiktD9qRlgc8NSmdKmxDUY/edit?usp=sharing.

## Checklist
- [ ] Code is well-documented

## Changes
- Added `NumPick()` and `NumPickByEtype()`.
- Added `GetNumPickFn()` to get `num_pick_fn`.
- Utilized lambda and template.


graphbolt/src/csc_sampling_graph.cc
```diff
@@ -131,6 +131,92 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
 }
 
+int64_t NumPick(
+    int64_t fanout, bool replace,
+    const torch::optional<torch::Tensor>& probs_or_mask, int64_t offset,
+    int64_t num_neighbors) {
+  int64_t actual_fanout = fanout == -1 ? num_neighbors : fanout;
+  int64_t max_count =
+      probs_or_mask.has_value()
+          ? torch::nonzero(
```
mfbalin, line 141: Can you use `torch::count_nonzero`? https://pytorch.org/docs/stable/generated/torch.count_nonzero.html

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -131,6 +131,92 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+ * equal to the number of edges in the graph.
+ *
+ * @return A lambda function which takes offset and num_neighbors as params and
+ * returns the number of neighbors-to-be-sampled.
+ */
+auto GetNumPickFn(
+    const std::vector<int64_t>& fanouts, bool replace,
+    const torch::optional<torch::Tensor>& type_per_edge,
+    const torch::optional<torch::Tensor>& probs_or_mask) {
+  // If fanouts.size() > 1, count with sampling for each edge type of each node,
```
peizhou001, line 206: Can you elaborate this description? It looks a little confusing.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -221,6 +308,11 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
                 // This number should be the same as the result of num_pick_fn.
                 num_picked_neighbors_per_node[i + 1] =
                     picked_neighbors_cur_thread[i - begin].size(0);
+                TORCH_CHECK(
```
peizhou001, line 311: For tensor read, use raw pointer as `item()` is very slow.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -221,6 +308,11 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
...
+                    num_picked_neighbors_per_node[i + 1].item<int64_t>() ==
+                        num_pick_fn(offset, num_neighbors),
+                    "Return value of num_pick_fn doesn't match the actual "
```
peizhou001, line 314: Actual picked number.

graphbolt/src/csc_sampling_graph.cc
RamonZhou: Is there any other faster way? Or maybe we can delete this check.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -131,6 +131,117 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+ * node.
+ * @param num_neighbors The number of neighbors of this node.
+ *
+ * @return The pick number of the given node.
+ */
+int64_t NumPick(
+    int64_t fanout, bool replace,
+    const torch::optional<torch::Tensor>& probs_or_mask, int64_t offset,
+    int64_t num_neighbors) {
+  int64_t actual_fanout = fanout == -1 ? num_neighbors : fanout;
```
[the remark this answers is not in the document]
frozenbugs: I just confirmed that the comment was not accurate for non-uniform pick, @peizhou001 will update.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -131,6 +131,117 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+        const scalar_t* type_per_edge_data = type_per_edge.data_ptr<scalar_t>();
+        while (etype_begin < end) {
+          scalar_t etype = type_per_edge_data[etype_begin];
+          TORCH_CHECK(
+              etype >= 0 && etype < (int64_t)fanouts.size(),
+              "Etype values exceed the number of fanouts.");
+          auto etype_end_it = std::upper_bound(
+              type_per_edge_data + etype_begin, type_per_edge_data + end,
+              etype);
+          etype_end = etype_end_it - type_per_edge_data;
```
frozenbugs, line 195: And remove L181?
```suggestion
          int64_t etype_end = etype_end_it - type_per_edge_data;
```

graphbolt/src/csc_sampling_graph.cc
peizhou001: raw pointer `data_ptr`.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -131,6 +131,117 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+      probs_or_mask.has_value()
+          ? torch::count_nonzero(
+                probs_or_mask.value().slice(0, offset, offset + num_neighbors))
+                .item<int64_t>()
+          : num_neighbors;
+  if (replace) {
+    return max_count == 0 ? 0 : actual_fanout;
+  } else {
+    return std::min(max_count, actual_fanout);
+  }
```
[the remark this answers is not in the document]
RamonZhou: I noticed an inconsistency between Neighbor sample algorithm (which has been in use for long) and Labor algorithm. The number of picked neighbors is different when `fanout = -1, replace = true` and `mask` is not full zero. Under this circumstance, the Neighbor sample algorithm only picks as many as the number of neighbors with **non-zero** probability, while the picked number of Labor algorithm is the same as the number of **all** neighbors. It should be consistent. @mfbalin which way from your view is more plausible?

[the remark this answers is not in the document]
mfbalin: Let's modify labor and equalize the behaviors. The implementation here is supposed to match neighbor fully.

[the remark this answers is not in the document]
mfbalin: It can be formalized by saying that for `fanout=-1`, the behaviour is simply the same as passing the `num_neighbors` as the fanout for each vertex. For weighted sampling, we can say that `num_neighbors` is treated as the number of nonzero weights.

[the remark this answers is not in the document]
frozenbugs: Semantically, the fanout = -1 with replace true is non-defined. There are 2 options:
- Simply throw an exception when it happens.
- Define a consistent behavior for both sampler and document it clearly (in both c++ and python doc).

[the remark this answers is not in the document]
RamonZhou: Let's modify Labor to match Neighbor.
There is another inconsistency: when `replace = true` and `fanout` is bigger than the number of neighbors, Labor only samples `num_neighbors` neighbors. I suppose the expected result is that `fanout` neighbors are sampled with replacement.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -199,7 +307,7 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
                   local_grain_size);
 
               for (scalar_t i = begin; i < end; ++i) {
-                const auto nid = nodes[i].item<int64_t>();
+                const auto nid = *nodes[i].data_ptr<int64_t>();
```
mfbalin, line 310: const auto nid = nodes.data_ptr<int64_t>()[i];
It is best if the dataptr is assigned outside of the loop.

mfbalin (changes requested):

frozenbugs (approved):

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -131,6 +131,112 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
...
+ * probabilities associated with each neighboring edge of a node in the original
+ * graph. It must be a 1D floating-point tensor with the number of elements
+ * equal to the number of edges in the graph.
+ * @param offset The starting edge ID for the connected neighbors of the given
+ * node.
+ * @param num_neighbors The number of neighbors of this node.
+ *
+ * @return The pick number of the given node.
+ */
+int64_t NumPick(
```
frozenbugs, line 157: Add NumPick, NumPickByEtype definition in the header file, and move the whole code section below L438: LoadFromSharedMemory.

PR #6114, opened 2023-08-08
[Graphbolt] Update the sampler algorithm documentation for fanout = -1.

RamonZhou:
## Description
Change the Labor sampling algorithm to match the behavior of Neighbor algorithm.
I simply changed the calculation of the actual picked number at the beginning of Labor process. However, since soon the picked number will be calculated in advance, this calculation will become useless and will be deleted.
Trying to solve https://github.com/dmlc/dgl/pull/6101/files/5e0de154b1d896e5c26be164b3480b9b81232c6c#r1285651084.

## Checklist
- [ ] Code is well-documented

## Changes
- Modified Labor.
- Modified comments.

### DIFF
--- a/graphbolt/src/csc_sampling_graph.cc
+++ b/graphbolt/src/csc_sampling_graph.cc
@@ -458,10 +458,10 @@ inline torch::Tensor UniformPick(
  * @param num_neighbors The number of neighbors to pick.
  * @param fanout The number of edges to be sampled for each node. It should be
  * >= 0 or -1.
- *  - When the value is -1, all neighbors (with non-zero probability, if
- * weighted) will be sampled once regardless of replacement. It is equivalent to
- * selecting all neighbors with non-zero probability when the fanout is >= the
- * number of neighbors (and replacement is set to false).
+ *  - When the value is -1, all neighbors with non-zero probability will be
+ * sampled once regardless of replacement. It is equivalent to selecting all
+ * neighbors with non-zero probability when the fanout is >= the number of
+ * neighbors (and replacement is set to false).
  *  - When the value is a non-negative integer, it serves as a minimum
  * threshold for selecting neighbors.
  * @param replace Boolean indicating whether the sample is performed with or
--- a/graphbolt/src/csc_sampling_graph.cc
+++ b/graphbolt/src/csc_sampling_graph.cc
@@ -329,10 +329,9 @@ c10::intrusive_ptr<CSCSamplingGraph> CSCSamplingGraph::LoadFromSharedMemory(
  * @param num_neighbors The number of neighbors to pick.
  * @param fanout The number of edges to be sampled for each node. It should be
  * >= 0 or -1.
- *  - When the value is -1, all neighbors (with non-zero probability, if
- * weighted) will be sampled once regardless of replacement. It is equivalent to
- * selecting all neighbors with non-zero probability when the fanout is >= the
- * number of neighbors (and replacement is set to false).
+ *  - When the value is -1, all neighbors will be sampled once regardless of
+ * replacement. It is equivalent to selecting all neighbors when the fanout is
+ * >= the number of neighbors (and replacement is set to false).
  *  - When the value is a non-negative integer, it serves as a minimum
  * threshold for selecting neighbors.
  * @param replace Boolean indicating whether the sample is performed with or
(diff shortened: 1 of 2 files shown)


graphbolt/src/csc_sampling_graph.cc
```diff
@@ -612,7 +615,22 @@ inline torch::Tensor LaborPick(
...
+                 : num_neighbors;
+  if (num_valid_neighbors == 0) {
+    return torch::tensor({}, options);
+  }
+  if constexpr (Replace) {
+    fanout = fanout < 0 ? num_valid_neighbors : fanout;
+  } else {
+    fanout = fanout < 0 ? num_valid_neighbors
+                        : std::min(fanout, num_valid_neighbors);
+  }
```
RamonZhou, line 633: According to the big plan, this calculation will be replaced by a call of `NumPick` function soon.

RamonZhou: Will this change affect the correctness/other behaviors of Labor algorithm? @mfbalin please confirm.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -329,9 +329,10 @@ c10::intrusive_ptr<CSCSamplingGraph> CSCSamplingGraph::LoadFromSharedMemory(
  * @param num_neighbors The number of neighbors to pick.
  * @param fanout The number of edges to be sampled for each node. It should be
  * >= 0 or -1.
- *  - When the value is -1, all neighbors will be chosen for sampling. It is
- * equivalent to selecting all neighbors with non-zero probability when the
- * fanout is >= the number of neighbors (and replacement is set to false).
+ *  - When the value is -1, all neighbors with non-zero probability will be
+ * sampled once regardless of replacement. It is equivalent to selecting all
+ * neighbors with non-zero probability when the fanout is >= the number of
+ * neighbors (and replacement is set to false).
```
frozenbugs, line 335: For UniformPick, we don't need to mention non-zero probability since there is no probability concept in UniformPick.

python/dgl/graphbolt/impl/csc_sampling_graph.py
```diff
@@ -395,10 +395,10 @@ def _sample_neighbors(
...
                 type of the nodes.
             The value of each fanout should be >= 0 or = -1.
-              - When the value is -1, all neighbors will be chosen for
-                sampling. It is equivalent to selecting all neighbors when
-                the fanout is >= the number of neighbors (and replace is set to
-                false).
+              - When the value is -1, all neighbors with non-zero probability
+                will be sampled once regardless of replacement. It is
+                equivalent to selecting all neighbors when the fanout is >=
+                the number of neighbors (and replace is set to false).
```
frozenbugs, line 401: It is equivalent to selecting all neighbors **with non-zero probability**.

frozenbugs: Also L458.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -612,6 +614,11 @@ inline torch::Tensor LaborPick(
     const torch::TensorOptions& options,
     const torch::optional<torch::Tensor>& probs_or_mask,
     SamplerArgs<SamplerType::LABOR> args) {
+  // TODO: fix inconsistency with Neighbor sampler.
+  // 1. Replace = true, fanout = -1. Expected: sample all neighbors with
+  // non-zero probility once regardless of replacement.
+  // 2. Replace = true, fanout > num_neighbors. Expected: sample fanout many
+  // neighbors.
```
RamonZhou, line 621: @mfbalin Could you please help modify Labor sampler? Since you have the best knowledge of this process. :)

frozenbugs (approved):

graphbolt/src/csc_sampling_graph.cc
mfbalin: Why don't we just call NeighborSampler if fanout is -1 and replace is true? I don't think replicating code makes a lot of sense. We can do this during dispatch. When fanout is -1, it doesn't matter whether labor or neighbor is used.

mfbalin: I think the change should look like after we ensure that labor is never called with `fanout == -1`:
`fanout = replace ? fanout : std::min(fanout, num_neighbors);`

mfbalin: And inside the `Pick<SamplerType::LABOR>` function, we can call `NonuniformPick` and `UniformPick` if `fanout < 0`. This way, there won't be duplicated code and will be a single defined behavior.

mfbalin: See this commit: https://github.com/mfbalin/dgl-1/commit/5c3f8de64e1a09447be05868a85a7998b279ae7e

Issue #6135, opened 2023-08-11
[Build] Failed to build labor sampling related code on windows_cu118


Rhett-Ying:
## To Reproduce

Builder: vs2019
<img width="983" alt="image" src="https://github.com/dmlc/dgl/assets/85214957/4f622bdd-dc28-405f-9a33-cd811fe022e2">

## Environment

 - DGL Version (e.g., 1.0): 1.1.x, master
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3):
 - OS (e.g., Linux): Windows
 - How you installed DGL (`conda`, `pip`, source):
 - Build command you used (if compiling from source): vs2019
 - Python version:
 - CUDA/cuDNN version (if applicable): cuda11.8
 - GPU models and configuration (e.g. V100):
 - Any other relevant information:

Rhett-Ying:
@mfbalin Hi, have you ever tried to build on windows with cuda? Do you have any ideas on the build error I posted above?

mfbalin:
Hi @Rhett-Ying, that line has a static_assert, static_assert with no error message in 2nd argument is available only in c++17. We can fix it by adding a second argument such as "alignment requirement is not met!".
https://en.cppreference.com/w/cpp/language/static_assert

Rhett-Ying:
let me try this

Rhett-Ying:
@mfbalin errors thrown with `thrust` still exist even resolve the `static_assert`. Maybe build with c++17 works.
<img width="1093" alt="image" src="https://github.com/dmlc/dgl/assets/85214957/03b41172-8da2-4b9e-8225-20da233bb4a8">

mfbalin:
I agree, it might help. We can attempt to upgrade to c++17 again.

Rhett-Ying:
let me try to update `thrust`.

Rhett-Ying:
after update `thurst` in https://github.com/dmlc/dgl/pull/6139, many errors are gone.

those lines are added in https://github.com/dmlc/dgl/pull/5648

<img width="1087" alt="image" src="https://github.com/dmlc/dgl/assets/85214957/fd2e603f-667b-4076-b51e-5ed513111254">

Rhett-Ying:
[...]

PR #6065, opened 2023-07-31
[dev] compile dgl with c++17

Rhett-Ying:
## Description

Try to fix errors like [this](https://github.com/dmlc/dgl/pull/4341#issuecomment-1653209799).
```
CMake Error at gpu_cache_generated_nv_gpu_cache.cu.o.cmake:276 (message):
  Error generating file
  /dgl/build/CMakeFiles/gpu_cache.dir/third_party/HugeCTR/gpu_cache/src/./gpu_cache_generated_nv_gpu_cache.cu.o


make[2]: *** [CMakeFiles/gpu_cache.dir/build.make:77: CMakeFiles/gpu_cache.dir/third_party/HugeCTR/gpu_cache/src/gpu_cache_generated_nv_gpu_cache.cu.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:182: CMakeFiles/gpu_cache.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```
## Checklist
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [ ] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [ ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

PR #6138, opened 2023-08-11
[Dev] Change CXX standard to 17

mfbalin:
## Description
Changing CXX standard to 17 hoping that issues similar to #6135 will be resolved.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,9 @@ cmake_minimum_required(VERSION 3.5)
 project(dgl C CXX)
 message(STATUS "Start configuring project ${PROJECT_NAME}")
 
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
 # cmake utils
 include(cmake/util/Util.cmake)
 include(cmake/util/MshadowUtil.cmake)
@@ -245,8 +248,6 @@ if(USE_CUDA)
 else(USE_CUDA)
   add_library(dgl SHARED ${DGL_SRC})
 endif(USE_CUDA)
-
-set_property(TARGET dgl PROPERTY CXX_STANDARD 14)
 
 # include directories
 target_include_directories(dgl PRIVATE "include")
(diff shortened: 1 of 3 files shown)


mfbalin:
@Rhett-Ying I don't know why we still need to pass the `-std=c++17` flag to NVCC manually, but let's see if everything will work with this patch. If CI passes, we can check if #6135 is resolved also.

mfbalin:
Surprisingly, this patch doesn't have the problem in #6065 .

Rhett-Ying:
let me try with this PR on windows

Rhett-Ying:
same issue on windows + vs2019 with this PR's branch
image


image

mfbalin:
@Rhett-Ying could you try with the newly added commits also? Also, if there is a chance I can setup the same windows build environment, how can I do that?

mfbalin:
Also, are we sure that the changes here enable c++17 compilation for the windows build as well?

mfbalin:
We need to revert #5648 to fix #6135. But this PR can still be merged if we want to upgrade to CXX17.

Rhett-Ying:
> We need to revert #5648 to fix #6135. But this PR will fix the static_assert issue encountered on the Windows build without modifying the labor sampling code. Also, it can still be merged if we want to upgrade to CXX17. @Rhett-Ying

Reverting #5648 is the only way to fix the build issue on windows_cuda? I hit this issue when cutting a new release, and for this release, I think it's ok to not cherry-pick #5648 . 

As for this PR, I tried it on windows+cu118, and it failed due to the #5648 . So if we revert it, this PR should be fine on windows.

mfbalin:
I suggested we revert it because I couldn't find a way to fix the build error on Windows+cuda case. If you think windows+cuda release can simply exclude #5648 , that sounds alright. I think they will add windows CI to the CCCL repository (thrust, cub and libcudacxx now lives there) soon so we can update CCCL and windows build should start working.

Rhett-Ying:
As discussed above, let's merge this into master and to see if CCCL resolves the build issue on windows+cuda.

Rhett-Ying (approved):

PR #6143, opened 2023-08-11
[Misc] passing C/CXX flags to tensoradapter's cmake subprocess

initzhang:
## Description
Currently Tensor adapter libraries are built [in an individual CMake build](https://github.com/dmlc/dgl/blob/2c64ae03303ce5d53019d0ed1583be328f6f8c0d/CMakeLists.txt#L337C33-L337C33). And if user has specified the cmake C/CXX compiler flags in the main build, these flags are not passed to the sub-build, which causes actually different versions of C/CXX are used for the main build and individual build. From my perspective, user would anticipate that only the specified version is used. 

## Checklist
- [ ] Code is well-documented


frozenbugs:
@dgl-bot

CMakeLists.txt
```diff
@@ -355,6 +355,8 @@ if(BUILD_TORCH)
       CMAKE_COMMAND=${CMAKE_CMD}
       CUDA_TOOLKIT_ROOT_DIR=${CUDA_TOOLKIT_ROOT_DIR}
       USE_CUDA=${USE_CUDA}
+      CMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
```
Rhett-Ying, line 358: `CMAKE_<LANG>_COMPILER` indicates the full path of compiler, but you're using them to pass flags? https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html

initzhang: Hi, do you recommend that we use different names to pass the flags ? such as `MAIN_CMAKE_{LANG}_COMPILER`?

Rhett-Ying:
Hi @initzhang  Actually I don't think it's a good idea to always blindly pass c/cxx flags defined at the top level down to tensoradaptor build as it may incur unexpected errors. We already have dedicated build script/makefile which controls the build behavior on its own.

initzhang:
> Hi @initzhang Actually I don't think it's a good idea to always blindly pass c/cxx flags defined at the top level down to tensoradaptor build as it may incur unexpected errors. We already have dedicated build script/makefile which controls the build behavior on its own.

Hi @Rhett-Ying , thanks for reply! maybe blindly passing flags is not desired, but I think there should at least be some warnings/notifications when user setting their C/CXX flags. I think it is not a rare case where users have multiple versions of gcc/g++ installed on their machine. And once user have specified one version, I think they should be reminded that __some part of the build__ does not use the specified version.

Rhett-Ying:
So the main question is how to enable user to control the build behavior of underlying separate modules such as `tensoradaptor` more freely. For now, DGL passes down several envs/variables such as `TORCH_PYTHON_INTERPS`, `TORCH_CUDA_ARCH_LIST`, `USE_CUDA` and so on.

As we don't see any other required envs/variables or any must-have support for now, I think let's keep it. Even if we needs to add some other controls on downstream module's build, we could follow the existing way: add into top-level `CMakeLists.txt` and update corresponding `build.sh/bat`.

initzhang:
Got it, thanks for the reply!

PR #6152, opened 2023-08-14
[Graphbolt] Rewrite sampling process to eliminate torch::cat

RamonZhou:
## Description
With the preparation of `pick_fn` `num_pick_fn` UDF and modified pick methods, we can rewrite the main process of neighbor sampling.

Outline:
1. Calculate pick number of each node in parallel. 
2. Allocate the final tensor with calculated total length. 
3. Pick neighbors in parallel. Put the picked neighbors directly into the allocated tensor.

Thus the slow `torch::cat` can be eliminated.

## Checklist
- [ ] Code is well-documented

## Changes
- [x] Rewrite `SampleNeighborsImpl`.
- [x] Add comments.
- [x] Test the improvement.

## Performances

Tested performance on 5 different graphs with settings:
Graph | Num of nodes | Num of neighbors per etype | Node types | Edge types | Fanouts
:-:|:-:|:-:|:-:|:-:|:-
G1 | 10,000,000 | 5 | 3 | 2 | [3,4]
G2 | 100 | 1,000,000 | 3 | 2 | [899999, 911111]
G3 | 10,000 | 100,000 | 3 | 2 | [6000, 9000]
G4 | 10,000 | 5,000 | 30 | 20 | [1000, 8000, ..., 3000]
G5 (homo) | 1,000,000 | 100 | 1 | 1 | [80]

4 different types of problem settings are tested:
1. Neighbor sampling algorithm, uniform
2. Neighbor sampling algorithm, non-uniform
3. Labor sampling algorithm, uniform
4. Labor sampling algorithm, non-uniform

The bar charts with means and errors are shown below.

## Update (8.21)

Based on the comments, I've made some small changes and did some improvements as well:
1. Add checks, which can also reduce the calls of `NumPick`.
2. Move `index_select` into parallel section.
3. Use handwrite for loop instead of `torch::count_nonzero` (which seems to be unexpectedly slow).

The boost in the performance is obvious!

It's much faster in uniform cases:
UF2
UTF2

In non-uniform cases, at least it is not slower:
NF2
NTF2

/ | G1 | G2 | G3 | G4 | G5
:-:|:-:|:-:|:-:|:-:|:-:
New - NU | 224961.4 | 5647376.7 | 1490749.0 | 1287764.4 | 112337.3
Old - NU | 8201105.6 | 8971188.0 | 11911894.6 | 12674839.6 | 1231373.4
Compare | **+3545.6%** | +58.9% | **+699.1%** | **+884.3%** | **+996.1%**
New - NN | 29084539.2 | 16515189.3 | 38836236.9 | 17375014.6 | 4801887.6
Old - NN | 38688025.1 | 18178196.6 | 49979084.7 | 26732285.2 | 5280066.0
Compare | +33.0% | +10.1% | +28.7% | +53.9% | +10.0%
New - LU | 891378.8 | 5543386.5 | 2979804.6 | 1680911.2 | 169718.7
Old - LU | 9142654.2 | 7668949.0 | 10316067.7 | 11305912.7 | 1417829.8
Compare | **+925.7%** | +38.3% | **+246.2%** | **+572.6%** | **+735.4%**
New - LN | 956316.8 | 5706404.6 | 2955571.1 | 1797973.8 | 188544.5
Old - LN | 9430210.8 | 8193670.2 | 11730630.6 | 10311480.4 | 1404687.7
Compare | **+886.1%** | +43.6% | **+296.9%** | **+473.5%** | **+645.0%**

We can see:
- The improvement is relatively small when fanout is large.
- After these changes, the performance is much more stable.
- The major reason of Labor non-uniform sampling being slow is `NumPick` function. As it being rewritten, the performance is much faster. (The time costs of `NumPick` is now reduced by over 90% on G1).

Possible future plans:
- The performance of non-uniform cases could be further improved (will do this in a new PR).
- Test more cases to cover all the branches.
- Figure out how to improve when fanout is large.

## 8.15

UF
NF
UTF
NTF

### Data

The `Compare` row is the improvement in speed. Calculation: `old_time / new_time - 1`.

/ | G1 | G2 | G3 | G4 | G5
:-:|:-:|:-:|:-:|:-:|:-:
New - NU | 1087005.6 | 8554792.9 | 10564663.2 | 10950551.4 | 581512.8
Old - NU | 8201105.6 | 8971188.0 | 11911894.6 | 12674839.6 | 1231373.4
Compare | +654.5% | +4.9% | +12.8% | +15.7% | +111.8%
New - NN | 45285264.4 | 20179410.5 | 54786576.6 | 28771414.8 | 5741047.2
Old - NN | 38688025.1 | 18178196.6 | 49979084.7 | 26732285.2 | 5280066.0
Compare | -14.6% | -9.9% | -8.8% | -7.1% | -8.0%
New - LU | 1410470.6 | 7690707.0 | 11836302.6 | 10111700.3 | 667031.3
Old - LU | 9142654.2 | 7668949.0 | 10316067.7 | 11305912.7 | 1417829.8
Compare | +548.2%  | -0.3%  | -12.8%  | +11.8%  | +112.6%
New - LN | 24745363.5 | 7726521.9 | 12286540.3 | 10284233.2 | 1065888.7
Old - LN | 9430210.8 | 8193670.2 | 11730630.6 | 10311480.4 | 1404687.7
Compare | -61.9% | +6.0% | -4.5% | +0.3% | +31.8%

### Analysis

- The quantity of the time spent on the final `torch::cat` is actually relatively small (mostly the ratio is between 0.1% to 5%) .
- The process of calculating pick number in fact brings more time cost. Sometimes can take up to 45% of the total time, and also sometimes unstable.
- The processing in the end (i.e. the `index_select` part) is often the most time-consuming part. (It could take up almost 94% of the total time).
- The `pick` process is another major time consumer. In the cases where new version behaves worse than the old version, we can notice that `pick` process are taking more time. Especially when it's non-uniform.

### Conclusion

- The boost in speed is not much as expected because the `torch::cat` is actually not the major target.
- The situation varies A LOT according to different settings including:
    - Graph topology
    - Graph size
    - Neighbor or Labor algorithm
    - Replace or not
    - Uniform or not
    - Fanouts
    - etc.
- The `index_select` part needs to be improved. (Move it into the parallel section first).
- The `num_pick_fn` needs to be improved.
- The sampling process when non-uniform needs to be improved. (esp. the `multinomial` branch. I've already worked out a solution to improve the performance of this `nultinomial` branch)


peizhou001 (changes requested):
As this is a performance critical change, add performance data like #6030

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -252,40 +249,55 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
...
+
+              const auto nodes_data_ptr = nodes.data_ptr<int64_t>();
+              for (int64_t i = begin; i < end; ++i) {
+                const auto nid = nodes_data_ptr[i];
+                const auto offset = indptr_data[nid];
+                const auto num_neighbors = indptr_data[nid + 1] - offset;
+                const auto picked_number = num_picked_neighbors_data_ptr[i + 1];
+                const auto picked_offset = subgraph_indptr_data_ptr[i];
+                if (picked_number > 0) {
+                  pick_fn(
```
mfbalin, line 292: This is a minor comment, do we want to add an error-checking mechanism to see if `picked_number` is equal to the number of picked edges by `pick_fn`? If `pick_fn` returned the number of sampled edges, we could add a `TORCH_CHECK(picked_number == pick_fn(), "Runtime error, raise an issue at DGL!")`. If `pick_fn` somehow samples fewer or higher numbers of edges, it will cause a crash. However, as long as there is no bug, this will never happen of course.

mfbalin (approved):
LGTM, minor comments that might not even require addressing.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -252,40 +249,55 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
...
+                      offset, num_neighbors,
+                      picked_eids_data_ptr + picked_offset);
+                }
+              }
+            });
+      }));
+
+  // Step 5. Calculate other attributes and return the subgraph.
   torch::Tensor subgraph_indices =
```
peizhou001, line 301: We can consider pre-allocate the memory and move the `index_select` to multi-thread section if it's really slow after test. But this could be a future work.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -252,40 +249,55 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
...
+
+  // Step 2. Calculate prefix sum to get total length and offsets of each node.
+  // It's also the indptr of the generated subgraph.
   torch::Tensor subgraph_indptr =
       torch::cumsum(num_picked_neighbors_per_node, 0);
 
-  torch::Tensor picked_eids = torch::cat(picked_neighbors_per_thread);
+  // Step 3. Allocate the tensor for picked neighbors.
+  torch::Tensor picked_eids;
+  AT_DISPATCH_INTEGRAL_TYPES(
```
peizhou001, line 265: Can we combine this dispatch with Line234?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -252,40 +249,55 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
...
+      subgraph_indptr.scalar_type(), "GetTotalLength", ([&] {
+        const auto total_length =
+            subgraph_indptr.data_ptr<scalar_t>()[num_nodes];
+        picked_eids = torch::empty({total_length}, indptr_options);
+      }));
+
+  // Step 4. Pick neighbors for each node.
+  AT_DISPATCH_INTEGRAL_TYPES(
```
peizhou001, line 273: Ditto

graphbolt/src/csc_sampling_graph.cc
frozenbugs: +1

graphbolt/src/csc_sampling_graph.cc
RamonZhou: As I'm testing the performance, this `index_select` is sometimes the most time-consuming part and seems to be unstable in terms of performance. So I think it needs improvement.

mfbalin: Moving this inside the multi-thread section would increase labor performance by a lot as labor already accesses the indices array while sampling.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -222,28 +222,27 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
     const torch::Tensor& nodes, bool return_eids, NumPickFn num_pick_fn,
     PickFn pick_fn) const {
   const int64_t num_nodes = nodes.size(0);
-  const int64_t num_threads = torch::get_num_threads();
-  std::vector<torch::Tensor> picked_neighbors_per_thread(num_threads);
+  const auto indptr_options = indptr_.options();
   torch::Tensor num_picked_neighbors_per_node =
-      torch::zeros({num_nodes + 1}, indptr_.options());
+      torch::zeros({num_nodes + 1}, indptr_options);
```
mfbalin, line 227: We better use torch::empty here. Every index except 0 is already overwritten in the num_pick_fn. So we simply set the 0 index to zero manually.

mfbalin (commented):
Just a minor comment that might slightly improve performance.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -385,9 +420,17 @@ int64_t NumPick(
     int64_t num_neighbors) {
   int64_t num_valid_neighbors =
       probs_or_mask.has_value()
-          ? *torch::count_nonzero(
-                 probs_or_mask.value().slice(0, offset, offset + num_neighbors))
-                 .data_ptr<int64_t>()
+          ? [&] {
+            int64_t nonzero_count = 0;
```
frozenbugs, line 424: nit: Instead using inline method, create a inline function separately its more readable.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -385,9 +420,17 @@ int64_t NumPick(
...
+            AT_DISPATCH_ALL_TYPES(
+                probs_or_mask.value().scalar_type(), "NumPick", ([&] {
+                  scalar_t* probs_data_ptr = probs_or_mask.value().data_ptr<scalar_t>() + offset;
```
frozenbugs, line 427: exceed 80 chars.

mfbalin:
In theory, Neighbor should always be faster than Labor. So, it is possible to keep optimizing neighbor until it is faster than labor.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -385,9 +420,17 @@ int64_t NumPick(
...
+                  for (auto i = offset; i < offset + num_neighbors; ++ i, ++ probs_data_ptr) {
```
mfbalin, line 428: We can use std::count here, https://en.cppreference.com/w/cpp/algorithm/count
we can count how many 0s there are by passing 0 to value parameter of std::count and subtract from num_neighbors.

mfbalin (changes requested):

mfbalin (approved):
LGTM. I am curious about the final performance with the latest changes.

RamonZhou:
@mfbalin I've updated the description :)

frozenbugs (approved):

mfbalin:
> @mfbalin I've updated the description :)

Are the benchmark scripts going to be available? I am thinking of improving Labor performance in the future.

peizhou001 (approved):

PR #6161, opened 2023-08-16
[dev] configure optimize options explicitly for different build types

Rhett-Ying:
## Checklist
- [ ] Code is well-documented


CMakeLists.txt
```diff
@@ -108,10 +108,24 @@ dgl_feature_option(
...
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O0 -g3 -ggdb")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O0 -g3 -ggdb")
+  endif()
+else()
+  if (MSVC)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /O2")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /O2")
+  else()
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O3")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3")
```
frozenbugs, line 126: Shall we use -O2 or -O3 here?

frozenbugs:
@dgl-bot

CMakeLists.txt
Rhett-Ying: `/O2` is the most optimized on windows but `-O3` is the most optimized in gcc

frozenbugs: https://wiki.gentoo.org/wiki/GCC_optimization/en

frozenbugs: -O3: enables -O2 as well as optimizations that are expensive in terms of compile time and memory usage. Compiling with -O3 is not a guaranteed way to improve performance, and in fact, in many cases, can slow down a system due to larger binaries and increased memory usage. -O3 is also known to break several packages. Using -O3 is not recommended. However, it also enables -ftree-vectorize so that loops in the code get vectorized and will use AVX YMM registers.

frozenbugs: Unless we have strong reasons, let's stay on O2.

Rhett-Ying: source? I didn't see it in https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

Rhett-Ying: let's use `-O2` then.

frozenbugs (approved):

PR #6203, opened 2023-08-23
[Graphbolt] Improve Labor performance

mfbalin:
#6152 showed less than ideal speedups for labor for the G1 test case. Speedup of Neighbor was up to 36x but Labor was limited at 9x. I suspect this is because of the calls to torch tensor allocation calls. So this is my attempt at reducing those calls by utilizing the stack for small memory allocations.

@RamonZhou how does the labor performance look after this change?


mfbalin:
@Rhett-Ying The MSVC version in CI doesn't seem to play well with modern C++ features. I recommend updating the MSVC version in CI.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -781,10 +782,16 @@ inline int64_t LaborPick(
     std::iota(picked_data_ptr, picked_data_ptr + num_neighbors, offset);
     return num_neighbors;
   }
-  torch::Tensor heap_tensor = torch::empty({fanout * 2}, torch::kInt32);
+#define STACK_SIZE_LABOR 1024
```
frozenbugs, line 785:
```suggestion
  static const int kLaborStackSize = 1024;
```

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -781,10 +782,16 @@ inline int64_t LaborPick(
...
   // Assuming max_degree of a vertex is <= 4 billion.
-  auto heap_data = reinterpret_cast<std::pair<float, uint32_t>*>(
-      heap_tensor.data_ptr<int32_t>());
+  std::array<std::pair<float, uint32_t>, STACK_SIZE_LABOR> heap_stack;
```
frozenbugs, line 787: just stack? heap? it is a little bit confusing. Maybe a more descriptive name?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -814,9 +821,14 @@ inline int64_t LaborPick(
           // is O((fanout + num_neighbors) log(fanout)). It is possible to
           // decrease the logarithmic factor down to
           // O(log(min(fanout, num_neighbors))).
-          torch::Tensor remaining =
-              torch::ones({num_neighbors}, torch::kFloat32);
-          float* rem_data = remaining.data_ptr<float>();
+          std::array<float, STACK_SIZE_LABOR> remaining_stack;
+          auto rem_data = remaining_stack.data();
```
frozenbugs, line 825:
```suggestion
          auto remaining_data = remaining_stack.data();
```

unnecessary abbr.

frozenbugs (approved):

RamonZhou:
I've tested the performance.

- `v0`: The original one.
- `v1`: After #6152 
- `v2`: After this PR.

L-UF
L-NF
L-UT
L-NT

Data:
data

There is a big boost in G1.
The small drops in other graphs may be fluctuations.

graphbolt/src/csc_sampling_graph.cc
mfbalin: I originally defined it as constexpr int, but msvc in CI doesn't support it properly. It is used as a template parameter for std::array so it needs to be defined at compile time.

graphbolt/src/csc_sampling_graph.cc
mfbalin: Storage is from stack but the datastructure is heap. Maybe heap_from_stack?

frozenbugs: Just heap should be enough, and L788 makes more sense `heap_data = heap.data();`

graphbolt/src/csc_sampling_graph.cc
frozenbugs: Weird, if that's the case, move the `#define` to the top, after L23. 

`#define kLaborStackSize  1024`

mfbalin:
@frozenbugs My goal is to keep Labor performance at parity with Neighbor as much as possible. Labor actually does a bit more work, always O(num_neighbors) compared to Neighbor possibly doing O(fanout), so it may not be possible for all use cases (high average degree graphs). Since I am proposing Labor to replace Neighbor, I don't want potential future users to say this is a complicated algorithm so it is bound to be slow. They keep saying this in my paper reviews all the time.

What would be nice is to give this algorithm a bit more visibility in examples etc. As it is, the API and behavior fully match the Neighbor API so one can go ahead and use this instead of Neighbor in some examples. I would love the feedback if somehow the end accuracy or result doesn't match the Neighbor baseline, which I could then investigate, but I don't expect such a thing to happen. When sampling multiple layers, Labor will sample significantly fewer vertices and edges which will speed up convergence.

In the future, Labor will support more advanced features which are already supported by the `dgl.sampling.sample_labors`. I am sure it will all look very fancy when we announce all these supported features to the wider GNN community.

graphbolt/include/graphbolt/csc_sampling_graph.h
```diff
@@ -406,8 +406,8 @@ int64_t PickByEtype(
     PickedType* picked_data_ptr);
 
 template <
-    bool NonUniform, bool Replace, typename ProbsType = float,
-    typename PickedType>
+    bool NonUniform, bool Replace, typename ProbsType, typename PickedType,
```
mfbalin, line 409: C++ requires template parameters with default values to come after the other template parameters. So, it is necessary to delete the default from ProbsType. I don't know how it was compiling before.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -781,10 +783,15 @@ inline int64_t LaborPick(
     std::iota(picked_data_ptr, picked_data_ptr + num_neighbors, offset);
     return num_neighbors;
   }
-  torch::Tensor heap_tensor = torch::empty({fanout * 2}, torch::kInt32);
   // Assuming max_degree of a vertex is <= 4 billion.
-  auto heap_data = reinterpret_cast<std::pair<float, uint32_t>*>(
-      heap_tensor.data_ptr<int32_t>());
+  std::array<std::pair<float, uint32_t>, StackSize> heap;
```
Rhett-Ying, line 787: so the main change here is use stack memory instead of heap memory for small size to achieve the improvement?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -781,10 +783,15 @@ inline int64_t LaborPick(
...
+  auto heap_data = heap.data();
+  torch::Tensor heap_tensor;
+  if (fanout > StackSize) {
+    heap_tensor = torch::empty({fanout * 2}, torch::kInt32);
```
Rhett-Ying, line 791: Is this memory safe? `heap_tensor` holds the memory and it ends it's life outside of the `if` statement. Such memory will be deleted automatically even the data pointer is saved to `heap_data`?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -781,10 +783,15 @@ inline int64_t LaborPick(
...
+    heap_data = reinterpret_cast<std::pair<float, uint32_t>*>(
```
Rhett-Ying, line 792: Are we assuming the memory size allocated by `torch::empty` equals `fanout * 2 * sizeof(float) `?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -781,10 +783,15 @@ inline int64_t LaborPick(
...
+        heap_tensor.data_ptr<int32_t>());
```
Rhett-Ying, line 793: `float`, `uint32_t`, `int32_t` are used in different places while we always assume all of them are in same size, namely `4` bytes. We may need a more secure way to violate it if someone else change one of them to different size such as `int64_t` some day.

graphbolt/src/csc_sampling_graph.cc
mfbalin: Yes

graphbolt/src/csc_sampling_graph.cc
mfbalin: heap_tensor is defined outside of if.

graphbolt/src/csc_sampling_graph.cc
mfbalin: we are assuming it equals fanout int32_t and fanout floats.

graphbolt/src/csc_sampling_graph.cc
mfbalin: This is already secure. My only assumption is that a vertex doesn't have more than 4 billion neighbors. There is no dataset that I know where a vertex has more than 4 billion neighbors.

mfbalin: I can change the code so that it allocates a tensor with enough bytes in guaranteed manner even if the types change. Let me do that.

graphbolt/src/csc_sampling_graph.cc
Rhett-Ying: utilizing `torch:empty` instead of `new` or `std::vector` is to ask torch to help us manage memory better?

mfbalin: Yeah, I thought that since we are relying on torch for everything, we can do the same here. Not sure if torch allocation is faster or new is faster.

graphbolt/src/csc_sampling_graph.cc
mfbalin: Just made a change.

Rhett-Ying (approved):

Rhett-Ying:
> @Rhett-Ying The MSVC version in CI doesn't seem to play well with modern C++ features. I recommend updating the MSVC version in CI.

it's VS2019 and may not play well with C++17 features.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -781,10 +783,16 @@ inline int64_t LaborPick(
...
-  torch::Tensor heap_tensor = torch::empty({fanout * 2}, torch::kInt32);
   // Assuming max_degree of a vertex is <= 4 billion.
-  auto heap_data = reinterpret_cast<std::pair<float, uint32_t>*>(
-      heap_tensor.data_ptr<int32_t>());
+  std::array<std::pair<float, uint32_t>, StackSize> heap;
+  auto heap_data = heap.data();
+  torch::Tensor heap_tensor;
+  if (fanout > StackSize) {
+    heap_tensor = torch::empty(
+        {fanout * (sizeof(heap_data[0]) / sizeof(int32_t))}, torch::kInt32);
```
Rhett-Ying: seems cause error when building on windows, `2` is ok for me as we explicitly declare `std::pair` above. Or try with `std::tuple_size<decltype(heap)::value_type>::value`?

mfbalin: I just fixed it. It was related to casting of unsigned int to int.

mfbalin:
We also could allocate the required memory inside the parallel for section but that seemed like a lot of code change. I didn't want to make the code more complicated so I resorted to using the stack as a quick hack.

mfbalin:
I am pretty happy with the current performance of this code. Made a last-minute minor optimization. Feel free to merge anytime.
ogbn-products dataset 20 fanout, 200k nodes:
```
{('NS', False, False): 13.961718496284448,
 ('LABOR', False, False): 22.257723700022325,
 ('NS', False, True): 264.3265006976435,
 ('LABOR', False, True): 27.74477390339598,
 ('NS', True, False): 198.85092989716213,
 ('LABOR', True, False): 47.24292550235987,
 ('NS', True, True): 1236.366263902164,
 ('LABOR', True, True): 54.78198000055272}
```
Benchmarked with following code:
```
import torch
import dgl.graphbolt as gb
from ogb.nodeproppred import DglNodePropPredDataset
# g = gb.from_dglgraph(RedditDataset()[0])
g = gb.from_dglgraph(DglNodePropPredDataset('ogbn-products')[0][0])
g = gb.from_csc(g.csc_indptr, g.indices, None, None, {'probs': torch.rand([g.num_edges])}, g.metadata)
from timeit import default_timer as timer
from collections import defaultdict
times = defaultdict(float)
trials = 10
for z in range(trials):
    nodes = torch.randperm(g.num_nodes)[:200000]
    # nodes = torch.arange(g.num_nodes)
    for replace in [False, True]:
        for w in [None, 'probs']:
            for f, name in zip([g.sample_neighbors, g.sample_layer_neighbors], ["NS", "LABOR"]):
                st = timer()
                out = f(nodes, torch.tensor([20]), replace, probs_name=w)
                t = timer() - st
                times[name, replace, w is not None] += t
print({k: v / trials * 1000 for k, v in times.items()})
```

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -781,10 +783,16 @@ inline int64_t LaborPick(
...
     return num_neighbors;
   }
-  torch::Tensor heap_tensor = torch::empty({fanout * 2}, torch::kInt32);
   // Assuming max_degree of a vertex is <= 4 billion.
-  auto heap_data = reinterpret_cast<std::pair<float, uint32_t>*>(
-      heap_tensor.data_ptr<int32_t>());
+  std::array<std::pair<float, uint32_t>, StackSize> heap;
+  auto heap_data = heap.data();
+  torch::Tensor heap_tensor;
+  if (fanout > StackSize) {
```
peizhou001, line 790: Is that means `fanout > StackSize`, memory allocated in Line 787 is wasted?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -814,41 +822,62 @@ inline int64_t LaborPick(
...
-              torch::ones({num_neighbors}, torch::kFloat32);
-          float* rem_data = remaining.data_ptr<float>();
+          std::array<float, StackSize> remaining;
+          auto remaining_data = remaining.data();
+          torch::Tensor remaining_tensor;
+          if (num_neighbors > StackSize) {
+            remaining_tensor = torch::empty({num_neighbors}, torch::kFloat32);
+            remaining_data = remaining_tensor.data_ptr<float>();
+          }
+          std::fill(remaining_data, remaining_data + num_neighbors, 1.f);
```
peizhou001, line 832: Will there have some gains to use memset?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -814,41 +822,62 @@ inline int64_t LaborPick(
...
           // decrease the logarithmic factor down to
           // O(log(min(fanout, num_neighbors))).
-          torch::Tensor remaining =
-              torch::ones({num_neighbors}, torch::kFloat32);
-          float* rem_data = remaining.data_ptr<float>();
+          std::array<float, StackSize> remaining;
+          auto remaining_data = remaining.data();
+          torch::Tensor remaining_tensor;
+          if (num_neighbors > StackSize) {
+            remaining_tensor = torch::empty({num_neighbors}, torch::kFloat32);
```
peizhou001, line 829: Ditto.

graphbolt/src/csc_sampling_graph.cc
RamonZhou: `memset` seems not able to do the same thing here. It fills the memory by bytes.

graphbolt/src/csc_sampling_graph.cc
mfbalin: What do you mean here?

mfbalin:
Can a significant portion of edge weights be zero? Should I optimize considering this case as well in the future? Will users use edge weights to separate train, validation or test edges etc.?

frozenbugs:
> @frozenbugs @Rhett-Ying @peizhou001 Can a significant portion of edge weights be zero? Should I optimize considering this case as well in the future? Will users use edge weights to separate train, validation or test edges etc. by setting a lot of edge weights to 0?

It is hard to predict how users will use the edge weight in a very specific way, but we usually only optimize for common usage, if a user pattern become a common pattern, we can always optimize it later.

peizhou001 (approved):

PR #6208, opened 2023-08-25
[Sparse] Add sparse matrix slicing operator implementation

xiangyuzhi:
## Description
Sparse slice operator implementation, include index_select and range_select.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -138,21 +138,52 @@ class SparseMatrix : public torch::CustomClassHolder {
       torch::Tensor value, const std::vector<int64_t>& shape);
 
   /**
-   * @brief Create a SparseMatrix by index slicing.
-   * @param dim Row-wise or column-wise select
-   * @param ids Ids of the selected rows or columns
+   * @brief Create a SparseMatrix by selecting rows or columns based on provided
+   * indices.
    *
-   * @return SparseMatrix
+   * This function allows you to create a new SparseMatrix by selecting specific
+   * rows or columns from the original SparseMatrix based on the provided
+   * indices. The selection can be performed either row-wise or column-wise,
+   * determined by the 'dim' parameter.
+   *
+   * @param dim Select rows (dim=0) or columns (dim=1).
+   * @param ids A tensor containing the indices of the selected rows or columns.
+   *
+   * @return A new SparseMatrix containing the selected rows or columns.
+   *
+   * @note The 'dim' parameter should be either 0 (for row-wise selection) or 1
+   * (for column-wise selection).
+   * @note The 'ids' tensor should contain valid indices within the range of the
+   * original SparseMatrix's dimensions.
+   * @note The returned SparseMatrix may share data with the original matrix,
+   * depending on the implementation.
    */
   c10::intrusive_ptr<SparseMatrix> IndexSelect(int64_t dim, torch::Tensor ids);
 
   /**
-   * @brief Create a SparseMatrix by range slicing.
-   * @param dim Row-wise or column-wise select
-   * @param start Inclusive start id of the selected rows or columns
-   * @param end Exclusive end id of the selected rows or columns
+   * @brief Create a SparseMatrix by selecting a range of rows or columns based
+   * on provided indices.
    *
-   * @return SparseMatrix
+   * This function allows you to create a new SparseMatrix by selecting a range
+   * of specific rows or columns from the original SparseMatrix based on the
+   * provided indices. The selection can be performed either row-wise or
+   * column-wise, determined by the 'dim' parameter.
+   *
+   * @param dim Select rows (dim=0) or columns (dim=1).
+   * @param start The starting index of the range.
+   * @param end The ending index (exclusive) of the range.
+   *
+   * @return A new SparseMatrix containing the selected range of rows or
+   * columns.
+   *
+   * @note The 'dim' parameter should be either 0 (for row-wise selection) or 1
+   * (for column-wise selection).
+   * @note The 'start' and 'end' parameters define the inclusive-exclusive range
+   * of indices to select.
+   * @note The function ensures that the 'start' and 'end' indices are within
+   * the valid range of the original SparseMatrix's dimensions.
+   * @note The returned SparseMatrix may share data with the original matrix,
+   * depending on the implementation.
    */
   c10::intrusive_ptr<SparseMatrix> RangeSelect(
       int64_t dim, int64_t start, int64_t end);
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -155,8 +155,6 @@ class SparseMatrix : public torch::CustomClassHolder {
    * (for column-wise selection).
    * @note The 'ids' tensor should contain valid indices within the range of the
    * original SparseMatrix's dimensions.
-   * @note The returned SparseMatrix may share data with the original matrix,
-   * depending on the implementation.
    */
   c10::intrusive_ptr<SparseMatrix> IndexSelect(int64_t dim, torch::Tensor ids);
 
@@ -170,7 +168,7 @@ class SparseMatrix : public torch::CustomClassHolder {
    * column-wise, determined by the 'dim' parameter.
    *
    * @param dim Select rows (dim=0) or columns (dim=1).
-   * @param start The starting index of the range.
+   * @param start The starting index (inclusive) of the range.
    * @param end The ending index (exclusive) of the range.
    *
    * @return A new SparseMatrix containing the selected range of rows or
@@ -178,12 +176,8 @@ class SparseMatrix : public torch::CustomClassHolder {
    *
    * @note The 'dim' parameter should be either 0 (for row-wise selection) or 1
    * (for column-wise selection).
-   * @note The 'start' and 'end' parameters define the inclusive-exclusive range
-   * of indices to select.
-   * @note The function ensures that the 'start' and 'end' indices are within
+   * @note The 'start' and 'end' indices should be valid indices within
    * the valid range of the original SparseMatrix's dimensions.
-   * @note The returned SparseMatrix may share data with the original matrix,
-   * depending on the implementation.
    */
   c10::intrusive_ptr<SparseMatrix> RangeSelect(
       int64_t dim, int64_t start, int64_t end);
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -149,8 +149,8 @@ class SparseMatrix : public torch::CustomClassHolder {
   /**
    * @brief Create a SparseMatrix by range slicing.
    * @param dim Row-wise or column-wise select
-   * @param start Start id of the selected rows or columns
-   * @param end End id of the selected rows or columns
+   * @param start Inclusive start id of the selected rows or columns
+   * @param end Exclusive end id of the selected rows or columns
    *
    * @return SparseMatrix
    */
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -138,40 +138,24 @@ class SparseMatrix : public torch::CustomClassHolder {
       torch::Tensor value, const std::vector<int64_t>& shape);
 
   /**
-   * @brief Create a SparseMatrix by slicing rows.
+   * @brief Create a SparseMatrix by index slicing.
+   * @param dim Row-wise or column-wise select.
    * @param ids Ids of the selected rows
    *
    * @return SparseMatrix
    */
-  c10::intrusive_ptr<SparseMatrix> RowwiseSelect(torch::Tensor ids);
-
-  //   /**
-  //    * @brief Create a SparseMatrix by slicing range rows.
-  //    * @param start Start id of the selected rows.
-  //    * @param end End id of the selected rows.
-  //    *
-  //    * @return SparseMatrix
-  //    */
-  //   c10::intrusive_ptr<SparseMatrix> RowwiseSelect(int64_t start, int64_t
-  //   end);
+  c10::intrusive_ptr<SparseMatrix> IndexSelect(int64_t dim, torch::Tensor ids);
 
   /**
-   * @brief Create a SparseMatrix by slicing columns.
-   * @param ids Ids of the selected columns
+   * @brief Create a SparseMatrix by range slicing.
+   * @param dim Row-wise or column-wise select.
+   * @param start Start id of the selected columns.
+   * @param end End id of the selected columns.
    *
    * @return SparseMatrix
    */
-  c10::intrusive_ptr<SparseMatrix> ColumnwiseSelect(torch::Tensor ids);
-
-  //   /**
-  //    * @brief Create a SparseMatrix by slicing columns.
-  //    * @param start Start id of the selected columns.
-  //    * @param end End id of the selected columns.
-  //    *
-  //    * @return SparseMatrix
-  //    */
-  //   c10::intrusive_ptr<SparseMatrix> ColumnwiseSelect(int64_t start, int64_t
-  //   end);
+  c10::intrusive_ptr<SparseMatrix> RangeSelect(
+      int64_t dim, int64_t start, int64_t end);
 
   /**
    * @brief Create a SparseMatrix from a SparseMatrix using new values.
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -138,6 +138,45 @@ class SparseMatrix : public torch::CustomClassHolder {
       torch::Tensor value, const std::vector<int64_t>& shape);
 
   /**
+   * @brief Create a SparseMatrix by slicing rows.
+   * @param ids Ids of the selected rows
+   *
+   * @return SparseMatrix
+   */
+  c10::intrusive_ptr<SparseMatrix> RowwiseSelect(
+    torch::Tensor ids);
+
+//   /**
+//    * @brief Create a SparseMatrix by slicing columns.
+//    * @param ids Ids of the selected columns
+//    *
+//    * @return SparseMatrix
+//    */
+//   static c10::intrusive_ptr<SparseMatrix> ColumnwiseSelect(
+//       torch::Tensor ids);
+
+//   /**
+//    * @brief Create a SparseMatrix by slicing range rows.
+//    * @param start Start id of the selected rows.
+//    * @param end End id of the selected rows.
+//    *
+//    * @return SparseMatrix
+//    */
+//   static c10::intrusive_ptr<SparseMatrix> RowwiseSelect(
+//       int64_t start, int64_t end);
+
+//   /**
+//    * @brief Create a SparseMatrix by slicing columns.
+//    * @param start Start id of the selected columns.
+//    * @param end End id of the selected columns.
+//    *
+//    * @return SparseMatrix
+//    */
+//   static c10::intrusive_ptr<SparseMatrix> ColumnwiseSelect(
+//       int64_t start, int64_t end);
+
+
+  /**
    * @brief Create a SparseMatrix from a SparseMatrix using new values.
    * @param mat An existing sparse matrix
    * @param value New values of the sparse matrix
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -143,38 +143,35 @@ class SparseMatrix : public torch::CustomClassHolder {
    *
    * @return SparseMatrix
    */
-  c10::intrusive_ptr<SparseMatrix> RowwiseSelect(
-    torch::Tensor ids);
+  c10::intrusive_ptr<SparseMatrix> RowwiseSelect(torch::Tensor ids);
 
-//   /**
-//    * @brief Create a SparseMatrix by slicing columns.
-//    * @param ids Ids of the selected columns
-//    *
-//    * @return SparseMatrix
-//    */
-//   static c10::intrusive_ptr<SparseMatrix> ColumnwiseSelect(
-//       torch::Tensor ids);
+  //   /**
+  //    * @brief Create a SparseMatrix by slicing range rows.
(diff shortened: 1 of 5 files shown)


dgl_sparse/src/python_binding.cc
```diff
@@ -33,7 +33,9 @@ TORCH_LIBRARY(dgl_sparse, m) {
       .def("transpose", &SparseMatrix::Transpose)
       .def("coalesce", &SparseMatrix::Coalesce)
       .def("has_duplicate", &SparseMatrix::HasDuplicate)
-      .def("is_diag", &SparseMatrix::HasDiag);
+      .def("is_diag", &SparseMatrix::HasDiag)
+      .def("rowwise_select", &SparseMatrix::RowwiseSelect);
```
czkkkkkk, line 37: Install `clang-format` in Python environment and use vscode to format your C++ code.

dgl_sparse/src/sparse_matrix.cc
```diff
@@ -122,6 +124,22 @@ c10::intrusive_ptr<SparseMatrix> SparseMatrix::FromDiag(
...
+  auto new_shape = std::vector<int64_t>{id_array.GetSize(), this->shape()[1]};
+  if (HasCSR() || !HasCOO()) {
+    auto dgl_csr = CSRToOldDGLCSR(this->CSRPtr());
+    auto slice_csr = dgl::aten::CSRSliceRows(dgl_csr, id_array);
+    auto slice_value = DGLArrayToTorchTensor(dgl::aten::IndexSelect(TorchTensorToDGLArray(this->value()), slice_csr.data));
+    return SparseMatrix::FromCSRPointer(CSRFromOldDGLCSR(slice_csr), slice_value, new_shape);
+  } else { // COO
+    auto dgl_coo = COOToOldDGLCOO(this->COOPtr());
+    auto slice_coo = dgl::aten::COOSliceRows(dgl_coo, id_array);
+    auto slice_value = DGLArrayToTorchTensor(dgl::aten::IndexSelect(TorchTensorToDGLArray(this->value()), slice_coo.data));
```
czkkkkkk, line 138: Pls use the index_select operator of Pytorch.

czkkkkkk (commented):
LGTM. Pls implement other interfaces in the same fasion.

dgl_sparse/src/python_binding.cc
xiangyuzhi: Applied.

dgl_sparse/src/sparse_matrix.cc
xiangyuzhi: Applied.

dgl_sparse/src/sparse_matrix.cc
```diff
@@ -122,6 +124,99 @@ c10::intrusive_ptr<SparseMatrix> SparseMatrix::FromDiag(
   return SparseMatrix::FromDiagPointer(diag, value, shape);
 }
 
+c10::intrusive_ptr<SparseMatrix> SparseMatrix::IndexSelect(
+    int64_t dim, torch::Tensor ids) {
+  auto id_array = TorchTensorToDGLArray(ids);
+  if (dim == 0) {  // row
```
czkkkkkk, line 130: The logic of dim = 0 and 1 are very similar. You can merge them for simplification. Basically, they all get the CSR/CSC/COO matrices, call DGL APIs and convert the result to sparse matrix. You only need to add some transpose operators according to the dim.

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -450,6 +450,52 @@ def test_has_duplicate():
     assert csc_A.has_duplicate()
 
 
+def test_select():
```
czkkkkkk, line 453: Test `index_select` and `range_select` independently

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -450,6 +450,52 @@ def test_has_duplicate():
...
+    ctx = F.ctx()
+
+    row = torch.tensor([0, 1, 1, 2, 3, 4]).to(ctx)
```
czkkkkkk, line 456: Hardcode test cases only cover very limited scenarios. You can try to parameterize the test cases. https://github.com/dmlc/dgl/blob/ec8225df4ab164cc37522fd1a903ba00d1aed26c/tests/python/pytorch/sparse/test_sparse_matrix.py#L88 is a good example. And you also need to use some other tool to produce the correct result for each case. You can try to convert your results to pytorch dense tensor and verify the correctness by Pytorch dense index_select.

czkkkkkk (changes requested):

tests/python/pytorch/sparse/test_sparse_matrix.py
xiangyuzhi: ok, I have extended tests by generating correct results automatically. please review.

dgl_sparse/src/sparse_matrix.cc
```diff
@@ -122,6 +124,63 @@ c10::intrusive_ptr<SparseMatrix> SparseMatrix::FromDiag(
   return SparseMatrix::FromDiagPointer(diag, value, shape);
 }
 
+c10::intrusive_ptr<SparseMatrix> SparseMatrix::IndexSelect(
+    int64_t dim, torch::Tensor ids) {
+  auto id_array = TorchTensorToDGLArray(ids);
+  if ((!dim && HasCSR()) || (dim && HasCSC()) || !HasCOO()) {
```
czkkkkkk, line 130: Define `bool rowwise = dim==0` in advance to make the later `if` conditions easier to understand.

dgl_sparse/src/sparse_matrix.cc
```diff
@@ -122,6 +124,63 @@ c10::intrusive_ptr<SparseMatrix> SparseMatrix::FromDiag(
...
+    auto csr = dim == 0 ? this->CSRPtr() : this->CSCPtr();
+    auto slice_csr = dgl::aten::CSRSliceRows(CSRToOldDGLCSR(csr), id_array);
+    auto slice_value =
+        this->value().index_select(0, DGLArrayToTorchTensor(slice_csr.data));
+    slice_csr.data = dgl::aten::NullArray();
```
czkkkkkk, line 135: Why do you explicitly reset the array here.

dgl_sparse/src/sparse_matrix.cc
```diff
@@ -122,6 +124,63 @@ c10::intrusive_ptr<SparseMatrix> SparseMatrix::FromDiag(
...
+    auto ret = CSRFromOldDGLCSR(slice_csr);
+    auto new_shape = dim == 0
+                         ? std::vector<int64_t>{ret->num_rows, ret->num_cols}
+                         : std::vector<int64_t>{ret->num_cols, ret->num_rows};
+    return dim == 0 ? SparseMatrix::FromCSRPointer(ret, slice_value, new_shape)
+                    : SparseMatrix::FromCSCPointer(ret, slice_value, new_shape);
```
czkkkkkk, line 141:
```suggestion
    if (rowwise) {
      return SparseMatrix::FromCSRPointer(ret, slice_value, {ret->num_rows, ret->num_cols});
    } else {
      return SparseMatrix::FromCSCPointer(ret, slice_value, {ret->num_cols, ret->num_rows);
    }
```

dgl_sparse/src/sparse_matrix.cc
czkkkkkk: Better to define the used sparse format earlier to simply this if condition.

```cpp
SparseFormat sparse_format = ...
if (sparse_format == SparseFormat::kCSR || sparse_format == SparseFormat::kCSC) {
...
}
```

dgl_sparse/src/sparse_matrix.cc
```diff
@@ -122,6 +124,63 @@ c10::intrusive_ptr<SparseMatrix> SparseMatrix::FromDiag(
...
+    if (dim == 1) slice_coo = aten::COOTranspose(slice_coo);
+    auto ret = COOFromOldDGLCOO(slice_coo);
+    auto new_shape = std::vector<int64_t>{ret->num_rows, ret->num_cols};
+    return SparseMatrix::FromCOOPointer(ret, slice_value, new_shape);
+  }
+}
+
+c10::intrusive_ptr<SparseMatrix> SparseMatrix::RangeSelect(
+    int64_t dim, int64_t start, int64_t end) {
+  if ((!dim && HasCSR()) || (dim && HasCSC()) || !HasCOO()) {
```
czkkkkkk, line 158: ditto.

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -450,6 +450,146 @@ def test_has_duplicate():
     assert csc_A.has_duplicate()
 
 
+@pytest.mark.parametrize("dense_dim", [None, 4])
+@pytest.mark.parametrize("row", [(0, 1, 1, 2, 3, 4), (0, 1, 2, 3, 4, 4)])
+@pytest.mark.parametrize("col", [(0, 2, 4, 3, 5, 0), (1, 2, 5, 3, 0, 1)])
+@pytest.mark.parametrize("dim", [0, 1])
+@pytest.mark.parametrize("index", [(0, 1, 4), (1, 1, 3)])
+def test_index_select(dense_dim, row, col, dim, index):
```
czkkkkkk, line 458: Please follow https://github.com/dmlc/dgl/blob/f281959ac70fb50a96a8b243133c9d013ccd3e9f/tests/python/pytorch/sparse/test_matmul.py#L32 to test different sparse formats. https://github.com/dmlc/dgl/blob/f281959ac70fb50a96a8b243133c9d013ccd3e9f/tests/python/pytorch/sparse/utils.py#L139 also provides utility function to convert a sparse matrix to a pytorch dense tensor.

czkkkkkk (changes requested):

dgl_sparse/src/sparse_matrix.cc
xiangyuzhi: Since when we use SparseMatrix::FromCOOPointer(coo, value, shape) to construct a COO matrix, the origin 'coo' can not contain a data array, otherwise it will occur an error. In the CSR format, it will not occur this error immediately, but if it call coo() function later, sparse will use the csr format to construct COO, input the csr data array, and occur this error. 
For example, if we use csr to constrcut a sparse and then want to see the row and col attributes of it, sparse will convert it to COO format and occur an error.

dgl_sparse/src/sparse_matrix.cc
xiangyuzhi: The sparse format may contain different formats at the same time, I have no idea whether sparse_format is suitable.

tests/python/pytorch/sparse/test_sparse_matrix.py
xiangyuzhi: I have modified the code. However, I find there are some problems with current DGL COO slice func, that is, it dose not support deplicate select ids: https://github.com/dmlc/dgl/blob/master/src/array/cpu/spmat_op_impl_coo.cc#L799-L820 
It uses a hashmap to store ids and scans rows once, which is different with the implementation of CSR slice: 
https://github.com/dmlc/dgl/blob/master/src/array/cpu/spmat_op_impl_csr.cc#L384-L486

dgl_sparse/src/sparse_matrix.cc
czkkkkkk: I mean you can simplify the code here. This if condition is not readable. You change the code like this
```cpp
SparseFormat sparse_format;
if (rowwise) {
  if (HasCSR() || !HasCOO()) {
    sparse_format = SparseFormat::kCSR;
  }
  else {
    sparse_format = SparseFormat::kCOO;
  }
} else {
  ...
}
```

We may have a utility function for deciding the format later.

tests/python/pytorch/sparse/test_sparse_matrix.py
czkkkkkk: That is a problem. You can only support slicing on CSR/CSC for now. If we think slicing on COO is important, we can implement it by sparse native operators.

dgl_sparse/src/sparse_matrix.cc
czkkkkkk: Got it. Then it is better to convert it to the new sparse format and then remove the index array. Please also explain the reason for doing this in the code.

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -450,6 +452,48 @@ def test_has_duplicate():
     assert csc_A.has_duplicate()
 
 
+@pytest.mark.parametrize("create_func", [rand_coo, rand_csr, rand_csc])
```
czkkkkkk, line 455: Also test `rand_diag`.

tests/python/pytorch/sparse/test_sparse_matrix.py
xiangyuzhi: OK, I just remove the coo test, since it supports neither duplicate ids nor cuda operations.

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -450,6 +452,48 @@ def test_has_duplicate():
...
+    dense = sparse_matrix_to_dense(A)
+    dense_select = torch.index_select(dense, select_dim, index)
+
+    A_select_to_dense = sparse_matrix_to_dense(A_select)
+
+    assert A_select_to_dense.shape == dense_select.shape
+    assert torch.allclose(A_select_to_dense, dense_select)
+
+
+@pytest.mark.parametrize("create_func", [rand_diag, rand_csr, rand_csc])
```
czkkkkkk, line 475: Testing on COO is necessary. Your should not perform select on the COO format. If the sparse matrix only has COO format, it should convert it to CSR/COO for slicing.

dgl_sparse/include/sparse/sparse_matrix.h
```diff
@@ -137,6 +137,26 @@ class SparseMatrix : public torch::CustomClassHolder {
   static c10::intrusive_ptr<SparseMatrix> FromDiag(
       torch::Tensor value, const std::vector<int64_t>& shape);
 
+  /**
+   * @brief Create a SparseMatrix by index slicing.
```
czkkkkkk, line 141: Could you ask chatgpt to polish the documentation? Here is what I got from it.
```cpp
/**
 * @brief Create a SparseMatrix by selecting rows or columns based on provided indices.
 *
 * This function allows you to create a new SparseMatrix by selecting specific rows
 * or columns from the original SparseMatrix based on the provided indices. The selection
 * can be performed either row-wise or column-wise, determined by the 'dim' parameter.
 *
 * @param dim Select rows (dim=0) or columns (dim=1).
 * @param ids A tensor containing the indices of the selected rows or columns.
 *
 * @return A new SparseMatrix containing the selected rows or columns.
 *
 * @note The 'dim' parameter should be either 0 (for row-wise selection) or 1 (for column-wise selection).
 * @note The 'ids' tensor should contain valid indices within the range of the original SparseMatrix's dimensions.
 * @note The returned SparseMatrix may share data with the original matrix, depending on the implementation.
 */
c10::intrusive_ptr<SparseMatrix> IndexSelect(int64_t dim, torch::Tensor ids);
```

xiangyuzhi: Sure, do you think yours is ok? I can just use it.

czkkkkkk: You still need to carefully check the content from chatgpt.

dgl_sparse/include/sparse/sparse_matrix.h
```diff
@@ -137,6 +137,57 @@ class SparseMatrix : public torch::CustomClassHolder {
...
+   * @param dim Select rows (dim=0) or columns (dim=1).
+   * @param ids A tensor containing the indices of the selected rows or columns.
+   *
+   * @return A new SparseMatrix containing the selected rows or columns.
+   *
+   * @note The 'dim' parameter should be either 0 (for row-wise selection) or 1
+   * (for column-wise selection).
+   * @note The 'ids' tensor should contain valid indices within the range of the
+   * original SparseMatrix's dimensions.
+   * @note The returned SparseMatrix may share data with the original matrix,
```
czkkkkkk, line 158: Remove this line. Your implementation would not share the data.

dgl_sparse/include/sparse/sparse_matrix.h
```diff
@@ -137,6 +137,57 @@ class SparseMatrix : public torch::CustomClassHolder {
...
+   * @return A new SparseMatrix containing the selected range of rows or
+   * columns.
+   *
+   * @note The 'dim' parameter should be either 0 (for row-wise selection) or 1
+   * (for column-wise selection).
+   * @note The 'start' and 'end' parameters define the inclusive-exclusive range
+   * of indices to select.
+   * @note The function ensures that the 'start' and 'end' indices are within
+   * the valid range of the original SparseMatrix's dimensions.
+   * @note The returned SparseMatrix may share data with the original matrix,
```
czkkkkkk, line 185: ditto.

dgl_sparse/src/sparse_matrix.cc
```diff
@@ -122,6 +124,57 @@ c10::intrusive_ptr<SparseMatrix> SparseMatrix::FromDiag(
...
+c10::intrusive_ptr<SparseMatrix> SparseMatrix::IndexSelect(
+    int64_t dim, torch::Tensor ids) {
+  auto id_array = TorchTensorToDGLArray(ids);
+  bool rowwise = dim == 0;
+  SparseFormat sparse_format;
+  if (rowwise) {
+    sparse_format = SparseFormat::kCSR;
+  } else {
+    sparse_format = SparseFormat::kCSC;
+  }
```
czkkkkkk, line 136:
```suggestion
  SparseFormat sparse_format = rowwise? SparseFormat::kCSR: SparseFormat::kCSC;
```

It is simply to write it in one line now.

czkkkkkk (approved):
LGTM with minor comments. @frozenbugs, please take a look at this PR.

dgl_sparse/include/sparse/sparse_matrix.h
```diff
@@ -137,6 +137,57 @@ class SparseMatrix : public torch::CustomClassHolder {
...
+   * depending on the implementation.
+   */
+  c10::intrusive_ptr<SparseMatrix> RangeSelect(
```
frozenbugs, line 188: Slice has 3 parameters: start, end, step. Do we plan to support step?
https://www.w3schools.com/python/ref_func_slice.asp

@czkkkkkk

python/dgl/sparse/sparse_matrix.py
```diff
@@ -575,7 +575,11 @@ def range_select(self, dim: int, index: slice):
         if dim not in (0, 1):
             raise ValueError("The selection dimension should be 0 or 1.")
         if isinstance(index, slice):
-            raise NotImplementedError
+            start = 0 if index.start is None else index.start
+            end = index.stop
```
frozenbugs, line 579: if we don't support step, raise error here.

dgl_sparse/src/sparse_matrix.cc
```diff
@@ -122,6 +124,57 @@ c10::intrusive_ptr<SparseMatrix> SparseMatrix::FromDiag(
   return SparseMatrix::FromDiagPointer(diag, value, shape);
 }
 
+c10::intrusive_ptr<SparseMatrix> SparseMatrix::IndexSelect(
+    int64_t dim, torch::Tensor ids) {
+  auto id_array = TorchTensorToDGLArray(ids);
+  bool rowwise = dim == 0;
+  SparseFormat sparse_format;
```
frozenbugs, line 131: The only usage of sparse_format is L144, why not check rowwise in L144 directly?

dgl_sparse/src/sparse_matrix.cc
```diff
@@ -122,6 +124,57 @@ c10::intrusive_ptr<SparseMatrix> SparseMatrix::FromDiag(
...
+  if (rowwise) {
+    sparse_format = SparseFormat::kCSR;
+  } else {
+    sparse_format = SparseFormat::kCSC;
+  }
+  auto csr = rowwise ? this->CSRPtr() : this->CSCPtr();
+  auto slice_csr = dgl::aten::CSRSliceRows(CSRToOldDGLCSR(csr), id_array);
+  auto slice_value =
+      this->value().index_select(0, DGLArrayToTorchTensor(slice_csr.data));
+  // Drop the index array to prevent potential construct error of COO format.
```
frozenbugs, line 141: Can you be more specific here, where does the error come from?

dgl_sparse/include/sparse/sparse_matrix.h
czkkkkkk: We don't support it for now.

dgl_sparse/src/sparse_matrix.cc
xiangyuzhi: OK, and the reason is here: https://github.com/dmlc/dgl/pull/6208#discussion_r1314173011.

python/dgl/sparse/sparse_matrix.py
xiangyuzhi: I think it is better to raise it when users input step is not 1 or None.

dgl_sparse/src/sparse_matrix.cc
xiangyuzhi: Ok, we have changed this many times, because we want to make it more readable at the very beginning.

python/dgl/sparse/sparse_matrix.py
frozenbugs: SGTM.

python/dgl/sparse/sparse_matrix.py
```diff
@@ -575,7 +575,13 @@ def range_select(self, dim: int, index: slice):
         if dim not in (0, 1):
             raise ValueError("The selection dimension should be 0 or 1.")
         if isinstance(index, slice):
-            raise NotImplementedError
+            if index.step not in (None, 1):
+                raise NotImplementedError
```
frozenbugs, line 579: Can you also add error message?

frozenbugs (approved):

python/dgl/sparse/sparse_matrix.py
xiangyuzhi: Done

PR #6231, opened 2023-08-29
[GraphBolt] use str etype in data block

Rhett-Ying:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/data_block.py
+++ b/python/dgl/graphbolt/data_block.py
@@ -30,7 +30,7 @@ class DataBlock:
     no node types, 'node_type' should be None.
     """
 
-    edge_feature: Dict[Tuple[str, str], torch.Tensor] = None
+    edge_feature: List[Dict[Tuple[str, str], torch.Tensor]] = None
     """Edge features associated with the 'sampled_subgraphs'.
     The keys are tuples in the format '(edge_type, feature_name)', and the
     values represent the corresponding features. In the case of a homogeneous
--- a/python/dgl/graphbolt/data_block.py
+++ b/python/dgl/graphbolt/data_block.py
@@ -37,9 +37,7 @@ class DataBlock:
     graph where no edge types exist, 'edge_type' should be set to None.
     """
 
-    input_nodes: Union[
-        torch.Tensor, Dict[str, torch.Tensor]
-    ] = None
+    input_nodes: Union[torch.Tensor, Dict[str, torch.Tensor]] = None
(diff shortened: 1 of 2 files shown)


python/dgl/graphbolt/data_block.py
```diff
@@ -30,17 +30,14 @@ class DataBlock:
     no node types, 'node_type' should be None.
     """
 
-    edge_feature: List[Dict[Tuple[str, str], torch.Tensor]] = None
```
peizhou001, line 33: This `List` is corresponding to layers, not edge type.

python/dgl/graphbolt/data_block.py
```diff
@@ -30,17 +30,14 @@ class DataBlock:
...
+    edge_feature: Dict[Tuple[str, str], torch.Tensor] = None
     """Edge features associated with the 'sampled_subgraphs'.
     The keys are tuples in the format '(edge_type, feature_name)', and the
     values represent the corresponding features. In the case of a homogeneous
     graph where no edge types exist, 'edge_type' should be set to None.
-    Note 'edge_type' are of format 'str:str:str'.
     """
 
-    input_nodes: Union[
```
peizhou001, line 41: For these changes, the related code also need to be changed. Same for all below fields.

python/dgl/graphbolt/data_block.py
Rhett-Ying: then why `node_feature` is not a list?

python/dgl/graphbolt/data_block.py
Rhett-Ying: the changes I made is about annotation only. the running code lies in the use of data block such as `NegativeSampler`. I will update them in a following PR.

python/dgl/graphbolt/data_block.py
frozenbugs: Because of id space. All nodes are shared the same id space, while edges owns its own id space per layer for multi-layer subgraph.

python/dgl/graphbolt/data_block.py
frozenbugs: Why it was Tuple[str, str, str]? Wasn't it node type?

frozenbugs (approved):

python/dgl/graphbolt/data_block.py
Rhett-Ying: I think it was a typo

PR #6254, opened 2023-08-31
[GraphBolt] add names for each item in ItemSet

Rhett-Ying:
## Description

added `Names` will be used to instantiate `MiniBatch`(previously called `DataBlock`).
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/itemset.py
+++ b/python/dgl/graphbolt/itemset.py
@@ -47,11 +47,25 @@ class ItemSet:
      (tensor(4), tensor(9), tensor([18, 19]))]
     """
 
-    def __init__(self, items: Iterable or Tuple[Iterable]) -> None:
+    def __init__(
+        self,
+        items: Iterable or Tuple[Iterable],
+        names: str or Tuple[str] = None,
+    ) -> None:
         if isinstance(items, tuple):
             self._items = items
         else:
             self._items = (items,)
+        if names is not None:
+            if isinstance(names, tuple):
+                self._names = names
+            else:
+                self._names = (names,)
+            assert len(self._items) == len(
+                self._names
+            ), f"Number of items ({len(self._items)}) and names ({len(self._names)}) must match."
+        else:
+            self._names = None
 
     def __iter__(self) -> Iterator:
         if len(self._items) == 1:
@@ -67,6 +81,10 @@ class ItemSet:
         raise TypeError(
             f"{type(self).__name__} instance doesn't have valid length."
         )
+
+    @property
+    def names(self) -> Tuple[str]:
+        return self._names
 
 
 class ItemSetDict:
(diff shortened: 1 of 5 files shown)


tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
```diff
@@ -61,6 +61,53 @@ def test_OnDiskDataset_TVTSet_exceptions():
             _ = gb.OnDiskDataset(test_dir).load()
 
 
+def test_OnDiskDataset_TVTSet_ItemSet_names():
```
frozenbugs, line 64: Instead creating a new test case, the name attribute should be added for all other test cases.

Rhett-Ying: I don't it's necessarily required as the name attributes parse/use logic are fully covered in the newly added testcase. No need to test it other existing testcases which will be redundant.

python/dgl/graphbolt/itemset.py
```diff
@@ -47,11 +47,25 @@ class requires each input itemset to be iterable.
...
         else:
             self._items = (items,)
+        if names is not None:
+            if isinstance(names, tuple):
+                self._names = names
+            else:
+                self._names = (names,)
+            assert len(self._items) == len(
+                self._names
+            ), f"Number of items ({len(self._items)}) and names ({len(self._names)}) must match."
```
frozenbugs, line 66: line too long.

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
frozenbugs: Test cases are also used by reader as an example of usage, it is better to align with the recommend way.

Rhett-Ying: added.

frozenbugs (approved):

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
```diff
@@ -220,36 +326,45 @@ def test_OnDiskDataset_TVTSet_ItemSet_node_pair_label():
                 train_set:
                   - type: null
                     data:
-                      - format: numpy
+                      - name: src
```
frozenbugs, line 329: We need to make sure that user knows our keywords, currently the only mis-match between the reserved name and mini-batch variables is src, dst with node_pairs. Is this the best option?

frozenbugs (approved):

frozenbugs (approved):

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
Rhett-Ying: create a ticket for this issue: https://github.com/dmlc/dgl/issues/6259.

Issue #6257, opened 2023-08-31
[GraphBolt] Regression in SampleNeighbors


mfbalin:
## 🐛 Bug

https://github.com/dmlc/dgl/blob/master/graphbolt/src/csc_sampling_graph.cc#L406

This line goes over all the weights each time the function is called. This is not good for performance.

Runtimes before:

{('LABOR', False, False): 21.5099017159082,
 ('LABOR', False, True): 28.73401760822162,
 ('LABOR', True, False): 46.43180200597271,
 ('LABOR', True, True): 55.90287307742983}

Runtimes after:

{('LABOR', False, False): 21.77980845561251,
 ('LABOR', False, True): 54.58833070937544,
 ('LABOR', True, False): 47.00140200788155,
 ('LABOR', True, True): 81.90738406265154}

@frozenbugs @peizhou001 @RamonZhou

mfbalin:
Regression testing for all branches of code is needed. Layer=True,False Replace=True,False NonUniform=True,False.

PR #6264, opened 2023-09-01
[Graphbolt] Delete the expensive check

RamonZhou:
## Description
Deleted an time-expensive check in NeighborSampling.
Solve #6257.

Only tested the affected branches:

Before:
Labor, False, True: 51.4157
Labor, True, True: 86.6231

After:
Labor, False, True: 33.5337
Labor, True, True: 66.5594

## Checklist
- [ ] Code is well-documented

## Changes
- Deleted a check.


frozenbugs (approved):

mfbalin (approved):
If we want to perform such a check, it can happen inside of the num_pick_fn function. We already go over the weights to check how many are positive there.

mfbalin:
Also, not only the Labor branch, but also the neighbor branch is affected by this change.

RamonZhou:
> If we want to perform such a check, it can happen inside of the num_pick_fn function. We already go over the weights to check how many are positive there.

@mfbalin We decided to put it in the offline script.

PR #6287, opened 2023-09-05
[Sparse] Add sparse sample python API

xiangyuzhi:
## Description
Define sparse sample operator API.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/sparse/sparse_matrix.py
+++ b/python/dgl/sparse/sparse_matrix.py
@@ -586,6 +586,74 @@ class SparseMatrix:
             )
         raise TypeError(f"{type(index).__name__} is unsupported input type.")
 
+    def sample(
+        self, dim: int, n_pick: int, replacement: Optional[bool] = False
+    ):
+        """Returns a sample matrix according to the given sample dim and number.
+
+        Parameters
+        ----------
+        dim : int
+            The dimension for sampling, should be 0 or 1. `dim = 0` for
+            rowwise selection and `dim = 1` for columnwise selection.
+        n_pick : int
+            The number of nodes to randomly sample on each row or column.
+        replacement : bool, optional
+            Whether to allow repeated sampling of the same neighbor.
+            `replacement = True` means allowing and `replacement = False`
+            means not.
+            NOTE: If `replacement = False` and there are fewer nodes
+            than n_pick, all nodes will be sampled.
+
+        The function does not support autograd.
+
+        Returns
+        -------
+        SparseMatrix
+            A submatrix with the same shape as the original matrix
+            containing the sampled nodes.
+
+        Examples
+        --------
+
+        >>> indices = torch.tensor([0, 0, 1, 1, 2, 2, 2], [0, 2, 0, 1, 0, 1, 2]])
+        >>> val = torch.tensor([0, 1, 2, 3, 4, 5, 6])
+        >>> A = dglsp.spmatrix(indices, val)
+
+        Case 1: Sample rows with the given number and disable repeated sampling.
+
+        >>> A.sample(0, 2)
+        SparseMatrix(indices=tensor([[0, 0, 1, 1, 2, 2],
+                                     [0, 2, 0, 1, 0, 2]]),
+                     values=tensor([0, 1, 2, 3, 4, 6]),
+                     shape=(3, 3), nnz=6)
+
+        Case 2: Sample cols with the given number and disable repeated sampling.
+
+        >>> A.sample(1, 2)
+        SparseMatrix(indices=tensor([[0, 1, 1, 2, 0, 2],
+                                     [0, 0, 1, 1, 2, 2]]),
+                     values=tensor([0, 2, 3, 5, 1, 6]),
+                     shape=(3, 3), nnz=4)
+
+        Case 3: Sample rows with the given number and enable repeated sampling.
+
+        >>> A.sample(0, 2, True)
+        SparseMatrix(indices=tensor([[0, 0, 1, 1, 2, 2],
+                                     [0, 2, 0, 0, 1, 2]]),
+                     values=tensor([0, 1, 2, 2, 5, 6]),
+                     shape=(3, 3), nnz=5)
+
+        Case 4: Sample cols with the given number and enable repeated sampling.
+
+        >>> A.sample(1, 2, True)
+        SparseMatrix(indices=tensor([[0, 1, 1, 1, 2, 2],
+                                     [0, 0, 1, 1, 2, 2]]),
+                     values=tensor([0, 2, 3, 3, 6, 6]),
+                     shape=(3, 3), nnz=4)
+        """
+        raise NotImplementedError
+
 
 def spmatrix(
     indices: torch.Tensor,
--- a/python/dgl/sparse/sparse_matrix.py
+++ b/python/dgl/sparse/sparse_matrix.py
@@ -613,7 +613,7 @@ class SparseMatrix:
             When `replace = True`, repeated sampling is permitted; when
             `replace = False`, it is not allowed.
             NOTE: If `replace = False` and there are fewer elements than
-            `n_pick`, all non-zero elements will be sampled.
+            `fanout`, all non-zero elements will be sampled.
         bias : bool, optional
             A boolean flag indicating whether to enable biasing during sampling.
             When `bias = True`, the values of the sparse matrix will be used as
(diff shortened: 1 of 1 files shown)


python/dgl/sparse/sparse_matrix.py
```diff
@@ -586,6 +586,74 @@ def range_select(self, dim: int, index: slice):
             )
         raise TypeError(f"{type(index).__name__} is unsupported input type.")
 
+    def sample(
+        self, dim: int, n_pick: int, replacement: Optional[bool] = False
+    ):
+        """Returns a sample matrix according to the given sample dim and number.
```
czkkkkkk, line 592:
```suggestion
        """Returns a sampled matrix on the given dimension and sample arguments.
```

python/dgl/sparse/sparse_matrix.py
```diff
@@ -586,6 +586,74 @@ def range_select(self, dim: int, index: slice):
...
+            NOTE: If `replacement = False` and there are fewer nodes
+            than n_pick, all nodes will be sampled.
+
+        The function does not support autograd.
+
+        Returns
+        -------
+        SparseMatrix
+            A submatrix with the same shape as the original matrix
+            containing the sampled nodes.
```
czkkkkkk, line 614:
```suggestion
            containing the sampled non-zero elements.
```

czkkkkkk (changes requested):
Did you try chatgpt to polish the documentation?

xiangyuzhi:
> Did you try chatgpt to polish the documentation?

Well, I pick answers from it. Maybe my prompts are not specific.

czkkkkkk:
Yeah. Specifically ask it to polish your documentation.

python/dgl/sparse/sparse_matrix.py
```diff
@@ -586,6 +586,74 @@ def range_select(self, dim: int, index: slice):
...
+    ):
+        """Returns a sample matrix according to the given sample dim and number.
+
+        Parameters
+        ----------
+        dim : int
+            The dimension for sampling, should be 0 or 1. `dim = 0` for
+            rowwise selection and `dim = 1` for columnwise selection.
+        n_pick : int
+            The number of nodes to randomly sample on each row or column.
```
czkkkkkk, line 600:
```suggestion
            The number of elements to randomly sample on each row or column.
```

python/dgl/sparse/sparse_matrix.py
```diff
@@ -586,6 +586,74 @@ def range_select(self, dim: int, index: slice):
             )
         raise TypeError(f"{type(index).__name__} is unsupported input type.")
 
+    def sample(
+        self, dim: int, n_pick: int, replacement: Optional[bool] = False
+    ):
```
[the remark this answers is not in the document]
czkkkkkk: OK. Then skip the return type annotation.

python/dgl/sparse/sparse_matrix.py
```diff
@@ -586,6 +586,75 @@ def range_select(self, dim: int, index: slice):
             )
         raise TypeError(f"{type(index).__name__} is unsupported input type.")
 
+    def sample(
+        self, dim: int, n_pick: int, replacement: Optional[bool] = False
```
czkkkkkk, line 590: I think the current DGL implementation supports sampling on some rows/columns and sampling bias. We may need to change this interface a bit.
```suggestion
        self, dim: int, n_pick: int, ids: Optional[torch.Tensor] = None, replacement: Optional[bool] = False, bias: Optional[bool] = False
```

`ids` indicates the row/columns ids to sample and `ids = None` means sampling on the whole matrix. Then this function may change the shape of a sparse matrix. `bias` means whether use the value of the sparse matrix as sampling bias or not.

xiangyuzhi: OK, I find this too when implement the API

python/dgl/sparse/sparse_matrix.py
```diff
@@ -586,6 +586,93 @@ def range_select(self, dim: int, index: slice):
             )
         raise TypeError(f"{type(index).__name__} is unsupported input type.")
 
+    def sample(
+        self,
+        dim: int,
+        n_pick: int,
```
frozenbugs, line 592: n_pick -> fanouts

python/dgl/sparse/sparse_matrix.py
```diff
@@ -586,6 +586,93 @@ def range_select(self, dim: int, index: slice):
...
+        ids: Optional[torch.Tensor] = None,
+        replacement: Optional[bool] = False,
```
frozenbugs, line 594:
```suggestion
        replace: Optional[bool] = False,
```

python/dgl/sparse/sparse_matrix.py
```diff
@@ -586,6 +586,93 @@ def range_select(self, dim: int, index: slice):
             )
         raise TypeError(f"{type(index).__name__} is unsupported input type.")
 
+    def sample(
+        self,
+        dim: int,
+        n_pick: int,
+        ids: Optional[torch.Tensor] = None,
```
frozenbugs, line 593:
```suggestion
        nodes: Optional[torch.Tensor] = None,
```

python/dgl/sparse/sparse_matrix.py
```diff
@@ -586,6 +586,93 @@ def range_select(self, dim: int, index: slice):
...
+            sample elements.
+            NOTE: If `ids` is not provided (i.e., `ids = None`), the function
+            will sample from all rows or columns.
+        replacement : bool, optional
+            Indicates whether repeated sampling of the same element is allowed.
+            When `replacement = True`, repeated sampling is permitted; when
+            `replacement = False`, it is not allowed.
+            NOTE: If `replacement = False` and there are fewer elements than
+            `n_pick`, all non-zero elements will be sampled.
+        bias : bool, optional
```
frozenbugs, line 617: Is it comment to use the value as weight or should the weight be passed in as a paramater?

python/dgl/sparse/sparse_matrix.py
czkkkkkk: GNN sampling does not have the case that different nodes sample different numbers of neighbors. So an integer `fanout` should be ok.

python/dgl/sparse/sparse_matrix.py
czkkkkkk: I thinks `ids` is more suitable for SparseMatrix, which does not have the concepts of `graph` or `node`.

python/dgl/sparse/sparse_matrix.py
czkkkkkk: Passing weight as a parameter is also an option. But I think using the value as bias is better because
* it avoids use to think about the logic of `val_like` in `sample`. If users have a `bias` tensor, they can construct a sparse matrix by `val_like` for sampling.
* it is more clean if users want to sample a sparse matrix by its value. Otherwise, it may looks like `A.sample(..., bias=A.val)`

python/dgl/sparse/sparse_matrix.py
frozenbugs: SGTM.

python/dgl/sparse/sparse_matrix.py
frozenbugs: SGTM

czkkkkkk (approved):

python/dgl/sparse/sparse_matrix.py
frozenbugs: SGMT.

python/dgl/sparse/sparse_matrix.py
```diff
@@ -586,6 +586,93 @@ def range_select(self, dim: int, index: slice):
             )
         raise TypeError(f"{type(index).__name__} is unsupported input type.")
 
+    def sample(
```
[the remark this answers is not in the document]
frozenbugs: still please swap the order of ids and fanout.

frozenbugs (approved):

python/dgl/sparse/sparse_matrix.py
```diff
@@ -586,6 +586,93 @@ def range_select(self, dim: int, index: slice):
...
+            An optional tensor containing row or column IDs from which to
+            sample elements.
+            NOTE: If `ids` is not provided (i.e., `ids = None`), the function
+            will sample from all rows or columns.
+        replace : bool, optional
+            Indicates whether repeated sampling of the same element is allowed.
+            When `replace = True`, repeated sampling is permitted; when
+            `replace = False`, it is not allowed.
+            NOTE: If `replace = False` and there are fewer elements than
+            `n_pick`, all non-zero elements will be sampled.
```
frozenbugs, line 616: Please double check other doc string.
```suggestion
            `fanout`, all non-zero elements will be sampled.
```

PR #6288, opened 2023-09-05
[Graphbolt] Add adding reverse edges helper

peizhou001:
## Description
Add a helper to find reverse edges for a set of given edges, it is always called before finding exclude edges.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/utils/sample_utils.py
+++ b/python/dgl/graphbolt/utils/sample_utils.py
@@ -19,7 +19,7 @@ def add_reverse_edges(
     This function finds the reverse edges of the given `edges` and returns the
     composition of them. In a homogeneous graph, reverse edges have inverted
     source and destination node IDs. While in a heterogeneous graph, reversing
-    also involves swapping node IDs and their types. This function is always
+    also involves swapping node IDs and their types. This function could be
     used before `exclude_edges` function to help find targeting edges.
     Note: The found reverse edges may not really exists in the original graph.
     And repeat edges could be added becasue reverse edges may already exists in


python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -145,7 +174,20 @@ def exclude_edges(
...
+    exclude_mode : Union[str, callable], optional
+        Whether and how to exclude dependencies related to the provided edges
+        in the argument ``edges``. Possible values are
+        - If None, for excluding only provided ``edges``.
+        - If "reverse", for excluding not only the provided ``edges`` but aslo
+          their reverse edges. In the case of a homogeneous graph, reverse
+          edges refer to edges with inverted source and destination node IDs.
+          In heterogeneous graphs, this encompasses reversing both node IDs
+          and types, as specified by the ``reverse_types`` argument.
+        - If User-defined exclusion rule. It is a callable with edges in the
```
frozenbugs, line 186: It is not necessary to allow callable, `exclude_edges(subgraph, edges, callable)` is equivalent to `exclude_edges(subgraph, callable(edges))`.

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -145,7 +174,20 @@ def exclude_edges(
...
         of edge types and the corresponding edges to exclude.
-
+    exclude_mode : Union[str, callable], optional
+        Whether and how to exclude dependencies related to the provided edges
+        in the argument ``edges``. Possible values are
+        - If None, for excluding only provided ``edges``.
+        - If "reverse", for excluding not only the provided ``edges`` but aslo
+          their reverse edges. In the case of a homogeneous graph, reverse
+          edges refer to edges with inverted source and destination node IDs.
+          In heterogeneous graphs, this encompasses reversing both node IDs
```
frozenbugs, line 184: There are some corner cases we need to be crystal clear in doc string:
1) If in heterogeneous graphs, the edges (A:E:A) has the same node type for both src and dst, users still need to specify the reverse_etypes map (A:E:A) - (A:E:A), is that right?
2) if in heterogeneous graphs, for edges (A:E:B), users want to exclude all reverse edge (B:*:A) regardless edge type, users need to input all type in reverse_etypes (A:E:B) - (B:*:A), is that right?

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
```diff
@@ -179,6 +221,8 @@ def exclude_edges(
         "The sampled subgraph and the edges to exclude should be both "
         "homogeneous or both heterogeneous."
     )
+    # Get edges to exclude.
+    edges = _find_exclude_edges(edges, exclude_mode, reverse_etypes)
```
czkkkkkk, line 225: The name `_find_exclude_edges` is confusing. It is more like expanding or transforming the edge set to exclude. Moreover, I think the logic of this function is independent from `exclude_edges`. What about we call it before `exclude_edges`?

python/dgl/graphbolt/impl/sampled_subgraph_impl.py
peizhou001: Totally correct, add this in the docstring for clearity.

python/dgl/graphbolt/utils/sample_utils.py
```diff
@@ -8,6 +8,63 @@
...
+        Edges to exclude. If sampled subgraph is homogeneous, then `edges`
+        should be a pair of tensors representing the edges to exclude. If
+        sampled subgraph is heterogeneous, then `edges` should be a dictionary
+        of edge types and the corresponding edges to exclude.
+    reverse_etypes : Dict[str, str], optional
+        The mapping from the original edge types to their reverse edge types.
+
+    Returns
+    -------
+    SampledSubgraphImpl
```
frozenbugs, line 35: Update the return type?

python/dgl/graphbolt/utils/sample_utils.py
```diff
@@ -8,6 +8,63 @@
 from ..base import etype_str_to_tuple
 
 
+def add_reverse_edges(edges, reverse_etypes=None):
```
frozenbugs, line 11: add types for parameter list?

python/dgl/graphbolt/utils/sample_utils.py
```diff
@@ -8,6 +8,63 @@
...
+    used before `exclude_edges` function to help find targeting edges.
+    Note: The found reverse edges may not really exists in the original graph.
+    And repeat edges could be added becasue reverse edges may already exists in
+    the `edges`.
+
+    Parameters
+    ----------
+    edges : Union[Dict[str, Tuple[torch.Tensor, torch.Tensor]],
+                Tuple[torch.Tensor, torch.Tensor]]
+        Edges to exclude. If sampled subgraph is homogeneous, then `edges`
```
frozenbugs, line 26: Does this limit to edges to exclude?

peizhou001: yes, it does.

python/dgl/graphbolt/utils/sample_utils.py
```diff
@@ -8,6 +8,72 @@
 from ..base import etype_str_to_tuple
 
 
+def find_reverse_edges(
```
czkkkkkk, line 11: What about `add_reverse_edges`?

python/dgl/graphbolt/utils/sample_utils.py
frozenbugs: No, it does not :)

python/dgl/graphbolt/utils/sample_utils.py
frozenbugs: add_reverse_edges is slightly better than find_reverse_edges. It is pretty hard to find a name :(

peizhou001: just changed back from `add_reverse_edges`... If both you think it's better, let me change back

python/dgl/graphbolt/utils/sample_utils.py
```diff
@@ -8,6 +8,72 @@
...
+        Tuple[torch.Tensor, torch.Tensor],
+    ],
+    reverse_etypes: Dict[str, str] = None,
+):
+    r"""
+    This function finds the reverse edges of the given `edges` and returns the
+    composition of them. In a homogeneous graph, reverse edges have inverted
+    source and destination node IDs. While in a heterogeneous graph, reversing
+    also involves swapping node IDs and their types. This function is always
+    used before `exclude_edges` function to help find targeting edges.
```
frozenbugs, line 23:
```suggestion
    also involves swapping node IDs and their types. This function could be used
    before `exclude_edges` function to help find targeting edges.
```

frozenbugs (approved):

PR #6300, opened 2023-09-07
[CMAKE] Move DGL cuda file declaration to the main CMakeLists.txt

czkkkkkk:
## Description

Move DGL cuda file declaration to the main CMakeLists.txt. Later both graphbolt and dgl sparse can use `dgl_config_cuda` to configure CUDA compilation

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -259,8 +259,22 @@ list(APPEND DGL_SRC ${DGL_RPC_SRC})
 
 # Configure cuda
 if(USE_CUDA)
-  dgl_config_cuda(DGL_CUDA_SRC)
+  file(GLOB_RECURSE DGL_CUDA_SRC
+    src/array/cuda/*.cc
+    src/array/cuda/*.cu
+    src/array/cuda/uvm/*.cc
+    src/array/cuda/uvm/*.cu
+    src/kernel/cuda/*.cc
+    src/kernel/cuda/*.cu
+    src/partition/cuda/*.cu
+    src/runtime/cuda/*.cc
+    src/runtime/cuda/*.cu
+    src/geometry/cuda/*.cu
+    src/graph/transform/cuda/*.cu
+    src/graph/sampling/randomwalks/*.cu
+  )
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -259,8 +259,22 @@ list(APPEND DGL_SRC ${DGL_RPC_SRC})
 
 # Configure cuda
 if(USE_CUDA)
-  dgl_config_cuda(DGL_CUDA_SRC)
+  file(GLOB_RECURSE DGL_CUDA_SRC
+    src/array/cuda/*.cc
```
frozenbugs, line 263: is it a better idea to move all cuda file to a single dir?

czkkkkkk: I think they are similar to the cc implementation files. So we also need to organize them in different folders.

yaox12:
Regarding this change, I wonder if it's a chance to switch to CMake 3.8+, list CUDA among the languages supported, and then add CUDA (.cu) sources directly to targets similar to other languages, since the current way to enable CUDA, i.e., `FindCUDA`, is deprecated.  
Reference: https://cmake.org/cmake/help/latest/module/FindCUDA.html

czkkkkkk:
> Regarding this change, I wonder if it's a chance to switch to CMake 3.8+, list CUDA among the languages supported, and then add CUDA (.cu) sources directly to targets similar to other languages, since the current way to enable CUDA, i.e., `FindCUDA`, is deprecated. Reference: https://cmake.org/cmake/help/latest/module/FindCUDA.html

Right. We should deprecate FindCUDA since it is deprecated from CMAKE 3.10. But I am not familiar with the new cmake syntax. @yaox12 Could you help to make such a change after this PR?

frozenbugs (approved):

CMakeLists.txt
nv-dlasalle: +1 to this. They are dispatched from the co-located files, so I think the current method of organizing them makes a lot of sense.

nv-dlasalle (approved):

PR #6303, opened 2023-09-08
[Sparse] Sparse sample implementation

xiangyuzhi:
## Description
The implementation of sparse 'sample' operator. The sparse 'sample' operator differs slightly from DGL's 'sample' API. In the sparse sample, it only returns the submatrix selected by the index, rather than the entire original matrix.

## Checklist
- [ ] Code is well-documented

## Changes

### DIFF
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -196,11 +196,13 @@ class SparseMatrix : public torch::CustomClassHolder {
    * @param ids An optional tensor containing row or column IDs from which to
    * sample elements.
    * @param replace Indicates whether repeated sampling of the same element
-   * is allowed. When 'replace = true', repeated sampling is permitted; when
-   * 'replace = false', it is not allowed.
+   * is allowed. If True, repeated sampling is allowed; otherwise, it is not
+   * allowed.
    * @param bias An optional boolean flag indicating whether to enable biasing
-   * during sampling. When 'bias = true', the values of the sparse matrix will
-   * be used as bias weights.
+   * during sampling. If True, the values of the sparse matrix will be used as
+   * bias weights, meaning that elements with higher values will be more likely
+   * to be sampled. Otherwise, all elements will be sampled uniformly,
+   * regardless of their value.
    *
    * @return A new SparseMatrix with the same shape as the original matrix
    * containing the sampled elements.
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -191,27 +191,26 @@ class SparseMatrix : public torch::CustomClassHolder {
    * with the specified sample count.
    *
    * @param dim Select rows (dim=0) or columns (dim=1) for sampling.
-   * @param n_pick The number of elements to randomly sample from each row or
+   * @param fanout The number of elements to randomly sample from each row or
    * column.
-   * @param replacement Indicates whether repeated sampling of the same element
-   * is allowed. When 'replacement = true', repeated sampling is permitted; when
-   * 'replacement = false', it is not allowed.
    * @param ids An optional tensor containing row or column IDs from which to
    * sample elements.
+   * @param replace Indicates whether repeated sampling of the same element
+   * is allowed. When 'replace = true', repeated sampling is permitted; when
+   * 'replace = false', it is not allowed.
    * @param bias An optional boolean flag indicating whether to enable biasing
    * during sampling. When 'bias = true', the values of the sparse matrix will
    * be used as bias weights.
    *
    * @return A new SparseMatrix with the same shape as the original matrix
    * containing the sampled elements.
-   * @note If 'replacement = false' and there are fewer elements than 'n_pick',
+   * @note If 'replace = false' and there are fewer elements than 'fanout',
    * all non-zero elements will be sampled.
    * @note If 'ids' is not provided, the function will sample from
    * all rows or columns.
    */
   c10::intrusive_ptr<SparseMatrix> Sample(
-      int64_t dim, int64_t n_pick, bool replacement, torch::Tensor ids,
-      bool bias);
+      int64_t dim, int64_t fanout, torch::Tensor ids, bool replace, bool bias);
 
   /**
    * @brief Create a SparseMatrix from a SparseMatrix using new values.
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -183,6 +183,29 @@ class SparseMatrix : public torch::CustomClassHolder {
       int64_t dim, int64_t start, int64_t end);
 
   /**
+   * @brief Create a SparseMatrix by sampling elements based on provided
+   * dimension and number.
+   *
+   * This function allows you to create a new SparseMatrix by randomly sampling
+   * elements along a specified dimension (row-wise or column-wise) with the
+   * given number.
+   *
+   * @param dim Select rows (dim=0) or columns (dim=1) for sampling.
+   * @param n_pick The number of elements to randomly sample from each row or
+   * column.
+   * @param replacement Indicates whether repeated sampling of the same element
+   * is allowed. When 'replacement = true', repeated sampling is permitted; when
+   * 'replacement = false', it is not allowed.
+   *
+   * @return A new SparseMatrix with the same shape as the original matrix
+   * containing the sampled elements.
+   * @note If 'replacement = false' and there are fewer elements than 'n_pick',
+   * all non-zero elements will be sampled.
+   */
+  c10::intrusive_ptr<SparseMatrix> Sample(
+      int64_t dim, int64_t n_pick, bool replacement = false);
+
+  /**
    * @brief Create a SparseMatrix from a SparseMatrix using new values.
    * @param mat An existing sparse matrix
    * @param value New values of the sparse matrix
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -186,9 +186,9 @@ class SparseMatrix : public torch::CustomClassHolder {
    * @brief Create a SparseMatrix by sampling elements based on the specified
    * dimension and sample count.
    *
-   * This function enables the creation of a new SparseMatrix by randomly
-   * sampling elements along a specified dimension (row-wise or column-wise)
-   * with the specified sample count.
+   * If `ids` is provided, this function samples elements from the specified
+   * set of row or column IDs, resulting in a sparse matrix containing only
+   * the sampled rows or columns.
    *
    * @param dim Select rows (dim=0) or columns (dim=1) for sampling.
    * @param fanout The number of elements to randomly sample from each row or
@@ -204,6 +204,7 @@ class SparseMatrix : public torch::CustomClassHolder {
    *
    * @return A new SparseMatrix with the same shape as the original matrix
    * containing the sampled elements.
+   *
    * @note If 'replace = false' and there are fewer elements than 'fanout',
    * all non-zero elements will be sampled.
    * @note If 'ids' is not provided, the function will sample from
--- a/dgl_sparse/include/sparse/sparse_matrix.h
+++ b/dgl_sparse/include/sparse/sparse_matrix.h
@@ -183,12 +183,12 @@ class SparseMatrix : public torch::CustomClassHolder {
       int64_t dim, int64_t start, int64_t end);
 
   /**
-   * @brief Create a SparseMatrix by sampling elements based on provided
-   * dimension and number.
+   * @brief Create a SparseMatrix by sampling elements based on the specified
+   * dimension and sample count.
    *
-   * This function allows you to create a new SparseMatrix by randomly sampling
-   * elements along a specified dimension (row-wise or column-wise) with the
-   * given number.
+   * This function enables the creation of a new SparseMatrix by randomly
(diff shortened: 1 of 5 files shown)


dgl_sparse/src/sparse_matrix.cc
```diff
@@ -167,6 +167,35 @@ c10::intrusive_ptr<SparseMatrix> SparseMatrix::RangeSelect(
   }
 }
 
+c10::intrusive_ptr<SparseMatrix> SparseMatrix::Sample(
+    int64_t dim, int64_t fanout, torch::Tensor ids, bool replace, bool bias) {
+  bool rowwise = dim == 0;
+  auto id_array = TorchTensorToDGLArray(ids);
+  auto csr = rowwise ? this->CSRPtr() : this->CSCPtr();
+
+  auto prob_or_mask =
```
czkkkkkk, line 176:
```suggestion
  auto prob =
```

python/dgl/sparse/sparse_matrix.py
```diff
@@ -671,7 +671,12 @@ def sample(
                      values=tensor([0, 2, 3, 3]),
                      shape=(3, 2), nnz=3)
         """
-        raise NotImplementedError
+        if ids is None:
+            num = self.shape[0] if dim == 0 else self.shape[1]
+            ids = torch.tensor(range(0, num))
```
czkkkkkk, line 676:
```suggestion
            dim_size = self.shape[0] if dim == 0 else self.shape[1]
            ids = torch.range(0, dim_size, dtype=torch.int64, device=self.device)
```

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -504,6 +504,60 @@ def test_range_select(create_func, shape, dense_dim, select_dim, rang):
     assert torch.allclose(A_select_to_dense, dense_select)
 
 
+@pytest.mark.parametrize(
+    "create_func", [rand_diag, rand_csr, rand_csc, rand_coo]
+)
+@pytest.mark.parametrize("sample_dim", [0, 1])
+@pytest.mark.parametrize("index", [None, (0, 1, 3)])
+@pytest.mark.parametrize("replace", [0, 1])
```
czkkkkkk, line 512:
```suggestion
@pytest.mark.parametrize("replace", [False, True])
```

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -504,6 +504,60 @@ def test_range_select(create_func, shape, dense_dim, select_dim, rang):
...
+@pytest.mark.parametrize("bias", [0, 1])
```
czkkkkkk, line 513:
```suggestion
@pytest.mark.parametrize("bias", [False, True])
```

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -504,6 +504,60 @@ def test_range_select(create_func, shape, dense_dim, select_dim, rang):
     assert torch.allclose(A_select_to_dense, dense_select)
 
 
+@pytest.mark.parametrize(
+    "create_func", [rand_diag, rand_csr, rand_csc, rand_coo]
+)
+@pytest.mark.parametrize("sample_dim", [0, 1])
+@pytest.mark.parametrize("index", [None, (0, 1, 3)])
```
czkkkkkk, line 511: Also test repeated indices.

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -504,6 +504,60 @@ def test_range_select(create_func, shape, dense_dim, select_dim, rang):
...
+@pytest.mark.parametrize("bias", [0, 1])
+def test_sample(create_func, sample_dim, index, replace, bias):
+    ctx = F.ctx()
+    shape = (5, 5)
+    sample_num = 3
+    A = create_func(shape, 6, ctx)
+    A_row = A.row
+    A_col = A.col
+    A_val = torch.abs(A.val)
+    A = from_coo(A_row, A_col, A_val, shape)
```
czkkkkkk, line 522: Why A is reconstructed?

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -504,6 +504,60 @@ def test_range_select(create_func, shape, dense_dim, select_dim, rang):
...
+
+    if index is not None:
+        index = torch.tensor(index).to(ctx)
+
+    A_sample = A.sample(sample_dim, sample_num, index, replace, bias)
+    A_dense = sparse_matrix_to_dense(A)
+    A_sample_to_dense = sparse_matrix_to_dense(A_sample)
+
+    if index == None:
```
czkkkkkk, line 531: This branch is not necessary. We can create the index array for testing.

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -504,6 +504,60 @@ def test_range_select(create_func, shape, dense_dim, select_dim, rang):
...
+        ans_shape = shape
+        for row in range(shape[0]):
+            ans_ele = list(A_dense[row, :].nonzero().reshape(-1))
+            ret_ele = list(A_sample_to_dense[row, :].nonzero().reshape(-1))
+            for e in ret_ele:
```
czkkkkkk, line 536: Could you also test whether the number of elements of each row or column is expected? For example, if a row to sample has `n` nonzero elements and we want to sample `m` among them without replacement, then the sampled size should be `min(n, m)`. You may also need to take care about the case when sampling a row without nonzero elements in a replacement manner.

dgl_sparse/src/sparse_matrix.cc
```diff
@@ -167,6 +167,35 @@ c10::intrusive_ptr<SparseMatrix> SparseMatrix::RangeSelect(
...
+      bias ? TorchTensorToDGLArray(this->value()) : dgl::aten::NullArray();
+  auto sample_coo = dgl::aten::CSRRowWiseSampling(
+      CSRToOldDGLCSR(csr), id_array, fanout, prob_or_mask, replace);
+  auto sample_csr = COOToCSR(sample_coo);
```
czkkkkkk, line 180: Is it better to do the slicing and then sampling? Then you don't need to convert sampled COO to CSR.

czkkkkkk (changes requested):

tests/python/pytorch/sparse/test_sparse_matrix.py
xiangyuzhi: Because the values of matrix generated by function may contain negative elements which can not be used as bias array, I use 'abs' function to make them non-negative.

dgl_sparse/src/sparse_matrix.cc
xiangyuzhi: I have a question regarding sampling. If we sample duplicate IDs, should the results for the same row always be the same? For instance, if we sample 2 elements from a row ```[1, 2, 3]``` three times, can the result be ```[1, 2], [2, 3], [1, 3]```? Alternatively, should it always be like ```[1, 2], [1, 2], [1, 2]```?

If we use slicing after sampling, the sampled rows from one row will always be the same. However, if we use sampling after slicing, the answers can differ.

dgl_sparse/include/sparse/sparse_matrix.h
```diff
@@ -182,6 +182,36 @@ class SparseMatrix : public torch::CustomClassHolder {
   c10::intrusive_ptr<SparseMatrix> RangeSelect(
       int64_t dim, int64_t start, int64_t end);
 
+  /**
+   * @brief Create a SparseMatrix by sampling elements based on the specified
+   * dimension and sample count.
+   *
+   * This function enables the creation of a new SparseMatrix by randomly
```
czkkkkkk, line 189: This sentence delivers the same message as the previous one. Let's remove it.

dgl_sparse/include/sparse/sparse_matrix.h
```diff
@@ -182,6 +182,36 @@ class SparseMatrix : public torch::CustomClassHolder {
   c10::intrusive_ptr<SparseMatrix> RangeSelect(
       int64_t dim, int64_t start, int64_t end);
 
+  /**
+   * @brief Create a SparseMatrix by sampling elements based on the specified
+   * dimension and sample count.
+   *
```
czkkkkkk, line 188:
```suggestion
   *
   * If `ids` is provided, this function samples elements from the specified set of row or column IDs, resulting in a sparse matrix containing only the sampled rows or columns.
```

dgl_sparse/include/sparse/sparse_matrix.h
```diff
@@ -182,6 +182,36 @@ class SparseMatrix : public torch::CustomClassHolder {
...
+   * sample elements.
+   * @param replace Indicates whether repeated sampling of the same element
+   * is allowed. When 'replace = true', repeated sampling is permitted; when
+   * 'replace = false', it is not allowed.
+   * @param bias An optional boolean flag indicating whether to enable biasing
+   * during sampling. When 'bias = true', the values of the sparse matrix will
+   * be used as bias weights.
+   *
+   * @return A new SparseMatrix with the same shape as the original matrix
+   * containing the sampled elements.
```
czkkkkkk, line 206:
```suggestion
   * containing the sampled elements.
   *
```

dgl_sparse/src/sparse_matrix.cc
```diff
@@ -167,6 +167,35 @@ c10::intrusive_ptr<SparseMatrix> SparseMatrix::RangeSelect(
...
+  // Slicing matrix.
+  auto slice_csr = dgl::aten::CSRSliceRows(CSRToOldDGLCSR(csr), id_array);
+  auto slice_value =
+      this->value().index_select(0, DGLArrayToTorchTensor(slice_csr.data));
+  // Reset value indices.
+  slice_csr.data = dgl::aten::NullArray();
+
+  auto prob =
+      bias ? TorchTensorToDGLArray(slice_value) : dgl::aten::NullArray();
+  std::vector<uint64_t> v(id_array.NumElements());
```
czkkkkkk, line 184: Please follow the `Range` interface of NDArray to create a range vector.

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -504,6 +504,67 @@ def test_range_select(create_func, shape, dense_dim, select_dim, rang):
...
+@pytest.mark.parametrize("index", [(0, 1, 2, 3, 4), (0, 1, 3), (1, 1, 2)])
+@pytest.mark.parametrize("replace", [False, True])
+@pytest.mark.parametrize("bias", [False, True])
+def test_sample(create_func, sample_dim, index, replace, bias):
+    ctx = F.ctx()
+    shape = (5, 5)
+    sample_num = 3
+    A = create_func(shape, 10, ctx)
+    A_row, A_col, A_val = A.row, A.col, torch.abs(A.val)
+    A = from_coo(A_row, A_col, A_val, shape)
```
czkkkkkk, line 520:
```suggestion
    A = val_like(torch.abs(A.val))
```

dgl_sparse/include/sparse/sparse_matrix.h
```diff
@@ -182,6 +182,37 @@ class SparseMatrix : public torch::CustomClassHolder {
...
+   * the sampled rows or columns.
+   *
+   * @param dim Select rows (dim=0) or columns (dim=1) for sampling.
+   * @param fanout The number of elements to randomly sample from each row or
+   * column.
+   * @param ids An optional tensor containing row or column IDs from which to
+   * sample elements.
+   * @param replace Indicates whether repeated sampling of the same element
+   * is allowed. When 'replace = true', repeated sampling is permitted; when
+   * 'replace = false', it is not allowed.
```
frozenbugs, line 200:
```suggestion
   * is allowed. If True, repeated sampling is allowed; otherwise, it is not allowed.
```

frozenbugs, line 200: If True, repeated sampling is allowed; otherwise, it is not allowed.

dgl_sparse/include/sparse/sparse_matrix.h
```diff
@@ -182,6 +182,37 @@ class SparseMatrix : public torch::CustomClassHolder {
...
+   * @param bias An optional boolean flag indicating whether to enable biasing
+   * during sampling. When 'bias = true', the values of the sparse matrix will
+   * be used as bias weights.
```
frozenbugs, line 203: If True, the values of the sparse matrix will be used as bias weights, meaning that elements with higher values will be more likely to be sampled. Otherwise, all elements will be sampled uniformly, regardless of their value.

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -504,6 +504,66 @@ def test_range_select(create_func, shape, dense_dim, select_dim, rang):
...
+    A = create_func(shape, 10, ctx)
+    A = val_like(A, torch.abs(A.val))
+
+    index = torch.tensor(index).to(ctx)
+
+    A_sample = A.sample(sample_dim, sample_num, index, replace, bias)
+    A_dense = sparse_matrix_to_dense(A)
+    A_sample_to_dense = sparse_matrix_to_dense(A_sample)
+
+    if sample_dim == 0:
```
frozenbugs, line 527: nit: The nested code makes it hard to read.
One option is separate the dim 1 and 0 to 2 different test cases since the primary code is not shared anyway.

def test_sample_rowwise(create_func, index, replace, bias):

def test_sample_columnwise(create_func, index, replace, bias):

frozenbugs (approved):

czkkkkkk (approved):

PR #6329, opened 2023-09-14
[GraphBolt] enable to download built-in dataset from s3

Rhett-Ying:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/impl/ondisk_dataset.py
+++ b/python/dgl/graphbolt/impl/ondisk_dataset.py
@@ -30,7 +30,7 @@ from .ondisk_metadata import (
 )
 from .torch_based_feature_store import TorchBasedFeatureStore
 
-__all__ = ["OnDiskDataset", "preprocess_ondisk_dataset", "BuiltinOnDiskDataset"]
+__all__ = ["OnDiskDataset", "preprocess_ondisk_dataset", "BuiltinDataset"]
 
 
 def _copy_or_convert_data(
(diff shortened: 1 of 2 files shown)


python/dgl/graphbolt/impl/ondisk_dataset.py
```diff
@@ -473,3 +474,31 @@ def _init_tvt_set(
                 )
             ret = ItemSetDict(data)
         return ret
+
+
+class BuiltinOnDiskDataset(OnDiskDataset):
```
frozenbugs, line 479: BuiltinDataset should be good, I can't image a BuiltinDataset which is not OnDiskDataset.
```suggestion
class BuiltinDataset(OnDiskDataset):
```

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
```diff
@@ -1694,3 +1694,26 @@ def test_OnDiskDataset_load_tasks():
         original_train_set = None
         modify_train_set = None
         dataset = None
+
+
+def test_BuiltinOnDiskDataset():
+    """Test BuiltinOnDiskDataset."""
+    with tempfile.TemporaryDirectory() as test_dir:
```
frozenbugs, line 1701: better to move to regression.

frozenbugs (approved):

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
Rhett-Ying: this `test-only` dataset is dedicated for unit test that makes sure the download and instantiate logic works as expected. The dataset is quite small and expect to finish very soon.
As for regression, I think what we need to offer is make sure all the built-in dataset(such as ogb-ones) can be downloaded and instantiated successfully.
In a word, both are require in CI and regression.

frozenbugs (approved):

PR #6341, opened 2023-09-17
[GraphBolt] Add DistributedItemSampler to support multi-gpu training

RamonZhou:
## Description
Create `DistributedItemSampler` as a replacement choice of `ItemSampler`. It will distribute inputs to each replica, supporting DDP training.
Added new params:
- `even_inputs`: It true, `DistributedItemSampler` will drop the uneven inputs so the number of batches is all the same for all replicas. Could be used in lightning. If False, user needs to use `Join` to avoid possible hanging.
- `num_replicas`: As the other torch APIs, user can choose to pass the number of replicas. If None, `DistributedItemSampler` will get this value from `dist`.

## Checklist
- [ ] Code is well-documented

## Changes
- `DistributedItemSampler`
- Unit test.


python/dgl/graphbolt/item_sampler.py
```diff
@@ -326,8 +329,104 @@ def __iter__(self) -> Iterator:
...
+    """Distributed Item Sampler.
+
+    This is useful with torch's DDP training. It will create distributed item
+    subset of data which could be node IDs, node pairs with or without labels,
+    node pairs with negative sources/destinations, DGLGraphs and heterogeneous
+    counterparts. The items of the original item set will be sharded and form
+    an exclusive subset for each replica (process).
+
+    Note: This class `DistributedItemSampler` is not decorated with
+    `torchdata.datapipes.functional_datapipe` on purpose. This indicates it
```
frozenbugs, line 348: I can still see functional_datapipe decoration, conflict?

python/dgl/graphbolt/item_sampler.py
```diff
@@ -326,8 +329,104 @@ def __iter__(self) -> Iterator:
...
+
+@functional_datapipe("distributed_sample_item")
+class DistributedItemSampler(ItemSampler):
+    """Distributed Item Sampler.
+
+    This is useful with torch's DDP training. It will create distributed item
+    subset of data which could be node IDs, node pairs with or without labels,
+    node pairs with negative sources/destinations, DGLGraphs and heterogeneous
+    counterparts. The items of the original item set will be sharded and form
+    an exclusive subset for each replica (process).
```
frozenbugs, line 345: Powered by Bard:

This sampler creates a distributed subset of items from the given data set, which can be used for training with PyTorch's Distributed Data Parallel (DDP). The items can be node IDs, node pairs with or without labels, node pairs with negative sources/destinations, DGLGraphs, or heterogeneous counterparts. The original item set is sharded such that each replica (process) receives an exclusive subset.

python/dgl/graphbolt/item_sampler.py
```diff
@@ -326,8 +329,104 @@ def __iter__(self) -> Iterator:
...
+        self._even_inputs = even_inputs
+        # Apply a sharding filter to distribute the items.
+        self._item_set = self._item_set.sharding_filter()
+        # Get world size.
+        if num_replicas is None:
+            if not dist.is_available():
+                raise RuntimeError(
+                    "Requires distributed package to be available"
+                )
+            num_replicas = dist.get_world_size()
```
frozenbugs, line 400: What if input num_replicas is conflict with dist.get_world_size()?

python/dgl/graphbolt/item_sampler.py
```diff
@@ -326,8 +329,108 @@ def __iter__(self) -> Iterator:
...
+    minibatcher : Optional[Callable]
+        A callable that takes in a list of items and returns a `MiniBatch`.
+    drop_last : bool
+        Option to drop the last batch if it's not full.
+    shuffle : bool
+        Option to shuffle before sample.
+    num_replicas: int
+        Number of processes participating in distributed training. By default,
+        `world_size` is retrieved from the current distributed group.
+    even_inputs : bool
```
frozenbugs, line 367: even_inputs is not very descriptive, what about drop_uneven_batches?
```suggestion
    drop_uneven_batches : bool
```

python/dgl/graphbolt/item_sampler.py
```diff
@@ -326,8 +329,108 @@ def __iter__(self) -> Iterator:
...
+        Option to make sure the numbers of batches for each replica are the
```
frozenbugs, line 368: Option to ensure that all replicas receive the same number of batches. If some replicas have more batches than the others, the redundant batches of those replicas will be dropped. If the drop_last parameter is also set to True, the last batch will be dropped before the redundant batches are dropped.
Note: When using Distributed Data Parallel (DDP) training, it is recommended to use the Join Context Manager provided by PyTorch to solve the uneven inputs problem. However, this option can be used if the Join Context Manager is not helpful for any reason.

python/dgl/graphbolt/item_sampler.py
```diff
@@ -326,8 +329,108 @@ def __iter__(self) -> Iterator:
...
+        same. If some of the processes have more batches than the others, the
+        redundant batches of those processes will be dropped. If drop_last and
+        even_inputs are both true, it will execute drop_last before even_inputs.
+        When using DDP training, using Join Context Manager provided by pytorch
+        is the recommended way to solve the uneven inputs problem. But if Join
+        is not helpful due to any reason, you could try this option.
+
+    Examples
```
frozenbugs, line 376: Remove if no example is provided.

python/dgl/graphbolt/item_sampler.py
```diff
@@ -326,8 +329,108 @@ def __iter__(self) -> Iterator:
...
+        Data to be sampled.
+    batch_size : int
+        The size of each batch.
+    minibatcher : Optional[Callable]
+        A callable that takes in a list of items and returns a `MiniBatch`.
+    drop_last : bool
+        Option to drop the last batch if it's not full.
+    shuffle : bool
+        Option to shuffle before sample.
+    num_replicas: int
```
frozenbugs, line 364: Is replica a good terminology to describe the processes?
num_processes

if you prefer num_replicas, then update the python doc:
`Number of model replicas created in Distributed Data Parallel (DDP) training.`

Or just world_size as used in the python doc.

python/dgl/graphbolt/item_sampler.py
```diff
@@ -326,8 +329,108 @@ def __iter__(self) -> Iterator:
...
+        Number of processes participating in distributed training. By default,
+        `world_size` is retrieved from the current distributed group.
```
frozenbugs, line 366: The number of model replicas that will be created during Distributed Data Parallel (DDP) training. By default, the world size is retrieved from the current distributed group.

python/dgl/graphbolt/item_sampler.py
```diff
@@ -326,8 +329,108 @@ def __iter__(self) -> Iterator:
...
+    even_inputs : bool
+        Option to make sure the numbers of batches for each replica are the
+        same. If some of the processes have more batches than the others, the
+        redundant batches of those processes will be dropped. If drop_last and
+        even_inputs are both true, it will execute drop_last before even_inputs.
+        When using DDP training, using Join Context Manager provided by pytorch
+        is the recommended way to solve the uneven inputs problem. But if Join
+        is not helpful due to any reason, you could try this option.
```
frozenbugs, line 374: What may happen if I did not use Join Context Manager to even the input, and also did not set this option to True?

frozenbugs: What may happen if I did not set drop_last and but set this option to True, in this case the last batches still will not have complete size.

python/dgl/graphbolt/item_sampler.py
```diff
@@ -326,8 +329,108 @@ def __iter__(self) -> Iterator:
...
+        # Get world size.
+        if num_replicas is None:
+            if not dist.is_available():
+                raise RuntimeError(
+                    "Requires distributed package to be available"
+                )
+            num_replicas = dist.get_world_size()
+        total_len = len(item_set)
+        # Calculate the number of batches for each replica.
+        self._num_batches = total_len // (num_replicas * batch_size) + (
```
frozenbugs, line 403: The _num_batches is used as the min_num_batches across each process, is this formula correct in calculate the min_num_batches?

tests/python/pytorch/graphbolt/test_item_sampler.py
```diff
@@ -610,3 +614,110 @@ def test_ItemSetDict_node_pairs_negative_dsts(batch_size, shuffle, drop_last):
...
+    num_batches = len(list(item_sampler))
+
+    # Calculate expected numbers of items and batches.
+    expected_num_items = num_ids // nprocs + (num_ids % nprocs > proc_id)
+    if drop_last and expected_num_items % batch_size > 0:
+        expected_num_items -= expected_num_items % batch_size
+    expected_num_batches = expected_num_items // batch_size + (
+        (not drop_last) and (expected_num_items % batch_size > 0)
+    )
+    if even_inputs:
```
frozenbugs, line 664: Consider to test whether this option can actually even the input in different cases.

python/dgl/graphbolt/item_sampler.py
RamonZhou: This is the same as ItemSampler (L90). I think it means it's not added **on purpose**.

frozenbugs: @Rhett-Ying can you clarify the conflict?

Rhett-Ying: I have removed the functional decorator in #6357. please remove it too

python/dgl/graphbolt/item_sampler.py
```diff
@@ -326,8 +329,114 @@ def __iter__(self) -> Iterator:
...
+        self._drop_uneven_inputs = drop_uneven_inputs
+        # Apply a sharding filter to distribute the items.
+        self._item_set = self._item_set.sharding_filter()
+        # Get world size.
+        if num_replicas is None:
+            assert (
+                dist.is_available()
+            ), "Requires distributed package to be available"
+            num_replicas = dist.get_world_size()
+        total_len = len(item_set)
```
Rhett-Ying, line 407: `__len__` is not always available though it's true for most cases. how to handle this scenario?

python/dgl/graphbolt/item_sampler.py
```diff
@@ -326,8 +329,114 @@ def __iter__(self) -> Iterator:
...
+        # Calculate the number of batches for each replica.
+        self._num_batches = total_len // (num_replicas * batch_size) + (
+            (not drop_last)
+            and (total_len % (num_replicas * batch_size) >= num_replicas)
         )
 
+    def __iter__(self) -> Iterator:
```
Rhett-Ying, line 414: below code is quite like to the implementation of `ItemSampler`. is it possible to remove such code duplication? In current way, I don't see much benefit of subclassing `ItemSampler`.

tests/python/pytorch/graphbolt/test_item_sampler.py
```diff
@@ -610,3 +614,121 @@ def test_ItemSetDict_node_pairs_negative_dsts(batch_size, shuffle, drop_last):
...
+
+    # Create a DistributedItemSampler.
+    item_sampler = gb.DistributedItemSampler(
+        item_set,
+        batch_size=batch_size,
+        shuffle=shuffle,
+        drop_last=drop_last,
+        drop_uneven_inputs=drop_uneven_inputs,
+    )
+    data_loader = gb.SingleProcessDataLoader(item_sampler)
```
Rhett-Ying, line 650: what about `MultiProcessDL`? cannot be used?

python/dgl/graphbolt/item_sampler.py
```diff
@@ -326,8 +329,114 @@ def __iter__(self) -> Iterator:
...
+        minibatcher: Optional[Callable] = minibatcher_default,
+        drop_last: Optional[bool] = False,
+        shuffle: Optional[bool] = False,
+        num_replicas: Optional[int] = None,
+        drop_uneven_inputs: Optional[bool] = False,
+    ) -> None:
+        super().__init__(item_set, batch_size, minibatcher, drop_last, shuffle)
+        self._drop_uneven_inputs = drop_uneven_inputs
+        # Apply a sharding filter to distribute the items.
+        self._item_set = self._item_set.sharding_filter()
```
Rhett-Ying, line 400: so the worflow in `DistribuedItemSampler` is `shard-> shuffle -> batch -> ...` ?

python/dgl/graphbolt/item_sampler.py
RamonZhou: It will use the given `num_replicas`.

frozenbugs: Let's me make the question clear, in which case num_replicas does not equal to get_world_size.
Intuitively we don't need to pass in num_replicas since we can get_world_size. The comment should focus on address this confusion.

RamonZhou: It could be used when `dist` is not available during the instantiation of `DistributedItemSampler`. Most of the cases we don't have to use this param. But the comment is clear I think: we only retrieve it from `dist.get_world_size()` when `num_replicas` is None.

python/dgl/graphbolt/item_sampler.py
RamonZhou: Yes

tests/python/pytorch/graphbolt/test_item_sampler.py
RamonZhou: Could be used but not perfect.

python/dgl/graphbolt/item_sampler.py
RamonZhou: The `drop_last` cares only about the completeness of the last batch, while `drop_uneven_inputs` cares only about the number of batches. I've added example cases of these situations.

python/dgl/graphbolt/item_sampler.py
RamonZhou: It's only used when `drop_uneven_inputs` is True and it's used for limiting the number of batches in the datapipe. It would be fine if tests are all passed.

python/dgl/graphbolt/item_sampler.py
```diff
@@ -306,27 +325,147 @@ def _collate(batch):
...
-        )
+        data_pipe = data_pipe.map(partial(self._minibatcher, names=self._names))
 
         return iter(data_pipe)
+
+
+class DistributedItemSampler(ItemSampler):
+    """Distributed Item Sampler.
+
+    This sampler creates a distributed subset of items from the given data set,
```
Rhett-Ying, line 343: I think we'd better clearly introduce the workflow: `shard -> shuffle -> ...` to help user have a clear understanding of this datapipe. And For this dist one, add note for the shuffle part which may lower the data diversity potentially.

python/dgl/graphbolt/item_sampler.py
frozenbugs: This is still a valid question, maybe we can not support multi-gpu training if we can't get the len of item_set.

@Rhett-Ying do you have other ideas?

frozenbugs (approved):

PR #6347, opened 2023-09-19
[Graphbolt] Support return reverse edge ids in sampling

peizhou001:
## Description
Support return reverse edge ids in sampling, which always used when need edge features.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/impl/csc_sampling_graph.py
+++ b/python/dgl/graphbolt/impl/csc_sampling_graph.py
@@ -8,7 +8,7 @@ from typing import Dict, Optional, Union
 
 import torch
 
-from ...base import ETYPE
+from ...base import EID, ETYPE
 from ...convert import to_homogeneous
 from ...heterograph import DGLGraph
 from ..base import etype_str_to_tuple, etype_tuple_to_str
@@ -230,6 +230,10 @@ class CSCSamplingGraph:
         )
         row = C_sampled_subgraph.indices
         type_per_edge = C_sampled_subgraph.type_per_edge
+        reverse_edge_ids = C_sampled_subgraph.reverse_edge_ids
+        return_eids = reverse_edge_ids is not None
+        if return_eids:
+            reverse_edge_ids = self.edge_attributes[EID][reverse_edge_ids]
         if type_per_edge is None:
             # The sampled graph is already a homogeneous graph.
             node_pairs = (row, column)
@@ -237,6 +241,7 @@ class CSCSamplingGraph:
             # The sampled graph is a fused homogenized graph, which need to be
             # converted to heterogeneous graphs.
             node_pairs = defaultdict(list)
+            reverse_hetero_edge_ids = {}
             for etype, etype_id in self.metadata.edge_type_to_id.items():
                 src_ntype, _, dst_ntype = etype_str_to_tuple(etype)
                 src_ntype_id = self.metadata.node_type_to_id[src_ntype]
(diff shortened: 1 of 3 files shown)


python/dgl/graphbolt/impl/csc_sampling_graph.py
```diff
@@ -230,13 +230,18 @@ def _convert_to_sampled_subgraph(
         )
         row = C_sampled_subgraph.indices
         type_per_edge = C_sampled_subgraph.type_per_edge
+        reverse_edge_ids = C_sampled_subgraph.reverse_edge_ids
```
frozenbugs, line 233: change all reverse to original? reverse_edge_ids -> original_edge_ids.

python/dgl/graphbolt/impl/csc_sampling_graph.py
```diff
@@ -8,7 +8,7 @@
 
 import torch
 
-from ...base import ETYPE
+from ...base import EID, ETYPE
```
frozenbugs, line 11: Graphbolt should not have dependency on DGL graph's EID definition.

Note: the ETYPE was added for converting DGL graph to csc_sampling_graph.

python/dgl/graphbolt/impl/csc_sampling_graph.py
```diff
@@ -303,6 +315,15 @@ def sample_neighbors(
             corresponding to each neighboring edge of a node. It must be a 1D
             floating-point or boolean tensor, with the number of elements
             equalling the total number of edges.
+        return_eids: bool
+            Boolean indicating whether the edge IDs of sampled edges, should be
```
frozenbugs, line 319: Do we always have eids as edge attribute in CSCSamplingGraph?

peizhou001: No, so the pre-check should check that.

frozenbugs: We should mention it here in the documentation, say it works only if eids is edge attribute.

Meanwhile, we need to ask users to specify whether the eids should be added as edge attribute in the preprocess or not.

python/dgl/graphbolt/impl/csc_sampling_graph.py
peizhou001: Will be resolved in another PR

python/dgl/graphbolt/impl/csc_sampling_graph.py
```diff
@@ -230,13 +230,20 @@ def _convert_to_sampled_subgraph(
         )
         row = C_sampled_subgraph.indices
         type_per_edge = C_sampled_subgraph.type_per_edge
+        reverse_edge_ids = C_sampled_subgraph.reverse_edge_ids
+        return_eids = reverse_edge_ids is not None
+        if return_eids:
+            reverse_edge_ids = self.edge_attributes[ORIGINAL_EDGE_ID][
+                reverse_edge_ids
+            ]
```
frozenbugs, line 238: We should check whether `ORIGINAL_EDGE_ID in self.edge_attributes` rather than `reverse_edge_ids is not None` here.
```suggestion
        has_original_eids = ORIGINAL_EDGE_ID in self.edge_attributes
        if has_original_eids:
            original_edge_ids = self.edge_attributes[ORIGINAL_EDGE_ID][
                original_edge_ids
            ]
```

frozenbugs (approved):

frozenbugs (approved):

PR #6364, opened 2023-09-21
[Sparse] Modify the test of sparse_matrix_str

yxy235:
## Description

Modify the test of sparse_matrix which has changed "print" to "assert".
## Checklist
- [ ] Code is well-documented


tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -596,14 +596,30 @@ def test_print():
     col = torch.tensor([2, 1, 3]).to(ctx)
     val = torch.tensor([1.0, 1.0, 2.0]).to(ctx)
     A = from_coo(row, col, val)
-    print(A)
+    expect_res = str(
```
frozenbugs, line 599:
```suggestion
    expected = str(
```

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -596,14 +596,30 @@ def test_print():
...
+        """SparseMatrix(indices=tensor([[1, 1, 3],
+                             [2, 1, 3]]),
+             values=tensor([1., 1., 2.]),
+             shape=(4, 4), nnz=3)"""
```
frozenbugs, line 603:
```suggestion
        """
SparseMatrix(indices=tensor([[1, 1, 3],
                             [2, 1, 3]]),
             values=tensor([1., 1., 2.]),
             shape=(4, 4), nnz=3)
        """
```

tests/python/pytorch/sparse/test_sparse_matrix.py
```diff
@@ -596,14 +596,30 @@ def test_print():
...
     # vector-shape non zero
     row = torch.tensor([1, 1, 3]).to(ctx)
     col = torch.tensor([2, 1, 3]).to(ctx)
-    val = torch.randn(3, 2).to(ctx)
+    val = torch.tensor(
+        [[1.3080, 1.5984], [-0.4126, 0.7250], [-0.5416, -0.7022]]
+    ).to(ctx)
     A = from_coo(row, col, val)
-    print(A)
+    expect_res = str(
```
frozenbugs, line 614: ditto.

tests/python/pytorch/sparse/test_sparse_matrix.py
frozenbugs: Not sure whether this passes the test or not, but looks better in this way.

frozenbugs (approved):

tests/python/pytorch/sparse/test_sparse_matrix.py
yxy235: > Not sure whether this passes the test or not, but looks better in this way.

I tried it locally. It does not pass the test, for there will be another '\n' before the expected result, which causes assertion error.

frozenbugs: I see, then ignore this.

PR #6374, opened 2023-09-22
[Graphbolt] Support get feature size.

yxy235:
## Description

Add size api to class Feature and class FeatureStore. 
By using feature.size(), you can get the size of the feature directly.
If you want to get the size of a unique feature in the featruestore, try to use featuresotre.size(domain, type_name, feature_name.
## Checklist
- [ ] Code is well-documented


python/dgl/graphbolt/feature_store.py
```diff
@@ -26,6 +26,22 @@ def read(self, ids: torch.Tensor = None):
         """
         raise NotImplementedError
 
+    def size(self, ids: torch.Tensor = None):
```
frozenbugs, line 29: The size api should not need id as parameter, feature of all ids should have the same size.

python/dgl/graphbolt/feature_store.py
```diff
@@ -26,6 +26,22 @@ def read(self, ids: torch.Tensor = None):
         """
         raise NotImplementedError
 
+    def size(self):
+        """Get the size of the feature.
+
+        Parameters
+        ----------
+        ids : torch.Tensor, optional
```
frozenbugs, line 34: Remove?

python/dgl/graphbolt/impl/basic_feature_store.py
```diff
@@ -57,6 +57,31 @@ def read(
...
+            The node or edge type name.
+        feature_name : str
+            The feature name.
+        Returns
+        -------
+        int
+            The size of the feature.
+        """
+        if domain is None and type_name is None and feature_name is None:
+            for value in self._features.values():
```
frozenbugs, line 81: We don't need to return the number of different features in feature store. The key should not be optional.

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -77,6 +77,15 @@ def read(self, ids: torch.Tensor = None):
...
         return self._tensor[ids]
 
+    def size(self):
+        """Get the size of the feature.
+
+        -------
+        int
+            The size of the feature.
+        """
+        return len(self._tensor)
```
frozenbugs, line 87: This is actually a incorrect implementation.

For 2-d tensor, it returns number of rows instead of number of column.
```suggestion
        return self._tensor.size(1)
```

tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
```diff
@@ -47,6 +50,12 @@ def test_torch_based_feature(in_memory):
             feature_b.read(torch.tensor([1])),
             torch.tensor([[4, 5, 6]]),
         )
+
+        # Test get the size of the entire feature.
+        assert feature_a.size() == 3
+        assert feature_b.size() == 2
```
frozenbugs, line 56: feature_b.size() should be 3 as well.

frozenbugs (approved):

tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
```diff
@@ -47,6 +53,13 @@ def test_torch_based_feature(in_memory):
             feature_b.read(torch.tensor([1])),
             torch.tensor([[4, 5, 6]]),
         )
+
+        # Test get the size of the entire feature.
+        assert feature_a.size() == torch.Size([3])
```
frozenbugs, line 58: Actually feature_a.size() should be torch.Size([1])

yxy235: > Actually feature_a.size() should be torch.Size([1]), see L44 test case.

Can I take it that size API is actually the size of the data read out by rows from the tensor, and do I need to additionally consider the case where the feature is empty?

frozenbugs: Yes.

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -85,6 +85,17 @@ def read(self, ids: torch.Tensor = None):
...
         return self._tensor[ids]
 
+    def size(self):
+        """Get the size of the feature.
+
+        -------
+        torch.Size
+            The size of the feature.
+        """
+        if len(self._tensor.shape) == 1:
```
frozenbugs, line 95: We don't need this anymore, right?

yxy235: Right. I just merge the `master` branch into this branch. I am modifying the class and the tests.

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -85,6 +85,17 @@ def read(self, ids: torch.Tensor = None):
...
+            return torch.Size([1])
+        return self._tensor[0].shape
```
frozenbugs, line 97: This should work?
```suggestion
        return self._tensor.size()[1:]
```

tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
```diff
@@ -57,13 +61,24 @@ def test_torch_based_feature(in_memory):
             feature_b.read(), torch.tensor([[[1, 2], [3, 4]], [[1, 2], [3, 4]]])
         )
 
+#         # Test get the size of the entire feature.
```
frozenbugs, line 64: Remove if we don't need this, never just comment out unused code.

yxy235: OK. This code I will modify later to test the size API.

python/dgl/graphbolt/impl/torch_based_feature_store.py
yxy235: Right.

PR #6409, opened 2023-10-09
[Sparse]Fix device check 

### LINKED ISSUE #6393: [Bug]DGL sparse run error message is incomplete
## 🐛 Bug

DGL sparse run error message is incomplete, make it hard to debug for user.

## To Reproduce

In DGL sparse, the op like bsddmm needs all inputs are on the same device, howerver, if inputs are not on the same device, it won't report "they are not on the same device" but report Segmentation fault, it is hard to debug for user.

```
>>> import torch
>>> import dgl.sparse as dglsp
>>> indices = torch.tensor([[1, 1, 2], [2, 3, 3]])
>>> val = torch.arange(1, 4).float()
>>> A = dglsp.spmatrix(indices, val, (3, 4))
>>> X1 = torch.arange(0, 3 * 5 * 2).view(3, 5, 2).float() 
>>> X2 = torch.arange(0, 5 * 4 * 2).view(5, 4, 2).float()
>>> X1 = X1.to("cuda:0")
>>> X2 = X2.to("cuda:0")
>>> dglsp.bsddmm(A, X1, X2) // A is still on the CPU
Segmentation fault (core dumped)
```

## Expected behavior

Should we add the input device check for the DGL sparse operator?

xiangyuzhi:
## Description
Add the device check for sparse operator. When user call the function and the device of input matrix is not the same as input tensor, raise an error.

Resolve #6393.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/sparse/matmul.py
+++ b/python/dgl/sparse/matmul.py
@@ -123,6 +123,7 @@ def spspmm(A: SparseMatrix, B: SparseMatrix) -> SparseMatrix:
     assert isinstance(
         B, SparseMatrix
     ), f"Expect A2 to be a SparseMatrix object, got {type(B)}."
+
     return SparseMatrix(
         torch.ops.dgl_sparse.spspmm(A.c_sparse_matrix, B.c_sparse_matrix)
     )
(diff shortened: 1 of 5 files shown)


python/dgl/sparse/matmul.py
```diff
@@ -123,7 +123,6 @@ def spspmm(A: SparseMatrix, B: SparseMatrix) -> SparseMatrix:
     assert isinstance(
         B, SparseMatrix
     ), f"Expect A2 to be a SparseMatrix object, got {type(B)}."
-
```
czkkkkkk, line 126: Remove this change.

czkkkkkk (approved):

PR #6449, opened 2023-10-16
[GraphBolt] Ensure the contiguity of the `TorchBasedFeature._tensor`.

keli-wen:
## Description

The features of the raw OGB dataset may not always be **C-contiguous**; sometimes, they are **Fortran-contiguous** (column-major order). A tensor derived from NumPy will be contiguous if the original numpy.ndarray is C-contiguous. Operations can significantly slow down when the tensor is not contiguous.

The time complexity of `is_contiguous` is $O(dim)$ ([refer to PyTorch](https://github.com/pytorch/pytorch/blob/a7f6b0ab4fa4e43a2fa7bcb53825277db88b55f0/torch/lib/TH/generic/THTensor.c#L639-L654)), which is essentially cost-free.

Therefore, we have added some checks and transformations in `TorchBasedFeature` to ensure contiguity.

---
The following code can be used to check the contiguity of a `numpy.ndarray`:
```shell
In [x]: raw_ogb_feat.flags
Out[x]: 
  C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
```

## Checklist
- [x] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/impl/torch_based_feature_store.py
+++ b/python/dgl/graphbolt/impl/torch_based_feature_store.py
@@ -69,6 +69,12 @@ class TorchBasedFeature(Feature):
             f"dimension of torch_feature in TorchBasedFeature must be greater "
             f"than 1, but got {torch_feature.dim()} dimension."
         )
+        if not torch_feature.is_contiguous():
+            Warning(
+                f"torch_feature in TorchBasedFeature is not contiguous, "
+                f"so it will be copied to contiguous memory."
+            )
+            torch_feature = torch_feature.contiguous()
         self._tensor = torch_feature
 
     def read(self, ids: torch.Tensor = None):
--- a/tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
+++ b/tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
@@ -86,6 +86,44 @@ def test_torch_based_feature(in_memory):
         a = b = None
         feature_a = feature_b = None
 
+        # Test loaded tensors' contiguity from C/Fortran contiguous ndarray.
+        C_contiguous_numpy = np.array([[1, 2, 3], [4, 5, 6]], order="C")
+        Fortran_contiguous_numpy = np.array([[1, 2, 3], [4, 5, 6]], order="F")
+        assert C_contiguous_numpy.flags["C_CONTIGUOUS"]
+        assert Fortran_contiguous_numpy.flags["F_CONTIGUOUS"]
+        np.save(
+            os.path.join(test_dir, "C_contiguous_numpy.npy"), C_contiguous_numpy
+        )
+        np.save(
+            os.path.join(test_dir, "Fortran_contiguous_numpy.npy"),
+            Fortran_contiguous_numpy,
+        )
+
+        cur_mmap_mode = None
+        if not in_memory:
+            cur_mmap_mode = "r+"
+        feature_a = gb.TorchBasedFeature(
+            torch.from_numpy(
+                np.load(
+                    os.path.join(test_dir, "C_contiguous_numpy.npy"),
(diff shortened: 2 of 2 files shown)


python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -69,6 +69,12 @@ def __init__(self, torch_feature: torch.Tensor):
             f"dimension of torch_feature in TorchBasedFeature must be greater "
             f"than 1, but got {torch_feature.dim()} dimension."
         )
+        if not torch_feature.is_contiguous():
```
peizhou001, line 72: Is this necessary? I suspect `contiguous` will check it inside.

tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
```diff
@@ -86,6 +86,44 @@ def test_torch_based_feature(in_memory):
         a = b = None
         feature_a = feature_b = None
 
+        # Test loaded tensors' contiguity from C/Fortran contiguous ndarray.
+        C_contiguous_numpy = np.array([[1, 2, 3], [4, 5, 6]], order="C")
```
peizhou001, line 90: Change the name to something like contiguous and non-contiguous for clarify.

tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
```diff
@@ -86,6 +86,44 @@ def test_torch_based_feature(in_memory):
...
+        Fortran_contiguous_numpy = np.array([[1, 2, 3], [4, 5, 6]], order="F")
```
peizhou001, line 91: Ditto

python/dgl/graphbolt/impl/torch_based_feature_store.py
keli-wen: Yes, I do want to alert the user to the weird phenomenon, so they will check their own dataset. Otherwise, the user may never realize that their features aren't contiguous.

tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
keli-wen: Done

frozenbugs (approved):

python/dgl/graphbolt/utils/internal.py
```diff
@@ -19,7 +19,14 @@ def _read_numpy_data(path, in_memory=True):
...
+        _tensor = _read_torch_data(path)
+        if not _tensor.is_contiguous():
+            Warning(
+                "The tensor read from disk is not contiguous, "
+                "so it will be copied to contiguous memory."
+            )
+            _tensor = _tensor.contiguous()
+        return _tensor
     elif fmt == "numpy":
         return _read_numpy_data(path, in_memory=in_memory)
```
Rhett-Ying, line 31: for `numpy` data, we don't need to make it `contiguous` when read? Instead, we do it when saving?

python/dgl/graphbolt/utils/internal.py
```diff
@@ -36,9 +36,22 @@ def save_data(data, path, fmt):
 
     # Perform necessary conversion.
     if fmt == "numpy" and isinstance(element, torch.Tensor):
-        element = element.cpu().numpy()
+        element = element.cpu()
```
frozenbugs, line 39: what's element?

frozenbugs: Do we have unit test for this function?

keli-wen: How should I add a new test? Do we need to create a new directory, such as `test/python/pytorch/graphbolt/utils/`?

PR #6459, opened 2023-10-17
[GraphBolt] enable indexing for ItemSetDict

Rhett-Ying:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/itemset.py
+++ b/python/dgl/graphbolt/itemset.py
@@ -266,6 +266,15 @@ class ItemSetDict:
         assert all(
             self._names == itemset.names for itemset in itemsets.values()
         ), "All itemsets must have the same names."
+        try:
+            # For indexable itemsets, we compute the offsets for each itemset
+            # in advance to speed up indexing.
+            offsets = [0] + [
+                len(itemset) for itemset in self._itemsets.values()
+            ]
+            self._offsets = torch.tensor(offsets).cumsum(0)
+        except TypeError:
+            self._offsets = None
 
     def __iter__(self) -> Iterator:
         for key, itemset in self._itemsets.items():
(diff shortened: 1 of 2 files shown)


python/dgl/graphbolt/itemset.py
```diff
@@ -255,6 +267,44 @@ def __iter__(self) -> Iterator:
...
+        except TypeError:
+            raise TypeError(
+                f"{type(self).__name__} instance doesn't support indexing."
+            )
+        offsets = [len(itemset) for itemset in self._itemsets.values()]
+        total_num = sum(offsets)
+        if isinstance(idx, int):
+            if idx < 0:
+                idx += total_num
+            assert idx < total_num, (
```
frozenbugs, line 282: should we assert idx >= 0?

python/dgl/graphbolt/itemset.py
```diff
@@ -255,6 +267,44 @@ def __iter__(self) -> Iterator:
...
+                f"Index {idx} out of range for {type(self).__name__} "
+                f"instance with length {total_num}."
+            )
+            for key, offset in zip(self._itemsets.keys(), offsets):
```
frozenbugs, line 286: make offset prefix sum and use binary search?

python/dgl/graphbolt/itemset.py
```diff
@@ -255,6 +267,44 @@ def __iter__(self) -> Iterator:
...
+                if idx < offset:
+                    return {key: self._itemsets[key][idx]}
+                idx -= offset
+        elif isinstance(idx, slice):
+            start, stop, step = idx.indices(total_num)
+            data = {}
+            for key, offset in zip(self._itemsets.keys(), offsets):
+                if start < offset:
```
frozenbugs, line 294: what if start is within a type and stop is within another type?

python/dgl/graphbolt/itemset.py
Rhett-Ying: no as negative index is converted to positive ones. and I'd like to support negative indexing

python/dgl/graphbolt/itemset.py
```diff
@@ -255,6 +267,44 @@ def __iter__(self) -> Iterator:
...
 
+    def __getitem__(self, idx: Union[int, slice, Iterable]) -> Dict[str, Tuple]:
+        try:
+            len(self)
+        except TypeError:
+            raise TypeError(
+                f"{type(self).__name__} instance doesn't support indexing."
+            )
+        offsets = [len(itemset) for itemset in self._itemsets.values()]
+        total_num = sum(offsets)
```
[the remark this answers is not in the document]
Rhett-Ying: I thought about it. pros of moving to init is to avoid `sum` in runtime. cons is it needs to check  if `len` supported and init if supported. multiple lines of code will be added into `__init__`. I don't think it's elegant. And the reason why I leave it here is the overhead should be trivial: instantiate offset and sum it.

python/dgl/graphbolt/itemset.py
Rhett-Ying: it's supported and covered in testcase.

python/dgl/graphbolt/itemset.py
frozenbugs: I meant after idx += total_num, the idx should be non-negative.

python/dgl/graphbolt/itemset.py
Rhett-Ying: the main reason I choose this implementation is to be easy to handle range that across multiple keys. and the complexity is trivial and binary search does not bring obvious improvement in most cases.

python/dgl/graphbolt/itemset.py
Rhett-Ying: right. should be added.

python/dgl/graphbolt/itemset.py
Rhett-Ying: oh, I was wrong. in this case, yes. binary search is better

Rhett-Ying: BS is applied

Rhett-Ying: But I'm not sure it really perform better than original one.

python/dgl/graphbolt/itemset.py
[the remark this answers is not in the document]
Rhett-Ying: moved to `__init__`

frozenbugs (approved):

PR #6507, opened 2023-10-30
[Graphbolt] Enable cuda optimizations for UVAIndexSelect.

czkkkkkk:
## Description

This PR copies the kernel optimizations from https://github.com/dmlc/dgl/blob/master/src/array/cuda/uvm/array_index_select_uvm.cu.
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/src/cuda/index_select_impl.cu
+++ b/graphbolt/src/cuda/index_select_impl.cu
@@ -9,15 +9,37 @@
 #include <numeric>
 
 #include "../index_select.h"
+#include "./utils.h"
 
 namespace graphbolt {
 namespace ops {
 
+/** @brief Index select operator implementation for feature size 1. */
+template <typename DType, typename IdType>
+__global__ void IndexSelectSingleKernel(
+    const DType* input, const int64_t input_len, const IdType* index,
+    const int64_t output_len, DType* output,
+    const int64_t* permutation = nullptr) {
+  int64_t out_row_index = blockIdx.x * blockDim.x + threadIdx.x;
+  int stride = gridDim.x * blockDim.x;
+  while (out_row_index < output_len) {
+    assert(index[out_row_index] >= 0 && index[out_row_index] < input_len);
+    const auto out_row =
+        permutation ? permutation[out_row_index] : out_row_index;
+    output[out_row] = input[index[out_row_index]];
+    out_row_index += stride;
+  }
+}
+
+/**
+ * @brief Index select operator implementation for feature size > 1.
+ */
 template <typename DType, typename IdType>
 __global__ void IndexSelectMultiKernel(
     const DType* const input, const int64_t input_len,
     const int64_t feature_size, const IdType* const index,
-    const int64_t output_len, DType* const output) {
+    const int64_t output_len, DType* const output,
+    const int64_t* permutation = nullptr) {
   int64_t out_row_index = blockIdx.x * blockDim.y + threadIdx.y;
 
   const int64_t stride = blockDim.y * gridDim.x;
@@ -26,10 +48,49 @@ __global__ void IndexSelectMultiKernel(
     int64_t column = threadIdx.x;
     const int64_t in_row = index[out_row_index];
     assert(in_row >= 0 && in_row < input_len);
+    const auto out_row =
+        permutation ? permutation[out_row_index] : out_row_index;
     while (column < feature_size) {
-      output[out_row_index * feature_size + column] =
+      output[out_row * feature_size + column] =
           input[in_row * feature_size + column];
       column += blockDim.x;
+    }
+    out_row_index += stride;
+  }
+}
+
+/**
+ * @brief Index select operator implementation for feature size > 1.
+ *
+ * @note This is a cross-device access version of IndexSelectMultiKernel. Since
+ * the memory access over PCIe is more sensitive to the data access aligment
+ * (cacheline), we need a separate version here.
+ */
+template <typename DType, typename IdType>
+__global__ void IndexSelectMultiKernelAligned(
+    const DType* const input, const int64_t input_len,
+    const int64_t feature_size, const IdType* const index,
+    const int64_t output_len, DType* const output,
+    const int64_t* permutation = nullptr) {
+  int64_t out_row_index = blockIdx.x * blockDim.y + threadIdx.y;
+
+  const int64_t stride = blockDim.y * gridDim.x;
+
+  while (out_row_index < output_len) {
+    int64_t col = threadIdx.x;
+    const int64_t in_row = index[out_row_index];
+    assert(in_row >= 0 && in_row < input_len);
+    const int64_t idx_offset =
+        ((uint64_t)(&input[in_row * feature_size]) % GPU_CACHE_LINE_SIZE) /
+        sizeof(DType);
+    col = col - idx_offset;
+    const auto out_row =
+        permutation ? permutation[out_row_index] : out_row_index;
+    while (col < feature_size) {
+      if (col >= 0)
+        output[out_row * feature_size + col] =
+            input[in_row * feature_size + col];
+      col += blockDim.x;
     }
     out_row_index += stride;
   }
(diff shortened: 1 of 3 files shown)


graphbolt/src/cuda/index_select_impl.cu
```diff
@@ -46,19 +107,41 @@ torch::Tensor UVAIndexSelectImpl_(torch::Tensor input, torch::Tensor index) {
...
+        input_ptr, input_len, index_sorted_ptr, return_len, ret_ptr,
+        permutation_ptr);
+  } else {
+    dim3 block(512, 1);
+    while (static_cast<int64_t>(block.x) >= 2 * feature_size) {
+      block.x >>= 1;
+      block.y <<= 1;
+    }
+    const dim3 grid((return_len + block.y - 1) / block.y);
+    if (feature_size * sizeof(DType) < 2 * GPU_CACHE_LINE_SIZE) {
```
czkkkkkk, line 133: @mfbalin Could you explain why we choose `IndexSelectMultiKernel` in this case? From my understanding, `IndexSelectMultiKernelAligned` should be better than `IndexSelectMultiKernel` in any cases.

graphbolt/src/cuda/index_select_impl.cu
```diff
@@ -46,19 +107,41 @@ torch::Tensor UVAIndexSelectImpl_(torch::Tensor input, torch::Tensor index) {
                                       .dtype(input.dtype())
                                       .device(c10::DeviceType::CUDA));
   DType* input_ptr = input.data_ptr<DType>();
-  IdType* index_ptr = index.data_ptr<IdType>();
   DType* ret_ptr = ret.data_ptr<DType>();
+
+  // Sort the index to improve the memory access pattern.
+  torch::Tensor sorted_index, permutation;
+  std::tie(sorted_index, permutation) = torch::sort(index);
```
mfbalin, line 114: Can we use CUB sort here? We can see the shape of the input, determine the number of bits required to radix sort the input, and use that for the sort. Should be faster. Like:
```
auto N = input.size(0);
int i = 0;
while ((1ll << i) < N)
  i++;
// pass i into cub::radixsort
```

mfbalin (changes requested):
This code can be optimized further for future use cases. I am planning on revisiting it later.

graphbolt/src/cuda/index_select_impl.cu
mfbalin: Using fewer SMs will be important when we want to overlap the copies with other ongoing computations. I will revisit this code to minimize SM usage further.

graphbolt/src/cuda/index_select_impl.cu
```diff
@@ -46,19 +107,41 @@ torch::Tensor UVAIndexSelectImpl_(torch::Tensor input, torch::Tensor index) {
...
+  const int64_t* permutation_ptr = permutation.data_ptr<int64_t>();
+
   cudaStream_t stream = 0;
-  dim3 block(512, 1);
-  // Find the smallest block size that can fit the feature_size.
-  while (static_cast<int64_t>(block.x) >= 2 * feature_size) {
-    block.x >>= 1;
-    block.y <<= 1;
+
+  if (feature_size == 1) {
```
frozenbugs, line 120: Just to confirm, it the feature_size here the size of feature for each node or each edge?
Can you add documentation at the top of the cu file to explain how each option here improves each situation?

frozenbugs:
Did you test the throughput of the new work compare to existing code?

graphbolt/src/cuda/index_select_impl.cu
czkkkkkk: > I didn't author the if checks in this code. In my opinion, the threshold should be modified to simply
<= GPU_CACHE_LINE_SIZE. I think aligned is faster otherwise. You can benchmark it for different values to find out which one is better for sure. When there is 1 or less than 1 cache line of feature dimension, then the nonaligned version is more resource-efficient, meaning has the potential to use fewer SMs.

I see. I will only check the size with one GPU_CACHE_LINE_SIZE.

> Can we use CUB sort here?

Yes, we can. But it requires large changes. I will enable it in a later PR.

graphbolt/src/cuda/index_select_impl.cu
czkkkkkk: > is the feature_size here the size of feature for each node or each edge?

Can be either node or edge feature. The index_select operator is designed for our feature store.

czkkkkkk:
> Did you test the throughput of the new work compare to existing code?

Not yet.

tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
```diff
@@ -136,44 +138,34 @@ def test_torch_based_feature(in_memory):
...
+    )
+    tensor = torch.arange(0, reduce(mul, shape), dtype=dtype).reshape(shape)
+    test_tensor = tensor.clone().detach()
+    test_tensor_cuda = test_tensor.cuda()
+    pinned_tensor = tensor.pin_memory()
 
-    feature_a = gb.TorchBasedFeature(a)
-    feature_b = gb.TorchBasedFeature(b)
-    feature_c = gb.TorchBasedFeature(c)
+    feature = gb.TorchBasedFeature(pinned_tensor)
```
frozenbugs, line 154: better to 
```
feature = gb.TorchBasedFeature(tensor).pin_memory_()
```
Instead of pin the tensor directly (better align with our API).

tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
```diff
@@ -136,44 +138,34 @@ def test_torch_based_feature(in_memory):
...
-        feature_b.read(torch.tensor([1], dtype=idtype).cuda()),
-        torch.tensor([[[4, 5], [6, 7]]], dtype=dtype).cuda(),
+        feature.read(torch.tensor([0], dtype=idtype).cuda()),
+        test_tensor_cuda[[0]],
     )
-    assert feature_b.read(torch.tensor([1], dtype=idtype).cuda()).is_cuda
+    assert feature.read(torch.tensor([0], dtype=idtype).cuda()).is_cuda
 
-    assert feature_c.read().is_cuda
+    assert feature.read().is_cuda
```
frozenbugs, line 164: duplicate with L157.
```suggestion
```

tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
```diff
@@ -136,44 +138,34 @@ def test_torch_based_feature(in_memory):
...
+    test_tensor = tensor.clone().detach()
+    test_tensor_cuda = test_tensor.cuda()
+    pinned_tensor = tensor.pin_memory()
 
-    feature_a = gb.TorchBasedFeature(a)
-    feature_b = gb.TorchBasedFeature(b)
-    feature_c = gb.TorchBasedFeature(c)
+    feature = gb.TorchBasedFeature(pinned_tensor)
 
+    assert torch.equal(feature.read(), test_tensor_cuda)
```
frozenbugs, line 156:
```suggestion
    # Test read entire pinned feature, the result should be on cuda.
    assert torch.equal(feature.read(), test_tensor_cuda)
```

tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
```diff
@@ -136,44 +138,34 @@ def test_torch_based_feature(in_memory):
...
-    )
-    assert feature_a.read(torch.tensor([0], dtype=idtype).cuda()).is_cuda
-    assert torch.equal(
-        feature_b.read(torch.tensor([1], dtype=idtype).cuda()),
-        torch.tensor([[[4, 5], [6, 7]]], dtype=dtype).cuda(),
+        feature.read(torch.tensor([0], dtype=idtype).cuda()),
+        test_tensor_cuda[[0]],
     )
-    assert feature_b.read(torch.tensor([1], dtype=idtype).cuda()).is_cuda
+    assert feature.read(torch.tensor([0], dtype=idtype).cuda()).is_cuda
```
frozenbugs, line 162:
```suggestion

    # Test read pinned feature with idx on cuda, the result should be on cuda. 
    assert feature.read(torch.tensor([0], dtype=idtype).cuda()).is_cuda
```

tests/python/pytorch/graphbolt/impl/test_torch_based_feature_store.py
```diff
@@ -136,44 +138,34 @@ def test_torch_based_feature(in_memory):
...
 
-    assert feature_c.read().is_cuda
+    assert feature.read().is_cuda
     assert torch.equal(
```
frozenbugs, line 165:
```suggestion
    # Test read pinned feature with idx on cpu, the result should be on cpu.
    assert torch.equal(
```

graphbolt/src/cuda/utils.h
```diff
@@ -0,0 +1,36 @@
...
+  while (ret > dim) {
+    ret = ret >> 1;
+  }
+  return ret;
+}
+
+}  // namespace cuda
+}  // namespace graphbolt
+
+#endif  // GRAPHBOLT_CUDA_UTILS_H_
```
frozenbugs, line 36: no newline.

graphbolt/src/cuda/utils.h
```diff
@@ -0,0 +1,36 @@
...
+ * @brief CUDA utilities.
+ */
+
+#ifndef GRAPHBOLT_CUDA_UTILS_H_
+#define GRAPHBOLT_CUDA_UTILS_H_
+
+namespace graphbolt {
+namespace cuda {
+
+// The cache line size of GPU
```
frozenbugs, line 14:
```suggestion
// The cache line size of GPU.
```

graphbolt/src/cuda/utils.h
```diff
@@ -0,0 +1,36 @@
...
+#define GPU_CACHE_LINE_SIZE 128
+// The max number of threads per block
```
frozenbugs, line 16:
```suggestion
// The max number of threads per block.
```

graphbolt/src/cuda/utils.h
```diff
@@ -0,0 +1,36 @@
...
+#define CUDA_MAX_NUM_THREADS 1024
+
+/** @brief Calculate the number of threads needed given the dimension length.
```
frozenbugs, line 19: What's "dimension length"? dimension or length? of length of the feature?

graphbolt/src/cuda/utils.h
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * It finds the biggest number that is smaller than min(dim, max_num_nthreads)
```
frozenbugs, line 21: It finds the largest power of two that is less than or equal to the minimum of dim and max_num_threads.

graphbolt/src/cuda/utils.h
```diff
@@ -0,0 +1,36 @@
...
+ * and is also power of two.
+ */
+inline int FindNumThreads(int dim, int max_num_threads = CUDA_MAX_NUM_THREADS) {
+  if (dim == 0) return 1;
+  int ret = max_num_threads;
+  while (ret > dim) {
+    ret = ret >> 1;
+  }
```
frozenbugs, line 29: note: when max_num_threads is not power of two, the result is not guaranteed to be power of two, which conflicts with the documentation.
```suggestion
  int ret = 1;
  while ((ret << 1) <= std::min(dim, max_num_threads)) {
    ret <<= 1;
  }
```

frozenbugs: consider rename dim to length.

graphbolt/src/cuda/utils.h
czkkkkkk: Should be "the size of the dimension to be processed". In the context of IndexSelectSingleKernel, it is the size of `index`.

graphbolt/src/cuda/utils.h
czkkkkkk: I used `size` to represent the size of a dimension.

frozenbugs (approved):

mfbalin (approved):
LGTM overall.

Noting for future revisit:
1) Sort can be changed to use fewer bits via cub sort
2) Regardless of the datatype, larger granularity accesses by each thread using a datatype such as float4 can be used to further reduce resource usage. In this case, the function will not depend on template parameter DType, but only on feature access granularity bytes. The types to use for copy are `uint8_t, uint16_t, uint32_t, float2, float4` and byte granularity being `1,2,4,8,16`. We will check if `gcd(feature dimension * sizeof DType, 16)`, and choose the appropriate type to perform the copy.
3) Limit the number of blocks used by taking a parameter indicating the max number of inflight accesses needed to saturate the PCI-e, as there will be diminishing returns because PCI-e bandwidth is limited.

PR #6515, opened 2023-11-01
[GraphBolt] Refactor CSCSamplingGraph and update the corresponding method.

keli-wen:
## Description

Since we will subsequently add a new class, `(non-fused)CSCSamplingGraph`, it would be better to refactor the current `CSCSamplingGraph` and its associated methods to facilitate the addition of the upcoming features.

However, two issues need to be addressed:

- [ ] The `BuiltinDataset` needs to be updated.
- [ ] We should determine whether to update the `SampledSubgraph` in `graphbolt/include/`. We can use `SampledSubgraph.type_per_edge == torch::nullopt` to distinguish the graph type or create a new class, such as `FusedSampledSubgraph`.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/docs/source/api/python/dgl.graphbolt.rst
+++ b/docs/source/api/python/dgl.graphbolt.rst
@@ -56,7 +56,7 @@ Standard Implementations
 
     OnDiskDataset
     BuiltinDataset
-    CSCSamplingGraph
+    FusedCSCSamplingGraph
     UniformNegativeSampler
     NeighborSampler
     LayerNeighborSampler
--- a/graphbolt/src/fused_csc_sampling_graph.cc
+++ b/graphbolt/src/fused_csc_sampling_graph.cc
@@ -1,10 +1,10 @@
 /**
  *  Copyright (c) 2023 by Contributors
- * @file csc_sampling_graph.cc
+ * @file fused_csc_sampling_graph.cc
  * @brief Source file of sampling graph.
  */
 
-#include <graphbolt/csc_sampling_graph.h>
+#include <graphbolt/fused_csc_sampling_graph.h>
 #include <graphbolt/serialize.h>
 #include <torch/torch.h>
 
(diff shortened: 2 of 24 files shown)


graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -173,7 +182,7 @@ CSCSamplingGraph::GetState() const {
   return state;
 }
 
-c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::InSubgraph(
+c10::intrusive_ptr<SampledSubgraph> FusedCSCSamplingGraph::InSubgraph(
```
frozenbugs, line 185: the c++ SampledSubgraph should be rename to FusedSampledSubgraph as well, let's do it in a separated PR.

frozenbugs (approved):

peizhou001 (approved):

docs/source/api/python/dgl.graphbolt.rst
```diff
@@ -56,7 +56,7 @@ Standard Implementations
 
     OnDiskDataset
     BuiltinDataset
-    CSCSamplingGraph
+    FusedCSCSamplingGraph
```
Rhett-Ying, line 59: will we add a new wrapper or interface class such as `CSCSamplingGraph` to incorporate fused and non-fused classes?

Rhett-Ying: non-fused one will be named as `CSCSamplingGraph`

graphbolt/src/fused_csc_sampling_graph.cc
keli-wen: Got

keli-wen:
@dgl-bot

keli-wen:
To ensure the CI test is successful, I have updated the `BuiltinDataset`: `test-only.zip` by regenerating it with the latest code. This may resolve the encountered `OnDiskGraphTopologyType` mismatch in the original dataset.

PR #6522, opened 2023-11-02
[GraphBolt] Fix Link Prediction eample evaluation function bug and Enable cuda execution

keli-wen:
## Description

This PR try to address the following issues:
- [x] Fix the bug in `evaluate` function. (In the original code, compacted_neg_src and compacted_neg_dst are 1D tensors, which will cause the bug in `evaluate` function. We need to reserve the original shape of neg_src and neg_dst to reshape the compacted_neg_src and compacted_neg_dst)
- [x] Enable the example cude execution. (Since, `neg_ratio` in eval/test set is $1000$, which is too large for GPU memory. We need to set different batch size for evaluation stage.)
- [x] Update the test code to keep consistent with the above changes.

Trainning Log:
```
$ python examples/sampling/graphbolt/link_prediction.py --device cuda --early-stop 1000 --epochs 10 --num-workers 8
...
later
```

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/examples/sampling/graphbolt/link_prediction.py
+++ b/examples/sampling/graphbolt/link_prediction.py
@@ -103,7 +103,9 @@ def create_dataloader(args, graph, features, itemset, is_train=True):
     # Initialize the ItemSampler to sample mini-batche from the dataset.
     ############################################################################
     datapipe = gb.ItemSampler(
-        itemset, batch_size=args.batch_size, shuffle=is_train
+        itemset,
+        batch_size=args.train_batch_size if is_train else args.eval_batch_size,
+        shuffle=is_train,
     )
 
     ############################################################################
@@ -311,7 +313,8 @@ def parse_args():
     parser.add_argument("--epochs", type=int, default=10)
     parser.add_argument("--lr", type=float, default=0.0005)
     parser.add_argument("--neg-ratio", type=int, default=1)
-    parser.add_argument("--batch-size", type=int, default=512)
+    parser.add_argument("--train-batch-size", type=int, default=512)
+    parser.add_argument("--eval-batch-size", type=int, default=2)
     parser.add_argument("--num-workers", type=int, default=4)
     parser.add_argument(
         "--early-stop",
--- a/examples/sampling/graphbolt/link_prediction.py
+++ b/examples/sampling/graphbolt/link_prediction.py
@@ -207,7 +207,7 @@ def create_dataloader(args, graph, features, itemset, is_train=True):
     return dataloader
 
 
-def to_binary_link_dgl_computing_pack(data: gb.MiniBatch):
+def to_binary_link_dgl_computing_pack(data: gb.DGLMiniBatch):
     """Convert the minibatch to a training pair and a label tensor."""
     pos_src, pos_dst = data.positive_node_pairs
     neg_src, neg_dst = data.negative_node_pairs
@@ -353,6 +353,7 @@ def main(args):
 
     in_size = features.size("node", None, "feat")[0]
     hidden_channels = 256
+    args.device = torch.device(args.device)
     model = SAGE(in_size, hidden_channels).to(args.device)
 
     # Model training.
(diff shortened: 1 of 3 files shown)


examples/sampling/graphbolt/link_prediction.py
```diff
@@ -311,7 +313,8 @@ def parse_args():
     parser.add_argument("--epochs", type=int, default=10)
     parser.add_argument("--lr", type=float, default=0.0005)
     parser.add_argument("--neg-ratio", type=int, default=1)
-    parser.add_argument("--batch-size", type=int, default=512)
+    parser.add_argument("--train-batch-size", type=int, default=512)
+    parser.add_argument("--eval-batch-size", type=int, default=2)
```
peizhou001, line 317: Why eval batch so small

keli-wen: Since the `neg_ratio` in evaluation stage is too large(1000), bigger eval batch_size will cause GPU OOM.

examples/sampling/graphbolt/link_prediction.py
```diff
@@ -350,7 +353,7 @@ def main(args):
 
     in_size = features.size("node", None, "feat")[0]
     hidden_channels = 256
-    model = SAGE(in_size, hidden_channels)
+    model = SAGE(in_size, hidden_channels).to(args.device)
```
Skeleton003, line 356: I suggest to add `args.device = torch.device(args.device)` before this.

examples/sampling/graphbolt/link_prediction.py
```diff
@@ -249,8 +251,8 @@ def evaluate(args, graph, features, itemset, model):
         )
```
frozenbugs, line 251: L210, the type should be DGLMiniBatch?

python/dgl/graphbolt/subgraph_sampler.py
```diff
@@ -116,8 +118,14 @@ def _node_pairs_preprocess(self, minibatch):
             compacted = compacted[2:]
             if has_neg_src:
                 compacted_negative_srcs = compacted.pop(0)
+                compacted_negative_srcs = compacted_negative_srcs.view(
+                    neg_src_shape
```
frozenbugs, line 122: Why not just neg_src.shape and remove L109?

python/dgl/graphbolt/subgraph_sampler.py
```diff
@@ -116,8 +118,14 @@ def _node_pairs_preprocess(self, minibatch):
...
+                )
             if has_neg_dst:
                 compacted_negative_dsts = compacted.pop(0)
+                compacted_negative_dsts = compacted_negative_dsts.view(
+                    neg_dst_shape
```
frozenbugs, line 127: ditto.

frozenbugs (approved):

python/dgl/graphbolt/subgraph_sampler.py
frozenbugs: Can you add a comment to elaborate why we need to reshape here?

examples/sampling/graphbolt/link_prediction.py
```diff
@@ -205,7 +207,7 @@ def create_dataloader(args, graph, features, itemset, is_train=True):
     return dataloader
 
 
-def to_binary_link_dgl_computing_pack(data: gb.MiniBatch):
+def to_binary_link_dgl_computing_pack(data: gb.DGLMiniBatch):
```
Skeleton003, line 210: Is this just a helper function or could we make it a member function of `gb.DGLMiniBatch` ?

Skeleton003 (approved):

examples/sampling/graphbolt/link_prediction.py
keli-wen: Since the `datapipe` do not generate corresponding label, we need to construct them manually. But, I do think it will be merged into the `datapipe` in the future.

Skeleton003 (approved):

PR #6523, opened 2023-11-03
[GraphBolt] enable more dtypes for sample_neighbors

Rhett-Ying:
## Description

Previously, `indptr` and `indices` are required to be `torch.int64` or runtime error shown below will be thrown:
```
RuntimeError: expected scalar type Long but found Int
```
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/src/fused_csc_sampling_graph.cc
+++ b/graphbolt/src/fused_csc_sampling_graph.cc
@@ -335,7 +335,7 @@ FusedCSCSamplingGraph::SampleNeighborsImpl(
         auto num_picked_neighbors_data_ptr =
             num_picked_neighbors_per_node.data_ptr<scalar_t>();
         num_picked_neighbors_data_ptr[0] = 0;
-        const auto nodes_data_ptr = nodes.data_ptr<int64_t>();
+        const auto nodes_data_ptr = nodes.data_ptr<scalar_t>();
 
         // Step 1. Calculate pick number of each node.
         torch::parallel_for(
@@ -356,7 +356,8 @@ FusedCSCSamplingGraph::SampleNeighborsImpl(
 
         // Step 2. Calculate prefix sum to get total length and offsets of each
         // node. It's also the indptr of the generated subgraph.
-        subgraph_indptr = torch::cumsum(num_picked_neighbors_per_node, 0);
+        subgraph_indptr =
+            num_picked_neighbors_per_node.cumsum(0, indptr_.scalar_type());
 
         // Step 3. Allocate the tensor for picked neighbors.
         const auto total_length =
--- a/graphbolt/src/fused_csc_sampling_graph.cc
+++ b/graphbolt/src/fused_csc_sampling_graph.cc
@@ -335,13 +335,12 @@ FusedCSCSamplingGraph::SampleNeighborsImpl(
         auto num_picked_neighbors_data_ptr =
             num_picked_neighbors_per_node.data_ptr<scalar_t>();
         num_picked_neighbors_data_ptr[0] = 0;
-        const auto nodes_data_ptr = nodes.data_ptr<indices_.scalar_type>();
 
         // Step 1. Calculate pick number of each node.
         torch::parallel_for(
             0, num_nodes, grain_size, [&](int64_t begin, int64_t end) {
               for (int64_t i = begin; i < end; ++i) {
-                const auto nid = nodes_data_ptr[i];
+                const auto nid = nodes[i].item<int64_t>();
                 TORCH_CHECK(
                     nid >= 0 && nid < NumNodes(),
                     "The seed nodes' IDs should fall within the range of the "
@@ -375,7 +374,7 @@ FusedCSCSamplingGraph::SampleNeighborsImpl(
         torch::parallel_for(
             0, num_nodes, grain_size, [&](int64_t begin, int64_t end) {
               for (int64_t i = begin; i < end; ++i) {
-                const auto nid = nodes_data_ptr[i];
+                const auto nid = nodes[i].item<int64_t>();
                 const auto offset = indptr_data[nid];
                 const auto num_neighbors = indptr_data[nid + 1] - offset;
                 const auto picked_number = num_picked_neighbors_data_ptr[i + 1];
--- a/graphbolt/src/fused_csc_sampling_graph.cc
+++ b/graphbolt/src/fused_csc_sampling_graph.cc
@@ -335,7 +335,7 @@ FusedCSCSamplingGraph::SampleNeighborsImpl(
         auto num_picked_neighbors_data_ptr =
             num_picked_neighbors_per_node.data_ptr<scalar_t>();
         num_picked_neighbors_data_ptr[0] = 0;
-        const auto nodes_data_ptr = nodes.data_ptr<scalar_t>();
+        const auto nodes_data_ptr = nodes.data_ptr<indices_.scalar_type>();
 
         // Step 1. Calculate pick number of each node.
         torch::parallel_for(
--- a/tests/python/pytorch/graphbolt/impl/test_fused_csc_sampling_graph.py
+++ b/tests/python/pytorch/graphbolt/impl/test_fused_csc_sampling_graph.py
@@ -683,8 +683,8 @@ def test_sample_neighbors_hetero(labor, indptr_dtype, indices_dtype):
 
     # Sample on both node types.
     nodes = {
-        "n1": torch.tensor([0], dtype=indptr_dtype),
-        "n2": torch.tensor([0], dtype=indptr_dtype),
+        "n1": torch.tensor([0], dtype=indices_dtype),
+        "n2": torch.tensor([0], dtype=indices_dtype),
     }
     fanouts = torch.tensor([-1, -1])
     sampler = graph.sample_layer_neighbors if labor else graph.sample_neighbors
--- a/tests/python/pytorch/graphbolt/impl/test_fused_csc_sampling_graph.py
+++ b/tests/python/pytorch/graphbolt/impl/test_fused_csc_sampling_graph.py
@@ -604,7 +604,10 @@ def test_in_subgraph_heterogeneous():
     F._default_context_str == "gpu",
     reason="Graph is CPU only at present.",
 )
-def test_sample_neighbors_homo():
+@pytest.mark.parametrize("labor", [False, True])
+@pytest.mark.parametrize("indptr_dtype", [torch.int32, torch.int64])
+@pytest.mark.parametrize("indices_dtype", [torch.int32, torch.int64])
+def test_sample_neighbors_homo(labor, indptr_dtype, indices_dtype):
     """Original graph in COO:
     1   0   1   0   1
     1   0   1   1   0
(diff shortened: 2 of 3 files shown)


python/dgl/graphbolt/impl/csc_sampling_graph.py
```diff
@@ -362,7 +362,7 @@ def _convert_to_homogeneous_nodes(self, nodes):
         for ntype, ids in nodes.items():
             ntype_id = self.metadata.node_type_to_id[ntype]
             homogeneous_nodes.append(ids + self.node_type_offset[ntype_id])
-        return torch.cat(homogeneous_nodes)
+        return torch.cat(homogeneous_nodes).to(self.csc_indptr.dtype)
```
frozenbugs, line 365: Ideally, we don't need to convert the data type during sampling, the dataset should use the correct dtype already.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -339,7 +339,8 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
 
         // Step 2. Calculate prefix sum to get total length and offsets of each
         // node. It's also the indptr of the generated subgraph.
-        subgraph_indptr = torch::cumsum(num_picked_neighbors_per_node, 0);
+        subgraph_indptr =
```
peizhou001, line 342: Does `.dtype `work?

python/dgl/graphbolt/impl/csc_sampling_graph.py
Rhett-Ying: Usually users are not aware of(even not care about it at all) the dtype requirement when working on self-curated dataset. Adding such dtype formatting is just an additional guard.

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -318,7 +318,7 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
         auto num_picked_neighbors_data_ptr =
             num_picked_neighbors_per_node.data_ptr<scalar_t>();
         num_picked_neighbors_data_ptr[0] = 0;
-        const auto nodes_data_ptr = nodes.data_ptr<int64_t>();
+        const auto nodes_data_ptr = nodes.data_ptr<scalar_t>();
```
peizhou001, line 321: Does it mean `seed nodes` should have same data type as `indices`?

graphbolt/src/csc_sampling_graph.cc
```diff
@@ -339,7 +339,8 @@ c10::intrusive_ptr<SampledSubgraph> CSCSamplingGraph::SampleNeighborsImpl(
 
         // Step 2. Calculate prefix sum to get total length and offsets of each
         // node. It's also the indptr of the generated subgraph.
-        subgraph_indptr = torch::cumsum(num_picked_neighbors_per_node, 0);
+        subgraph_indptr =
+            num_picked_neighbors_per_node.cumsum(0, indptr_.scalar_type());
```
peizhou001, line 343: Ditto

python/dgl/graphbolt/impl/csc_sampling_graph.py
```diff
@@ -444,6 +444,9 @@ def sample_neighbors(
         return self._convert_to_sampled_subgraph(C_sampled_subgraph)
 
     def _check_sampler_arguments(self, nodes, fanouts, probs_name):
+        assert (
```
peizhou001, line 447: If required, we can move this ahead to python.

graphbolt/src/csc_sampling_graph.cc
Rhett-Ying: whose `.dtype()`?

graphbolt/src/csc_sampling_graph.cc
Rhett-Ying: yes.

graphbolt/src/csc_sampling_graph.cc
Rhett-Ying: yes or it's always int `int64`.

python/dgl/graphbolt/impl/csc_sampling_graph.py
Rhett-Ying: where do mean exactly? checking here looks ok to me as it's for sample argument

Rhett-Ying:
As discussed offline, we'd better handle dtype issue in `preprocess` stage and throw exception/warning if mismatched instead of hacking and formatting silently.

graphbolt/src/csc_sampling_graph.cc
peizhou001: indptr_

graphbolt/src/csc_sampling_graph.cc
mfbalin: But indptr holds offsets into edges, which should be a datatype that can hold the total number of edges. Indices holds vertex ids, which should be a datatype that can represent maximum vertex id. I would expect seed nodes to have the same type as the indices array.

mfbalin: Then people can use int64 for `indptr` and int32 for `indices` if graph has less than 2B vertices but more than 2B edges.

graphbolt/src/csc_sampling_graph.cc
Rhett-Ying: not works

graphbolt/src/csc_sampling_graph.cc
Rhett-Ying: Now, I always format it to `int64_t` as it's for indexing only no matter the underlying data format is `int32_t` or `int64_t`. I think handling in this way is more general. What do you think?

python/dgl/graphbolt/impl/csc_sampling_graph.py
Rhett-Ying: removed such check as we don't need it any more in the latest implementation.

Rhett-Ying:
> As discussed offline, we'd better handle dtype issue in `preprocess` stage and throw exception/warning if mismatched instead of hacking and formatting silently.

For now, no explicit formatting is applied. So we don't need to handle dtype in `preprocess` stage as we previously discussed.

python/dgl/graphbolt/impl/csc_sampling_graph.py
Rhett-Ying: such format is removed as we don't need it any more in the latest implementation. please let me know if you have any more concerns.

tests/python/pytorch/graphbolt/impl/test_fused_csc_sampling_graph.py
```diff
@@ -615,17 +618,21 @@ def test_sample_neighbors_homo():
...
 
     # Generate subgraph via sample neighbors.
-    nodes = torch.LongTensor([1, 3, 4])
-    subgraph = graph.sample_neighbors(nodes, fanouts=torch.LongTensor([2]))
+    nodes = torch.tensor([1, 3, 4], dtype=indices_dtype)
+    sampler = graph.sample_layer_neighbors if labor else graph.sample_neighbors
+    subgraph = sampler(nodes, fanouts=torch.LongTensor([2]))
 
     # Verify in subgraph.
     sampled_num = subgraph.node_pairs[0].size(0)
```
frozenbugs, line 638: Maybe also assert the edges are actually in the graph in addition to only assert sampled_num?

frozenbugs (approved):

graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -335,13 +335,12 @@ FusedCSCSamplingGraph::SampleNeighborsImpl(
...
             num_picked_neighbors_per_node.data_ptr<scalar_t>();
         num_picked_neighbors_data_ptr[0] = 0;
-        const auto nodes_data_ptr = nodes.data_ptr<int64_t>();
 
         // Step 1. Calculate pick number of each node.
         torch::parallel_for(
             0, num_nodes, grain_size, [&](int64_t begin, int64_t end) {
               for (int64_t i = begin; i < end; ++i) {
-                const auto nid = nodes_data_ptr[i];
+                const auto nid = nodes[i].item<int64_t>();
```
mfbalin, line 343: I think this might have a negative effect on runtime if it uses dynamic polymorphism, I would have preferred casting nodes to the same dtype as indices before calling this function. Or utilizing AT_DISPATCH_INTEGRAL_TYPES for nodes as well.

graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -374,7 +374,7 @@ FusedCSCSamplingGraph::SampleNeighborsImpl(
...
```
mfbalin, line 377: same as above.

graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -335,13 +335,12 @@ FusedCSCSamplingGraph::SampleNeighborsImpl(
         auto num_picked_neighbors_data_ptr =
```
mfbalin, line 335: using indptr_ptr_t = scalar_t;

mfbalin: And after that `AT_DISTPATCH_INTEGRAL_TYPES` for nodes as well.

mfbalin (changes requested):
I think we can utilize one more dispatch for the `nodes` tensor accesses as well.

graphbolt/src/csc_sampling_graph.cc
mfbalin: I think it is good, but we can do it via using a dispatch and use the original type of `nids` tensor, such as `const int64_t nid = nids[i]`;

graphbolt/src/fused_csc_sampling_graph.cc
mfbalin: We can then write `const int64_t nid = nodes_ptr[i];`. `nodes_ptr` will be a pointer with original dtype of `nodes`.

graphbolt/src/fused_csc_sampling_graph.cc
Rhett-Ying: I though about it but I don't think we really need to do like this. And I think it's `static_cast<>` behind `item<>`. No overhead in runtime. We do need to make sure the underlying data type could be safely casted to the `int64_t`, but such assumption is also required even using `scalar_t`. What's more, if the underlying dtype is float, `const auto nid = scalar_t` returns float while `item<>` casts it to `int64_t`.

mfbalin: The problem is that with the `.item` member, the code needs to look at the underlying tensor type information each time it is called (inside of the for loop), go into the corresponding branch in the code, and then return the value. (This is performed for every access.)

mfbalin: This is basically a (minor, should be benchmarked) performance concern, the code is correct of course.

tests/python/pytorch/graphbolt/impl/test_fused_csc_sampling_graph.py
Rhett-Ying: we'd better do. As for now, we just check num only in all the related testcases, I'd like to add such assertion in another PR.

graphbolt/src/fused_csc_sampling_graph.cc
Rhett-Ying: Thanks for your suggestion.

Rhett-Ying: I haven't checked the implementation of `item<>`. But I guess `nodes[i].item<int64_t>()` is like
```
static_cast<int64_t>(nodes.data_ptr<scalar_t>()[i])
```
Is this workable?

mfbalin: Sorry, it is my mistake. I was slightly mistaken about the above description. Let me correct it below, the performance issue is basically due to the `nodes[i]` access:

https://github.com/pytorch/pytorch/blob/d70857bd9e425ffc5b1a32639fb6651462eeef97/aten/src/ATen/templates/TensorBody.h#L321 has the definition of the tensor::operator[], which is being used in `nodes[i].item<int64_t>();`. It calls the tensor::select function. Which is defined here: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/TensorShape.cpp#L1816.

I am writing it for referencing and self-correcting.

mfbalin: The array access needs to happen after C++ knows the type information, which means we need to access it via a pointer whose type C++ already knows, which means we have to unfortunately use a dispatch for maximum performance especially because it is performed in a for loop.

mfbalin: I hope I made my point clear, sorry for the confusing explanation, I am not the best communicator.

Rhett-Ying: I see. So point is we'd better always prefer `nodes.data_ptr<sclar_t>()[i]` than `nodes[i]` for indexing on tensor as the latter one requires to call `select()` each time while the previous one does not as we have already accessing on the underlying data pointer directly. Right?

mfbalin: Exactly, and `nodes.data_ptr<scalar_t>()` unfortunately is inconvenient and requires a dispatch to get `scalar_t`. But this way, we pay the dispatch cost outside any loops.

PR #6531, opened 2023-11-04
[Graphbolt] Add MiniBatchBase

### LINKED ISSUE #6497: [Graphbolt] Feature fetcher support DGLMinibatch.
## 🔨Work Item

**IMPORTANT:**

czkkkkkk:
## Description
Resolve #6497 .

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/minibatch.py
+++ b/python/dgl/graphbolt/minibatch.py
@@ -1,5 +1,6 @@
 """Unified data structure for input and ouput of all the stages in loading process."""
 
+from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from typing import Dict, List, Tuple, Union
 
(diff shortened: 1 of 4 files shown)


czkkkkkk:
@frozenbugs , @peizhou001 , I added a abstract class for `MiniBatch` and `DGLMiniBatch`. But I still need your help for several issues when fetching features.
1. Seems that `DGLBlock` does not have `original_edge_ids`. How can I retrieve the edge ids when fetching edge features for `DGLBlock`?
2. I moved as many attributes of `MiniBatch` and `DGLMiniBatch` to `MiniBatchBase` as I can. Please check whether there are missing attributes that are shared by the two classes.
3. To align with `DGLMiniBatch`, I renamed the `seed_nodes` in `MiniBatch` to `output_nodes`. I may need to refactor all the usages of `seed_nodes` in`MiniBatch` later. Please check whether this is reasonable.

python/dgl/graphbolt/minibatch.py
```diff
@@ -62,12 +64,23 @@ class DGLMiniBatch:
...
     """
 
-    labels: Union[torch.Tensor, Dict[str, torch.Tensor]] = None
-    """Labels associated with seed nodes / node pairs in the graph.
-    - If `labels` is a tensor: It indicates the graph is homogeneous. The value
-      are corresponding labels to given 'output_nodes' or 'node_pairs'.
-    - If `labels` is a dictionary: The keys are node or edge type and the value
-      should be corresponding labels to given 'output_nodes' or 'node_pairs'.
+    @abstractmethod
+    def get_num_layers(self):
```
Rhett-Ying, line 68: `num_layers()`?

frozenbugs:
> @frozenbugs , @peizhou001 , I added a abstract class for `MiniBatch` and `DGLMiniBatch`. But I still need your help for several issues when fetching features.
> 
> 1. Seems that `DGLBlock` does not have `original_edge_ids`. How can I retrieve the edge ids when fetching edge features for `DGLBlock`?
> 2. I moved as many attributes of `MiniBatch` and `DGLMiniBatch` to `MiniBatchBase` as I can. Please check whether there are missing attributes that are shared by the two classes.
> 3. To align with `DGLMiniBatch`, I renamed the `seed_nodes` in `MiniBatch` to `output_nodes`. I may need to refactor all the usages of `seed_nodes` in`MiniBatch` later. Please check whether this is reasonable.

- The edge should be in dglblock.edata[dgl.EID], here is the [code](https://github.com/dmlc/dgl/blob/master/python/dgl/graphbolt/minibatch.py#L297C40-L297C40) that assigning the value. 
- Please don't rename the variable in MiniBatch to align with DGLMiniBatch, we intentionally kept them different, since the downstream of MiniBatch could be PyG or other library as well

python/dgl/graphbolt/minibatch.py
peizhou001: how about a property instead of method?

python/dgl/graphbolt/minibatch.py
```diff
@@ -62,12 +64,23 @@ class DGLMiniBatch:
...
+        pass
+
+    @abstractmethod
+    def get_sampled_subgraphs(self):
```
peizhou001, line 72: Ditto?

czkkkkkk:
> Please don't rename the variable in MiniBatch to align with DGLMiniBatch, we intentionally kept them different, since the downstream of MiniBatch could be PyG or other library as well

I see. Do you think it is a better to keep `MiniBatchBase` as an interface class without any data, such as `input_nodes`, `node_features`, etc?

czkkkkkk:
@frozenbugs, @Rhett-Ying, @peizhou001, I refactored the necessary interfaces of `MinibatchBase` to functions instead of properties. This is because defining them as properties needs to rename all the attributes of `MiniBatch` and `DGLMinibatch` to avoid naming conflict. For example, `input_nodes` will become `_input_nodes`. Then all the usages of these attributes, including constructors, will need to include the underscores. Take converting `MiniBatch` to `DGLMiniBatch` as an example.
```
minibatch = DGLMiniBatch(
    _blocks=self._to_dgl_blocks(),
    _node_features=self._node_features,
    _edge_features=self._edge_features,
    _labels=self._labels,
)
```
Please leave your comments if you have a better solution. If we think this solution is fine, I will go ahead to improve the docstring and add unittests.

python/dgl/graphbolt/minibatch.py
```diff
@@ -15,6 +16,57 @@
...
+        """A representation of input nodes in the outermost layer. Conatins all
+        nodes in the 'sampled_subgraphs'.
+        - If `input_nodes` is a tensor: It indicates the graph is homogeneous.
+        - If `input_nodes` is a dictionary: The keys should be node type and the
+          value should be corresponding heterogeneous node id.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_num_layers(self) -> int:
```
frozenbugs, line 33: ditto, omit get.

frozenbugs (approved):

PR #6545, opened 2023-11-08
[Example] Add early stop option for better regression test.

keli-wen:
## Description

Given that the entire training set in `ogbl-citation2` is excessively large, requiring substantial computational resources, we have introduced an `early-stop` option in the new GraphBolt link prediction example to manage and reduce the time cost during the training stage.

Currently, I need to implement a regression test. To maintain consistency, I believe it is crucial to include this option in the DGL version example as well.

## Checklist
- [x] Code is well-documented

### DIFF
--- a/examples/sampling/link_prediction.py
+++ b/examples/sampling/link_prediction.py
@@ -370,6 +370,8 @@ def train(args, device, g, reverse_eids, seed_edges, model, use_uva):
             loss.backward()
             opt.step()
             total_loss += loss.item()
+            if (it + 1) == args.early_stop:
+                break
         print(f"Epoch {epoch:05d} | Loss {total_loss / (it + 1):.4f}")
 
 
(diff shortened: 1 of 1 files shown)


frozenbugs (approved):

examples/sampling/link_prediction.py
```diff
@@ -370,6 +370,8 @@ def train(args, device, g, reverse_eids, seed_edges, model, use_uva):
             loss.backward()
             opt.step()
             total_loss += loss.item()
+            if (it + 1) == args.early_stop:
```
frozenbugs, line 373: I would break it at L355.

keli-wen: The `args.early_stop` parameter specifies the expected number of steps for each training epoch. Therefore, I would like to place it in this position, **as is done in the original DGL link prediction example**.

Issue #6561, opened 2023-11-14
[CPU-sampling] Avoid access/init cuda instances at each sampling (child) process


chang-l:
In PyTorch dataloader (cpu sampling), worker processes will never initialize CUDA context, as CUDA runtime does not support `fork` start method (https://pytorch.org/docs/stable/notes/multiprocessing.html#cuda-in-multiprocessing). I think DGL dataloader should also follow this convention, if possible. 

However, here, when creating coo/csr matrix, it would actually call this `cudaPointerGetAttributes` CUDA runtime API inside `IsPinned` function. As a result, each worker would call this constructor and start init/access cuda instances.

https://github.com/dmlc/dgl/blob/fedaa36da50dd39bb97ef33d04dc5921b08e1eb2/include/dgl/aten/coo.h#L68-L71

It's not a bug and will not error out as such behavior is guarded here by clearing the cuda error msg (see below):
https://github.com/dmlc/dgl/blob/b35757a0068d7c5faa2e9954f65308a436420e58/src/runtime/cuda/cuda_device_api.cc#L295-L301

Nevertheless, I believe it would still be preferable to adhere to PyT's convention by removing the `IsPinned` function from the constructor of the coo/csr matrix.

I can come up with a PR for the fix later. cc. @nv-dlasalle @yaox12 @frozenbugs @TristonC

nv-dlasalle:
This has caused a lot of trouble in the past, so really glad you've caught this and will fix it, @chang-l. It will also save a lot of developer time as the resulting bugs from this issue take a while to track down.

PR #6568, opened 2023-11-15
[Misc.] Avoid calling `IsPinned` in the coo/csr constructor from every sampling process

chang-l:
## Description
This PR addresses https://github.com/dmlc/dgl/issues/6561.
It moves the `IsPinned` checking block out of the coo/csr constructor, to avoid initialize cuda instances at sampling(child) processes and align with PyTorch dataloader convention.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/include/dgl/aten/coo.h
+++ b/include/dgl/aten/coo.h
@@ -56,19 +56,16 @@ struct COOMatrix {
   /** @brief constructor */
   COOMatrix(
       int64_t nrows, int64_t ncols, IdArray rarr, IdArray carr,
-      IdArray darr = NullArray(), bool rsorted = false, bool csorted = false)
+      IdArray darr = NullArray(), bool rsorted = false, bool csorted = false,
+      bool pin_memory = false)
       : num_rows(nrows),
         num_cols(ncols),
         row(rarr),
         col(carr),
         data(darr),
         row_sorted(rsorted),
-        col_sorted(csorted) {
-    if (!IsEmpty()) {
-      is_pinned = (aten::IsNullArray(row) || row.IsPinned()) &&
-                  (aten::IsNullArray(col) || col.IsPinned()) &&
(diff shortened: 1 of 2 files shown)


include/dgl/aten/coo.h
```diff
@@ -134,6 +131,12 @@ struct COOMatrix {
            aten::IsNullArray(data);
   }
 
+  inline bool IsPinned() const {
```
nv-dlasalle, line 134: Since we set internal member `is_pinned` in the constructor, it seems odd to me this member function doesn't use that. If this is only used internally for a check, can we make it private and change the name to make clear it will initialize a cuda context? e.g. `CheckIfPinnedInCUDA()`

include/dgl/aten/csr.h
```diff
@@ -128,6 +125,12 @@ struct CSRMatrix {
            aten::IsNullArray(data);
   }
 
+  inline bool IsPinned() const {
```
nv-dlasalle, line 128: Same here.

nv-dlasalle (approved):
I think in general it looks good, but as I noted I think the `IsPinned()` function should be changed.

include/dgl/aten/coo.h
chang-l: Good point.
I changed the function name to `CheckIfPinnedInCUDA`. So, there are two ways to check if the coo/csr is under pinned memory:
- The 'soft' way: `is_pinned` flag. This method does not need to init. cuda context and flexible. It can be updated by the 'hard' method or by developer when memory is pinned by DGL (not pytorch).
- The 'hard/strict' way: check via CUDA runtime API. This is a strict way but involves CUDA context. I use it to update the flag, only if necessary (when the pinned memory pool is not controlled by DGL and we need this sanity check on DGL side).

frozenbugs (approved):

PR #6576, opened 2023-11-16
[Graphbolt] Support feature metadata.

frozenbugs:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/feature_store.py
+++ b/python/dgl/graphbolt/feature_store.py
@@ -52,6 +52,16 @@ class Feature:
         """
         raise NotImplementedError
 
+    def metadata(self):
+        """Get the metadata of the feature.
+
+        Returns
+        -------
+        Dict
+            The metadata of the feature.
+        """
+        return {}
+
 
 class FeatureStore:
     r"""A store to manage multiple features for access."""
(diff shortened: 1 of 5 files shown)


python/dgl/graphbolt/feature_store.py
```diff
@@ -52,6 +52,16 @@ def update(self, value: torch.Tensor, ids: torch.Tensor = None):
         """
         raise NotImplementedError
 
+    def metadata(self):
```
Rhett-Ying, line 55: existing `.size()` is one of the metadata? Or a special metadata that lies standalone?

python/dgl/graphbolt/feature_store.py
```diff
@@ -110,6 +120,29 @@ def size(
         """
         raise NotImplementedError
 
+    def metadata(
```
Rhett-Ying, line 123: do we need to think more about the `metadata`? In this implementation, metadata is always per triplet of `domain/type_name/feature_name`. do need some metadata that's per `domain`, or even per `feature_store`?

python/dgl/graphbolt/feature_store.py
frozenbugs: Size is actually inferred from the tensor.

python/dgl/graphbolt/feature_store.py
frozenbugs: Good point, maybe in the future the API can be extended to making domain / type / feature optional.

Rhett-Ying (approved):

Issue #6577, opened 2023-11-16
Get no error while using `list` object in `bg.set_batch_num_nodes()` but will get an AttributeError when using AvgPool to pool the batched graph


NormanZyq:
## 🐛 Bug

If use a list object as the argument of set_batch_num_nodes(), we get no error, and it seems the graph is successfully batched with the number of nodes. However, if I use AvgPooling to pool the batched graph, I would get an error:

```
AttributeError: 'list' object has no attribute 'dtype'
```
It's obvious that the list is not supported in this case. 
If I use tensor, everything works fine, as the doc mentioned.

## To Reproduce

Please refer to the notebook:
https://colab.research.google.com/drive/1n-tfuPp3R59olfxgN4I9NJ95bNUmdsdJ?usp=sharing

## Expected behavior

If I use a list, there should be an error raised, or it can be transformed into tensor automatically

## Environment

 - DGL Version (e.g., 1.0): 1.1.2
 - Backend Library & Version (e.g., PyTorch 0.4.1, MXNet/Gluon 1.3): 2.1.0+cu118
 - OS (e.g., Linux): Linux
 - How you installed DGL (`conda`, `pip`, source): pip
 - Build command you used (if compiling from source):
 - Python version: 3.10.12
 - CUDA/cuDNN version (if applicable):
 - GPU models and configuration (e.g. V100): 
 - Any other relevant information:

Rhett-Ying:
Thanks for reporting this. We do need to avoid passing unexpected data.

PR #6578, opened 2023-11-17
[BugFix] return batch related ids in g.idtype

Rhett-Ying:
## Description

Fix for https://github.com/dmlc/dgl/issues/6577. What's more, I use `g.idtype` instead of `F.int64` for return values of `batch_num_nodes/edges()`.
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/heterograph.py
+++ b/python/dgl/heterograph.py
@@ -1511,7 +1511,7 @@ class DGLGraph(object):
             self._batch_num_nodes = {}
             for ty in self.ntypes:
                 bnn = F.copy_to(
-                    F.tensor([self.num_nodes(ty)], F.int64), self.device
+                    F.tensor([self.num_nodes(ty)], self.idtype), self.device
                 )
                 self._batch_num_nodes[ty] = bnn
         if ntype is None:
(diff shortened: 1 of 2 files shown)


python/dgl/heterograph.py
```diff
@@ -890,7 +890,7 @@ def remove_nodes(self, nids, ntype=None, store_ids=False):
             self._batch_num_nodes[
                 target_ntype
             ] = c_ntype_batch_num_nodes - F.astype(
-                batch_num_removed_nodes, F.int64
+                batch_num_removed_nodes, self.idtype
```
frozenbugs, line 893: Just wondering L884 hard-coded dtype (F.float32) is used, how do we decide when to use hard-coded, when we allow user specifying self.idtype?

frozenbugs (approved):

python/dgl/heterograph.py
Rhett-Ying: I think the underlying call `segment_reduce()` operates on float data, so it's specified as `F.float32` although `F.float64` works as well.

PR #6586, opened 2023-11-20
[CUDA] Fix issue about integer overflow

### LINKED ISSUE #6585: is there has max edge count limit?
DGLError: [23:02:45] /opt/dgl/src/array/cuda/./utils.h:53: Check failed: (range - 1) >> bits == 0 (1181544903 vs. 0) :

mfbalin:
Fixes #6585.

### DIFF
--- a/src/array/cuda/utils.h
+++ b/src/array/cuda/utils.h
@@ -46,7 +46,7 @@ int _NumberOfBits(const T& range) {
   }
 
   int bits = 1;
-  while (bits < static_cast<int>(sizeof(T) * 8) && (1 << bits) < range) {
+  while (bits < static_cast<int>(sizeof(T) * 8) && (1ll << bits) < range) {
     ++bits;
   }
 
--- a/src/array/cuda/utils.h
+++ b/src/array/cuda/utils.h
@@ -50,7 +50,9 @@ int _NumberOfBits(const T& range) {
     ++bits;
   }
 
-  CHECK_EQ((range - 1) >> bits, 0);
+  if (bits < static_cast<int>(sizeof(T) * 8)) {
+    CHECK_EQ((range - 1) >> bits, 0);
+  }
   CHECK_NE((range - 1) >> (bits - 1), 0);
 
   return bits;
--- a/src/array/cuda/utils.h
+++ b/src/array/cuda/utils.h
@@ -46,7 +46,7 @@ int _NumberOfBits(const T& range) {
   }
 
   int bits = 1;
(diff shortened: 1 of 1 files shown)


mfbalin:
@Rhett-Ying could you review this quick fix?

Rhett-Ying (commented):
could you add a testcase for this issue?

Rhett-Ying:
> could you add a testcase for this issue?

It's probably not easy to add testcase for this issue in python code path as it needs to create a large graph.

src/array/cuda/utils.h
```diff
@@ -46,7 +46,7 @@ int _NumberOfBits(const T& range) {
   }
 
   int bits = 1;
-  while (bits < static_cast<int>(sizeof(T) * 8) && (1 << bits) < range) {
+  while (bits < static_cast<int>(sizeof(T) * 8) && (1ll << bits) < range) {
```
Rhett-Ying, line 49: what if `int64_t range = std::numeric_limits<int64_t>::max()`? and what if `T is unsigned`?

mfbalin: range is of type T, so if T is unsigned, range can't be int64_t max.

Rhett-Ying: image

mfbalin: Did you test with `range = std::numeric_limits<uint64_t>::max()`?

mfbalin: I think the problem is that when we try to shift 64 times for the unsigned case, it is like no shifting. Thus, an if check is required to see if the bits == maximum bit width before the first assertion.

mfbalin: See here: https://godbolt.org/z/T4fTEsc5K
Pushing with this fix. Thanks for the catch!

mfbalin: In C/C++ [bitwise shifting](https://en.wikipedia.org/wiki/Logical_shift) a value by a number of bits which is either a negative number or is greater than or equal to the total number of bits in this value results in undefined behavior.
Source: https://en.wikipedia.org/wiki/Undefined_behavior#:~:text=In%20C%2FC%2B%2B%20bitwise%20shifting,value%20results%20in%20undefined%20behavior.

Rhett-Ying (approved):

Rhett-Ying:
@mfbalin could you help resolve below warning which is the side effect of this PR?
```
utils.h:49:73: warning: comparison of integer expressions of different signedness: ‘long long unsigned int’ and ‘const long int’ [-Wsign-compare]
   49 |   while (bits < static_cast<int>(sizeof(T) * 8) && (1ull << bits) < range) {
```

PR #6588, opened 2023-11-21
[doc] update cpu_best_practices

Rhett-Ying:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/tutorials/cpu/cpu_best_practises.py
+++ b/tutorials/cpu/cpu_best_practises.py
@@ -30,8 +30,7 @@ OpenMP settings
 ---------------------------
 
 As `OpenMP` is the default parallel backend, we could control performance
-including sampling and training via environment variable `OMP_NUM_THREADS` or
-`torch.set_num_threads() <https://pytorch.org/docs/stable/generated/torch.set_num_threads.html>`__.
+including sampling and training via `dgl.utils.set_num_threads()`.
 
 If number of OpenMP threads is not set and `num_workers` in dataloader is set
 to 0, the OpenMP runtime typically use the number of available CPU cores by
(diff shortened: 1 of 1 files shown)


tutorials/cpu/cpu_best_practises.py
```diff
@@ -29,17 +29,31 @@
...
-Best performance of parallelization in OpenMP
-can be achieved by setting up the optimal number of working threads and dataloading workers.
-Nodes with high number of CPU cores may benefit from higher number of dataloading workers.
-A good starting point could be setting num_threads=4 in Dataloader constructor for nodes with 32 cores or more.
-If number of cores is rather small, the best performance might be achieved with just one
-dataloader worker or even with dataloader num_threads=0 for dataloading and trainig performed
-in the same process
+As `OpenMP` is the default parallel backend, we could control performance
+including sampling and training via environment variable `OMP_NUM_THREADS` or
+`torch.set_num_threads() <https://pytorch.org/docs/stable/generated/torch.set_num_threads.html>`__.
```
anko-intel, line 34: Why not https://github.com/dmlc/dgl/blob/master/python/dgl/utils/internal.py#L989 ?
Maybe remove internal set_num_threads if it's not needed any more.

Rhett-Ying: The one you mentioned is another way to control omp threads. The reason why I mention `torch::set_num_threads` is that `torch::parallel_for()` is widely used in GraphBolt.
Indeed, we have too many ways to configure num_threads for now and it could be confusing.

Rhett-Ying: In order to avoid confusion, I prefer to mention `dgl.utils.set_num_threads()` only. And after we fully deprecate other backend(mxnet, tf), we could unify the mechanics to control thread pool.

frozenbugs (approved):

PR #6589, opened 2023-11-21
[DO NOT MERGE][Graphbolt] Profiling the throughput of UVA index select

czkkkkkk:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/src/cuda/index_select_impl.cu
+++ b/graphbolt/src/cuda/index_select_impl.cu
@@ -6,6 +6,7 @@
 #include <c10/cuda/CUDAException.h>
 #include <torch/script.h>
 
+#include <cub/cub.cuh>
 #include <numeric>
 
 #include "../index_select.h"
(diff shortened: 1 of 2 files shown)


test_gb_index_select.py
```diff
@@ -0,0 +1,192 @@
...
+    for _ in range(3):
+        for index in indices:
+            torch.ops.graphbolt.index_select(feature, index)
+    torch.cuda.synchronize()
+    n_iter = 5
+    start = time.time()
+    for _ in range(n_iter):
+        for index in indices:
+            torch.ops.graphbolt.index_select(feature, index)
+    torch.cuda.synchronize()
```
mfbalin, line 72: The use of cuda events here would give more reliable results, especially for smaller test cases. But for larger ones, it should be good enough.

graphbolt/src/cuda/index_select_impl.cu
```diff
@@ -14,6 +15,68 @@
...
+
+template <typename T>
+int NumberOfBits(const T& range) {
+  if (range <= 1) {
+    // ranges of 0 or 1 require no bits to store
+    return 0;
+  }
+
+  int bits = 1;
+  while (bits < static_cast<int>(sizeof(T) * 8) && (1 << bits) < range) {
```
mfbalin, line 40: There was a minor bug which I fixed in #6586. It might be relevant when testing 3.5B int8 tensor test case.

mfbalin: But since we are testing up to 4B, the current code may still work though.

mfbalin:
This benchmark has already been incorporated into the regression test.

PR #6601, opened 2023-11-22
[GraphBolt] replace item<> with raw pointer

Rhett-Ying:
## Description

wrapped with extra `AT_DISPATCH_INTEGRAL_TYPES`.
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/src/fused_csc_sampling_graph.cc
+++ b/graphbolt/src/fused_csc_sampling_graph.cc
@@ -330,96 +330,112 @@ FusedCSCSamplingGraph::SampleNeighborsImpl(
   torch::optional<torch::Tensor> subgraph_type_per_edge = torch::nullopt;
 
   AT_DISPATCH_INTEGRAL_TYPES(
-      indptr_.scalar_type(), "SampleNeighborsImpl", ([&] {
-        const scalar_t* indptr_data = indptr_.data_ptr<scalar_t>();
-        auto num_picked_neighbors_data_ptr =
-            num_picked_neighbors_per_node.data_ptr<scalar_t>();
-        num_picked_neighbors_data_ptr[0] = 0;
+      indptr_.scalar_type(), "SampleNeighborsImplWrappedWithIndptr", ([&] {
+        using indptr_t = scalar_t;
+        AT_DISPATCH_INTEGRAL_TYPES(
+            nodes.scalar_type(), "SampleNeighborsImplWrappedWithNodes", ([&] {
+              using nodes_t = scalar_t;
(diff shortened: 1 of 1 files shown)


mfbalin (approved):
LGTM.

graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -330,96 +330,112 @@ FusedCSCSamplingGraph::SampleNeighborsImpl(
   torch::optional<torch::Tensor> subgraph_type_per_edge = torch::nullopt;
 
   AT_DISPATCH_INTEGRAL_TYPES(
-      indptr_.scalar_type(), "SampleNeighborsImpl", ([&] {
-        const scalar_t* indptr_data = indptr_.data_ptr<scalar_t>();
-        auto num_picked_neighbors_data_ptr =
-            num_picked_neighbors_per_node.data_ptr<scalar_t>();
-        num_picked_neighbors_data_ptr[0] = 0;
+      indptr_.scalar_type(), "SampleNeighborsImplWrappedWithIndptr", ([&] {
```
peizhou001, line 333: Don't change name?

graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -330,96 +330,112 @@ FusedCSCSamplingGraph::SampleNeighborsImpl(
...
+        using indptr_t = scalar_t;
```
peizhou001, line 334: why need this

graphbolt/src/fused_csc_sampling_graph.cc
Rhett-Ying: just to be consistent with following wrapper with `nodes`. Any concern about this name? any other dependency on this name beside its uniqueness requirement?

peizhou001:
The change makes sense, but one thing needs discussion is do we need to carefully deal with the type nodes? or just do a explicit conversion in python side at tensor level, or just do implicit conversion in C side at variable level?

Rhett-Ying:
> The change makes sense, but one thing needs discussion is do we need to carefully deal with the type nodes? or just do a explicit conversion in python side at tensor level, or implicit conversion in C side at variable level?

The main purpose I added an extra dispatch macro is to avoid conversion. If conversion is applied always that makes sure `nodes.dtype` equals `indices.dtype`, that is the safest. Or user is responsible for indexing on graph with `nodes` IDs is dtype-safe.

frozenbugs (approved):

PR #6605, opened 2023-11-23
[Misc] Align dgl multi-gpu example to graphbolt

RamonZhou:
## Description
Added a `benchmark` mode disabling uva, gpu sampling and fused sampling.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/examples/multigpu/node_classification_sage.py
+++ b/examples/multigpu/node_classification_sage.py
@@ -123,12 +123,13 @@ class SAGE(nn.Module):
         return y
 
 
-def evaluate(model, g, num_classes, dataloader):
+def evaluate(device, model, g, num_classes, dataloader):
     model.eval()
     ys = []
     y_hats = []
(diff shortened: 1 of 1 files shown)


examples/multigpu/node_classification_sage.py
```diff
@@ -145,6 +146,8 @@ def layerwise_infer(
 ):
     model.eval()
     with torch.no_grad():
+        if not use_uva:
+            g = g.to(device)
```
peizhou001, line 150: It means in our benchmark, entire graph wil be put into GPU?

RamonZhou: Yes. It's because dgl uses inference to do the final testing so this part won't be considered in the comparison. We can optimize this part by putting the graph into GPU. It's also the original way.

peizhou001 (approved):

PR #6613, opened 2023-11-24
[Graphbolt] Support set seed for graphbolt.

frozenbugs:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/src/random.cc
+++ b/graphbolt/src/random.cc
@@ -1,0 +1,22 @@
+/**
+ *  Copyright (c) 2023 by Contributors
+ * @file random.cc
+ * @brief Random Engine.
+ */
+
+#include "./random.h"
+
+#include <torch/torch.h>
+
+namespace graphbolt {
+
+void SetSeed(int64_t seed) {
+  int64_t num_threads = torch::get_num_threads();
+  torch::parallel_for(0, num_threads, 1, [&](size_t start, size_t end) {
+    for (auto i = start; i < end; ++i) {
+      RandomEngine::ThreadLocal()->SetSeed(seed);
+    }
+  });
+}
+
+}  // namespace graphbolt
--- a/graphbolt/src/python_binding.cc
+++ b/graphbolt/src/python_binding.cc
@@ -10,6 +10,7 @@
 #include <graphbolt/unique_and_compact.h>
 
 #include "./index_select.h"
+#include "./random.h"
 
 namespace graphbolt {
 namespace sampling {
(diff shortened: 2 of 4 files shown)


graphbolt/src/random.cc
```diff
@@ -0,0 +1,61 @@
...
+    std::lock_guard<std::mutex> guard(mutex);
+    id = num_threads;
+    num_threads++;
+  }
+  return id;
+}
+
+};  // namespace
+
+std::optional<uint64_t> RandomEngine::manual_seed;
```
peizhou001, line 31: Is this repeat with static variable in header file?

graphbolt/src/random.cc
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/** @brief Set the seed. */
+void RandomEngine::SetSeed(uint64_t seed) { SetSeed(seed, GetThreadId()); }
+
+void RandomEngine::SetSeed(uint64_t seed, uint64_t stream) {
+  rng_.seed(seed, stream);
+}
+
+void SetSeed(int64_t seed) { RandomEngine::manual_seed = seed; }
```
peizhou001, line 59: int64_t -> uint64_t?

peizhou001, line 59: Can we enable null here so that user can recover the randomness of seed?

graphbolt/src/random.cc
frozenbugs: This is initialization of static variable which declared in the header file.

graphbolt/src/random.cc
frozenbugs: [Only int64_t and bool are supported as an integral argument type when building a torchscript extension](https://discuss.pytorch.org/t/invalid-type-only-int64-t-and-bool-are-supported-as-an-integral-argument-type-when-building-a-torchscript-extension/95342)

frozenbugs: We can add it later if people really need it.

graphbolt/src/random.cc
```diff
@@ -0,0 +1,64 @@
...
+/** @brief Set the seed. */
+void RandomEngine::SetSeed(uint64_t seed) { SetSeed(seed, GetThreadId()); }
+
+/** @brief Set the seed. */
+void RandomEngine::SetSeed(uint64_t seed, uint64_t stream) {
+  rng_.seed(seed, stream);
+}
+
+/** @brief Manually fix the seed. */
+void RandomEngine::SetManualSeed(int64_t seed) { manual_seed = seed; }
```
peizhou001, line 62: It's more complete to support null here so that user can clear the manual seed to recover randomness.

peizhou001 (approved):

frozenbugs:
@dgl-bot

PR #6629, opened 2023-11-27
[DoNotMerge][C++Seminar] define all in header file

Rhett-Ying:
## Description

What's the problem you hit if this PR is applied? @frozenbugs 
## Checklist
- [ ] Code is well-documented


graphbolt/src/random.h
```diff
@@ -17,28 +17,54 @@
...
   /** @brief Set the seed. */
-  void SetSeed(uint64_t seed);
-  void SetSeed(uint64_t seed, uint64_t stream);
+  void SetSeed(uint64_t seed) { SetSeed(seed, GetThreadId()); }
+  void SetSeed(uint64_t seed, uint64_t stream) { rng_.seed(seed, stream); }
 
   /** @brief Manually fix the seed. */
-  static std::optional<uint64_t> manual_seed;
-  static void SetManualSeed(int64_t seed);
+  static inline std::optional<uint64_t> manual_seed;
```
frozenbugs, line 66: I didn't use inline previously, does this change the scope of the variable?

Rhett-Ying: it's for allowing definition within class instead of outside. As for scope, I think every translation unit has the identical instance of `::manual_seed` no matter definitions are in header file or cc file.

PR #6632, opened 2023-11-27
[Graphbolt][CUDA] Optimize UVA index select

mfbalin:
Here, we make index select independent of the input type and use the largest possible type to perform the copies since for example copying 2 uint32_t equals copying 1 uint64_t. Should help a lot especially for smaller types such as int8, float16, etc. to reduce resource requirements and also potentially improve performance. It also allows supporting all native torch dtypes while using minimum compilation time by sharing the code path for dtypes with the same alignment.

### DIFF
--- a/graphbolt/src/cuda/index_select_impl.cu
+++ b/graphbolt/src/cuda/index_select_impl.cu
@@ -99,12 +99,13 @@ __global__ void IndexSelectMultiKernelAligned(
 }
 
 template <typename DType, typename IdType>
-torch::Tensor UVAIndexSelectImpl_(
-    torch::Tensor input, torch::Tensor index, int64_t feature_size) {
+torch::Tensor UVAIndexSelectImpl_(torch::Tensor input, torch::Tensor index) {
   const int64_t input_len = input.size(0);
   const int64_t return_len = index.size(0);
   const int64_t original_feature_size = std::accumulate(
-      input.sizes().begin() + 1, input.sizes().end(), 1, std::multiplies<>());
+      input.sizes().begin() + 1, input.sizes().end(), 1ll, std::multiplies<>());
+  const auto aligned_feature_size =
+      input.element_size() * original_feature_size / sizeof(DType);
   torch::Tensor ret = torch::empty(
       {return_len, original_feature_size}, torch::TensorOptions()
                                                .dtype(input.dtype())
@@ -120,7 +121,7 @@ torch::Tensor UVAIndexSelectImpl_(
 
   cudaStream_t stream = torch::cuda::getDefaultCUDAStream();
 
-  if (feature_size == 1) {
+  if (aligned_feature_size == 1) {
     // Use a single thread to process each output row to avoid wasting threads.
     const int num_threads = cuda::FindNumThreads(return_len);
     const int num_blocks = (return_len + num_threads - 1) / num_threads;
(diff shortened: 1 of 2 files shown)


graphbolt/src/cuda/index_select_impl.cu
```diff
@@ -160,14 +164,37 @@ torch::Tensor UVAIndexSelectImpl_(torch::Tensor input, torch::Tensor index) {
...
-              return UVAIndexSelectImpl_<scalar_t, index_t>(input, index);
-            });
-      });
+  return AT_DISPATCH_INDEX_TYPES(
+      index.scalar_type(), "UVAIndexSelectImpl", ([&] {
+        const auto ptr = (size_t)input.data_ptr();
+        const int64_t feature_size = std::accumulate(
+            input.sizes().begin() + 1, input.sizes().end(), 1ll,
+            std::multiplies<>());
+        const int alignment =
```
frozenbugs, line 173: Add a comment about what's alignment and what's the magic number 16?

graphbolt/src/cuda/index_select_impl.cu
```diff
@@ -160,14 +164,37 @@ torch::Tensor UVAIndexSelectImpl_(torch::Tensor input, torch::Tensor index) {
  * The supporting index types are: int, int64_t.
```
frozenbugs, line 164: Do we need to update the supporting types?

graphbolt/src/cuda/utils.h
```diff
@@ -31,6 +31,22 @@ inline int FindNumThreads(int size) {
   return ret;
 }
 
+template <typename T>
+int _NumberOfBits(const T& range) {
```
frozenbugs, line 35: This is not used in this PR, maybe add it when we need it?

graphbolt/src/cuda/common.h
```diff
@@ -0,0 +1,54 @@
...
+inline bool is_zero(T size) {
+  return size == 0;
+}
+
+template <>
+inline bool is_zero<dim3>(dim3 size) {
+  return size.x == 0 || size.y == 0 || size.z == 0;
+}
+
+#define CUDA_DRIVER_CALL(x)                                             \
```
frozenbugs, line 24: This is not used in this PR, maybe add it when we need it?

graphbolt/src/cuda/common.h
```diff
@@ -0,0 +1,54 @@
...
+  {                                                                     \
+    CUresult result = x;                                                \
+    if (result != CUDA_SUCCESS && result != CUDA_ERROR_DEINITIALIZED) { \
+      const char* msg;                                                  \
+      cuGetErrorName(result, &msg);                                     \
+      LOG(FATAL) << "CUDAError: " #x " failed with error: " << msg;     \
+    }                                                                   \
+  }
+
+#define CUDA_CALL(func)                                      \
```
frozenbugs, line 34: ditto.

graphbolt/src/cuda/index_select_impl.cu
mfbalin: However, index types are int32_t and int64_t still.

graphbolt/src/cuda/index_select_impl.cu
```diff
@@ -156,18 +160,47 @@ torch::Tensor UVAIndexSelectImpl_(torch::Tensor input, torch::Tensor index) {
...
+            std::multiplies<>());
+        // We perform the copy with datatype of size powers of 2, and the
+        // maximum data type we use has 16 bytes. We check the alignment of the
+        // pointer and the feature dimensionality to determine the largest
+        // type to use for the copy to minimize the number of CUDA threads used.
+        // Alignment denotes the maximum suitable alignment and datatype size
+        // for the copies.
+        const int alignment =
+            std::gcd(16, std::gcd(ptr, input.element_size() * feature_size));
+        const auto new_feature_size =
```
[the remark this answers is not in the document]
czkkkkkk: I think it is better to define `aligned_feature_size` in `UVAIndexSelectImpl_`. The interface will be cleaner and it is easier to understand what are `aligned_feature_size` and `original_feature_size`.

graphbolt/src/cuda/index_select_impl.cu
```diff
@@ -156,18 +160,47 @@ torch::Tensor UVAIndexSelectImpl_(torch::Tensor input, torch::Tensor index) {
...
+            input.element_size() * feature_size / alignment;
+        switch (alignment) {
```
czkkkkkk, line 183: I am not sure whether we need to support all these data types. If so, we need to add unit tests for them. @frozenbugs, what do you think?

czkkkkkk (changes requested):

graphbolt/src/cuda/index_select_impl.cu
```diff
@@ -156,18 +160,47 @@ torch::Tensor UVAIndexSelectImpl_(torch::Tensor input, torch::Tensor index) {
...
+            return UVAIndexSelectImpl_<uint16_t, index_t>(
+                input, index, new_feature_size);
+          case 4:
+            return UVAIndexSelectImpl_<uint32_t, index_t>(
+                input, index, new_feature_size);
+          case 8:
+            return UVAIndexSelectImpl_<uint64_t, index_t>(
+                input, index, new_feature_size);
+          case 16:
+            return UVAIndexSelectImpl_<float4, index_t>(
```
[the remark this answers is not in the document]
mfbalin: Float4 is a cuda native datatype made of 4 floats packed together. The important thing is that it occupies 16 bytes.

graphbolt/src/cuda/index_select_impl.cu
frozenbugs: Yes, it is better to start from what we really need, if we foresee those are important for our future work, we can add it and unit test is required in this case.

mfbalin: Let me add the unit tests for all different types of sizes 1 2 4 8 and 16 from torch. I have the code already so easy to extend the unit test with more datatypes.

graphbolt/src/cuda/index_select_impl.cu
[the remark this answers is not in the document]
mfbalin: Yeah, just changed it.

graphbolt/src/cuda/index_select_impl.cu
mfbalin: I added the unit tests for all possible types of different sizes.

czkkkkkk (approved):
LGTM

PR #6689, opened 2023-12-05
[GraphBolt] Add option to select node_pairs or csc_format.

yxy235:
## Description

Add option to select node_pairs or csc_format.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
+++ b/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
@@ -577,9 +577,7 @@ class FusedCSCSamplingGraph(SamplingGraph):
         C_sampled_subgraph = self._sample_neighbors(
             nodes, fanouts, replace, probs_name
         )
-        if deduplicate is True:
-            if is_cscformat:
-                raise RuntimeError("Not implemented yet.")
+        if deduplicate and not is_cscformat:
             return self._convert_to_fused_sampled_subgraph(C_sampled_subgraph)
         else:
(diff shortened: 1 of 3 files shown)


python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -506,6 +506,7 @@ def sample_neighbors(
         replace: bool = False,
         probs_name: Optional[str] = None,
         deduplicate=True,
+        is_cscformat=False,
```
frozenbugs, line 509:
```suggestion
        # TODO: clean up once the migration is done.
        output_cscformat=False,
```

frozenbugs (approved):

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -505,8 +505,9 @@ def sample_neighbors(
         fanouts: torch.Tensor,
         replace: bool = False,
         probs_name: Optional[str] = None,
+        # TODO: clean up once the migration is done.
```
frozenbugs, line 508: Move this above L510, I don't think we need to clean up `deduplicate`, right?

yxy235: > Move this above L510, I don't think we need to clean up `deduplicate`, right?

After migration, both unique and not_deduplicate use csc_foramt sampled subgraph, so we don't need `deduplicate` to select which function to use.

PR #6761, opened 2023-12-15
[GraphBolt] de-duplicate code for reading data from achive

Rhett-Ying:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/include/graphbolt/serialize.h
+++ b/graphbolt/include/graphbolt/serialize.h
@@ -57,47 +57,18 @@ inline serialize::OutputArchive& operator<<(
 namespace graphbolt {
 
 /**
- * @brief Read data from archive.
+ * @brief Read data from archive and format to specified type.
  * @param archive Input archive.
  * @param key Key name of data.
  *
  * @return data.
  */
-torch::IValue read_from_archive(
-    torch::serialize::InputArchive& archive, const std::string& key);
-
-/**
- * @brief Read dictionary from archive.
- * @param archive Input archive.
- * @param key Key name of dictionary.
- *
- * @return dictionary.
- */
-template <typename DictT>
-DictT read_dict_from_archive(
+template <typename T>
+T read_from_archive(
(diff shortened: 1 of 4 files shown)


graphbolt/include/graphbolt/serialize.h
```diff
@@ -66,6 +66,40 @@ namespace graphbolt {
...
     torch::serialize::InputArchive& archive, const std::string& key);
 
+/**
+ * @brief Read dictionary from archive.
+ * @param archive Input archive.
+ * @param key Key name of dictionary.
+ *
+ * @return dictionary.
+ */
+template <typename DictT>
```
czkkkkkk, line 76: The PR makes a lot of sense to me. A minor suggestion is to make both key and value types as the templates for this function.
```cpp
template <typename KeyType, typename ValueType>
```
Then the function is general to all dictionary types and we don't need to specially handle `int64_t` and `torch::Tensor`.

czkkkkkk (approved):

graphbolt/include/graphbolt/serialize.h
Rhett-Ying: I wish I could achieve. The major blocker is the caller function name when converting IValue to real data such as `toInt()`, `toTensor()`. Is there any template-like calls for these ones? For example, `to<int>`, `to<Tensor>`?

czkkkkkk: I see. I am not sure whether [this](https://github.com/pytorch/pytorch/blob/4ea7430ffb0b37e84108c70f90f402f5fd37262e/aten/src/ATen/core/ivalue.h#L1004) is workable.

Rhett-Ying: this works. let me further refine related code.

Rhett-Ying: I used generalized `to<>` for further de-duplication

PR #6766, opened 2023-12-18
[GraphBolt] Fix MiniBatch blocks.

yxy235:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/minibatch.py
+++ b/python/dgl/graphbolt/minibatch.py
@@ -395,7 +395,11 @@ class MiniBatch:
                             (
                                 v.indptr,
                                 v.indices,
-                                torch.tensor([]),
+                                torch.arange(
+                                    0,
+                                    len(v.indices),
+                                    device=v.indptr.device,
+                                ),
                             ),
                         )
                         for etype, v in subgraph.node_pairs.items()
@@ -421,7 +425,11 @@ class MiniBatch:
                         (
                             node_pairs.indptr,
                             node_pairs.indices,
-                            torch.tensor([]),
+                            torch.arange(
+                                0,
+                                len(node_pairs.indices),
+                                device=node_pairs.indptr.device,
+                            ),
                         ),
                     )
                 num_src_nodes = original_row_node_ids.size(0)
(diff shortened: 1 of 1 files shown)


python/dgl/graphbolt/minibatch.py
```diff
@@ -395,7 +395,11 @@ def blocks(self):
                             (
                                 v.indptr,
                                 v.indices,
-                                torch.tensor([]),
+                                torch.arange(
+                                    0,
+                                    len(v.indices),
+                                    device=v.indptr.device,
```
Rhett-Ying, line 401: use length of `indices` while device is retrieved from `indptr`? though `indptr` and `indices` are supposed to be on same device?

yxy235: `indptr` and `indices` are supposed to be on the same device because they are all in the `CSCFormatBase`. Should I change it to `indices`？

Rhett-Ying: the 3rd argument is supposed to be edge ID? https://docs.dgl.ai/en/latest/generated/dgl.create_block.html#dgl.create_block

yxy235: Yes, I have tried to use `original_edge_ids` before, however it ran segmentation fault.

Rhett-Ying: to be consistent, it's better to keep them same. retrieve device type from `indices` too.

Rhett-Ying: could we replace `len(v.indices)` with `v.indptr[-1]` for better performance?

yxy235: > to be consistent, it's better to keep them same. retrieve device type from `indices` too.

Due to use indptr[-1], I still retrieve device type from `indptr`.

python/dgl/graphbolt/minibatch.py
```diff
@@ -421,7 +425,11 @@ def blocks(self):
                         (
                             node_pairs.indptr,
                             node_pairs.indices,
-                            torch.tensor([]),
+                            torch.arange(
+                                0,
+                                node_pairs.indptr[-1],
+                                device=node_pairs.indptr.device,
```
Rhett-Ying, line 431: how about specifying `dtype` as well?

Rhett-Ying (approved):

PR #6738, opened 2023-12-12
[GraphBolt][CUDA] Unique and Compact

mfbalin:
## Description
An implementation of a Unique and compact operator for CUDA. I will benchmark it against the version in DGL and see which one is more efficient. Then, we can determine the best approach and merge the fastest version.

## Benchmark results on A100 GPU

If you can run the benchmark script on other hardware, that would be helpful too.

```
* params:  n_rows=16777216, avg_degree=8, indptr_dtype=torch.int64, tensor_dtype=torch.int32, num_indices=1000, indptr_device=Device.Pinned, tensor_device=Device.Pinned, indices_device=Device.GPU, regular=Regularity.IRREGULAR

torch.ops.graphbolt.unique_and_compact --- Runtime in us: 506, Throughput in MiB/s: 120


gb.unique_and_compact_csc_formats --- Runtime in us: 548, Throughput in MiB/s: 111


dgl.to_block --- Runtime in us: 3114, Throughput in MiB/s: 19


* params:  n_rows=16777216, avg_degree=8, indptr_dtype=torch.int64, tensor_dtype=torch.int32, num_indices=100000, indptr_device=Device.Pinned, tensor_device=Device.Pinned, indices_device=Device.GPU, regular=Regularity.IRREGULAR

torch.ops.graphbolt.unique_and_compact --- Runtime in us: 764, Throughput in MiB/s: 7983


gb.unique_and_compact_csc_formats --- Runtime in us: 711, Throughput in MiB/s: 8579


dgl.to_block --- Runtime in us: 648, Throughput in MiB/s: 9402


* params:  n_rows=16777216, avg_degree=8, indptr_dtype=torch.int64, tensor_dtype=torch.int32, num_indices=1000000, indptr_device=Device.Pinned, tensor_device=Device.Pinned, indices_device=Device.GPU, regular=Regularity.IRREGULAR

torch.ops.graphbolt.unique_and_compact --- Runtime in us: 4750, Throughput in MiB/s: 12845


gb.unique_and_compact_csc_formats --- Runtime in us: 3785, Throughput in MiB/s: 16120


dgl.to_block --- Runtime in us: 5680, Throughput in MiB/s: 10743


* params:  n_rows=16777216, avg_degree=16, indptr_dtype=torch.int64, tensor_dtype=torch.int32, num_indices=1000, indptr_device=Device.Pinned, tensor_device=Device.Pinned, indices_device=Device.GPU, regular=Regularity.IRREGULAR

[...]

PR #6770, opened 2023-12-18
[GraphBolt][CUDA] SampleNeighbors (Without replacement for now)

mfbalin:
## Description
The CUDA equivalent of `FusedCSCSamplingGraph::SampleNeighbors`. Implements all the branches with a single implementation. Can be specialized later for potential optimization. My goal is to get end-to-end examples working as soon as possible. After this PR is merged, node and link prediction examples should just all work, I modified `lightning/node_prediction.py` temporarily to show that it works. Will remove changes to the example before merging and open a separate PR to enable GPU sampling on the GraphBolt examples.

```
mfbalin@BALIN-PC:~/dgl-1/examples/sampling/graphbolt/lightning$ time python node_classification.py 
The dataset is already preprocessed.
GPU available: True (cuda), used: True
TPU available: False, using: 0 TPU cores
IPU available: False, using: 0 IPUs
HPU available: False, using: 0 HPUs
You are using a CUDA device ('NVIDIA GeForce RTX 4090') that has Tensor Cores. To properly utilize them, you should set `torch.set_float32_matmul_precision('medium' | 'high')` which will trade-off precision for performance. For more details, read https://pytorch.org/docs/stable/generated/torch.set_float32_matmul_precision.html#torch.set_float32_matmul_precision
LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0]

  | Name      | Type               | Params
-------------------------------------------------
0 | layers    | ModuleList         | 206 K 
1 | dropout   | Dropout            | 0     
2 | train_acc | MulticlassAccuracy | 0     
3 | val_acc   | MulticlassAccuracy | 0     
-------------------------------------------------
206 K     Trainable params
0         Non-trainable params
206 K     Total params
0.828     Total estimated model params size (MB)
Sanity Checking: 0it [00:00, ?it/s]/home/mfbalin/.local/lib/python3.10/site-packages/pytorch_lightning/trainer/connectors/data_connector.py:432: PossibleUserWarning: The dataloader, val_dataloader, does not have many workers which may be a bottleneck. Consider increasing the value of the `num_workers` argument` (try 32 which is the number of cpus on this machine) in the `DataLoader` init to improve performance.
  rank_zero_warn(
/home/mfbalin/.local/lib/python3.10/site-packages/pytorch_lightning/trainer/connectors/data_connector.py:432: PossibleUserWarning: The dataloader, train_dataloader, does not have many workers which may be a bottleneck. Consider increasing the value of the `num_workers` argument` (try 32 which is the number of cpus on this machine) in the `DataLoader` init to improve performance.
  rank_zero_warn(
Epoch 14: : 192it [00:07, 26.47it/s, v_num=119, train_acc=0.911, num_nodes/0=5.12e+5, num_edges/0=9.36e+5, num_nodes/1=94886.0, num_edges/1=1.07e+5, num_nodes/2=10882.0, num_edges/2=1e+4, num_nodes/3=1024.0, val_acc=0.906]   
                                                  
real    1m56.086s
user    1m51.387s
sys     0m18.968s
```

Right now, ~~3x~~ 4x faster than CPU sampling on my 4090, however should become much faster when we are done optimizing. In the profile below on an A100, there are a lot of occasions where the GPU is simply idle. We should eliminate places where GPU & CPU synchronizes.

image


mfbalin:
I was able to run the training example with the GPU sampling successfully, everything seems to be working. In order for me to enable the tests, I need #6738 to move forward.

graphbolt/include/graphbolt/cuda_ops.h
```diff
@@ -97,6 +98,51 @@ torch::Tensor ExclusiveCumSum(torch::Tensor input);
...
+ * @param probs_or_mask An optional tensor with (unnormalized) probabilities
+ * corresponding to each neighboring edge of a node. It must be
+ * a 1D tensor, with the number of elements equaling the total number of edges.
+ * @param random_seed An optional random_seed parameter to fix the random_seed
+ * of the sampling operation. (Useful when layer=True)
+ *
+ * @return An intrusive pointer to a FusedSampledSubgraph object containing
+ * the sampled graph's information.
+ */
+c10::intrusive_ptr<sampling::FusedSampledSubgraph> SampleNeighbors(
```
frozenbugs, line 138: Can we make this an class method of fused_csc_sampling_graph?

graphbolt/include/graphbolt/cuda_ops.h
```diff
@@ -97,6 +98,51 @@ torch::Tensor ExclusiveCumSum(torch::Tensor input);
...
+    torch::Tensor indptr, torch::Tensor indices, torch::Tensor nodes,
+    const std::vector<int64_t>& fanouts, bool replace, bool layer,
+    bool return_eids,
+    torch::optional<torch::Tensor> type_per_edge = torch::nullopt,
+    torch::optional<torch::Tensor> probs_or_mask = torch::nullopt,
+    torch::optional<int64_t> random_seed = torch::nullopt);
```
frozenbugs, line 144: why do we need to pass in random_seed, is there anyway to set global seed for CUDA?

graphbolt/src/cuda/neighbor_sampler_impl.cu
```diff
@@ -0,0 +1,250 @@
...
+#include "./utils.h"
+
+namespace graphbolt {
+namespace ops {
+
+constexpr int BLOCK_SIZE = 128;
+
+template <
+    typename float_t, typename indptr_t, typename indices_t, typename weights_t>
+__global__ void _ComputeRowRandomPairs(
```
frozenbugs, line 31: It is good for readability to create util function, can you also add a comment about what does this util do? It is not obvious to me from the name of the method.

graphbolt/src/cuda/neighbor_sampler_impl.cu
```diff
@@ -0,0 +1,250 @@
...
+  const auto num_edges = coo_rows.size(0);
+  auto allocator = cuda::GetAllocator();
+  const auto stream = cuda::GetCurrentStream();
+  if (!random_seed.has_value()) {
+    random_seed = RandomEngine::ThreadLocal()->RandInt(
+        static_cast<int64_t>(0), std::numeric_limits<int64_t>::max());
+  }
+  torch::Tensor picked_eids;
+
+  AT_DISPATCH_INTEGRAL_TYPES(
```
frozenbugs, line 131: nit: can you polish the code a little bit according to the comments I made to the previous PR?
- Add blank line between section, comment each section
- Keep the comment neutral and objective, avoiding subjective pronouns like "we" and "you."
- Pick better name for the local variables, if the name is not contextual enough, add comment.

graphbolt/include/graphbolt/cuda_ops.h
mfbalin: It is already being dispatched inside `FusedCSCSamplingGraph::SampleNeighbors`. This way, keeping different implementations in different files is easier, which helps compile times greatly. For example, this implementation takes a long time to compile due to the three nested type dispatches. Especially because the dispatch for `probs_or_mask` is done for all possible types (bool. int8, uint8, int16, int32, int64, float16, bfloat16, float32, float64, for a total of 10 types times 4 times 4 = 40 type combinations).

graphbolt/include/graphbolt/cuda_ops.h
mfbalin: I can leave it out in this PR. Labor sampling has a layer-dependency option for which it is important to fix the seed of the sampler for all layers. This option is not yet ported from DGL so I will remove this parameter for now.

graphbolt/include/graphbolt/cuda_ops.h
mfbalin: Also, I might reuse this implementation for temporal sampling as well, in the FusedCSCSamplingGraph code, temporal sampling code path is separate.

mfbalin: So TemporalSampleNeighbors might call this function as well.

frozenbugs: My concern is as an cuda_op in cuda_ops.h, it is overwhelming for the code to know the data type FusedSampledSubgraph, which belongs to fused_csc_sampling_graph.

mfbalin: One solution is then to return a `std::tuple` and then form a fused_csc_sampling_graph inside the CPU source file.

mfbalin: Encapsulating the implementation into a class will overall reduce readability I am afraid.

mfbalin: Moreover, I have plans for later using this function when we have a graph cache.

mfbalin: Or I can have `cuda_ops.h` as usual but create another header file called `cuda_sampling_ops.h`. Only this file can include the definition of the FusedSamplingGraph.

mfbalin: Then cuda_ops.h does not have to import the contents of the FusedSampledSubgraph.

mfbalin: made the changes.

python/dgl/graphbolt/impl/neighbor_sampler.py
```diff
@@ -118,9 +118,15 @@ def sample_subgraphs(self, seeds):
         # Enrich seeds with all node types.
         if isinstance(seeds, dict):
             ntypes = list(self.graph.node_type_to_id.keys())
+            device = None
+            dtype = None
+            for _, seed in seeds.items():
```
frozenbugs, line 123: it is not intuitive to use loop here, add comment.

frozenbugs:
LGTM overall

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -444,7 +444,9 @@ def _convert_to_homogeneous_nodes(self, nodes, timestamps=None):
         homogeneous_timestamps = []
         for ntype, ids in nodes.items():
             ntype_id = self.node_type_to_id[ntype]
-            homogeneous_nodes.append(ids + self.node_type_offset[ntype_id])
+            homogeneous_nodes.append(
```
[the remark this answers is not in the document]
mfbalin: Yes, it was not working without this chance due to trying to add two tensors, one is on GPU, the other on CPU.

graphbolt/src/cuda/neighbor_sampler.cu
```diff
@@ -0,0 +1,319 @@
...
+
+c10::intrusive_ptr<sampling::FusedSampledSubgraph> SampleNeighbors(
+    torch::Tensor indptr, torch::Tensor indices, torch::Tensor nodes,
+    const std::vector<int64_t>& fanouts, bool replace, bool layer,
+    bool return_eids, torch::optional<torch::Tensor> type_per_edge,
+    torch::optional<torch::Tensor> probs_or_mask) {
+  TORCH_CHECK(
+      fanouts.size() == 1, "Heterogenous sampling is not supported yet!");
+  TORCH_CHECK(!replace, "Sampling with replacement is not supported yet!");
+  // Assume that indptr, indices, nodes, type_per_edge and probs_or_mask
```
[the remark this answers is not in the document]
mfbalin: The code still works if some of these tensors are pinned. Thus, an assert is not needed.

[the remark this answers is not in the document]
mfbalin: It is just that I didn't specialize the implementation in any way thinking some of these tensors might be pinned. Might want to revisit and optimize for the pinned cases.

examples/sampling/graphbolt/lightning/node_classification.py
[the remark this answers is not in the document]
mfbalin: Yes, I will right before merging.

graphbolt/src/cuda/neighbor_sampler.cu
```diff
@@ -0,0 +1,319 @@
...
+      constexpr uint64_t kCurandSeed = 999961;
+      curand_init(kCurandSeed, random_seed, indices[in_idx], &rng);
+    }
+
+    const auto rnd = curand_uniform(&rng);
+    const auto prob = weights ? weights[in_idx] : static_cast<weights_t>(1);
+    const auto exp_rnd = -__logf(rnd);
+    const float_t adjusted_rnd = prob > 0
+                                     ? static_cast<float_t>(exp_rnd / prob)
+                                     : std::numeric_limits<float_t>::infinity();
```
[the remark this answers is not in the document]
mfbalin: linter keeps moving it to the back for some reason.

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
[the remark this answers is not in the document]
frozenbugs: Can you fix it in a separate PR?

frozenbugs (approved):

graphbolt/src/cuda/neighbor_sampler.cu
```diff
@@ -0,0 +1,319 @@
...
+                  sub_indptr.data_ptr<indptr_t>(),
+                  sub_indptr.data_ptr<indptr_t>() + 1, stream));
+              auto tmp_storage =
+                  allocator.AllocateStorage<char>(tmp_storage_size);
+              CUDA_CALL(cub::DeviceSegmentedSort::SortPairs(
+                  tmp_storage.get(), tmp_storage_size, randoms.get(),
+                  randoms_sorted.get(), edge_id_segments.get(),
+                  sorted_edge_id_segments.get(), num_edges, num_rows,
+                  sub_indptr.data_ptr<indptr_t>(),
+                  sub_indptr.data_ptr<indptr_t>() + 1, stream));
```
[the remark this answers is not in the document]
mfbalin: I will think of ways to get around this inconveniency.

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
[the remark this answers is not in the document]
mfbalin: Ok, let me open a quick PR with only this fix now.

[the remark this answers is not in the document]
mfbalin: #6856 has this fix. After it is merged, this change should go away from this PR.

python/dgl/graphbolt/impl/neighbor_sampler.py
mfbalin: added

graphbolt/src/cuda/neighbor_sampler.cu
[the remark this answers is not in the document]
mfbalin: #6860 tracks this issue.

PR #6784, opened 2023-12-20
[Graphbolt] Implement Temporal Neighbor Sampling.

czkkkkkk:
## Description
Implemented TemporalNeighborSampler.

Since there are more and more samplers in Graphbolt, we may need to refactor the sampling logic. The current PickFn and NumPickFn are lambdas, which are pretty tedious since they forward a lot of arguments multiple times during the sampling logic. And it is not straightforward to support template in lambda. I'd suggest to use a abstract class for various sampling methods, which offer NumPick and Pick functions but with private attributes to support different sampling logics.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/src/fused_csc_sampling_graph.cc
+++ b/graphbolt/src/fused_csc_sampling_graph.cc
@@ -349,6 +349,31 @@ auto GetNumPickFn(
   };
 }
 
+auto GetTemporalNumPickFn(
+    torch::Tensor seed_timestamp, torch::Tensor csc_indices,
+    const std::vector<int64_t>& fanouts, bool replace,
+    const torch::optional<torch::Tensor>& type_per_edge,
+    const torch::optional<torch::Tensor>& probs_or_mask,
+    const torch::optional<torch::Tensor>& node_timestamp,
+    const torch::optional<torch::Tensor>& edge_timestamp) {
+  // If fanouts.size() > 1, returns the total number of all edge types of the
+  // given node.
+  return [&seed_timestamp, &csc_indices, &fanouts, replace, &probs_or_mask,
+          &type_per_edge, &node_timestamp, &edge_timestamp](
+             int64_t seed_offset, int64_t offset, int64_t num_neighbors) {
+    if (fanouts.size() > 1) {
+      return TemporalNumPickByEtype(
+          seed_timestamp, csc_indices, fanouts, replace, type_per_edge.value(),
+          probs_or_mask, node_timestamp, edge_timestamp, seed_offset, offset,
+          num_neighbors);
+    } else {
+      return TemporalNumPick(
+          seed_timestamp, csc_indices, fanouts[0], replace, probs_or_mask,
+          node_timestamp, edge_timestamp, seed_offset, offset, num_neighbors);
+    }
+  };
+}
+
 /**
  * @brief Get a lambda function which contains the sampling process.
  *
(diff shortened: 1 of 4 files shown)


graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -669,6 +744,42 @@ int64_t NumPick(
...
+}
+
+int64_t TemporalNumPick(
+    torch::Tensor seed_timestamp, torch::Tensor csc_indics, int64_t fanout,
+    bool replace, const torch::optional<torch::Tensor>& probs_or_mask,
+    const torch::optional<torch::Tensor>& node_timestamp,
+    const torch::optional<torch::Tensor>& edge_timestamp, int64_t seed_offset,
+    int64_t offset, int64_t num_neighbors) {
+  auto mask = TemporalMask(
+      seed_timestamp[seed_offset].item<int64_t>(), csc_indics, probs_or_mask,
```
peizhou001, line 776: Avoid use `item`. From experience it will singificantly slow down the program.

graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -669,6 +744,42 @@ int64_t NumPick(
...
+      node_timestamp, edge_timestamp, {offset, offset + num_neighbors});
+  int64_t num_valid_neighbors = mask.sum().item<int64_t>();
```
peizhou001, line 778: Ditto

graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -983,6 +1128,34 @@ int64_t Pick(
...
+    const torch::optional<torch::Tensor>& probs_or_mask,
+    const torch::optional<torch::Tensor>& node_timestamp,
+    const torch::optional<torch::Tensor>& edge_timestamp,
+    PickedType* picked_data_ptr) {
+  auto mask = TemporalMask(
+      seed_timestamp[seed_offset].item<int64_t>(), csc_indices, probs_or_mask,
+      node_timestamp, edge_timestamp, {offset, offset + num_neighbors});
+  torch::Tensor masked_prob;
+  if (probs_or_mask.has_value()) {
+    masked_prob =
```
peizhou001, line 1145: If `prob_or_mask `is Bool, will `* ` cause error?

graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -983,6 +1128,34 @@ int64_t Pick(
...
+        probs_or_mask.value().slice(0, offset, offset + num_neighbors) * mask;
+  } else {
+    masked_prob = mask.to(torch::kFloat32);
+  }
+  auto picked_indices = NonUniformPickOp(masked_prob, fanout, replace);
+  auto picked_indices_ptr = picked_indices.data_ptr<int64_t>();
+  for (int i = 0; i < picked_indices.numel(); ++i) {
+    picked_data_ptr[i] =
```
peizhou001, line 1153: Why not use tensor op here to leverage the underlying vectolization optimization?

peizhou001:
LGTM overall, please also add the test for new added function.

graphbolt/src/fused_csc_sampling_graph.cc
czkkkkkk: I see.

graphbolt/src/fused_csc_sampling_graph.cc
czkkkkkk: prob_or_mask has been casted to float32 before this function.

graphbolt/src/fused_csc_sampling_graph.cc
czkkkkkk: Do you mean we can use `picked_data_ptr` as a tensor? I just follow other sampling interface here and not sure why we need to pass value to return by a raw pointer.

czkkkkkk:
> LGTM overall, please also add the test for new added function.

Unittests are in another PR.

peizhou001 (approved):

PR #6792, opened 2023-12-20
[GraphBolt][CUDA] Pass GPU architecture option from DGL

mfbalin:
## Description
Setting the same `CUDA_ARCHITECTURES` as DGL build. Without this change, passing `-DCUDA_ARCH_NAME=All` to DGL does not trigger all architectures for the Graphbolt build.


mfbalin:
My latest commit seems to have resolved the multiple architecture case. I will use `cuobjdump` to ensure that the generated binary contents match the passed architecture list.

```
mfbalin@BALIN-PC:~/dgl-1$ cuobjdump /home/mfbalin/dgl-1/build/graphbolt/libgraphbolt_pytorch_2.0.1.so -sass -ptx | grep "code for sm_"
        code for sm_89
        code for sm_50
        code for sm_60
        code for sm_70
        code for sm_75
        code for sm_80
        code for sm_86
        code for sm_90
        code for sm_89
        code for sm_50
        code for sm_60
        code for sm_70
        code for sm_75
        code for sm_80
        code for sm_86
        code for sm_90
        code for sm_89
        code for sm_50
        code for sm_60
        code for sm_70
        code for sm_75
        code for sm_80
        code for sm_86
        code for sm_90
```

cmake/modules/CUDA.cmake
```diff
@@ -199,6 +202,7 @@ function(dgl_select_nvcc_arch_flags out_variable)
   string(REPLACE ";" " " __nvcc_archs_readable "${__nvcc_archs_readable}")
   set(${out_variable}          ${__nvcc_flags}          PARENT_SCOPE)
   set(${out_variable}_readable ${__nvcc_archs_readable} PARENT_SCOPE)
+  set(CUDA_ARCHITECTURES       ${__archs}               PARENT_SCOPE)
```
Rhett-Ying, line 205: what's the relationship between `CUDA_ARCHITECTURES` and existing `CUDA_ARCH_NAME`: https://github.com/dmlc/dgl/blob/3cf8246291d2ab559caae7f3bfafcb25a436405b/cmake/modules/CUDA.cmake#L124C8-L124C22. For now, we're using `CUDA_ARCH_NAME` alongside with `CUDA_ARCH_BIN` and `TORCH_CUDA_ARCH_LIST` to compile cuda/ptx for different archs.

mfbalin: `CUDA_ARCH_NAME` is a name like `Auto`, `All`, or a specific architecture like `Kepler`. However, specifying this option for DGL doesn't affect Graphbolt at all. Thus, I extract the SM version such as 50 or 80 etc. into `CUDA_ARCHITECTURES` and pass it to Graphbolt via `CUDAARCHS`. Then, setting `CMAKE_CUDA_ARCHITECTURES` for Graphbolt enables the same targets as DGL.

Rhett-Ying: could you check with `libdgl.so` and `dgl/dgl_sparse/libdgl_sparse_pytorchxxx` as well?

mfbalin: dgl_sparse doesn't have any `.cu` files at all in its src directory.

mfbalin: `libdgl.so` has all architectures when `All` option is passed. And:
```
mfbalin@BALIN-PC:~/dgl-1$ cuobjdump /home/mfbalin/dgl-1/build/dgl_sparse/libdgl_sparse_pytorch_2.0.1.so -sass -ptx | grep "code for sm_"
cuobjdump info    : File '/home/mfbalin/dgl-1/build/dgl_sparse/libdgl_sparse_pytorch_2.0.1.so' does not contain device code
```

Rhett-Ying:
@TristonC  @yaox12  could you help review this PR.

graphbolt/CMakeLists.txt
```diff
@@ -49,6 +49,9 @@ if(USE_CUDA)
     ${BOLT_DIR}/cuda/*.cu
   )
   list(APPEND BOLT_SRC ${BOLT_CUDA_SRC})
+  if(DEFINED ENV{CUDAARCHS})
+    set(CMAKE_CUDA_ARCHITECTURES $ENV{CUDAARCHS})
```
Rhett-Ying, line 53: Do we need to set it explicitly? According to the extracted from https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_ARCHITECTURES.html: Initialized by the [CUDAARCHS](https://cmake.org/cmake/help/latest/envvar/CUDAARCHS.html#envvar:CUDAARCHS) environment variable if set.

mfbalin: Only for cmake version greater than 3.20. to remove this line, we need to increase minimum cmake version to 3.20.

mfbalin: I added it because the current minimum is 3.18.

mfbalin: Also, for some reason, the `find_package(Torch REQUIRED)` line empties the contents of the `CMAKE_CUDA_ARCHITECTURES` variable, so I had to place this after that line.

mfbalin: Issue tracking this: https://github.com/pytorch/pytorch/issues/115937

Rhett-Ying (approved):

PR #6796, opened 2023-12-21
[GraphBolt][CUDA] Use Current Stream and introduce `CopyScalar` class

mfbalin:
## Description
Part 1 of #6770, Neighbor Sampling PR.

Adding a convenience function to get the current stream, was using the default stream before which is not something we want to do. Finally, adding a convenience class to overlap scalar device to host copies so that we don't experience kernel launch latency overhead. Their uses can be found in #6770.


graphbolt/src/cuda/common.h
```diff
@@ -88,6 +91,54 @@ inline bool is_zero<dim3>(dim3 size) {
...
+
+// This includes all integer, float and boolean types.
+#define GRAPHBOLT_DISPATCH_CASE_ALL_TYPES(...)            \
+  AT_DISPATCH_CASE_ALL_TYPES(__VA_ARGS__)                 \
+  AT_DISPATCH_CASE(at::ScalarType::Half, __VA_ARGS__)     \
+  AT_DISPATCH_CASE(at::ScalarType::BFloat16, __VA_ARGS__) \
+  AT_DISPATCH_CASE(at::ScalarType::Bool, __VA_ARGS__)
+
+#define GRAPHBOLT_DISPATCH_ALL_TYPES(TYPE, NAME, ...) \
+  AT_DISPATCH_SWITCH(TYPE, NAME, GRAPHBOLT_DISPATCH_CASE_ALL_TYPES(__VA_ARGS__))
```
frozenbugs, line 140: Unrelated to this PR?

graphbolt/src/cuda/common.h
```diff
@@ -88,6 +91,54 @@ inline bool is_zero<dim3>(dim3 size) {
...
+ * into scalar_t, the value can be read.
+ *
+ * auto num_edges = cuda::ReadScalar(indptr.data_ptr<scalar_t>() +
+ * indptr.size(0) - 1);
+ * // Perform many operations here, they will run as normal.
+ * // We finally need to read num_edges.
+ * auto indices = torch::empty(static_cast<scalar_t>(num_edges));
+ */
+template <typename scalar_t>
+struct ReadScalar {
```
frozenbugs, line 106: ReadScalar is a pretty bad name here, how come cudaMemcpyAsync happens in a method named read? Can you come up with a better name?

frozenbugs:
rename the header of the pr to more contextual.

graphbolt/src/cuda/common.h
frozenbugs: I would suggest we can create a separate PR for this code, which can be approved blindly, which is faster than adding confusion and then realize this PR did 3 things.

frozenbugs:
Note to reviewers: LGTM, expect for the name of ReadScalar.

graphbolt/src/cuda/common.h
mfbalin: Changed to CopyScalar, in the code, it is used as: `cuda::CopyScalar`.

graphbolt/src/cuda/common.h
```diff
@@ -90,6 +91,44 @@ inline bool is_zero<dim3>(dim3 size) {
...
+      copy_event.synchronize();
+      is_ready = true;
+    }
+    return reinterpret_cast<scalar_t*>(pinned_scalar.data_ptr())[0];
+  }
+
+ private:
+  torch::Tensor pinned_scalar;
+  at::cuda::CUDAEvent copy_event;
+  bool is_ready;
```
frozenbugs, line 129: nit: all internal variable should end with _, and rename is_ready -> synchronized_.
```suggestion
  torch::Tensor pinned_scalar_;
  at::cuda::CUDAEvent copy_event_;
  bool synchronized_;
```

frozenbugs (approved):

PR #6807, opened 2023-12-21
[GraphBolt] exclude nothing if edge is not found in node pairs

Rhett-Ying:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/sampled_subgraph.py
+++ b/python/dgl/graphbolt/sampled_subgraph.py
@@ -1,4 +1,5 @@
 """Graphbolt sampled subgraph."""
+
 # pylint: disable= invalid-name
 from typing import Dict, Tuple, Union
 
@@ -207,7 +208,7 @@ class SampledSubgraph:
                     )
                 index[etype] = _exclude_homo_edges(
                     reverse_edges,
-                    edges.get(etype),
+                    edges.get(etype, None),
                     assume_num_node_within_int32,
                 )
             if is_cscformat:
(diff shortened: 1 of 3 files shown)


peizhou001 (approved):

python/dgl/graphbolt/sampled_subgraph.py
```diff
@@ -266,8 +267,15 @@ def _relabel_two_arrays(lhs_array, rhs_array):
...
-def _exclude_homo_edges(edges, edges_to_exclude, assume_num_node_within_int32):
+def _exclude_homo_edges(
+    edges: Tuple[torch.Tensor, torch.Tensor],
+    edges_to_exclude: Tuple[torch.Tensor, torch.Tensor],
+    assume_num_node_within_int32: bool,
+):
     """Return the indices of edges that are not in edges_to_exclude."""
+    if edges_to_exclude is None:
+        # No edges are excluded, so return all indices of edges.
+        return torch.arange(len(edges[0]))
```
jermainewang, line 278: Creating a large array tensor may incur some overhead. My local fix skips `_exclude_homo_edges` into the `index` dictionary, and then in the later `_slice_subgraph` and `_slice_subgraph_node_pairs` functions also skips the slicing operator. The basic idea is to use `None` as a special value for identity mapping. Not sure if the fix will impact other parts. Just a suggestion.

Rhett-Ying: the overhead exists even for `Not None`. But we could remove such overhead for `None`.

Rhett-Ying: `index[etype]=None` looks hacky as `index` indicates it stores the index to be selected.

Rhett-Ying: But for the performance issue, I do agree that we need to take into consideration. Probably we should do the reverse: `excluded_index` instead of `index` as excluded edges are usually a small portion of original edges?

Rhett-Ying: For now, I handle `None` specially as you suggested.

jermainewang: > index[etype]=None looks hacky as index indicates it stores the index to be selected.

Agree. So I skip `index[etype] = ...` if the etype does not exist.

> But for the performance issue, I do agree that we need to take into consideration. Probably we should do the reverse: excluded_index instead of index as excluded edges are usually a small portion of original edges?

Yes. That could be a better option, but I'm not sure how will this complicate other parts.

python/dgl/graphbolt/sampled_subgraph.py
```diff
@@ -266,7 +271,11 @@ def _relabel_two_arrays(lhs_array, rhs_array):
     return mapping[: lhs_array.numel()], mapping[lhs_array.numel() :]
 
 
-def _exclude_homo_edges(edges, edges_to_exclude, assume_num_node_within_int32):
+def _exclude_homo_edges(
+    edges: Tuple[torch.Tensor, torch.Tensor],
+    edges_to_exclude: Tuple[torch.Tensor, torch.Tensor],
+    assume_num_node_within_int32: bool,
+):
     """Return the indices of edges that are not in edges_to_exclude."""
```
jermainewang, line 279: Nit: `"""Return the indices of edges to be included."""`

jermainewang (approved):

PR #6814, opened 2023-12-22
[Graphbolt] Add TemporalNeighborSampler.

czkkkkkk:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/impl/temporal_neighbor_sampler.py
+++ b/python/dgl/graphbolt/impl/temporal_neighbor_sampler.py
@@ -89,7 +89,7 @@ class TemporalNeighborSampler(SubgraphSampler):
         self.edge_timestamp_attr_name = edge_timestamp_attr_name
         self.sampler = graph.temporal_sample_neighbors
 
-    def sample_subgraphs(self, seeds, seeds_timestamp):
+    def sample_subgraphs(self, seeds, seeds_timestamp=None):
         subgraphs = []
         num_layers = len(self.fanouts)
         # Enrich seeds with all node types.
--- a/python/dgl/graphbolt/impl/temporal_neighbor_sampler.py
+++ b/python/dgl/graphbolt/impl/temporal_neighbor_sampler.py
@@ -1,0 +1,132 @@
+"""Temporal neighbor subgraph samplers for GraphBolt."""
+import torch
+from torch.utils.data import functional_datapipe
+
+from ..internal import compact_csc_format
+
+from ..subgraph_sampler import SubgraphSampler
+from .sampled_subgraph_impl import SampledSubgraphImpl
+
+
+__all__ = ["TemporalNeighborSampler"]
+
+
+@functional_datapipe("temporal_sample_neighbor")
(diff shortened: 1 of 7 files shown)


python/dgl/graphbolt/impl/temporal_neighbor_sampler.py
```diff
@@ -0,0 +1,132 @@
...
+        Boolean indicating whether the sample is preformed with or
+        without replacement. If True, a value can be selected multiple
+        times. Otherwise, each value can be selected only once.
+    prob_name: str, optional
+        The name of an edge attribute used as the weights of sampling for
+        each node. This attribute tensor should contain (unnormalized)
+        probabilities corresponding to each neighboring edge of a node.
+        It must be a 1D floating-point or boolean tensor, with the number
+        of elements equalling the total number of edges.
+    node_timestamp_attr_name: str, optional
```
peizhou001, line 54: Do we need to fix the string ?

python/dgl/graphbolt/internal/sample_utils.py
```diff
@@ -294,12 +294,29 @@ def unique_and_compact_csc_formats(
...
     """
-    Compact csc formats and return original_row_ids (per type).
+    Relabel the row (source) IDs in the csc formats into a contiguous range from
+    0 and return the original row node IDs per type.
+
+    Note that
+    1. The column (destination) IDs are included in the relabeled row IDs.
+    2. If there are repeated row IDs, they would not be uniqued and will be
+    treated as different nodes.
+    3. If `dst_timestamps` is given, the timestamp of each destination node will
```
peizhou001, line 318: Is this by design? That is, source nodes' timestamps is determined by corresponding destination nodes?

peizhou001 (approved):

python/dgl/graphbolt/internal/sample_utils.py
```diff
@@ -394,4 +463,20 @@ def compact_csc_format(
                     + offset
                 ),
             )
+            if use_timestamp:
+                src_timestamps[src_type] = torch.cat(
```
peizhou001, line 467: A little confused about this logic, can you add a comment for clafity?

czkkkkkk: Sure. This is to append the timestamps of sampled neighbors to `src_timestamps`. And `_broadcast_timestamps` broadcast destination timestamps to sources.

python/dgl/graphbolt/impl/temporal_neighbor_sampler.py
czkkkkkk: Do you mean the examples? I will add them later.

python/dgl/graphbolt/internal/sample_utils.py
czkkkkkk: Yes. The the source nodes are used as the frontier nodes in the next layer. They should retain the timestamps of the frontier nodes in this layer to filter neighbors based on the timestamps of the seed nodes.

czkkkkkk:
@dgl-bot

PR #6816, opened 2023-12-24
[GraphBolt] Extend temporal sampling to labor

mfbalin:
## Description
Extends temporal sampling to LayerNeighborSampler.


graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -426,37 +426,39 @@ auto GetPickFn(
...
+          return TemporalPickByEtype(
+              seed_timestamp, csc_indices, seed_offset, offset, num_neighbors,
+              fanouts, replace, options, type_per_edge.value(), probs_or_mask,
+              node_timestamp, edge_timestamp, args, picked_data_ptr);
+        } else {
+          int64_t num_sampled = TemporalPick(
+              seed_timestamp, csc_indices, seed_offset, offset, num_neighbors,
+              fanouts[0], replace, options, probs_or_mask, node_timestamp,
+              edge_timestamp, args, picked_data_ptr);
+          if (type_per_edge) {
```
peizhou001, line 456: If it is a tensor, but all value are 0, will it return false?

mfbalin: I think it is a torch::optional, so it is checking if is has a value. I didn't author this code, it was like this before the change.

I can change it to
`if (type_per_edge.value())` if you will like that better.

mfbalin:
@czkkkkkk Could you provide a review if you have the time?

graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -1152,22 +1171,22 @@ int64_t Pick(
     SamplerArgs<SamplerType::NEIGHBOR> args, PickedType* picked_data_ptr) {
   if (probs_or_mask.has_value()) {
     return NonUniformPick(
-        offset, num_neighbors, fanout, replace, options, probs_or_mask,
+        offset, num_neighbors, fanout, replace, options, probs_or_mask.value(),
```
czkkkkkk, line 1174: Is this a bug?

czkkkkkk (approved):
Actually I cannot follow the logic of labor sampling since our sampling code is a bit complicated. I plan to refactor the sampling code after a DDL in February. Since this PR doesn't affect the logic of temporal sampling, I will approve it and have more discussions during the refactor.

graphbolt/src/fused_csc_sampling_graph.cc
mfbalin: It is not, when a function expects a tensor, but we pass a `torch::optional<torch::Tensor>`, it automatically calls the `.value()` function of the optional.

mfbalin: I can take back the change if you so prefer.

mfbalin: Sorry, I wrote this code weeks ago so I misremembered, it is due to changing the type of the probs_or_mask argument of NonUniformPick in the change above from an optional to a Tensor.

graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -1130,11 +1148,12 @@ static torch::Tensor NonUniformPickOp(
 template <typename PickedType>
 inline int64_t NonUniformPick(
     int64_t offset, int64_t num_neighbors, int64_t fanout, bool replace,
-    const torch::TensorOptions& options,
-    const torch::optional<torch::Tensor>& probs_or_mask,
+    const torch::TensorOptions& options, const torch::Tensor& probs_or_mask,
```
mfbalin, line 1151: @czkkkkkk Since NonUniformPick requires a probs_or_mask argument, I changed its type from an optional to a plain tensor.

graphbolt/src/fused_csc_sampling_graph.cc
mfbalin: Changed it as suggested above to make it more readable.

PR #6829, opened 2023-12-26
[GraphBolt][CUDA] `gb.isin` implementation

mfbalin:
## Description
CUDA implementation of IsIn. Required to enable the Link tests of #6770, which I temporarily disabled on the GPU.

### DIFF
--- a/graphbolt/src/cuda/isin.cu
+++ b/graphbolt/src/cuda/isin.cu
@@ -1,0 +1,38 @@
+/**
+ *  Copyright (c) 2023 by Contributors
+ *  Copyright (c) 2023, GT-TDAlab (Muhammed Fatih Balin & Umit V. Catalyurek)
+ * @file cuda/isin.cu
+ * @brief IsIn operator implementation on CUDA.
+ */
+#include <graphbolt/cuda_ops.h>
+#include <thrust/binary_search.h>
+
+#include <cub/cub.cuh>
+
+#include "./common.h"
+
+namespace graphbolt {
+namespace ops {
+
+torch::Tensor IsIn(torch::Tensor elements, torch::Tensor test_elements) {
+  auto sorted_test_elements = Sort(test_elements).first;
+  auto allocator = cuda::GetAllocator();
+  auto stream = cuda::GetCurrentStream();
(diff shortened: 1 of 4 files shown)


graphbolt/src/isin.cc
```diff
@@ -16,6 +20,12 @@ namespace sampling {
 
 torch::Tensor IsIn(
     const torch::Tensor& elements, const torch::Tensor& test_elements) {
+  if (utils::is_accessible_from_gpu(elements) &&
```
frozenbugs, line 23: I would rename the IsIn to IsInCPU, and create a IsIn to do just dispatch.
IsIn() {
  if (blabla) {
    GRAPHBOLT_DISPATCH_CUDA_ONLY_DEVICE(...)
  } else {
    return IsInCPU(...)
  }
}

frozenbugs (approved):

graphbolt/src/cuda/isin.cu
```diff
@@ -0,0 +1,38 @@
...
+torch::Tensor IsIn(torch::Tensor elements, torch::Tensor test_elements) {
+  auto sorted_test_elements = Sort(test_elements).first;
+  auto allocator = cuda::GetAllocator();
+  auto stream = cuda::GetCurrentStream();
+  const auto exec_policy = thrust::cuda::par_nosync(allocator).on(stream);
+  auto result = torch::empty_like(elements, torch::kBool);
+
+  AT_DISPATCH_INTEGRAL_TYPES(
+      elements.scalar_type(), "IsInOperation", ([&] {
+        thrust::binary_search(
```
peizhou001, line 26: Is this required the data is sorted?

peizhou001:
A concern is is this implementation outperform torch version?
CPU version benefits from the multi thread, but GPU seems doesn't have this advantage.

mfbalin:
> A concern is is this implementation outperform torch version? CPU version benefits from the multi thread, but GPU seems doesn't have this advantage.

I will increase the performance of this operation in a future PR. The current implementation may not be faster but it will be.

Possible optimizations are:

1. Add a sort that does not return the second output, which is currently ignored anyway.
2. Before sort, compute num_bits.

PR #6841, opened 2023-12-26
[GraphBolt][CUDA] Reduce and hide `unique_and_compact` synchronizations.

mfbalin:
## Description
We eliminate one unnecessary GPU CPU synchronization from the code when computing the number of bits. When we run two `thrust::reduce` back to back, we synch at both which is unnecessary. So, we switch to using `cub` for the first one so when the second one synchs, we can have both results.

Profiling in #6770 shows that rather than the performance of the kernels, we are more bound by how many synchronizations exist in the code.


graphbolt/src/cuda/unique_and_compact_impl.cu
```diff
@@ -51,14 +51,26 @@ std::tuple<torch::Tensor, torch::Tensor, torch::Tensor> UniqueAndCompact(
...
+              temp.get(), workspace_size, src_ids_ptr, max_id_src_ptr,
+              src_ids.size(0));
+          auto max_id_dst = thrust::reduce(
               exec_policy, unique_dst_ids_ptr,
-              unique_dst_ids_ptr + unique_dst_ids.size(0), max_id,
-              thrust::maximum<scalar_t>{});
-          num_bits = cuda::NumberOfBits(max_id + 1);
+              unique_dst_ids_ptr + unique_dst_ids.size(0),
+              static_cast<scalar_t>(0), thrust::maximum<scalar_t>{});
+          // thrust::reduce syncs, so we can also safely access max_id_src_ptr
```
frozenbugs, line 71:
```suggestion
          // thrust::reduce syncs, which ensures safe access to max_id_src_ptr.
```

graphbolt/src/cuda/unique_and_compact_impl.cu
```diff
@@ -51,14 +51,26 @@ std::tuple<torch::Tensor, torch::Tensor, torch::Tensor> UniqueAndCompact(
         // If the given num_bits argument is not in the reasonable range,
         // we recompute it to speedup the expensive sort operations.
         if (num_bits <= 0 || num_bits > sizeof(scalar_t) * 8) {
-          auto max_id = thrust::reduce(
-              exec_policy, src_ids_ptr, src_ids_ptr + src_ids.size(0),
-              static_cast<scalar_t>(0), thrust::maximum<scalar_t>{});
-          max_id = thrust::reduce(
+          auto max_id_src = torch::empty(
```
frozenbugs, line 54: Are you re-implementing the max operation?
- why cuda does not provide the optimal op?
- what about move it out as a util function?

mfbalin: `thrust::reduce` synchronizes the GPU with the CPU. I am refactoring the pattern I use here into CopyScalar so that I don't have to allocate pinned tensors myself anymore. That is why, I re-marked this PR as a draft.

mfbalin: I will work on it a little more and ping you when it is ready to review again.

mfbalin: Refactored the max operator and refactored and optimized the whole implementation. Could you take another look?

graphbolt/src/cuda/unique_and_compact_impl.cu
```diff
@@ -48,17 +66,13 @@ std::tuple<torch::Tensor, torch::Tensor, torch::Tensor> UniqueAndCompact(
...
-          auto max_id = thrust::reduce(
-              exec_policy, src_ids_ptr, src_ids_ptr + src_ids.size(0),
-              static_cast<scalar_t>(0), thrust::maximum<scalar_t>{});
-          max_id = thrust::reduce(
-              exec_policy, unique_dst_ids_ptr,
-              unique_dst_ids_ptr + unique_dst_ids.size(0), max_id,
-              thrust::maximum<scalar_t>{});
-          num_bits = cuda::NumberOfBits(max_id + 1);
+        // If num_bits is not given, compute it to speedup the expensive sort
+        // operations later.
```
frozenbugs, line 70: The num_bits assignment is not happening in this block, consider rephrase the comment?

graphbolt/src/cuda/common.h
```diff
@@ -104,26 +104,41 @@ inline bool is_zero<dim3>(dim3 size) {
...
     }
-    return reinterpret_cast<scalar_t*>(pinned_scalar_.data_ptr())[0];
+    return *get();
   }
 
  private:
+  void init_pinned_storage() {
+    pinned_scalar_ = torch::empty(
+        sizeof(scalar_t),
+        c10::TensorOptions().dtype(torch::kBool).pinned_memory(true));
```
peizhou001, line 139: Why use Bool here

mfbalin: So that I can use this class with scalar_t types that have no corresponding torch dtype, such as std::pair<int,float>.

graphbolt/src/cuda/unique_and_compact_impl.cu
```diff
@@ -32,6 +33,23 @@ struct EqualityFunc {
...
+#define DefineReductionFunction(reduce_fn, name)                              \
+  template <typename scalar_iterator_t>                                       \
+  auto name(const scalar_iterator_t input, int64_t size) {                    \
+    auto allocator = cuda::GetAllocator();                                    \
+    auto stream = cuda::GetCurrentStream();                                   \
+    using scalar_t = std::remove_reference_t<decltype(input[0])>;             \
+    cuda::CopyScalar<scalar_t> result;                                        \
+    size_t workspace_size = 0;                                                \
+    reduce_fn(nullptr, workspace_size, input, result.get(), size, stream);    \
+    auto temp = allocator.AllocateStorage<char>(workspace_size);              \
```
peizhou001, line 45: avoid naming like `temp`

graphbolt/src/cuda/unique_and_compact_impl.cu
```diff
@@ -32,6 +33,23 @@ struct EqualityFunc {
...
 
+#define DefineReductionFunction(reduce_fn, name)                              \
+  template <typename scalar_iterator_t>                                       \
+  auto name(const scalar_iterator_t input, int64_t size) {                    \
+    auto allocator = cuda::GetAllocator();                                    \
+    auto stream = cuda::GetCurrentStream();                                   \
+    using scalar_t = std::remove_reference_t<decltype(input[0])>;             \
+    cuda::CopyScalar<scalar_t> result;                                        \
+    size_t workspace_size = 0;                                                \
+    reduce_fn(nullptr, workspace_size, input, result.get(), size, stream);    \
```
peizhou001, line 44: Add a comment to show this is not get the required size of temporal storage.

graphbolt/src/cuda/unique_and_compact_impl.cu
```diff
@@ -32,6 +33,23 @@ struct EqualityFunc {
...
+    auto temp = allocator.AllocateStorage<char>(workspace_size);              \
+    reduce_fn(temp.get(), workspace_size, input, result.get(), size, stream); \
```
peizhou001, line 46: A strange thing is `CopyScalar` here only works as a storage but copy nothing, please correct if misunderstanding.

peizhou001:
Is the benefits of this PR comes from the fact that : `cub` doesn't have implicit synchronization while `thrust` have?

graphbolt/src/cuda/common.h
peizhou001: Book make sense, ignore previous comment.

peizhou001 (approved):

graphbolt/src/cuda/unique_and_compact_impl.cu
mfbalin: Yes, in this case we can use it as storage as the API takes a scalar as input. Other times, such as copying the last element of indptr to get the num edges, it performs a copy when constructed with a device pointer. It also has the facilities to synchronize the stream for the copy. However, we use the least functionality we can depending on the situation to maximize performance.

graphbolt/src/cuda/unique_and_compact_impl.cu
mfbalin: This is a common pattern in CUB. I will add a CUB_CALL so that we do not have to keep writing this pattern of calls. Opening an issue to track it.

mfbalin: #6860

graphbolt/src/cuda/unique_and_compact_impl.cu
mfbalin: The copy happens implicitly as the underlying kernel directly writes the result into the CPU RAM, since the storage is pinned, we can do that.

mfbalin:
> s PR comes from the fact that : `cub` doesn't have implicit synchron

Yes, but not all `thrust` API are synchronizing. Only the ones such as `thrust::reduce` that return the reduction result as the function return value are synchronizing. So I am eliminating those ones only so that I can manage the copies myself and optimize them.

mfbalin:
@dgl-bot

PR #6846, opened 2023-12-27
[Graphbolt] Support temporal sampling in SubgraphSampler.

czkkkkkk:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/impl/temporal_neighbor_sampler.py
+++ b/python/dgl/graphbolt/impl/temporal_neighbor_sampler.py
@@ -117,10 +117,10 @@ class TemporalNeighborSampler(SubgraphSampler):
                 original_row_node_ids,
                 compacted_csc_formats,
                 row_timestamps,
-            ) = compact_csc_format(subgraph.node_pairs, seeds, seeds_timestamp)
+            ) = compact_csc_format(subgraph.sampled_csc, seeds, seeds_timestamp)
 
             subgraph = SampledSubgraphImpl(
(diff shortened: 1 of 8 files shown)


czkkkkkk:
@dgl-bot

czkkkkkk:
@dgl-bot

python/dgl/graphbolt/impl/temporal_neighbor_sampler.py
```diff
@@ -90,6 +90,7 @@ def __init__(
         self.sampler = graph.temporal_sample_neighbors
 
     def sample_subgraphs(self, seeds, seeds_timestamp=None):
+        assert seeds_timestamp is not None, "seeds_timestamp must be provided."
```
Rhett-Ying, line 93: can we remove the default value in above line as it looks weird that `seeds_timestamp` is None in default but it shouldn't be None.

Rhett-Ying: I find it's the same method signature of base `SubgraphSampler`. But such signature has to be for `NeighborSampler` looks weird as it seems such additional argument is for `TemporalNS` only.

Rhett-Ying (approved):
LGTM

python/dgl/graphbolt/impl/temporal_neighbor_sampler.py
czkkkkkk: Make sense. I removed the default value.

PR #6847, opened 2023-12-27
[DistDGL] fix distributed partition issue

Rhett-Ying:
## Description

The bug in previous implementation of `srcids update` is the line that recover the sorted `srcids` back to original order: `srcids[sort_ids] = srcids`.  Such in-place is incorrect as it sequentially overwrites elements in the sorted array, leading to incorrect results and loss of data. This error will results in unexpected value of `NID/EID`. The correct way is to create a new array to store like:
```
new_srcids = np.zeros_like(srcids)
new_srcids[sort_ids] = srcids
srcids = new_srcids
```

But in this PR, I[Rui] just replace previous implementation via a single call: `srcids = np.searchsorted(uniques, srcids, side="left")`. From my understanding, it should be more efficient in both performance and memory compared to iterating over manually. But for the record, I haven't profiled it.

**Previous Implementation**
```
# build inverse idxes for srcids, dstids and nids
    srcids = np.searchsorted(uniques, srcids, side="left")
    # over-write the srcids and dstids arrays.
    sort_ids = np.argsort(srcids)
    srcids = srcids[sort_ids]

    # TODO: check if wrapping this while loop in a c++ wrapper
    # helps in speeding up the code
    idx1 = 0
    idx2 = 0
    while (idx1 < len(srcids)) and (idx2 < len(uniques)):
        if srcids[idx1] == uniques[idx2]:
            srcids[idx1] = idx2
            idx1 += 1
        elif srcids[idx1] < uniques[idx2]:
            idx1 += 1
        else:
            idx2 += 1

    assert idx1 >= len(srcids), (
        f"Failed to locate all srcids in the uniques array "
        f" len(srcids) = {len(srcids)}, idx1 = {idx1} "
        f" len(uniques) = {len(uniques)}, idx2 = {idx2}"
    )
    srcids[sort_ids] = srcids
```
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/tests/scripts/task_distributed_test.sh
+++ b/tests/scripts/task_distributed_test.sh
@@ -36,4 +36,4 @@ export DMLC_LOG_DEBUG=1
 
 python3 -m pytest -v --capture=tee-sys --junitxml=pytest_distributed.xml --durations=100 tests/distributed/*.py || fail "distributed"
 
-#PYTHONPATH=tools:tools/distpartitioning:$PYTHONPATH python3 -m pytest -v --capture=tee-sys --junitxml=pytest_tools.xml --durations=100 tests/tools/*.py || fail "tools"
+PYTHONPATH=tools:tools/distpartitioning:$PYTHONPATH python3 -m pytest -v --capture=tee-sys --junitxml=pytest_tools.xml --durations=100 tests/tools/*.py || fail "tools"
--- a/tests/tools/test_change_etype_to_canonical_etype.py
+++ b/tests/tools/test_change_etype_to_canonical_etype.py
@@ -1,6 +1,7 @@
 import json
 import os
 import tempfile
+import unittest
 from collections import Counter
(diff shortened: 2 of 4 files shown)


Rhett-Ying:
@thvasilo could you try with this patch(you could just directly change the code in your installed path)? it passes the verify script I previously created.

thvasilo:
Yeah seems to solve the particular error we were seeing. I'd like to test with some large dataset with distributed execution as well, I'll have the time to do that tomorrow.

thvasilo:
Don't see any regression in our large scale test as well, so this should be good to merge. Thanks @Rhett-Ying !

mfbalin:
@Rhett-Ying Is it possible that this PR is the cause of the CI failures in #6865?

Rhett-Ying:
It's another known issue. let me fix it

PR #6850, opened 2023-12-27
[GraphBolt][CUDA] Extend SamplerNeighbors to heterogenous graphs

mfbalin:
## Description
We are extending the CUDA implementation for heterogeneous graphs. It depends on #6770 so it includes its changes.


mfbalin:
Will open another PR from master.

PR #6851, opened 2023-12-28
[GraphBolt] `CopyTo` node classification fix

mfbalin:
## Description
When running GPU sampling with #6770, I got an error saying that `seed_nodes` are not on the CUDA device. Thus, this should fix the issue.


peizhou001:
This could be intentionally set to save bandwidth, @RamonZhou for confirm.

mfbalin:
> This could be intentionally set to save bandwidth, @RamonZhou for confirm.

Right below, `seed_nodes` is being copied for `node_inference`. What is the different between the two?

PR #6861, opened 2023-12-28
[GraphBolt][CUDA] Enable GPU sampling in examples

mfbalin:
## Description
Making modifications to enable GPU sampling in the graphbolt examples. All modified examples seem to just work, haven't tested the multi-GPU lightning scenario yet (`num_gpus > 1`). I would appreciate it if you could help see if they all work and match the CPU sampling results.

Depends on #6957.

Google colab demo here: https://colab.research.google.com/drive/1AMnK9YlEP5dSWiz1Kg6V6DdVjsiSeN75?usp=sharing


python/dgl/graphbolt/minibatch.py
```diff
@@ -227,7 +227,7 @@ def blocks(self):
                         sampled_csc.indices,
                         torch.arange(
                             0,
-                            sampled_csc.indptr[-1],
+                            len(sampled_csc.indices),
```
Skeleton003, line 230: what's the difference?

mfbalin: You can check #6862 and #6863. The old one initiates a device-to-host transfer (from GPU to CPU), the new one does not initiate any such transfer. This is because `indptr` is on the CUDA device when sampling on the GPU.

mfbalin: I updated the links in #6862 since GitHub stopped rendering the images I had attached for some reason.

mfbalin:
@frozenbugs can you check `graphbolt/node_classification.py` and `graphbolt/link_prediction.py` files to see if the mixed mode option looks good?

Also if others can also test the examples with all the different available modes, that would be great. One thing is that FeatureStore does not have a `.to()` method, only has a `pin_memory_()` method. So even if the pair is `cuda-cuda`, we pin the memory for the features.

mfbalin:
Feel free to push to my branch as you like. The examples other than R-GCN all work. `node_classification.py` and `link_prediction.py` have the 4 different mixed modes enabled.

examples/sampling/graphbolt/node_classification.py
```diff
@@ -349,28 +362,35 @@ def parse_args():
...
     print("Loading data...")
     dataset = gb.BuiltinDataset("ogbn-products").load()
 
     graph = dataset.graph
-    # Currently the neighbor-sampling process can only be done on the CPU,
-    # therefore there is no need to copy the graph to the GPU.
     features = dataset.feature
+    # Move the dataset to the selected storage.
+    if args.storage_device == "pinned":
+        graph.pin_memory_()
```
Rhett-Ying, line 389: as we discussed, we'd better figure out a better way that requires less code in user side. maybe adding copy_to datapipe and pinning graph within `gb.DataLoader` is feasible.

mfbalin: How does this file look with the updated way of moving the graph and features?

mfbalin: The reason why there are 2 `copy_to`s in the dataloader is that I wanted the CPU sampling code path to be selectable by setting `--mode=cpu-cuda`. If this mode is deleted, then the last `copy_to` can be deleted safely. However, `pinned-cuda` will be preferable to `cpu-cuda` for all scenarios I can think of.

mfbalin: I used `args.storage_device` instead of having separate `args.feature_device` and `args.graph_device` to not overcomplicate it.

Rhett-Ying:
@mfbalin could you run below examples(run all the examples that are changed in this PR is preferable) with GPU sampling and share the accuracy results to see if it's comparable to the counterparts in DGL. We're looking into the accuracy issue in https://github.com/dmlc/dgl/issues/6941.

`python3 https://github.com/dmlc/dgl/tree/master/examples/sampling/graphbolt/rgcn/hetero_rgcn.py --dataset ogbn-mag --num_gpus 1`
`python3 https://github.com/dmlc/dgl/blob/master/examples/sampling/graphbolt/node_classification.py`

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -950,21 +950,23 @@ def _apply_to_members(self, fn):
 
         return self
 
-    def to(self, device: torch.device) -> None:  # pylint: disable=invalid-name
+    def to(
```
mfbalin, line 953: @frozenbugs if we can merge these changes in a separate PR, the examples will be runnable with DGL nightly.

examples/sampling/graphbolt/node_classification.py
```diff
@@ -309,9 +322,9 @@ def train(args, graph, features, train_set, valid_set, num_classes, model):
 
             total_loss += loss.item()
 
-        t1 = time.time()
         # Evaluate the model.
         acc = evaluate(args, model, graph, features, valid_set, num_classes)
+        t1 = time.time()
```
frozenbugs, line 327: Why change this? I think we are using this log in regression test to monitor the performance.

mfbalin: The time measurement did not take the validation stage into account and I thought it was misleading. When we print information about an epoch, we print the time as well and I think that it should be the whole epoch time.

mfbalin: I understand that the less changes, the easier to merge so we can consider this in a separate PR.

examples/sampling/graphbolt/node_classification.py
```diff
@@ -92,6 +92,18 @@ def create_dataloader(
...
     # [Step-2]:
+    # self.copy_to()
+    # [Input]:
+    # 'device': The device to copy the data to.
+    # 'extra_attrs': The extra attributes to copy.
+    # [Output]:
+    # A CopyTo object to copy the data to the specified device.
+    ############################################################################
+    if args.storage_device != "cpu":
+        datapipe = datapipe.copy_to(device=device, extra_attrs=["seed_nodes"])
```
frozenbugs, line 103: Can you add a comment to clarify this is for using CUDA sampling?

Also I am wondering whether we need a CUDA version of item sampler as well, such that we can move all seed nodes to CUDA, and shuffle from there.

mfbalin: I actually talked to @Rhett-Ying about it some time ago. However, since we wrap with a Prefetcher before the CopyTo in the dataloader, it actually doesn't cause a performance issue as the itemsampler runs on a separate thread with buffering.

examples/sampling/graphbolt/link_prediction.py
```diff
@@ -144,6 +144,15 @@ def create_dataloader(args, graph, features, itemset, is_train=True):
...
     )
 
+    ############################################################################
+    # [Input]:
+    # 'device': The device to copy the data to.
+    # [Output]:
+    # A CopyTo object to copy the data to the specified device.
+    ############################################################################
+    if args.storage_device != "cpu":
+        datapipe = datapipe.copy_to(device=args.device)
```
frozenbugs, line 154: Same to graphbolt/node_classification.py.

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
frozenbugs: Yes, let's merge this in a separate PR, the change LGTM.

examples/sampling/graphbolt/lightning/node_classification.py
frozenbugs, line 1: The lightning example is to demonstrate how GB works with lightning, let's keep it simple as it is.

examples/sampling/pyg/node_classification.py
frozenbugs, line 1: Let's also leave PYG out for same reason.

examples/sampling/graphbolt/rgcn/hetero_rgcn.py
frozenbugs, line 1: Same to rgcn.

examples/sampling/graphbolt/node_classification.py
mfbalin: Took back the change.

frozenbugs (approved):

examples/sampling/graphbolt/node_classification.py
mfbalin: Minimized the changed code by adding a .to method that can also pin object.

frozenbugs (approved):

PR #6867, opened 2023-12-29
[GraphBolt] Optimize `_convert_to_sampled_subgraph`.

yxy235:
## Description

Locally test `hetero_rgcn`.
Device: GPU.
Time(1 epoch): [00:59, 10.35it/s]
image

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
+++ b/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
@@ -425,75 +425,45 @@ class FusedCSCSamplingGraph(SamplingGraph):
             # The sampled graph is already a homogeneous graph.
             sampled_csc = CSCFormatBase(indptr=indptr, indices=indices)
         else:
-            # The sampled graph is a fused homogenized graph, which need to be
-            # converted to heterogeneous graphs.
-            # Pre-calculate the number of each etype
-            num = {}
-            for etype in type_per_edge:
-                num[etype.item()] = num.get(etype.item(), 0) + 1
-            # Preallocate
-            subgraph_indice_position = {}
-            subgraph_indice = {}
-            subgraph_indptr = {}
-            node_edge_type = defaultdict(list)
+            # 1. Find node types for each nodes in column.
+            node_types = (
+                torch.searchsorted(self.node_type_offset, column, right=True)
+                - 1
+            )
+
             original_hetero_edge_ids = {}
-            for etype, etype_id in self.edge_type_to_id.items():
-                subgraph_indice[etype] = torch.empty(
-                    (num.get(etype_id, 0),),
-                    dtype=indices.dtype,
-                    device=indices.device,
-                )
-                if has_original_eids:
-                    original_hetero_edge_ids[etype] = torch.empty(
-                        (num.get(etype_id, 0),),
-                        dtype=original_edge_ids.dtype,
-                        device=original_edge_ids.device,
-                    )
-                subgraph_indptr[etype] = [0]
-                subgraph_indice_position[etype] = 0
-                # Preprocessing saves the type of seed_nodes as the edge type
-                # of dst_ntype.
-                _, _, dst_ntype = etype_str_to_tuple(etype)
-                dst_ntype_id = self.node_type_to_id[dst_ntype]
-                node_edge_type[dst_ntype_id].append((etype, etype_id))
-            # construct subgraphs
-            for i, seed in enumerate(column):
-                l = indptr[i].item()
-                r = indptr[i + 1].item()
-                node_type = (
-                    torch.searchsorted(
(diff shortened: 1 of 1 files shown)


mfbalin:
Looks good to me, thank you for the quick fix!

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -425,75 +424,45 @@ def _convert_to_sampled_subgraph(
             # The sampled graph is already a homogeneous graph.
             sampled_csc = CSCFormatBase(indptr=indptr, indices=indices)
         else:
```
mfbalin, line 426: Right at the beginning of the else block, can we add this line:
`node_type_offset = self.node_type_offset.to(column.device)`

For GPU sampling, when we pin the graph, the node_type_offset is on the CPU. We need to temporarily move it to the GPU and use the moved version in the code below.

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -425,75 +424,45 @@ def _convert_to_sampled_subgraph(
...
-            for etype in type_per_edge:
-                num[etype.item()] = num.get(etype.item(), 0) + 1
-            # Preallocate
-            subgraph_indice_position = {}
-            subgraph_indice = {}
-            subgraph_indptr = {}
-            node_edge_type = defaultdict(list)
+            # 1. Find node types for each nodes in column.
+            node_types = (
+                torch.searchsorted(self.node_type_offset, column, right=True)
```
mfbalin, line 429: `torch.searchsorted(node_type_offset, column, right=True)`

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -425,75 +424,45 @@ def _convert_to_sampled_subgraph(
...
+                    eids = torch.nonzero(type_per_edge == etype_id).view(-1)
                     src_ntype_id = self.node_type_to_id[src_ntype]
-                    num_edges = torch.searchsorted(
-                        type_per_edge[l:r], etype_id, right=True
-                    ).item()
-                    end = num_edges + l
-                    subgraph_indptr[etype].append(
-                        subgraph_indptr[etype][-1] + num_edges
+                    sub_indices[etype] = (
+                        indices[eids] - self.node_type_offset[src_ntype_id]
```
mfbalin, line 448: `indices[eids] - node_type_offset[src_ntype_id]`

mfbalin (changes requested):
I am suggesting some changes so that GPU sampling works when the graph is in pinned_memory. When that is the case, we need to have a move like:
`self.node_type_offset = self.node_type_offset.to(column.device)` at the beginning.

This is a temporary workaround until we find a better solution, see how examples pin the graph in #6861. When we are not using GPU sampling, this extra line will not affect performance as the `to` operation will do nothing as it returns the passed tensor if it is already on the same device. Moreover, the movement should happen only once.

mfbalin:
Or, is it also acceptable to move the original tensor to the GPU, so instead of `node_type_offset = self.node_type_offset.to(column.device)`, we can do `self.node_type_offset = self.node_type_offset.to(column.device)`.

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -425,75 +424,45 @@ def _convert_to_sampled_subgraph(
...
+                    )
+                    cum_edges = torch.searchsorted(
+                        eids, indptr[nids + 1], right=False
```
peizhou001, line 451: move `indptr[nids + 1]` out from the inner loop as it keeps same for one node type.

peizhou001 (approved):

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
yxy235: > Right at the beginning of the else block (inside else), can we add this line: `self.node_type_offset = self.node_type_offset.to(column.device)`
> 
> For GPU sampling, when we pin the graph, the `self.node_type_offset` is on the CPU. We need to move it to the GPU before the following code starts to run. After the initial movement, the later calls will not do anything as it will already be on the device.

Make sense. The line has been added.

mfbalin (approved):
LGTM, GPU sampling should work for all cases now, whether the graph is pinned or is already on the GPU. Should we add a test for it in a separate PR?

yxy235:
> LGTM, GPU sampling should work for all cases now, whether the graph is pinned or is already on the GPU. @yxy235 could you add a test for it in a separate PR?

OK, I will add it next week.

PR #6870, opened 2023-12-29
[GraphBolt][CUDA] Refactor codebase with `CUB_CALL` macro

### LINKED ISSUE #6860: [GraphBolt][CUDA] Add a `CUB_CALL` macro and refactor codebase
## 🔨Work Item

**IMPORTANT:**

## Description

CUB requires us to call its functions twice, first call is to learn the workspace size, then we allocate workspace storage and call it again. A macro can do this seamlessly.

@frozenbugs @peizhou001

mfbalin:
## Description
Fixes #6860. We ensure that all kernel launches, CUB function calls and thrust function calls are run on the current stream, and with the proper workspace allocator. This makes the code much more concise and less error-prone, as the less we have to type, the less the chance of making a typo or mistake.


mfbalin:
@Rhett-Ying do you think we can merge this soon so that I can utilize the functionality here in my future and existing PRs?

graphbolt/src/cuda/common.h
```diff
@@ -82,15 +83,34 @@ inline bool is_zero<dim3>(dim3 size) {
...
+    auto allocator = graphbolt::cuda::GetAllocator();                         \
+    auto stream = graphbolt::cuda::GetCurrentStream();                        \
+    size_t workspace_size = 0;                                                \
+    CUDA_CALL(cub::fn(nullptr, workspace_size, __VA_ARGS__, stream));         \
+    auto workspace = allocator.AllocateStorage<char>(workspace_size);         \
+    CUDA_CALL(cub::fn(workspace.get(), workspace_size, __VA_ARGS__, stream)); \
+  }
+
+#define THRUST_CALL(fn, ...)                                                 \
+  [&] {                                                                      \
```
Rhett-Ying, line 107: wrap with `{}` and end with `;` before `}`?

Rhett-Ying (approved):

graphbolt/src/cuda/common.h
mfbalin: This is like a function call and we expect the users of this macro to end their statements with `;` of their own. If we add `;` at the end, then the user can finish their lines with no `;` at the end of `THRUST_CALL`:
```
int b = f(3);
THRUST_CALL(...)
int a = f(2);
```
But we want:
```
int b = f(3);
THRUST_CALL(...);
int a = f(2);
```

mfbalin: I needed to use `lambda` to be able to handle return values from thrust functions.

mfbalin: I hope we are on the same page, if not you can suggest a change here and I will make sure to fix it in a future PR immediately. Merging so that I can adapt other PRs on top of this PR.

PR #6886, opened 2024-01-02
[GraphBolt][CUDA] Add a `node_type_offset_list` property.

### LINKED ISSUE #6885: [GraphBolt][CUDA] `self.node_type_offset` should always be on the CPU
## 🔨Work Item

**IMPORTANT:**

## Description

https://github.com/dmlc/dgl/blob/d32a5980ca553f70fd55600b28dd8bb1017704c6/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py#L429
https://github.com/dmlc/dgl/blob/d32a5980ca553f70fd55600b28dd8bb1017704c6/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py#L920

When using GPU sampling either via when the graph is in pinned memory or GPU memory, `node_type_offset` is on the GPU due to the two lines above. However, it is required to be on the GPU only for these lines:

https://github.com/dmlc/dgl/blob/d32a5980ca553f70fd55600b28dd8bb1017704c6/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py#L431-L434

In other places, it is always better if `self.node_type_offset` is always on the CPU, otherwise, a lot of CPU GPU synchronization happens for each access to `self.node_type_offset`. I will think of an elegant solution. Opening an issue so that we don't forget about it.

@yxy235 @Rhett-Ying

mfbalin:
## Description
Adding a list version of `node_type_offset` that is called `node_type_offset_list`. Whenever we access it from the code with CPU indexes one-by-one, we should make sure to use `node_type_offset_list` so that there is no unnecessary CPU GPU transfer and synchronization as lists are guaranteed to be stored on the CPU and are fast to access element-wise. Should fix #6885.

### DIFF
--- a/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
+++ b/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
@@ -34,9 +34,6 @@ class FusedCSCSamplingGraph(SamplingGraph):
     ):
         super().__init__()
         self._c_csc_graph = c_csc_graph
-        self._node_type_offset = self.node_type_offset
-        if self._node_type_offset is not None:
-            self._node_type_offset = self._node_type_offset.tolist()
 
     @property
     def total_num_nodes(self) -> int:
@@ -95,7 +92,7 @@ class FusedCSCSamplingGraph(SamplingGraph):
         {'N0': 2, 'N1': 3}
         """
 
-        offset = self._node_type_offset
+        offset = self.node_type_offset_list
 
         # Homogenous.
         if offset is None or self.node_type_to_id is None:
@@ -215,15 +212,37 @@ class FusedCSCSamplingGraph(SamplingGraph):
         """
         return self._c_csc_graph.node_type_offset()
 
+    @property
+    def node_type_offset_list(self) -> Optional[list]:
+        """Returns the node type offset list if present.
+
+        Returns
+        -------
+        list or None
+            If present, returns a 1D integer list of shape
+            `(num_node_types + 1,)`. The list is in ascending order as nodes
+            of the same type have continuous IDs, and larger node IDs are
+            paired with larger node type IDs. The first value is 0 and last
+            value is the number of nodes. And nodes with IDs between
+            `node_type_offset_[i]~node_type_offset_[i+1]` are of type id 'i'.
+
+        """
+        if (
+            not hasattr(self, "_node_type_offset_list")
+            or self._node_type_offset_list is None
+        ):
+            self._node_type_offset_list = self.node_type_offset
+            if self._node_type_offset_list is not None:
+                self._node_type_offset_list = self.node_type_offset.tolist()
+        return self._node_type_offset_list
+
     @node_type_offset.setter
     def node_type_offset(
         self, node_type_offset: Optional[torch.Tensor]
     ) -> None:
         """Sets the node type offset tensor if present."""
-        self._c_csc_graph.set_node_type_offset(node_type_offset)
-        self._node_type_offset = None
-        if node_type_offset is not None:
-            self._node_type_offset = node_type_offset.tolist()
+        self._c_csc_graph.set_node_type_offset_list(node_type_offset)
+        self._node_type_offset_list = None
 
     @property
     def type_per_edge(self) -> Optional[torch.Tensor]:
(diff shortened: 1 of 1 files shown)


python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -34,6 +34,9 @@ def __init__(
     ):
         super().__init__()
         self._c_csc_graph = c_csc_graph
+        self._node_type_offset = self.node_type_offset
+        if self._node_type_offset:
```
mfbalin, line 38:
```suggestion
        if self._node_type_offset is not None:
```

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -218,6 +221,9 @@ def node_type_offset(
     ) -> None:
         """Sets the node type offset tensor if present."""
         self._c_csc_graph.set_node_type_offset(node_type_offset)
+        self._node_type_offset = None
+        if node_type_offset:
```
mfbalin, line 225:
```suggestion
        if node_type_offset is not None:
```

mfbalin (commented):
To fix the CI.

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -426,6 +454,7 @@ def _convert_to_sampled_subgraph(
             # The sampled graph is already a homogeneous graph.
             sampled_csc = CSCFormatBase(indptr=indptr, indices=indices)
         else:
+            # UVA sampling requires us to move node_type_offset to GPU.
             self.node_type_offset = self.node_type_offset.to(column.device)
```
yxy235, line 458: Why not we only change this line to `node_type_offset=self.node_type_offset.to(column.device)`, and still use `self.node_type_offset` in other places that prefer CPU?

mfbalin: Because when we move the graph to GPU, `self.node_type_offset` is moved as well. Then, there is synchronizations at every access to `self.node_type_offset`. If we keep moving to the GPU here at every iteration, then we pay the copy cost every iteration.

mfbalin: So we basically need to have two versions, one on the GPU, one on the CPU.

mfbalin: So that we use the approproate one depending on the operation.

yxy235: Make sense.

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -212,12 +214,39 @@ def node_type_offset(self) -> Optional[torch.Tensor]:
         """
         return self._c_csc_graph.node_type_offset()
 
+    @property
+    def node_type_offset_list(self) -> Optional[list]:
```
Rhett-Ying, line 218: I don't think we need a dedicated method to obtain the list version of `node_type_offset`. user is free to call `node_type_offset = g.node_type_offset.tolist()` if `tensor` is not what wanted.

Rhett-Ying: I think we'd better define the required APIs only to keep the API set to be concise.

Rhett-Ying (changes requested):
I don't think we need such a dedicated API for list version of `node_type_offset`

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
mfbalin: Can we make it a private property and call it `_node_type_offset_list`? We basically need a guaranteed CPU version of `node_type_offset` for use in GraphBolt internals.

mfbalin: For performance reasons.

Rhett-Ying: why do we need a guaranteed CPU version(list?)? we're free to call `tolist()` instead of `.item()` internally.

mfbalin: When graph is moved to GPU, node type offset is moved as well. However, when we access it in graphbolt internals, it triggers CPU GPU copies that are totally unnecessary and hurts performance.

mfbalin: All modifications in this PR are examples of CPU accesses to node type offset.

mfbalin: If we call tolist when we access, we copy it at every minibatch train iteration. Introducing this property let's us cache the copied version.

Rhett-Ying: ok, then let's make it private for now.

mfbalin: Pushed the private version.

Rhett-Ying (approved):

PR #6895, opened 2024-01-03
[GraphBolt][CUDA] Multiple fanout hetero sampling support.

mfbalin:
## Description
GPU should support multiple fanout values per layer with this PR.


mfbalin:
@yxy235 Can you help double-check the tests?

mfbalin:
@Rhett-Ying I found a bug in the `gb.seed()` functionality, can you take a quick look? I can open a separate PR for it after that.

Rhett-Ying:
> @Rhett-Ying I found a bug in the `gb.seed()` functionality, can you take a quick look? I realized it after setting the graphbolt seed but seeing that I keep getting different results. Opened a separate PR for it in #6900.

could you elaborate more about the root cause? It's about racing if multiple threads are setting seed simultaneously? and when such scenario happens? Calling `gb.seed()` at the begin of python causes racing? Does it work well in CPU sampling? Why such issue is exposed in GPU sampling?

mfbalin:
> > @Rhett-Ying I found a bug in the `gb.seed()` functionality, can you take a quick look? I realized it after setting the graphbolt seed but seeing that I keep getting different results. Opened a separate PR for it in #6900.
> 
> could you elaborate more about the root cause? It's about racing if multiple threads are setting seed simultaneously? and when such scenario happens? Calling `gb.seed()` at the begin of python causes racing? Does it work well in CPU sampling? Why such issue is exposed in GPU sampling?

The GPU random seed is derived from the CPU random engine. In the tests, I was setting the seed but kept getting different results. The issue mainly results from the fact that setting the manual seed doesn't set the existing randomEngine seed.

Rhett-Ying:
Yes. I totally aware of and agree with what you said. #6900 looks good to me. 

The reason why we didn't apply mutex on `manual_seed` is the assumption we call `gb.seed()` at the very beginning and it's on main thread of current process. No multi-threading is involved at that time. As for the root cause of you what you hit now is probably caused by `gb.seed()` is called in current test and when we run multiple tests, we're running on same process. that makes `ThreadLocal->Get()` returns previously set seed. I am wondering if this is the case. And if you run the test in your PR alone instead of all graphbolt tests, will you hit the issue? could you confirm this?

Anyway I want to figure why this issue is hit in this PR, not found before.

mfbalin:
> Yes. I totally aware of and agree with what you said. #6900 looks good to me.
> 
> The reason why we didn't apply mutex on `manual_seed` is the assumption we call `gb.seed()` at the very beginning and it's on main thread of current process. No multi-threading is involved at that time. As for the root cause of you what you hit now is probably caused by `gb.seed()` is called in current test and when we run multiple tests, we're running on same process. that makes `ThreadLocal->Get()` returns previously set seed. I am wondering if this is the case. And if you run the test in your PR alone instead of all graphbolt tests, will you hit the issue? could you confirm this?
> 
> Anyway I want to figure why this issue is hit in this PR, not found before.

Before I resorted to sorting the sampled edges in the tests, I was trying to see if I can set the seed and have GPU specialized expected test result arrays. But I noticed that the GPU results kept changing. A small investigation showed that the problem was with `gb.seed()` not setting the current thread random engine seed.

But now that I am sorting the sampled edges, we don't need to set the seed in the test file anymore. Let me delete it.

graphbolt/src/cuda/neighbor_sampler.cu
```diff
@@ -123,24 +124,112 @@ struct SegmentEndFunc {
...
+  new_indegree = new_indegree.slice(0, 1, num_rows + 2);
+  return {new_sub_indptr, new_indegree, new_sliced_indptr};
+}
+
 c10::intrusive_ptr<sampling::FusedSampledSubgraph> SampleNeighbors(
     torch::Tensor indptr, torch::Tensor indices, torch::Tensor nodes,
     const std::vector<int64_t>& fanouts, bool replace, bool layer,
     bool return_eids, torch::optional<torch::Tensor> type_per_edge,
     torch::optional<torch::Tensor> probs_or_mask) {
+  TORCH_CHECK(fanouts.size() >= 1, "At least one fanout value is needed.");
```
Rhett-Ying, line 196: such sanity checks are already done in upper caller https://github.com/dmlc/dgl/blob/25eabbb4efbd8a9b56f4a607b0a1f5c62b197c13/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py#L566-L583.

graphbolt/src/cuda/neighbor_sampler.cu
```diff
@@ -123,24 +124,112 @@ struct SegmentEndFunc {
...
+    const auto degree = sub_indptr[homo_i + 1] - indptr_i;
+    const etype_t etype = i % num_fanouts;
+    auto offset = cuda::LowerBound(etypes + indptr_i, degree, etype);
+    new_sub_indptr[i] = indptr_i + offset;
+    new_sliced_indptr[i] = sliced_indptr[homo_i] + offset;
+    if (i == num_rows - 1) new_sub_indptr[num_rows] = indptr_i + degree;
+  }
+};
+
+std::tuple<torch::Tensor, torch::Tensor, torch::Tensor> SliceCSCIndptrHetero(
```
Rhett-Ying, line 148: why not place it together with `SliceCSCIndptr` in `index_select_csc_impl.cu`? or is it possible to merge them into unified one?

tests/python/pytorch/graphbolt/impl/test_fused_csc_sampling_graph.py
```diff
@@ -698,36 +694,35 @@ def test_in_subgraph_homo():
     # Initialize data.
     total_num_nodes = 5
     total_num_edges = 12
-    indptr = torch.LongTensor([0, 3, 5, 7, 9, 12])
-    indices = torch.LongTensor([0, 1, 4, 2, 3, 0, 1, 1, 2, 0, 3, 4])
+    indptr = torch.tensor([0, 3, 5, 7, 9, 12], device=F.ctx())
+    indices = torch.tensor([0, 1, 4, 2, 3, 0, 1, 1, 2, 0, 3, 4], device=F.ctx())
```
yxy235, line 698: > @yxy235 Can you help double-check the tests?

LGTM. Maybe we do not need to set device for indptr and indices when constructing csc_sampling_graph.

graphbolt/src/cuda/neighbor_sampler.cu
mfbalin: Maybe we can take them out to a different file together?

tests/python/pytorch/graphbolt/impl/test_fused_csc_sampling_graph.py
mfbalin: Yeah, I must have blindly moved them to the device.

graphbolt/src/cuda/neighbor_sampler.cu
```diff
@@ -123,24 +124,107 @@ struct SegmentEndFunc {
...
+  for (size_t i = 0; i < fanouts.size(); i++) {
+    fanouts_pinned_ptr[i] =
+        fanouts[i] >= 0 ? fanouts[i] : std::numeric_limits<int64_t>::max();
+  }
+  // Finally, copy the adjusted fanout values to the device memory.
+  auto fanouts_device = allocator.AllocateStorage<int64_t>(fanouts.size());
+  CUDA_CALL(cudaMemcpyAsync(
+      fanouts_device.get(), fanouts_pinned_ptr,
+      sizeof(int64_t) * fanouts.size(), cudaMemcpyHostToDevice, stream));
   auto in_degree_and_sliced_indptr = SliceCSCIndptr(indptr, nodes);
```
Rhett-Ying, line 216: `auto [in_degree, sliced_indptr]`?

mfbalin: When I do that, the variables can not be captured by the dispatch lambdas unfortunately.

graphbolt/src/cuda/neighbor_sampler.cu
Rhett-Ying: yes, utility functions is better.

Rhett-Ying: even in another separate PR works for me.

graphbolt/src/cuda/utils.h
```diff
@@ -51,6 +51,31 @@ int NumberOfBits(const T& range) {
...
+ * of the first element which compares greater than or equal to value.
+ *
+ * This function assumes 0-based index
+ * @param A: ascending sorted array
+ * @param n: size of the A
+ * @param x: value to search in A
+ * @return index, i, of the first element st. A[i]>=x. If x>A[n-1] returns n.
+ * if x<A[0] then it returns 0.
+ */
+template <typename indptr_t, typename indices_t>
```
Rhett-Ying, line 65: is there any utility in cuda/thrust/cccl library instead of implementing on your own?

mfbalin: I checked quickly but wasn't able to find it.

graphbolt/src/cuda/neighbor_sampler.cu
Rhett-Ying: it seems a bug in c++ lambda?

mfbalin: I checked and they say that when you unpack variables with `auto [in_degree, sliced_indptr]`, they are not proper variables. That is why they can not be captured. Probably a bug in the specification that needs to be fixed.

graphbolt/src/cuda/neighbor_sampler.cu
mfbalin: It would be hard for me to test the correctness of the functionality in 2 separate PRs. If you don't mind, could we review the changes together? I moved the SliceCSCIndptr function from one file to another so we don't have to check it at all.

graphbolt/src/cuda/utils.h
Rhett-Ying: https://thrust.github.io/doc/group__binary__search_ga19da452d0b760bd927865ad2e0f45e41.html#ga19da452d0b760bd927865ad2e0f45e41 ?

mfbalin: I was using this documentation website. https://nvidia.github.io/cccl/thrust/api.html

graphbolt/src/cuda/sampling_utils.cu
```diff
@@ -0,0 +1,132 @@
...
+            stream);
+        auto tmp_storage = allocator.AllocateStorage<char>(tmp_storage_size);
+        cub::DeviceAdjacentDifference::SubtractLeftCopy(
+            tmp_storage.get(), tmp_storage_size,
+            new_sub_indptr.data_ptr<indptr_t>(),
+            new_indegree.data_ptr<indptr_t>(), num_rows + 1, cub::Difference{},
+            stream);
+      }));
+  // Discard the first element of the SubtractLeftCopy result.
+  new_indegree = new_indegree.slice(0, 1, num_rows + 2);
```
Rhett-Ying, line 127: what's the magic number `2` for?

Rhett-Ying: equivalent to `[1:]` in python?

mfbalin: Yes, I make sure to have the indegree tensor of size `num_rows + 1` so that doing an `ExclusiveCumSum` on it gives us the compacted `sub_indptr`.

mfbalin: The first number is garbage due to AdjacentDifference not having a pair of values to subtract for the first value.

graphbolt/src/cuda/utils.h
mfbalin: I will try and see if they work. If it works, I will replace UpperBound as well.

graphbolt/src/cuda/sampling_utils.cu
Rhett-Ying: does `new_indegree.slice(0, 1)` work?

graphbolt/src/cuda/utils.h
mfbalin: I think we are not supposed to use them. I got errors like this:
```
/home/mfbalin/dgl-1/graphbolt/src/cuda/sampling_utils.cu(83): warning #20011-D: calling a __host__ function("T1 thrust::THRUST_200300_890_NS::lower_bound<const unsigned char *, unsigned char> (T1, T1, const T2 &)") from a __host__ __device__ function("graphbolt::ops::EdgeTypeSearch<unsigned char, unsigned char> ::operator ()") is not allowed

Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"

/home/mfbalin/dgl-1/graphbolt/src/cuda/sampling_utils.cu(83): warning #20011-D: calling a __host__ function("T1 thrust::THRUST_200300_890_NS::lower_bound<const signed char *, signed char> (T1, T1, const T2 &)") from a __host__ __device__ function("graphbolt::ops::EdgeTypeSearch<unsigned char, signed char> ::operator ()") is not allowed

/home/mfbalin/dgl-1/graphbolt/src/cuda/sampling_utils.cu(83): warning #20011-D: calling a __host__ function("T1 thrust::THRUST_200300_890_NS::lower_bound<const int *, int> (T1, T1, const T2 &)") from a __host__ __device__ function("graphbolt::ops::EdgeTypeSearch<unsigned char, int> ::operator ()") is not allowed

/home/mfbalin/dgl-1/graphbolt/src/cuda/sampling_utils.cu(83): warning #20011-D: calling a __host__ function("T1 thrust::THRUST_200300_890_NS::lower_bound<const long *, long> (T1, T1, const T2 &)") from a __host__ __device__ function("graphbolt::ops::EdgeTypeSearch<unsigned char, long> ::operator ()") is not allowed

/home/mfbalin/dgl-1/graphbolt/src/cuda/sampling_utils.cu(83): warning #20011-D: calling a __host__ function("T1 thrust::THRUST_200300_890_NS::lower_bound<const short *, short> (T1, T1, const T2 &)") from a __host__ __device__ function("graphbolt::ops::EdgeTypeSearch<unsigned char, short> ::operator ()") is not allowed

/home/mfbalin/dgl-1/graphbolt/src/cuda/sampling_utils.cu(83): warning #20011-D: calling a __host__ function("T1 thrust::THRUST_200300_890_NS::lower_bound<const unsigned char *, unsigned char> (T1, T1, const T2 &)") from a __host__ __device__ function("graphbolt::ops::EdgeTypeSearch<signed char, unsigned char> ::operator ()") is not allowed

/home/mfbalin/dgl-1/graphbolt/src/cuda/sampling_utils.cu(83): warning #20011-D: calling a __host__ function("T1 thrust::THRUST_200300_890_NS::lower_bound<const signed char *, signed char> (T1, T1, const T2 &)") from a __host__ __device__ function("graphbolt::ops::EdgeTypeSearch<signed char, signed char> ::operator ()") is not allowed

/home/mfbalin/dgl-1/graphbolt/src/cuda/sampling_utils.cu(83): warning #20011-D: calling a __host__ function("T1 thrust::THRUST_200300_890_NS::lower_bound<const int *, int> (T1, T1, const T2 &)") from a __host__ __device__ function("graphbolt::ops::EdgeTypeSearch<signed char, int> ::operator ()") is not allowed

/home/mfbalin/dgl-1/graphbolt/src/cuda/sampling_utils.cu(83): warning #20011-D: calling a __host__ function("T1 thrust::THRUST_200300_890_NS::lower_bound<const long *, long> (T1, T1, const T2 &)") from a __host__ __device__ function("graphbolt::ops::EdgeTypeSearch<signed char, long> ::operator ()") is not allowed

/home/mfbalin/dgl-1/graphbolt/src/cuda/sampling_utils.cu(83): warning #20011-D: calling a __host__ function("T1 thrust::THRUST_200300_890_NS::lower_bound<const short *, short> (T1, T1, const T2 &)") from a __host__ __device__ function("graphbolt::ops::EdgeTypeSearch<signed char, short> ::operator ()") is not allowed
```

graphbolt/src/cuda/sampling_utils.cu
mfbalin: I will add comments about it.

mfbalin: https://pytorch.org/cppdocs/api/function_namespaceat_1ae6bdafd2514da71bc5a656a649bcb575.html

mfbalin: The arguments are dimension, start, and end. So that is the best I can do, unfortunately.

mfbalin: Sorry, you are right. `.slice(0, 1)` should work.

mfbalin: Yeah, it works. Fixed it. Is there anything else you want me to add here?

Rhett-Ying (approved):

yxy235 (approved):

PR #6900, opened 2024-01-04
[GraphBolt] Setting manual seed also sets the seed now.

mfbalin:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/src/random.cc
+++ b/graphbolt/src/random.cc
@@ -62,7 +62,7 @@ void RandomEngine::SetSeed(uint64_t seed, uint64_t stream) {
 
 /** @brief Manually fix the seed. */
 void RandomEngine::SetManualSeed(int64_t seed) {
-  // We intentionally set the seed for current thread also.
+  // Intentionally set the seed for current thread also.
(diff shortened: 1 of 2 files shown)


graphbolt/src/random.cc
```diff
@@ -59,6 +61,10 @@ void RandomEngine::SetSeed(uint64_t seed, uint64_t stream) {
 }
 
 /** @brief Manually fix the seed. */
-void RandomEngine::SetManualSeed(int64_t seed) { manual_seed = seed; }
+void RandomEngine::SetManualSeed(int64_t seed) {
+  RandomEngine::ThreadLocal()->SetSeed(seed);
```
Rhett-Ying, line 65: could you add a comment here that we set seed for current thread intentionally. this is the root cause of the issue?

Rhett-Ying (approved):

graphbolt/src/random.cc
mfbalin: Yes, that was the root cause. Let me do that. The mutex is just to make the code future-proof.

PR #6907, opened 2024-01-06
[DO NOT MERGE][GraphBolt][CUDA] Reduce `SampleNeighbors` compilation time.

mfbalin:
## Description
Should fix #6903. The CI normally took less than 50 minutes but started taking over an hour because of the long compilation time of GraphBolt CUDA SampleNeighbors. We split the dispatch of `indptr.scalar_type()` into different files so that their compilation can be parallelized. I would much rather prefer the solution in #6912 though.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [ ] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [ ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

PR #6912, opened 2024-01-07
[GraphBolt] Switch to using `AT_DISPATCH_INDEX_TYPES` for graph

### LINKED ISSUE #6903: [GraphBolt] Reduce compilation times.
## 🔨Work Item

**IMPORTANT:**

## Description

Reduce the compile time of GPU neighbor sampling by refactoring the main indptr dispatch and instantiating each type (uint8, int8, int16, int32, int64) in a separate file.

### LINKED ISSUE #6911: [GraphBolt] Dispatch types for different tensors
## 🔨Work Item

**IMPORTANT:**

## Description

Currently, we use `AT_DISPATCH_INTEGRAL_TYPES` for indptr, and indices/nodes tensors. It can accept `uint8`, `int8`, `int16`, `int32` and `int64` types. There are 25 combinations of types assuming `indices` and `nodes` are constrained to be the same type, otherwise, there are 125 combinations. However, as you can imagine, int8 and uint8 don't make much sense for `indptr` array as it requires that there are less than 127 or 255 edges in the graph. Even int16 doesn't make much sense for `indptr` as it requires that the graph has less than `32767` edges.

The same discussion applies to the nodes/indices tensors as well. The potential use of `int8` types implies that there are less than `255` vertices in the graph.

I propose to use the `AT_DISPATCH_INDEX_TYPES` (`int32` and `int64`) for both `indptr` and `indices/nodes` tensors so that there are only 4 combinations of template instantiations. Due to the GPU Neighbor Sampling implementation, the CI times increased from 45 minutes to more than 1 hour. I attempted to fix it in #6907, but if we switch to using `AT_DISPATCH_INDEX_TYPES` for `indptr` and `indices/nodes` tensors, we don't need to go forward with #6907 as the issue will be automatically resolved. If we want to, we can additionally support int16 for indices/nodes tensors (more realistic to have less than 32k vertices in the graph and results into only 6 combinations), while only supporting int32 and int64 for the `indptr` tensor.

mfbalin:
For `indptr`, `indices` and `nodes` tensors, let's use `AT_DISPATCH_INDEX_TYPES`. Should fix #6911 and fix #6903, and is a much better alternative to #6907. I might have missed other files and places, will add them when we build consensus on what solution to go forward with. The CI time is down to 43 minutes from well over an hour with this PR.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/src/cuda/neighbor_sampler.cu
+++ b/graphbolt/src/cuda/neighbor_sampler.cu
@@ -164,16 +164,16 @@ c10::intrusive_ptr<sampling::FusedSampledSubgraph> SampleNeighbors(
   auto max_in_degree = torch::empty(
       1,
       c10::TensorOptions().dtype(in_degree.scalar_type()).pinned_memory(true));
-  AT_DISPATCH_INTEGRAL_TYPES(
+  AT_DISPATCH_INDEX_TYPES(
       indptr.scalar_type(), "SampleNeighborsInDegree", ([&] {
         size_t tmp_storage_size = 0;
         cub::DeviceReduce::Max(
-            nullptr, tmp_storage_size, in_degree.data_ptr<scalar_t>(),
-            max_in_degree.data_ptr<scalar_t>(), num_rows, stream);
+            nullptr, tmp_storage_size, in_degree.data_ptr<index_t>(),
+            max_in_degree.data_ptr<index_t>(), num_rows, stream);
         auto tmp_storage = allocator.AllocateStorage<char>(tmp_storage_size);
         cub::DeviceReduce::Max(
-            tmp_storage.get(), tmp_storage_size, in_degree.data_ptr<scalar_t>(),
-            max_in_degree.data_ptr<scalar_t>(), num_rows, stream);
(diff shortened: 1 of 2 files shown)


graphbolt/src/cuda/neighbor_sampler.cu
```diff
@@ -164,16 +164,16 @@ c10::intrusive_ptr<sampling::FusedSampledSubgraph> SampleNeighbors(
   auto max_in_degree = torch::empty(
       1,
       c10::TensorOptions().dtype(in_degree.scalar_type()).pinned_memory(true));
-  AT_DISPATCH_INTEGRAL_TYPES(
+  AT_DISPATCH_INDEX_TYPES(
```
Rhett-Ying, line 167: the main advantage of this new macro is only generating code for `int32_t` and `int64_t` only, instead of all possible integral data types? so less code, less compile time, less compiled lib size?

mfbalin: Exactly. When we have nested dispatches, the combinations become exponential.

Rhett-Ying (approved):

PR #6932, opened 2024-01-11
[GraphBolt] change default value of persistent_workers to False

Rhett-Ying:
## Description

to be consistent with default value in `torch.utils.data.DataLoader` which is `False` in default.

I don't specify `persistent_workers=True` explicitly for our examples as `num_workers=0` is used in default. And I'd like to see if such change triggers regression.
## Checklist
- [ ] Code is well-documented


tests/python/pytorch/graphbolt/test_integration.py
```diff
@@ -1,10 +1,13 @@
 import dgl
 import dgl.graphbolt as gb
 import dgl.sparse as dglsp
+import pytest
 import torch
 
 
-def test_integration_link_prediction():
+@pytest.mark.parametrize("num_workers", [0])
```
RamonZhou, line 8: Why not set num_workers as [0, 4]?

Rhett-Ying: if `num_workers>0`, the return order of each iteration on dataloader is not determined.

Rhett-Ying: https://pytorch.org/docs/stable/notes/randomness.html#dataloader

PR #6939, opened 2024-01-11
[GraphBolt][CUDA] Add native `GPUCachedFeature` instead of using DGL

mfbalin:
## Description
`gb.GPUCachedFeature` was using `dgl.cuda.GPUCache` but GraphBolt or any of its components should not have a dependency on DGL. Hence, we link GraphBolt directly against the HugeCTR gpu_cache target and add its wrapper natively into GraphBolt. Test coverage is also increased and ensures that the cache works for any given datatype.

## Checklist
- [ ] Code is well-documented


graphbolt/src/cuda/gpu_cache.h
```diff
@@ -0,0 +1,64 @@
...
+  std::tuple<torch::Tensor, torch::Tensor, torch::Tensor> Query(
+      torch::Tensor keys);
+
+  void Replace(torch::Tensor keys, torch::Tensor values);
+
+  static c10::intrusive_ptr<GpuCache> Create(
+      const std::vector<int64_t>& shape, torch::ScalarType dtype);
+
+ private:
+  std::vector<int64_t> shape;
```
frozenbugs, line 49:
```suggestion
  std::vector<int64_t> shape_;
```
ditto for others.

graphbolt/src/cuda/gpu_cache.h
```diff
@@ -0,0 +1,64 @@
...
+  torch::ScalarType dtype;
+  std::unique_ptr<gpu_cache_t> cache;
+  int64_t num_bytes;
+  int64_t num_float_feats;
+  torch::DeviceIndex device_id;
+};
+
+// The cu file in HugeCTR gpu cache uses unsigned int and long long.
+// Changing to int64_t results in a mismatch of template arguments.
+static_assert(sizeof(long long) == 8);  // NOLINT
```
frozenbugs, line 59: Add error message? When this assert will be triggered, in compiling or runtime?

mfbalin: Compile time. It will never fail.

mfbalin: So long as `long long` is 8 bytes when compiling for any given system.

mfbalin: I need to ensure that `sizeof(long long) == sizeof(int64_t)`, otherwise code won't work. But for all systems I know, this is the case.

PR #6948, opened 2024-01-15
[GraphBolt] Add check about whether edge IDs are saved when edge feature is stored.

yxy235:
## Description

If edge IDs are not saved when edge feature is stored, make warnings to user.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/impl/ondisk_dataset.py
+++ b/python/dgl/graphbolt/impl/ondisk_dataset.py
@@ -163,6 +163,10 @@ def preprocess_ondisk_dataset(
                         graph_feature["name"]
                     ] = node_data
             if graph_feature["domain"] == "edge":
+                if not include_original_edge_id:
+                    dgl_warning(
+                        "Edge feature is stored, but edge IDs are not saved."
+                    )
                 edge_data = read_data(
                     os.path.join(dataset_dir, graph_feature["path"]),
                     graph_feature["format"],
--- a/python/dgl/graphbolt/impl/ondisk_dataset.py
+++ b/python/dgl/graphbolt/impl/ondisk_dataset.py
@@ -218,6 +218,7 @@ def preprocess_ondisk_dataset(
 
     # 7. Load the node/edge features and do necessary conversion.
     if input_config.get("feature_data", None):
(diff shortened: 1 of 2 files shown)


python/dgl/graphbolt/impl/ondisk_dataset.py
```diff
@@ -146,6 +146,10 @@ def preprocess_ondisk_dataset(
                         graph_feature["name"]
                     ] = node_data
             if graph_feature["domain"] == "edge":
+                if include_original_edge_id is False:
```
Skeleton003, line 149:
```suggestion
                if not include_original_edge_id:
```

python/dgl/graphbolt/impl/ondisk_dataset.py
```diff
@@ -146,6 +146,10 @@ def preprocess_ondisk_dataset(
                         graph_feature["name"]
                     ] = node_data
             if graph_feature["domain"] == "edge":
+                if not include_original_edge_id:
+                    dgl_warning(
+                        "Edge feature is stored, but edge IDs are not saved."
+                    )
```
frozenbugs, line 152: Why it is a problem that there are edge attributes in sampling graph while no eid?

yxy235: This check was proposed by Rui, if I have understood him, the reason we raise warning is that edge feature can not be fetched without eids. Loading edge feature with `include_original_edge_id=False` may cause confusion, especially when somebody prepare to use them in the follow-up process.

python/dgl/graphbolt/impl/ondisk_dataset.py
```diff
@@ -229,6 +229,10 @@ def preprocess_ondisk_dataset(
             in_memory = (
                 True if "in_memory" not in feature else feature["in_memory"]
             )
+            if feature["domain"] == "edge" and not include_original_edge_id:
```
frozenbugs, line 232: nit: this will raise multiple warnings.
What about add has_edge_feature_data in the for loop, and raise one warning after the for loop?

frozenbugs (approved):

PR #6954, opened 2024-01-15
[GraphBolt][CUDA] Overlap feature fetcher

mfbalin:
## Description

UVA feature fetching is overlapped with other operations. The picture below shows that `IndexSelectMultiKernelAligned` executes simultaneously as the GNN model forward/backward operations. We may need to tune the 6144 default value in a later PR for optimal performance (Might also want to implement autotuning logic). If it is too high, then too much unnecessary GPU resources will be used for the UVA copy, if it is too little, we can't fully utilize the PCI-e connection. However, it is safer for overall performance to set it to a higher value to ensure the PCI-e can be fully utilized in all different systems.
image

NOTE: The memory leak was caused by multiple CUDA streams being created when we created multiple dataloaders. Making uva_stream a global variable resolved the issue.

## Checklist
- [ ] Code is well-documented


graphbolt/src/cuda/max_uva_threads.h
```diff
@@ -0,0 +1,24 @@
...
+#define GRAPHBOLT_MAX_UVA_THREADS_H_
+
+#include <cstdint>
+#include <optional>
+
+namespace graphbolt {
+namespace cuda {
+
+/** @brief Set a limit on the number of CUDA threads for UVA accesses. */
+inline std::optional<int64_t> max_uva_threads;
```
frozenbugs, line 17: I don't really understand the keyword inline, for defining global variables should we use `extern std::optional<int64_t> max_uva_threads;`?

mfbalin: using inline in the header file ensures that the linker only keeps one instantiation of the variable even if the header is included in multiple source files.

mfbalin: That is what I learned from the internet from my brief research.

mfbalin: I tested the code, it works even though multiple source files include this header.

mfbalin: https://en.cppreference.com/w/cpp/language/inline, Explanation section. It looks like it is a C++17 feature.

python/dgl/graphbolt/dataloader.py
```diff
@@ -122,7 +187,35 @@ def __init__(self, datapipe, num_workers=0, persistent_workers=True):
             persistent_workers=persistent_workers,
         )
 
-        # (3) Cut datapipe at CopyTo and wrap with prefetcher. This enables the
+        # (3) Overlap UVA feature fetching by buffering and using an alternative
+        # stream.
```
frozenbugs, line 191: why do we need a separate stream?

mfbalin: Without seperate stream, overlap is not possible. Need at least 2 streams of execution to overlap them.

python/dgl/graphbolt/dataloader.py
```diff
@@ -122,7 +187,35 @@ def __init__(self, datapipe, num_workers=0, persistent_workers=True):
...
+        if (
+            overlap_feature_fetch
+            and num_workers == 0
+            and torch.cuda.is_available()
+        ):
+            torch.ops.graphbolt.set_max_uva_threads(max_uva_threads)
```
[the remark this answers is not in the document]
mfbalin: if torch cuda is available, can't we assume that GRAPHBOLT_USE_CUDA is turned on?

python/dgl/graphbolt/dataloader.py
```diff
@@ -86,14 +143,22 @@ class DataLoader(torch.utils.data.DataLoader):
...
     """
 
-    def __init__(self, datapipe, num_workers=0, persistent_workers=True):
+    def __init__(
+        self,
+        datapipe,
+        num_workers=0,
+        persistent_workers=True,
+        overlap_feature_fetch=True,
+        max_uva_threads=6144,
```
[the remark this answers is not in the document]
mfbalin: added context

python/dgl/graphbolt/dataloader.py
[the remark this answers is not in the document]
mfbalin: I can compile the newly added cc file even if GRAPHBOLT_USE_CUDA is off if you want, I just need to move the addition of the source file outside of the USE_CUDA guards in both cmake and the python_binding.cc

python/dgl/graphbolt/dataloader.py
```diff
@@ -35,6 +38,52 @@ def _find_and_wrap_parent(
...
+
+    def __iter__(self):
+        for data in self.datapipe:
+            yield data
+
+
+class Bufferer(dp.iter.IterDataPipe):
+    """Buffers items before yielding them."""
+
+    def __init__(self, datapipe, buffer_size=2):
```
frozenbugs, line 55: Add a doc about why buffer_size is default to 2 and how can user decide what's the best for their cases.

python/dgl/graphbolt/dataloader.py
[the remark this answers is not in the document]
frozenbugs: I am actually suggestion remove the flag since I don't really understand the benefit. We probably only need to make sure that for pure CPU build, we can ignore the cuda code.

frozenbugs (approved):

PR #6962, opened 2024-01-17
[GraphBolt][CUDA] Inplace pin memory for Graph and TorchFeatureStore

mfbalin:
## Description
torch pin_memory method creates a copy of the tensor. When we work with large datasets or use multi-GPU training, we don't want copies to be made. So, this PR ensures that `pin_memory_()` method is in-place by using `cudaHostRegister`.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
+++ b/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
@@ -36,6 +36,9 @@ class FusedCSCSamplingGraph(SamplingGraph):
         self._c_csc_graph = c_csc_graph
 
     def __del__(self):
+        # torch.Tensor.pin_memory() is not an inplace operation. To make it
+        # truly in-place, we need to use cudaHostRegister. Then, we need to use
+        # cudaHostUnregister to unpin the tensor in the destructor.
         # https://github.com/pytorch/pytorch/issues/32167#issuecomment-753551842
         if hasattr(self, "_is_inplace_pinned"):
             cudart = torch.cuda.cudart()
@@ -983,6 +986,9 @@ class FusedCSCSamplingGraph(SamplingGraph):
         """Copy `FusedCSCSamplingGraph` to the pinned memory in-place."""
         self._is_inplace_pinned = set()
 
+        # torch.Tensor.pin_memory() is not an inplace operation. To make it
+        # truly in-place, we need to use cudaHostRegister. Then, we need to use
+        # cudaHostUnregister to unpin the tensor in the destructor.
         # https://github.com/pytorch/pytorch/issues/32167#issuecomment-753551842
         cudart = torch.cuda.cudart()
 
--- a/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
+++ b/python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
@@ -1006,7 +1006,7 @@ class FusedCSCSamplingGraph(SamplingGraph):
 
             return x
 
-        return self._apply_to_members(_pin)
+        self._apply_to_members(_pin)
 
 
 def fused_csc_sampling_graph(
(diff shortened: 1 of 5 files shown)


python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -169,14 +176,25 @@ def metadata(self):
 
     def pin_memory_(self):
         """In-place operation to copy the feature to pinned memory."""
-        self._tensor = self._tensor.pin_memory()
+        self._is_inplace_pinned = set()
+        # https://github.com/pytorch/pytorch/issues/32167#issuecomment-753551842
```
frozenbugs, line 180: Add context instead just paste the link here.

# blababla, see more details:
# https://github.com/pytorch/pytorch/issues/32167#issuecomment-753551842

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -84,6 +84,13 @@ def __init__(self, torch_feature: torch.Tensor, metadata: Dict = None):
         self._tensor = torch_feature.contiguous()
         self._metadata = metadata
 
+    def __del__(self):
+        # https://github.com/pytorch/pytorch/issues/32167#issuecomment-753551842
+        if hasattr(self, "_is_inplace_pinned"):
```
frozenbugs, line 89: ditto.

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -974,9 +981,30 @@ def _pin(x):
 
     def pin_memory_(self):
         """Copy `FusedCSCSamplingGraph` to the pinned memory in-place."""
+        self._is_inplace_pinned = set()
+
+        # https://github.com/pytorch/pytorch/issues/32167#issuecomment-753551842
```
frozenbugs, line 986: ditto.

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -35,6 +35,13 @@ def __init__(
         super().__init__()
         self._c_csc_graph = c_csc_graph
 
+    def __del__(self):
+        # https://github.com/pytorch/pytorch/issues/32167#issuecomment-753551842
```
frozenbugs, line 39: ditto.

python/dgl/graphbolt/impl/torch_based_feature_store.py
mfbalin: Added context, can you check?

frozenbugs (approved):

mfbalin:
@frozenbugs, I measured the memory consumption of the multi-GPU example in #6961. Without this PR, the consumption grows as more GPUs are used. With this PR, adding more GPUs does not significantly change the memory consumption. The tests pass as well.

TristonC:
We saw better scale up performance one a single DGX node from 1GU to 8GPUs with this PR.

mfbalin:
> We saw better scale up performance one a single DGX node from 1GU to 8GPUs with this PR.

Why does this PR lead to better performance? I thought it would only help lower the memory usage.

TristonC:
> > We saw better scale up performance one a single DGX node from 1GU to 8GPUs with this PR.
> 
> Why does this PR lead to better performance? I thought it would only help lower the memory usage.

Will find out if is just his PR or other PR combined effect.

python/dgl/graphbolt/impl/fused_csc_sampling_graph.py
```diff
@@ -974,9 +985,31 @@ def _pin(x):
...
 
         def _pin(x):
-            return x.pin_memory() if hasattr(x, "pin_memory") else x
+            if hasattr(x, "pin_memory_"):
+                x.pin_memory_()
+            elif (
+                isinstance(x, torch.Tensor)
+                and not x.is_pinned()
+                and x.device.type == "cpu"
+                and x.is_contiguous()
```
czkkkkkk, line 1001: If a tensor is not contiguous, it will cause some unexpected errors if users try to pin it. So I'd suggest to raise an error if the device type is not GPU or the tensor is not contiguous.

mfbalin: Addressed it. Thanks!

czkkkkkk (approved):

Issue #6978, opened 2024-01-18
[GraphBolt][CUDA] MultiGPU example crashes


mfbalin:
## 🔨Work Item

**IMPORTANT:**
* This template is only for dev team to track project progress. For feature request or bug report, please use the corresponding issue templates.
* DO NOT create a new work item if the purpose is to fix an existing issue or feature request. We will directly use the issue in the project tracker.

Project tracker: https://github.com/orgs/dmlc/projects/2

## Description

There seems to be a bug in the multi-GPU graphbolt example. Enabling overlap makes it crash while I can't profile the code even with overlap disabled.

mfbalin:
Error disappers when the multiGPU example is launched with `CUDA_LAUNCH_BLOCKING=1`.

mfbalin:
It appears that the multi-GPU example is buggy even with overlap_fetch_feature disabled.

mfbalin:
@RamonZhou any insights?

PR #6979, opened 2024-01-18
[GraphBolt][CUDA] Handle edge case for fetch feature overlap

mfbalin:
## Description
If the user runs the following:
```python
for data in dataloader:
    pass
```

Moreover, this PR fixes the multi-GPU example crash tracked by #6978. I believe the crash is due to dropping uneven inputs, basically meaning that we discard the last minibatch coming from the dataloader. It is like the case above but only just for the last minibatch.

Because the feature fetch operation doesn't synchronize, the edge id and vertex id tensors might get freed while the fetch operation is still ongoing. If there are forward/backward operations inside the loop, this doesn't happen because we make the current stream wait for the UVA stream via CUDA events.

https://github.com/dmlc/dgl/blob/f7e065f7672198f033bd50da8694a5ec20b4a160/python/dgl/graphbolt/dataloader.py#L85-L94
https://github.com/dmlc/dgl/blob/f7e065f7672198f033bd50da8694a5ec20b4a160/python/dgl/graphbolt/feature_fetcher.py#L165-L172

I reproduced a crash by replacing the training loop of the `graphbolt/node_classification.py` with a simple `pass` expression to simulate edge case above.

```
/localscratch/dgl-1/graphbolt/src/cuda/index_select_impl.cu:84: void graphbolt::ops::IndexSelectMultiKernelAligned(const DType *, signed long, signed long, const IdType *, signed long, DType *, const signed long *) [with DType = float4; IdType = signed long]: block: [6,0,0], thread: [26,12,0] Assertion `in_row >= 0 && in_row < input_len` failed.
/localscratch/dgl-1/graphbolt/src/cuda/index_select_impl.cu:84: void graphbolt::ops::IndexSelectMultiKernelAligned(const DType *, signed long, signed long, const IdType *, signed long, DType *, const signed long *) [with DType = float4; IdType = signed long]: block: [6,0,0], thread: [27,12,0] Assertion `in_row >= 0 && in_row < input_len` failed.
/localscratch/dgl-1/graphbolt/src/cuda/index_select_impl.cu:84: void graphbolt::ops::IndexSelectMultiKernelAligned(const DType *, signed long, signed long, const IdType *, signed long, DType *, const signed long *) [with DType = float4; IdType = signed long]: block: [6,0,0], thread: [28,12,0] Assertion `in_row >= 0 && in_row < input_len` failed.
/localscratch/dgl-1/graphbolt/src/cuda/index_select_impl.cu:84: void graphbolt::ops::IndexSelectMultiKernelAligned(const DType *, signed long, signed long, const IdType *, signed long, DType *, const signed long *) [with DType = float4; IdType = signed long]: block: [6,0,0], thread: [29,12,0] Assertion `in_row >= 0 && in_row < input_len` failed.
/localscratch/dgl-1/graphbolt/src/cuda/index_select_impl.cu:84: void graphbolt::ops::IndexSelectMultiKernelAligned(const DType *, signed long, signed long, const IdType *, signed long, DType *, const signed long *) [with DType = float4; IdType = signed long]: block: [6,0,0], thread: [30,12,0] Assertion `in_row >= 0 && in_row < input_len` failed.
/localscratch/dgl-1/graphbolt/src/cuda/index_select_impl.cu:84: void graphbolt::ops::IndexSelectMultiKernelAligned(const DType *, signed long, signed long, const IdType *, signed long, DType *, const signed long *) [with DType = float4; IdType = signed long]: block: [6,0,0], thread: [31,12,0] Assertion `in_row >= 0 && in_row < input_len` failed.
```

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/feature_fetcher.py
+++ b/python/dgl/graphbolt/feature_fetcher.py
@@ -171,7 +171,6 @@ class FeatureFetcher(MiniBatchTransformer):
         if self.stream is not None:
             current_stream = torch.cuda.current_stream()
             self.stream.wait_stream(current_stream)
-            self.feature_store.record_stream(self.stream)
         with torch.cuda.stream(self.stream):
             data = self._read_data(data, current_stream)
             if self.stream is not None:
--- a/python/dgl/graphbolt/feature_fetcher.py
+++ b/python/dgl/graphbolt/feature_fetcher.py
@@ -134,6 +134,8 @@ class FeatureFetcher(MiniBatchTransformer):
                         edges = original_edge_ids.get(type_name, None)
                         if edges is None:
                             continue
+                        if edges.is_cuda:
+                            edges.record_stream(torch.cuda.current_stream())
                         for feature_name in feature_names:
                             edge_features[i][
                                 (type_name, feature_name)
@@ -143,6 +145,10 @@ class FeatureFetcher(MiniBatchTransformer):
                                 )
                             )
                 else:
+                    if original_edge_ids.is_cuda:
+                        original_edge_ids.record_stream(
+                            torch.cuda.current_stream()
+                        )
                     for feature_name in self.edge_feature_keys:
                         edge_features[i][feature_name] = record_stream(
                             self.feature_store.read(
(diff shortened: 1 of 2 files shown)


TristonC:
LGTM, Good catch to cover this corner case @mfbalin

TristonC (approved):

PR #6980, opened 2024-01-18
[GraphBolt][CUDA] Refine the multi-GPU example

### LINKED ISSUE #6978: [GraphBolt][CUDA] MultiGPU example crashes

mfbalin:
## Description
When training, we should have the same batch size for each iteration so it is a better practice to enable `drop_last`, `drop_uneven_inputs`, and `shuffle` only when `is_train=True`.

## Checklist
- [ ] Code is well-documented


examples/multigpu/graphbolt/node_classification.py
```diff
@@ -252,14 +247,13 @@ def train(
         dist.reduce(tensor=acc, dst=0)
         total_loss /= step + 1
         dist.reduce(tensor=total_loss, dst=0)
-        dist.barrier()
```
RamonZhou, line 255: Here it could also use `torch.cuda.synchronize()` to prevent the wrong accuracy issue. Or is it not necessary?

mfbalin: When you reduce a tensor and access it, it handles the sychronization automatically.

mfbalin: Also, below, I removed another barrier because reduction is a barrier itself.

mfbalin: I am not sure if the latest `torch.cuda.synchronize()` is needed or not, but having the synchronize there ensures that the different processes synch on the CPU as well, `dist.barrier()` only synchronizes the GPUs asynchronously.

RamonZhou: I think we'd better add them because it caused issues before. See #6544. Your point here is probably the exact reason that caused this.

RamonZhou (approved):

examples/multigpu/graphbolt/node_classification.py
mfbalin: We either need to add it or not. When we run reduce and then call torch.cuda.synchronize, it will wait for the reduction kernel to finish running on the GPUs. I will run this example as is and provide its outputs. If it matches the previous version, we should be good? Because I think it extra barriers after reduction calls are definitely not needed.

mfbalin:
<details><summary>Output of the code</summary>
<p>

Training with 4 gpus.
The dataset is already preprocessed.
Training...
48it [00:04, 10.66it/s]
Validating...
10it [00:00, 10.12it/s]
Epoch 00000 | Average Loss 2.3168 | Accuracy 0.7865 | Time 5.5491
48it [00:01, 28.29it/s]
Validating...
10it [00:00, 36.07it/s]
Epoch 00001 | Average Loss 0.9710 | Accuracy 0.8419 | Time 2.2026
48it [00:01, 24.33it/s]
Validating...
10it [00:00, 32.25it/s]
Epoch 00002 | Average Loss 0.7345 | Accuracy 0.8605 | Time 2.3079
48it [00:01, 28.20it/s]
Validating...
10it [00:00, 37.16it/s]
Epoch 00003 | Average Loss 0.6306 | Accuracy 0.8700 | Time 1.9945
48it [00:02, 21.81it/s]
Validating...
10it [00:00, 36.24it/s]
Epoch 00004 | Average Loss 0.5645 | Accuracy 0.8772 | Time 2.5011
48it [00:01, 28.22it/s]
Validating...
10it [00:00, 24.57it/s]
Epoch 00005 | Average Loss 0.5239 | Accuracy 0.8832 | Time 2.1315
48it [00:01, 28.49it/s]
Validating...
10it [00:00, 38.09it/s]
Epoch 00006 | Average Loss 0.4918 | Accuracy 0.8830 | Time 2.5510
48it [00:01, 27.96it/s]
Validating...
10it [00:00, 36.42it/s]
Epoch 00007 | Average Loss 0.4692 | Accuracy 0.8884 | Time 2.0129
48it [00:01, 28.27it/s]
Validating...
10it [00:00, 35.34it/s]
Epoch 00008 | Average Loss 0.4505 | Accuracy 0.8897 | Time 2.0001
48it [00:01, 28.69it/s]
Validating...
10it [00:00, 19.44it/s]
Epoch 00009 | Average Loss 0.4357 | Accuracy 0.8913 | Time 2.2071
Testing...
541it [00:18, 29.33it/s]
Test Accuracy 0.7410

</p>
</details>

examples/multigpu/graphbolt/node_classification.py
mfbalin: I posted the output. There seems to be no accuracy issue with these changes. If there is indeed an issue, we can add the barrier calls back.

mfbalin:
To double check, I printed before reduce and after reduce:
```
Training...
48it [00:03, 12.19it/s]
Validating...
10it [00:00, 15.82it/s]
0 tensor(0.7791, device='cuda:0')
1 tensor(0.7715, device='cuda:1')
3 tensor(0.7722, device='cuda:3')
2 tensor(0.7752, device='cuda:2')
Epoch 00000 | Average Loss 2.4012 | Accuracy 0.7745 | Time 4.7326
```
The average is correctly being computed.

PR #6982, opened 2024-01-19
[GraphBolt] fix random generator for shuffle among all workers

Rhett-Ying:
## Description

image

```
Test accuracy 45.8261

Epoch: 01, Loss: 2.3362, Valid accuracy: 48.41%, Time 68.9185
Training~Epoch 02: 0it [00:00, ?it/s]rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
rank None shuffled indices: tensor([404999, 305221, 507294,  ..., 354460, 285126, 516371])
Training~Epoch 02: 615it [02:25,  4.23it/s]
Evaluating the model on the validation set.
Inference: 16it [00:03,  4.22it/s]
Finish evaluating on validation set.
Epoch: 02, Loss: 1.5541, Valid accuracy: 48.49%, Time 145.3738
Training~Epoch 03: 0it [00:00, ?it/s]rank None shuffled indices: tensor([ 87977, 530066, 189900,  ..., 302993,  15348, 600035])
rank None shuffled indices: tensor([ 87977, 530066, 189900,  ..., 302993,  15348, 600035])
rank None shuffled indices: tensor([ 87977, 530066, 189900,  ..., 302993,  15348, 600035])
rank None shuffled indices: tensor([ 87977, 530066, 189900,  ..., 302993,  15348, 600035])
rank None shuffled indices: tensor([ 87977, 530066, 189900,  ..., 302993,  15348, 600035])
rank None shuffled indices: tensor([ 87977, 530066, 189900,  ..., 302993,  15348, 600035])
rank None shuffled indices: tensor([ 87977, 530066, 189900,  ..., 302993,  15348, 600035])
```
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/item_sampler.py
+++ b/python/dgl/graphbolt/item_sampler.py
@@ -115,6 +115,8 @@ class ItemShufflerAndBatcher:
     rank : int
         The rank of the current replica. Applies only when `distributed` is
         True.
+    rng : np.random.Generator
+        The random number generator to use for shuffling.
     """
 
     def __init__(
@@ -128,6 +130,7 @@ class ItemShufflerAndBatcher:
         drop_uneven_inputs: Optional[bool] = False,
         world_size: Optional[int] = 1,
         rank: Optional[int] = 0,
+        rng: Optional[np.random.Generator] = None,
     ):
         self._item_set = item_set
         self._shuffle = shuffle
@@ -142,6 +145,7 @@ class ItemShufflerAndBatcher:
         self._drop_uneven_inputs = drop_uneven_inputs
         self._num_replicas = world_size
         self._rank = rank
+        self._rng = rng
 
     def _collate_batch(self, buffer, indices, offsets=None):
         """Collate a batch from the buffer. For internal use only."""
@@ -216,7 +220,7 @@ class ItemShufflerAndBatcher:
             buffer = self._item_set[start_offset + start : start_offset + end]
             indices = torch.arange(end - start)
             if self._shuffle:
-                np.random.shuffle(indices.numpy())
+                self._rng.shuffle(indices.numpy())
             offsets = self._calculate_offsets(buffer)
             for i in range(0, len(indices), self._batch_size):
                 if output_count <= 0:
@@ -494,6 +498,7 @@ class ItemSampler(IterDataPipe):
         self._drop_uneven_inputs = False
         self._world_size = None
         self._rank = None
+        self._rng = np.random.default_rng()
 
     def _organize_items(self, data_pipe) -> None:
         # Shuffle before batch.
@@ -529,6 +534,7 @@ class ItemSampler(IterDataPipe):
 
     def __iter__(self) -> Iterator:
         if self._use_indexing:
+            seed = self._rng.integers(0, np.iinfo(np.int32).max)
             data_pipe = IterableWrapper(
                 ItemShufflerAndBatcher(
                     self._item_set,
@@ -540,6 +546,7 @@ class ItemSampler(IterDataPipe):
                     drop_uneven_inputs=self._drop_uneven_inputs,
                     world_size=self._world_size,
                     rank=self._rank,
+                    rng=np.random.default_rng(seed),
                 )
             )
         else:


mfbalin:
Is there a particular reason we use np rng and np shuffle? If we used torch, the code could be enabled to work on the GPU later.

Rhett-Ying:
> Is there a particular reason we use `np.rng` and `np.shuffle`? If we used torch, the code could be enabled to work on the GPU later.

`torch.Generator()` cannot be pickled. and is this targeted on GPU? why do we need that?

mfbalin:
> > Is there a particular reason we use `np.rng` and `np.shuffle`? If we used torch, the code could be enabled to work on the GPU later.
> 
> `torch.Generator()` cannot be pickled. and is this targeted on GPU? why do we need that?

I see. @frozenbugs was also asking about whether we should consider enabling GPU for the ItemSampler. Now I know what are the limitations. Thanks!

There is still a way to utilize torch shuffle while utilizing np.rng. We store an np.rng object and whenever we need to shuffle, we get a random number from np.rng, and seed a new torch rng. Then we shuffle with torch shuffle with torch rng.

python/dgl/graphbolt/item_sampler.py
```diff
@@ -216,7 +220,7 @@ def __iter__(self):
             buffer = self._item_set[start_offset + start : start_offset + end]
             indices = torch.arange(end - start)
             if self._shuffle:
-                np.random.shuffle(indices.numpy())
+                self._rng.shuffle(indices.numpy())
```
mfbalin, line 223:
```suggestion
                seed = self._rng.integers(low=0)
                torch_rng = torch.random.manual_seed(seed)
                permutation = torch.randperm(len(indices), generator=torch_rng)
                indices = indices[permutation]
```

mfbalin: I thought of something like this.

mfbalin: actually, since we have contiguous range, it can be improved, let me edit the suggestion above.

mfbalin: Later, we can add a device argument to this class and it can be made to work on the GPU easily, if we decide to do so.

Rhett-Ying: what's the benefit of using `torch` and `np` together?

Rhett-Ying:
I don't know whether moving `ItemSampler` to GPU is a good idea or best practice. Does `torch.DataLoader` ever think about it? Putting all workloads onto GPU may lose overlapping. The best practice may be combining CPU and GPU together instead of relying on single one of them.

python/dgl/graphbolt/item_sampler.py
mfbalin: We can store np as a member of the class since it can be pickled. We use torch for the actual operation so that it can be made to work on the GPU later if we decide to do so.

Rhett-Ying: If we really need to turn to `torch.Generator()` some day, we could refactor the `ItemSampler`.

mfbalin: We can pass device arguments to torch operations.

mfbalin: With my suggestion above, we don't need to turn to torch.Generator. We can keep the np generator as before. And this part wouldn't even need to change if we refactor this code as it will already be working via torch operations.

mfbalin:
> I don't know whether moving `ItemSampler` to GPU is a good idea or best practice. Does `torch.DataLoader` ever think about it? Putting all workloads onto GPU may lose overlapping. The best practice may be combining CPU and GPU together instead of relying on single one of them.

There is one more reason the solution I propose is better than the numpy one. If we use numpy, it will keep its own thread pool separate from torch thread pool. It is best if we use torch for all operations if we can, even if it runs on the CPU.

python/dgl/graphbolt/item_sampler.py
mfbalin: Is there any disadvantage going forward with my suggestion above? I feel that as a torch library, we should use torch for almost all operations if we can so that everything can utilize torch internals, torch memory pools etc. even if things run on the CPU.

mfbalin:
I benchmarked the two approaches and seems like numpy is faster: https://colab.research.google.com/drive/1_5M2aLPrjLcHSrnTyT6GpB0FMyRhuCbO?usp=sharing

Numpy, torch:
[2.3219032287597656, 3.6355559825897217]

python/dgl/graphbolt/item_sampler.py
mfbalin: Looks like numpy's way of shuffling is more performance on the CPU, so maybe we can do this in a later PR if we decide to run the operation on the GPU.

mfbalin:
Actually, on my machine, torch seems to be the more performant one:
numpy, torch runtimes:
[0.5068767070770264, 0.31662845611572266]

mfbalin:
The reason why we might want to move the shuffling operation to the GPU is as you can see, it starts taking very long as the itemset size grows. That is why, it might be a good idea to have GPU option. Otherwise, GPUs will idle between epochs.

mfbalin:
In the updated results above, it can be seen that `torch.randperm` is not a GPU accelerated operation.

frozenbugs (approved):

PR #6993, opened 2024-01-22
[GraphBolt][CUDA] Make nodes optional for sampling

mfbalin:
## Description
We make `nodes` parameter optional for SampleNeighbors so that when it is None, it is assumed to be sampling from the whole graph. This will be useful to implement overlap logic for GPU sampling operations. If the `nodes` is None, then we don't need extra slicing operations because the graph will have already been sliced.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/include/graphbolt/cuda_ops.h
+++ b/graphbolt/include/graphbolt/cuda_ops.h
@@ -121,7 +121,7 @@ std::tuple<torch::Tensor, torch::Tensor> IndexSelectCSCImpl(
  * on it gives the output indptr.
  */
 std::tuple<torch::Tensor, torch::Tensor> SliceCSCIndptr(
-    torch::Tensor indptr, torch::Tensor nodes);
+    torch::Tensor indptr, torch::optional<torch::Tensor> nodes);
 
 /**
  * @brief Given the compacted sub_indptr tensor, edge type tensor and
--- a/graphbolt/src/cuda/neighbor_sampler.cu
+++ b/graphbolt/src/cuda/neighbor_sampler.cu
@@ -130,16 +130,18 @@ struct SegmentEndFunc {
 };
 
 c10::intrusive_ptr<sampling::FusedSampledSubgraph> SampleNeighbors(
-    torch::Tensor indptr, torch::Tensor indices, torch::Tensor nodes,
-    const std::vector<int64_t>& fanouts, bool replace, bool layer,
-    bool return_eids, torch::optional<torch::Tensor> type_per_edge,
+    torch::Tensor indptr, torch::Tensor indices,
+    torch::optional<torch::Tensor> nodes, const std::vector<int64_t>& fanouts,
+    bool replace, bool layer, bool return_eids,
+    torch::optional<torch::Tensor> type_per_edge,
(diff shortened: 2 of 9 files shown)


graphbolt/include/graphbolt/cuda_ops.h
```diff
@@ -121,7 +121,7 @@ std::tuple<torch::Tensor, torch::Tensor> IndexSelectCSCImpl(
  * on it gives the output indptr.
  */
 std::tuple<torch::Tensor, torch::Tensor> SliceCSCIndptr(
-    torch::Tensor indptr, torch::Tensor nodes);
+    torch::Tensor indptr, torch::optional<torch::Tensor> nodes);
```
frozenbugs, line 124: Update doc?

mfbalin:
@dgl-bot

graphbolt/src/cuda/neighbor_sampler.cu
```diff
@@ -166,29 +168,44 @@ c10::intrusive_ptr<sampling::FusedSampledSubgraph> SampleNeighbors(
             DeviceReduce::Max, in_degree.data_ptr<index_t>(),
             max_in_degree.data_ptr<index_t>(), num_rows);
       }));
+  // Protect access to max_in_degree with a CUDAEvent
+  at::cuda::CUDAEvent max_in_degree_event;
+  max_in_degree_event.record();
   torch::optional<int64_t> num_edges_;
```
frozenbugs, line 174:
```suggestion
  torch::optional<int64_t> num_edges;
```

graphbolt/src/fused_csc_sampling_graph.cc
```diff
@@ -607,27 +607,37 @@ FusedCSCSamplingGraph::SampleNeighborsImpl(
...
+      !replace) {
     GRAPHBOLT_DISPATCH_CUDA_ONLY_DEVICE(
         c10::DeviceType::CUDA, "SampleNeighbors", {
           return ops::SampleNeighbors(
               indptr_, indices_, nodes, fanouts, replace, layer, return_eids,
               type_per_edge_, probs_or_mask);
         });
+  } else if (!nodes.has_value()) {
+    // Initialize the nodes tensor for the CPU execution.
+    nodes = torch::arange(NumNodes(), indices_.options());
```
frozenbugs, line 640: Shall we just throw error for cpu cases?

frozenbugs (approved):

graphbolt/src/fused_csc_sampling_graph.cc
mfbalin: That will complicate the tests a little. Would need to make a special case for the CPU.

mfbalin: I prefer keeping it this way but if you think it is better to throw an error for the CPU, I can do that before merging.

mfbalin: The disk case is not the same as the GPU. Let me throw an error.

mfbalin: Added a check for the CPU. CPU requires nodes to be passed. We can change it later if needed.

PR #7039, opened 2024-01-26
[GraphBolt][CUDA] Pipelined sampling optimization

mfbalin:
## Description

We overlap the UVA graph accesses with the rest of the computations by launching them in the UVA stream. We use a separate thread becase the `index_select_csc` is blocking, unlike `index_select`, which is nonblocking. I suspect the GIL in python is making multithreading inefficient a little bit, so more performance can be unlocked in the future after python with no GIL is available, see https://peps.python.org/pep-0703/#gpu-heavy-workloads-require-multi-core-processing

I am getting %10-%20 end-to-end speedup for `sample_layer_neighbor`. `sample_neighbor` is currently not getting much speedup because it does not need to access the whole indices tensor, which we are fetching through this optimization. Will need to specialize for `sample_neighbor` later.

This optimization will be the core of us providing the users a way to define their own custom samplers using `dgl.sparse`. My next PR will show how to implement LADIES while still taking advantage of the GraphBolt optimizations.

image

## Checklist
- [ ] Code is well-documented


graphbolt/src/cuda/index_select_csc_impl.cu
frozenbugs, line 1: split out the cu change to separate PR.

python/dgl/graphbolt/dataloader.py
```diff
@@ -226,6 +258,30 @@ def __init__(
...
+            and torch.cuda.is_available()
+        ):
+            torch.ops.graphbolt.set_max_uva_threads(max_uva_threads)
+            samplers = dp_utils.find_dps(
+                datapipe_graph,
+                SamplePerLayer,
+            )
+            executor = ThreadPoolExecutor(max_workers=1)
+            for sampler in samplers:
+                datapipe_graph = dp_utils.replace_dp(
```
frozenbugs, line 273: It is a little bit weird to see FetcherAndSampler in the dataloader.py, but not sure what's the best way to do. I am worried about we can't maintain the complexity with all this dp replacement.

Is it possible to code this in neighbor_sampler.py in some way?

mfbalin: It will lead to circular dependency in that case. I can remove FetcherAndSampler if you want and manually chain the datapipes here.

mfbalin: Pushed my changes.

graphbolt/src/cuda/index_select_csc_impl.cu
mfbalin: #7056

python/dgl/graphbolt/dataloader.py
mfbalin: Or, we can move the Waiter and Bufferer into a different file so that dataloader and the neighborsampler can import it.

mfbalin: Moved the datapipe components into base.py and the FetcherAndSampler into sample_neighbors.py.

mfbalin:
@frozenbugs I am hoping that with my most recent commits, the main concerns related to your previous reviews should be addressed now.

python/dgl/graphbolt/impl/neighbor_sampler.py
```diff
@@ -1,18 +1,125 @@
...
+        super().__init__(datapipe, self._fetch_per_layer)
+        self.graph = sample_per_layer_obj.sampler.__self__
+        self.prob_name = sample_per_layer_obj.prob_name
+        self.stream = stream
+        if executor is None:
+            self.executor = ThreadPoolExecutor(max_workers=1)
+        else:
+            self.executor = executor
+
+    def _fetch_per_layer_helper(self, minibatch, stream):
```
frozenbugs, line 37:
```suggestion
    def _fetch_per_layer_impl(self, minibatch, stream):
```

python/dgl/graphbolt/impl/neighbor_sampler.py
```diff
@@ -1,18 +1,125 @@
...
 from .sampled_subgraph_impl import SampledSubgraphImpl
 
 
-__all__ = ["NeighborSampler", "LayerNeighborSampler"]
+__all__ = ["NeighborSampler", "LayerNeighborSampler", "SamplePerLayer"]
+
+
+@functional_datapipe("fetch_insubgraph_data")
+class FetchInsubgraphData(Mapper):
+    """Fetches the insubgraph and wraps it in a FusedCSCSamplingGraph object."""
```
frozenbugs, line 23: Worth to clarify what does this function do. it fetch insubgraph neighbors for seed nodes, but for probs_or_mask are you only get for seed nodes or all insubgraph neighbors?

python/dgl/graphbolt/impl/neighbor_sampler.py
```diff
@@ -1,18 +1,125 @@
...
+                if stream is not None and tensor.is_cuda:
+                    tensor.record_stream(stream)
+
+            indptr, indices = index_select(self.graph.indices, index, None)
+            record_stream(indptr)
+            record_stream(indices)
+            output_size = len(indices)
+            if self.graph.type_per_edge is not None:
+                _, type_per_edge = index_select(
+                    self.graph.type_per_edge, index, output_size
```
frozenbugs, line 56: Is this correct to use index or indices?

python/dgl/graphbolt/impl/neighbor_sampler.py
frozenbugs: Also please add unit test for this method.

frozenbugs:
Overall LGTM, questions on python/dgl/graphbolt/impl/neighbor_sampler.py.

python/dgl/graphbolt/impl/neighbor_sampler.py
mfbalin: Index because we defined `index_select` above as:
```python
            index_select = partial(
                torch.ops.graphbolt.index_select_csc, self.graph.csc_indptr
            )
```

mfbalin: We are getting the vaue for the insubgraph neighbors of seed nodes when used like this. Do you think I can make this more clear in the code?

frozenbugs: I see, index_select is a reserved name for the specific op, reuse it here causes confusion. Consider rename it to index_select_csc_with_indptr or better ones.

frozenbugs (approved):

mfbalin:
@dgl-bot

mfbalin:
@dgl-bot

mfbalin:
@Rhett-Ying Does the CI have a problem?

mfbalin:
@dgl-bot

Rhett-Ying:
@dgl-bot

Rhett-Ying:
@dgl-bot

Rhett-Ying:
@dgl-bot

PR #7051, opened 2024-01-31
[GraphBolt][CUDA] Add `gb.index_select` and fix example inferencing.

mfbalin:
## Description
The existing example was moving the feature memory to the device memory. We change it to the storage_device instead.

When I started this PR, I thought I would use `gb.index_select` to load features for inferencing. Then I realized that I can just update the features and rely on the dataloader to fetch the features for inferencing. Otherwise, the copies for all the different modes become too complicated and ugly, which the dataloader automatically handles for us. However, I am sure that `gb.index_select` will be useful to us down the road, it is a very simple wrapper over `torch.ops.graphbolt.index_select`.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/examples/sampling/graphbolt/link_prediction.py
+++ b/examples/sampling/graphbolt/link_prediction.py
@@ -84,12 +84,6 @@ class SAGE(nn.Module):
         pin_memory = storage_device == "pinned"
         buffer_device = torch.device("cpu" if pin_memory else storage_device)
 
-        def move_to_storage_device(tensor):
-            if pin_memory:
-                return tensor.pin_memory()
-            else:
-                return tensor.to(buffer_device)
-
         print("Start node embedding inference.")
         for layer_idx, layer in enumerate(self.layers):
             is_last_layer = layer_idx == len(self.layers) - 1
@@ -99,6 +93,7 @@ class SAGE(nn.Module):
                 self.hidden_size,
                 dtype=torch.float32,
                 device=buffer_device,
+                pin_memory=pin_memory,
             )
             for data in tqdm.tqdm(dataloader):
                 # len(blocks) = 1
@@ -110,7 +105,7 @@ class SAGE(nn.Module):
                     buffer_device, non_blocking=True
                 )
             if not is_last_layer:
-                features.update("node", None, "feat", move_to_storage_device(y))
+                features.update("node", None, "feat", y)
 
         return y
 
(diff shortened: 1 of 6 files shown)


python/dgl/graphbolt/base.py
```diff
@@ -102,6 +103,33 @@ def expand_indptr(indptr, dtype=None, node_ids=None, output_size=None):
     )
 
 
+def index_select(tensor, index):
```
frozenbugs, line 106: Does this support tensor on CUDA as well?

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -144,11 +145,6 @@ def update(self, value: torch.Tensor, ids: torch.Tensor = None):
             updated.
         """
         if ids is None:
-            assert self.size() == value.size()[1:], (
```
frozenbugs, line 147: why remove this check? not related to index_select.

examples/sampling/graphbolt/link_prediction.py
frozenbugs, line 1: Will the change affects the performance of the example?

python/dgl/graphbolt/base.py
mfbalin: Yes, it supports CPU, Pinned-GPU and GPU cases all.

python/dgl/graphbolt/impl/torch_based_feature_store.py
mfbalin: This is required if we want to be able to update the stored tensor for the inferencing use case. When inferencing, the dimensions between different layers are different, without this change, we can't update the stored feature.

examples/sampling/graphbolt/link_prediction.py
mfbalin: We were previously moving the whole feature tensor to the GPU. Thus, it probably will slow down for the `cpu-cuda` case.

frozenbugs (approved):

examples/sampling/graphbolt/link_prediction.py
frozenbugs: Can we avoid that for cpu-cuda case?

mfbalin: I mean because it is cpu-cuda or pinned-cuda, we can not move everything to the GPU, that should be performed only for cuda-cuda. Thus, it is bound to be slower a little.

mfbalin:
Let's wait for #7083 to be merged first so that I can test this PR properly.

mfbalin:
Just tested all 4 different modes for the node_prediction example, all work as expected.

mfbalin:
Just tested the link_prediction example with the fix in #7083, all work as expected.

PR #7052, opened 2024-02-01
[GraphBolt] Improving `ondisk_dataset` tests.

drivanov:
## Description

The goal of this PR is to resolve the following 24 warnings that appear when running tests from `test_ondisk_dataset.py`:
```
tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py: 24 warnings
  /usr/local/lib/python3.10/dist-packages/dgl/graphbolt/impl/ondisk_dataset.py:260: DGLWarning: Edge feature is stored, but edge IDs are not saved.
    dgl_warning("Edge feature is stored, but edge IDs are not saved.")
```

To achieve this goal, I created and started using a new local function  `_on_disk_dataset(...) which uses the method already used several times in this file:
```
    with pytest.warns(
        DGLWarning,
        match="Edge feature is stored, but edge IDs are not saved.",
    ):
```

## Checklist


Rhett-Ying (changes requested):
I think it's better to explicitly capture warnings directly in the testcase. and check if we really need to enable `include_original_edge_id` or remove edge feature before adding such capture. We only capture for those where warning is inevitable.

drivanov:
@Rhett-Ying :  Sorry, I'm not sure I understand what you're talking about. We already have similar captures of that warning [here](https://github.com/dmlc/dgl/blob/master/tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py#L2097-L2103) and [here](https://github.com/dmlc/dgl/blob/master/tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py#L2986C1-L2990C71). They are generated only for `include_original_edge_id=False`. I created a generic function for these two and 24 other cases.

Note that I'm not simply suppressing this warning even though it won't be generated. With the following lines:
```
    with pytest.warns(
        DGLWarning,
        match="Edge feature is stored, but edge IDs are not saved.",
    ):
```
the program **expects** that this warning will be generated, therefore  it should be considered as  **inevitable**

Moreover, if the program **expects** to see, but won't see, this warning for some test, that test will fail with the following error message:
```
================================================================================= short test summary info =================================================================================
FAILED tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py::test_OnDiskDataset_preprocess_auto_force_preprocess - Failed: DID NOT WARN. No warnings of type (<class 'dgl.base.DGLWarning'>,) were emitted.
```

I just got it with the following changes:
```
        # 4. Change nothing.
        with pytest.warns(
            DGLWarning,
            match="Edge feature is stored, but edge IDs are not saved.",
        ):
            preprocessed_metadata_path = (
                gb.ondisk_dataset.preprocess_ondisk_dataset(
                    test_dir, include_original_edge_id=True
                )
            )
```
I made to  `test_OnDiskDataset_preprocess_auto_force_preprocess`. As you could see, here `include_original_edge_id=True`, therefore, this warning should not have been generated, and when it was not captured, this test failed.

Rhett-Ying:
@drivanov Sorry for the vagueness. What I mean is if we mean to verify such warning is thrown or inevitable, then directly wrap with `with pytest.warns()` instead of a wrapper like `_ondist_xx()` you did. For other testcases that throw warning, we just remove the edge feature or specify `include_original_eids=True` to eliminate the warnings.

drivanov:
@Rhett-Ying :  I have implemented the changes you suggested.

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
```diff
@@ -1830,7 +1835,12 @@ def test_OnDiskDataset_preprocess_yaml_content_windows():
         with open(yaml_file, "w") as f:
             f.write(yaml_content)
 
-        preprocessed_metadata_path = gb.preprocess_ondisk_dataset(test_dir)
+        with pytest.warns(
+            DGLWarning,
+            match="Edge feature is stored, but edge IDs are not saved.",
+        ):
+            preprocessed_metadata_path = gb.preprocess_ondisk_dataset(test_dir)
```
Rhett-Ying, line 1842: how about either specifying `include_original_edge_id=True` or remove the edge feature instead of capturing the warning? As edge feature are usually added, I prefer `include_original_edge_id=True`. From my understanding, we just need to capture warning for the tests that disable the `include_xx`.

Rhett-Ying:
@dgl-bot

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
```diff
@@ -2158,7 +2169,11 @@ def test_OnDiskDataset_load_feature(edge_fmt):
             f.write(yaml_content)
 
         # Case1. Test modify the `in_memory` field.
-        dataset = gb.OnDiskDataset(test_dir).load()
+        with pytest.warns(
+            DGLWarning,
+            match="Edge feature is stored, but edge IDs are not saved.",
+        ):
+            dataset = gb.OnDiskDataset(test_dir).load()
```
Rhett-Ying, line 2176: add `, include_original_edge_id=True` works well on my side.

Rhett-Ying, line 2176: why not adding `include_original_edge_id=True`? please note there are 2 calls of `gb.OnDiskDataset()` in this testcase.

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
```diff
@@ -2253,7 +2268,11 @@ def test_OnDiskDataset_load_graph(edge_fmt):
         )
 
         # Case1. Test modify the `type` field.
-        dataset = gb.OnDiskDataset(test_dir)
+        with pytest.warns(
+            DGLWarning,
+            match="Edge feature is stored, but edge IDs are not saved.",
+        ):
+            dataset = gb.OnDiskDataset(test_dir)
```
Rhett-Ying, line 2275: ditto

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
```diff
@@ -2346,7 +2369,11 @@ def test_OnDiskDataset_load_tasks(edge_fmt):
             f.write(yaml_content)
 
         # Case1. Test modify the `name` field.
-        dataset = gb.OnDiskDataset(test_dir)
+        with pytest.warns(
+            DGLWarning,
+            match="Edge feature is stored, but edge IDs are not saved.",
+        ):
+            dataset = gb.OnDiskDataset(test_dir)
```
Rhett-Ying, line 2376: `, include_original_edge_id=True`?

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
```diff
@@ -2636,10 +2663,16 @@ def test_OnDiskDataset_homogeneous(include_original_edge_id, edge_fmt):
         with open(yaml_file, "w") as f:
             f.write(yaml_content)
 
-        dataset = gb.OnDiskDataset(
-            test_dir, include_original_edge_id=include_original_edge_id
-        ).load()
+        if include_original_edge_id:
```
Rhett-Ying, line 2666: I think it's ok to wrap directly instead of `if-else`

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
drivanov: In fact, this test contains not 2, but 5 calls of `gb.OnDiskDataset()`, and one way to resolve these warnings is to use `include_original_edge_id=True` for all of them. The other way would be the use 
```
        with pytest.warns(
            DGLWarning,
            match="Edge feature is stored, but edge IDs are not saved.",
        ):
            dataset = gb.OnDiskDataset(test_dir).load()
```
only on the first such call.

We have similar situations in
```
# of calls               TestName:
    5                test_OnDiskDataset_load_feature                                       - 
    5                test_OnDiskDataset_load_graph                                    
    6                test_OnDiskDataset_load_tasks
    6                test_OnDiskDataset_load_tasks_selectively
```
Do you want me to use `include_original_edge_id=True` everywhere?

Rhett-Ying: use include_original_edge_id=True everywhere

drivanov: DONE

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
drivanov: DONE

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
```diff
@@ -3067,7 +3104,11 @@ def test_OnDiskDataset_load_tasks_selectively():
             f.write(yaml_content)
 
         # Case1. Test load all tasks.
-        dataset = gb.OnDiskDataset(test_dir).load()
+        with pytest.warns(
+            DGLWarning,
+            match="Edge feature is stored, but edge IDs are not saved.",
+        ):
+            dataset = gb.OnDiskDataset(test_dir).load()
```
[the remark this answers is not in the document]
drivanov: DONE

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
drivanov: DONE

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
drivanov: DONE

tests/python/pytorch/graphbolt/impl/test_ondisk_dataset.py
drivanov: Done

Rhett-Ying:
@dgl-bot

Rhett-Ying (approved):

Rhett-Ying:
this PR could be abandoned if https://github.com/dmlc/dgl/pull/7110# is merged.

PR #7073, opened 2024-02-02
[GraphBolt][CUDA] GPUCache performance fix.

### LINKED ISSUE #6955: [GraphBolt][CUDA] Showcase `GPUCachedFeature` usage in an example.

mfbalin:
## Description
The missing_len argument to the GPU cache needs to be on the GPU because the cache performs atomic operations on it. Results in a 1000x performance improvement. Before this change, it turns out that the GPUCache was practically not usable at all.

Testable with #7074, @TristonC could you test it on multiGPU machines with `--gpu-cache-size=1000000`, either before or after this PR is merged?

When the GPUCachedFeature is used, the feature fetching operation has a GPU synchronization. We might want to run the feature fetching operation in another thread. Hoping that the main thread will make progress while the feature fetching thread is waiting for synchronization. However, if the GIL is not released while waiting, then it may not do what we expect, which will pose a problem, and the overlap optimization might not work as expected.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/examples/multigpu/graphbolt/node_classification.py
+++ b/examples/multigpu/graphbolt/node_classification.py
@@ -284,6 +284,10 @@ def run(rank, world_size, args, devices, dataset):
     hidden_size = 256
     out_size = num_classes
 
+    if args.gpu_cache_size > 0:
+        input_feature = dataset.feature._features[("node", None, "feat")]
+        dataset.feature._features[("node", None, "feat")] = gb.GPUCachedFeature(input_feature)
+
     # Create GraphSAGE model. It should be copied onto a GPU as a replica.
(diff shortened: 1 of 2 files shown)


mfbalin:
@TristonC

TristonC:
Will test it ASAP

TristonC (commented):
With both PR 7073 and 7074, we saw good performance on an 8GPU runs. On the same DGX1V, per epoch time dropped from ~1.8 seconds to less than 1.2 seconds with --gpu-cache-size=1000000. Well done.

TristonC (approved):

PR #7088, opened 2024-02-05
[GraphBolt][CUDA] Pipelined sampling accuracy fix

mfbalin:
## Description
There was a bug related to the permutation of the vertices and recording the inputs to the graph fetching datapipe. This PR fixes these issues.

Now, enabling graph fetch in the dataloader or not enabling it gives the identical accuracies.

I would like to modify the node_classification example to add the overlap_graph_fetch option so that we can run the regression test for this case as well.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/tests/python/pytorch/graphbolt/impl/test_neighbor_sampler.py
+++ b/tests/python/pytorch/graphbolt/impl/test_neighbor_sampler.py
@@ -41,8 +41,12 @@ def get_hetero_graph():
 @unittest.skipIf(F._default_context_str != "gpu", reason="Enabled only on GPU.")
 @pytest.mark.parametrize("hetero", [False, True])
 @pytest.mark.parametrize("prob_name", [None, "weight", "mask"])
-def test_NeighborSampler_GraphFetch(hetero, prob_name):
-    items = torch.arange(3)
+@pytest.mark.parametrize("sorted", [False, True])
+def test_NeighborSampler_GraphFetch(hetero, prob_name, sorted):
+    if sorted:
+        items = torch.arange(3)
+    else:
+        items = torch.tensor([2, 0, 1])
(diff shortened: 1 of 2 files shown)


tests/python/pytorch/graphbolt/impl/test_neighbor_sampler.py
```diff
@@ -41,8 +41,12 @@ def get_hetero_graph():
...
 @pytest.mark.parametrize("hetero", [False, True])
 @pytest.mark.parametrize("prob_name", [None, "weight", "mask"])
-def test_NeighborSampler_GraphFetch(hetero, prob_name):
-    items = torch.arange(3)
+@pytest.mark.parametrize("sorted", [False, True])
+def test_NeighborSampler_GraphFetch(hetero, prob_name, sorted):
+    if sorted:
+        items = torch.arange(3)
+    else:
+        items = torch.tensor([2, 0, 1])
```
frozenbugs, line 49: I am a little bit confused by the unit test case, does it suggest that regardless whether the items are sorted or not, the result will be the same?

frozenbugs (approved):

tests/python/pytorch/graphbolt/impl/test_neighbor_sampler.py
mfbalin: Yes, however, the code path may be a little different. Sorting is required when using hetero graphs because the extracted InSubgraph needs to have the node_type_offset field. When nodes are not sorted, it is not possible to assign a value to that field.

mfbalin: Whether the nodes are sorted or not, the output should match with the nonpipelined version.

PR #7047, opened 2024-01-30
[GraphBolt] Improving `subgraph_sampler` tests.

drivanov:
## Description

The main goal of this PR is to resolve the following 13 warnings that appear when running tests from `test_subgraph_sampler.py`:
```
tests/python/pytorch/graphbolt/test_subgraph_sampler.py: 13 warnings
  /usr/local/lib/python3.10/dist-packages/dgl/graphbolt/item_sampler.py:67: DGLWarning: Unknown item name 'timestamp' is detected and added into `MiniBatch`. You probably need to provide a customized `MiniBatcher`.
    dgl_warning(
```
Some improvements have also been made to eliminate code duplication.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [x] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [x] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [x] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

PR #7110, opened 2024-02-09
[Misc] Suppress dgl warning in unit tests.

frozenbugs:
## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/base.py
+++ b/python/dgl/base.py
@@ -53,7 +53,7 @@ def dgl_warning(message, category=DGLWarning, stacklevel=2):
     """DGL warning wrapper that defaults to ``DGLWarning`` instead of ``UserWarning`` category."""
     # Skip warning in unit test.
     if "unittest" in sys.modules.keys():
-        return
+        return None
     return warnings.warn(message, category=category, stacklevel=stacklevel)
 
 
--- a/python/dgl/base.py
+++ b/python/dgl/base.py
@@ -1,6 +1,7 @@
 """Module for base types and utilities."""
 from __future__ import absolute_import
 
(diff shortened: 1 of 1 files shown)


python/dgl/base.py
```diff
@@ -50,6 +51,9 @@ def dgl_warning_format(message, category, filename, lineno, line=None):
 
 def dgl_warning(message, category=DGLWarning, stacklevel=2):
     """DGL warning wrapper that defaults to ``DGLWarning`` instead of ``UserWarning`` category."""
+    # Skip warning in unit test.
+    if "unittest" in sys.modules.keys():
```
Rhett-Ying, line 55: skip all warnings is too mighty...

Rhett-Ying (approved):

python/dgl/base.py
frozenbugs: The problem is "did we ever care about warnings in unit test", I can move this to the item_sampler.py related to https://github.com/dmlc/dgl/pull/7047, and we can see whether the warning in unit test is helpful or just more trouble.

Rhett-Ying: I think we should care about warnings even in unit test in default as it reminds us we may call APIs in not-recommended way. For those we really don't care about it, we'd better capture warning explicitly right in the testcase which indicates: we know warning is thrown and we don't care about it. As the `timestamp` warning in #7047, I am wondering should we add it into canonical attribute list to suppress the warning.

mfbalin: I caught some bugs thanks to the warnings in the tests.

PR #7117, opened 2024-02-13
Skip test when atomic operations are not supported on GPU.

drivanov:
## Description

Following tests 
```
FAILED tests/python/common/ops/test_ops.py::test_gather_mm_idx_b[dtype0-0.01-256] - RuntimeError: CUDA error: unspecified launch failure
FAILED tests/python/common/ops/test_ops.py::test_gather_mm_idx_b[dtype0-0.01-64] - RuntimeError: CUDA error: unspecified launch failure
FAILED tests/python/common/ops/test_ops.py::test_gather_mm_idx_b[dtype0-0.01-16] - RuntimeError: CUDA error: unspecified launch failure
FAILED tests/python/common/ops/test_ops.py::test_gather_mm_idx_b[dtype0-0.01-8] - RuntimeError: CUDA error: unspecified launch failure
FAILED tests/python/common/ops/test_ops.py::test_gather_mm_idx_b[dtype0-0.01-1] - RuntimeError: CUDA error: unspecified launch failure
```
are failing for `dtype=torch.bfloat16` with the error message:
```
Atomic operations are not supported for bfloat16 (BF16) on GPUs with compute capability less than 8.0.
```

## Checklist

### DIFF
--- a/tests/python/common/ops/test_ops.py
+++ b/tests/python/common/ops/test_ops.py
@@ -407,22 +407,19 @@ def test_segment_mm(idtype, feat_size, dtype, tol):
 def test_gather_mm_idx_b(feat_size, dtype, tol):
     if F._default_context_str == "cpu" and dtype == torch.float16:
         pytest.skip("float16 is not supported on CPU.")
-    if (
-        F._default_context_str == "gpu"
-        and dtype == torch.bfloat16
-        and not torch.cuda.is_bf16_supported()
-    ):
-        pytest.skip("BF16 is not supported.")
-    if F._default_context_str == "gpu" and (
-        dtype == torch.float16
-        and torch.cuda.get_device_capability() < (7, 0)
-        or dtype == torch.bfloat16
-        and torch.cuda.get_device_capability() < (8, 0)
-    ):
-        pytest.skip(
-            f"{dtype} is not supported for atomic operations on GPU with "
-            f"cuda capability ({torch.cuda.get_device_capability()})."
-        )
+
+    if  F._default_context_str == "gpu":
+        if dtype == torch.bfloat16 and not torch.cuda.is_bf16_supported():
+            pytest.skip("BF16 is not supported.")
+
+        device_capability = torch.cuda.get_device_capability()
+        if (dtype == torch.float16 and  device_capability < (7, 0) or
+            dtype == torch.bfloat16 and device_capability < (8, 0)
+        ):
+            pytest.skip(
+                f"{dtype} is not supported for atomic operations on GPU with "
+                f"cuda capability ({torch.cuda.get_device_capability()})."
+            )
 
     dev = F.ctx()
     # input
--- a/tests/python/common/ops/test_ops.py
+++ b/tests/python/common/ops/test_ops.py
@@ -413,13 +413,11 @@ def test_gather_mm_idx_b(feat_size, dtype, tol):
         and not torch.cuda.is_bf16_supported()
     ):
         pytest.skip("BF16 is not supported.")
-    if (
-        F._default_context_str == "gpu"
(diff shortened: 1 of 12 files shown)


chang-l:
Btw, let me provide a bit more background here: 

Recently, PyTorch introduces this commit: https://github.com/pytorch/pytorch/commit/fc5fda14bcc41008e5610bf39c53eb066933ea4e, which changes the behavior of `torch.cuda.is_bf16_supported()`. Previously, `torch.cuda.is_bf16_supported()` will be `False` for any GPUs older than A100 (<sm80). But after this commit, it could return `True` even with V100 GPUs (sm70).  As a result, it can bypass the check here even with a V100 (sm70) GPU: https://github.com/dmlc/dgl/blob/8e6cbd621f34c0892ab5adac5259cd61401a6d2b/tests/python/common/ops/test_ops.py#L410-L415
Sadly, even if we know that A100 is the first gen GPU that supports bf16 operations [[1]](https://en.wikipedia.org/wiki/Ampere_(microarchitecture)) [[2]](https://images.nvidia.com/aem-dam/en-zz/Solutions/data-center/nvidia-ampere-architecture-whitepaper.pdf), we could not use this API(`torch.cuda.is_bf16_supported()`) to query for the bf16 operation support any more.  cc. @TristonC @nv-dlasalle @frozenbugs 

As of now, this commit has not been populated to any pytorch release branch yet. So, it can only be reproduced from a nightly PyTorch build.

nv-dlasalle:
@chang-l to clarify the issue, because DGL has some operations only implemented via atomics, we cannot support these GPUs, but pytorch has non-atomic versions of its operators and thus can implement operations for BF16 on these GPUs?

chang-l:
@nv-dlasalle 
After checking `cuda_bf16.h` (12.3) and the [doc](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH____BFLOAT162__ARITHMETIC.html#group__CUDA__MATH____BFLOAT162__ARITHMETIC_1g550f52c89d672213390e9bfd8a3c42bf), it seems starting from cuda12.2, an 'emulated' support of many bf16 operations have been added to older devices (sm70), which includes atomicAdd operation. I think, technically, all sm70-sm80 devices should support bf16 operations via 'emulation path' from driver side (with cuda 12.2+). 

Then, do we need to update https://github.com/dmlc/dgl/blob/8e6cbd621f34c0892ab5adac5259cd61401a6d2b/src/array/cuda/atomic.cuh to remove the assertions?  cc. @yaox12

frozenbugs:
@dgl-bot

Rhett-Ying:
@dgl-bot

tests/python/common/ops/test_ops.py
```diff
@@ -413,14 +413,14 @@ def test_gather_mm_idx_b(feat_size, dtype, tol):
...
         pytest.skip("BF16 is not supported.")
-
-    if (
-        F._default_context_str == "gpu"
-        and dtype == torch.float16
+    if F._default_context_str == "gpu" and (
+        dtype == torch.float16
         and torch.cuda.get_device_capability() < (7, 0)
+        or dtype == torch.bfloat16
+        and torch.cuda.get_device_capability() < (8, 0)
```
frozenbugs, line 420: Sorry I am not good at understanding the priority of `or` and `and`, can you add parentheses?

chang-l:
I am okay with the current fix in this PR.

chang-l (approved):

chang-l:
@dgl-bot

tests/python/common/ops/test_ops.py
drivanov: Done

frozenbugs (approved):

Issue #7005, opened 2024-01-24
[GraphBolt][CUDA] `cpu-cuda` optimization


mfbalin:
## 🔨Work Item

**IMPORTANT:**
* This template is only for dev team to track project progress. For feature request or bug report, please use the corresponding issue templates.
* DO NOT create a new work item if the purpose is to fix an existing issue or feature request. We will directly use the issue in the project tracker.

Project tracker: https://github.com/orgs/dmlc/projects/2

## Description

We should consider moving the `copy_to` operation before the feature fetch stage so that the overlap optimization is enabled for the `cpu-cuda` mode as well. However, the features need to be pinned for this to work. On my machine, I got a 2.2x speedup by doing so.

```python
if args.storage_device == "cpu":
    datapipe = datapipe.copy_to(device=device, extra_attrs=["input_nodes"])
```

@Rhett-Ying @frozenbugs

mfbalin:
The main examples take two parameters, storage device and device. To take advantage of this optimization, we need to move the features to the pinned memory even if the storage device is cpu.

Should we add one more parameter to the `args.mode` that denotes the feature_Storage? Then, we would have the following modes:

graph, features, device: `cpu-cpu-cpu`, `cpu-cpu-cuda`, `cpu-pinned-cuda`, `pinned-pinned-cuda`, `cuda-pinned-cuda`, `cuda-cuda-cuda`.

@frozenbugs @Rhett-Ying

mfbalin:
Also, for each different mode of graph and feature storages, we need copy_to inserted. This makes the examples quite bulky. We could probably insert it only at the end and let the dataloader move it further up.

PR #7123, opened 2024-02-19
[GraphBolt][PyG] Modify PyG example with `to_pyg_data`

RamonZhou:
## Description
Apply `to_pyg_data` in the example and implement layer-wise inference.

For one train epoch:
- [PyG original example](https://github.com/pyg-team/pytorch_geometric/blob/25b2f208e671eeec285bfafa2e246ea0a234b312/examples/ogbn_products_sage.py) costs 49s
- Our example costs 50s

For inference:
- PyG original example costs 1m51s
- Out example costs 1m28s

Acc after 10 epoch:
- PyG original example 0.7807
- Our example 0.7808

## Checklist
- [ ] Code is well-documented


examples/sampling/pyg/node_classification.py
```diff
@@ -144,35 +167,37 @@ def train(model, dataloader, optimizer, criterion, device, num_classes):
     num_batches = 0  # Counter for the number of mini-batches processed
 
     for minibatch in dataloader:
-        node_features = minibatch.node_features["feat"]
-        labels = minibatch.labels
+        pyg_data = minibatch.to_pyg_data()
```
Rhett-Ying, line 170: add a few comments here about converting GB.MiniBatch to pyg ones. Just highlight the key calls.

Rhett-Ying (approved):

examples/sampling/pyg/node_classification.py
```diff
@@ -75,19 +75,35 @@ def __init__(self, in_size, hidden_size, out_size):
...
-        h = x
-        for i, (layer, block) in enumerate(zip(self.layers, blocks)):
-            src, dst = block.edges()
-            edge_index = torch.stack([src, dst], dim=0)
-            h_src, h_dst = h, h[: block.number_of_dst_nodes()]
-            h = layer((h_src, h_dst), edge_index)
-            if i != len(blocks) - 1:
-                h = F.relu(h)
-        return h
+    def forward(self, x, edge_index):
```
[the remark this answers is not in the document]
RamonZhou: The logic is the same as the example from PyG but I rewrote the code. I can add a link in the comment.

examples/sampling/pyg/node_classification.py
```diff
@@ -120,7 +145,7 @@ def create_dataloader(dataset_set, graph, feature, device, is_train):
     return dataloader
 
 
-def train(model, dataloader, optimizer, criterion, device, num_classes):
+def train(model, dataloader, optimizer, num_classes):
     #####################################################################
```
frozenbugs, line 149: Too many unnecessary HIGHLIGHT, the only highlight in this example is to "highlight how to adopt GB with PYG"

frozenbugs:
Did you test the performance compare to original pyg example?

RamonZhou:
> Did you test the performance compare to original pyg example?

I have updated in the description

mfbalin (commented):
LGTM overall. What is the performance like compared to the previous version of this example? @RamonZhou

mfbalin:
What is the blocker for this PR?

mfbalin:
There were some changes in master, we should update the branch and resolve any potential issue w.r.t. dtype use of int32.

mfbalin:
It might help improve end-to-end performance if we move the `copy_to` operation so that it is before `fetch_feature` but after `sample_neighbor` as discussed in #7005.

Another thing is to imitate the inference function in our official node_classification example so that we use the dataloader's functionality to load the features. Below, we update the featurestore to be the next layer's features.
https://github.com/dmlc/dgl/blob/dbafbe41599b48a66e14dba4589ad257d32155ac/examples/sampling/graphbolt/node_classification.py#L199-L227

mfbalin:
Proposed modification to to_pyg_data.
```python
    def to_pyg_data(self):
        """Construct a PyG Data from `MiniBatch`. This function only supports
        node classification task on a homogeneous graph and the number of
        features cannot be more than one.
        """
        from torch_geometric.data import Data

        if self.sampled_subgraphs is None:
            edge_index = None
        else:
            col_nodes = []
            row_nodes = []
            prev_num_edges = 0
            for subgraph in reversed(self.sampled_subgraphs):
                if subgraph is None:
                    continue
                sampled_csc = subgraph.sampled_csc
                indptr = sampled_csc.indptr
                indices = sampled_csc.indices
                expanded_indptr = expand_indptr(
                    indptr, dtype=indices.dtype, output_size=len(indices)
                )
                # Drop the edges sampled for dst nodes in later layers like PyG.
                my_slice = slice(prev_num_edges, expanded_indptr.size(0))
                prev_num_edges = indices.size(0)
                col_nodes.append(expanded_indptr[my_slice])
                row_nodes.append(indices[my_slice])
            col_nodes = torch.cat(col_nodes)
            row_nodes = torch.cat(row_nodes)
            edge_index = torch.stack((row_nodes, col_nodes))
```

RamonZhou:
> @RamonZhou It might help improve end-to-end performance if we move the `copy_to` operation so that it is before `fetch_feature` but after `sample_neighbor` as discussed in #7005. We will need to pin the features in this case. I see close to 2x speedup just by doing this.
> 
> Another thing is to imitate the inference function in our official node_classification example so that we use the dataloader's functionality to load the features. Below, we update the featurestore to be the next layer's features, and no manual feature fetch indexing operation needed.
> 
> https://github.com/dmlc/dgl/blob/dbafbe41599b48a66e14dba4589ad257d32155ac/examples/sampling/graphbolt/node_classification.py#L199-L227

This suggestion can probably improve the inference performance but here we hope not to modify the PyG model too much because if PyG users want to use our GB dataloader, they won't want to do much modification to their original model.

mfbalin:
PyG original example uses different fanouts compared to this example:
https://github.com/pyg-team/pytorch_geometric/blob/25b2f208e671eeec285bfafa2e246ea0a234b312/examples/ogbn_products_sage.py#L20-L28

examples/sampling/pyg/node_classification.py
```diff
@@ -182,52 +182,96 @@ def evaluate(model, dataloader, device, num_classes):
...
     num_classes = dataset.tasks[0].metadata["num_classes"]
-    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
 
     train_dataloader = create_dataloader(
-        train_set, graph, feature, device, is_train=True
+        train_set,
+        graph,
+        feature,
+        args.batch_size,
+        [10, 10, 10],
```
mfbalin, line 237:
```suggestion
        [15, 10, 5],
```
Should we use same fanout values as the PyG original example?
https://github.com/pyg-team/pytorch_geometric/blob/25b2f208e671eeec285bfafa2e246ea0a234b312/examples/ogbn_products_sage.py#L20-L28

examples/sampling/pyg/node_classification.py
```diff
@@ -182,52 +182,96 @@ def evaluate(model, dataloader, device, num_classes):
...
+        device,
+        job="train",
     )
     valid_dataloader = create_dataloader(
-        valid_set, graph, feature, device, is_train=False
+        valid_set,
+        graph,
+        feature,
+        args.batch_size,
+        [10, 10, 10],
```
mfbalin, line 246:
```suggestion
        [15, 10, 5],
```
Same as above.

mfbalin:
If we are moving features to the GPU, then why not do the same for the graph as well as usually, the graph takes even less space than the features? @frozenbugs @Rhett-Ying

frozenbugs (approved):

Rhett-Ying:
> If we are moving features to the GPU, then why not do the same for the graph as well; as usually, the graph takes even less space than the features? @frozenbugs @Rhett-Ying

Leaving graph on CPU is consistent with PyG canonical example?

mfbalin:
> > If we are moving features to the GPU, then why not do the same for the graph as well; as usually, the graph takes even less space than the features? @frozenbugs @Rhett-Ying
> 
> Leaving graph on CPU is consistent with PyG canonical example?

Then we could still leave it on CPU and pin its memory? I also tested and saw no performance difference between pinning features vs moving them to the GPU. @RamonZhou what performance difference do you see between pinning features and moving them to the GPU?

examples/sampling/pyg/node_classification.py
mfbalin: I am not sure if the ordering of fanouts is the same as PyG's though.

examples/sampling/pyg/node_classification.py
```diff
@@ -75,104 +80,100 @@ def __init__(self, in_size, hidden_size, out_size):
...
 
+    def inference(self, args, dataloader, x_all, device):
+        """Conduct layer-wise inference to get all the node embeddings."""
+        for i, layer in tqdm(enumerate(self.layers), "inference"):
+            xs = []
+            for minibatch in dataloader:
+                # Call `to_pyg_data` to convert GB Minibatch to PyG Data.
+                pyg_data = minibatch.to_pyg_data()
+                n_ids = minibatch.node_ids().to("cpu")
+                x = x_all[n_ids].to(device)
```
mfbalin, line 99:
```suggestion
                x = gb.index_select(x_all, minibatch.node_ids())
```
Works from GPU when `x_all` is pinned and node ids are on GPU.

mfbalin: However, in our case, because we use the read function, `x_all` is probably on the GPU?

examples/sampling/pyg/node_classification.py
```diff
@@ -182,52 +183,96 @@ def evaluate(model, dataloader, device, num_classes):
...
+        "--batch-size", type=int, default=1024, help="Batch size for training."
+    )
     args = parser.parse_args()
+
+    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
     dataset_name = args.dataset
     dataset = gb.BuiltinDataset(dataset_name).load()
-    graph = dataset.graph
-    feature = dataset.feature
+    graph = dataset.graph.pin_memory_()
```
mfbalin, line 225: If we are going to pin the graph, we should move the copy_to so that it is before sample_neighbor. Otherwise, there is no need to pin the graph.

examples/sampling/pyg/node_classification.py
RamonZhou: Here we hope to use the same inference function as the PyG example with least modification. In their example, `x_all` is on the CPU

mfbalin: Got it. We might want to get the initial `x_all` in a different way then because it moves the whole feature tensor to the GPU by default when features are pinned.

mfbalin: We might want to directly access the stored tensor.

RamonZhou: I think we can put this for later since it's not a blocker.

PR #7128, opened 2024-02-20
[Graphbolt] Disk feature fetcher

xiangyuzhi:
## Description
Add basic disk feature store based on io_uring and replace original mmap based method.
For more benchmark details, pls see the [doc](https://quip-amazon.com/cjn5AXrarlYa/Disk-Feature-Store-and-Fetcher). 

## Checklist
- [ ] Code is well-documented

## Changes


python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -221,9 +303,8 @@ def __init__(self, feat_data: List[OnDiskFeatureData]):
                     torch.load(spec.path), metadata=metadata
                 )
             elif spec.format == "numpy":
-                mmap_mode = "r+" if not spec.in_memory else None
-                features[key] = TorchBasedFeature(
-                    torch.as_tensor(np.load(spec.path, mmap_mode=mmap_mode)),
+                features[key] = DiskBasedFeature(
```
frozenbugs, line 306: Do we already have unit test to cover this case?

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,76 @@
...
+  }
+  /**
+   * @brief Parse numpy meta data.
+   */
+  void parse_npy_header(FILE *fp);
+
+  /**
+   * @brief Get the feature shape of numpy data according to meta data.
+   */
+  torch::Tensor feature_size() { return feature_shape; }
```
frozenbugs, line 57: shape or size, make it consistent.

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * @brief Read disk numpy file based on given index and transform to
+   tensor.
```
frozenbugs, line 61:
```suggestion
   * @brief Read disk numpy file based on given index and transform to
   * tensor.
```

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,76 @@
...
+   */
+  torch::Tensor index_select_iouring(torch::Tensor idx);
+
+ private:
+  char *data_holder;
+  std::string filename;
+  std::vector<int64_t> shape;
+  size_t word_size;
```
frozenbugs, line 69: what's word_size?

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,76 @@
...
+  torch::Tensor feature_size() { return feature_shape; }
+
+  /**
+   * @brief Read disk numpy file based on given index and transform to
+   tensor.
+   */
+  torch::Tensor index_select_iouring(torch::Tensor idx);
+
+ private:
+  char *data_holder;
```
frozenbugs, line 66: I did not see any initialization of this field, can you check? also add comment, what's data_holder?

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,76 @@
...
+  std::string filename;
```
frozenbugs, line 67: class variable ends with _, e.g. filename_, such that you can rename the method parameter from _filename to filename

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,76 @@
...
+  std::vector<int64_t> shape;
```
frozenbugs, line 68: There are a lot of shapes, can you clarify?

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,76 @@
...
+   */
+  torch::Tensor feature_size() { return feature_shape; }
+
+  /**
+   * @brief Read disk numpy file based on given index and transform to
+   tensor.
+   */
+  torch::Tensor index_select_iouring(torch::Tensor idx);
+
+ private:
```
frozenbugs, line 65: please add comment for all variables.

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,192 @@
...
+ * @brief Read disk numpy file based on given index and transform to tensor.
+ */
+torch::Tensor OnDiskNpyArray::index_select_iouring(torch::Tensor idx) {
+  int feature_fd = open(filename.c_str(), O_RDONLY | O_DIRECT);
+
+  int64_t feature_size = feat_dim * word_size;
+  int64_t num_idx = idx.numel();
+
+  // Multi-thread number.
+  int numthd = 4;
```
peizhou001, line 36: numthd -> num_threads

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,192 @@
...
+  omp_set_num_threads(numthd);
+  // Each group size for reader.
+  int64_t group_size = 512;
+
+  char *read_buffer =
+      (char *)aligned_alloc(ALIGNMENT, ALIGNMENT * 2 * group_size * numthd);
+  char *result_buffer =
+      (char *)aligned_alloc(ALIGNMENT, feature_size * num_idx);
+
+  auto idx_data = idx.data_ptr<int64_t>();
```
peizhou001, line 46: Data dtype could also be `int32`,  add dispatch here.
And also, please add corresponding tests for it.

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,192 @@
...
+
+  // Init io_uring fetcher.
+  struct io_uring ring[numthd];
+  for (int64_t t = 0; t < numthd; t++) {
+    io_uring_queue_init(512, &ring[t], 0);
```
peizhou001, line 51: Use static member for constant.

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  // Record the inside offsets of feteched features.
+  int64_t residual[group_size * numthd];
+
+#pragma omp parallel for num_threads(numthd)
```
peizhou001, line 57: Why not use torch parallel?

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,192 @@
...
+          sqe, feature_fd,
+          read_buffer + (ALIGNMENT * 2 * group_size * thd_id) +
+              (ALIGNMENT * 2 * m),
+          read_size, aligned_offset);
+    }
+    // Submit I/O requests.
+    io_uring_submit(&ring[thd_id]);
+
+    // Wait for completion of I/O requests.
+    int64_t finish_num = 0;
```
peizhou001, line 87: `num_finish` for consistent

graphbolt/src/cnumpy.cc
xiangyuzhi: I have not used it yet, can you show me an example?

graphbolt/src/cnumpy.cc
xiangyuzhi: We use unified index data type here https://github.com/dmlc/dgl/pull/7128/files#diff-145c291ca0abd8d19cee18d755fcbe7d9886c5f018fe7528b31b4f37e7459004R28 . Maybe there is no need for type check :)

python/dgl/graphbolt/impl/torch_based_feature_store.py
xiangyuzhi: Added it.

graphbolt/src/cnumpy.cc
mfbalin: If we use torch parallel, we need a way to limit the number of threads used later for performance. We will want to have the rest of the threads running other ops.

examples/sampling/graphbolt/rgcn/hetero_rgcn.py
```diff
@@ -554,7 +554,11 @@ def train(
 
 def main(args):
     device = torch.device(
-        "cuda" if args.num_gpus > 0 and torch.cuda.is_available() else "cpu"
+        "cuda"
+        if args.num_gpus > 0
+        and torch.cuda.is_available()
+        and args.dataset != "ogb-lsc-mag240m"  # Large dataset cannot be pinned.
```
mfbalin, line 560: It can be pinned on my machine, how about we make it optional to enable disk feature fetching?

graphbolt/src/index_select.cc
```diff
@@ -21,6 +22,25 @@ torch::Tensor IndexSelect(torch::Tensor input, torch::Tensor index) {
   return input.index({index.to(torch::kLong)});
 }
 
+torch::Tensor DiskIndexSelect(
+    std::string path, torch::Tensor index, torch::ScalarType dtype) {
+#ifdef __linux__
+  storage::OnDiskNpyArray arr(path);
```
mfbalin, line 28: What is the overhead of opening the file everytime DiskIndexSelect is called?

mfbalin: We could keep an open handle instead of working with paths.

graphbolt/src/index_select.cc
```diff
@@ -21,6 +22,25 @@ torch::Tensor IndexSelect(torch::Tensor input, torch::Tensor index) {
...
+  return arr.index_select_iouring({index.to(torch::kLong)}, dtype);
+#else
+  return torch::empty({0});
+#endif
+}
+
+torch::Tensor DiskFeatureShape(std::string path) {
```
mfbalin, line 35: If we keep an open handle, then we could cache the size of the tensor instead of rereading it everytime.

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -12,7 +13,168 @@
...
+        Returns
+        -------
+        Dict
+            The metadata of the feature.
+        """
+        return (
+            self._metadata if self._metadata is not None else super().metadata()
+        )
+
+    def pin_memory_(self):
```
mfbalin, line 119: We are not going to support pinning DiskBasedFeature so we can remove these sections.

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -12,7 +13,168 @@
...
+            self._is_inplace_pinned.add(x)
+            self._inplace_unpinner = cudart.cudaHostUnregister
+
+        return self
+
+    def is_pinned(self):
+        """Returns True if the stored feature is pinned."""
+        return self._tensor.is_pinned()
+
+    def to(self, device):  # pylint: disable=invalid-name
```
mfbalin, line 148: Or movign to another device.

graphbolt/src/cnumpy.cc
mfbalin: Example: https://github.com/dmlc/dgl/blob/1ec0092e1598fdc43ccc5baf22872386d7ad865e/graphbolt/src/fused_csc_sampling_graph.cc#L306-L322

mfbalin: https://github.com/pytorch/pytorch/blob/5cc511f72fe073bbd8c10d796d72dce67f5cd5c4/aten/src/ATen/ParallelFuture.h#L10

We might want to use this to make the disk read asyncronous so that we can overlap it with other computations.

xiangyuzhi: Sure, for the asynchronous execution we may optimize it in the future.

examples/sampling/graphbolt/rgcn/hetero_rgcn.py
xiangyuzhi: OK, I think there is no need for using a new class for feature fetcher, and I will integrate it within torch feature.

graphbolt/src/index_select.cc
xiangyuzhi: Good idea.

python/dgl/graphbolt/impl/torch_based_feature_store.py
xiangyuzhi: I do not add it or retain it as NotImplement at first , but there are many tests and examples needs this interface. I will try to integrate within the torchfeature to avoid such problem.

mfbalin: Why not use BasicFeatureStore to store the DiskBasedFeatures? Then, we have two feature stores, one is TorchBasedFeatureStore, one BasicFeatureStore if needed.

@frozenbugs What is the best way to do this?

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,65 @@
...
+#include <iostream>
+#include <string>
+
+namespace graphbolt {
+namespace storage {
+
+/**
+ * @brief Disk Numpy Fetecher class.
+ */
+class OnDiskNpyArray {
```
mfbalin, line 23: Let's expose this class to torch so that it can be returned to python. Let's also make sure to open the file once and keep it open until destruction.

graphbolt/src/index_select.cc
```diff
@@ -21,6 +22,25 @@ torch::Tensor IndexSelect(torch::Tensor input, torch::Tensor index) {
   return input.index({index.to(torch::kLong)});
 }
 
+torch::Tensor DiskIndexSelect(
+    std::string path, torch::Tensor index, torch::ScalarType dtype) {
```
mfbalin, line 26: Why do we need dtype here?

xiangyuzhi: It is complex to parse the data type based on the meta data in the binary numpy file (so many types and needs a specialized function), a simple method is using high level python parse and transfer to C.

mfbalin: When we expose OnDiskNpyArray to torch, we can take dtype in the constructor and DiskIndexSelect can take OnDiskNpyArray as its first argument instead of path.

mfbalin: It can then use tensor.dtype to get it, so that getting the dtype separately is not necessary.

examples/sampling/graphbolt/rgcn/hetero_rgcn.py
xiangyuzhi: @frozenbugs Should we decide to use disk fetcher or make it optional for user ?

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,65 @@
...
+   * tensor.
+   */
+  torch::Tensor index_select_iouring(
+      torch::Tensor idx, torch::ScalarType dtype);
+
+ private:
+  std::string filename;           // Path to numpy file.
+  torch::Tensor feat_shape;       // Shape of feature, e.g. (M,K,L).
+  signed long feat_len;           // Length of feature, e.g. M*K*L.
+  std::vector<int64_t> feat_dim;  // Shape of all features, e.g. {N,M,K,L}.
```
frozenbugs, line 58: Are these all duplicated info? Do we really need to store them all?

examples/sampling/graphbolt/rgcn/hetero_rgcn.py
```diff
@@ -554,7 +554,12 @@ def train(
 
 def main(args):
     device = torch.device(
-        "cuda" if args.num_gpus > 0 and torch.cuda.is_available() else "cpu"
+        "cuda"
```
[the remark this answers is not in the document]
xiangyuzhi: OK. This seems to be a bug, because the memory of GPU cannot cantain the large graph. Maybe you can reproduce it by running: 
```
python3 examples/sampling/graphbolt/rgcn/hetero_rgcn.py --dataset ogb-lsc-mag240m
```
I will discard my change to it, and may modify it in another PR if needed.

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * @brief Read disk numpy file based on given index and transform to
+   * tensor.
+   */
+  torch::Tensor index_select_iouring(
+      torch::Tensor idx, torch::ScalarType dtype);
+
+ private:
+  std::string filename;           // Path to numpy file.
```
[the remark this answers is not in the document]
xiangyuzhi: OK

graphbolt/src/cnumpy.h
xiangyuzhi: Hi guy, I have expose this class and change to use torch parallel, maybe you can take a look at it. :)

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,217 @@
...
+#include <torch/torch.h>
+
+#include <cstring>
+#include <regex>
+#include <stdexcept>
+
+namespace graphbolt {
+namespace storage {
+
+#define ALIGNMENT 4096
```
mfbalin, line 18:
```suggestion
constexpr int disk_alignment_size = 4096;
```
If you want, you could use all caps too. Depends on the code style.

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,217 @@
...
+      (char *)aligned_alloc(ALIGNMENT, feature_size_ * num_idx);
+
+  auto idx_data = idx.data_ptr<int64_t>();
+
+  // Record the inside offsets of feteched features.
+  int64_t residual[group_size_ * num_thd_];
+
+  // Indicator for index error.
+  bool error_flag = false;
+  torch::parallel_for(0, num_idx, group_size_, [&](int64_t begin, int64_t end) {
```
mfbalin, line 123: Before we call any parallel_for, we need to add a check so that the number of threads is the same as in the constructor, such as:
`TORCH_CHECK(num_thd_ == torch::get_num_threads(), "The number of threads can not be changed after a disk feature fetcher is constructed.");`

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -72,7 +73,13 @@ class TorchBasedFeature(Feature):
     device(type='cuda', index=0)
     """
 
-    def __init__(self, torch_feature: torch.Tensor, metadata: Dict = None):
+    def __init__(
```
mfbalin, line 76: I don't quite like that TorchBasedFeature is also the DiskBasedFeature. I feel the two should be seperated. @frozenbugs what do you think?

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -117,7 +129,13 @@ def read(self, ids: torch.Tensor = None):
             if self._tensor.is_pinned():
                 return self._tensor.cuda()
             return self._tensor
-        return index_select(self._tensor, ids)
+        elif self._disk_fetcher is True and platform.system() == "Linux":
+            try:
+                return self._fetcher.index_select(ids.to("cpu")).to(ids.device)
```
mfbalin, line 134: You can simply use `.cpu()` to move a tensor to the cpu.

mfbalin (changes requested):
This PR is getting together. We need to figure out a few more things though.

python/dgl/graphbolt/impl/torch_based_feature_store.py
mfbalin: It feels like a hack. If our abstractions are right, we should not need any hacks like this.

frozenbugs: I agree, we should introduce a new DiskBasedFeature instead of keep overloading TorchBasedFeature.

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,71 @@
...
+
+ private:
+  std::string filename_;           // Path to numpy file.
+  torch::ScalarType dtype_;        // Feature data type.
+  int feature_fd_;                 // File description.
+  std::vector<int64_t> feat_dim_;  // Shape of features, e.g. {N,M,K,L}.
+  int num_thd_;                    // Default thread number.
+  int64_t group_size_ = 512;       // Default group size.
+  int64_t feature_size_;           // Number of bytes of feature size.
+  size_t prefix_len_;              // Length of head data in numpy file.
```
frozenbugs, line 60:
```suggestion
  std::string filename_;           // Path to numpy file.
  int feature_fd_;                 // File description.
  size_t prefix_len_;              // Length of head data in numpy file.
  std::vector<int64_t> feat_dim_;  // Shape of features, e.g. {N,M,K,L}.
  torch::ScalarType dtype_;        // Feature data type.
  int64_t feature_size_;           // Number of bytes of feature size.
  int num_thd_;                    // Default thread number.
  int64_t group_size_ = 512;       // Default group size.
```
nit: regroup related field together.

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,71 @@
...
+   * tensor.
+   */
+  torch::Tensor IndexSelectIOuring(torch::Tensor idx);
+
+ private:
+  std::string filename_;           // Path to numpy file.
+  torch::ScalarType dtype_;        // Feature data type.
+  int feature_fd_;                 // File description.
+  std::vector<int64_t> feat_dim_;  // Shape of features, e.g. {N,M,K,L}.
+  int num_thd_;                    // Default thread number.
```
frozenbugs, line 57: nit: `feat_dim_` -> feature_dim_ since all the `feature` here are not using abbr.

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,71 @@
...
+  int64_t group_size_ = 512;       // Default group size.
+  int64_t feature_size_;           // Number of bytes of feature size.
+  size_t prefix_len_;              // Length of head data in numpy file.
+
+#ifdef __linux__
+  io_uring* ring_;  // io_uring queue.
```
frozenbugs, line 63: ring is not a good name, io_uring_queue or whatever more descriptive.

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -86,6 +93,11 @@ def __init__(self, torch_feature: torch.Tensor, metadata: Dict = None):
         # Make sure the tensor is contiguous.
         self._tensor = torch_feature.contiguous()
         self._metadata = metadata
+        self._disk_fetcher = disk_fetcher
+        if self._disk_fetcher is True:
+            self._fetcher = torch.ops.graphbolt.disk_fetcher(
```
frozenbugs, line 98: disk_fetcher is wired, maybe just on_disk_npy_array.

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,226 @@
...
+
+namespace graphbolt {
+namespace storage {
+
+constexpr int disk_alignment_size = 4096;
+
+c10::intrusive_ptr<OnDiskNpyArray> CreateDiskFetcher(
+    std::string path, torch::ScalarType dtype) {
+  return c10::make_intrusive<OnDiskNpyArray>(path, dtype);
+}
```
frozenbugs, line 23: Make it a static class function named `Create` inside the OnDiskNpyArray?
Similar to https://github.com/dmlc/dgl/blob/2bda158238371231452c8449a4b144f1c736b786/graphbolt/src/python_binding.cc#L84C10-L84C34
and
https://github.com/dmlc/dgl/blob/2bda158238371231452c8449a4b144f1c736b786/graphbolt/include/graphbolt/fused_csc_sampling_graph.h#L100

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,226 @@
...
+                 .clone();
+  } else {
+    throw std::runtime_error("IndexError: Index out of range.");
+  }
+
+  free(read_buffer);
+  free(result_buffer);
+
+  return result;
+#else
```
frozenbugs, line 220:
```suggestion
#else // __linux__
```

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,226 @@
...
+  return torch::empty({0});
+#endif
```
frozenbugs, line 222: ditto.

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * @brief Parse numpy meta data.
+   */
+  void ParseNumpyHeader();
+
+  /**
+   * @brief Read disk numpy file based on given index and transform to
+   * tensor.
+   */
+  torch::Tensor IndexSelectIOuring(torch::Tensor idx);
```
frozenbugs, line 50: If io_uring is guarded by `#if def __linux__`, then the `IndexSelectIOuring` should too.
I'd suggest that changing this to IndexSelect, and make IndexSelectIOuring a separate method, and dispatch in IndexSelect.

frozenbugs: nit: IndexSelectIOUring

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,226 @@
...
+
+  // Get system max thread number.
+  num_thd_ = torch::get_num_threads();
+  ring_ = new io_uring[num_thd_];
+
+  // Init io_uring queue.
+  for (int64_t t = 0; t < num_thd_; t++) {
+    io_uring_queue_init(group_size_, &ring_[t], 0);
+  }
+#endif
```
frozenbugs, line 42:
```suggestion
#endif // __linux__
```

graphbolt/src/cnumpy.h
frozenbugs: `num_thd_` -> `num_thread_` unnecessary abbr.

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,71 @@
...
+
+ private:
+  std::string filename_;           // Path to numpy file.
+  torch::ScalarType dtype_;        // Feature data type.
+  int feature_fd_;                 // File description.
```
frozenbugs, line 55: I would rename this to `file_description_` since it does not really related to feature, just a file property.

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,226 @@
...
+  // IO queue exit.
+  for (int64_t t = 0; t < num_thd_; t++) {
+    io_uring_queue_exit(&ring_[t]);
+  }
+  close(feature_fd_);
+#endif
+}
+
+void OnDiskNpyArray::ParseNumpyHeader() {
+  FILE *fp = fopen(filename_.c_str(), "rb");
```
frozenbugs, line 56: Can we use open here instead of using c fopen?

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,226 @@
...
+#ifdef __linux__
+  // IO queue exit.
+  for (int64_t t = 0; t < num_thd_; t++) {
+    io_uring_queue_exit(&ring_[t]);
+  }
+  close(feature_fd_);
+#endif
+}
+
+void OnDiskNpyArray::ParseNumpyHeader() {
```
frozenbugs, line 55: Can you add doc string about this method, what does it parse and give examples?

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,226 @@
...
+  if (!fp)
+    throw std::runtime_error("npy_load: Unable to open file " + filename_);
+  char buffer[256];
+  size_t res = fread(buffer, sizeof(char), 11, fp);
+  if (res != 11) throw std::runtime_error("parse_npy_header: failed fread");
+  std::string header = fgets(buffer, 256, fp);
+  assert(header[header.size() - 1] == '\n');
+  prefix_len_ = 11 + header.size();
+
+  size_t loc1, loc2;
```
frozenbugs, line 66: please avoid using abbreviation as much as you can to avoid confusion.

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,226 @@
...
+
+  auto idx_data = idx.data_ptr<int64_t>();
+
+  // Record the inside offsets of feteched features.
+  int64_t residual[group_size_ * num_thd_];
+
+  // Indicator for index error.
+  bool error_flag = false;
+  TORCH_CHECK(
+      num_thd_ == torch::get_num_threads(),
```
frozenbugs, line 126: why do we need this check? you did not expose any api to change this private variable.

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,226 @@
...
+#include <torch/torch.h>
+
+#include <cstring>
+#include <regex>
+#include <stdexcept>
+
+namespace graphbolt {
+namespace storage {
+
+constexpr int disk_alignment_size = 4096;
```
frozenbugs, line 18: static constexpr int kDiskAlignmentSize = 4096;

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,226 @@
...
+
+  // Indicator for index error.
+  bool error_flag = false;
+  TORCH_CHECK(
+      num_thd_ == torch::get_num_threads(),
+      "The number of threads can not be changed after a disk feature fetcher "
+      "is constructed.");
+  torch::parallel_for(0, num_idx, group_size_, [&](int64_t begin, int64_t end) {
+    auto thd_id = torch::get_thread_num();
+    if (!error_flag) {
```
frozenbugs, line 131: @peizhou001 is this thread safe? why we need to set a flag rather than throw error directly?

graphbolt/src/cnumpy.cc
mfbalin: We need it because the constructor initializes a io_uring_queue for each thread. If the user changes the thread count after constructing this object, for any additional threads, we won't have the io_uring queue and there will be an out of bounds access. Maybe the check can be changed so that `num_thd_ >= torch::get_num_threads()`.

graphbolt/src/cnumpy.h
xiangyuzhi: Are the IndexSelect and IndexSelectIOuring both internal methods? It looks like:
cnumpu.h
```
  torch::Tensor IndexSelect(torch::Tensor idx);
#ifdef __linux__
  torch::Tensor IndexSelectIOuring(torch::Tensor idx);
#endif  // __linux__
```

cnumpy.cc
```
torch::Tensor OnDiskNpyArray::IndexSelect(torch::Tensor idx) {
#ifdef __linux__
  return IndexSelectIOuring(idx);
#else
  return torch::empty({0});
#endif  // __linux__

#ifdef __linux__
torch::Tensor OnDiskNpyArray::IndexSelectIOuring(torch::Tensor idx) {
...
}
#endif  // __linux__
```

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -13,7 +13,162 @@
...
+    >>> feature = gb.DiskBasedFeature(torch_feat)
+    >>> feature.read(torch.tensor([0]))
+    tensor([[0, 1, 2, 3, 4]])
+    >>> feature.size()
+    torch.Size([5])
+    """
+
+    def __init__(self, path: str, metadata: Dict = None):
+        super().__init__()
+        self._is_inplace_pinned = set()
```
mfbalin, line 43:
```suggestion
```

mfbalin: Did you test if pinning works with mmap'ed tensor? If it doesn't work, let's remove pinning functionality. Let's not add any untested code.

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -12,7 +13,162 @@
...
+        )
+
+    def __del__(self):
+        # torch.Tensor.pin_memory() is not an inplace operation. To make it
+        # truly in-place, we need to use cudaHostRegister. Then, we need to use
+        # cudaHostUnregister to unpin the tensor in the destructor.
+        # https://github.com/pytorch/pytorch/issues/32167#issuecomment-753551842
+        for tensor in self._is_inplace_pinned:
+            assert self._inplace_unpinner(tensor.data_ptr()) == 0
+
```
mfbalin, line 61:
```suggestion
```
Needs to be removed if pinning doesn't work.

mfbalin (commented):
Let's add pinning tests. If it doesn't work, let's remove the pinning code.

python/dgl/graphbolt/impl/torch_based_feature_store.py
xiangyuzhi: For datasets larger than memory, pinning them on memory is theoretically impossible, so let's just remove it.

xiangyuzhi:
@dgl-bot

graphbolt/src/cnumpy.h
```diff
@@ -0,0 +1,117 @@
...
+   * The header contains three descriptors: 'descr' for data format description
+   * 'fortran_order' for row or column order preference, and 'shape' for data
+   * shape. The data format description includes the endianness, data type, and
+   * data length, e.g. '<i8' denotes little-endian storage, integer type, and
+   * 8-byte size (64 bit integer).
+   *
+   * For example, for a tensor `A` having 2 features with following entries:
+   * [[1, 2, 3], [4, 5, 6]]
+   * When `A` is saved as an numpy file, the data is stored as:
+   *   �NUMPY'descr': '<i8', 'fortran_order': False, 'shape': (2, 3), }\n
```
frozenbugs, line 66: Can you solve the Garbled characters?

frozenbugs, line 66: what is the first?

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,257 @@
...
+#endif  // __linux__
+}
+
+void OnDiskNpyArray::ParseNumpyHeader() {
+#ifdef __linux__
+  // Parse numpy file header to get basic info of feature.
+  int file_description = open(filename_.c_str(), O_RDONLY);
+  if (file_description == -1)
+    throw std::runtime_error(
+        "ParseNumpyHeader: Unable to open file " + filename_);
```
frozenbugs, line 63:
```suggestion
  if (file_description == -1) {
    throw std::runtime_error(
        "ParseNumpyHeader: Unable to open file " + filename_);
  }
```

frozenbugs: nit: As long as change lines, use {}

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,257 @@
...
+
+  char buffer[kMaxHeaderSize];
+  // Prefix of length kSkipHeaderSize do not contain information.
+  ssize_t res = read(file_description, buffer, kSkipHeaderSize);
+  if (res != kSkipHeaderSize)
```
frozenbugs, line 68: ditto.

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,257 @@
...
+#ifdef __linux__
+  // Parse numpy file header to get basic info of feature.
+  int file_description = open(filename_.c_str(), O_RDONLY);
+  if (file_description == -1)
+    throw std::runtime_error(
+        "ParseNumpyHeader: Unable to open file " + filename_);
+
+  char buffer[kMaxHeaderSize];
+  // Prefix of length kSkipHeaderSize do not contain information.
+  ssize_t res = read(file_description, buffer, kSkipHeaderSize);
```
frozenbugs, line 67: nit: ssize_t bytes_read

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,257 @@
...
+  if (res != kSkipHeaderSize)
+    throw std::runtime_error("ParseNumpyHeader: failed read numpy header");
+  // Store header string.
+  std::string header;
+  while (read(file_description, buffer, 1) == 1) {
```
frozenbugs, line 72: If the header is just the first line of data, did you consider use readline?

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,257 @@
...
+    if (buffer[0] == '\n') break;
+  }
+  assert(header[header.size() - 1] == '\n');
+  // Get prefix length for computing feature offset.
+  prefix_len_ = kSkipHeaderSize + header.size();
+  close(file_description);
+
+  // Get header description string.
+  size_t descr_begin = header.find("descr");
+  size_t descr_end = descr_begin + 9;
```
frozenbugs, line 83: explain in doc why + 9

graphbolt/src/cnumpy.h
xiangyuzhi: It seems to be a little binary data that cannot be represented.

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,221 @@
...
+  prefix_len_ = header.size() + 1;
+  file.close();
+#endif  // __linux__
+}
+
+torch::Tensor OnDiskNpyArray::IndexSelect(torch::Tensor index) {
+#ifdef __linux__
+  return IndexSelectIOUring(index);
+#else
+  return torch::empty({0});
```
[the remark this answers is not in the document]
mfbalin: @frozenbugs do you think we can have such a check in the constructor instead?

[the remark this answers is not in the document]
frozenbugs: Yes, it is better to throw error, than randomly return an empty value.

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -117,7 +243,7 @@ def read(self, ids: torch.Tensor = None):
             if self._tensor.is_pinned():
                 return self._tensor.cuda()
             return self._tensor
-        return index_select(self._tensor, ids)
+        return torch.ops.graphbolt.index_select(self._tensor, ids)
```
mfbalin, line 246: why is this change needed?

mfbalin: If it is not necessary, let's roll it back.

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -271,6 +397,7 @@ class TorchBasedFeatureStore(BasicFeatureStore):
 
     def __init__(self, feat_data: List[OnDiskFeatureData]):
         features = {}
+
```
mfbalin, line 400: Let's take unrelated changes back.

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -12,7 +13,132 @@
...
+    def pin_memory_(self):
+        """Disk based feature does not support pinning memory for now."""
+        raise NotImplementedError
+
+    def is_pinned(self):
+        """Returns True if the stored feature is pinned."""
+        return False
+
+    def to(self, device):  # pylint: disable=invalid-name
```
mfbalin, line 112: We most likely should not support the `to` function either. Let's clean the code. Would the user want to convert between TorchBasedFeature and DiskBasedFeature?

xiangyuzhi: OK, I just want to avoid possible errors of the call, so make it resemble TorchBaseFeature as much as possible. Let's remain them as NotImplement.

mfbalin: Where do you get an error? https://docs.dgl.ai/en/latest/generated/dgl.graphbolt.BasicFeatureStore.html#dgl.graphbolt.BasicFeatureStore
BasicFeatureStore does not have pin_memory_ or to methods. If there will be an error because there is no `to` method, then it means we are calling the method. But we will get an error anyway because it is not implemented. So, I don't see the benefit of keeping these methods. Could you elaborate?

xiangyuzhi: At the very beginning, I find it will occur errors if we direct apply DiskFeature to run this example https://github.com/dmlc/dgl/blob/master/examples/sampling/graphbolt/rgcn/hetero_rgcn.py#L573, for small dataset, it should be pinned on memory, but for large dataset it's impossible. I want Disk Feature to keep the ability to deal with small datasets (e.i. the same as TorchBased). However, if we want to split it as extra method to only deal with large datasets, we can throw all useless function.

xiangyuzhi: I think DiskBasedFeature's APIs should be kept consistent with TorchBasedFeature, as we consider the two classes to be equivalent.

mfbalin: Why do we consider them equivalent? One is a torch tensor, the other is tensor data stored on the disk.

mfbalin: We can instead write a method to get a TorchBasedFeature out of a DiskBasedFeature.

mfbalin: read_into_ram() -> TorchBasedFeature

Something like above.

xiangyuzhi: Because we want to replace the mmap-based method in the TorchBasedFeatureStore at the beginning: https://github.com/dmlc/dgl/blob/master/python/dgl/graphbolt/impl/torch_based_feature_store.py#L287

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -12,7 +13,126 @@
...
+        """Disk based feature does not support pinning memory."""
+        raise NotImplementedError
+
+    def is_pinned(self):
+        """Returns True if the stored feature is pinned."""
+        return False
+
+    def to(self, device):  # pylint: disable=invalid-name
+        """Disk based feature does not support copying to device."""
+        raise NotImplementedError
```
mfbalin, line 114:
```suggestion
    def read_into_memory(self) -> TorchBasedFeature:
        return TorchBasedFeature()
```

python/dgl/graphbolt/impl/torch_based_feature_store.py
mfbalin: Then we can modify TorchBasedFeatureStore's to and pin_memory_ methods so that it skips over DiskBasedFeature s in it.

xiangyuzhi: OK, it is a possible solution. let's do as you say.

mfbalin: If we have read_into_memory, then we could simply implement `.to` as:

```
def to(self, device): -> TorchBasedFeature
    return self.read_into_memory().to(device)
```

mfbalin: We can implement `.to` as above but `pin_memory_` is a bit complicated because it is an in-place operation. We can't convert a DiskBasedFeature to a TorchBasedFeature in-place.

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -234,6 +235,116 @@ def __repr__(self) -> str:
...
+        -------
+        Dict
+            The metadata of the feature.
+        """
+        return (
+            self._metadata if self._metadata is not None else super().metadata()
+        )
+
+    def read_into_memory(self) -> TorchBasedFeature:
+        return TorchBasedFeature(self._tensor, self.metadata)
```
mfbalin, line 324: Is self._tensor read into memory here? What function should be called so that self._tensor is not mmap'ed anymore but is on CPU?

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -234,6 +235,116 @@ def __repr__(self) -> str:
...
+        """Get the metadata of the feature.
+        Returns
+        -------
+        Dict
+            The metadata of the feature.
+        """
+        return (
+            self._metadata if self._metadata is not None else super().metadata()
+        )
+
```
mfbalin, line 322:
```suggestion
    def to(self, device) -> TorchBasedFeature:
        return self.read_into_memory().to(device)
```

python/dgl/graphbolt/impl/torch_based_feature_store.py
mfbalin: Do we need to call `.detach().clone()` when passing `self._tensor` into the TorchBasedFeature constructor?

mfbalin: I think it could be best to use https://pytorch.org/docs/main/generated/torch.tensor.html#torch.tensor here.

mfbalin: However it could create a copies, @xiangyuzhi is it fine to keep it as it is?

mfbalin: Or construct it again, this time without mmap:
```python
torch.from_numpy(np.load(self.path)).contiguous()
```

mfbalin:
@xiangyuzhi Do you think it is a better idea to make DiakBasedFeature a subclass of TorchBasedFeature if we want to treat the two the same?

xiangyuzhi:
> @xiangyuzhi Do you think it is a better idea to make DiakBasedFeature a subclass of TorchBasedFeature if we want to treat the two the same?

It looks wired. I think we can use the current solution for now. Since we need a series of adjustments to use this class later, including making it optional, or automatically selecting the feature class based on data and memory size. We can modify and use it in the future PRs.

graphbolt/src/cnumpy.cc
```diff
@@ -0,0 +1,221 @@
...
+  // Indicator for index error.
+  std::atomic<bool> error_flag{};
+  TORCH_CHECK(
+      num_thread_ >= torch::get_num_threads(),
+      "The number of threads can not be changed to larger than the number of "
+      "threads when a disk feature fetcher is constructed.");
+  torch::parallel_for(
+      0, num_index, group_size_, [&](int64_t begin, int64_t end) {
+        auto thread_id = torch::get_thread_num();
+        if (!error_flag.load()) {
```
peizhou001, line 123: According to test, `parallel_for`  can automatically catch the error after all threads are done. So this `error_flag`  is almost meaningless.

xiangyuzhi: Yes, it can automatically catch. But it doesn't stop the execution in the parallel and exit immediately, which will cause error when use io_uring.

peizhou001: > Yes, it can automatically catch. But it doesn't stop the execution in the parallel and cannot exit immediately, which will cause error when use io_uring.

Could you explain more what error could be caused under this circumanstance?

xiangyuzhi: For example, an out-of-range index would lead to accessing an offset beyond the file length, causing a crash. If we skip it, we also need to adjust the number of IO_uring requests asynchronously received later, or else it will result in an indefinite wait for accomplish.

xiangyuzhi: do you have some ideas about avoiding it ?

> > Yes, it can automatically catch. But it doesn't stop the execution in the parallel and cannot exit immediately, which will cause error when use io_uring.
> 
> Could you explain more what error could be caused under this circumanstance?

peizhou001: > do you have some ideas about avoiding it ?
> 
> > > Yes, it can automatically catch. But it doesn't stop the execution in the parallel and cannot exit immediately, which will cause error when use io_uring.
> > 
> > 
> > Could you explain more what error could be caused under this circumanstance?

Is this possible to handle this in IO_URING serving side, likely set the timeout.

peizhou001: > For example, an out-of-range index would lead to accessing an offset beyond the file length, causing a crash. If we skip it, we also need to adjust the number of IO_uring requests asynchronously received later, or else it will result in an indefinite wait for accomplish.

Firstly we need to know if this code is an improvement for fast fail or a fix for potential bugs. 
If it is fast fail, it's okay.
Otherwise, we need to see if it can handle all potential exceptions.

peizhou001 (approved):

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -234,6 +235,117 @@ def __repr__(self) -> str:
...
+    ----------
+    path : string
+        The path to the numpy feature file.
+        Note that the dimension of the numpy should be greater than 1.
+    Examples
+    --------
+    >>> import torch
+    >>> from dgl import graphbolt as gb
+    >>> torch_feat = torch.arange(10).reshape(2, -1)
+    >>> feature = gb.DiskBasedFeature(torch_feat)
```
Rhett-Ying, line 253: mismatch with `__init__`

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -234,6 +235,117 @@ def __repr__(self) -> str:
         )
 
 
+class DiskBasedFeature(Feature):
+    r"""A wrapper of disk based feature.
+
+    Initialize a disk based feature fetcher by a numpy file.
```
Rhett-Ying, line 241: we'd better mention the different underlying behavior on Windows and Linux

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -234,6 +235,117 @@ def __repr__(self) -> str:
...
+    >>> feature.read(torch.tensor([0]))
+    tensor([[0, 1, 2, 3, 4]])
+    >>> feature.size()
+    torch.Size([5])
+    """
+
+    def __init__(self, path: str, metadata: Dict = None):
+        super().__init__()
+        mmap_mode = "r+"
+        self._tensor = torch.from_numpy(
```
Rhett-Ying, line 263: can this be wrapped into existing `read_data()` for wider usage such as `OnDiskDataset`?
https://github.com/dmlc/dgl/blob/8e3d810127e01dff388cea588e917c3f1c9943a8/python/dgl/graphbolt/internal/utils.py#L25

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -234,6 +235,117 @@ def __repr__(self) -> str:
...
+    >>> feature = gb.DiskBasedFeature(torch_feat)
+    >>> feature.read(torch.tensor([0]))
+    tensor([[0, 1, 2, 3, 4]])
+    >>> feature.size()
+    torch.Size([5])
+    """
+
+    def __init__(self, path: str, metadata: Dict = None):
+        super().__init__()
+        mmap_mode = "r+"
```
Rhett-Ying, line 262: is `writable` really required? `r` should be ok?

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -234,6 +235,117 @@ def __repr__(self) -> str:
...
+        torch.Tensor
+            The read feature.
+        """
+        if ids is None:
+            if self._tensor.is_pinned():
+                return self._tensor.cuda()
+            return self._tensor
+        elif platform.system() == "Linux":
+            try:
+                return self._ondisk_npy_array.index_select(ids.cpu()).to(
```
Rhett-Ying, line 291: Is it possible and better to make the index select API of `ondisk_npy_array` be compatible with torch? Namely, improve `index_select()` to support `ondisk_npy_array`.

python/dgl/graphbolt/impl/torch_based_feature_store.py
```diff
@@ -234,6 +235,117 @@ def __repr__(self) -> str:
...
+        Returns
+        -------
+        Dict
+            The metadata of the feature.
+        """
+        return (
+            self._metadata if self._metadata is not None else super().metadata()
+        )
+
+    def read_into_memory(self) -> TorchBasedFeature:
```
Rhett-Ying, line 323: should we mention that when we'd better call this API?
And it seems no specific testcase for this API.

PR #7129, opened 2024-02-20
[Benchmark] Modify node_classification test to enable `sample_layer_neighbors`.

caojy1998:
## Checklist
- [x] Code is well-documented


Skeleton003 (approved):
LGTM.

examples/sampling/graphbolt/node_classification.py
```diff
@@ -117,9 +117,14 @@ def create_dataloader(
...
-        graph, fanout if job != "infer" else [-1]
-    )
+    if args.sample_mode == "sample-neighbors":
+        datapipe = datapipe.sample_neighbor(
+            graph, fanout if job != "infer" else [-1]
+        )
+    else:
+        datapipe = datapipe.sample_layer_neighbor(
+            graph, fanout if job != "infer" else [-1]
+        )
```
mfbalin, line 127: Do you think it would be better to have the following while enforcing `args.sample_mode` to be either `sample_neighbor` or `sample_layer_neighbor`?
```suggestion
    datapipe = getattr(datapipe, args.sample_mode)(
        graph, fanout if job != "infer" else [-1]
    )
```

mfbalin: Or is this suggestion making the code too obscure?

examples/sampling/graphbolt/node_classification.py
```diff
@@ -364,6 +373,21 @@ def parse_args():
         help="Dataset storage placement and Train device: 'cpu' for CPU and RAM,"
         " 'pinned' for pinned memory in RAM, 'cuda' for GPU and GPU memory.",
     )
+    parser.add_argument(
+        "--sample-mode",
+        default="sample-neighbors",
+        choices=["sample-neighbors", "sample-layer-neighbors"],
```
mfbalin, line 379:
```suggestion
        default="sample_neighbor",
        choices=["sample_neighbor", "sample_layer_neighbor"],
```

mfbalin: If we apply the suggestion above, we might want to apply this one too.

examples/sampling/graphbolt/node_classification.py
```diff
@@ -364,6 +373,21 @@ def parse_args():
...
+        help="The sampling function when doing layerwise sampling.",
+    )
+    parser.add_argument(
+        "--overlap-graph-fetch",
+        type=bool,
+        default=False,
```
mfbalin, line 385: action='store_true',

mfbalin, line 385:
```suggestion
        action='store_true',
```

mfbalin: So that we can write `--overlap-graph-fetch` to enable it instead of having to write `--overlap-graph-fetch=True`.

examples/sampling/graphbolt/node_classification.py
```diff
@@ -364,6 +373,21 @@ def parse_args():
...
+        help="An option for enabling overlap_graph_fetch in graphbolt dataloader."
+        "If True, the data loader will overlap the UVA graph fetching operations"
+        "with the rest of operations by using an alternative CUDA stream. Default"
+        "is False.",
```
mfbalin, line 389:
```suggestion
        "with the rest of operations by using an alternative CUDA stream. Disabled"
        "by default.",
```

mfbalin (approved):
Consider the comments I made "nit" comment, you can judge whether they would make the code better.

examples/sampling/graphbolt/node_classification.py
caojy1998: OK, I have changed here.

examples/sampling/graphbolt/node_classification.py
caojy1998: Because the args.sample_mode is `sample-neighbor` and the function is `sample_neighbor`. So using `getattr` may not work here.

mfbalin: My suggestion below was making `-` into `_` so that `getattr` works.
https://github.com/dmlc/dgl/pull/7129#discussion_r1496229697

caojy1998: OK, I will take this approach.

examples/sampling/graphbolt/node_classification.py
caojy1998: OK, I have changed here.

PR #7127, opened 2024-02-20
[GraphBolt] Support saving graph and TVT_Sets in `int32`

Skeleton003:
## Description
Depends on #7131.

## Checklist
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [ ] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [ ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

PR #7131, opened 2024-02-20
[GraphBolt] Cast sampled data to minimum dtype.

mfbalin:
## Description
We cast the sampled subgraphs into int32 if they fit into it. When we have mixed indptr and indices dtypes (int64 and int32), this casting is needed to ensure that DGL receives tensors that are of the same dtype as DGL doesn't support mixed dtypes.

## Checklist
- [ ] Code is well-documented


python/dgl/graphbolt/minibatch.py
```diff
@@ -231,6 +231,20 @@ def blocks(self):
             self.sampled_subgraphs[0].sampled_csc, Dict
         )
 
+        # casts to minimum dtype in-place and returns self.
+        def cast_to_minimum_dtype(v: CSCFormatBase):
+            # Checks if number of vertices and edges fit into an int32.
+            if (
+                max(v.indptr.size(0) - 2, v.indices.size(0))
```
Rhett-Ying, line 238: `-1` or `-2`?

python/dgl/graphbolt/minibatch.py
```diff
@@ -242,6 +256,8 @@ def blocks(self):
                 original_column_node_ids is not None
             ), "Missing `original_column_node_ids` in sampled subgraph."
             if is_heterogeneous:
+                for v in subgraph.sampled_csc.values():
+                    cast_to_minimum_dtype(v)
```
Rhett-Ying, line 260: `v = cast_to_minimum_dtype(v)`? or omit the return value if update on input argument?

Rhett-Ying (commented):
pls add testcase for this change

python/dgl/graphbolt/minibatch.py
mfbalin: -1 gives num_nodes and -2 gives max vertex id stores in the indices array. Thus, we need -2. Size of indices gives the max value stores in indptr array. So we are taking max of the max value stored in either array.

python/dgl/graphbolt/minibatch.py
mfbalin: Returning self was helpful in the below case to avoid an extra assignment.

mfbalin: Otherwise, the modification is in-place so this code works.

python/dgl/graphbolt/minibatch.py
```diff
@@ -231,6 +231,20 @@ def blocks(self):
...
+        # casts to minimum dtype in-place and returns self.
+        def cast_to_minimum_dtype(v: CSCFormatBase):
+            # Checks if number of vertices and edges fit into an int32.
+            if (
+                max(v.indptr.size(0) - 2, v.indices.size(0))
+                <= torch.iinfo(torch.int32).max
+            ):
+                dtype = torch.int32
+            else:
+                dtype = torch.int64
```
Skeleton003, line 243: Could these lines be written as
```python
dtype = torch.int32 if max(v.indptr.size(0) - 2, v.indices.size(0)) <= torch.iinfo(torch.int32).max else torch.int64
```

mfbalin: It could, I rewrote it in another way. How does it look now?

mfbalin: Updated.

Skeleton003:
LGTM. These changes may be helpful in addressing the CI failure I met in #7127 .

mfbalin:
> LGTM. These changes may be helpful in addressing the CI failure I met in #7127 .

I think it definitely will.

Rhett-Ying (approved):

python/dgl/graphbolt/minibatch.py
```diff
@@ -231,6 +231,19 @@ def blocks(self):
             self.sampled_subgraphs[0].sampled_csc, Dict
         )
 
+        # casts to minimum dtype in-place and returns self.
+        def cast_to_minimum_dtype(v: CSCFormatBase):
```
frozenbugs, line 235: Should we do it in the compact stage?

mfbalin: We can consider doing it in the compact stage as an optimization to fuse the type casting into the compact operation to save runtime.

python/dgl/graphbolt/minibatch.py
```diff
@@ -231,6 +231,19 @@ def blocks(self):
             self.sampled_subgraphs[0].sampled_csc, Dict
         )
 
+        # casts to minimum dtype in-place and returns self.
```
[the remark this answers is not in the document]
mfbalin: Merging it in #7147.

PR #7137, opened 2024-02-22
[GraphBolt][CUDA] RGCN example

mfbalin:
## Description
Adding GPU sampling support for the R-GCN example.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/examples/sampling/graphbolt/rgcn/hetero_rgcn.py
+++ b/examples/sampling/graphbolt/rgcn/hetero_rgcn.py
@@ -114,6 +114,11 @@ def create_dataloader(
     #   Whether to shuffle the items in the dataset before sampling.
     datapipe = gb.ItemSampler(item_set, batch_size=batch_size, shuffle=shuffle)
 
+    # Move the mini-batch to the appropriate device.
+    # `device`:
+    #   The device to move the mini-batch to.
+    datapipe = datapipe.copy_to(device, extra_attrs=["seed_nodes"])
+
     # Sample neighbors for each seed node in the mini-batch.
     # `graph`:
     #   The graph(FusedCSCSamplingGraph) from which to sample neighbors.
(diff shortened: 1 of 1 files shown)


frozenbugs:
LGTM

examples/sampling/graphbolt/rgcn/hetero_rgcn.py
```diff
@@ -657,7 +658,7 @@ def main(args):
     )
     parser.add_argument("--num_epochs", type=int, default=3)
     parser.add_argument("--num_workers", type=int, default=0)
-    parser.add_argument("--num_gpus", type=int, default=0)
+    parser.add_argument("--num_gpus", type=int, default=1)
```
Rhett-Ying, line 661: default to 0?

examples/sampling/graphbolt/rgcn/hetero_rgcn.py
```diff
@@ -133,17 +138,6 @@ def create_dataloader(
         node_feature_keys["institution"] = ["feat"]
     datapipe = datapipe.fetch_feature(features, node_feature_keys)
 
-    # Move the mini-batch to the appropriate device.
```
Rhett-Ying, line 136: so in the latest example, we only support
1. cpu sampling+ cpu training
2. gpu sampling + gpu training
The hack here is for cpu sampling + gpu training. If so, it's ok to remove the hack.

Rhett-Ying (approved):

examples/sampling/graphbolt/rgcn/hetero_rgcn.py
mfbalin: Default to 1 because if there is no GPU available, it will be set back to 0 automatically.

mfbalin: Why do we need this parameter? We could simply remove it?

examples/sampling/graphbolt/rgcn/hetero_rgcn.py
mfbalin: Yeah, that seems to be the case.

PR #7183, opened 2024-03-01
[GraphBolt] Add recommended PyG example.

mfbalin:
## Description
Alternative recommended (for performance) way to use PyG with GraphBolt is demonstrated in this example. Should showcase most of the GraphBolt functionalities that we also make available for PyG users. We can use this example when comparing performance against the [PyG original example](https://github.com/pyg-team/pytorch_geometric/blob/25b2f208e671eeec285bfafa2e246ea0a234b312/examples/ogbn_products_sage.py) in the release blog.

Feel free to push to my branch to change comments, the print format, ordering of functions, addition of timing or minor things like that.

Speedups on a DGX-A100 machine with two 64 core AMD EPYC CPUs and 4 A100's. We use only a single A100 in the runs.

node_classification_vs_pyg

## Checklist
- [ ] Code is well-documented


frozenbugs (approved):

mfbalin:
I was able to move the coo conversion into the model definition, making it a bit simpler to use. Since #7188 was merged, it is okay to do that now because torch compile still works even when `gb.expand_indptr` is being used in model forward. @frozenbugs

Rhett-Ying (approved):

PR #7208, opened 2024-03-12
[GraphBolt] Implement labor dependent minibatching - python side.

mfbalin:
## Description
Depends on #7207.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/impl/neighbor_sampler.py
+++ b/python/dgl/graphbolt/impl/neighbor_sampler.py
@@ -146,12 +146,17 @@ class SamplePerLayerFromFetchedSubgraph(MiniBatchTransformer):
 
     def _sample_per_layer_from_fetched_subgraph(self, minibatch):
         subgraph = minibatch.sampled_subgraphs[0]
-
+        kwargs = {
+            key[1:]: getattr(minibatch, key)
+            for key in ["_random_seed", "_seed2_contribution"]
+            if hasattr(minibatch, key)
+        }
         sampled_subgraph = getattr(subgraph, self.sampler_name)(
             minibatch._subgraph_seed_nodes,
             self.fanout,
             self.replace,
             self.prob_name,
+            **kwargs,
         )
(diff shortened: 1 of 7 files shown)


python/dgl/graphbolt/impl/neighbor_sampler.py
```diff
@@ -244,10 +258,56 @@ def __init__(
         prob_name,
         deduplicate,
         sampler,
+        layer_dependency=None,
+        batch_dependency=None,
     ):
+        if sampler.__name__ == "sample_layer_neighbors":
```
frozenbugs, line 264: Instead of checking the sampler name here, why not move the self._init_seed(batch_dependency) to __init__ of LayerNeighborSampler?

python/dgl/graphbolt/impl/neighbor_sampler.py
```diff
@@ -277,11 +337,22 @@ def _set_input_nodes(minibatch):
...
+        deduplicate,
+        sampler,
+        layer_dependency,
     ):
         datapipe = datapipe.transform(
             partial(self._prepare, graph.node_type_to_id)
         )
+        is_labor = sampler.__name__ == "sample_layer_neighbors"
+        if is_labor:
+            datapipe = datapipe.transform(self._set_seed)
```
frozenbugs, line 355: Same here, is it possible to move this sample_layer_neighbors specific code to its own class, instead of hack in the parent class.

mfbalin: How do you propose we do that without replicating the whole implementation?

mfbalin: I understand that. But I feel anti-oop pattern is better than replicating the code. I plan to base Temporal samplers into the NeighborSamplerImpl as well, so that the GPU sampling optimizations can be enabled for the temporal case as well.

PR #7217, opened 2024-03-15
[GraphBolt] Make unique_and_compact deterministic

RamonZhou:
## Description
Doc: https://docs.google.com/document/d/1fWX6GhQySniFCSmXnKw8mxZQ26WngD5jEi8vt2H41Ss

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/graphbolt/src/concurrent_id_hash_map.cc
+++ b/graphbolt/src/concurrent_id_hash_map.cc
@@ -68,7 +68,7 @@ torch::Tensor ConcurrentIdHashMap<IdType>::Init(
   // Fill in the first `num_seeds` ids.
   torch::parallel_for(0, num_seeds, kGrainSize, [&](int64_t s, int64_t e) {
     for (int64_t i = s; i < e; i++) {
-      InsertAndSet(ids_data[i], static_cast<IdType>(i));
+      InsertAndSetSmaller(ids_data[i], static_cast<IdType>(i));
     }
   });
   // Place the first `num_seeds` ids.
@@ -85,10 +85,20 @@ torch::Tensor ConcurrentIdHashMap<IdType>::Init(
   // Insert all elements in this loop.
   torch::parallel_for(
       num_seeds, num_ids, kGrainSize, [&](int64_t s, int64_t e) {
+        for (int64_t i = s; i < e; i++) {
+          InsertAndSetSmaller(ids_data[i], static_cast<IdType>(i));
+        }
+      });
+  
+  // Count the valid numbers in each thread.
+  torch::parallel_for(
+      num_seeds, num_ids, kGrainSize, [&](int64_t s, int64_t e) {
         size_t count = 0;
         for (int64_t i = s; i < e; i++) {
-          valid[i] = Insert(ids_data[i]);
-          count += valid[i];
+          if (MapId(ids_data[i]) == i) {
+            count ++;
+            valid[i] = 1;
+          }
         }
         auto thread_id = torch::get_thread_num();
         block_offset[thread_id + 1] = count;
@@ -197,6 +207,30 @@ inline void ConcurrentIdHashMap<IdType>::InsertAndSet(IdType id, IdType value) {
   }
 
   hash_map_.data_ptr<IdType>()[getValueIndex(pos)] = value;
+}
+
+template <typename IdType>
+void ConcurrentIdHashMap<IdType>::InsertAndSetSmaller(IdType id, IdType value) {
+  IdType pos = (id & mask_), delta = 1;
+  IdType* hash_map_data = hash_map_.data_ptr<IdType>();
+  InsertState state = AttemptInsertAt(pos, id);
+  while (state == InsertState::OCCUPIED) {
+    Next(&pos, &delta);
+    state = AttemptInsertAt(pos, id);
+  }
+
+  if (state == InsertState::INSERTED) {
+    hash_map_data[getValueIndex(pos)] = value;
+    return;
+  }
+
+  IdType val_pos = getValueIndex(pos);
+  IdType old_val = hash_map_data[val_pos];
+  while (old_val > value) {
+    old_val = CompareAndSwap(&(hash_map_data[val_pos]), old_val, value);
+  }
+
+  return;
 }
 
 template <typename IdType>
--- a/graphbolt/src/concurrent_id_hash_map.cc
+++ b/graphbolt/src/concurrent_id_hash_map.cc
@@ -89,14 +89,14 @@ torch::Tensor ConcurrentIdHashMap<IdType>::Init(
           InsertAndSetSmaller(ids_data[i], static_cast<IdType>(i));
         }
       });
-  
+
   // Count the valid numbers in each thread.
   torch::parallel_for(
       num_seeds, num_ids, kGrainSize, [&](int64_t s, int64_t e) {
         size_t count = 0;
         for (int64_t i = s; i < e; i++) {
           if (MapId(ids_data[i]) == i) {
-            count ++;
+            count++;
             valid[i] = 1;
           }
         }
--- a/graphbolt/src/concurrent_id_hash_map.cc
+++ b/graphbolt/src/concurrent_id_hash_map.cc
@@ -65,8 +65,8 @@ torch::Tensor ConcurrentIdHashMap<IdType>::Init(
   // This code block is to fill the ids into hash_map_.
   auto unique_ids = torch::empty_like(ids);
   IdType* unique_ids_data = unique_ids.data_ptr<IdType>();
-  // Fill in the first `num_seeds` ids.
-  torch::parallel_for(0, num_seeds, kGrainSize, [&](int64_t s, int64_t e) {
+  // Insert all ids into the hash map.
+  torch::parallel_for(0, num_ids, kGrainSize, [&](int64_t s, int64_t e) {
     for (int64_t i = s; i < e; i++) {
       InsertAndSetSmaller(ids_data[i], static_cast<IdType>(i));
     }
@@ -82,13 +82,6 @@ torch::Tensor ConcurrentIdHashMap<IdType>::Init(
 
   const int64_t num_threads = torch::get_num_threads();
   std::vector<size_t> block_offset(num_threads + 1, 0);
-  // Insert all elements in this loop.
-  torch::parallel_for(
-      num_seeds, num_ids, kGrainSize, [&](int64_t s, int64_t e) {
-        for (int64_t i = s; i < e; i++) {
-          InsertAndSetSmaller(ids_data[i], static_cast<IdType>(i));
-        }
-      });
 
   // Count the valid numbers in each thread.
   torch::parallel_for(
@@ -219,14 +212,11 @@ void ConcurrentIdHashMap<IdType>::InsertAndSetSmaller(IdType id, IdType value) {
     state = AttemptInsertAt(pos, id);
   }
 
-  if (state == InsertState::INSERTED) {
-    hash_map_data[getValueIndex(pos)] = value;
-    return;
-  }
-
+  IdType empty_key = static_cast<IdType>(kEmptyKey);
   IdType val_pos = getValueIndex(pos);
-  IdType old_val = hash_map_data[val_pos];
-  while (old_val > value) {
+  IdType old_val =
+      state == InsertState::INSERTED ? empty_key : hash_map_data[val_pos];
+  while (old_val == empty_key || old_val > value) {
     old_val = CompareAndSwap(&(hash_map_data[val_pos]), old_val, value);
   }
 
--- a/graphbolt/src/concurrent_id_hash_map.cc
+++ b/graphbolt/src/concurrent_id_hash_map.cc
@@ -68,7 +68,7 @@ torch::Tensor ConcurrentIdHashMap<IdType>::Init(
   // Insert all ids into the hash map.
   torch::parallel_for(0, num_ids, kGrainSize, [&](int64_t s, int64_t e) {
     for (int64_t i = s; i < e; i++) {
-      InsertAndSetSmaller(ids_data[i], static_cast<IdType>(i));
+      InsertAndSetMin(ids_data[i], static_cast<IdType>(i));
     }
   });
   // Place the first `num_seeds` ids.
@@ -203,7 +203,7 @@ inline void ConcurrentIdHashMap<IdType>::InsertAndSet(IdType id, IdType value) {
 }
 
 template <typename IdType>
-void ConcurrentIdHashMap<IdType>::InsertAndSetSmaller(IdType id, IdType value) {
+void ConcurrentIdHashMap<IdType>::InsertAndSetMin(IdType id, IdType value) {
   IdType pos = (id & mask_), delta = 1;
   IdType* hash_map_data = hash_map_.data_ptr<IdType>();
   InsertState state = AttemptInsertAt(pos, id);
@@ -216,10 +216,13 @@ void ConcurrentIdHashMap<IdType>::InsertAndSetSmaller(IdType id, IdType value) {
   IdType val_pos = getValueIndex(pos);
   IdType old_val = empty_key;
   while (old_val == empty_key || old_val > value) {
-    old_val = CompareAndSwap(&(hash_map_data[val_pos]), old_val, value);
+    IdType replaced_val =
+        CompareAndSwap(&(hash_map_data[val_pos]), old_val, value);
+    if (old_val == replaced_val)
+      break;
+    else
+      old_val = replaced_val;
   }
-
-  return;
 }
 
 template <typename IdType>
--- a/graphbolt/src/concurrent_id_hash_map.cc
+++ b/graphbolt/src/concurrent_id_hash_map.cc
@@ -214,8 +214,7 @@ void ConcurrentIdHashMap<IdType>::InsertAndSetSmaller(IdType id, IdType value) {
 
   IdType empty_key = static_cast<IdType>(kEmptyKey);
   IdType val_pos = getValueIndex(pos);
-  IdType old_val =
-      state == InsertState::INSERTED ? empty_key : hash_map_data[val_pos];
+  IdType old_val = empty_key;
   while (old_val == empty_key || old_val > value) {
     old_val = CompareAndSwap(&(hash_map_data[val_pos]), old_val, value);
   }
--- a/graphbolt/src/concurrent_id_hash_map.cc
+++ b/graphbolt/src/concurrent_id_hash_map.cc
@@ -218,10 +218,8 @@ void ConcurrentIdHashMap<IdType>::InsertAndSetMin(IdType id, IdType value) {
   while (old_val == empty_key || old_val > value) {
     IdType replaced_val =
         CompareAndSwap(&(hash_map_data[val_pos]), old_val, value);
-    if (old_val == replaced_val)
-      break;
-    else
-      old_val = replaced_val;
+    if (old_val == replaced_val) break;
+    old_val = replaced_val;
   }
 }
 
--- a/graphbolt/src/concurrent_id_hash_map.h
+++ b/graphbolt/src/concurrent_id_hash_map.h
@@ -159,6 +159,16 @@ class ConcurrentIdHashMap {
   inline void InsertAndSet(IdType key, IdType value);
 
   /**
+   * @brief Insert a key into the hash map. If the key exists, set the value
+   * with the smaller value.
+   *
+   * @param id The key to be inserted.
+   * @param value The value to be set for the `key`.
+   *
+   */
+  inline void InsertAndSetSmaller(IdType id, IdType value);
+
+  /**
    * @brief Attempt to insert the key into the hash map at the given position.
    *
    * @param pos The position in the hash map to be inserted at.
--- a/graphbolt/src/concurrent_id_hash_map.h
+++ b/graphbolt/src/concurrent_id_hash_map.h
@@ -166,7 +166,7 @@ class ConcurrentIdHashMap {
    * @param value The value to be set for the `key`.
    *
    */
-  inline void InsertAndSetSmaller(IdType id, IdType value);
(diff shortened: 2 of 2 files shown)


mfbalin:
LGTM, as per our diacussion with @peizhou001.

mfbalin:
How much slower is this new implementation compared to the old one?

RamonZhou:
> How much slower is this new implementation compared to the old one?

@mfbalin The time complexity should be the same. I also tested on my local machine, there's no significant slow down in any of `pinned-cuda`, `cpu-cuda` and `cuda-cuda` modes.

mfbalin:
> > How much slower is this new implementation compared to the old one?
> 
> @mfbalin The time complexity should be the same. I also tested on my local machine, there's no significant slow down in any of `pinned-cuda`, `cpu-cuda` and `cuda-cuda` modes.

@RamonZhou Could you also test the advanced pyg example with `cpu-pinned-cuda`? That one stresses the sampling and unique_and_compact the most.

graphbolt/src/concurrent_id_hash_map.cc
```diff
@@ -83,12 +83,22 @@ torch::Tensor ConcurrentIdHashMap<IdType>::Init(
   const int64_t num_threads = torch::get_num_threads();
   std::vector<size_t> block_offset(num_threads + 1, 0);
   // Insert all elements in this loop.
+  torch::parallel_for(
+      num_seeds, num_ids, kGrainSize, [&](int64_t s, int64_t e) {
+        for (int64_t i = s; i < e; i++) {
+          InsertAndSetSmaller(ids_data[i], static_cast<IdType>(i));
+        }
+      });
```
mfbalin, line 91: Can't we just insert all elements in the previous loop? Because we use the SetSmaller variant, the first num_seeds will be guaranteed to be inserted anyway.

mfbalin, line 91:
```suggestion
```

graphbolt/src/concurrent_id_hash_map.cc
```diff
@@ -68,7 +68,7 @@ torch::Tensor ConcurrentIdHashMap<IdType>::Init(
   // Fill in the first `num_seeds` ids.
   torch::parallel_for(0, num_seeds, kGrainSize, [&](int64_t s, int64_t e) {
```
mfbalin, line 69:
```suggestion
  torch::parallel_for(0, num_ids, kGrainSize, [&](int64_t s, int64_t e) {
```

graphbolt/src/concurrent_id_hash_map.cc
```diff
@@ -199,6 +209,30 @@ inline void ConcurrentIdHashMap<IdType>::InsertAndSet(IdType id, IdType value) {
...
+  IdType pos = (id & mask_), delta = 1;
+  IdType* hash_map_data = hash_map_.data_ptr<IdType>();
+  InsertState state = AttemptInsertAt(pos, id);
+  while (state == InsertState::OCCUPIED) {
+    Next(&pos, &delta);
+    state = AttemptInsertAt(pos, id);
+  }
+
+  if (state == InsertState::INSERTED) {
+    hash_map_data[getValueIndex(pos)] = value;
```
mfbalin, line 223: There is a data race here. We need to perform an atomic store operation here.

graphbolt/src/concurrent_id_hash_map.cc
```diff
@@ -199,6 +209,30 @@ inline void ConcurrentIdHashMap<IdType>::InsertAndSet(IdType id, IdType value) {
...
+    return;
+  }
+
+  IdType val_pos = getValueIndex(pos);
+  IdType old_val = hash_map_data[val_pos];
```
mfbalin, line 228: This needs to be an atomic load operation as well.

graphbolt/src/concurrent_id_hash_map.cc
mfbalin: Actually, all store operations need to be atomicMax operations so this if statement I think should be removed.

graphbolt/src/concurrent_id_hash_map.cc
mfbalin, line 228:
```suggestion
  IdType old_val = std::numeric_limits<IdType>::max();
```

mfbalin: We always need to access hash_map_data atomically, setting it to a max value ensures that it is initialized in the while loop below.

graphbolt/src/concurrent_id_hash_map.cc
mfbalin: Instead, we need to ensure that `empty_key == std::numeric_limits<IdType>::max()`;

graphbolt/src/concurrent_id_hash_map.cc
```diff
@@ -199,6 +209,30 @@ inline void ConcurrentIdHashMap<IdType>::InsertAndSet(IdType id, IdType value) {
...
+  while (old_val > value) {
```
mfbalin, line 229:
```suggestion
  while (old_val == kEmptyKey || old_val > value) {
```

graphbolt/src/concurrent_id_hash_map.h
```diff
@@ -158,6 +158,16 @@ class ConcurrentIdHashMap {
...
 
+  /**
+   * @brief Insert a key into the hash map. If the key exists, set the value
+   * with the smaller value.
+   *
+   * @param id The key to be inserted.
+   * @param value The value to be set for the `key`.
+   *
+   */
+  inline void InsertAndSetSmaller(IdType id, IdType value);
```
frozenbugs, line 169: Do we still need InsertAndSet after this work? If not, why not user replace?

RamonZhou:
> > > How much slower is this new implementation compared to the old one?
> > 
> > 
> > @mfbalin The time complexity should be the same. I also tested on my local machine, there's no significant slow down in any of `pinned-cuda`, `cpu-cuda` and `cuda-cuda` modes.
> 
> @RamonZhou Could you also test the advanced pyg example with `cpu-pinned-cuda`? That one stresses the CPU sampling and unique_and_compact the most.

@mfbalin  I tested it and the average epoch time is 7.94s before and 7.99s after

mfbalin:
> > > > How much slower is this new implementation compared to the old one?
> > > 
> > > 
> > > @mfbalin The time complexity should be the same. I also tested on my local machine, there's no significant slow down in any of `pinned-cuda`, `cpu-cuda` and `cuda-cuda` modes.
> > 
> > 
> > @RamonZhou Could you also test the advanced pyg example with `cpu-pinned-cuda`? That one stresses the CPU sampling and unique_and_compact the most.
> 
> @mfbalin I tested it and the average epoch time is 7.94s before and 7.99s after

Thank you. Let's test it again before we merge it. The changes during the review process may impact performance. I will check if there is a faster atomicmin operation available in gcc.

graphbolt/src/concurrent_id_hash_map.cc
RamonZhou: Store operations should be atomic but why does read operation also need to be atomic?

frozenbugs:
> > > > How much slower is this new implementation compared to the old one?
> > > 
> > > 
> > > @mfbalin The time complexity should be the same. I also tested on my local machine, there's no significant slow down in any of `pinned-cuda`, `cpu-cuda` and `cuda-cuda` modes.
> > 
> > 
> > @RamonZhou Could you also test the advanced pyg example with `cpu-pinned-cuda`? That one stresses the CPU sampling and unique_and_compact the most.
> 
> @mfbalin I tested it and the average epoch time is 7.94s before and 7.99s after

we may also add an option to decide which path to go depends on whether user need deterministic or not.

graphbolt/src/concurrent_id_hash_map.cc
mfbalin: All operations need to be atomic if you modify a memory location atomically. Otherwise, there will be a data race.

RamonZhou: Yes sorry I got it wrong. They should be atomic. But I think `hash_map_data[pos]` is atomic itself?

mfbalin: It is a plain pointer, it is not an atomic pointer. If we want the code to work on multiple platforms, we can't access the pointer directly like that.

graphbolt/src/concurrent_id_hash_map.cc
```diff
@@ -199,6 +202,27 @@ inline void ConcurrentIdHashMap<IdType>::InsertAndSet(IdType id, IdType value) {
...
+  InsertState state = AttemptInsertAt(pos, id);
+  while (state == InsertState::OCCUPIED) {
+    Next(&pos, &delta);
+    state = AttemptInsertAt(pos, id);
+  }
+
+  IdType empty_key = static_cast<IdType>(kEmptyKey);
+  IdType val_pos = getValueIndex(pos);
+  IdType old_val =
+      state == InsertState::INSERTED ? empty_key : hash_map_data[val_pos];
```
mfbalin, line 218:
```suggestion
  IdType old_val = empty_key;
```

mfbalin: We can't access hash_map_data in a non-atomic way.

graphbolt/src/concurrent_id_hash_map.h
RamonZhou: It's different from `InsertAndSet` so I created another function. Now `InsertAndSet` is not used. We can delete it or save it for possible future use because it's a common operation for a hash map.

graphbolt/src/concurrent_id_hash_map.cc
RamonZhou: Ok I see

peizhou001:
> > > > > How much slower is this new implementation compared to the old one?
> > > > 
> > > > 
> > > > @mfbalin The time complexity should be the same. I also tested on my local machine, there's no significant slow down in any of `pinned-cuda`, `cpu-cuda` and `cuda-cuda` modes.
> > > 
> > > 
> > > @RamonZhou Could you also test the advanced pyg example with `cpu-pinned-cuda`? That one stresses the CPU sampling and unique_and_compact the most.
> > 
> > 
> > @mfbalin I tested it and the average epoch time is 7.94s before and 7.99s after
> 
> we may also add an option to decide which path to go depends on whether user need deterministic or not.

Maybe not? because it keeps the same speed while making the result sable, there is no reason why user want the result non-deterministic w/o performance benefits.

mfbalin (approved):
LGTM. If the performance is same as before or even better (we insert all in a single parallel for now, should be a little faster), we can go ahead and merge it.

graphbolt/src/concurrent_id_hash_map.cc
```diff
@@ -199,6 +202,26 @@ inline void ConcurrentIdHashMap<IdType>::InsertAndSet(IdType id, IdType value) {
...
+  }
+
+  IdType empty_key = static_cast<IdType>(kEmptyKey);
+  IdType val_pos = getValueIndex(pos);
+  IdType old_val = empty_key;
+  while (old_val == empty_key || old_val > value) {
+    old_val = CompareAndSwap(&(hash_map_data[val_pos]), old_val, value);
+  }
+
+  return;
```
mfbalin, line 222: nit:
```suggestion
```

mfbalin:
> > > > > > How much slower is this new implementation compared to the old one?
> > > > > 
> > > > > 
> > > > > @mfbalin The time complexity should be the same. I also tested on my local machine, there's no significant slow down in any of `pinned-cuda`, `cpu-cuda` and `cuda-cuda` modes.
> > > > 
> > > > 
> > > > @RamonZhou Could you also test the advanced pyg example with `cpu-pinned-cuda`? That one stresses the CPU sampling and unique_and_compact the most.
> > > 
> > > 
> > > @mfbalin I tested it and the average epoch time is 7.94s before and 7.99s after
> > 
> > 
> > we may also add an option to decide which path to go depends on whether user need deterministic or not.
> 
> Maybe not? because it keeps the same speed while making the result sable, there is no reason why user want the result non-deterministic w/o performance benefits.

I see around 5% performance loss with this PR on my machine.

frozenbugs (approved):

mfbalin:
@RamonZhou it might be best to measure the performance by making the feature dimension smaller in case the feature fetch is the bottleneck on your system. Like `features = features[:, 32]`.

graphbolt/src/concurrent_id_hash_map.cc
```diff
@@ -65,10 +65,10 @@ torch::Tensor ConcurrentIdHashMap<IdType>::Init(
   // This code block is to fill the ids into hash_map_.
   auto unique_ids = torch::empty_like(ids);
   IdType* unique_ids_data = unique_ids.data_ptr<IdType>();
-  // Fill in the first `num_seeds` ids.
-  torch::parallel_for(0, num_seeds, kGrainSize, [&](int64_t s, int64_t e) {
+  // Insert all ids into the hash map.
+  torch::parallel_for(0, num_ids, kGrainSize, [&](int64_t s, int64_t e) {
     for (int64_t i = s; i < e; i++) {
-      InsertAndSet(ids_data[i], static_cast<IdType>(i));
+      InsertAndSetSmaller(ids_data[i], static_cast<IdType>(i));
```
peizhou001, line 71: A little confused with the name, could you change to a more meaningful name?

peizhou001 (approved):

RamonZhou:
@mfbalin I tested again with the latest master branch merged. It's 7.911s (before) vs. 7.919s (after)

graphbolt/src/concurrent_id_hash_map.cc
RamonZhou: How about `InsertAndSetIfSmaller` or `SetIfSmaller`

frozenbugs: InsertAndSetMin is clearer to me, this loop is finding the minimum value, right?
```
  while (old_val == empty_key || old_val > value) {
    old_val = CompareAndSwap(&(hash_map_data[val_pos]), old_val, value);
  }
```

graphbolt/src/concurrent_id_hash_map.cc
```diff
@@ -199,6 +202,29 @@ inline void ConcurrentIdHashMap<IdType>::InsertAndSet(IdType id, IdType value) {
...
+  IdType empty_key = static_cast<IdType>(kEmptyKey);
+  IdType val_pos = getValueIndex(pos);
+  IdType old_val = empty_key;
+  while (old_val == empty_key || old_val > value) {
+    IdType replaced_val =
+        CompareAndSwap(&(hash_map_data[val_pos]), old_val, value);
+    if (old_val == replaced_val)
+      break;
+    else
+      old_val = replaced_val;
```
mfbalin, line 224: I would take the changes here back, there is no reason to make it more complicated. The performance won't change either.
```suggestion
    old_val = CompareAndSwap(&(hash_map_data[val_pos]), old_val, value);
```

RamonZhou: But in the old code, after a successful `CompareAndSwap`, the loop won't stop, it will do an extra `CompareAndSwap`, which could be slower

mfbalin: Makes sense, good catch! How much faster is this new implementation? We can do this because CompareAndSwap is the https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange strong variant, right? It is guaranteed to succeed if there is no contention.

frozenbugs, line 224:
```suggestion
    if (old_val == replaced_val) break;
    old_val = replaced_val;
```

graphbolt/src/concurrent_id_hash_map.cc
```diff
@@ -199,6 +202,29 @@ inline void ConcurrentIdHashMap<IdType>::InsertAndSet(IdType id, IdType value) {
...
+  IdType* hash_map_data = hash_map_.data_ptr<IdType>();
+  InsertState state = AttemptInsertAt(pos, id);
+  while (state == InsertState::OCCUPIED) {
+    Next(&pos, &delta);
+    state = AttemptInsertAt(pos, id);
+  }
+
+  IdType empty_key = static_cast<IdType>(kEmptyKey);
+  IdType val_pos = getValueIndex(pos);
+  IdType old_val = empty_key;
```
mfbalin, line 217: We might want to perform an atomic load operation here to initialize with `hash_map_data[val_pos]` instead of empty key. We will need to implement atomic load similar to how we implemented CompareAndSwap for that though. Can be implemented in a future PR.

RamonZhou:
I did another test and the results are:
- Old version: 5.817s
- Current PR (with the `if` inside the loop): 5.802s
- Current PR (without the `if` inside the loop): 5.783s

So it seems that deleting the `if` might be faster!

mfbalin:
> I did another test and the results are (features are sliced with `[:, :32]`):
> 
> * Old version: 5.817s
> * Current PR (with the `if` inside the loop): 5.802s
> * Current PR (without the `if` inside the loop): 5.783s
> 
> So it seems that deleting the `if` might be faster!

The fact that all the numbers are so close is suspicious to me. Could we test with no feature fetch and no forward pass?

RamonZhou:
> > I did another test and the results are (features are sliced with `[:, :32]`):
> > 
> > * Old version: 5.817s
> > * Current PR (with the `if` inside the loop): 5.802s
> > * Current PR (without the `if` inside the loop): 5.783s
> > 
> > So it seems that deleting the `if` might be faster!
> 
> The fact that all the numbers are so close is suspicious to me. Could we test with no feature fetch and no forward pass?

Without feature fetching and forward pass: 4.319s (before) and 4.707s (after). About 9% slower

mfbalin:
> > > I did another test and the results are (features are sliced with `[:, :32]`):
> > > 
> > > * Old version: 5.817s
> > > * Current PR (with the `if` inside the loop): 5.802s
> > > * Current PR (without the `if` inside the loop): 5.783s
> > > 
> > > So it seems that deleting the `if` might be faster!
> > 
> > 
> > The fact that all the numbers are so close is suspicious to me. Could we test with no feature fetch and no forward pass?
> 
> Without feature fetching and forward pass: 4.319s (before) and 4.707s (after). About 9% slower

Do you want to look into implementinf atomic load and see if it helps before the cas loop? Or do you want to merge this first, then continue to optimize?

RamonZhou:
> > > > I did another test and the results are (features are sliced with `[:, :32]`):
> > > > 
> > > > * Old version: 5.817s
> > > > * Current PR (with the `if` inside the loop): 5.802s
> > > > * Current PR (without the `if` inside the loop): 5.783s
> > > > 
> > > > So it seems that deleting the `if` might be faster!
> > > 
> > > 
> > > The fact that all the numbers are so close is suspicious to me. Could we test with no feature fetch and no forward pass?
> > 
> > 
> > Without feature fetching and forward pass: 4.319s (before) and 4.707s (after). About 9% slower
> 
> Do you want to look into implementing atomic load and see if it helps before the cas loop? Or do you want to merge this first, then continue to optimize?

@mfbalin I tried to implement it but I think the atomic load can just be a `CompareAndSwap`. Like `old_val = CompareAndSwap(&(hash_map_data[val_pos]), empty_key, value);` is a way to load it atomically. But in the end it's the same with the current code since we are doing the same thing in the loop.

mfbalin:
> > > > > I did another test and the results are (features are sliced with `[:, :32]`):
> > > > > 
> > > > > * Old version: 5.817s
> > > > > * Current PR (with the `if` inside the loop): 5.802s
> > > > > * Current PR (without the `if` inside the loop): 5.783s
> > > > > 
> > > > > So it seems that deleting the `if` might be faster!
> > > > 
> > > > 
> > > > The fact that all the numbers are so close is suspicious to me. Could we test with no feature fetch and no forward pass?
> > > 
> > > 
> > > Without feature fetching and forward pass: 4.319s (before) and 4.707s (after). About 9% slower
> > 
> > 
> > Do you want to look into implementing atomic load and see if it helps before the cas loop? Or do you want to merge this first, then continue to optimize?
> 
> @mfbalin I tried to implement it but I think the atomic load can just be a `CompareAndSwap`. Like `old_val = CompareAndSwap(&(hash_map_data[val_pos]), empty_key, value);` is a way to load it atomically. But in the end it's the same with the current code since we are doing the same thing in the loop.

https://gcc.gnu.org/onlinedocs/gcc-4.1.0/gcc/Atomic-Builtins.html
You can use __sync_fetch_and_or with 2nd argument as 0 to perform an atomic load operation.

RamonZhou:
> > > > > > I did another test and the results are (features are sliced with `[:, :32]`):
> > > > > > 
> > > > > > * Old version: 5.817s
> > > > > > * Current PR (with the `if` inside the loop): 5.802s
> > > > > > * Current PR (without the `if` inside the loop): 5.783s
> > > > > > 
> > > > > > So it seems that deleting the `if` might be faster!
> > > > > 
> > > > > 
> > > > > The fact that all the numbers are so close is suspicious to me. Could we test with no feature fetch and no forward pass?
> > > > 
> > > > 
> > > > Without feature fetching and forward pass: 4.319s (before) and 4.707s (after). About 9% slower
> > > 
> > > 
> > > Do you want to look into implementing atomic load and see if it helps before the cas loop? Or do you want to merge this first, then continue to optimize?
> > 
> > 
> > @mfbalin I tried to implement it but I think the atomic load can just be a `CompareAndSwap`. Like `old_val = CompareAndSwap(&(hash_map_data[val_pos]), empty_key, value);` is a way to load it atomically. But in the end it's the same with the current code since we are doing the same thing in the loop.
> 
> https://gcc.gnu.org/onlinedocs/gcc-4.1.0/gcc/Atomic-Builtins.html You can use __sync_fetch_and_or with 2nd argument as 0 to perform an atomic load operation.

I see. I tried `__sync_fetch_and_or` and tested. The performance is about the same (4.717s)

RamonZhou:
Let's merge this first and optimize in the future.

PR #7211, opened 2024-03-13
[DO NOT MERGE] Add `compute_mrr_seeds`.

yxy235:
## Description

Add `compute_mrr_seeds` to link_prediction examples. This function will be used after the modification of the datasets.
The time is longer than original due to len(dataloader_seeds) > len(dataloader_original) as neg_dst are flattend in new dataset.
Also `RetrievalMRR()` provided by torchmetrics will cost longer time compared with original implementation.

### original:
node_pairs: [[1,2], [3,4], [5,6], [7,8]]
neg_dst: [[1,2,3,4,5], [11,12,13,14,15], [21,22,23,24,25], [31,32,33,34,35]]
### seeds:
seeds: [[1,2], [3,4], [5,6], [7,8], [1,1], [1,2], [1,3], [1,4], [1,5], [3,11], … , [7,34], [7,35]]
labels: [1, 1, 1, 1, 0, 0, … , 0, 0]
indexes: [0, 1, 2, 3, 0, 0, 0, 0, 0, 1, 1, … , 3, 3]

## Performance
| | original | seeds|
|---|---|---|
Training | 17.96 it/s ; 55:05 min | 18.45 it/s ; 53:37 min
Node embedding inference | 223.89 / 212.75 / 215.00 it/s ; 12 /13 /13 (s) | 209.67 / 197.59 / 195.86 it/s ; 13 / 14 / 14 (s)
evaluation | 37 s | 66 s
validation MRR | 0.7983 | 0.7949
test MRR | 0.7987 | 0.7943

## Checklist
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [$CATEGORY] (such as [NN], [Model], [Doc], [Feature]])
- [ ] I've leverage the [tools](https://docs.google.com/document/d/1iHyj7zlmygKSk5gBPsqIqL5ASPzJSPREaNT_QdsiYA4/edit) to beautify the python and c++ code.
- [ ] The PR is complete and small, read the [Google eng practice (CL equals to PR)](https://google.github.io/eng-practices/review/developer/small-cls.html) to understand more about small PR. In DGL, we consider PRs with less than 200 lines of core code change are small (example, test and documentation could be exempted).
- [ ] All changes have test coverage
- [ ] Code is well-documented
- [ ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change
- [ ] Related issue is referred in this PR
- [ ] If the PR is for a new model/paper, I've updated the example index [here](../examples/README.md).

PR #7231, opened 2024-03-19
[GraphBolt] Modify examples to use `seeds`.

yxy235:
## Description

There are major changes to `link_prediction`， you can see more details in #7211 .

Something still wait to be addressed:
1. dataset:
- [x] ogb-lsc-mag240m
2. examples:
- [x] lightning/node_classification.py
- [x] sparse/graphsage.py

## Checklist
- [ ] Code is well-documented


examples/sampling/graphbolt/node_classification.py
```diff
@@ -312,6 +312,7 @@ def train(args, graph, features, train_set, valid_set, num_classes, model):
             y = data.labels
 
             y_hat = model(data.blocks, x)
+            y_hat = y_hat[data.compacted_seeds]
```
frozenbugs, line 315: Why do we need to add this?

yxy235: When `seeds` go through `unique_and_compact` on cuda mode, it will be sorted in ascending order which may be different from original order and may mismatch with `labels`. So, we need to use `compacted_seeds` to restore it to its original order.

frozenbugs: Does this caused by your refactor? Or it was an existing bug?

frozenbugs (approved):

examples/sampling/graphbolt/rgcn/hetero_rgcn.py
```diff
@@ -578,7 +578,7 @@ def main(args):
     # `institution` are generated in advance and stored in the feature store.
     # For `ogbn-mag`, we generate the features on the fly.
     embed_layer = None
-    if args.dataset == "ogbn-mag":
+    if args.dataset == "ogbn-mag-seeds":
```
Rhett-Ying, line 581: This looks weird to me. User should not care about `seeds` suffix at all. the dataset name exposed to user should always be `ogbn-mag`. In the near future, we will fully deprecate existing`BuiltinDataset(ogbn-mag)`, just replace it with `ogbn-mag-seeds`.

Rhett-Ying: and README is not updated accordingly.

PR #7240, opened 2024-03-24
[CUDA] Make sanity check optional for `dgl.create_block`.

mfbalin:
## Description
The hetero sampling case has too many GPU synchronizations, we need to hunt them down one by one if we want to get good performance.

I would appreciate any help in this direction as some of these are inside DGL itself, not GraphBolt.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/utils/data.py
+++ b/python/dgl/utils/data.py
@@ -191,39 +191,28 @@ def graphdata2tensors(
                 data.format, tuple(F.tensor(a) for a in data.arrays)
             )
 
+    num_src, num_dst = None, None
     if isinstance(data, SparseAdjTuple):
         if idtype is not None:
             data = SparseAdjTuple(
                 data.format, tuple(F.astype(a, idtype) for a in data.arrays)
             )
-        num_src, num_dst = (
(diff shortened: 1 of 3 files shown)


python/dgl/utils/data.py
```diff
@@ -116,7 +116,9 @@ def networkx2tensor(nx_graph, idtype, edge_id_attr_name=None):
 SparseAdjTuple = namedtuple("SparseAdjTuple", ["format", "arrays"])
 
 
-def graphdata2tensors(data, idtype=None, bipartite=False, **kwargs):
+def graphdata2tensors(
```
frozenbugs, line 119: Is there any test coverage?

frozenbugs (approved):

python/dgl/utils/data.py
mfbalin: There is no specific test for this method. The added arguments make it so that the sanity check is enabled by default.

mfbalin: Feel free to merge if you think this is OK. I profiled the code and checked whether the GPU synchronizations are eliminated for the GraphBolt rgcn example.

PR #7245, opened 2024-03-27
[GraphBolt] Fix `to` when using `seeds`.

yxy235:
## Description

In examples, we will use `copy_to` after `ItemSampler` and `fetch_feature`.
When using `copy_to` after `ItemSampler`, we will copy all attributes because at this time, only `seed_nodes`, `node_pairs` and `labels` have value in minibatch.
When using `copy_to` after `fetch_feature`, we will only copy attributes we need.
These two situations are selected by checking `seed_nodes` and `compacted_pairs` to determine the current stage and the task. However, when using `seeds`, it is difficult to determine these through `seeds` and `compacted_seeds`. For example, when running evaluation in quickstart/link_prediction, `seeds` will not be transfered due to existing `labels`, which may cause error. So, I have changed the `transfer_attr` to confirm `seeds` will be transfered every situation to avoid such error.

## Checklist
- [ ] Code is well-documented

### DIFF
--- a/python/dgl/graphbolt/minibatch.py
+++ b/python/dgl/graphbolt/minibatch.py
@@ -591,15 +591,10 @@ class MiniBatch:
                 "sampled_subgraphs",
                 "node_features",
                 "edge_features",
+                "compacted_seeds",
+                "indexes",
+                "seeds",
             ]
-            # Link/edge related tasks.
-            if self.compacted_seeds is not None:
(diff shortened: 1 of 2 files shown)


mfbalin:
copy_to can be between sample neighbors and fetch feature as well.

frozenbugs (approved):

